Issue #2, opened 2015-10-24
Code Conventions


schauder:
It would be nice to have some code conventions defined, to make it easier for non-core-committers to provide useful pull requests.

Ideally they would come with formatters/checkstyle configurations.

sbrannen:
Don't worry.

We will **definitely** having coding conventions in place once we get to the point where we are accepting pull requests.

At this time we are simply focusing on the functionality and feasibility of the prototype.

Since most of the code will likely be rewritten or thrown away, coding conventions are a moot point for the time being. :wink:

kcooney:
JUnit moved to.the Google Java Style Guide, except we kept the indentation of existing files at four spaces.

See https://github.com/junit-team/junit/blob/master/CODING_STYLE.txt

kcooney:
FYI: there is a checkstyle config for the Google Java Style Guide: http://checkstyle.sourceforge.net/google_style.html

See also https://agilewarrior.wordpress.com/2015/08/07/how-to-add-intellij-google-style-guide-to-intellij/ and https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml

nedtwigg:
Whitespace is certainly not the most important part of a project, but if you put stylechecking into the build while the project is still young, then it ensures that your project history from will be clean of formatting noise so you can focus on the content.  I submitted two PR's above:
- the first enforces a style guide, using rules that closely match the exising style
- the second is the same as the first, but with an extra commit to switch to the Google style

nedtwigg:
Saw this change in the README:

> The goal of the prototype is to get feedback on the API. Focusing on code style, formatting and other details will distract the community's (and our) attention.

I agree that formatting is a minor concern, so I won't distract further after this comment.  But the pull-requests I submitted to help with formatting now have a merge conflict because of a [meaningless whitespace change](https://github.com/junit-team/junit-lambda/commit/cbf97077bae0db71dc3a7cade28c3b00587a8699).

As someone who doesn't care what style the code is in, but just wants to follow the evolution of the code, it's distracting to filter through formatting changes.  The commit which introduced conflicts for the PRs also swapped the order of imports in several places, without actually changing what was imported.

> Be regular and orderly in your life, so that you may be violent and original in your work.
>     - Gustave Flaubert

[...]

PR #3, opened 2015-10-26
Enforce style rules using Spotless

nedtwigg:
Added [Spotless](https://github.com/diffplug/spotless), a Gradle plugin that checks and enforces style rules, to help with #2.  There are lots of tools that can alert you to files that don't match the style rules.  Spotless is unique in that it can also automatically apply them.  This makes it really easy for contributors to format their work correctly, while also ensuring that the project history isn't polluted by dirty whitespace.

`spotlessCheck` is added as a dependency to `build`, and it alerts the user if any files don't meet the formatting rules.  If there are any formatting errors, `spotlessApply` will fix them.  Includes rules for Java source code and project metadata.

For this PR, I found rules that closely matched the existing format.  In a follow-up PR, I'll apply the Google style rules (personally I dislike the Google style, but so long as a project provides a formatter such as Spotless, it doesn't matter what the project uses, because I can always convert back and forth when submitting commits).

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -9,7 +9,7 @@ THE ACCOMPANYING PROGRAM IS PROVIDED UNDER THE TERMS OF THIS ECLIPSE PUBLIC LICE
 * **a)** in the case of the initial Contributor, the initial code and documentation distributed under this Agreement, and
 * **b)** in the case of each subsequent Contributor:
 	* **i)** changes to the Program, and
-	* **ii)** additions to the Program; 
+	* **ii)** additions to the Program;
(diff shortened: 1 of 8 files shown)


sbrannen:
@nedtwigg, thanks so much for making us aware of Spotless. The plugin seems very useful, and it's so easy to set up that I've gone ahead and implemented it myself. The formatter settings are obviously not yet set in stone, but the status quo (i.e., my latest commits) is much better than having inconsistent formatting. :wink:

nedtwigg:
Glad it was useful :+1:

PR #4, opened 2015-10-26
Switch to Google style

nedtwigg:
Also for Issue #2, this is the same as PR #3, but with an extra commit which uses the Google style rules.  Personally I'm not a huge fan of the Google styleguide, but to each their own :)


sbrannen:
Closing... see issue #3 for details.

PR #7, opened 2015-11-05
Simplified specification resolution

jlink:
OK, guys. I've been refactoring the whole spec resolution part.
Doing that 
- I introduced a new little type hierarchy: JUnit5Testable|Method|Class which allowed me to get rid of the resolvers and resolver registry. 
- The JUnit5TestableFactory contains all logic for creating uniqueIDs and backwards
- In doing the above I introduced an AbstractTestDescriptor which has the possibility to add a child. addChild and getChildren is not exposed in TestDescriptor so nothing has changed from the perspective of the launcher. Having the hierarchy greatly simplified the code for resolving from TestSpecificationElement to TestDescriptor, so bear with me.

The current implementation can deal with both nested test classed (each level becomes a containing test descriptor) and test methods with params (even if parameter injection is still missing).


junit-engine-api/src/main/java/org/junit/gen5/engine/AbstractTestDescriptor.java
```diff
@@ -0,0 +1,64 @@
...
+		if (other == null)
+			return false;
+		if (this.getClass() != other.getClass())
+			return false;
+		TestDescriptor otherDescriptor = (TestDescriptor) other;
+		return this.getUniqueId().equals(otherDescriptor.getUniqueId());
+	}
+
+	@Override
+	public int hashCode() {
```
bechte: I didn't want to write anything about this, but I just can't stop myself, sorry: Even if it complies to Java's equals/hashCode, I think we could use this.getUniqueId().hashCode() which should return a more useful hash. ;-) With 42, we wouldn't be able to use the Descriptors in a HashSet/HashMap/etc. I think the hash of the uniqueId is a reasonable fallback, if subclasses do not implement hashCode() their own.

junit-launcher/src/main/java/org/junit/gen5/launcher/Launcher.java
```diff
@@ -38,7 +38,7 @@ public void registerTestPlanExecutionListeners(TestExecutionListener... testList
 	public TestPlan discover(TestPlanSpecification specification) {
 		TestPlan testPlan = new TestPlan();
 		for (TestEngine testEngine : lookupAllTestEngines()) {
-			TestDescriptor engineDescriptor = new EngineDescriptor(testEngine);
+			EngineDescriptor engineDescriptor = new EngineDescriptor(testEngine);
```
bechte: .oO( I still dislike the creation of a EngineDescriptor inside the Launcher )

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
```diff
@@ -0,0 +1,175 @@
...
+import org.junit.gen5.commons.util.ReflectionUtils;
+
+public class JUnit5TestableFactory {
+
+	private static final String SEPARATORS = ":$#";
+
+	public JUnit5Testable fromUniqueId(String uniqueId, String engineId) {
+		Preconditions.notEmpty(uniqueId, "uniqueId must not be empty");
+		List<String> parts = split(uniqueId);
+		Preconditions.condition(parts.remove(0).equals(engineId), "uniqueId must start with engineId");
```
bechte: As this precondition is absolutely necessary, I support the idea of encapsulating the whole concept of engines from the Launcher, such that the launcher does not handle engine IDs separately. For a given unique ID, the Launcher should simply call fromUniqueId(String uniqueId), which having to extract the engineId.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
```diff
@@ -0,0 +1,175 @@
...
+import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.junit.gen5.commons.util.Preconditions;
+import org.junit.gen5.commons.util.ReflectionUtils;
+
+public class JUnit5TestableFactory {
```
bechte: I like the idea of having this factory. :+1: 

I wonder if we could generalize this idea such that each engine must support some kind of factory for all the configuration options that we allow in the Launcher (className, classes, packages, packageNames, folders, files, uniqueIds, etc.).

I would love to have a registry that extension providers can register themselves at, so that they can contribute within the "factory" process. They might want to intercept in order to decorate or to replace the test descriptor.

But it's probably a "next step"... Thanks for preparing.

bechte:
Hi Johannes

I like the refactoring. It provides a lot of progress. Thanks for that.

We were also thinking about introducing a JUnit5Testable interface and your approach is close to what I did in the http://github.com/bechte/JUT case. I think it is a good direction but we still have to keep in mind, that there must be some kind of registration to allow extension providers to contribute to the whole process. This is probably engine specific, but we will see later.

For me, we could merge this pull request into master and continue our work from there. I made some comments within the code where I think we should investigate some more time. I could do so on the weekend, would that be fine with you?

Best
Stefan

jlink:
If you all agree with Stefan, I could merge the pull request today with a
few changes suggested by Stefan and adding appropriate todo comments to
reevaluate the decisions under debate later.

Johannes

2015-11-06 9:07 GMT+01:00 Stefan Bechtold <notifications@github.com>:

sbrannen:
I don't have time to review the PR at the moment. So I will defer to your collective judgment.

cheers,

Sam

mmerdes:
without being able to review the details, 
introducing factories and especially explicit tree children
is a step in the right direction for me.

junit-engine-api/src/main/java/org/junit/gen5/engine/AbstractTestDescriptor.java
jlink: With 42 you'd be able to use it in a HashSet etc. because all hashing collections fall back on equals in case of collition. Anyway, I changed it as you suggested.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
jlink: I don't quite understand. Do you mean that the engine ID should not leak into the specific engine at all? Then I'm with you. If you mean something else I need to see the exact code. Maybe a pull request ;)

bechte: Nope, I don't mean that. What I was saying is, that the engine id was introduced to guarantee, that TestDescriptors of the same kind are still unique, even if two engines produce a TestDescriptor with the identical local or engine-specific unique id. Still, the engine id is not a concept that any other class than the engine itself should work with.

Therefore, in a very first step, the resolver should be able to resolve the correct engine by using the first part as the engine's id and look it up in the "application context" or an engine registry.

jlink: Why should an engine even know about other engines' IDs or unique IDs aimed at other engines?

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
bechte: see parts of this idea within https://github.com/junit-team/junit-lambda/pull/8

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
bechte: I don't understand your question. Engines should not know about each others, at all.

Still the engine ID is part of the unique ID of a descriptor (at least for me it is a must-have). With this in place, we easily guarantee that descriptors created by different engines will never conflict each other. If we had unique ids without the engine id in it, we cannot guarantee uniqueness over several engines out-of-the-box. For me, the engine must be responsibility for uniqueness...

Please see my pull request.

jlink: Only the launcher can guarantee uniqueness since it sees all ids, an engine sees only its own. But again, let's stop working in this corner until we can discuss it with the rest.

Von meinem iPad gesendet

bechte: Just to make my point: The Launcher should not influence the resolution of the unique id for an engine at all. How would it guarantee, that subsequent test runs will always have the same ids?

One example: Let's say we have two engines that have a conflicting local/engine-specific unique id (other cases dont lead to problems anyways). What could a Launcher do to make them unique? The only feasible solution is to add the engine id in front of the local/engine-specific id. If there is still a conflict, the Launcher must give up and throw an exception telling that the engines are not compatible. Why?

If the Launcher influences the unique id by some kind of "magic" due to the given engines on the classpath, the returned full unique id would not be stable. It might be, that a subsequent call to the Launcher has only one engine on the classpath leading to another mapping/building of the full unique id.

So I definitely go for the solution, that the engine must guarantee the globally uniqueness of its descriptors, because it just doesnt make sense to move this responsibility somewhere else. The launcher only has to prove, that the ids are unique, and if it detects a conflict, it must throw some kind of exception complaining about the incompatibility of the engines.

But yes, we (you and me) have different ideas about this. So we both made our points, let's have someone else to decide on it...

jlink: > One example: Let's say we have two engines that have a conflicting local/engine-specific unique id (other cases dont lead to problems anyways). What could a Launcher do to make them unique? The only feasible solution is to add the engine id in front of the local/engine-specific id. If there is still a conflict, the Launcher must give up and throw an exception telling that the engines are not compatible. Why?
> 

If the launcher prefixes a "locally unique ID" with a _unique_ engine ID, how could the resulting ID be NOT globally unique?

bechte: If it has a rather simple algorithm that only asks the engine for its id and we have two engines that have the same id? If we throw an exception during startup, that avoids this situation, then still: why should the Launcher be responsible for adding a engine's id? what are the benefits? Why shouldnt the engine itself be responsible for it?

If it has a more complex algorithm, that takes into account the engines loaded in the current run, than the unique id depends on all engines on the classpath and, therefore, it is not stable.

jlink: Yes, engine IDs must be unique. If there are two engines with same ID on the classpath you have a problem. Most likely with Gradle/Maven config. 
And now I'll really stop ;)

junit-engine-api/src/main/java/org/junit/gen5/engine/AbstractTestDescriptor.java
kcooney: Yea, hashing collections fall back to equals, but 1) that could be inefficient, and 2) other code using hashCode() could be even less efficient with many collisions.

jlink: My point: Without a known usage pattern nobody can tell what efficiency means. Calculating a hash code also takes time. If you're not reasonably certain that it will prevent collisions in most cases, calculating the hash at all is inefficient.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
kcooney: If you are looking for a tie breaker...

I haven't looked at all the code, but I am leaning towards the position that the launcher should add the engine ID to the unique ID generated by the engine (or in an object that encapsulates the engine-generated ID and the engine ID). Another option would be that some code external to the engine (possibly the launcher) checks that the IDs generated by the engine are prefixed with the engine ID, but that leaks implementation details into the engine (why should the engine care how we make sure that IDs generated by multiple engines never conflict, or be forced to use a naming convention?)

Either way, if we want third-parties to add engines, then the engine ID would need to be a full class name, so the full IDs will be quite long. We should discuss how to make sure that most output that contains test IDs are easy for the user to parse (at least for the majority of users who never have a single test run containing tests from more that two engines). If there is a pass where all tests are enemerated before they are run, JUnit could see if there are conflicts ignoring engine ID, and if not only use the engine-generated ID in any stack traces. XML output would of course need either the full ID or the pieces necessary to construct it.

junit-engine-api/src/main/java/org/junit/gen5/engine/AbstractTestDescriptor.java
kcooney: If hashing is expensive and the use case for adding the value to a set (or making it a key in a map) are unclear, then the Javadoc could explain that the class isn't designed to be a keg i a map.

In any case, glad we could a solution.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
jlink: Nice to have you in the discussion, Kevin!

> I haven't looked at all the code, but I am leaning towards the position that the launcher should add the engine ID to the unique ID generated by the engine (or in an object that encapsulates the engine-generated ID and the engine ID). Another option would be that some code external to the engine (possibly the launcher) checks that the IDs generated by the engine are prefixed with the engine ID, but that leaks implementation details into the engine (why should the engine care how we make sure that IDs generated by multiple engines never conflict, or be forced to use a naming convention?)
> 
That's my standpoint exactly. 

> Either way, if we want third-parties to add engines, then the engine ID would need to be a full class name, so the full IDs will be quite long. We should discuss how to make sure that most output that contains test IDs are easy for the user to parse (at least for the majority of users who never have a single test run containing tests from more that two engines). If there is a pass where all tests are enemerated before they are run, JUnit could see if there are conflicts ignoring engine ID, and if not only use the engine-generated ID in any stack traces. XML output would of course need either the full ID or the pieces necessary to construct it.
> 
I consider the engine ID to be a means for an engine to announce what role it wants to play. Having some de facto agreement about which ID means what (eg "junit5" means "I handle all j5 tests")  can give the launcher the possibility to react in different ways:
- Call out the conflict and stop
- Warn about the conflict but choose one engine over the other
- Allow several engines with the same role and add an additional index to the engine ID before handing it out

Using a fully qualified classname as ID is brittle in this respect when implementation details change. 

Anyway, I think we'll have to delay the decision till our next meeting. Maybe, Kevin, you'd like to join us then?


Johannes

kcooney: Interesting ideas about an engine ID being a role.

After sending my message I had the same realization about brittleness of using the fill class name. Perhaps we could recommend or require that third parties have the engine IDs prefixed by the first two parts of the class name (ex: "com.google.nifty").

Totally agree on the idea of putting disagreements asside so we continue the momentum, and tracking the concerns by adding TODO comments or issues (or both; ex: "TODO(b/12345): is this necessary?"). Often the right path becomes clear on its own later.

PR #8, opened 2015-11-06
Changed Responsibility for EngineId

bechte:
I moved the responsibility for taking care about the engineId into the Engine class and its helpers. Any class outside of the junit5-engine package should not operate on the engine id.

With this change I also introduce delegate methods on the engine class which might be a good indicator in which way we should move on:

```
public JUnit5Testable fromUniqueId(String uniqueId) {
		return testableFactory.fromUniqueId(uniqueId);
	}

	public JUnit5Testable fromClassName(String className) {
		return testableFactory.fromClassName(className);
	}

	public JUnit5Testable fromClass(Class<?> clazz) {
		return testableFactory.fromClass(clazz);
	}

	public JUnit5Testable fromMethod(Method testMethod, Class<?> clazz) {
		return testableFactory.fromMethod(testMethod, clazz);
	}
```

We could extract these methods into the test engine interface, i.e. each engine is responsible for the resolution. With that approach we could extract the more generic part of the resolver into a separate class that operates on and interacts with all engines. Still a way to go, but I wanted to point into this direction. What do you think?

### DIFF
--- a/junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
+++ b/junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
@@ -184,7 +184,7 @@ public class ReflectionUtils {
 		return true;
 	}
 
-	public static Class[] findAllClassesInIackage(String packageName) {
+	public static Class[] findAllClassesInPackage(String packageName) {
 		return new ReflectionPackage(packageName).findAllClasses();
 	}
 
--- a/junit-commons/src/test/java/org/junit/gen5/commons/util/FindClassesInPackageTest.java
+++ b/junit-commons/src/test/java/org/junit/gen5/commons/util/FindClassesInPackageTest.java
@@ -12,7 +12,6 @@ package org.junit.gen5.commons.util;
 
 import java.io.IOException;
 import java.util.Arrays;
-import java.util.Collections;
 
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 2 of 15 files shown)


jlink:
For me this is going the wrong way :-/

Stefan, we're playing a ping pong development game here. I change it one way, you change it back, and so on. I suggest we stop fiddling with this part of the engine API until we meet again to come to an agreement within the whole team. 

Johannes

Von meinem iPad gesendet

bechte:
Just to make my point: The Launcher should not influence the resolution of the unique id for an engine at all. How would it guarantee, that subsequent test runs will always have the same ids?

One example: Let's say we have two engines that have a conflicting local/engine-specific unique id (other cases dont lead to problems anyways). What could a Launcher do to make them unique? The only feasible solution is to add the engine id in front of the local/engine-specific id. If there is still a conflict, the Launcher must give up and throw an exception telling that the engines are not compatible. Why?

If the Launcher influences the unique id by some kind of "magic" due to the given engines on the classpath, the returned full unique id would not be stable. It might be, that a subsequent call to the Launcher has only one engine on the classpath leading to another mapping/building of the full unique id.

So I definitely go for the solution, that the engine must guarantee the globally uniqueness of its descriptors, because it just doesnt make sense to move this responsibility somewhere else. The launcher only has to prove, that the ids are unique, and if it detects a conflict, it must throw some kind of exception complaining about the incompatibility of the engines.

But yes, we (you and me) have different ideas about this. So we both made our points, let's have someone else to decide on it...

bechte:
Yeah the cyclic dependency isn't good. It shows that there is still some responsibility misplaced. I think the resolver and the engine have to be refactored as well. I just haven't done it yet.

The factories of course have a relation to the engine. For me the engine has to provide such a mechanism so that extensions can contribute and communicate with the Engine. The first approach was way to static and did leave lot of space for extensibility. But we will get to this point later I think.

PR #9, opened 2015-11-07
Method argument injection

mmerdes:
just merged master into the experimental method-argument-injection branch.
the latter is modeled roughly after the respective spring argument injection solution.
it includes two sample argument resolvers (by type and by annotation) 
and a couple of test cases to demonstrate these two resolvers.
examples for overloaded test methods are included as well.
there is a dummy registry for resolvers which will have to be replaced when there is a proper extension mechanism.
some polishing still needs to be done (e.g., failing in the case of too many competing resolvers)


jlink:
A couple of comments:
1. You should use AnnotationUtils in the SimpleAnnotationBasedMethodArgumentResolver for direct support of meta annotations
2. Looking at the myriad of exception that can occur when resolving a parameter, we should introduce s.t. like ArgumentResolutionException. I'm all for unchecked exceptions but that's my personal taste.
3. before/after methods should also be resolved. Not sure how much the mechanism must be tweeked for this. There might be occasion when arguments are only available after initialization. 
4. We should have some kind of TestContext object available for injection by default
After fixing 1, 2, and 4 I'd vote for merging the PR in.

junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
```diff
@@ -47,12 +48,18 @@ private ReflectionUtils() {
 		return constructor.newInstance();
 	}
 
-	public static Object invokeMethod(Method method, Object testInstance)
+	public static Object invokeMethod(Method method, Object testInstance, Object[] arguments)
```
marcphilipp: How about using varargs here, i.e. `Object... arguments`. Then you wouldn't have to overload the method.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/MethodTestExecutionNode.java
```diff
@@ -88,6 +89,25 @@ else if (exceptionThrown instanceof TestAbortedException) {
...
+		List<Object> arguments = this.prepareArguments(methodTestDescriptor);
+
+		Method testMethod = methodTestDescriptor.getTestMethod();
+		ReflectionUtils.invokeMethod(testMethod, testInstance, arguments.toArray());
+	}
+
+	private List<Object> prepareArguments(MethodTestDescriptor methodTestDescriptor)
+			throws NoSuchMethodException, InstantiationException, IllegalAccessException, InvocationTargetException {
+
+		//should probably not be instantiated here, maybe be brought along by the executionContext
```
marcphilipp: Should be marked with `TODO`, right?

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/injection/MethodArgumentResolver.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.junit.gen5.commons.util.*;
+
+// for a 'real' solution see: org.springframework.web.method.support.HandlerMethodArgumentResolver
+public interface MethodArgumentResolver {
+
+	boolean supports(Parameter parameter);
+
+	default public Object resolveArgumentForMethodParameter(Parameter parameter)
+			throws NoSuchMethodException, InstantiationException, IllegalAccessException, InvocationTargetException {
```
marcphilipp: Do we need to declare all those exception types in the signature?

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/injection/MethodArgumentResolver.java
```diff
@@ -0,0 +1,30 @@
...
+package org.junit.gen5.engine.junit5.execution.injection;
+
+import java.lang.reflect.*;
+
+import org.junit.gen5.commons.util.*;
+
+// for a 'real' solution see: org.springframework.web.method.support.HandlerMethodArgumentResolver
+public interface MethodArgumentResolver {
+
+	boolean supports(Parameter parameter);
```
marcphilipp: I think it's fine to use `Parameter` for now but I would rather pass in our own abstraction of it so we can make sure annotations are read consistently etc. in a next step.

marcphilipp:
I agree with Johannes on (1). I think (3) and (4) can wait. For item (2) it could live with just throwing `Exception` -- since the only caller should be the internals pf the JUnit 5 engine.

mmerdes:
Thanks for the Feedback so far,
I will try to fix Things tomorrow

Good Night
Matthias
Marc Philipp <notifications@github.com> schrieb am Mo., 9. Nov. 2015 um
20:22:

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/injection/MethodArgumentResolver.java
marcphilipp: Use `ArgumentResolutionException` now?

kcooney:
Very excited to see this happen!

A few things I would like to see as we polish this feature

1. It's not obvious from looking at tests that use injection what's going on (where do the parameters come from? is it from dependency injection, or is this some kind of parameterized test?) I feel that this violates the principle that tests should be easy to understand (see https://github.com/junit-team/junit-lambda/wiki/Core-Principles). When I was thinking of creating an extension point for this in JUnit4, I was playing with an annotation like this:

    public class MyTest {
      @Parameterizer
      public final MethodParameterizer parameterizer = new GuiceMethodParameterizer(
          new MyTestGuiceModule());

Someone reading the code would clearly see some kind of method parameterization was going on, can look at the particular class being created to see how the parameterization works, and look at the Javadoc for the `@Parameterizer` annotation to see how to write your own parameterization extension.

We could, of course, use a class annotation:

    @ParameterizeWith(GuiceMethodParameterizer.class)
    @GuiceModule(MyTestGuiceModule.class)
    public class MyTest {
       ...

...or a meta annotation:

    @Target(ElementType.CLASS)
    @Documented
    @MethodParameterizer
    public @interface InjectWithGuice {
	Class<? extends Module> module();
    }

    @InjectWithGuice(MyTestGuiceModule.class)
    public class MyTest {
      ..
The meta-annotation might make it a bit hard to understand what's going on (especially if the author of the custom annotation doesn't add Javadoc) but it does allow the author of the custom annotation provide an easy-to-use API.

3) We should consider supporting both method injection and constructor injection. Personally I would love to have my integration tests use constructor injection with Guice/Dagger, and have my test methods parameterized with JUnitParams

3) We should eventually support scopes (for dependency injections that support them; not necessarily in the default dependency injection system built into JUnit). Possible scopes include:

* prototype (new instance every time)
* singleton (same instance every time for the life of the DI container)
* class (one instance created before any of the tests in the class are run, and reused during the test
* method ? (one instance per test method; same instance reused if the test method is called multiple times (for instance, if there is a retry-once-on-failure rule, or if the method is parameterized

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/injection/MethodArgumentResolver.java
mmerdes: fixed by sam in the meantime

jlink:
Kevin, thank for your comments and suggestions. We're definitely thinking
of some annotation along the lines you described. It will be one of the
next steps.
@TestName was the simplest useful example we could think of. It may or may
not stay as an implicit "parameterizer".

As for constructor injection, it hasn't come up in discussions so far - or
I've forgotten :-/
When you say "scope", you mean different life cycles of injected objects?

For the time being we wanted to get the mechanism in place so that
different people could build on that without having to constantly rebase or
merge a branch in order to stay in sync with master.

cheers,
Johannes


2015-11-12 18:27 GMT+01:00 Kevin Cooney <notifications@github.com>:

kcooney:
@jlink yes, by "scope" I mean "different life cycles of injected objects"

One crazy idea: what if `@BeforeClass` methods that have return values automatically acted as providers for injected objects with a lifetime of the test?

    public void ItemDaoTest {
      private final ItemDao itemDao;
      private final LocalDb localDb;

      public ItemDaoTest(LocalDb localDb) {
        this.localDb = localDb;
        this.itemDao = new ItemDao(localDb);
    }

    @BeforeClass
    public LocalDb createLocalDb() {
        InMemoryLocalDb localDb = InMemoryLocalDb.create()
    }

    @After
    public void rollbackTransactions() {
      localDb.rollbackOpenTransations();
    }

    @Test
    public void insertAndGet() {
      Item item = new Item(3);
      itemDao.insert(item);
      Item fetchedItem = itemDao.getById(3);
      assertThat(fetchedItem).isEqualTo(item);
    }

jlink:
Looks quite similar to a proposal that's already been on the table:

@Scope(MyScope.class)
public class SampleBeforeAfterClassTestCase {

    static class MyScope {

        String aString;

        @Before
        void init() {
              aString = "";
        }

        @After
        void release() {
              System.out.println(aString);
        }
    }

    int myvalue = 1;

    @Test
    void test1(MyScope scope) {
        scope.aString += "test1";
    }

    @Test
    void test2(MyScope scope) {
        assert(scope.aString == "test1");
    }

}


2015-11-13 3:06 GMT+01:00 Kevin Cooney <notifications@github.com>:

mmerdes:
kevin,
thanks a lot for your comments - highly appreciated.
argument injection in both @before and @after methods is planned - just not done yet.
i also like your idea of having @before methods act as implicit argument providers.
we might want to add an extra @ArgumentProvider annotation in order to communicate this more clearly.
as for scopes: i am not sure to what extend this could and should be handled by the underlying DI infrastructure (spring, guice, cdi, whatever). in a, say, spring environment it might be confusing to have both flavors of scopes if we introduced custom junit scopes. 

cheers,
matthias

sbrannen:
> argument injection in both @before and @after methods is planned - just not done yet.

Actually, I just pushed that support to `master`. :wink:

Issue #27, opened 2015-12-02
Less restrictive Spotless settings


marcphilipp:
- Try to make Spotless less strict on additional or missing newlines.
- Needs to work in Eclipse and IntelliJ IDEA.

PR #36, opened 2015-12-03
#27: Changed code formatting settings to be less restrictive

bechte:
- raised line lengths to 160 characters
- allow new lines without enforcing them

Note: This pull requests requires a `gradle spotlessApply` afterwards, reformatting the code base with the new formatting settings. Otherwise the build will break. This is also why the build breaks for this branch, but I didn't want to change to files. I could do a rebase after the cleanup of the master branch and apply the formatter afterwards.


schauder:
I don't want to start a discussion (and I at least won't continue it) but IMHO 160characters is way to wide. 120 is kind of ok.

At least I work often with two or even three editor windows side by side on a single monitor, which will cut of lines that long.

And nobody should really enjoy reading 160 characters in a single line.

dinesh707:
wider it gets, developers can write bad codes. I think google has 80. And currently i'm working with 100. As @schauder suggest the width should allow peer-reviewers to review the code easily in a standard monitor side by side.

kcooney:
FWIW, Google uses 100

For these kinds of issues, we are always going to have disagreements. You'll never find a style that everyone loves. That's why I suggested using Google Java style for the current JUnit project; it's detailed, well documents and reasonable.

(The one change from Google was using a 4-character indent, so we didn't have to reformat every line)

Issue #33, opened 2015-12-02
Split TestDescriptor into two interfaces


marcphilipp:
- `TestDescriptor`: immutable API for users of Launcher (in junit-engine-api)
- `MutableTestDescriptor` extends `TestDescriptor`: additional methods to add children etc.

kcooney:
I would prefer a builder. `TestDescriptor` can have a `builder()` static method and a `toBuilder()` instance method.

sbrannen:
Hi Kevin,

A builder wouldn't help us in this particular situation. The current contract of the `TestDescriptor` interface in the prototype contains methods for mutating the state of a descriptor (e.g., `addChild(TestDescriptor)` and `removeChild(TestDescriptor)`). This was, however, never the intention. These types of methods were only added in order to allow the team to make quick progress with the prototype. Now that the prototype phase is over, we are therefore removing such leaky abstractions.

Make sense?

dsaff:
@sbrannen, when will the MutableTestDescriptor be used, then?

jlink:
> Am 02.12.2015 um 22:23 schrieb David Saff <notifications@github.com>:
> @sbrannen, when will the MutableTestDescriptor be used, then?
> 
During the build phase of the test descriptors tree, currently called "test discovery phase". Usage might be hidden behind some kind of builder but unless you want to do it in two phases, you'll have to mutate (add children to) an existing descriptor at some point. Maybe you have different proposal?

PR #43, opened 2015-12-03
Split TestDescriptor into two interfaces

marcphilipp:
Attempt at #33. 

@junit-team/junit-lambda Please review and comment.

### DIFF
--- a/junit-engine-api/src/main/java/org/junit/gen5/engine/TestDescriptor.java
+++ b/junit-engine-api/src/main/java/org/junit/gen5/engine/TestDescriptor.java
@@ -10,6 +10,7 @@
 
 package org.junit.gen5.engine;
 
+import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
@@ -17,11 +18,6 @@ import java.util.concurrent.atomic.AtomicLong;
 /**
  * @since 5.0
  */
-// TODO Divide into public facing TestDescriptor and engine-internal MutableTestDescriptor.
-//
-// The tree of TestDescriptors should be parallel to (and not just a super type of)
-// MutableTestDescriptor so that clients won't rely on implementation details of
-// MutableTestDescriptor.
 public interface TestDescriptor {
 
 	/**
@@ -34,7 +30,7 @@ public interface TestDescriptor {
 
 	String getDisplayName();
 
-	Optional<TestDescriptor> getParent();
+	Optional<? extends TestDescriptor> getParent();
 
 	boolean isTest();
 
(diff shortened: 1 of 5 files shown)


jlink:
OK from my side. One thought: Should MutableTD be moved to "internal" package or something like that?

marcphilipp:
In JUnit 4 we've had a really bad experience using "internal" or "experimental" packages... 

But it would be nice to split the package into Launcher API consumer facing and Engine API implementor facing parts.

junit-engine-api/src/main/java/org/junit/gen5/engine/TestDescriptor.java
```diff
@@ -44,11 +40,16 @@ default boolean isRoot() {
 
 	Set<TestTag> getTags();
 
-	void addChild(TestDescriptor descriptor);
+	Set<? extends TestDescriptor> getChildren();
 
-	void removeChild(TestDescriptor descriptor);
-
-	Set<TestDescriptor> getChildren();
+	default Set<TestDescriptor> allDescendants() {
```
jlink: Shouldn't those return Set<? extends TestDescriptor> ?

marcphilipp: You mean `allDescendants`?

jlink: yep

2015-12-03 15:16 GMT+01:00 Marc Philipp <notifications@github.com>:

marcphilipp: I've used `Set<? extends TestDescriptor>` for methods that are overridden in `MutableTestDescriptor`. I wasn't sure where `allDescendants` will be used. Probably only through `MutableTestDescriptor`? In this case, it could even return `Set<MutableTestDescriptor>`.

Anyway, I'll change it for consistency now... whereever it may end up later on.

dsaff:
Quick FYI note about "internal" and "experimental": in Android, similar effects are achieved by using annotations, to much better effect than our package-based scheme had in JUnit 4 (many IDEs can be configured to treat @Internal public as private for code-completion and other purposes).

jlink:
Is @Internal a built-in annotation?

dsaff:
Oops, turns out I misremembered.  They actually use three strategies:

a) internal packages for "we don't think this will ever be useful externally"
b) javadoc @hide: https://code.google.com/p/doclava/wiki/JavadocTags#@hide for "not for external use at the moment, maybe in the future"
c) Beta annotation (http://developer.android.com/reference/android/support/test/annotation/Beta.html) for "will likely be public, but maybe not in its current form, caveat programmor."

Issue #29, opened 2015-12-02
Investigate more extensible solution for test resolution


marcphilipp:

schauder:
I wrote a mail to the mailing list last week, which among other things touched on the test resolution & execution process. 

To which @bechte replied (https://groups.yahoo.com/neo/groups/junit/conversations/messages/24734)

> The concept of a test engine covers both, lookup and execute tests. The core framework is not supposed to execute the tests itself, because a test engine could provide very specific test behavior, e.g. running cucumber tests, requiring a very specific execution path. Therefore, we decided that the engine is responsible for both. In case of the JUnit5TestEngine, we will provide extension points that can be used by extension providers to hook into the standard „Java testing“ test run circle. We are still working on this extension model, therefore, there is not much within the code, yet. Please be patient about this feature. I hope that we are able to provide it within the next iteration, too.

I still think that executing the tests should be done by some internal Java component, not some pluggable mechanism like a TestEngine.

Tools like IDEs will very soon rely heavily on the correct events/information delivered through some kind of listener API. If the TestEngines have to do this, because they are executing the tests, this will be broken for many TestEngines to come. But if it has to be the same all the time, it should be implemented only once.

I like the Cucumber example. I fully agree that it should be possible to execute write an Engine that registers tests as Cucumber does. And there is some challenging work in this, to find the right abstraction (like: What exactly is a Test, and do we need some Granularity above and below Tests.) But this needs to be represented in a TestSpecifcation, so JUnit can properly report on it. And in the end a Test just gets executed, succeeds, fails or gets skipped, there is not much any Engine can or should change about that.

And if TestEngine actually needs itself for the execution, there is no problem in having a Statement (using the Term from JUnit4 Rules) to reference and use the TestEngine it was created by.

Keep up the good work.

jlink:
If I  understand you right you'd prefer something along the following:

    interface  ContainerDescriptor extends TestDescriptor {
        Executable beforeAll();
        Executable afterAll();
    }

    interface LeafDescriptor extends TestDescriptor {
        Executable runTest();
    }

Well worth a consideration IMO.

schauder:
Yes.

jlink:
Additional complexity: For discovery engines must be given an object / lambda with which they can do things like create dynamic tests, forward additional test data and like during test execution later.

I still think we should consider this approach and see if all types of engines that we think of would fit into such a model.

jlink:
[...]

PR #52, opened 2015-12-07
Introduce TestIdentifier (DTO for TestDescriptor)

marcphilipp:
This pull request introduces `TestIdentifier`, an immutable DTO that contains all the information IDEs and build tools should need. We could even make it `Serializable` in a next step, currently `TestSource` prevents this.

A `TestIdentifier` knows only about the ID of its parent and does not know about its children. The `TestPlan` can be used to query this information. `TestPlan` no longer implements `TestDescriptor` but rather manages the `TestIdentifier` tree structure.

The `TestPlan` is returned by `Launcher.discoverTests()` and passed to `TestExecutionListener` methods. I've merged `TestPlanExecutionListener` and `TestExecutionListener` and added `EngineExecutionListener`. Launcher now uses an adapter that maps from `TestDescription` to `TestIdentifier` using the `TestPlan`.

However, `EngineExecutionListener` is IMHO just an intermediate solution. I would rather remove the need for engines to call the right listener methods as discussed in #29.


jlink:
Just to make sure that I understand the approach fully.

Advantages:
- We don't need to split TestDescriptor and MutableTestDescriptor any more. TestDescriptor can be mutable and used by engines directly.
- TestIdentifier can be serialized (if we make Source serializable) and used by tools directly to transmit information between test running JVM and tool JVM

Disadvantages:
- One additional layer of indirection through which every test and test container has to go.
- Clients cannot navigate a tree of descriptors directly but have to use the TestPlan instance for navigation.

Anything I forgot?

jlink:
I don't quite understand the role of TestId. Couldn't we use the uniqueId string directly?

marcphilipp:
Regarding `TestId`: Yes, we could use the String directly. I found it awkward that `TestPlan` has many methods that take `String` arguments and `TestDescriptor` returns `String` and `Optional<String>` for its parent.

I thought about using TestId in TestDescriptor, too. But then it would have to be moved down into junit-engine-api again.

I could live with `String` if need be.

marcphilipp:
**More advantages:**
- No way to downcast it. :wink: 
- Hopefully more natural support for dynamic tests.

jlink:
Could you elaborate on "more natural support for dynamic tests"?

jlink:
About `TestId` in `TestDescriptor`: Last week there was the idea to have some kind of UniqueId builder that frees an engine of thinking about uniqueness outside of its own context. Maybe an immutable object like that:

    class UniqueId {
        public static UniqueId withEngineId(String eid) { return new UniqueId(eid); }
        private String engineId;
        private List<IdPart> parts = new ArrayList<>();
        private UniqueId(String engineId) { this.engineId = engineId }
        private UniqueId(String engineId, List<IdPart> parts) { ... }
        UniqueId addPart(char separator, String part) { ... }
        IdPart popPart() { ... }
        TestId toTestId() { ... }
        String toString() { ... }
    }

    class IdPart { 
        char separator;
        String part;
    }

This could then serve as the bridge from `TestDescriptor` to `TestId`.

mmerdes:
already merged

PR #53, opened 2015-12-07
Issue28: Alternative approach to JUnit 5 extensions and extension ordering

jlink:
See my last comment for the new proposed extension model.


marcphilipp:
When would `@BeforeEach` etc. methods be called? `INNERMOST`?

marcphilipp:
With the old `BeforeEachCallbacks` you could execute code `postBeforeEach`. Is this still possible with this approach?

marcphilipp:
Would we implement calling `@BeforeEach` and `@AfterEach` as a `TestLifecycleExtension`?

junit5-api/src/main/java/org/junit/gen5/api/extension/TestLifecycleExtension.java
```diff
@@ -0,0 +1,70 @@
...
+	default Object postProcessTestInstance(TestExtensionContext testExtensionContext) throws Exception {
+		return testExtensionContext.getTestInstance();
+	}
+
+	default void beforeEach(TestExtensionContext testExtensionContext) throws Exception {
+		/* no-op */
+	}
+
+	/**
+	 * Should only be implemented by a single extension.
```
marcphilipp: How would we call this method? Since its a default method it is implemented by all `TestLifecycleCallbacks`, isn't it?

jlink:
It can be called in any position. Using INNERMOST etc. defines the order of
several extensions. Think of it as a stack of extensions. Actually two
stacks: One for container lifecycle, one for test lifecycle. All of them
will be applied.

2015-12-07 16:20 GMT+01:00 Marc Philipp <notifications@github.com>:

jlink:
Yes. You can be even more precise - but you don't have to be.

Use INNERMOST to mark it closest to the method invocation. Or just INSIDE
to be closer to method invocation than those with order OUTERMOST, OUTSIDE
and MIDDLE.

2015-12-07 16:22 GMT+01:00 Marc Philipp <notifications@github.com>:

marcphilipp:
I' probably missing something: When would a method annotated with `@BeforeEach` be called when I have the following `TestLifecycleExtensions`?

- OUTERMOST
- OUTSIDE
- MIDDLE
- INSIDE
- INNERMOST

I'm guessing before or after MIDDLE?

marcphilipp:
Is `DEFAULT` just `MIDDLE`?

jlink:
If we use the same extension mechanism for the annotated methods - what I'd recommend- they would all be MIDDLE. 

I guess I'll have to draw a diagram to make the model's intent clear.

jlink:
Maybe trying to express everything with annotations is not a good idea, since it seems difficult - at least for me - to communicate the intention of the annotations.

What about something like this:

    enum OrderPosition {
        FIRST, BEFORE, MIDDLE, AFTER, LAST
    }

    interface TestExtension {
        void registerCallbacks(CallbackRegistry registry);
    }

    class MyExtension implements TestExtension {
        void registerCallbacks(CallbackRegistry registry) {
            registry.registerMethodParameterResolver(new MyResolver());
            registry.registerBeforeAll(containerContext -> doSomething(containerContext), OrderPosition.FIRST);
            ...
        }
        class MyResolver implements MethodParameterResolver { ... }
    }

Callbacks  could be functional interfaces.

This would allow any extension to register the same callback several times for different order positions. The CallbackRegistry could handle (or announce) all conflicts - for example two extensions registering the same callback as FIRST. The registry would be context-specific and not global.

Looking at it, it seems to communicate much better than the annotation-based variant. What's missing is a way to change the orderPosition from outside in case of a conflict. Maybe one of you has an idea how this could be added?

jlink:
I just implemented the model suggested in previous comment.

jlink:
Will implement a similiar approach (after discussing this one in the team) in another branch.

Issue #39, opened 2015-12-03
Re-implement JUnit 5 engine


marcphilipp:
- [ ] Resolution (in #29)
- [x] ExtensionRegistry honors ordering
- [x] Implement getParent() of ExtensionContext
- [ ] Encapsulate attributes for ExtensionContext, do not expose Map
- [x] Invocation of ExtensionPoints
  - [x] BeforeEach
  - [x] AfterEach
  - [x] BeforeAll
  - [x] AfterAll
  - [x] PostProcessTestInstance
  - [X] MethodParameterResolver
  - [x] Conditional test execution
- [ ] Clean up integrated tests (subclasses of AbstractJUnit5TestEngineTestCase)

mmerdes:
johannes and i will try to come up with a sketch of a new engine tomorrow.
should we use this issue or a different (yet to be created) one?

marcphilipp:
I think referencing this issue is fine.

mmerdes:
thanks for the clarification :)

marcphilipp:
I'm working on cleaning up HierarchicalTestEngine.

jlink:
I'm working on Before and After extension points.

bechte:
Hi guys,

I have almost finished the implementation of the resolvers. Still there are some (minor) issues, that I would like to talk about:

1) Given a user has hand in an unique id to a method, the test cases require that both (the method descriptor and the class descriptor) are returned. According to JUnit4, this is a change. Do we really want to change this behavior? I somehow feel, that it is reasonable to run nodes that have been given by an unique id are run on the top level. Still not sure about this one…

From an execution’s point of view, our current implementation requires the test methods be be placed below a test class node, otherwise the beforeAll|Each/afterAll|Each will not get executed. Do we want this strong relationship? How about the package level? Shall we also introduce nodes for each package? This would be a consequence, at least with java 9… What do you think?


2) Handling of duplication: How do we want to handle duplication, i.e. if a test is specified more than once according to the hierarchy. Where do we want to place the test in the tree? Example:

TestClass {
  void test1();
  void test2();
}

Given a test plan specification with a class element „TestClass“ and an unique id element „junit5:TestCass.test1()“, how should the resulting tree look like? The most reasonable result I can think of is:

- junit5
-- TestClass
--- test1()
--- test2()


[...]

PR #65, opened 2015-12-15
Feature 29 extensible discovery

bechte:
Resolved and fixed all issues from #29 and #39.

Possible improvements that we can work on after this merge:
- [ ] Define a concept of a "path" along the unique id
- [ ] Use "path" concept to
  - [ ] remove tight coupling between resolvers
  - [ ] split `ClassResolver` into two classes: `TopLevelClassResolver` & `NestedClassResolver`
  - [ ] extract `NestedClassResolver` & `NestedClassTestDescriptor` as an extension
- [ ] Clarify whether or not to keep static inner classes on the top level
- [ ] Rework the `TestPlanSpecification`: Extensibility vs. Data Structure

I would like to handle those improvements if necessary in separate tickets working on the master branch. This will allow us to work in parallel on the topics.


junit-commons/src/main/java/org/junit/gen5/commons/util/ObjectUtils.java
```diff
@@ -29,4 +31,8 @@ public static String nullSafeToString(Class<?>... classes) {
 		return stream(classes).map(Class::getName).collect(joining(", "));
 	}
 
+	public static void verifyNonNull(Object obj, String message) {
+		if (obj == null)
+			throw new IllegalArgumentException(message);
+	}
```
marcphilipp: This is exactly the same as `Preconditions.notNull()`. Please remove this method again and use the existing one instead. 😉

junit-tests/src/test/java/org/junit/gen5/engine/DummyTestEngineDescriptor.java
```diff
@@ -10,20 +10,21 @@
 
 package org.junit.gen5.engine;
 
-class DummyTestDescriptor extends AbstractTestDescriptor implements Leaf<DummyEngineExecutionContext> {
-
-	public static final String ENGINE_ID = "dummy";
-
+class DummyTestEngineDescriptor extends AbstractTestDescriptor implements Leaf<DummyEngineExecutionContext> {
```
marcphilipp: Why did you rename this class? I think the new name is confusing because it's not actually an `EngineDescriptor`.

bechte: Based on the unique id of this dummy, it represents a test engine node. That is why I renamed it. I agree the new name does not reflect it either. `implements Leaf<bla>` was added after I changed the name, so the intent of this dummy might have changed. I can rename it back to what it was. Still I think we should talk about a naming convention for test doubles.

junit-commons/src/main/java/org/junit/gen5/commons/util/ObjectUtils.java
bechte: True :) - it was late. Thanks.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/MethodTestDescriptor.java
```diff
@@ -57,6 +54,17 @@
...
 	@Override
+	public String getUniqueId() {
+		// @formatter:off
+		return String.format("%s:%s#%s(%s)",
+				getTestEngine().getId(),
+				getTestClass().getName(),
+				getTestMethod().getName(),
+				ObjectUtils.nullSafeToString(getTestMethod().getParameterTypes()));
+		// @formatter:on
+	}
```
marcphilipp: Is this correct for methods in nested non-static classes?

Here and in the other `getUniqueId()` methods: Have you considered implementing this like the following (pseudo-code)?
```java
@Override
public String getUniqueId() {
	// @formatter:off
	return String.format("%s#%s(%s)",
			getParent().map(TestDescriptor::getUniqueId).orElse("")
			getTestMethod().getName(),
			ObjectUtils.nullSafeToString(getTestMethod().getParameterTypes()));
	// @formatter:on
}
```

junit-tests/src/test/java/org/junit/gen5/engine/DummyTestEngineDescriptor.java
marcphilipp: Well, at the moment it is used for the engine node and its children. The unique IDs are build like this: `ENGINE_ID + ":someString"`. I agree that this could be cleaned up, but would prefer the original class name.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/MethodTestDescriptor.java
bechte: Yep, it is correct (we have tests for this). Anyways, I am with you, we should investigate some more time on this. I would like to have a separate ticket for it, as it is not as easy as it looks in the first place. If you like, we could pair on this topic and have a look at your "path" idea.

I have considered to create the unique id using the parent, but in my first version this did not work out, as I moved methods from the specification on the top level. After we decided to create the whole tree for every node, it is now possible to use the parent for id generation. As I said, I would like to consider this topic in an own issue, as I think it needs some more time to find out all dependencies.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/resolver/TestResolver.java
```diff
@@ -0,0 +1,26 @@
...
+package org.junit.gen5.engine.junit5.resolver;
+
+import org.junit.gen5.engine.TestDescriptor;
+import org.junit.gen5.engine.TestEngine;
+import org.junit.gen5.engine.TestPlanSpecification;
+
+/**
+ * A {@link TestResolver} is responsible for resolving different kind of test representatives. Each resolver can
+ * contribute to the list of children of the given parent. It may only resolve children, that accomplish the
+ * {@link TestPlanSpecification}. The children are returned as list.
```
marcphilipp: I believe the last sentence is outdated.

bechte:
This feature will be discussed internally and rebuilt onto master. Therefore, I am closing this pull request.

PR #70, opened 2015-12-17
Issue57 extensible reporting

mmerdes:
@jlink and i prepared the pull request for the infrastructural aspects of a generic reporting facility.
the reporting entries are currently of type Map<String, String> but will be replaced by a dedicated value object.
please review and provide feedback.


junit-engine-api/src/main/java/org/junit/gen5/engine/EngineExecutionListener.java
```diff
@@ -10,12 +10,17 @@
...
 
+import java.util.*;
+
 /**
  * @since 5.0
  */
 public interface EngineExecutionListener {
 
-	void dynamicTestRegistered(TestDescriptor testDescriptor);
+	void reportingEntryPublished(TestDescriptor testDescriptor, Map<String, String> entry);
```
jlink: I think `entry` should actually be some kind of class to wrap the key-value pairs and maybe add something like a timestamp? This would also simplify evolution.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/JUnit5EngineExecutionContext.java
```diff
@@ -49,11 +52,22 @@ public static Builder builder(JUnit5EngineExecutionContext context) {
 		return new Builder(context.state, null);
 	}
 
+	@Override
+	public void registerPublishHandler(PublishHandler handler) {
+		state.publishHandler = handler;
+	}
+
+	public void publishReportEntry(TestDescriptor descriptor, Map<String, String> entry) {
```
jlink: Alternatively the executionContext could provide a getter for `PublishHandler`. Then `AbstractDescriptorBasedExtensionContext` would only need the publish handler instead of the full engine context.

jlink:
You could enhance the console runner to report publish events. Or does it already?

marcphilipp:
Merged manually in b4957a135bb9b9eba12c8bcb154c8cfbfb4cc566.

PR #71, opened 2015-12-19
Testing package dependencies with Degraph

schauder:
Adds a test that tests for dependency cycles on package an module level.

Modules in this sense are all classes that start with the same package: org.junit.gen5.x

In order to make the test pass I had to break to dependencies from test to the JUnit5TestEngine. Since the engine wasn't really part of the test, I replaced it with a mock.

I agree to https://github.com/junit-team/junit-lambda/blob/master/CONTRIBUTING.md

### DIFF
--- a/junit-tests/build.gradle
+++ b/junit-tests/build.gradle
@@ -24,6 +24,7 @@ dependencies {
 	testCompile("de.schauderhaft.degraph:degraph-check:${degraphVersion}")
 
 	testRuntime("org.apache.logging.log4j:log4j-core:${log4JVersion}")
+	testRuntime("org.apache.logging.log4j:log4j-jul:${log4JVersion}")
 }
 
 jacoco {
--- a/junit-tests/build.gradle
+++ b/junit-tests/build.gradle
@@ -21,9 +21,9 @@ dependencies {
 	testCompile("org.assertj:assertj-core:${assertJVersion}")
 	testCompile("junit:junit:${junit4Version}")
 	testCompile("org.mockito:mockito-core:${mockitoVersion}")
+	testCompile("de.schauderhaft.degraph:degraph-check:${degraphVersion}")
 
 	testRuntime("org.apache.logging.log4j:log4j-core:${log4JVersion}")
-	testRuntime("org.apache.logging.log4j:log4j-jul:${log4JVersion}")
 }
 
 jacoco {
--- a/junit-tests/src/test/java/org/junit/gen5/commons/util/org/junit/gen5/meta/DependencyTests.java
+++ b/junit-tests/src/test/java/org/junit/gen5/commons/util/org/junit/gen5/meta/DependencyTests.java
@@ -27,8 +27,9 @@ public class DependencyTests {
 
     @Test
     public void noCycles(){
+        // we can't use noJar(), because with gradle the dependencies of other modules are
+        // included as jar files in the path.
         Assert.assertThat(classpath()
-                .noJars()
                 .printTo("dependencies.graphml")
                 .including("org.junit.**")
                 .withSlicing("module", "org.junit.gen5.(*).**"),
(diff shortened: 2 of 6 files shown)


junit-tests/build.gradle
```diff
@@ -21,9 +21,9 @@ dependencies {
 	testCompile("org.assertj:assertj-core:${assertJVersion}")
 	testCompile("junit:junit:${junit4Version}")
 	testCompile("org.mockito:mockito-core:${mockitoVersion}")
+	testCompile("de.schauderhaft.degraph:degraph-check:${degraphVersion}")
 
 	testRuntime("org.apache.logging.log4j:log4j-core:${log4JVersion}")
-	testRuntime("org.apache.logging.log4j:log4j-jul:${log4JVersion}")
```
marcphilipp: Why did you remove that?

marcphilipp:
@schauder Thanks for the PR! Just one minor question in the diff.

junit-tests/src/test/java/org/junit/gen5/commons/util/org/junit/gen5/meta/DependencyTests.java
```diff
@@ -0,0 +1,37 @@
...
+ * Copyright 2015 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.gen5.commons.util.org.junit.gen5.meta;
```
marcphilipp: Should be just `org.junit.gen5.meta`, right?

junit-tests/build.gradle
schauder: Ooops, by accident. I'll put it back in.

marcphilipp:
@schauder Is there a way to tell degraph that `org.junit.gen5.engine.junit4` (junit4-engine), `org.junit.gen5.engine.junit5` (junit5-engine) and everything else in `org.junit.gen5.engine.**` (engine-api) are separate modules? I tried using `NamedPattern` but couldn't get it to work.

jlink:
For legal reasons you'll have to add a paragraph to the pull request description that states that you agree to the terms in https://github.com/junit-team/junit-lambda/blob/master/CONTRIBUTING.md

schauder:
@marcphilipp it is possible and NamedPatterns are the way to go (although you could achieve it without them) You probably used the parameters in the wrong oder. No idea why I put them the way they are ...

schauder:
@jlink done. Pulling out org.junit.Assert right now and replacing it with import static org.hamcrest.MatcherAssert.assertThat;

marcphilipp:
@schauder Thanks! I've merged this pull request manually. Great to finally have this check in place. :+1:

jlink:
I'd like to pull `DependencyTests` out of junit-tests since it takes 13 secs and slows down the suite.
Any counter arguments? Pull out to a new project?

sbrannen:
I agree with @jlink: `DependencyTests` should not be a part of the standard _pre-commit_ build. In fact, I think it shouldn't be part of the standard CI build either. Rather, I would prefer that it be part of a _nightly_ build.

schauder:
While I agree that it is to slow for the first build step of a CI pipeline, I prefer to let it run after every push, but in a separate build step.

Another alternative would be to include such a test in each project. Since it would only run on the classes (not the jar files) it should be much faster, although still not as fast as one would wish for a proper Unit test.

jlink:
I don't really mind to have it as a precommit step or in the ci build. But I also want a _very fast_ test suite to run (more or less) continuously.

@schauder Did you think about externalizing the use of the tool into a gradle/maven plugin that gets a DSL like configuration block? This would give users the freedom to use it in the way best fit for their project setup.

schauder:
@jlink Yes: https://github.com/schauder/degraph/issues/62

Issue #67, opened 2015-12-16
Agree and document on conventions for Javadoc


marcphilipp:
### Where to use Javadoc?

- Every class/method that we expose as API?
- Every class/method that is public?

### How to write/format Javadoc?

- How to format paragraphs?
  - `<p>` on a separate line or at the beginning of the new paragraph?
- Where to put whitespace?
  - blank line between paragraphs?
  - blank line before tags?
- Define/check order of tags?

There is some good advice here:
- http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html
- http://www.oracle.com/technetwork/java/javase/documentation/index-142372.html
- https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style#javadoc

marcphilipp:
@junit-team/junit-lambda Here's another issue we should discuss.

jlink:
I think we should be quite strict and formal with all things we expose as
API and should be liberal with the rest.

As for the concrete format, you guys all have more experience with writing
JavaDoc than I. I'll go with your recommendations.

2015-12-16 9:28 GMT+01:00 Marc Philipp <notifications@github.com>:

kcooney:
The Google Java Style Guide covers all of these questions:

https://google.github.io/styleguide/javaguide.html#s7-javadoc

To avoid having to debate minor stylistic issues (now and from future contributors) I suggest we just adopt the Google Java Style Guide.

jlink:
I like the idea to adopt a well-known standard. I seem to remember that in
the kick-off some arguments against using the Google style guide were
raised; that's why we didn't use it. I cannot remember which ones, though.

2015-12-17 4:58 GMT+01:00 Kevin Cooney <notifications@github.com>:

marcphilipp:
I can remember the 2-space indentation being a major objection. This issue is just about Javadoc guidelines... In this regard I find the Google style guide helpful. 😉

marcphilipp:
Team discussion results:

> - Prefer putting `<p>` on same line as text with preceding blank line.
> - Blank line before at-clauses/tags.
> - Prefer to use `{@code foo}` over `<code>foo</code>` (better escaping).
> - Use `@since 5.0` instead of `@since 5.0.0`.

PR #73, opened 2015-12-19
Checkstyle for Javadoc

marcphilipp:
This pull request is related to #67 and adds Checkstyle to the Gradle build to check the following basics for Javadoc:

- Paragraphs are separated by blank lines and prefixes with `<p>` immediately before the first word of the paragraph (`JavadocParagraph`).
- Exceptions documented using `@throws` on methods must be present on their signature (`JavadocMethod`).
- The order of "tags"/"at clauses" must be `@author`, `@param`, `@return`, `@throws`, `@exception`, `@since`, `@see` (`AtclauseOrder`).
- All "tags"/"at clauses" must have a description. (`NonEmptyAtclauseDescription`).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ subprojects {
 	apply plugin: 'maven'
 	apply plugin: 'signing'
 	apply plugin: 'com.diffplug.gradle.spotless'
+	apply plugin: 'checkstyle'
 
 	repositories {
 		mavenCentral()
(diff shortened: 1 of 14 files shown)


sbrannen:
Can we please have `@since` tags immediately follow `@author` tags?

jlink:
Does it enforce Javadoc to be present?

marcphilipp:
@sbrannen Sure.

@jlink No, it does not.

jlink:
Then go ahead and merge it.

Issue #64, opened 2015-12-13
ReflectionUtils.handleException() should not wrap a target exception in an IllegalStateException


jlink:
## Status Quo

The `handleException()` method in `ReflectionUtils` currently extracts a _target exception_ from an `InvocationTargetException`, and if the _target exception_ is a checked `Exception` it will be wrapped in an `IllegalStateException` with the message `"Unhandled exception"`.

This behavior, however, is inappropriate for a testing framework that reflectively invokes test methods.

Specifically, any exception thrown by a test method must be reported in `TestExecutionResult`. If such an exception is wrapped in an `IllegalStateException`, that makes it very difficult to _retrieve_ the original exception, since it is impossible to know that some `IllegalStateExceptions` should be treated differently than others.

## Analysis

We do not want to have to reason about the contents of an `IllegalStateException`. Instead, we need an explicit way to retrieve an exception thrown by a test method.

## Deliverables

- [x] Change everything that `throws Throwable` to `throws Exception`
- [x] Change ReflectionUtlils.invokeMethod()/newInstance() to _mask_ checked exceptions as unchecked exceptions without physically wrapping them in a `RuntimeException`.
- [x] Refactor `ReflectionUtils.handleException()` accordingly.

sbrannen:
Thanks for raising this issue.

I've been aware of it for quite some time; I just didn't get around to creating a ticket for it.

We actually do not want to propagate an `InvocationTargetException` since it is a checked exception. Instead, we should wrap either the `InvocationTargetException` or its _target exception` in a custom `RuntimeException` that we can then correctly handle higher up the call stack.

sbrannen:
I have completely revised this issue's title and description.

jlink:
I'm just about adding some tests for correct exception propagation in the engine. Fixing the tests will force me to come up with a proposal.

jlink:
What's the problem with a checked exception? If we have to catch exceptions when calling invoke anyway, we could even throw Throwsble...

marcphilipp:
I assumed it would be sufficient to rethrow the target exception. Why would we want to wrap it?

sbrannen:
Hi guys,

Those are naturally very valid questions.

The question here is a fundamental one: namely, should our utility classes throw checked exceptions?

[...]

PR #77, opened 2015-12-21
ReflectionUtils.invoke now throws Throwable.

jlink:
Gets rid of Wrapper exception and a couple of catch clauses, but also requires (at least for reasons of symmetry) that ReflectionUtils.newInstance() throws Throwable.

### DIFF
--- a/junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
+++ b/junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
@@ -85,7 +85,7 @@ public final class ReflectionUtils {
 		return Modifier.isStatic(member.getModifiers());
 	}
 
-	public static <T> T newInstance(Class<T> clazz, Object... args) {
+	public static <T> T newInstance(Class<T> clazz, Object... args) throws Throwable {
 		Preconditions.notNull(clazz, "class must not be null");
 
 		try {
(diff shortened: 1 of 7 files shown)


junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
```diff
@@ -405,23 +405,7 @@ private static void handleException(Throwable ex) {
...
-		private final Throwable targetException;
-
-		private TargetExceptionWrapper(Throwable targetException) {
-			this.targetException = targetException;
-		}
-
-		public Throwable getTargetException() {
-			return targetException;
-		}
+		throw ex;
```
marcphilipp: Now we don't need the two `ifs` checking for `RuntimeException` and `Error` anymore, do we?

marcphilipp:
I think this version is an improvement over `TargetExceptionWrapper`. :+1:

junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
jlink: Right.

jlink:
I also vote for merging the PR. Since ReflectionUtils are only internally used I think we could remove Throwable again later without too much hassle.

junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
```diff
@@ -405,23 +405,7 @@ private static void handleException(Throwable ex) {
 		if (ex instanceof Error) {
```
bechte: After removing this if block, I think it is ready to merge. Thanks.

jlink:
Open questions:
- throws Throwable or Exception in public interfaces
- What to do about special exception, e.g. ParameterResolutionException. Checked or Unchecked?

jlink:
Pull request not compatible with ideas in #64. Implementor should open new one.

PR #89, opened 2016-01-06
Executable.execute() should allow throwing Throwable

lukaseder:
`Assertions.assertThrows()` also allows for `Throwable` classes to be checked for:
https://github.com/junit-team/junit-lambda/blob/8a6a13c58d9e4a5f734037cdb088902364d1ba43/junit5-api/src/main/java/org/junit/gen5/api/Assertions.java#L204

Thus, `Throwable` must be the declared type on this `execute()` method.

This is particularyl interesting for Scala usage, where (I believe) exceptions derive from `Throwable`, not from `Exception`

### DIFF
--- a/junit5-api/src/main/java/org/junit/gen5/api/Executable.java
+++ b/junit5-api/src/main/java/org/junit/gen5/api/Executable.java
@@ -13,10 +13,10 @@ package org.junit.gen5.api;
 /**
  * {@code Executable} is a functional interface that can be used to
  * implement any generic block of code that potentially throws an
- * {@link Exception}.
+ * {@link Throwable}.
  *
  * <p>The {@code Executable} interface is similar to {@link java.lang.Runnable},
- * except that an {@code Executable} can throw an {@link Exception}.
+ * except that an {@code Executable} can throw an {@link Throwable}.
  *
  * @since 5.0
  * @see Assertions#assertAll(Executable...)
@@ -27,6 +27,6 @@ package org.junit.gen5.api;
 @FunctionalInterface
 public interface Executable {
 
-	void execute() throws Exception;
+	void execute() throws Throwable;
 
 }


jlink:
Just yesterday we changed it from Throwable to Exception arguing that no Java API code should deliberately throw Throwable. 

I'm wondering how relevant the Scala argument is. Does any serious Scala dev use plain JUnit? I'd rather think ScalaTest would implement its own test engine which has all liberties. Other thoughts?

lukaseder:
Why be opinionated about this? In other words: What do you gain from this restriction, apart from the occasional edge case not working in client code (which will inevitably lead to this discussion again)?

Some edge cases:

- I want to test for the presence or absence of a `StackOverflowError`
- I'm writing some OSGi components and I want to make sure that `NoClassDefFoundError` is thrown after some operation
- I want to write a test that checks for some weird third party library's behaviour
- I want to make sure that illegal arguments result in `java.lang.AssertionError`

Tons of use cases would be impossible, and I keep wondering: What for? :-)

mmerdes:
Considering that usage and handling of Exception and Throwable in the broader JVM world are less consistent than would be desirable we might indeed go with the less restrictive variant - even if the more restrictive one would be more satisfactory in theory...

sbrannen:
FYI: this is the commit from yesterday that changed the signature for `Executable.execute()`: https://github.com/junit-team/junit-lambda/commit/58d455c6da6ce482e079d27892c058dd759cc6a7

sbrannen:
@lukaseder,

It's great to see you chiming in here and that you are actively working with the JUnit 5 snapshots!

But now... to switch roles for a second... 

[begin:devil's advocate]

> - I want to test for the presence or absence of a StackOverflowError
> - I'm writing some OSGi components and I want to make sure that NoClassDefFoundError is thrown after some operation
> - I want to make sure that illegal arguments result in java.lang.AssertionError

That' not a problem. `NoClassDefFoundError`, `StackOverflowError`, and `java.lang.AssertionError` are all subclasses of `Error` and as such are _unchecked_.

In other words, `Executable.execute()` doesn't need to declare that it `throws Throwable` if the invocation of the method throws a `java.lang.Error` or `java.lang.Exception`.

For example, the following test passes as you would expect (with the current code in `master`):

```java
@Test
void assertThrowsStackOverflowError() {
	assertThrows(StackOverflowError.class, this::recurseIndefinitely);
}

private void recurseIndefinitely() {
	// simulate infinite recursion
	throw new StackOverflowError();
}
```

[end:devil's advocate]

> - I want to write a test that checks for some weird third party library's behaviour

In my opinion, this is a valid argument in favor of having `Executable.execute()` declare that it `throws Throwable`.

In general we strive to avoid allowing methods to declare `throws Throwable` since it's typically not a recommended practice in Java. However, since it is in fact possible that user code or third-party libraries explicitly declare `throws Throwable`, we will follow the less restrictive path and revert the changes from yesterday.

Cheers,

Sam

kcooney:
I am actually almost convinced by Sam's "devil's advocate"  argument that the method should throw `Exception`. The JDK `Callable` class is even more generic (it has a result) but still `Callable.call()` throws `Exception`.

Could someone expand on why "weird third party library's behaviour" might require throwing `Throwable`?

sbrannen:
Resolved by reverting previous changes in dc5b6d7d78167f417291c13c141ab2d68cf28e6d.

lukaseder:
@sbrannen: Nice to hear from you too :-)
@kcooney: OK, conversely, two more arguments:

- Why do you catch `Throwable` and declare `<T extends Throwable>` in `Assertions`? What do you want to communicate with this API inconsistency?
- `Callable` establishes a contract on what is allowed within an `ExecutorService`. That's different from you providing convenience API for arbitrary third party testing. `Callable` is useful for a much narrower scope than `Assertions.expectThrows()`

Again, I'm not convinced by the fact that JUnit should be opinionated by this. JUnit gains nothing by adding a restriction here.

Java has warts and legacy. Let's just accept them, and offer the most functionality possible.

sbrannen:
@kcooney,

As a purist, I would also vote for throwing `Exception` rather than `Throwable`, but the crux of this issue is that people _can_ throw `Throwable`. They can even create their own custom subclasses of `Throwable`. I would not condone that anyone do that, but... since we can't prevent anyone from declaring that a method throws `Throwable`, one can argue that an `Executable` should be permitted to throw `Throwable`. Otherwise, we limit the utility of having `Executable` as a functional interface: if the code someone wants to pass to `assertThrows` declares that it throws `Throwable`, it then becomes quite ugly to use lambdas or method references within `assertThrows`.

There are pros and cons to both sides of this coin, but in the end... I think we will have fewer issues in the long term if we follow the less restrictive path.

lukaseder:
> As a purist

Depends on what "purity" you mean. Subjective "best practices"? -> `Exception` (says who, btw!?). Type theory and API consistecy?  -> `Throwable` :)

bechte:
Actually, we don't know and we cannot avoid that someone declares Throwable (whether it makes sense or not). There is no reason for us to limit the TestExecutable to only throw Exception, other than we thought this would be suffient. The reasons we discussed (also 1-3) were all pointing towards this limitation but argument 4 convinces me, that we should use Throwable instead of Exception, and if it is "only" to not limit test writers in any case (which is a great thing writing a test framework).

sbrannen:
@lukaseder, the _purity_ of which I dreamt is based on a desire that `Throwable` not be a _checked_ exception in the Java language, but... some desires will forever remain dreams. :wink:

kcooney:
Okay, I am convinced :-) I did say "almost convinced"

One question: should we make `Executor` a top-level class? Don't we have the exact same interface in `Assertions` and `org.JUnit.Assert`?

lukaseder:
@sbrannen: I see, I have dreamed about that as well ;-)

By the way: Good job on this work, folks!

sbrannen:
@kcooney, I'm assuming you mean `Executable` instead of `Executor` -- right?

If that's what you're referring to then, yes, we have semantically equivalent functional interfaces in JUnit 4.13 and 5.0.

From `org.junit.Assert`:
```java
public interface ThrowingRunnable {
    void run() throws Throwable;
}
```

JUnit 5's `org.junit.gen5.api.Executable`:

```java
@FunctionalInterface
public interface Executable {
	void execute() throws Throwable;
}
```

I have used this form of `Executable` on various projects, and I much prefer it over the `ThrowingRunnable`. If I were a core committer for JUnit 4, I would have already morphed `ThrowingRunnable` into `Executable`, but as it stands I'll have to leave that to you in case you feel so inclined. ;)

Note, however, that `org.junit.gen5.api.Executable` is already a top-level class in the JUnit 5 API.

So, what exactly are you proposing when you ask, "should we make `Executor` a top-level class?".

kcooney:
@sbrannen sorry, you are right that `Executable` is a top-level API. I misread the diff of the commit.

I am a committer, so I could "morph" `ThrowingRunnable` to look more like `Executable`. I'm tempted to keep the name as is. It's likely that many teams will have both JUnit and JUnit-Lambda in their classpath, so having too many classes with the same name is slightly annoying.

marcphilipp:
@kcooney Since `ThrowingRunnable` has not been released yet, we could indeed change it. But what would you want to change? The method name?

sbrannen:
The name `ThrowingRunnable` implies that it is a type of `Runnable` (i.e., that it is intended to be executed in a separate thread). So in that sense the name is IMO very misleading. If you don't rename it to `Executable` in JUnit 4, I would still recommend choosing a more suitable name.

sbrannen:
> It's likely that many teams will have both JUnit and JUnit-Lambda in their classpath, so having too many classes with the same name is slightly annoying.

That's why I asked if you think `Executable` should become a part of the Open Test Alliance.

If we move `Executable` to the OTA, JUnit 4.13, JUnit 5, and any other testing framework could rely on it.

Thoughts?

marcphilipp:
I don't see a lot of benefits, but one major disadvantage: coupling. If we ever wanted to change it, e.g. add default methods to it, we would have to do it in a way that makes sense to all possible users. In addition OTA can only use JDK 6 fearures, right?

jlink:
Hm. Executable is so unrelated to testing that it feels weird to put it into OTA. Moreover, if it's being used lambda-style the concrete type doesn't matter. I'd be happy with Executable in both J4 and J5, it's not the only duplicate name.

sbrannen:
Yes, the OTA can only use JDK 6 language features, so no `default` methods.

`Executable` is certainly generic in the sense that `Runnable` and `Callable` are generic, but the need to be able to throw a `Throwable` from such a code block is more closely related to testing frameworks.

In any case, I'm not going to push to move `Executable` to the OTA. I'm also fine with having it _duplicated_ in J4 and J5. And based on the collective feedback here, I think we can agree to just leave it in JUnit 5 as it is now. :wink:

kcooney:
> The name ThrowingRunnable implies that it is a type of
> Runnable (i.e., that it is intended to be executed in a
> separate thread).

That's a good point (the `Runnable` Javadoc summary specifically mentioned threads). That makes me think that we should rename the JUnit 5 version.

I can think of use cases where this might make sense to have this in OTA (for example, OTA could have an `ErrorCollector` interface that could have a version of `assertThrows()`). That being said,  I don't think I would want JUnit 4.13 to have new dependencies (since it is a 4.x release).

sbrannen:
@kcooney, ummmm.... methinks that's a typo. Surely you meant, "we should rename the JUnit **4** version." Right? :wink:

sbrannen:
@kcooney, have you had a look at our [`ThrowableCollector`](https://github.com/junit-team/junit-lambda/blob/master/junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/ThrowableCollector.java)?

kcooney:
@sbrannen yes, I meant we should rename the JUnit 4 version :-)

I haven't seen `ThrowableCollector` but that's not quite what I had in mind. I was thinking of creating an interface for `ErrorCollector` in OTA and adding a method like this:

    void checkThrows(
        Class<T> expectedThrowable, Executable lambda);

That's the only use case I could see for adding an interface like this to OTA, and it's not an extremely compelling reason

PR #94, opened 2016-01-09
Change Preconditions.notNull() to throw NullPointerException

kcooney:
Change Preconditions.notNull() to throw NullPointerException
instead of IllegalArgumentException. Also updated it to return
the passed-in value.

I thought this might be controversial, so here's my thoughts:

According to Joshua Bloch in Effective Java, 2nd Edition, Item 60,
"If a caller passes null in some parameter for which null values
are prohibited, convention dictates that NullPointerException be thrown
rather than IllegalArgumentException."

Throwing NPE in these cases is also quite common in the JDK.
(see java.util.Collections for many examples).

The Guava Preconditions.checkNotNull() methods also throws a NPE.

Finally, if somehere someone deferences a null variable without
checking it, a NPE would naturally be thrown. If we later add an
explicit check, we wouldn't want the code to start throwing an
IllegalArgumentException.

### DIFF
--- a/junit-commons/src/main/java/org/junit/gen5/commons/util/Preconditions.java
+++ b/junit-commons/src/main/java/org/junit/gen5/commons/util/Preconditions.java
@@ -17,9 +17,6 @@ import java.util.function.Supplier;
  * Collection of utilities for asserting preconditions for method and
  * constructor arguments.
  *
- * <p>Each method in this class throws an {@link IllegalArgumentException}
- * if the precondition fails.
- *
  * @since 5.0
  */
 public final class Preconditions {
@@ -33,10 +30,12 @@ public final class Preconditions {
 	 *
 	 * @param object the object to check
 	 * @param message precondition failure message
+	 * @return the passed in object
+	 * @throws NullPointerException if the object is {@code null}
 	 * @see #notNull(Object, Supplier)
 	 */
-	public static void notNull(Object object, String message) throws IllegalArgumentException {
-		notNull(object, () -> message);
+	public static <T> T notNull(T object, String message) throws NullPointerException {
+		return notNull(object, () -> message);
 	}
 
 	/**
@@ -44,10 +43,15 @@ public final class Preconditions {
 	 *
 	 * @param object the object to check
 	 * @param messageSupplier precondition failure message supplier
+	 * @return the passed in object
+	 * @throws NullPointerException if the object is {@code null}
 	 * @see #condition(boolean, Supplier)
 	 */
-	public static void notNull(Object object, Supplier<String> messageSupplier) throws IllegalArgumentException {
-		condition(object != null, messageSupplier);
+	public static <T> T notNull(T object, Supplier<String> messageSupplier) throws NullPointerException {
+		if (object == null) {
+			throw new NullPointerException(messageSupplier.get());
+		}
+		return object;
 	}
 
 	/**
@@ -55,6 +59,7 @@ public final class Preconditions {
 	 *
 	 * @param str the string to check
 	 * @param message precondition failure message
+	 * @throws IllegalArgumentException if the string is {@code null} or empty
 	 * @see #notEmpty(String, Supplier)
 	 */
 	public static void notEmpty(String str, String message) throws IllegalArgumentException {
@@ -67,6 +72,7 @@ public final class Preconditions {
 	 * @param str the string to check
 	 * @param messageSupplier precondition failure message supplier
 	 * @see StringUtils#isNotEmpty(CharSequence)
+	 * @throws IllegalArgumentException if the string is {@code null} or empty
 	 * @see #condition(boolean, Supplier)
 	 */
 	public static void notEmpty(String str, Supplier<String> messageSupplier) throws IllegalArgumentException {
@@ -78,6 +84,7 @@ public final class Preconditions {
 	 *
 	 * @param collection the collection to check
 	 * @param message precondition failure message
+	 * @throws IllegalArgumentException if the collection is {@code null} or empty
 	 * @see #condition(boolean, Supplier)
 	 */
 	public static void notEmpty(Collection<?> collection, String message) throws IllegalArgumentException {
@@ -89,6 +96,7 @@ public final class Preconditions {
 	 *
 	 * @param str the string to check
 	 * @param message precondition failure message
+	 * @throws IllegalArgumentException if the string is {@code null} or blank
 	 * @see #notBlank(String, Supplier)
 	 */
 	public static void notBlank(String str, String message) throws IllegalArgumentException {
@@ -100,6 +108,7 @@ public final class Preconditions {
 	 *
 	 * @param str the string to check
 	 * @param messageSupplier precondition failure message supplier
+	 * @throws IllegalArgumentException if the string is {@code null} or blank
 	 * @see StringUtils#isNotBlank(String)
 	 * @see #condition(boolean, Supplier)
 	 */
@@ -112,6 +121,7 @@ public final class Preconditions {
 	 *
 	 * @param predicate the predicate to check
 	 * @param message precondition failure message
+	 * @throws IllegalArgumentException if the predicate is {@code false}
 	 * @see #condition(boolean, Supplier)
 	 */
 	public static void condition(boolean predicate, String message) throws IllegalArgumentException {
--- a/junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
+++ b/junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
@@ -155,7 +155,11 @@ class JUnit5TestableFactory {
 	}
 
 	private Class<?> loadClassByName(String className) {
-		return ReflectionUtils.loadClass(className).orElse(null);
+		return ReflectionUtils.loadClass(className).orElseThrow(() -> createCannotLoadClassException(className));
+	}
+
+	private RuntimeException createCannotLoadClassException(String className) {
+		return new IllegalArgumentException(String.format("Cannot load class '%s'", className));
 	}
 
 	private Class<?> loadRequiredClass(String className, String fullUniqueId, String uniqueIdPart) {
--- a/junit-commons/src/main/java/org/junit/gen5/commons/util/CollectionUtils.java
+++ b/junit-commons/src/main/java/org/junit/gen5/commons/util/CollectionUtils.java
@@ -28,8 +28,9 @@ public final class CollectionUtils {
 	 *
 	 * @param collection the collection to get the element from
 	 * @return the only element of the collection
-	 * @throws IllegalArgumentException if the collection is {@code null}
-	 * or does not contain exactly one element
+	 * @throws NullPointerException if the collection is {@code null}
+	 * @throws IllegalArgumentException if the collection does not contain
+	 * exactly one element
 	 */
 	public static <T> T getOnlyElement(Collection<T> collection) {
 		Preconditions.notNull(collection, "collection must not be null");
--- a/junit-commons/src/main/java/org/junit/gen5/commons/util/ToStringBuilder.java
+++ b/junit-commons/src/main/java/org/junit/gen5/commons/util/ToStringBuilder.java
@@ -35,8 +35,7 @@ public class ToStringBuilder {
 	}
 
 	public ToStringBuilder(Class<?> type) {
-		Preconditions.notNull(type, "Class must not be null");
-		this.type = type;
+		this.type = Preconditions.notNull(type, "Class must not be null");
 	}
 
 	public ToStringBuilder append(String name, Object value) {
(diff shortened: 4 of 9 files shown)


jlink:
I tend to disagree. Preconditions IMO are a way to show the programmer that they messed up an explicit contract - DBC-like.  Since an NPE could come from anywhere it doesn't give this message, but I have to check the stack trace to find out. When I introduce a precondition I intentionally want the behaviour to change in a way that makes errors more obvious.

I would even recommend that we use something more explicit like a PreconditionViolatedException or a ProgrammerMessedUpExcpetion.

Von meinem iPad gesendet

junit-commons/src/main/java/org/junit/gen5/commons/util/Preconditions.java
```diff
@@ -67,6 +72,7 @@ public static void notEmpty(String str, String message) throws IllegalArgumentEx
 	 * @param str the string to check
 	 * @param messageSupplier precondition failure message supplier
 	 * @see StringUtils#isNotEmpty(CharSequence)
+	 * @throws IllegalArgumentException if the string is {@code null} or empty
```
marcphilipp: There's a Checkstyle check that fails now:
```
:junit-commons:checkstyleMain[ant:checkstyle] /home/travis/build/junit-team/junit-lambda/junit-commons/src/main/java/org/junit/gen5/commons/util/Preconditions.java:75: error: At-clauses have to appear in the order '[@param, @return, @throws, @exception, @author, @since, @see]'.
```

The tag `@throws` should be above `@see`.

marcphilipp:
Thanks for the pull request, @kcooney!

I like that you've changed it to return the object that was passed in. That's really useful in constructors and setters! :+1: 

As can be seen by numerous discussions on [Stack Overflow](http://stackoverflow.com/questions/3322638/is-it-okay-to-throw-nullpointerexception-programatically) and elsewhere, whether to throw a `NullPointerException` from your own code is a highly controversial topic. I know Joshua Bloch recommends to use it and has done so in the JDK, but I still tend to not using it because it will confuse some (if not a lot of) people.

For me, `IllegalArgumentException` conveys the meaning Johannes is describing by `PreconditionViolatedException` already. But I would not object to make it explicit by introducing such a custom exception type. If we do, I think it should derive directly from `RuntimeException`.

junit-commons/src/main/java/org/junit/gen5/commons/util/Preconditions.java
```diff
@@ -33,10 +30,12 @@ private Preconditions() {
...
 	 * @param message precondition failure message
+	 * @return the passed in object
+	 * @throws NullPointerException if the object is {@code null}
 	 * @see #notNull(Object, Supplier)
 	 */
-	public static void notNull(Object object, String message) throws IllegalArgumentException {
-		notNull(object, () -> message);
+	public static <T> T notNull(T object, String message) throws NullPointerException {
+		return notNull(object, () -> message);
 	}
```
sbrannen: That's a nice improvement. I've seen that elsewhere, but I somehow forgot about implementing it like that in JUnit 5.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
```diff
@@ -155,7 +155,11 @@ private Method findMethod(Class<?> clazz, String methodName, Class<?>[] paramete
 	}
 
 	private Class<?> loadClassByName(String className) {
-		return ReflectionUtils.loadClass(className).orElse(null);
+		return ReflectionUtils.loadClass(className).orElseThrow(() -> createCannotLoadClassException(className));
+	}
```
sbrannen: Why did you make this change?

Do you have a concrete use case for this that fails otherwise?

sbrannen:
I realize that Josh Bloch and certain libraries within the JDK condone the action of throwing NPEs from user code; however, I strongly disagree with this _convention_.

As has been mentioned here and elsewhere, **most** developers interpret a NPE to mean that there is a bug within the invoked code; whereas, an `IllegalArgumentException` conveys something totally different, namely that the developer is using the API incorrectly. 

In JUnit 5, the sole purpose of `Preconditions` is to assert that the caller has obeyed the documented contract, and we want to be **very** clear about any violation of our documented contracts. Throwing an `IllegalArgumentException` is a clear signal to the user; throwing a `NullPointerException` is ambiguous at best and often requires that the developer study or debug the source code to discern the actual cause.

With this in mind, I stand behind my original decision to throw `IllegalArgumentException` consistently within `Preconditions`.

sbrannen:
With regard to introducing a dedicated `PreconditionViolatedException`, I don't feel there is much benefit over a standard `IllegalArgumentException`... unless we introduce precondition checks for purposes other than arguments (e.g., checking state within an algorithm), but in the latter case I would then recommend such methods throw an `IllegalStateException`.

sbrannen:
> Change Preconditions.notNull() to throw NullPointerException instead of IllegalArgumentException. Also updated it to return the passed-in value.

These are two separate concerns. So I'll split off the latter into a dedicated issue.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
kcooney: Yes. There was a test that was expecting `IllegalArguementExxeption` that broke because we tripped on a `notNull()` further down the code.

This is another example of why the difference matters.

kcooney:
I had an extremely hard time finding *any* common JDK API that throws an IEA on an unexpected `null`. Can any of you find examples?

Besides the advice from Joshua Bloch (who was the head architect of the collection APIs) there is the concern of code that should check for `null` but doesn't. If we release code like that, we can't later change it to use the current precondition APIs to provide a useful message, because that would be an externally visible API change.

I understand the argument that the class is named ` Preconditions`, but I don't find that argument compelling. The question is what should our APIs throw if you passed in an unexpected `null`. The JDK has a clear convention here and we should follow it.

Generally if I get a NPE when calling a Java API, that gives me a clear indication that a `null` reference was passed into my API and I didn't check for it. Getting an IAE generally means that I used the API incorrectly.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
sbrannen: Are you actively using JUnit 5 on a project already?

sbrannen: Based on your explanation, this sounds like an actual bug.

Would you be willing to raise a report against it a separate issue or pull request?

If so, that would allow us to merge your fix into `master` rather quickly.

mmerdes:
Looking at NPE's Javadoc ("Thrown when an application attempts to use null in a case where an object is required.") a NPE seems a good fit for a not-null precondition. But then I can also see the argument that we want to communicate an illegal usage of the api in the sense of a breach of contract.
So should we maybe consider to use an NPE as the exception type and use its message to communicate  the breach of contract flavor?

kcooney:
I decided to do some research on how some common third-party libraries handle this.

I looked at Mockito ([Checks.java](https://github.com/mockito/mockito/blob/3ec1008fffad541beab504677db286616fda040b/src/main/java/org/mockito/internal/util/Checks.java)), Joda Time ([LongConverter](https://github.com/JodaOrg/joda-time/blob/5846d0569dcf64cdfd024f6bf58d707f1a679f64/src/main/java/org/joda/time/convert/LongConverter.java) | [ConverterSet](https://github.com/JodaOrg/joda-time/blob/5846d0569dcf64cdfd024f6bf58d707f1a679f64/src/main/java/org/joda/time/convert/ConverterSet.java)), as well as Guava, Spring and Apache Commons Lang ([compared in this blog post](http://www.sw-engineering-candies.com/blog-1/comparison-of-ways-to-check-preconditions-in-java))

Of those, only Spring throws an `IllegalArgumentException` when checking if a parameter is `null`; all others throw `NullPointerException`

In the JDK, besides the Collection classes, parameter checks throw NPE for `null` in [Optional]([https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#of-T-) and [Consumer](https://docs.oracle.com/javase/8/docs/api/java/util/function/Consumer.html#andThen-java.util.function.Consumer).

In addition, many JDK interfaces that we might implement require throwing a NPE on `null`, including [Comparable](https://docs.oracle.com/javase/8/docs/api/java/lang/Comparable.html#compareTo-T-) and [Executor](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executor.html#execute-java.lang.Runnable-). If we kept our `Precoditions` class as-is, we couldn't use it when implementing APIs like these.

Reading the `NullPointerException` Javadoc also makes it clear that this is an expected use of that exception:

> Applications should throw instances of this class to indicate other illegal uses of the null object.

Finally, there is the JDK `Objects` class, which has this method:

    public static <T> T requireNonNull(T obj, String message)

The Javadoc shows states this:

> This method is designed primarily for doing parameter validation in methods and constructors with multiple parameters, as demonstrated below:

...with this code snippet:

    public Foo(Bar bar, Baz baz) {
         this.bar = Objects.requireNonNull(bar, "bar must not be null");
         this.baz = Objects.requireNonNull(baz, "baz must not be null");
     }

`Objects.requireNonNull` throws a `NullPointerException` when a `null` was passed in. In JDK 8, they even added this method:

    public static <T> T requireNonNull(T obj, Supplier<String> messageSupplier)

...which is the exact signature of `Preconditions.notNull()` now.

In summary, the JDK gives a strong preference for throwing a `NullPointerException` when an unexpected `null` is passed to a method or constructor. Many common third-party libraries follow this convention. Therefore, most users should expect this behavior when calling methods. I think JUnit-Lambda APIs should do the same.

I don't think we should introduce our own exceptions for these cases. The existing ones are well-known, and there are even [answers on Stack Overflow](http://stackoverflow.com/questions/218384/what-is-a-null-pointer-exception-and-how-do-i-fix-it) detailing what to do when you see the exception being thrown

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
kcooney: I'll send a separate pull with just this change and an update to the associated test. It might take a few days for me to have time though :-)

sbrannen: Great!

thanks

sbrannen:
Hi everybody,

Thanks to your very valuable and very detailed input, this discussion has resulted in the following changes.

1. Methods in `Preconditions` now return the original value when feasible (see #95).
1. A bug has been discovered in `JUnit5TestableFactory` for which @kcooney will soon submit a fix.
1. We have introduced a dedicated `JUnitException` hierarchy, and all methods in `Preconditions` now throw `PreconditionViolationException` (see #97).

As explained in #97, after thorough discussions of the possible solutions and their related pros and cons, the JUnit 5 team came to a consensus this week that methods in `Preconditions` will throw _neither_
`IllegalArgumentException` _nor_ `NullPointerException`. The rationale for this decision is ultimately quite simple: JUnit is a testing framework, and that makes it _very_ special with regard to exception handling and propagation. Thus, in order to completely avoid ambiguities with exceptions that are typically thrown by the JDK, third-party libraries, or client code, JUnit code will only throw JUnit-specific unchecked exceptions.

This decision brings with it the following benefits.

1. Tests which assert that an `IllegalArgumentException` or `NullPointerException` was thrown will never suffer from a false positive due to improper usage of JUnit APIs.
1. The presence of a `JUnitException` (or subclass thereof) in a stack trace will make the cause readily apparent to the viewer.
1. Any `IllegalArgumentException`, `NullPointerException`, or related JDK exception thrown from JUnit code will typically point to an actual bug within the JUnit code base.

Due to the fact that #95 and #97 have already been closed, I will now close this issue as well.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestableFactory.java
sbrannen: Hi @kcooney,

Any chance you'll get around to creating that issue/PR this week?

If not, I think we'll go ahead and push the proposed change anyway.

cheers

Sam

kcooney: @sbrannen I'll send a PR today or tomorrow

sbrannen: Aha. I see you submitted #139.  Cheers!

PR #119, opened 2016-01-16
Feature #116 - Provide better names for the test plan specification

bechte:
I will need to rebase it onto master, but don't have time right now. Still, I wanted to provide it to you guys for review purposes. Therefore, I've already created this pull request.

PR #120, opened 2016-01-16
Rebase of #119

marcphilipp:
This is merely a rebase of #119 onto master.


bechte:
Thanks for rebasing! :+1: Shall we merge this, and I will resolve the open issues from #119 in master directly? I will have time tomorrow to solve it. This might be the easiest way. What do you think?

marcphilipp:
Sounds good to me!

PR #129, opened 2016-01-20
Introduced interface Launcher and renamed old to JUnit5Launcher

jlink:
This was done to address yesterday's discussion that `Launcher` should be a top-level abstraction in module `launcher-api`.

Moreover, it further decouples clients from implementation.

### DIFF
--- a/junit-launcher/src/main/java/org/junit/gen5/launcher/main/package-info.java
+++ b/junit-launcher/src/main/java/org/junit/gen5/launcher/main/package-info.java
@@ -1,8 +1,8 @@
 /**
- * The {@link org.junit.gen5.launcher.main.Launcher Launcher} class is the
+ * The {@link org.junit.gen5.launcher.main.JUnit5Launcher Launcher} class is the
  * main starting point for running all JUnit tests.
  *
- * <p>The {@link org.junit.gen5.launcher.main.DiscoveryRequestBuilder
+ * <p>The {@link org.junit.gen5.launcher.main.TestDiscoveryRequestBuilder
  * DiscoveryRequestBuilder} serves as a small DSL for creating
  * {@link org.junit.gen5.launcher.TestDiscoveryRequest TestDiscoveryRequests}.
  */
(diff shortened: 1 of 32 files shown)


sbrannen:
The Javadoc in several locations needs to be updated to reference the `Launcher` instead of the concrete `JUnit5Launcher`.

jlink:
Hopefully all locations found.

2016-01-20 13:52 GMT+01:00 Sam Brannen <notifications@github.com>:

junit-launcher/src/main/java/org/junit/gen5/launcher/main/package-info.java
```diff
@@ -1,8 +1,8 @@
 /**
- * The {@link org.junit.gen5.launcher.main.Launcher Launcher} class is the
+ * The {@link org.junit.gen5.launcher.main.JUnit5Launcher Launcher} class is the
```
sbrannen: Half missed this one. :wink:

marcphilipp:
I think separating the interface from the implementation is a good idea even though we will only provide a single implementation. 

However, I think the name `JUnit5Laucher` is misleading since it can be used to execute tests for any engine. How about `DefaultLauncher`?

sbrannen:
> However, I think the name `JUnit5Laucher` is misleading since it can be used to execute tests for any engine. How about `DefaultLauncher`?

I agree: `DefaultLauncher` or `StandardLauncher` would be more more suitable names.

jlink:
I also thing that the number 5 should mostly be restricted to
"junit5-engine". "DefaultLauncher" suggests that many exist, what about
just "JUnitLauncher"?

2016-01-20 16:33 GMT+01:00 Marc Philipp <notifications@github.com>:

marcphilipp:
Well, we (or someone else) might create another one at some point, e.g. A `RemoteLaucher` proxy, or a new one that uses a different engine registration model.

jlink:
So, do you vote for DefaultLauncher?

I could do the change in the PR. Problem, with all the polishing going on,
rebasing becomes an increasing problem since the PR is all over the place.

2016-01-20 17:08 GMT+01:00 Marc Philipp <notifications@github.com>:

marcphilipp:
At present, yes. 😀

I wouldn't mind if you did the rename in master after merging this pull request.

jlink:
Since I fixed all know errors in the PR and just rebased it, feel free to
go ahead and accept it.

2016-01-20 17:18 GMT+01:00 Marc Philipp <notifications@github.com>:

sbrannen:
`DefaultLauncher` sounds good.

Thanks for making the changes in the branch, @jlink.

I'll merge this into `master` this evening.

jlink:
Cheers.

Von meinem iPad gesendet

sbrannen:
Merged into `master` in c3c93beab822d6ea83683d53bcc7456d74f2d59e.

PR #134, opened 2016-01-24
Resolver based discovery mechanism

bechte:
This pull request contains all changes that are required to switch to another discovery process. The new process is built upon resolvers and follows a clear structure for the unique id. It encapsulates all logic that is required to create and map a `TestDescriptor` with an guaranteed unique id within the `TestResolver`.

Advantages over the current discovery process:
- No hard-coded rules for test discovery: The current approach needs to be changed in code in order to add new `TestDescriptors` within the test plan tree. The new approach allows to add children to a given parent any time during the test discovery phase. In this way, extension provider could also write their own `TestResolvers` that can parse arbitrary syntax trees (e.g. lambda, XML files, etc.)
- the concept of a `TestResolver` simplifies the dependencies: A `TestResolver` has no dependency on other `TestResolver`s. It only knows about the `TestDescriptor` that it want to contribute new children to, and knows how to describe them in form of a `DiscoverySelector`. In this way, discovery can be done top-down (recursively) or bottom-up (exact match only).
- it provides better maintainability by adding a clear separation of concerns: A `TestResolver` is responsible for creating the executable `TestDescriptors` and to resolve the `UniqueId` for it. It needs not to know about any other `TestResolver` in the system, but only about which `TestDescriptor` it wants to add children to.
- Concentration of rules for `UniqueId`s: `TestResolver`s are responsible for creating and resolving `UniqueId`s. No other module will need to contain rules for modifying or changing an `UniqueId`.
- easier integration of `DiscoveryFilter`: Each `TestResolver` might provide its own `DiscoveryFilters` that it will apply during discovery. This gains performance, since a `DiscoveryFilter` is applied upfront by its `TestResolver` and no `TestDescriptor` needs to be generated, at all. `DiscoveryFilter` might be understood as 'Tuning or Configuration parameters' for the `TestResolvers`

Team decision: Everybody should review this pull request until the next team Skype call on Tuesday. Then we will decide whether or not we will switch to the new discovery process.


bechte:
Team decision:
- We choose to take pull request #218 instead of this one to be able to ship M1 faster.
- Therefore, this PR will be closed and not merged
- I will open a new issue for extending the behaviour according to this PR that we can work on in a later milestone

Issue #112, opened 2016-01-13
Revisit extension registration model


sbrannen:
## Status Quo

The extension registration model is currently a work in progress and needs to be revisited to make sure it aligns with the needs of the extension authors and the community at large.

## Topics to Review and Revisit

- [x] `TestExtension` vs. `ExtensionPoint`
  - What purpose does `TestExtension` serve since the introduction of `ExtensionPoint`? Do we need two maker interfaces for extensions?
- [x] `ExtensionPoint` vs `RegisteredExtensionPoint`
- [x] `ExtensionPointSorter`
  - Why limit sorting to implementations of `ExtensionPoint`? Why not allow any type of `TestExtension` to be sorted?
- [x] `ExtensionRegistry` vs. `TestExtensionRegistry`
  - At the very least, the names are both misleading and confusing.
  - An `ExtensionRegistry` currently allows one to register _only_ instances of `ExtensionPoint`. So why isn't it called an `ExtensionPointRegistry`?
- [x] `ExtensionPoint.Position`
  - Why is `Position` specific to `ExtensionPoint`?
  - An extension author would likely never want an extension to be invoked "in the middle" (i.e., in the `DEFAULT` position) alongside user code. So what purpose does `DEFAULT` serve?
  - `Position` defines only 5 options for extension authors. Why should JUnit be so restrictive? Why not allow any numerical ordering (e.g., the entire range of integers, perhaps other than `0`) beyond the hard coded 5 current values?
  - Consider a solution such as `@Order` that can be reused across multiple use cases within the framework.
- [x] Verify (via appropriate testing) that the current algorithm properly sorts extensions when multiple extensions exist within test class hierarchies and within a given extension execution context.
- [x] The current extension model allows _additional_ extensions to be registered programmatically via the `ExtensionRegistry`; however, there is no mechanism for _overriding_ extensions that have already been registered (e.g., in superclasses).
  - We need a way to override extensions that have been registered in an enclosing context.

jlink:
Some team decisions:
- [ ] Rename `TestExtension` to `Extension`. 
- [ ] `ExtensionRegistrar` implements `Extension`

jlink:
A thing we talked about but did not decide on:
- [ ] Reintroduce extension name into extension registration - and `RegisteredExtensionPoint` to allow for better logging and reporting.

sbrannen:
> Reintroduce extension name into extension registration - and RegisteredExtensionPoint to allow for better logging and reporting.

[...]

PR #135, opened 2016-01-24
Introduce ExtensionPointConfiguration

jlink:
This is an attempt to tackle some of the points discussed in #112 .

Every concrete subtype of `ExtensionPoint` can now register a configuration of type `ExtensionPointConfiguration`. For example:

```java
public interface BeforeEachExtensionPoint extends ExtensionPoint...
	/**
	 * Configuration for {@code BeforeEachExtensionPoint}
	 */
	ExtensionPointConfiguration CONFIG = new ExtensionPointConfiguration(
		new ExtensionPointRegistry.Position[] { OUTERMOST, OUTSIDE_DEFAULT, DEFAULT, INSIDE_DEFAULT, INNERMOST },
		ExtensionPointRegistry.ApplicationOrder.FORWARD);
```
This configuration is used in `ExtensionRegistry` two things:
- Determine the application order of an extension point. Formerly that was done by the applier of an extension point who had to know.
- Allow only certain values for `Position` when registering an extension. All other values will lead to an `ExtensionConfigurationException`.

The concept has been introduced, because not all values make sense for all types of extension points. For example, `ExceptionHandlerExtensionPoint` allows `FIRST`, `DEFAULT` and `LAST`.

Extension point types that do not come with their own configuration only allow `DEFAULT` as position and are thus ordered by their position in the source only.

### DIFF
--- a/junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
+++ b/junit-commons/src/main/java/org/junit/gen5/commons/util/ReflectionUtils.java
@@ -253,6 +253,31 @@ public final class ReflectionUtils {
 		return Arrays.stream(clazz.getDeclaredClasses()).filter(predicate).collect(toList());
 	}
 
+	//TODO: Search in hierarchy
+	public static Optional<Field> findField(Class<?> clazz, String fieldName) {
+		try {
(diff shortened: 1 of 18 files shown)


marcphilipp:
I'm closing this issue since we've decided not to ship any support for extension ordering in M1.

PR #19, opened 2016-02-02
Introduce support for before/after callbacks once per entire test run

ttddyy:
Initially wrote it [here](https://github.com/junit-team/junit-lambda/wiki/OLD-Prototype-Feedback#lack-of-beforealltests-afteralltests-or-suitetags-level-annotations).

Nice to have callback interfaces once per entire test run or engine/tag/etc. (e.g.: _@BeforeAllTests/@AfterAllTests_)
Would be nice that they receive `TestExecutionContext` in argument, so that implementations can populate beans which could be resolved/used/injected across all tests.

### DIFF
--- a/junit-engine-api/src/main/java/org/junit/gen5/engine/GlobalExtensionPoint.java
+++ b/junit-engine-api/src/main/java/org/junit/gen5/engine/GlobalExtensionPoint.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.gen5.engine;
+
+import java.util.Map;
+
+/**
+ * @author Tadaya Tsuyukubo
+ */
+public interface GlobalExtensionPoint {
+
+	default void beforeExecute(Map<String, Object> requestAttributes) {
+		// no-op
+	}
+
+	default void afterExecute(Map<String, Object> requestAttributes) {
+		// no-op
+	}
+
+}
--- a/junit-launcher/src/main/java/org/junit/gen5/launcher/main/DefaultLauncher.java
+++ b/junit-launcher/src/main/java/org/junit/gen5/launcher/main/DefaultLauncher.java
@@ -10,13 +10,18 @@
 
 package org.junit.gen5.launcher.main;
 
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
+import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.logging.Logger;
 
 import org.junit.gen5.commons.JUnitException;
+import org.junit.gen5.commons.util.ReflectionUtils;
 import org.junit.gen5.engine.ExecutionRequest;
 import org.junit.gen5.engine.FilterResult;
+import org.junit.gen5.engine.GlobalExtensionPoint;
 import org.junit.gen5.engine.TestDescriptor;
 import org.junit.gen5.engine.TestEngine;
 import org.junit.gen5.launcher.Launcher;
@@ -95,11 +100,25 @@ class DefaultLauncher implements Launcher {
 		testExecutionListener.testPlanExecutionStarted(testPlan);
 		ExecutionListenerAdapter engineExecutionListener = new ExecutionListenerAdapter(testPlan,
 			testExecutionListener);
+
+		Iterable<GlobalExtensionPoint> globalExtensionPoints = getGlobalExtensionPoints();
+
+		Map<String, Object> requestAttributes = new HashMap<>();
+		globalExtensionPoints.forEach(globalExtensionPoint -> globalExtensionPoint.beforeExecute(requestAttributes));
+
 		for (TestEngine testEngine : root.getTestEngines()) {
 			TestDescriptor testDescriptor = root.getTestDescriptorFor(testEngine);
-			testEngine.execute(new ExecutionRequest(testDescriptor, engineExecutionListener));
+			ExecutionRequest request = new ExecutionRequest(testDescriptor, engineExecutionListener);
+			request.getAttributes().putAll(requestAttributes);
+			testEngine.execute(request);
 		}
+
+		globalExtensionPoints.forEach(globalExtensionPoint -> globalExtensionPoint.afterExecute(requestAttributes));
 		testExecutionListener.testPlanExecutionFinished(testPlan);
+	}
+
+	private Iterable<GlobalExtensionPoint> getGlobalExtensionPoints() {
+		return ServiceLoader.load(GlobalExtensionPoint.class, ReflectionUtils.getDefaultClassLoader());
 	}
 
 }
--- a/junit5-api/src/main/java/org/junit/gen5/api/extension/EngineExtensionPoint.java
+++ b/junit5-api/src/main/java/org/junit/gen5/api/extension/EngineExtensionPoint.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.gen5.api.extension;
+
+/**
+ *
+ * @author Tadaya Tsuyukubo
+ */
+public interface EngineExtensionPoint extends ExtensionPoint {
+
+	default void beforeEngine(ExtensionContext extensionContext) {
+		// no-op
+	}
+
+	default void afterEngine(ExtensionContext extensionContext) {
+		// no-op
+	}
+
+}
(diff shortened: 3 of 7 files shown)


ttddyy:
Hi,

I wrote POC(or a pull request) for per engine extension point as well as global extension point using `ServiceLoader`.
I attache the commit to this issue above.

I have mostly focused on how to populate `ExtensionContext#Store`.
Many times, I had usecases to create objects very beginning of all tests (and may clean-up at the very end). 
For example, create a `HttpClient` and uses(inject) it into all tests(may be resolve to method parameter in junit5 way).

So, it doesn't do anything with `TestExecutionListener` but may be tweaked to do something about it or have separate mechanism for listeners.



some of the items need to think about:
- should this be named extension point?
  - mechanism is in a way different from other extension points.
- which package to place extension point APIs?

Thanks,

jlink:
I might be missing the right context but a global/per engine callback should IMO be outside the JUnit 5 API so ExtensionContext wouldn't be touched at all. Instead it would live in the launcher or engine module.

sbrannen:
@jlink, there are numerous use cases where a global/per engine callback would need to provide something to tests or extensions (e.g., server name, port number, JDBC connection URL, a physical `DataSource`, etc.).

This is what @ttddyy is talking about with regard to the _store_ in the `ExtensionContext`. Such a global callback may not interact directly with the `ExtensionContext`, but it would need a mechanism for propagating information to the tests and extensions.

ttddyy:
@sbrannen Thanks for the explanation. Yes, that was my intention.

For global callback, I used `ExecutionRequest#getAttributes`, which is not used in current master, as a means to pass objects from callback to root level `ExtensionContext`.

jlink:
Sorry for my premature comment earlier. I had a look at the code now and think I get the basic point.

It's a big change in so far that it introduces a new - and IMO very strong - coupling, namly pushing information, uncontrolled by JUnit, from the outside to an inside test extension. This has a couple of drawbacks:
- A test extension might only work if a global execution point has succeeded to do its task and put in the proper information using the proper key. This will make it more difficult to run individual tests or part of a test suite.
- A global extension that does essential things - like starting a web server etc - is doing magic from the POV of the test code itself. A reader will not see this dependency just by reading the code.
- Some tools, e.g. maven-sure-fire, have means to parallelize test execution by starting individual JVMs. This opens a whole new can of potential problems with global extension points.
- Registering global extensions using a service provider makes it difficult to switch them on and off in cases of conflict.

I think we'll have to discuss this suggestion thoroughly to see if the advantages outweigh the drawbacks.

ttddyy:
Aside from how to implement, I believe there are two requirements for this:

- callback(or extension point) for before/after entire test run
  - may or may not be analogous to `@BeforeSuite`, `@AfterSuite` in TestNG
- having parent/root ExtensionContext (parent of `ClassBasedContainerExtensionContext`) and populate/pass it

I think these two points can be combined together or can be separated.

Another approach I can think of is having interfaces like this:

```java

// extension point for global before/after
public interface GlobalCallback {
  void beforeExecute(...);
  void afterExecute(...);
}


public interface ExtensionContextSourceProvider {

  // return maybe Map<String, Object> but additionally needs to give id/name of it
  ExtensionContext createTestContext();
}


// this class needs to be registered or discovered in discovery phase
public class MyTestContextProvider implements GlobalCallback, ExtensionContextSourceProvider {

  @Override
  public void beforeExecute(...) {
  }

  @Override
  public TestContext createTestContext() {
     // create context & populate data for ExtensionContext#store
     // the context should have id/name.
     //   ex: "my-context" specified in MyExtensionPoint below
     return ...;
  }

  @Override
  public void afterExecute(...) {
  }

}


@ParentExtensionContext("my-context")  // specify parent context
public class MyExtensionPoint implements BeforeAllExtensionPoint, AfterAllExtensionPoint {

   @Override
   void beforeAll(ContainerExtensionContext context) throws Exception {
       // the given context should have parent context set
   };

   ....
}


```

All classes that implements `GlobalCallback` needs to be discovered/registered before test execution, so that it can be invoked before running tests.  Any tool that runs test class needs to be aware of the class which implements `GlobalCallback`. (IDE, surefire, etc when specifying single test class/method to run)

`ExtensionContextSourceProvider` is to return `ExtensionContext` or just a map which goes to `ExtensionContext#Store`.

`@ParentExtensionContext` annotation is used in class level `ExtensionPoint` (`BeforeAllExtensionPoint` or `AfterAllExtensionPoint`).
When it is used, specified name of `ExtensionContext` or map(`ExtensionContext#Store`) would be set as a parent `ExtensionContext`.
Now `ExtensionPoint` annotated with `@ParentExtensionContext` has dependency to the specified context, `ExtensionContextSourceProvider` needs to be called beforehand, which I think it is natural that all `ExtensionContextSourceProvider` get invoked before starting tests which overwraps to the feature of `GlobalCallback`.



> Some tools, e.g. maven-sure-fire, have means to parallelize test execution by starting individual JVMs. This opens a whole new can of potential problems with global extension points.

I believe this is certainly a good point to discuss. 
I just wonder how it is handling `@BeforeSuite` and `@AfterSuite` in TestNG. I haven't verified but my guess is each JVM performs `@BeforeSuite` and `@AfterSuite`.

marcphilipp:
Closed in favor of #671.

RomanIovlev:
@marcphilipp Is any way in JUnit 5 now to Run some code once per test run? Or you completely deny this ability?

marcphilipp:
@RomanIovlev Yes, there is a way to do that: you can write an extension like the following and [register it globally](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-automatic).

```java
public class OncePerTestRunExtension implements BeforeAllCallback {
    @Override
    public void beforeAll(ExtensionContext context) throws Exception {
        context.getRoot().getStore(ExtensionContext.Namespace.GLOBAL)
            .getOrComputeIfAbsent(MyFixture.class);
    }

    static class MyFixture implements ExtensionContext.Store.CloseableResource {
        public MyFixture() {
            // setup
        }

        @Override
        public void close() {
            // teardown
        }
    }
}
```

RomanIovlev:
Thanks, but Why you don't want to implement this as Annotation?
Call it BeforeSuite/AfterSuite or BeforeTestRun/AfterTestRun whatever...
In testing this is quite often to have something before all tests in test run and sometimes more important cleanup this data after all tests. code before/after all tests in class not so popular...

marcphilipp:
How would we discover classes with such annotations? It sounds like we'd have to scan the classpath even if only a single class was requested.

SergiiNik:
hi All!

Above advises do **not** work for me, so I solved this problem like this:

Add to your Base abstract class (I mean abstract class where you initialize your driver in *setUpDriver()* method) this part of code:

    private static boolean started = false;
    static{
        if (!started) {
            started = true;
            try {
                setUpDriver();  //method where you initialize your driver
            } catch (MalformedURLException e) {
            }
        }
    }

And now, if your test classes will **extends** from Base abstract class -> *setUpDriver()* method will be executed before first @Test only **ONE** time per project run.

jbwyatt4:
I have written an extension. Just to be sure it isn't a pom.xml issue. How would you add the junit.jupiter.extensions.autodetection.enabled property=true to the maven-surefire-plugin?

This is the XML I have so far in pom.xml
```
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>3.0.0-M3</version>
    <configuration>
        <systemPropertyVariables>
            <junit.jupiter.extensions.autodetection.enabled>true</junit.jupiter.extensions.autodetection.enabled>
        </systemPropertyVariables>
    </configuration>
</plugin>
```

marcphilipp:
@jbwyatt4 That looks like it should work. Could you debug or print all system properties in your test to see if Surefire hands them to the forked JVM?

jbwyatt4:
The XML property reports being true.
junit.jupiter.extensions.autodetection.enabled: true

Posting code for extension class:

```
@ExtendWith({YourJUnitExtension.class})
public class YourJUnitExtension implements BeforeAllCallback, ExtensionContext.Store.CloseableResource {
  private static boolean started = false;
  
  @Override
  public void beforeAll(ExtensionContext context) {
    if (!started) {
      started = true;
      CU.enable(); // custom class to control which print statements print
      CU.print("Print from Extension Context");
      CU.disable();
    }
  }
  
  @Override
  public void close() {
    // Your "after all tests" logic goes here
  }
}
```

marcphilipp:
If os, the `beforeAll` method should be called. Are you sure it isn't?

However, for `close` to be called, the resource needs to be added to an `ExtensionContext.Store`. Moreover, the `@ExtendWith` on the extension does not make sense as you cannot yet extend extensions and to extend sth. with itself would put you into an endless loop. 😉 

This should work and eventually be documented in #1555:

```java
public class YourJUnitExtension implements BeforeAllCallback {

  @Override
  public void beforeAll(ExtensionContext context) {
    ExtensionContext.Store rootStore = context.getRoot().getStore(ExtensionContext.Namespace.GLOBAL);
    rootStore.getOrComputeIfAbsent("myResource", key -> {
      CU.enable(); // custom class to control which print statements print
      CU.print("Print from Extension Context");
      CU.disable();
      return new MyResource();
    });
  }

  static class MyResource implements ExtensionContext.Store.CloseableResource {
    @Override
    public void close() throws Throwable {
      // Your "after all tests" logic goes here
    }
  }
}
```

marcphilipp:
@jbwyatt4 If you have additional questions, please let's continue the discussion in #1555 since this issue is closed.

jbwyatt4:
Just got the notification from GitHub on this.

Your code works. I believe you were right about @ExtendWith.

Thank you for your help @marcphilipp.

PR #176, opened 2016-03-09
Use the current Gradle 2.11 instead of 2.10.

stefanbirkner:
## Overview

Use the current Gradle 2.11 instead of 2.10.

---


marcphilipp:
Thanks!

PR #178, opened 2016-03-10
Just caches list size for performance improvement.

hotchemi:
## Overview

- Just a tiny little bit performance improvement.

---


marcphilipp:
A generally accepted rule regarding performance improvements is: no premature optimization. Do you have experienced a performance problem there? Because in 99.99% `size()` will return `1` and thus only called once. In addition, `ArrayList#size` does not compute anything but just returns the value of an instance variable. So even if it were called more often, it probably wouldn't make a difference.

hotchemi:
Umm, actually I've not experienced any problem. But I think there is no demerit when you approve this change right? It's just an optimization but doesn't lost any interoperability.

marcphilipp:
Well, we don't know if it's actually an optimization. And it makes the code slightly harder to read. Therefore, I would rather not merge it.

hotchemi:
OK.

PR #179, opened 2016-03-10
Add space before method name.

hotchemi:
## Overview

- Add space before method name.

---

### DIFF
--- a/junit4-runner/src/main/java/org/junit/gen5/junit4/runner/Classes.java
+++ b/junit4-runner/src/main/java/org/junit/gen5/junit4/runner/Classes.java
@@ -37,6 +37,6 @@ public @interface Classes {
 	/**
 	 * One or more test classes.
 	 */
-	Class<?>[]value();
+	Class<?>[] value();
 
 }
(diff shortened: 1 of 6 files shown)


hotchemi:
Oh is this your formatter rule? Didn't know that....

marcphilipp:
The build fails because our formatter settings are enforced by the Spotless Gradle plugin. However, I would actually like to have spaces where you've added them. If you want, you might want to look into changing the formatter settings.

hotchemi:
Sorry I'm not sure which id is correct to fix in the junit-eclipse-formatter-settings.xml. Could u point out soon?

sbrannen:
We are currently using Spotless 1.3.2 which uses the formatting functionality of Eclipse 4.5, and the behavior seen here is due to a [bug in Eclipse 4.5](https://bugs.eclipse.org/bugs/show_bug.cgi?id=472962#c5).

This bug has been fixed in Eclipse 4.5.1 and 4.6.

Thus, we will simply have to wait on a release of Spotless that uses Eclipse 4.5.1+.

sbrannen:
FYI: I have raised an issue with Spotless: https://github.com/diffplug/spotless/issues/15

marcphilipp:
Thanks, Sam!

sbrannen:
This pull request has been superseded by #183.

I am therefore closing this PR.

hotchemi:
OK.

Issue #104, opened 2016-01-12
DefaultLauncher must ensure uniqueness of engine IDs


sbrannen:
## Status Quo

`DefaultLauncher.discoverRootDescriptor()` currently delegates to a `TestEngineRegistry` to retrieve all registered test engines; however, no check is performed to ensure that no two engines claim to be the same engine (i.e., have the same unique ID).

## Deliverables

- [ ] Ensure the uniqueness of engine IDs for registered engines within the `DefaultLauncher`.

sbrannen:
Resolved in commit bc1770895674f4b22c7db1068d2b33b3699f9aa8.

PR #184, opened 2016-03-10
Engines for DefaultLauncher must have different ids.

stefanbirkner:
## Overview

A `DefaultLauncher` with `TestEngine`s that have equal ids is unusable. It fails to discover or execute tests. Hence it should not be possible to create such a launcher.

---

### DIFF
--- a/junit-tests/src/test/java/org/junit/gen5/launcher/main/DefaultLauncherTests.java
+++ b/junit-tests/src/test/java/org/junit/gen5/launcher/main/DefaultLauncherTests.java
@@ -46,10 +46,8 @@ class DefaultLauncherTests {
 
 	@Test
 	void constructLauncherWithMultipleTestEnginesWithDuplicateIds() {
-		DefaultLauncher launcher = createLauncher(new DummyTestEngine(), new DummyTestEngine());
-
 		JUnitException exception = expectThrows(JUnitException.class,
-			() -> launcher.discover(request().select(forUniqueId("foo")).build()));
+			() -> createLauncher(new DummyTestEngine("dummy id"), new DummyTestEngine("dummy id")));
 
 		assertThat(exception).hasMessageContaining("multiple engines with the same ID");
 	}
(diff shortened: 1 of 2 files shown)


junit-tests/src/test/java/org/junit/gen5/launcher/main/DefaultLauncherTests.java
```diff
@@ -46,12 +47,8 @@ void discoverEmptyTestPlanWithoutAnyEngines() {
 
 	@Test
 	void constructLauncherWithMultipleTestEnginesWithDuplicateIds() {
-		DefaultLauncher launcher = createLauncher(new DummyTestEngine(), new DummyTestEngine());
-
-		JUnitException exception = expectThrows(JUnitException.class,
-			() -> launcher.discover(request().select(forUniqueId("foo")).build()));
-
-		assertThat(exception).hasMessageContaining("multiple engines with the same ID");
```
sbrannen: Please add back verification of the contents of the exception message.

sbrannen:
How did you instantiate a `DefaultLauncher` with engines with duplicate IDs?

sbrannen:
This is related to #104.

junit-tests/src/test/java/org/junit/gen5/launcher/main/DefaultLauncherTests.java
stefanbirkner: Fixed.

stefanbirkner:
Two `TestRegistry`s can have the same id if the author overrides `getId()`. This leads to two questions:
* should JUnit care about not having `TestEngine`'s with the same id? (= remove the code for checking this)
* is this the responsibility of `DefaultLauncher`? (= move the code to another class)

What do you think?

sbrannen:
First of all, `ServiceLoaderTestEngineRegistry` is a private implementation detail of how engines are discovered. So there cannot be more than one.

In addition, `DefaultLauncher` is also a private implementation detail, only obtainable via a static factory method in `LauncherFactory`.

Thus, at the moment, it is not possible for multiple engines to be registered with the same ID... at least not in the default case.

To answer your first question: Yes, JUnit must care about uniqueness of engine IDs. That's why I created issue #104.

Regarding your second question: one could argue that the validation logic should be moved to `LauncherFactory.create()`; however, I think the best solution is to make it part of the contract of any `Launcher` implementation. I will therefore update the Javadoc for `Launcher` accordingly.

Cheers,

Sam

sbrannen:
FYI: I updated the Javadoc in https://github.com/junit-team/junit5/commit/426468f5b24969dcd0cccdd9c225dd60a69a0763.

stefanbirkner:
Thanks for the explanation and updating the Javadoc.

PR #185, opened 2016-03-10
Explicitly join ids with ", ".

stefanbirkner:
## Overview

IMHO this is a better approach then printing a list.

---


marcphilipp:
Thanks! :-)

PR #187, opened 2016-03-11
Add assertEquals for primitive types

### LINKED ISSUE #110: Add overloaded assertEquals methods for primitive types

hotchemi:
## Overview

- This pr fixes #110 .
- I just added assetEquals for following primitive types.
  - byte
  - short
  - int
  - long
  - char
  - float
  - double

---

### DIFF
--- a/junit-tests/src/test/java/org/junit/gen5/api/AssertionsTests.java
+++ b/junit-tests/src/test/java/org/junit/gen5/api/AssertionsTests.java
@@ -275,6 +275,302 @@ public class AssertionsTests {
 	// --- assertEquals -------------------------------------------------
 
 	@Test
+	void assertEqualsByte() {
+		byte expected = 1;
+		byte actual = 1;
+		assertEquals(expected, actual);
+	}
+
+	@Test
+	void assertEqualsByteWithOneVsTwo() {
+		byte expected = 1;
+		byte actual = 2;
+		try {
+			assertEquals(expected, actual);
+			expectAssertionFailedError();
+		}
+		catch (AssertionFailedError ex) {
+			assertMessageEquals(ex, "expected: <1> but was: <2>");
+		}
+	}
+
+	@Test
+	void assertEqualsByteWithOneVsTwoAndMessage() {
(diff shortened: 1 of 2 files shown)


bechte:
Thank you for the pull request. Would you please add additional tests for values that are not equal, as well as tests for the calls to assertEquals(int,int,String) and assertEquals(int,int,Supplier<String>) accordingly? Please see the tests for the other methods. Thanks.

hotchemi:
@bechte 
Thx, I added test cases.
One point I'd like to confirm is that I use `org.junit.gen5.api.Assertions#failNotEqual` internal assertion but is it good design? failNotEqual converts primitive arguments to String so it might take a cost than normal primitive comparison.

bechte:
@hotchemi: Actually, for the assertion you use the int primitives. `failNotEqual` is only called in case of this comparison fails. I don't think it is a tremendous overhead, especially because Integers are being cached and the auto-boxing feature makes use of this cache. Plus, this only happens for failing tests. So, I don't have any concerns in this directions.

bechte:
@hotchemi: Thanks for adding the tests. I think the `int` case is ready to be merged. Would it be possible for you to also add support for other primitive types as of #110 ? That would be great. If not, we probably merge this PR together with the changes from #110 .

hotchemi:
@bechte Sure. I'll add another methods and test cases on this pull request tonight. So changed the title to work in progress. Thank you for your review.

hotchemi:
@bechte I added following types of methods. Could you review?

- byte
- short
- int
- long
- char
- float
- double

sbrannen:
@hotchemi, the work now looks complete. Thanks!

Before we merge it, however, can you please squash all commits into a single commit (e.g., via `git rebase --interactive --autosquash`) and force push your changes for this PR?

hotchemi:
@sbrannen Did it. Sure :smile:

sbrannen:
@hotchemi, thanks!

We will merge this in the near future.

hotchemi:
:bow:

PR #191, opened 2016-03-11
Refactor getDisplayName of JUnit4TestDescriptor.

hotchemi:
## Overview

- Just replace ternary operation to Optional syntax.
  - You don't have to merge this if you don't like it.

---


sbrannen:
Thanks for the PR; however, in general we do not wish to replace the use of the ternary operator or simple if/else blocks with `Optional` since doing so would introduce a slight performance hit while not providing much (if any) value in terms of readability or maintainability.

hotchemi:
OK.

PR #194, opened 2016-03-12
Simplify interface of the TestDescriptor.Visitor.

stefanbirkner:
## Overview

Every visitor gets the instance of the TestDescriptor and a Runnable for deleting it from the hierarchy. Everybody who has access to a TestDescriptor can therefore delete it, but with an approach that is not obvious. That's why I added the delete method to the TestDescriptor interface because I think that it is easier to understand compared to the Runnable.

---

### DIFF
--- a/junit-engine-api/src/main/java/org/junit/gen5/engine/TestDescriptor.java
+++ b/junit-engine-api/src/main/java/org/junit/gen5/engine/TestDescriptor.java
@@ -55,6 +55,8 @@ public interface TestDescriptor {
 
 	void removeChild(TestDescriptor descriptor);
 
+	void removeFromHierarchy();
+
 	default Set<? extends TestDescriptor> allDescendants() {
 		Set<TestDescriptor> all = new LinkedHashSet<>();
 		all.addAll(getChildren());
@@ -72,7 +74,7 @@ public interface TestDescriptor {
 
 	interface Visitor {
 
-		void visit(TestDescriptor descriptor, Runnable remove);
+		void visit(TestDescriptor descriptor);
 	}
 
 	void accept(Visitor visitor);
--- a/junit-engine-api/src/main/java/org/junit/gen5/engine/support/descriptor/AbstractTestDescriptor.java
+++ b/junit-engine-api/src/main/java/org/junit/gen5/engine/support/descriptor/AbstractTestDescriptor.java
@@ -68,7 +68,8 @@ public abstract class AbstractTestDescriptor implements TestDescriptor {
 		child.setParent(null);
 	}
 
-	protected void removeFromHierarchy() {
+	@Override
+	public void removeFromHierarchy() {
 		if (isRoot()) {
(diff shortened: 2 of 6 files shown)


jlink:
Thanks for the suggestion. What I don't like about it is that it adds another method to TestDescriptor that is not needed by most clients. The visitor might not be obvious but it succeeds in keeping separate responsibilities separate.

schauder:
But it really is part of the API, just in a really strange hidden way. I stumbled over it. And I know others that stumbled over it. (We did some code reviewing at JavaLand with the JUnit 5 code).

If one wants to keep it separate from the rest of the API, I guess one could introduce a new separate interface: VisitableTestDescriptor

jlink:
VisitableTestDescriptor sounds like a viable alternative.

stefanbirkner:
I agree that the `TestDescriptor` interface may have too many methods, but hiding the API in a visitor doesn't solve this problem. IMHO making the method visible is an intermediate step for a better solution.

I'm confused about the responsibility of the visitor. It is used for different things: counting tests, removing descriptors and collecting ids. What is the purpose of the visitor if it is not visiting every node of the `TestDescriptor` hierarchy?

jlink:
2016-03-13 13:15 GMT+01:00 Stefan Birkner <notifications@github.com>:

> I agree that the TestDescriptor interface may have too many methods, but
> hiding the API in a visitor doesn't solve this problem. IMHO making the
> method visible is an intermediate step for a better solution.
>

Not sure I agree. Hiding part of an API behind a different interface seems
reasonable to me. It might not be the best split up, though, but a better
one has not been suggested yet. The only other suggestion was to introduce
a VisitableTestDescriptor. I'd like to see that in action to get a feel for
it.


> I'm confused about the responsibility of the visitor. It is used for
> different things: counting tests, removing descriptors and collecting ids.
> What is the purpose of the visitor if it is not visiting every node of the
> TestDescriptor hierarchy?
>

"Visitor" is the implementation technique; counting, pruning and collecting
is the purpose for what it's being used. I still prefer it to adding yet
another method to TestDescriptor.

What do the other team members think?

mmerdes:
I agree with @stefanbirkner that the current solution tends to be confusing.
The VisitableTestDescriptor approach might be a viable solution.

marcphilipp:
How would the solution with a separate interface look like? Can someone sketch that? Would it be a sub-interface of TestDescriptor? If so, how would the default methods in TestDescriptor use it?

marcphilipp:
@stefanbirkner We want to merge this pull request. Do you have time to rebase on master?

stefanbirkner:
Yes. I rebase it tomorrow.

stefanbirkner:
The pull request is ready to be merged.

bechte:
Due to merging #218 into master, there is an error in ```DiscoverySelectorResolver```. I pulled the changes from @stefanbirkner into a new branch and fixed the issue.

This PR is now merged into master.

PR #198, opened 2016-03-12
Refactor commons module

hotchemi:
## Overview

- Make AnnotationCacheKey's constructor package local.
- Modify to use method reference.
- Modify to use normal String in the toString of ToStringBuilder.

---

### DIFF
--- a/junit-commons/src/main/java/org/junit/gen5/commons/util/AnnotationUtils.java
+++ b/junit-commons/src/main/java/org/junit/gen5/commons/util/AnnotationUtils.java
@@ -223,7 +223,7 @@ public final class AnnotationUtils {
 		private final AnnotatedElement element;
 		private final Class<? extends Annotation> annotationType;
 
-		public AnnotationCacheKey(AnnotatedElement element, Class<? extends Annotation> annotationType) {
+		AnnotationCacheKey(AnnotatedElement element, Class<? extends Annotation> annotationType) {
 			this.element = element;
 			this.annotationType = annotationType;
 		}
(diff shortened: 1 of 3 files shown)


jlink:
Both the StringBuilder and the method call improve readability IMO. It's on a level of detail where we allow differences in style and respect the original developer's design choice.

stefanbirkner:
I think for a project like JUnit it is beneficial to have a consistent code style. JUnit's code will be very often read by people who are not JUnit developers. We should make their lives easy.

I think that @hotchemi's change improve readability (but readability depends on the style of the code you're used to read.)

jlink:
I agree that a consistent style is a plus. We therefore already have a style guide and will accept PRs that make code follow it. 

The question is: To what level of depth do we want to go: Always using method refs regardless of the context? Sometimes one is better readable than the other and sometimes not. Sometimes there are compelling reasons for one style over the other and sometimes there are not. 

I'd like to give implementors more leeway to still have a style of their own. Consistency also comes with a price. Focusing too much on detail - especially in a phase like now where quite a lot of fundamentals are unclear - is a real distraction.

hotchemi:
Sorry for the late.
When it comes to the consistency, in my opinion you should not accept various coding style.
I have no confidence how reader feel about that there is some method reference and not, but think about programmer, it helps us to remove their burden that thinking about which is good for "readability".
It's the golang philosophy(a.k.a go fmt). But of course there is another thinking like Perl or Ruby way.
It depends for junit team guys but mixing is always the cause of confusion right?

hotchemi:
And actually Intellij always show these point...

jlink:
There is no general hard line between _style_ and _implementation_, so let's get concrete:
- Should we have "always use method refs where possible" in our style guide?
- Should we have "always use string concatenation instead of string builder" in our style guide?
- Should we have "always use streams instead of iteration" in our style guide?
My personal opinion is "none of these". 

What do the other junit5 team members (@marcphilipp @sbrannen @mmerdes @bechte) think? We need some decision to resolve this PR and future ones along the same line...

marcphilipp:
I agree in all three points with @jlink.

hotchemi:
OK, should I close this PR?

jlink:
Yes. For me it was an important pull request because it helped to get my thoughts about style and cleaning up clearer. Thanks!

PR #199, opened 2016-03-12
Replace displayHelp call with method reference.

hotchemi:
## Overview

- Replace displayHelp call with method reference.

---


jlink:
Thanks for all the detailed code clan up. However, using method calls is often more readable than method references and/or a matter of style. That's why we allow differen styles in that respect and leave it as the original author coded it.

Issue #204, opened 2016-03-15
Allow Gradle plugin to support Android projects


mannodermaus:
At the moment, the `junit-gradle` plugin doesn't work on Android environments, since the handling of `SourceSet` containers and the general notion of the `runtime classpath` differ from plain Java, both syntactically and functionally. For instance, there is no [testRuntime configuration](https://github.com/junit-team/junit5/blob/master/junit-gradle/src/main/groovy/org/junit/gen5/gradle/JUnit5Plugin.groovy#L28) in Android projects, and the [testClasses task](https://github.com/junit-team/junit5/blob/master/junit-gradle/src/main/groovy/org/junit/gen5/gradle/JUnit5Plugin.groovy#L66) is missing as well.

I have [started working on Android support in my fork](https://github.com/aurae/junit5/blob/android/junit-gradle/src/main/groovy/org/junit/gen5/gradle/JUnit5Plugin.groovy), however that isn't done yet. Maybe one of you guys has already had Android support on their radar all along?

Thanks!

jlink:
At the time JUnit5's current design was conceived (Oct 2015) it looked like
Android wouldn't be on Java 8 for another couple of years. So we considered
Android support to be out of reach for some time and didn't bother.

Now that prospects have changed, it might be time to think about Android
again...

2016-03-15 11:21 GMT+01:00 Marcel Schnelle <notifications@github.com>:

aurae:
Thanks for your reply! The greatest thing about JUnit tests for Android developers has always been the fact that they are decoupled from the limitations that Android imposes on developers (mainly related to being bound to Java 6, and in parts, Java 7). This allows us to enforce TDD by not having to wait around for instrumentation tests to deploy to devices etc. That being said, Android getting Java 8 with the N release shouldn't even necessarily be the main reason that JUnit 5 should come to the platform, since a lot of devs have been using Java 8 as their internal foundation for quite some time now!

jlink:
Maybe you can add a few details to the points you mention in the issue description, so that we can estimate the effort and schedule it in one of the next releases. I'm quite sure, though, that it won't be part of https://github.com/junit-team/junit5/milestones/5.0%20M1 :-/

aurae:
I have made some good progress already, so that all relevant JUnit5 tasks are being initialized properly, and the most basic of tests are already picked up correctly. However, I am somewhat running into a wall when trying to set up the Android equivalent of the runtime classpath.

[...]

PR #213, opened 2016-03-22
Add Android support to the Gradle plugin

aurae:
## Overview

See #204 for a discussion on the topic.

- Extended `JUnit5Plugin` to detect applied Android plugins and tailor task creation to its environment
- Added tests for Android environment (**Note:** This requires an Android SDK installation on the target machine)

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -307,7 +307,7 @@ configure(rootProject) {
 	}
 
 	task wrapper(type: Wrapper) {
-		distributionUrl = 'https://services.gradle.org/distributions/gradle-2.11-all.zip'
+		distributionUrl = 'https://services.gradle.org/distributions/gradle-2.12-all.zip'
 	}
 
 }
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
-#Sun Mar 13 16:08:09 CET 2016
+#Mon Mar 14 16:18:48 CET 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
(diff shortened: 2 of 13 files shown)


marcphilipp:
@aurae Thanks for submitting the pull request! Have you noticed that the build on Travis failed?

aurae:
Yes, I have! After checking the `.travis.yaml` config file used in this project, it failing on the Android-related tests is most definitely related to the VM not having access to an installation of the Android SDK. I have verified these errors on a machine without the proper setup (that is, a missing `ANDROID_HOME` environment variable, as well as the SDK located in the folder it's pointing to).

Usually, Travis configuration for Android projects is set up by specifying `language: android`, along with some additional properties. [Here's a minimal example of what that looks like](https://github.com/journeyapps/android-maven-example/blob/master/.travis.yml). I'm hesitant to change JUnit's `.travis.yaml` to that on a global level, though, just because the Gradle plugin happens to be compatible with Android. Then again, the alternative would be [to `wget` the SDK and configure everything manually](http://gmariotti.blogspot.de/2014/04/integrating-android-github-repo-with.html), which would pollute Travis' configuration as well.

I will perform some tests regarding the change to `language: android` for this project's modules - in theory, the image should be identical to the Java one, but extended with the components which are required for the new tests.

jlink:
You could make the android tests conditional so that they are only executed if an android sdk is available. 

Von meinem iPad gesendet

aurae:
Great hint! I have added conditional execution to the new test cases now, and made them dependent on the presence of the required environment variable.

aurae:
I'm closing this in favor of #499, which honors the current codebase better than this older version.

Issue #58, opened 2015-12-11
Add mechanism for registering dynamic tests


marcphilipp:
## Further Resources

- Related to #14.

## Subtasks

- [ ] #278 
- [ ] #275 
- [ ] #280

jlink:
@mmerdes and I are starting on some experiments with it.

sbrannen:
Awesome!

Just skimmed the work in the branch, and... it's great that you guys are making progress here.

Looking forward to the result...

Good Luck!

jlink:
I don't think we'll go much further with branch (
https://github.com/junit-team/junit5/tree/issue58-dynamic-tests) unless we
agree that's the direction we want to go.

Here's a summary of what we learned so far:

- Adding a lambda-based approach for registering and executing dynamic
tests was rather straight-forward. We introduced a new annotation
`@Dynamic` for methods that return `Stream<DynamicTest>`,
`Collection<DynamicTest>` or `Iterator<DynamicTest>`. Those tests are
created, registered and executed on the fly and annotated methods follow
the same behaviour as test methods as for `Before/AfterEach` and parameter
injection.

- The lambda-based however do currently not have a life cycle / extension
model of their own. Just reusing the standard model from test methods did
not seem intuitive nor possible for various reasons (e.g. early binding of
the test instance to a lambda at creation time). But maybe we don't need
that.

- Re-running such a lambda-based test by using its unique ID has not been
implemented (yet). We started doing it but realized that the resulting code
(as we imagined it) would be rather involved. We think, that's due to the
current engine's focus on a statically created test hierarchy which does
not lend itself well to dynamically created tests. However, it would be
possible to add this functionality.

- The biggest drawback in the branch's implementation is that we don't
think parameterized tests through parameter injection (see #l14) can be
done with the same approach and the same life cycle/extension model as
@Test annotated methods. Same goes for quick-check-like tests of the same
kind. But maybe we're wrong.

Our suggestion is that you guys look at the programming model and the
implementation and tell us what alternatives you see.

2016-03-04 15:26 GMT+01:00 Sam Brannen <notifications@github.com>:

schauder:
Had a quick look. I like it for one big (at least for me) reason: It seems to easily allow writing Tests as closures. With a simple class using these dynamic tests, it should be able write tests like:

    class MyTests extends ClosureBasedTestRegistry {{
        test("my test name"){() -> {
            // actual test goes here
        });
    }}

ttddyy:
[...]

PR #218, opened 2016-03-31
New discovery mechanism

jlink:
When working on dynmaic tests #58 (https://github.com/junit-team/junit5/tree/issue58-dynamic-tests) and jqwik (https://github.com/jlink/jqwik/) I learned that introducing methods with different annotation (e.g. @Dynamic and @Property) and different semantics can be more easily done by adding new test descriptors than by implementing a new test engine. In this regard this is similar to #134 but implemented with (I hope) less interdependencies.

## Overview
- New discovery mechanism that lends itself more easily to extension
- Alternative suggestion to #134 from @bechte 

## Todos
- Add more microtests for resolvers and `MethodFinder`
- Open registration of `ElementResolver` as extension point

---


smoyer64:
I'm a heavy user of JUnitParams and love the idea that you can register multiple resolvers (and that you aren't limited to one as JUnitParams is via its reliance on a runner).  Here's a use case that I'd like to propose as a strawman.  How would I create and register additional (Method?) resolvers to run the following test 12 times (the cross-product of StringValues and NumericValues):

```
  @NumberValues({1, 2, 3})
  @StringValues({"A", "B", "C", "D"})
  void testParameterExpansion(int numberValue, String stringValue) {
    ...
  }
```

Using the current SNAPSHOT build I can obviously create parameter resolvers that read the annotations, but since I'm only passed one instance (executing from one TestDescriptor) I can only run a single iteration of the test.

One other quick idea (for parameterized tests) is to allow AfterDiscovery extensions to be registered.  The AfterDiscovery interface would look like:

```
public interface AfterDiscovery {
  TestDescriptor alter(TestDescriptor descriptor);
}
```

If a single method's TestDescriptor is passed into this extension, it could use the context to instead create a container with the 12 children defined above.

Thanks for thinking parameterized tests through so thoroughly ... I'm going to be a big fan!

bechte:
I updated / rebased branch jl-issue106-unique-id onto master.

PR #220, opened 2016-04-02
StringUtils: clean up nullSafeToString

mureinik:
## Overview

A minor cleanup in `StringUtils#nullSafeToString`: Streaming and joining an empty class list will return an empty string (`""`), so explicitly checking for it redundant.

---


sbrannen:
Thanks for submitting the PR; however, the current code is the way it is for a good reason: there is no need to use streams and joining if we know upfront that it's pointless.

In other words, this is intentionally like this for performance: it's a shortcut, which is behavior that we desire in core utility code.

I am therefore closing this issue.

Issue #142, opened 2016-02-01
Add Microtests for MethodInvoker


mmerdes:

mmerdes:
@signed: are you still working on this?

signed:
I intend to, just waiting for feedback on #223 to see if I can keep using [`ExtensionRegistry`](https://github.com/junit-team/junit5/blob/master/junit5-engine/src/main/java/org/junit/gen5/engine/junit5/extension/ExtensionRegistry.java) for the tests (preferred) or if I better move to a test double.
[Here](https://github.com/signed/junit5/blob/sandbox/junit-tests/src/test/java/org/junit/gen5/engine/junit5/execution/MethodInvokerTests.java) is what I have so far. I'll squash the commits before the pull request.

marcphilipp:
#223 has been merged.

sbrannen:
Merged PR #248 into `master` in 3ebaf5fc328931679943bb55a4007a64d5cf7b85.

PR #223, opened 2016-04-11
replace constructor with intention revealing factory method

signed:
While working on #142 I was surprised by the fact that a `new ExtensionRegistry()` already contains `MethodParameterResolver`s. I was expecting an empty `ExtensionRegistry`.

I replaced the default constructor with the factory method `ExtensionRegistry.newRootRegistryWithDefaultExtensions()` to better inform about the default extensions being added. With the single constructor remaining it is now possible to create an empty `ExtensionRegistry` what makes the tests in #142 easier to write.

## Overview
* replaced default constructor of `ExtensionRegistry` with factory method `newRootRegistryWithDefaultExtensions()`
* enable instantiation of empty `ExtensionRegistry`

---

### DIFF
--- a/junit5-engine/src/main/java/org/junit/gen5/engine/junit5/extension/ExtensionRegistry.java
+++ b/junit5-engine/src/main/java/org/junit/gen5/engine/junit5/extension/ExtensionRegistry.java
@@ -62,10 +62,24 @@ public class ExtensionRegistry {
 	 */
 	public static ExtensionRegistry newRegistryFrom(ExtensionRegistry parentRegistry,
 			List<Class<? extends Extension>> extensionTypes) {
-
-		ExtensionRegistry newExtensionRegistry = new ExtensionRegistry(parentRegistry);
+		// @formatter:off
+		ExtensionRegistry newExtensionRegistry = Optional.ofNullable(parentRegistry)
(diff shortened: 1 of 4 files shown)


junit5-engine/src/main/java/org/junit/gen5/engine/junit5/extension/ExtensionRegistry.java
```diff
@@ -62,12 +62,26 @@
 	 */
 	public static ExtensionRegistry newRegistryFrom(ExtensionRegistry parentRegistry,
 			List<Class<? extends Extension>> extensionTypes) {
-
-		ExtensionRegistry newExtensionRegistry = new ExtensionRegistry(parentRegistry);
+		// @formatter:off
+		ExtensionRegistry newExtensionRegistry = Optional.ofNullable(parentRegistry)
+				.map(parent -> new ExtensionRegistry(Optional.of(parent)))
+				.orElseGet(ExtensionRegistry::newRootRegistryWithDefaultExtensions);
+		// @formatter:on
```
marcphilipp: I don't think we need to support `parentRegistry` being `null` here.

marcphilipp:
@signed I've cherry-picked your commit and cleaned up the tests. Thanks!

Issue #222, opened 2016-04-09
Lists of talks about JUnit 5


nicolaiparlog:
I am currently writing an article about JUnit 5 and would like to raise awareness by pointing towards the talks that people start giving about JUnit 5. It would be great if there were a resource (e.g. a wiki page) that I could link to.

Would the project be interested in creating one?

(If only speakers could then enter themselves by opening pull requests.)

marcphilipp:
That's a very good idea! As long as JUnit 5 does not have a website I'd suggest we could use the wiki.

@junit-team/junit-lambda Any objections?

sbrannen:
Nope. No objections.

Sounds like a good idea. Go for it!

Here are three talks I'll be giving that can be added to the list (in case you make the page):

- Skills Matter in London, England: https://twitter.com/sam_brannen/status/718417999158517760
- JUG in Zurich, Switzerland: https://twitter.com/sam_brannen/status/718418900153667584
- Spring I/O in Barcelona, Spain: https://twitter.com/sam_brannen/status/718421554422484992

nicolaiparlog:
Here's one I'll be giving:

* May 21st, JEEConf 2016 in Kyiv, Ukraine [[Link](http://jeeconf.com/program/junit-5-next-generation-testing-on-the-jvm/)]

nicolaiparlog:
Because I don't know where else to put it, I'll do it here: I'll be talking about JUnit 5 at Devoxx Poland: http://cfp.devoxx.pl/2016/talk/OOT-4758

Unrelated: If there were a repo/branch with the Jekyll sources (or did I miss it?), we could add these via PRs.

marcphilipp:
Thanks for letting us know. I've added your talk to the list in 88787cec05ce2a2244be05a67a7b6a366a79ecaf.

The Jekyll sources are in the `gh-pages` branch in this repository. To add a talk you'll have to edit [events.json](https://github.com/junit-team/junit5/blob/gh-pages/data/events.json) and (if adding a new speaker) [speakers.json](https://github.com/junit-team/junit5/blob/gh-pages/data/speakers.json).

nicolaiparlog:
Oops, didn't look for the sources _in_ `gh-pages`. If there are more talks, I will make a PR.

Issue #205, opened 2016-03-16
New junit.org Website


jlink:
Create new website:
- More modern site generation, e.g. Jekyll
- More modern, responsive layout
- Split website in areas for JUnit4 and JUnit5

marcphilipp:
I've started to work on this in my fork:
https://github.com/marcphilipp/junit5/tree/gh-pages

marcphilipp:
TODOs:
- [ ] better icon for Javadoc
- [ ] id for events heading
- [ ] iCal feed?

marcphilipp:
Even though we will probably continue to work on the website, I'm closing this issue for now as we have an initial version.

PR #227, opened 2016-04-17
Jekyll-based website

marcphilipp:
## Overview

This adds a Jekyll-based website that only consists of a single page so far. It contains links to User Guide, Javadoc, and GitHub. In addition, it includes a "Upcoming Events" section that has been prefilled using data from #222.

Related issue: #205.

---


nicolaiparlog:
I like the upcoming events. :+1:

If they could get an ID, the link could be more specific, which might help if they move someplace else on the page.

marcphilipp:
You mean the "Upcoming Events" heading should get an id?

nicolaiparlog:
Yes, exactly. Sorry for being vague, it's early. ;)

marcphilipp:
That's perfectly fine. I've added a TODO on #205 to add an ID to the Upcoming Events heading. Thanks for the suggestion!

PR #230, opened 2016-04-18
Remove redundant toString() calls

mureinik:
## Overview

String.toString() just returns the string itself, so calling toString()
on a String instance is pointless.

This patch removes the occurrences of this faulty pattern from the
codebase, and replaces them with a straight-forward use of the strings
themselves.

---

Signed-off-by: Allon Mureinik <mureinik@gmail.com>


sbrannen:
Thanks for submitting the PR!

PR #231, opened 2016-04-19
Fix indentation for failed tests in ConsoleRunner

### LINKED ISSUE #160: ConsoleRunner uses wrong indentation for failed grouped test
If a test like this fails
```java
@Test
public void testNames() throws Exception {
  Person person = new Person();
  
  assertAll( "names", () -> {
    assertEquals( "John", person.getFirstName() );
    assertEquals( "Doe", person.getLastName() );
  } );
}
```

... the ConsoleRunner uses the wrong indentation like shown below. It seems that the `expected` line is indented with a single tab whereas the line above uses spaces.
```
Finished:    testNames [junit5:com.codeaffine.junit5.GroupAssertionsTest#testNames()]
             => Exception: names (1 failure)
	expected: <John> but was: <something else>
```

lutovich:
## Overview

Indentation for failed tests was a bit off because printed exception message could be a multi line string.

This PR fixes indentation by changing exception messages to contain same amount of spaces as the rest of the output.

Fixes #160 

---


marcphilipp:
Thanks! :)

Issue #232, opened 2016-04-29
Remove ExtensionPoint, ExtensionRegistrar, Position etc. and simplify implementation of JUnit 5 TestEngine


marcphilipp:

sbrannen:
_in progress_

sbrannen:
The _work in progress_ can be viewed here: https://github.com/sbrannen/junit5/commits/issue232-simplified-extension-model

And an official PR is forthcoming.

sbrannen:
See PR #237

sbrannen:
Merged into `master` in 434a3609f4b0a80fdeeac4a35f7c4b03bd2138d5.

PR #237, opened 2016-05-03
Revert to a simplified extension model

sbrannen:
In order to postpone support for programmatic extension registration, the JUnit Team has decided to revert to a _simplified extension model_ with a focus on _extensions that are applied at extension points_.

The commits in this pull request therefore address the following issues.

- Removal of `ExtensionPoint`, `ExtensionPointRegistry`, and `ExtensionRegistrar`
- Removal of the suffix `ExtensionPoint` in extension interface names
- Introduction of _before_ and _after_ test method execution lifecycle callbacks
- Simplification of the implementation of JUnit 5 test descriptors
- Separation of concerns regarding user code and extension code
- Eager lookup and validation of user-defined lifecycle callback methods during the discovery phase

Issue: #232

---


mmerdes:
I did a short analysis of potential conflicts between this PR and the changes in branch issue58-dynamic-tests-with-new-resolvers. There are only 2 files changed by both branches. The difference in MethodInvoker is trivial, the conflicts in MethodTestDescriptor are more substantial but should probably  be resolvable...

mmerdes:
BTW
I like the unformatted 'indentation art' in MethodTestDescriptor.execute()
While not 'spotless' it sure helps expressing a meaning which would not be achievable with merely syntactic means.
:)

sbrannen:
Thanks for checking for potential conflicts.

I hope the merging with `MethodTestDescriptor` isn't too time consuming for you!

sbrannen:
> I like the unformatted 'indentation art' in MethodTestDescriptor.execute() :)

"Indentation art" --> you've coined a new term!

We've already been using that technique in our tests when we want to verify extension execution order, so it only seemed fitting to do the same in the production code for exactly the same reason: visualization of the wrapping nature of extensions and lifecycle callbacks. So I'm glad you like it. :wink:

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/LifecycleMethodUtils.java
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.gen5.commons.JUnitException;
+import org.junit.gen5.commons.util.ReflectionUtils;
+import org.junit.gen5.commons.util.ReflectionUtils.MethodSortOrder;
+
+/**
+ * Collection of utilities for working with test lifecycle methods.
+ *
+ * @since 5.0
+ */
+final class LifecycleMethodUtils {
```
bechte: Nice. I like the idea of having a central place of how to lookup and find the different pieces. We also do have some more of these things in extra classes. I am wondering if we could consolidate those all into a "function" library. This will not be part of this PR, just an idea I wanted to mention.

bechte:
Great job! Thanks. I support the changes in this PR and vote for merging it into master. Let's talk about it in our call.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/ClassTestDescriptor.java
```diff
@@ -78,6 +76,8 @@ public ClassTestDescriptor(UniqueId uniqueId, Class<?> testClass) {
 
 		this.beforeAllMethods = findBeforeAllMethods(testClass);
 		this.afterAllMethods = findAfterAllMethods(testClass);
+		this.beforeEachMethods = findBeforeEachMethods(testClass);
+		this.afterEachMethods = findAfterEachMethods(testClass);
```
marcphilipp: Where are the exceptions thrown by these methods caught?

documentation/src/docs/asciidoc/extensions.adoc
```diff
@@ -178,25 +99,27 @@ resolved by _name_, _type_, _annotation_, or any combination thereof. For concre
...
+`{extension-api-package}` package for further details.
 
-* `BeforeEachExtensionPoint`
-* `AfterEachExtensionPoint`
-* `BeforeAllExtensionPoint`
-* `AfterAllExtensionPoint`
+* `BeforeEachCallback`
+* `AfterEachCallback`
+* `BeforeAllCallback`
+* `AfterAllCallback`
```
marcphilipp: The two new extension points (`BeforeTestMethodCallback` and `AfterTestMethodCallback`) are missing here, right?

marcphilipp:
I think we need a follow-up issue to take care of exception handling in the JUnit 5 TestEngine. E.g. invalid before-each methods should not lead to a complete abortion of the discovery process.

marcphilipp:
Team Decisions:
- Exception handling will be addressed in #242
- @sbrannen will update the user guide (wrt. the two new extension points) and merge this pull requests
- @sbrannen will create a follow-up issue regarding naming of `Extension` interfaces

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/ClassTestDescriptor.java
sbrannen: To be addressed in #242. :wink:

Issue #238, opened 2016-05-05
Add project metadata to JAR manifest


aalmiray:
It would be great if JAR manifests contained additional information related to the project such as

    Manifest-Version: 1.0
    Built-By: ${user.name}
    Created-By: 1.8.0_92 (Oracle Corporation 25.92-b14)
    Build-Date: 2016-05-05
    Build-Time: 18:44:48.377+0200
    Build-Revision: 0ca4beaa67bd484f98fd46e776e31c89adfea234
    Specification-Title: ${project.name}
    Specification-Version: 5.0.0-ALPHA
    Specification-Vendor: junit.org

sbrannen:
Merged PR into `master` in febd4b10d63358a1a1b9bec8cfede02c0abc4653.

PR #239, opened 2016-05-05
Configure additional metadata on JAR manifest.

aalmiray:
Fix for #238 by adding project metadata to JAR manifest.
Relies on the `versioning` plugin to pull in the latest SCM hash.

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,7 @@ Date buildTimeAndDate = new Date()
 ext {
 	buildDate = new SimpleDateFormat('yyyy-MM-dd').format(buildTimeAndDate)
 	buildTime = new SimpleDateFormat('HH:mm:ss.SSSZ').format(buildTimeAndDate)
+	builtByValue = project.hasProperty('builtBy') ? project.builtBy : project.defaultBuiltBy
 }
 
 allprojects {
@@ -230,18 +231,29 @@ subprojects {
 			}
 		}
 	}
+	
+	def normalizeVersion = { versionLiteral ->
+		try {
+			(versionLiteral =~ /(\d+)\.(\d+)\.(\d+).*/)[0][1..3].join('.')
+		} catch(x) {
+			throw new GradleException("Version '$versionLiteral' does not match version pattern, e.g. 5.0.0-QUALIFIER")
+		}
+	}
 
 	jar {
 		manifest {
 			attributes(
-				'Built-By': System.properties['user.name'],
+				'Built-By': builtByValue,
 				'Created-By': "${System.properties['java.version']} (${System.properties['java.vendor']} ${System.properties['java.vm.version']})".toString(),
 				'Build-Date': buildDate,
 				'Build-Time': buildTime,
 				'Build-Revision': versioning.info.commit,
 				'Specification-Title': project.name,
-				'Specification-Version': project.version,
+				'Specification-Version': normalizeVersion(project.version),
 				'Specification-Vendor': 'junit.org',
+				'Implementation-Title': project.name,
+				'Implementation-Version': project.version,
+				'Implementation-Vendor': 'junit.org'
 			)
 		}
 	}
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import java.text.SimpleDateFormat
+
 buildscript {
 	repositories {
 		// mavenLocal()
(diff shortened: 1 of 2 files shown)


sbrannen:
I'm not so sure we should include the user name in the build artifacts, but otherwise I think it's a great idea to have the additional metadata in the manifest.

So unless you have a really compelling reason to include the user name, I think we can merge this without the user name.

sbrannen:
Regarding the `net.nemerosa:versioning` plugin, how reliable is that?

sbrannen:
By the way, it's cool to see that you're playing around with JUnit 5! :+1: 

Have you tried it out in anger yet?

aalmiray:
Ideally, the published artifacts would come from a CI environment, in which the username may be good to have, as it can attest to the artifact's origin.

There are quite a good number of plugins out there that deal with SCM metadata, I've found `net.nemerosa:versioning`to be the most complete and stable of the bunch.

aalmiray:
I was looking at the `junit5-gradle` plugin; noticed that there's no close integration with Gradle's test reporting yet. I thought I'd have a look at that too.

sbrannen:
> Ideally, the published artifacts would come from a CI environment, in which the username may be good to have, as it can attest to the artifact's origin.

OK. sounds reasonable.

sbrannen:
> There are quite a good number of plugins out there that deal with SCM metadata, I've found net.nemerosa:versioningto be the most complete and stable of the bunch.

Fair enough. I don't have a problem with giving it a try, but let's see what the rest of the team thinks.

Cheers

sbrannen:
@aalmiray, that would be totally awesome if you spruced up the `junit5-gradle` plugin!

Go for it!

sbrannen:
@junit-team/junit-lambda, what do the rest of you think about adding extra metadata to the JARs?

marcphilipp:
I'm ok with adding additional metadata to the manifest.

Regarding the Gradle plugin: We want the Gradle team to take over after M1. It's not clear whether they will have time to do it themselves, so they might appreciate a pull requests for native JUnit 5 support in Gradle. I think that would be the way forward.

mmerdes:
adding the standard meta-data to the manifest is good practice, so I would welcome it.
IMHO, it should not contain a personal user name.
can the build be set up to use the JUnit Team account or similar?

aalmiray:
Of course, the value of the `Built-by` entry can be taken from anywhere. I picked `user.name` as it could give additional hints on the artifact's origin. But it may very well be changed to a fix value obtained from a project property (set in a gradle build file, gradle.properties or supplied as a command line option).

marcphilipp:
I think reading `Built-by` from a project property that is initialized to the `user.name` system property sounds good.

sbrannen:
> I think reading `Built-by` from a project property that is initialized to the `user.name` system property sounds good.

👍 

And I'd recommend defaulting to "JUnit Team" or "unknown" in case the project property hasn't been set.

build.gradle
```diff
@@ -221,6 +231,20 @@ subprojects {
...
+	jar {
+		manifest {
+			attributes(
+				'Built-By': System.properties['user.name'],
+				'Created-By': "${System.properties['java.version']} (${System.properties['java.vendor']} ${System.properties['java.vm.version']})".toString(),
+				'Build-Date': buildDate,
+				'Build-Time': buildTime,
+				'Build-Revision': versioning.info.commit,
+				'Specification-Title': project.name,
+				'Specification-Version': project.version,
```
sbrannen: According to the JLS, `project.version` is not appropriate for the `Specification-Version` since `project.version` contains non-digits (e.g., `5.0.0-ALPHA`, `5.0.0-M1`, `5.0.0-SNAPSHOT`, etc.). 

We should therefore use `project.version` for the `Implementation-Version` and a parsed version (with the "-modifier" dropped off) for the `Specification-Version`.

We should also set the `Implementation-Title` and `Implementation-Vendor` for completeness.

aalmiray: AFAIK Gradle has a built-in version parsing utility that can be used to correctly detect the version number.

sbrannen: OK. If it works auto-magically, then... cool.

sbrannen:
This has been merged into `master` in febd4b10d63358a1a1b9bec8cfede02c0abc4653.

Thanks again for all of your Gradle help!

Issue #252, opened 2016-05-10
Apply InstancePostProcessor extensions for every test instance


sbrannen:
## Status Quo

`InstancePostProcessor` extensions currently only get applied during the invocation of `MethodTestDescriptor.execute()`; however the actual _instance_ is created in `MethodTestDescriptor.prepare()`. Furthermore, outer instances for `@Nested` test classes are created transparently in `NestedClassTestDescriptor.testInstanceProvider()`.

## Analysis

For _classic_ tests, this does not pose a problem; however, for `@Nested` tests it means that `InstancePostProcessor` extensions are not applied to instances of outer classes when test methods in a nested class are executed.

See _TODOs_ in `InstancePostProcessorTests`.

## Deliverables

- [x] Ensure that `InstancePostProcessor` extensions are applied immediately after instance creation for _all_ test instances (including outer instances), not just when test methods are executed.

sbrannen:
_in progress_

sbrannen:
pull request submitted: #253

sbrannen:
Merged PR #253 into `master` in 2f8496577b8e19156884cd752851736ca16341e1.

PR #253, opened 2016-05-11
Apply InstancePostProcessor extensions for every test instance

sbrannen:
Prior to this commit, InstancePostProcessor extensions were only applied
during the execution of a MethodTestDescriptor instead of at the point
when such an instance was physically instantiated. Consequently,
InstancePostProcessor extensions were never applied to instances of
outer classes when test methods in a nested class were executed, which
lead to improperly initialized outer instances (e.g,
NullPointerExceptions, etc.).

This commit fixes this bug by relocating the invocation of
InstancePostProcessor extensions to places in the framework where test
instances are physically instantiated (i.e., in ClassTestDescriptor and
NestedClassTestDescriptor).

Furthermore, the argument list for the postProcessTestInstance() method
in the InstancePostProcessor API has been modified to accept the test
instance and a generic ExtensionContext.

Issue: #252

---


sbrannen:
@junit-team/junit-lambda, please review when you get a chance.

Thanks,

Sam

marcphilipp:
Looks good to me, thanks!

sbrannen:
Merged into `master` in 2f8496577b8e19156884cd752851736ca16341e1.

PR #257, opened 2016-05-12
name variable properly

signed:
## proper name for lambda parameter
The parameter is a descriptor, not an engine.

---


sbrannen:

Thanks

PR #259, opened 2016-05-13
Configure shadow plugin to package junit5-console as a fat JAR

### LINKED ISSUE #258: Package junit-platform-console as a fat JAR
https://github.com/johnrengelman/shadow plugin can be used to achieve this goal. This plugjn is capable of merging metadata files (such as META-INF/services) and others. It's also friendly with the `application` plugin; it can produce an executable fat jar when both plugins are applied.

aalmiray:
The `shadowJar` task takes care of merging service files.
This task will generate an artifact with "all" as classifier;
this artifact is not yet configured for publication to a maven
compatible repository.
## Fixes #258


sbrannen:
Assigned to M2 for now.

smoyer64:
There's an interesting article regarding fat JARs from HubSpot - http://product.hubspot.com/blog/the-fault-in-our-jars-why-we-stopped-building-fat-jars.

I've traditionally used OneJar to get around the problem of files clobbering each other (e.g. META-INF/MANIFEST.MF), but I'm not convinced that the new SlimFast plugin is the right solution either.  Where are you Java 9?

sbrannen:
Depends on #309.

Issue #233, opened 2016-04-29
Introduce mechanism to pass parameters to Launcher which passes it to the TestEngines


marcphilipp:
This is a prerequisite for #78.

bechte:
I assigned this issue to myself and start working on it. I will cover the following topics:

- Provide an handy way to specify additional launch parameters from the ```TestDiscoveryRequestBuilder```
- Add a way to fetch the launch parameters from the ```EngineDiscoveryRequest```

sbrannen:
Closing this issue since it was addressed in conjunction with PR #261 which was merged into `master` in 3a137040dde7e1bf7a6871d69f05ace9d99cda7f.

Issue #78, opened 2015-12-21
Introduce deactivation mechanism for conditional test execution


sbrannen:
## Status Quo

There is currently no way to deactivate or turn off support for conditional test execution.

For example, if a test class or test method is annotated with `@Disabled`, it is always disabled, and there is currently no way to override this behavior. However, there are times when a developer would like to run all _disabled_ tests to see if they are still _broken_ without having to manually delete or comment out all declarations of `@Disabled`.

Similarly, a developer may wish to disable _all_ or _specific_ conditions for a given test run.

## Prerequisites

- #233

## Deliverables

- [x] Introduce a mechanism for turning off conditions -- for example, via a Java system property and/or a _configuration property_ which is supplied to a `Launcher` and propagated to engines via the `ExecutionRequest`.

jlink:
I assumed overriding @Disabled would happen through a launcher parameter.

2015-12-21 17:05 GMT+01:00 Sam Brannen <notifications@github.com>:

sbrannen:
@jlink, yes, that's likely the most common mechanism for this feature.

When I wrote "an attribute supplied to ... the `ExtensionContext`", I was envisioning that the attribute would be provided to the `Launcher` as a launch parameter and passed from the `Launcher` to the engine and on down to the `ExtensionContext`.

And as an alternative, I'm proposing that we also provide the same support via a system property for command-line usage and in builds where it might not (yet) be possible to supply custom launch parameters.

Does that make sense?

marcphilipp:
Makes sense to me.

jlink:
> Am 21.12.2015 um 17:28 schrieb Sam Brannen <notifications@github.com>:
> 
> @jlink, yes, that's likely the most common mechanism for this feature.
> 
> When I wrote "an attribute supplied to ... the ExtensionContext", I was envisioning that the attribute would be provided to the Launcher as a launch parameter and passed from the Launcher to the engine and on down to the ExtensionContext.
> 
Would the evaluator itself decide to ignore the decision or would the condition extension point not be executed? In other words: Only override @Disabled or all conditional execution?

bechte:
Nice feature. I totally agree with the parameter solution.

I could also think of running tests by their unique id should reactivate them, too. At least it would be a nice-to-have. This only applies for the components that were given by an unique id directly and not for any of their children.

But it would raise complexity (we need to track how a test was discovered) and I am not sure if we want to go that far...

schauder:
How about using tagging for this?

[...]

PR #261, opened 2016-05-13
Add support for supplying configuration parameters to test engines

bechte:
## Overview

- Added support for adding a launch parameters, see Issue: #233

---


junit-launcher/src/main/java/org/junit/gen5/launcher/main/DefaultLaunchParameter.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.junit.gen5.commons.meta.API;
+import org.junit.gen5.commons.util.ToStringBuilder;
+import org.junit.gen5.launcher.LaunchParameter;
+
+/**
+ * @since 5.0
+ */
+@API(Experimental)
+public class DefaultLaunchParameter implements LaunchParameter {
```
sbrannen: Does this need to `public`?

junit-launcher/src/main/java/org/junit/gen5/launcher/main/DefaultLaunchParameter.java
```diff
@@ -0,0 +1,74 @@
...
+	private static final long serialVersionUID = 1L;
+
+	public static LaunchParameter keyValuePair(String key, String value) {
```
sbrannen: Does this factory method need to reside in the internal implementation class?

junit-launcher/src/main/java/org/junit/gen5/launcher/main/TestDiscoveryRequestBuilder.java
```diff
@@ -51,6 +54,12 @@
  *     .filter(byEngineIds("junit5"))
  *     .filter(byNamePattern("org\.junit\.gen5\.tests.*"), byNamePattern(".*Test[s]?"))
  *     .filter(requireTags("fast"), excludeTags("flow"))
+ *     .launchParameter("key1", "value1")
+ *     .launchParameters(
+ *       keyValuePair("key2", "value2"),
```
sbrannen: I prefer `parameter()` or `param()` over `keyValuePair()`.

Thoughts?

marcphilipp:
I'm a bit confused here. Why do we pass parameters in the discovery request when we need them in the execution request?

sbrannen:
`TestDiscoveryRequest` is the one and only parameter passed to both `discover()` and `execute()` in the `Launcher` API.

So in that sense, it's a bit of a misnomer.

Thanks for pointing out the inconsistency.

junit-launcher/src/main/java/org/junit/gen5/launcher/main/DefaultLaunchParameter.java
marcphilipp: Actually, I think we can get rid of the interface and just have this class. Moreover, since `LaunchParameter` is only used in the builder we can just use strings?

sbrannen:
To avoid ambiguity, we should likely rename `TestDiscoveryRequest` or introduce something else that can be passed to `Launcher.execute(TestDiscoveryRequest)`.

sbrannen:
I guess the crucial question is:

Do we want "launch parameters" to be available during discovery, just during execution, or during both?

junit-launcher/src/main/java/org/junit/gen5/launcher/main/DefaultLaunchParameter.java
sbrannen: If `LaunchParameter` is truly only used internally within the builder, then you're right: we can get rid of it altogether.

sbrannen: Currently there is a `launchParameters(LaunchParameter... launchParameters)` method in the builder, but that's likely unnecessary.

marcphilipp:
I think it might make sense to have parameters for discovery, too. Our current use case only requires them for the execution, though.

marcphilipp:
I think we can leave it as in this pull request but need to think about how to change `TestEngine` to pass the parameters. `ExecutionRequest` already has a `Map<String, String>` that is unused at the moment.

marcphilipp:
... but maybe the name `LaunchParameter` is a bit confusing then.

sbrannen:
Yes, that's why I recently referred to them as "configuration parameters" which happen to be provided via a launcher. 😉

bechte:
Team discussion / decision:

- Remove ```LaunchParameter```
- Rename "LaunchParameter" to "ConfigurationParameter" and wrap it into an "Configuration object"
- Pass "ConfigurationParameter" into ExecutionRequest

sbrannen:
I think we agreed on something like `Configuration` or `ConfigurationParameters` as the name of the new interface that encapsulates the map of configuration parameters.

sbrannen:
@bechte, the changes look good.

From my perspective, feel free to merge it into `master`.

junit-engine-api/src/main/java/org/junit/gen5/engine/ExecutionRequest.java
```diff
@@ -25,12 +25,16 @@
 
 	private final EngineExecutionListener engineExecutionListener;
 
+	private ConfigurationParameters configurationParameters;
+
 	private final Map<String, Object> attributes = new HashMap<>();
```
marcphilipp: I think `configurationParameters` should replace `attributes` or do you want to do this in a later step?

bechte: When I coded around that line I was also like "let's remove them", but as we not sure about those parameters in our discussion, I didn't want to remove them right away.

Readying your comment now convinces me, that we should remove them now, and if we need them later, we could always add them again.

What about the others?

marcphilipp: IIRC `attributes` were added for exactly the purpose this pull request is solving.

BTW please make the instance variable `final`.

bechte:
I just wonder if we should pass the configuration parameters onto to the extensions? Will this be fully addressed in #78 ?

junit-engine-api/src/main/java/org/junit/gen5/engine/ExecutionRequest.java
bechte: And off they go... 😄 

I removed them and declared the variable final. I will merge this PR later this day.

sbrannen:
> I just wonder if we should pass the configuration parameters onto to the extensions? Will this be fully addressed in #78 ?

Doing so is not intended for #78, but it's something we can consider later.

PR #265, opened 2016-05-15
Introduce deactivation mechanism for conditions

sbrannen:
Prior to this commit, there was no way to deactivate or turn off
support for conditional test execution. In other words, there was no
way to execute a `@Disabled` test and no way to deactivate custom
implementations of `ContainerExecutionCondition` or
`TestExecutionCondition`.

This commit addresses this issue by introducing support for supplying a
`junit.conditions.deactivate` pattern to the `Launcher` via a
configuration parameter in the `TestDiscoveryRequest`.

If the `junit.conditions.deactivate` pattern consists solely of an
asterisk (`*`), all conditions will be deactivated. Otherwise, the
pattern will be used to match against the fully qualified class name
(FQCN) of each registered condition. Any dot (`.`) in the pattern will
match against a dot (`.`) or a dollar sign (`$`) in the FQCN. Any asterisk
(`*`) will match against one or more characters in the FQCN. All other
characters in the pattern will be matched one-to-one against the FQCN.

Examples:

- `*`: deactivates all conditions.
- `org.junit.*`: deactivates every condition under the `org.junit` base
   package and any of its subpackages..
- `*.MyCondition`: deactivates every condition whose simple class
   name is exactly `MyCondition`.
- `*System*`: deactivates every condition whose simple class name
   contains `System`.
- `org.example.MyCondition`: deactivates the condition whose FQCN is
  exactly `org.example.MyCondition`.

Issue: #78

---

### DIFF
--- a/junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/ConditionEvaluator.java
+++ b/junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/ConditionEvaluator.java
@@ -11,15 +11,26 @@
 package org.junit.gen5.engine.junit5.execution;
 
 import static org.junit.gen5.commons.meta.API.Usage.Internal;
+import static org.junit.gen5.engine.junit5.Constants.DEACTIVATE_ALL_CONDITIONS_PATTERN;
+import static org.junit.gen5.engine.junit5.Constants.DEACTIVATE_CONDITIONS_PATTERN_PROPERTY_NAME;
 
+import java.util.function.BiFunction;
+import java.util.function.Predicate;
 import java.util.logging.Logger;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.junit.gen5.api.extension.ConditionEvaluationResult;
 import org.junit.gen5.api.extension.ContainerExecutionCondition;
 import org.junit.gen5.api.extension.ContainerExtensionContext;
+import org.junit.gen5.api.extension.Extension;
+import org.junit.gen5.api.extension.ExtensionContext;
 import org.junit.gen5.api.extension.TestExecutionCondition;
 import org.junit.gen5.api.extension.TestExtensionContext;
 import org.junit.gen5.commons.meta.API;
+import org.junit.gen5.commons.util.StringUtils;
+import org.junit.gen5.engine.ConfigurationParameters;
+import org.junit.gen5.engine.junit5.Constants;
 import org.junit.gen5.engine.junit5.extension.ExtensionRegistry;
 
 /**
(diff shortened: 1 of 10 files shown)


junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/ConditionEvaluator.java
```diff
@@ -108,4 +137,42 @@ private void logResult(Class<?> conditionType, ConditionEvaluationResult result)
...
+	 */
+	private String convertToRegEx(String pattern) {
+		pattern = Matcher.quoteReplacement(pattern.trim());
+
+		// Match "." against "." and "$" since users may declare a "." instead of a
+		// "$" as the separator between classes and nested classes.
+		pattern = pattern.replace(".", "[.$]");
+
+		// Convert our "*" wildcard into a proper RegEx pattern.
+		pattern = pattern.replace("*", ".+");
```
marcphilipp: According to the documentation an asterisk "will match against zero or more characters in the FQCN" while `.+` requires at least one character.

sbrannen: Oops. Yep, you're right.

I'll fix the documentation.

Thanks for pointing that out.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/ConditionEvaluator.java
```diff
@@ -108,4 +137,42 @@ private void logResult(Class<?> conditionType, ConditionEvaluationResult result)
 		LOG.finer(() -> String.format("Evaluation of condition [%s] resulted in: %s", conditionType.getName(), result));
 	}
 
+	private Predicate<Object> conditionIsActivated(ConfigurationParameters configurationParameters) {
+		Pattern deactivatePattern = getDeactivatePattern(configurationParameters);
+		if (deactivatePattern != null) {
+			if (DEACTIVATE_ALL_CONDITIONS_PATTERN.equals(deactivatePattern)) {
```
marcphilipp: This is always `false` because `DEACTIVATE_ALL_CONDITIONS_PATTERN` is a `String`, isn't it?

sbrannen: Right again!

This was the result of some refactoring, and unfortunately the tests continued to pass.

Thanks for pointing this out.

I'll fix it.

marcphilipp:
Looks good now, thanks! 👍

PR #270, opened 2016-05-16
Improve assertEquals for floats and doubles

### LINKED ISSUE #269: Introduct 'delta' support in Assertions for doubles and floats
## Status Quo

Unlike JUnit 4, the `assertEquals()` variants in `org.junit.gen5.api.Assertions` that accept `double` and `float` arguments currently do not support a _delta_. This omission was simply an oversight on our part.
## Deliverables
- [x] Add a `delta` argument to each of the existing `assertEquals()` variants in `org.junit.gen5.api.Assertions` that accept `double` and `float` arguments.

lutovich:
## Overview

This PR includes two commits:

 - Handle NaNs when comparing doubles and floats

 This commit makes methods `assertEquals(double, double)` and
`assertEquals(float, float)` treat NaNs as equal. Same applies to variations
of those methods that accept a message or a message supplier.

 Such behaviour is consistent with `Double#compare(double, double)` and
`Float#compare(float, float)` but not consistent with `==` and `!=`. It will
cause less confusion for users that execute code like
`assertEquals(Double.NaN, Double.NaN)`.

 - Added assertEquals for doubles & floats with delta

 This commit introduces `assertEquals(double, double, double)` and
`assertEquals(float, float, float)` where the third parameter is delta. First
two parameters are compared within this given delta. Also added methods with
delta parameter that accept a message and a message supplier.

 Such methods are needed because exact comparison of floating point numbers
does not always make sense.

Fixes #269 

---

### DIFF
--- a/junit-tests/src/test/java/org/junit/gen5/api/AssertionsTests.java
+++ b/junit-tests/src/test/java/org/junit/gen5/api/AssertionsTests.java
@@ -545,6 +545,18 @@ public class AssertionsTests {
 	}
 
 	@Test
+	void assertEqualsFloatWithIllegalDelta() {
+		AssertionFailedError e1 = expectThrows(AssertionFailedError.class, () -> assertEquals(0.1f, 0.2f, -0.9f));
+		assertMessageEndsWith(e1, "positive delta expected but was: <-0.9>");
+
+		AssertionFailedError e2 = expectThrows(AssertionFailedError.class, () -> assertEquals(.0f, .0f, -10.5f));
+		assertMessageEndsWith(e2, "positive delta expected but was: <-10.5>");
+
+		AssertionFailedError e3 = expectThrows(AssertionFailedError.class, () -> assertEquals(4.5f, 4.6f, Float.NaN));
+		assertMessageEndsWith(e3, "positive delta expected but was: <NaN>");
+	}
+
+	@Test
 	void assertEqualsFloatWithDeltaWithUnequalValues() {
 		AssertionFailedError e1 = expectThrows(AssertionFailedError.class, () -> assertEquals(0.5f, 0.2f, 0.2f));
 		assertMessageEndsWith(e1, "expected: <0.5> but was: <0.2>");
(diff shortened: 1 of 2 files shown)


junit-tests/src/test/java/org/junit/gen5/api/AssertionsTests.java
```diff
@@ -531,8 +538,59 @@ void assertEqualsFloatWithUnequalValuesAndMessageSupplier() {
...
+		assertMessageEndsWith(e1, "expected: <0.5> but was: <0.2>");
+
+		AssertionFailedError e2 = expectThrows(AssertionFailedError.class, () -> assertEquals(0.1f, 0.2f, 0.000001f));
+		assertMessageEndsWith(e2, "expected: <0.1> but was: <0.2>");
+
+		AssertionFailedError e3 = expectThrows(AssertionFailedError.class, () -> assertEquals(100.0f, 50.0f, 10.0f));
+		assertMessageEndsWith(e3, "expected: <100.0> but was: <50.0>");
+
+		AssertionFailedError e4 = expectThrows(AssertionFailedError.class, () -> assertEquals(-3.5f, -3.3f, -42.0f));
+		assertMessageEndsWith(e4, "expected: <-3.5> but was: <-3.3>");
```
marcphilipp: I'm wondering whether we should actually not allow negative deltas. What do you think?

lutovich: This is probably a good idea. JUnit 4 javadoc even says that delta should be positive but code does not check this. I'll add such check.

sbrannen: Agreed: only positive delta values make sense.

lutovich: Added check that delta is not NaN and positive: https://github.com/junit-team/junit5/pull/270/commits/0e44a3f9216e5c359fed4fbf98ecb9ed01f3333d

marcphilipp:
Looks good to me, thanks!

@junit-team/junit-lambda Please take a look! :-)

sbrannen:
Looks good!

I'll go ahead and merge it into `master`.

sbrannen:
This has been merged into `master` in 4ce7a0fc1f588948abfc334598533add203f2830.

Thanks!

Issue #138, opened 2016-01-27
Use 'actual' and 'expected' properties when instantiating AssertionFailedError


jlink:
Use `new AssertionFailedError(message, expected, actual)` in the following cases:
- [X] Change `expected` and `actual` instance variables and parameters to only store `Serializable` values.
- [x] `Assertions.assertEquals`
- [x] ~~`Assertions.assertNotEquals`~~ (see #282)
- [x] `Assertions.assertSame`
- [x] ~~`Assertions.assertNotSame`~~ (see #282)
- [x] `Assertions.assertNull`
- [x] ~~`Assertions.assertNotNull`~~ (see #282)

sbrannen:
Since the API for `AssertionFailedError` in the OTA is currently unstable (i.e., slated to be changed), we should hold off on changing JUnit 5's `Assertions` to depend on this feature.

jlink:
What about the IDEs that would like to provide a comparison view?

bechte:
Wouldn't they be able to extract that information the same way they do it today?

jlink:
Nowadays they use a JUnit 4 exception:
http://junit.sourceforge.net/javadoc/org/junit/ComparisonFailure.html

2016-01-28 18:45 GMT+01:00 Stefan Bechtold <notifications@github.com>:

bechte:
This exception is thrown by JUnit only. The IDEs support comparison views also for `AssertionError`s that are raised by libraries like Hamcrest or AssertJ. As far as I know, none of them are using  `ComparisonFailure`. Therefore, I think it would be fine to use an `AssertionFailedError` that provides a message with a formatting closed to what Hamcrest or AssertJ produces. This would probably be handled by the IDEs wisely and we would not depend on an unstable API.

jlink:
Since @akozlova wanted to use expected/actual in IntelliJ support, I suggest we should implemented.
Since we do already rely on `AssertionFailedError` from opentest4j, actually using the fields does not change the dependency strength much.

I additionally suggest to stringify expected and actual values if they are not serializable.

marcphilipp:
I'm starting to work on this issue.

marcphilipp:
I've addressed the first subtask in opentest4j, see ota4j-team/opentest4j@dada365668cd918d0273f22bbfdcdadd8eac5643 for details. Since I won't have time to work on this issue in the next few days, I'll remove the *in progress* tag.

marcphilipp:
[...]

PR #273, opened 2016-05-21
assertNull/Same/Equals: Fill expected/actual values in AssertionFailedError

marcphilipp:
Issue: #138 

This change fills expected/actual values in AssertionFailedError for 
- `assertNull`
- `assertSame`
- `assertEquals`

Contrary to the subtasks of #138 I don't think it makes sense to fill `expected` and `actual` for `assertNotNull`, `assertNotSame`, and `assertNotEquals` because there's no expected value in these cases. We could change `AssertionFailedError` to allow only specifying an actual value.

---


sbrannen:
@marcphilipp, this looks good to me.

Feel free to merge the PR into `master`.

mmerdes:
fine with me as well

Issue #271, opened 2016-05-19
Introduce a mechanism for associating multiple TestSources with a single test


signed:
With cucumber you have a feature file with the scenario definition and a Java class that implements the steps used in the scenario. If a scenario run fails and I have the test tree in my IDE I like to have the choice to navigate either to the feature file or the implementation of the failing step.

How would an implementation of TestSource look like to support this?

marcphilipp:
Thanks for raising this issue!

I see two possibilities:

1. Provide a `TestSource` implementation that represents multiple sources, e.g. a class `MultiSource` that contains a list of test sources.
2. Replace `TestIdentifier#getSource` and `TestDescriptor#getSource` with `getSources() : List<TestSource>`.

@signed @junit-team/junit-lambda Which one would you prefer?

sbrannen:
I would prefer the first option, and I would suggest introducing a `CompositeTestSource` that is a direct subinterface of `TestSource` with a `List<TestSource> getSources()` method.

sbrannen:
FYI: I have renamed this issue.

mmerdes:
I would prefer the classic composite pattern as well

sbrannen:
_in progress_

sbrannen:
@junit-team/junit-lambda, please review PR #274.

sbrannen:
No longer _in progress_: awaiting team feedback.

signed:
I know too little about the consumers of the API to have a preference. As long as it is possible its fine with me.

sbrannen:
Merged into `master` in 798beef089df5368a335d541140adbf79f665046.

PR #274, opened 2016-05-22
Introduce composite TestSource

sbrannen:
This commit introduces a `CompositeTestSource` that can be used to hold one or more `TestSources`. This can be useful, for example, if the programming model for a particular testing framework relies on Java classes and associated scripts in the file system.

Issue: #271

---

### DIFF
--- a/junit-engine-api/src/main/java/org/junit/gen5/engine/support/descriptor/CompositeTestSource.java
+++ b/junit-engine-api/src/main/java/org/junit/gen5/engine/support/descriptor/CompositeTestSource.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2015-2016 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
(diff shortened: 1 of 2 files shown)


mmerdes:
Looks good: defensive and immutable.

junit-engine-api/src/main/java/org/junit/gen5/engine/support/descriptor/CompositeTestSource.java
```diff
@@ -0,0 +1,82 @@
...
+ * <p>{@code CompositeTestSource} and its {@link #getSources sources} are immutable.
+ *
+ * @since 5.0
+ */
+@API(Experimental)
+public class CompositeTestSource implements TestSource {
+
+	private static final long serialVersionUID = 1L;
+
+	private final List<TestSource> sources;
```
signed: Where do you see ordering becoming relevant for the test sources?

marcphilipp: I think it makes sense to preserve the order because users that are presented with the choice between multiple sources should always see the same order.

However, since we copy the input anyway (which I think is a good idea!), I think we should change the constructor to accept `Collection<? extends TestSource>`.

Issue #153, opened 2016-02-12
Clarify differences between DisplayName & Name throughout JUnit 5


schauder:
I understand the need for a UniqueId and a DisplayName, which I guess is only unique among the children of single a parent. But what is the Name for?

----
### Related To
- #145
- #162

marcphilipp:
Related to #145.

jlink:
We discussed the issue today in a team call. I try to summarize.

Current situation:
- The `DisplayName` property describes the (hopefully) easy to read description of a test case or a test container. It currently defaults to a test method's simple name (without class); in the test container case it describes the classes fully qualified name. The default display name can be overridden by `@DisplayName("my display name")`. 
- The `Name` property is independent from display name. For a container it's the same as the default display name, but in the case of a test it's a concatenation of class name, method name and method parameters.
- The `Source` property points to a class or a method object.

The behaviour described above seems inconsistent. I therefore suggest the following changes:
- The contents of `Name` should be the default value for `DisplayName`.
- `Name` should result in the simple class name for containers (also for nested containers) and the method name (without params) for tests. 

    The information currently coded in the `Name` property can be retrieved via `TestDescriptio.getSource()` and - in case of a test - through an additional `getParent().getSource()`. The downside is that the Eclipse runner will no longer display the package name of a container; but this is already the case if the display name is being changed using `@DisplayName` (IntelliJ only shows the 'simple' form in all situations). This will no longer be an issue as soon as Eclipse provides native integration for JUnit 5. Having packages as inbetween containers would also mitigate this problem.
- `TestInfo` gets two additional properties: 
  1. `Optional<Class> getTestClass`
  2. `Optional<Method> getTestMethod`

Having a simpler form of default display name will also simplify a computed "prettification" of display names (see #162).

sbrannen:
@jlink, thanks for the detailed write up!

I haven't investigated the proposal in detail with regard to the code and any possible ramifications, but in general the proposal seems like a good idea.

jlink:
Rethinking the whole issue, I think that `getName()` is too strongly focused on class-based containers and method based tests (as used by JUnit4 and JUnit5). It does not really make sense for other types of tests like file-based or lambda-based ones. I therefore suggest to get rid of `getName()` in `TestIdentifier` and `TestDescriptor`. The information lost can be got via `getSource()` (see also #175).

[...]

PR #279, opened 2016-05-23
Overhaul display name semantics

sbrannen:
Previously there was considerable confusion with regard to the
differences between names and display names within JUnit 5.

This pull request addresses these issues by consolidating the two
competing concepts into a single "display name" concept. Test classes
and test methods may be assigned custom display names via the
`@DisplayName` annotation. Otherwise, a default display name will be
generated.

Specifically, the default display name for a test class is the fully
qualified name of the class.

The default display name for a test method is the name of the method
concatenated with a comma-separated list of parameter types in
parentheses, and the names of parameter types are retrieved using
`Class#getSimpleName()`. For example, the default display name for the
following test method is `testUser(TestInfo, User)`.

```java
  @Test
  void testUser(TestInfo testInfo, @Mock User user) { ... }
```

Issue: #153

---


jlink:
Using FQCN as default display name for test classes makes it difficult for test writers to just make some test classes more readable via "@DisplayName" since the more readable version should also contain the package(s). Otherwise the mixture of long and short names will look either strange in the IDE or the IDE will have to display it differently depending on if the package(s) are present or not.

marcphilipp:
@jlink I agree. However, since packages are not present in the test tree at the moment it might be confusing to use simple names for classes. I think we could use the simple names only if we added packages (as @sbrannen [suggested in #153](https://github.com/junit-team/junit5/issues/153#issuecomment-220847441)).

But I don't see us doing that for M1, especially because it's unclear how. There are at least two ways. For example for a class org.foo.SomeTest and org.foo.bar.AnotherTest the test tree could look like

1. Flat
  - org.foo
    - SomeTest
  - org.foo.bar
    - AnotherTest
2. Deep
  - org
    - foo
      - bar
        - AnotherTest
      - SomeTest

I'm undecided as to which one I would prefer.

jlink:
Having the package now and remove it later in M2 will force the IDEs to change their display policy. If an IDE wants to display the package name it can always use the test source.

sbrannen:
> If an IDE wants to display the package name it can always use the test source.

Yes, I think that's the crucial thing to keep in mind: an IDE or build tool can always use the `TestSource`.

That is in fact what I had to do in our XML report generator for this PR since `getName()` is no longer available. Consequently, we are now using the `JavaClassSource` to retrieve the FQCN for use in XML reports.

In general, IDEs and build tools should only use the _display name_ for display purposes, not for building a tree in the UI. From that perspective, it should not matter what is _in_ the display name.

However, I do see the point that using a FQCN as the default display name leads to rather long display names in the UI.

In fact, my original proposal in #153 stripped the package names from class names. So I am not opposed to doing so. Plus, as you mentioned @jlink, it is probably better to do it in M1 than changing it in M2. The only downside is that navigation back to a test class from the UI tree in IDEs likely will not work at all. But perhaps it's better to bite the bullet now.

sbrannen:
@marcphilipp, I think we should defer the decision regarding what the test tree will physically look like (i.e., in terms of flattening packages or not) until M2.

marcphilipp:
> However, I do see the point that using a FQCN as the default display name leads to rather long display names in the UI.
> 
> In fact, my original proposal in #153 stripped the package names from class names. So I am not opposed to doing so. Plus, as you mentioned @jlink, it is probably better to do it in M1 than changing it in M2. The only downsides are that navigation back to a test class from the UI tree in IDEs likely will not work well, and the trees in UIs will be flat without the package name in the display name (at least in Eclipse). But perhaps it's better to bite the bullet now.

Ok, you've convinced me. I think we should use the simple name and document how to retrieve the FQCN in case IDEs want to display it in M1.

> I think we should defer the decision regarding what the test tree will physically look like (i.e., in terms of flattening packages or not) until M2.

Point taken. I was just trying to illustrate what I was talking about. 😉

sbrannen:
Thanks for the feedback, guys!

I'll go ahead and switch to simple names for classes, except that I'll include the enclosing top-level class(es) for nested static classes (in line with my original proposal).

marcphilipp:
👍

junit-tests/src/test/java/org/junit/gen5/console/tasks/TempDirectory.java
```diff
@@ -71,7 +71,15 @@ private Namespace localNamespace(ExtensionContext context) {
 
 	private Path createTempDirectory(ExtensionContext context) {
 		try {
-			return Files.createTempDirectory(context.getName());
+			String tempDirName;
+			if (context.getTestMethod().isPresent()) {
+				tempDirName = context.getTestMethod().get().getName();
+			}
+			else {
+				tempDirName = context.getTestClass().get().getName();
```
marcphilipp: This call to `get()` strikes me as not being save in all cases. Can we use an additional fallback if it the test class is not present, e.g. `getDisplayName()`?

junit-console/src/main/java/org/junit/gen5/console/tasks/XmlReportWriter.java
```diff
@@ -125,6 +126,17 @@ private void writeTestcase(TestIdentifier test, NumberFormat numberFormat, XMLSt
...
 	}
 
+	private String getParentClassName(TestIdentifier test) {
+		// @formatter:off
+		return reportData.getTestPlan().getParent(test)
+				.flatMap(TestIdentifier::getSource)
+				.filter(JavaClassSource.class::isInstance)
+				.map(JavaClassSource.class::cast)
+				.map(source -> source.getJavaClass().getName())
+				.orElse("<unrooted>");
```
marcphilipp: Can we use the display name of the parent as a fallback? I think `<unrooted>` is really misleading. We can do it in a separate issue if you'd prefer.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestDescriptor.java
```diff
@@ -55,12 +56,12 @@ protected JUnit5TestDescriptor(UniqueId uniqueId) {
...
 	}
 
-	protected String determineDisplayName(AnnotatedElement element, String defaultName) {
+	protected String determineDisplayName(AnnotatedElement element, Supplier<String> defaultNameGenerator) {
 		// @formatter:off
 		return findAnnotation(element, DisplayName.class)
 				.map(DisplayName::value)
 				.filter(StringUtils::isNotBlank)
-				.orElse(defaultName);
+				.orElse(defaultNameGenerator.get());
```
marcphilipp: I think this should be `orElseGet(defaultNameGenerator)`, right?

junit5-api/src/main/java/org/junit/gen5/api/DynamicTest.java
```diff
@@ -40,31 +43,33 @@
 @API(Experimental)
 public class DynamicTest {
 
-	private final String name;
+	private final String displayName;
 	private final Executable executable;
 
 	public DynamicTest(String name, Executable executable) {
```
marcphilipp: Nitpick: Please also change the parameter name.

junit5-api/src/main/java/org/junit/gen5/api/DynamicTest.java
```diff
@@ -40,31 +43,33 @@
...
-		this.name = name;
+		this.displayName = name;
 		this.executable = executable;
 	}
 
-	public String getName() {
-		return name;
+	public String getDisplayName() {
+		return this.displayName;
```
marcphilipp: Good catch BTW! 👍

marcphilipp:
I commented a few minor things inline. Besides those, this PR looks really good! 👍

sbrannen:
Update: I just pushed the changes for shorter default display names for test classes.

junit5-api/src/main/java/org/junit/gen5/api/DynamicTest.java
sbrannen: done

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/descriptor/JUnit5TestDescriptor.java
sbrannen: Ummmm.... yeah.... except that the argument is now `Function<AnnotatedElement, String>`. 😉

junit-console/src/main/java/org/junit/gen5/console/tasks/XmlReportWriter.java
sbrannen: Well, this was the previous code:

```java
writer.writeAttribute("classname", parent.map(TestIdentifier::getName).orElse("<unrooted>"));
```

So I didn't want (and don't want) to change the existing behavior in this PR.

I assumed it was done that way for a reason.

In any case, feel free to raise a new issue if you think this should be changed.

junit-tests/src/test/java/org/junit/gen5/console/tasks/TempDirectory.java
sbrannen: It's _only_ test code, but sure... I'll make it more robust. 😉

PR #289, opened 2016-05-29
Array assertions

lutovich:
## Overview

This commit introduces all possible variants of `assertArrayEquals` method for:
 - primitive types
 - `float` arrays within the given `float` delta
 - `double` arrays within the given `double` delta
 - object arrays, which are examined deeply like in `Arrays#deepEquals()`

---

### DIFF
--- a/junit-tests/src/test/java/org/junit/gen5/api/AssertionsTests.java
+++ b/junit-tests/src/test/java/org/junit/gen5/api/AssertionsTests.java
@@ -11,6 +11,7 @@
 package org.junit.gen5.api;
 
 import static org.junit.gen5.api.Assertions.assertAll;
+import static org.junit.gen5.api.Assertions.assertArrayEquals;
 import static org.junit.gen5.api.Assertions.assertEquals;
 import static org.junit.gen5.api.Assertions.assertFalse;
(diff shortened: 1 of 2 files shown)


lutovich:
Array assertions are in `Assertions` class and not in a dedicated helper like `ArrayAssertions` because:
 1) this reminds JUnit 4 and might feel more native to users
 2) they reuse private helper methods from `Assertions` to format values, compare numbers, etc.

At the same time, users mostly static import all assertion methods anyway so `ArrayAssertions` could be a good place. This will keep JUnit code a bit cleaner and make `Assertions` class a bit smaller. Item (2) could be solved by some helper class that holds all message formatting functionality.

Please let me know if array assertions should be moved to a separate class.

sbrannen:
_in progress_

sbrannen:
This has been merged into `master` in 0ef3701114f5b11fb463d8e42b998f7be98072b7.

sbrannen:
@lutovich,

Thanks for the contribution.

Well implemented and thoroughly tested! 👍 

Regarding creating a new `ArrayAssertions` class, I agree that `Assertions` has become quite large in the meanwhile, but let's hold off on that for now.

PR #302, opened 2016-06-11
Introduce ParameterContext

### LINKED ISSUE #300: ParameterResolver has Optional parameter in `supports` and `resolve` methods
Optional was designed to provide a limited mechanism for library method return types where there needed to be a clear way to represent "no result"

marcphilipp:
Fixes #300.

---


junit5-api/src/main/java/org/junit/gen5/api/extension/ParameterResolver.java
```diff
@@ -32,61 +31,46 @@
...
+ * @see ParameterContext
  */
 @API(Experimental)
 public interface ParameterResolver extends Extension {
 
 	/**
-	 * Determine if this resolver supports resolution of the given {@link Parameter}
-	 * for the supplied {@code target} and {@link ExtensionContext}.
+	 * Determine if this resolver supports resolution of the given
+	 * {@link ParameterContext} in the supplied {@link ExtensionContext}.
```
sbrannen: I would use wording in line with the original text.

A `ParameterResolver` is still intended to resolve parameters, not parameter contexts.

So that would lead to wording like this:

> Determine if this resolver supports resolution of the Parameter in the supplied ParameterContext for the supplied ExtensionContext.

The same applies to the wording of the `resolve()` method.

junit5-api/src/main/java/org/junit/gen5/api/extension/ParameterResolver.java
```diff
@@ -32,61 +31,46 @@
...
 	 *
-	 * <p>The {@link java.lang.reflect.Method} or {@link java.lang.reflect.Constructor}
-	 * in which the {@code parameter} is declared can be retrieved via
-	 * {@link Parameter#getDeclaringExecutable()}.
```
sbrannen: I think this tip is very useful for the average extension developer.

As such, I would not remove it but rather reword it to point to the new `getDeclaringExecutable()` method you've introduced in the `ParameterContext`.

The same applies to the `resolve()` method.

junit5-engine/src/main/java/org/junit/gen5/engine/junit5/execution/ParameterContextImpl.java
```diff
@@ -0,0 +1,58 @@
...
+package org.junit.gen5.engine.junit5.execution;
+
+import java.lang.reflect.Parameter;
+import java.util.Optional;
+
+import org.junit.gen5.api.extension.ParameterContext;
+import org.junit.gen5.commons.util.Preconditions;
+import org.junit.gen5.commons.util.ToStringBuilder;
+
+class ParameterContextImpl implements ParameterContext {
```
sbrannen: We don't use the "Impl" suffix anywhere in the JUnit 5 code base.

So please change the name to `DefaultParameterContext`.

sbrannen:
@marcphilipp, I've reviewed the PR and added a few comments.

Otherwise, the PR looks good and can be merged into `master`. 👍

marcphilipp:
@sbrannen Thanks for reviewing this pull request! I made the proposed changes (thanks!) and merged to master.

Issue #144, opened 2016-02-01
Introduce tools for @API annotation processing


jlink:
Do something useful with `@API` annotations, e.g.
- Extract all public APIs
- Compare two versions of JUnit 5 for release notes
- Check correct usage of APIs (internally and externally)

sbrannen:
FYI: this issue is related to #436.

smoyer64:
As noted in PR #564, I've started working on a small set of tools to perform the functions listed above.  As I've been working through that code, I found that I have a few more questions:

-   Should @API also be applied to protected-scoped and package-scoped elements as well as publicly-scoped elements?  I can argue both ways on this one but if the policy is that you only use public fields, methods, constructors that are not annotated with Usage.INTERNAL (or Usage.EXPERIMENTAL/Usage.DEPRECATED at your own risk), then it's relatively safe to assume the elements needed are going to be public.

-   It's clearly stated that if an element isn't annotated explicitly with @API, it's inherited from the enclosing class.  I'll also walk the inheritance hierarchy to find the annotation if needed.  But what about methods and constructors that are @Override?  I think one of the important uses of @API is to encourage extension authors to continue to use @API but the application of the annotation will need to be unambiguously define.  The implementation of interfaces in some ways recreates the "diamond problem" for @API.

marcphilipp:
> Should @API also be applied to protected-scoped and package-scoped elements as well as publicly-scoped elements?

There's (at least) two ways to look at it: (1) protected-/package-scoped elements are implicitly annotated with `@API(Internal)`, or (2) we only look at public elements. I think either way, the outcome is the same, right?

> It's clearly stated that if an element isn't annotated explicitly with @API, it's inherited from the enclosing class. I'll also walk the inheritance hierarchy to find the annotation if needed. But what about methods and constructors that are @Override?

I think we should enforce that an overridden method needs to explicitly re-declare an `@API` annotation.

@junit-team/junit-lambda What do you think?

smoyer64:
@marcphilipp 

>   I think we should enforce that an overridden method needs to explicitly re-declare an @API annotation.

As I've gotten a bit deeper into this coding, I agree that explicitly overridden methods should re-declare the @API annotation.  This should also apply to interface method implementations as it eliminates a form of "the diamond problem".

I was hoping to be able to "lint" M3 but I'm not sure I'll finish in time for that.

33epico:
Hi all:

[...]

PR #317, opened 2016-06-18
Make @API public to allow Extensions and TestEngines to adopt API semantics

smoyer64:
Making @API public allows Extension and TestEngine developers to adopt JUnit 5's API "typing".  This also makes the API tools proposed in #144 more useful as they'd be applicable to the other projects.

---

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/meta/API.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/meta/API.java
@@ -10,7 +10,7 @@
 
 package org.junit.platform.commons.meta;
 
-import static org.junit.platform.commons.meta.API.Usage.Internal;
+import static org.junit.platform.commons.meta.API.Usage.Experimental;
 
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
@@ -31,7 +31,7 @@ import java.lang.annotation.Target;
 @Target({ ElementType.TYPE, ElementType.METHOD, ElementType.CONSTRUCTOR })
 @Retention(RetentionPolicy.CLASS)
 @Documented
-@API(Internal)
+@API(Experimental)
 public @interface API {
 
 	Usage value();


sbrannen:
Interesting proposal!

We'll consider it for M2.

volkovs:
@smoyer64 if `@API` is accepted to be public, should not it be moved from junit-commons to junit-api module? Usage in 3-rd party code will require production code dependency on junit-api, but at least junit-api is just interfaces (super lightweight).

smoyer64:
@volkovs The junit-jupiter-api module already has a dependency on junit-platform-commons.  My initial thought was that I'd like to use @API and it's related tools elsewhere - in which case a completely standalone module would be better.  When Java 9 is released, project jigsaw is supposed to take care of hiding internal-only API calls (per [JEP 260: Encapsulate Most Internal APIs](http://openjdk.java.net/jeps/260)).  Will we get enough tools to replace @API?  I have no idea.

sbrannen:
@volkovs, moving `@API` to the `junit-jupiter-api` module is forbidden: various parts of the JUnit Platform use `@API`, and Platform modules must *not* have any dependencies on JUnit Jupiter or JUnit Vintage modules.

I agree with @smoyer64: one viable option would be to move `@API` to its own artifact. The JUnit team has in fact already discussed such a possibility.

sbrannen:
*Team Decision*: make _Experimental_.

sbrannen:
Would you mind rebasing on `master` and squashing into a single commit?

marcphilipp:
We could just squash and merge through GitHub, couldn't we?

sbrannen:
I'm not sure if that new GitHub _squash_ feature physically rebases on `master`.

smoyer64:
I tried to use git commit --squash (but now know I should stick with git rebase).  Sorry!

sbrannen:
I usually do something like this:

```
> git rebase master
> git rebase --interactive --autosquash HEAD~4
```

Then I select "f" (for _fix_) for the last three commits and reword the remaining commit if necessary.

smoyer64:
Okay ... I'm sticking with my favorite method (for branches) - it's back to a single commit!

```
git reset master
git add ...
git commit -m "<altered message>"
git push -f
```

sbrannen:
Aye aye, captain!

Whatever floats your boat... and gets your commits clean. 😉

smoyer64:
Someone seems to have noticed my avatar?

sbrannen:
ummmmmmm

no....

not consciously at least.

But now that you mention it, perhaps that was my subconscious telling me what to type. 😉

sbrannen:
FYI: this issue is related to #436.

PR #324, opened 2016-06-20
Don't state implementation details in Javadoc.

stefanbirkner:
## Overview

The algorithm for creating the returned filter is an implementation detail that should not be fixed by the API specification (= Javadoc). This makes it easier to modify the implementation in the future.

---


sbrannen:
Thanks! I'll merge this as soon as the CI build passes.

stefanbirkner:
Tests are green again.

PR #325, opened 2016-06-20
Make DiscoveryRequest immutable.

stefanbirkner:
## Overview

Avoid code duplication with `TestDiscoveryRequestBuilder`. Having two objects with similar responsibility is confusing, too. For the reader it is unclear which one to use.

---


marcphilipp:
Thanks!

PR #326, opened 2016-06-20
Delete EngineDiscoveryRequest.getSelectors().

stefanbirkner:
## Overview

This method is only used by tests. Production code always uses the method `getSelectorsByType`. By removing the method `getSelectors()` we have one method less that has to be supported and may constrain future decisions.

---


sbrannen:
Good point.

I'll go ahead and merge this, and if somebody requests a simple `getSelectors()` method later on we could always add it back.

PR #327, opened 2016-06-20
Remove section about semantic versioning.

stefanbirkner:
## Overview

This project uses the `@API` annotation for information about API compatibility. The rules defined for this purpose do not comply with the [semantic versioning specification](http://semver.org/).

---

### DIFF
--- a/documentation/src/docs/asciidoc/api-evolution.adoc
+++ b/documentation/src/docs/asciidoc/api-evolution.adoc
@@ -11,7 +11,8 @@ classes, and methods.
 
 === API Annotations
 
-All publicly available interfaces, classes, and methods are annotated with {API}. The
+Every published artifact has a version number `<major>.<minor>.<patch>` and all
+publicly available interfaces, classes, and methods are annotated with {API}. The
 annotation's `Usage` value can be assigned one of the following five values:
 
 [cols="20,80"]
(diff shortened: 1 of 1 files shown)


marcphilipp:
This may be bending the rules: But if we'd define public API to be all classes that are annotated with `@API(Stable)` I think we'd be fine, right?

stefanbirkner:
I don't agree. We are setting wrong expectations by using the term "semantic versioning". Why should we (mistakable) use this term, even though there is no necessity to use it?

sbrannen:
I agree with @marcphilipp.

Java 9 will support modules and visibility with Project Jigsaw, but we are not there yet. We are on Java 8, and as such we have the _problem_ that `public` types and methods may in fact **not** be for "public consumption".

Thus, IMHO, we have to do the best we can do given the resources at hand. In other words, I think it is legitimate to claim that we follow the rules of _semantic versioning_ for things we explicitly declare for "public consumption." And we achieve the latter via `@API` declarations.

Consequently, I agree with Marc's proposal to use `@API(Stable)` to signify exactly what is for public consumption following the rules of semantic versioning.

stefanbirkner:
Could you explain why you want to keep the section about "semantic versioning". I would like to understand the reason behind.

marcphilipp:
@stefanbirkner Giving it a few days to sink in, I think you're right. We don't need the label "semantic versioning". Since our definition of "public API" isn't obvious it might be confusing to some people and thus do more harm than good.

PR #328, opened 2016-06-21
Remove abandoned TODO about configuring Launcher.

stefanbirkner:
## Overview

The TODO is there since the beginning. The current `Launcher` interface does no allow any configuration except of adding listeners.

---


sbrannen:
thanks!

PR #338, opened 2016-06-21
Always use status code on exit.

stefanbirkner:
## Overview

In general I consider it helpful to always return a status code that is not 0 in case of a failure. This makes it easy for scripts to decide whether all tests succeeded. In addition I cannot imagine any use case that relies on returning 0 in case of a failure.

The old implementation returned the number of failing tests as status code if the flag `enable-exit-code` has been set. But this is only possible for a small number (< 256) of failing tests. The problem was that n*256 failing tests have been reported as 0 failing tests and therefore could not be distinguished from a successful run via status code.

The parser in `AvailableOptions` still accepts the `enable-exit-code` flag because the project is build with an older version of the Gradle plugin from https://oss.sonatype.org/content/repositories/snapshots The parser does not to have accept the flag anymore once a new version of the Gradle plugin is pushed to the snapshot repository.

---

### DIFF
--- a/documentation/src/docs/asciidoc/running-tests.adoc
+++ b/documentation/src/docs/asciidoc/running-tests.adoc
@@ -232,6 +232,9 @@ Test run finished after 29 ms
 [         0 tests failed    ]
 ----
 
+The `{ConsoleRunner}` exits with status code 0 if no test failed. In case of at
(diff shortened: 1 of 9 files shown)


sbrannen:
Good catch!

PR #339, opened 2016-06-21
Add missing case to Javadoc.

stefanbirkner:
## Overview

-

---


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterAll.java
```diff
@@ -30,7 +30,8 @@
  * of {@code @AfterAll}.
  *
  * <p>In contrast to {@link AfterEach @AfterEach} methods, {@code @AfterAll}
- * methods are only executed once for a given test class.
+ * methods are only executed once for a given test class or test class
+ * hierarchy.
```
sbrannen: Hmmmmm....

This is not actually true.

An `@AfterAll` method will be executed for each test class within a test class hierarchy.

So... it's once per test class and N number of times per hierarchy, where N is the number of concrete test classes in the hierarchy.

sbrannen:
Closing this PR since it has been superseded by #340.

PR #342, opened 2016-06-21
Don't accept flag enable-exit-code.

stefanbirkner:
## Overview

This is a follow-up to e568a3563c417829a20eec988daa98797c8391b3. That commit removed support for the flag `enable-exit-code`. It is already deployed to the snapshot repository. Therefore we can now remove support for this flag from `AvailableOptions`.

---


sbrannen:
Thanks!

PR #365, opened 2016-06-28
Applied skeletal implementation to interface migration refactoring.

khatchad:
## Overview
This is a semantics-preserving refactoring.
- We are evaluating a research prototype automated refactoring Eclipse plug-in called [Migrate Skeletal Implementation to Interface](https://github.com/khatchad/Migrate-Skeletal-Implementation-to-Interface-Refactoring). We have applied the tool to your project in the hopes of receiving feedback.
- The approach is very conservative and you should not observe many source code changes.
- The source code should be semantically equivalent to the original
- We have run tests prior to applying to the tool and following the application. All tests pass.

Thank you for your help in this evaluation!

---

### DIFF
--- a/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ExtensionRegistryTests.java
+++ b/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ExtensionRegistryTests.java
@@ -120,37 +120,23 @@ public class ExtensionRegistryTests {
 
 	interface MyExtensionApi extends Extension {
 
-		void doNothing(String test);
+		default void doNothing(String test) {
+		}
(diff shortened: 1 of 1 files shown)


sbrannen:
@khatchad, if you could make a Gradle plugin out of it (analogous to [Spotless](https://github.com/diffplug/spotless)), that could be quite interesting.

As for the results of your application of the refactoring to the JUnit 5 code base, it only made changes to dummy interfaces and classes used within the test suite. In light of that, we won't be merging this PR.

However, the tool looks quite promising. Keep up the good work!

khatchad:
Thank you for the feedback, @sbrannen! We we have a look at Spotless.

Issue #376, opened 2016-07-04
Update MockitoExtension example in User Guide


marcphilipp:
The user guide still uses the `@InjectMock` annotation which has been replaced by the `@Mock` annotation of Mockito 2.x.

This issue was originally raised in junit-team/junit-team.github.io#1.

PR #377, opened 2016-07-04
User guide fix: @InjectMock -> @Mock

volkovs:
User guide fix: #376 
I have found only 2 instances of `@InjectMock`.

---


marcphilipp:
Thanks! 👍

PR #385, opened 2016-07-09
Fix Surefire plugin group ID in documentation

nicolaiparlog:


sbrannen:
Good catch. 👍 

Thanks!

PR #389, opened 2016-07-10
Use assertThrows instead of expectThrows.

stefanbirkner:
According to our own documentation `assertThrows` should be used if the test does not deal with the return value of `expectThrows`.

---


sbrannen:

Thanks

PR #391, opened 2016-07-11
Use diamond operator for creating Stack objects.

stefanbirkner:
Java 8 code should look like Java 8 code.

---


sbrannen:
Thanks for paying attention to detail. 👍

PR #394, opened 2016-07-11
Use meaningful names for @BeforeEach methods.

stefanbirkner:
## Overview

Makes it easier for the reader to understand the purpose of the methods. Additionally it shows the reader of the user guide that they can use meaningful names for `@BeforeEach` methods.

---


marcphilipp:
Thanks!

Issue #399, opened 2016-07-14
Support classpath scanning within JAR files


nilsga:
`DiscoverySelector.selectPackage` does not work when the package resides within a jar file. This is due to https://github.com/junit-team/junit5/blob/master/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java#L122, since `listFiles` does not work for a "folder" within a jar file.

marcphilipp:
_in progress_

marcphilipp:
@nilsga I've decided to only use the test you've provided in #400 (I cherry-picked your commit, see 2205313c19223ae05aa98928f270ae005b639378) and implement it using the `Path` API (as proposed in #466) so there's only one implementation for both cases.

Please take a look at the current [implementation](https://github.com/junit-team/junit5/blob/cbfdd3039a91578a09290d895621defc263bbbba/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java). Basically, it collects all base `URI` for a package, converts them into a `Path` based on their scheme and uses `Files.walkFileTree` for traversal.

Please confirm that this fix solves your problem, so we can close this issue.

nilsga:
Looks good. The problem seems to be resolved with this fix.

sbrannen:
Awesome! Thanks for letting us know it works for you.

PR #400, opened 2016-07-14
Support classpath scanning within JAR files

nilsga:
`DiscoverySelectors.selectPackage` does not currently work with packages
within JAR files.

This commit resolves this issue by traversing the entries within a match,
if it is a JAR file, and adding the candidate classes for discovery. 

Issue: #399

---


marcphilipp:
Can you please add a test for finding classes within a JAR file?

nilsga:
Yes, I was planning to. But I'm unsure how to write a repeatable, platform-independent test for this case. I could try to load a package from one of the other submodules, or a dependency. But am I guaranteed that these always will resolve as a jar file? I have a solution where I create a temporary jar file with a runtime compiled class, and add this class to a custom classloader. It works, but seems like overkill...

nilsga:
Test added, @marcphilipp.

PR #413, opened 2016-07-19
Fix nullSafeToString NullException

smallyard:
## Overview

-

Fix that when call like StringUtils.nullSafeToString(String.class, null, Integer.class) there throws a NullException.

---

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
@@ -93,7 +93,7 @@ public final class StringUtils {
 		if (classes == null || classes.length == 0) {
 			return "";
 		}
-		return stream(classes).map(mapper).collect(joining(", "));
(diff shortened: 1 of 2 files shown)


sbrannen:
_in progress_

sbrannen:
Thanks for the pull request!

I've merged it into `master` with minor modifications.

sbrannen:
FYI: `StringUtils` is an unsupported _Internal_ utility.

If you encountered the NPE in your own code that uses `StringUtils`, then that's simply unsupported.

However, if you encountered the NPE via standard use of supported APIs in JUnit, that would make this a _bug_, and we would be grateful to know under what circumstances the NPE was thrown.

So, out of curiosity, how did you run into the issue with the `NullPointerException`?

smallyard:
@sbrannen Thank you for your reply,  I just encountered the NPE in my own code.

sbrannen:
> I just encountered the NPE in my own code.

OK. Thanks for letting us know.

But... be careful about using anything in JUnit annotated with `@API(Internal)` (such as `StringUtils`), since doing so is literally completely unsupported and subject to change and/or deletion without notice.

PR #424, opened 2016-07-23
Add Preconditions.notBlank of key and value in ReportEntry.from

smallyard:
## Overview

-
Add Preconditions.notBlank of key and value in ReportEntry.from
---


sbrannen:
I am closing this issue as _invalid_ since the preconditions are already checked in `ReportEntry.add(String, String)` which is invoked by the factory method you modified in your pull request.

PR #427, opened 2016-07-23
HierarchicalTestEngine is no longer a generic type.

stefanbirkner:
## Overview

The engine did not use the type parameter. Making HierarchicalTestEngine a simple type reduces complexity.

---

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/JupiterTestEngine.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/JupiterTestEngine.java
@@ -21,13 +21,14 @@ import org.junit.platform.engine.EngineDiscoveryRequest;
 import org.junit.platform.engine.ExecutionRequest;
 import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.hierarchical.EngineExecutionContext;
 import org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine;
 
 /**
  * @since 5.0
  */
 @API(Experimental)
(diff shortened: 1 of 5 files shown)


marcphilipp:
Hmm... 🤔

I would rather want to enforce that the `TestDescriptors` a `HierarchicalTestEngine<C>` returns that implement `Node` also use `<C>` as type parameter.

sbrannen:
I'm afraid we cannot accept this PR since it removes compile-time type checking in implementations of `TestDescriptor` that also implement `Node`.

Specifically, with the changes suggested in this PR it is possible to compile such a `TestDescriptor`/`Node` that accepts a different type of `EngineExecutionContext` than the one created by the `HierarchicalTestEngine`.

As a proof of concept, I just implemented such a scenario, and as I suspected it results in a `ClassCastException` similar to the following:

```
java.lang.ClassCastException: org.junit.platform.engine.support.hierarchical.DummyEngineExecutionContext cannot be cast to org.junit.platform.engine.support.hierarchical.ForeignEngineExecutionContext
```

- `DummyEngineExecutionContext` is created by the engine.
- `ForeignEngineExecutionContext` is the type used by the `TestDescriptor`/`Node`.

So, although the proposal in this PR results in less code, it unfortunately removes the compile-time type safety that we already have in place.

sbrannen:
FYI: in case you read it, I have retracted my previous comment. We encounter a `ClassCastException` at run time both on `master` and for this PR.

marcphilipp:
Thanks again for the pull request, @stefanbirkner!

We've discussed it in the team and have decided against merging it. Even though it does not provide real compile-time type checking, we think it reveals the intention of using the same `EngineExecutionContext` subtype throughout the whole implementation of an `HierarchicalTestEngine`, i.e. also in the TestDescriptor the engine creates.

Issue #438, opened 2016-07-26
Introduce lambda based timeout assertions


sbrannen:
## Status Quo

`Assertions` provides `assertThrows` and `expectThrows` for expected exceptions but does not provide support for asserting timeouts.
## Deliverables
- [x] Introduce lambda based timeout support in `Assertions`.

sbrannen:
Merged into `master` in 0aa724bb2f7c07143ffef6bddba5f9771c629f1d.

PR #439, opened 2016-07-26
Introduce lambda based timeout assertions

sbrannen:
## Overview

This commit introduces `assertTimeout()` methods in `Assertions` that support both lazy and preemptive timeouts for blocks of code supplied as lambda expressions.

For example, the following assertion would pass:

`assertTimeout(ofMillis(500), () -> Thread.sleep(10));`

Whereas, the following assertion would fail:

`assertTimeout(ofMillis(10), () -> Thread.sleep(500));`

The above examples execute the code block in the same thread as the caller. To perform the same assertion as the last example and have the code executed in a separate thread and then preemptively aborted, the above can be rewritten as follows.

`assertTimeoutPreemptively(ofMillis(10), () -> Thread.sleep(500));`

Issue: #438

---

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
@@ -15,10 +15,17 @@ import static org.junit.platform.commons.meta.API.Usage.Experimental;
 import static org.junit.platform.commons.meta.API.Usage.Maintained;
 import static org.junit.platform.commons.util.ReflectionUtils.isArray;
 
+import java.time.Duration;
 import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.Deque;
 import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


sbrannen:
@junit-team/junit-lambda, please let me know your thoughts on the PR.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -996,6 +1005,154 @@ public static void assertThrows(Class<? extends Throwable> expectedType, Executa
 			String.format("Expected %s to be thrown, but nothing was thrown.", expectedType.getName()));
 	}
 
+	// --- assert timeouts -----------------------------------------------
+
+	/**
+	 * <em>Asserts</em> that execution of the supplied {@code executable}
+	 * completes before the given {@code timeout} is exceeded.
+	 *
+	 * <p>Note: the executable will be executed in the same thread as	 that
```
ttddyy: nit: space between "as" and "that". same for other timeout methods

sbrannen: good catch! (y)

I'll fix it.

marcphilipp:
Looks good to me! Great to finally have support for timeouts! 👍

sbrannen:
Thanks for the review, and... I couldn't agree more: it definitely is great to finally have official support for timeouts in JUnit 5!

PR #440, opened 2016-07-26
Change use of "JUnit Platform launcher" to "runner"

nicolaiparlog:
One slight error: you've mixed up junit-platform-runner and junit-platform-launcher. The former contains the JUnitPlatform class which is a JUnit 4 based `Runner`.

:stuck_out_tongue_winking_eye: 

---

### DIFF
--- a/documentation/src/docs/asciidoc/migration-from-junit4.adoc
+++ b/documentation/src/docs/asciidoc/migration-from-junit4.adoc
@@ -19,7 +19,7 @@ developers have plenty of time to migrate to JUnit Jupiter on their own schedule
 === Running JUnit 4 Tests on the JUnit Platform
 
 Just make sure that the `junit-vintage-engine` artifact is in your test runtime path. In that
-case JUnit 3 and JUnit 4 tests will automatically be picked up by the JUnit Platform launcher.
+case JUnit 3 and JUnit 4 tests will automatically be picked up by the JUnit Platform runner.
 
 See the example projects in the {junit5-samples-repo}[`junit5-samples`] repository to
 find out how this is done with Gradle and Maven.


marcphilipp:
This section is not about the `JUnitPlatform` class that implements `Runner`, though. It's about the junit-vintage-engine...

sbrannen:
Hi @nicolaiparlog,

This is obviously in response to my comment on your [blog on InfoQ](https://www.infoq.com/articles/JUnit-5-Early-Test-Drive). 😉 

@marcphilipp is, however, correct: in the passage you have modified, the `Runner` does not come into play. It is in fact the `Launcher` doing the work in collaboration with the `VintageTestEngine`.

I am therefore closing this issue as _invalid_.

nicolaiparlog:
Damn it, you're right (of course). I came to the docs to make sure I understood the launcher vs runner thingy and stumbled across this line. It never occurred to me that it had the meaning you intended. And then, I guess, Schadenfreude got the best of me. ;)

By the way, we're fixing that post (and the ones on my blog) as we speak. And sorry for the misspelled name. I am usually very exact with that.

sbrannen:
> It never occurred to me that it had the meaning you intended. 

LOL.

We'd definitely never claim that the documentation is without errors. So it could well be that we slipped up, but in this case.... well... that's not the case. 😉 

> And then, I guess, Schadenfreude got the best of me. ;)

Understood. 

> By the way, we're fixing that post (and the ones on my blog) as we speak. And sorry for the misspelled name. I am usually very exact with that.

No worries. There are about 5 different ways to spell my last name, and people have been getting it wrong my entire life. So I'm used to it.

sbrannen:
BTW, do you now understand the difference between the `Launcher`, `TestEngine`, and the `JUnitPlatform` `Runner`?

Are the User Guide and Javadoc adequate in that regard?

nicolaiparlog:
The conceptual difference is well developed and very clear. I keep mixing up launcher and runner, though, because the words are almost interchangeable when it comes to code. And didn't the console launcher used to be the console runner? (I know that the new term makes more sense but that doesn't mean it's not well-suited to screw up my memory.) But I never made the connection from "runner" to the JUnit 4 concept "Runner". With that in mind it should become easier.

> Are the User Guide and Javadoc adequate in that regard?

I have to admit that it's a long time I read the user guide in full. Nowadays I just scan it and look for specific details. I didn't read the section on that split, yet.

I may have some time over the next weeks to give the guide a good read and come back with PRs.

sbrannen:
> I may have some time over the next weeks to give the guide a good read and come back with PRs.

Sounds good!

cheers

PR #441, opened 2016-07-26
Document interoperability with third-party assertion Matchers in User Guide

### LINKED ISSUE #428: Document interoperability with third-party assertion Matchers in User Guide
I've been reading about JUnit 5 for a while and there's one thing I'm wondering about.

The [`Assertions`](http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/Assertions.html) class seems to be the JUnit 5 equivalent of the [JUnit 4 `Assert` class](http://junit.org/junit4/javadoc/latest/org/junit/Assert.html)

I noticed that it doesn't support any replacement of the `assertThat` method which accepts a custom matcher that one can use to make the assertion more simple and readable.

I like the power of expressions that matchers provide, I'm a fan of the `assertThat` syntax and it's something that I'd like to keep using.

I've had a look at the [User Guide](http://junit.org/junit5/docs/current/user-guide) as well as the list of packages in the [Javadoc](http://junit.org/junit5/docs/current/api/overview-summary.html) but I couldn't find a single mention of matchers.

What's the team's view on using matchers? Are they going to be supported or replaced in favour of a different mechanism? Or is it that they are not to be part of the JUnit standard and should be confined to external libraries such as Hamcrest  (Given how [`assertThat` in JUnit 4 `Assert`](http://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertThat%28java.lang.String, T, org.hamcrest.Matcher%29) class introduces a dependency on Hamcrest)

If anything, it would be nice to refer to the subject somewhere in the documentation.

toniedzwiedz:
## Overview

Added a paragraph documenting the use of 3rd party matchers in JUnit Jupiter in section 3.4 of the user guide.

This resolves #428 

---

### DIFF
--- a/documentation/src/docs/asciidoc/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/writing-tests.adoc
@@ -87,6 +87,36 @@ methods in the `{Assertions}` class.
 include::{testDir}/example/AssertionsDemo.java[tags=user_guide]
 ----
 
+Just like in JUnit 4, matchers can be used to make assertions more descriptive and
+readable. Contrary to JUnit 4 and its
+http://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertThat[`org.junit.Assert`]
+class, JUnit Jupiter does not provide a static method accepting a
+http://junit.org/junit4/javadoc/latest/org/hamcrest/Matcher.html[`Matcher`]
+and is, therefore, not dependent on http://hamcrest.org/JavaHamcrest/[Hamcrest].
+
+In order to use the `assertThat` syntax in JUnit Jupiter, you can add a dependency on
+http://hamcrest.org/JavaHamcrest/[Hamcrest], http://joel-costigliola.github.io/assertj/[AssertJ]
+or https://github.com/google/truth[Truth] and statically import the necessary methods supported
+by your chosen framework.
+
+[source,java,indent=0]
+[subs="verbatim"]
+----
+import org.junit.jupiter.api.Test;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+public class MyHamcrestTest() {
+
+    @Test
+    public void simpleTestWithHamcrest() {
+        assertThat(2 + 1, is(3));
+    }
+}
+----
+
+Naturally, legacy tests using the JUnit 4 based Runner can continue using `org.junit.Assert#assertThat`.
+
 [[writing-tests-assumptions]]
 === Assumptions
 


nicolaiparlog:
Instead of leading with "matchers can be used to make assertions more descriptive and readable" it may be more general to go for "any third-party assertion library" and then maybe pick Hamcrest and AssertJ as examples.

documentation/src/docs/asciidoc/writing-tests.adoc
```diff
@@ -87,6 +87,33 @@ methods in the `{Assertions}` class.
...
+In order to use the `assertThat` syntax in JUnit Jupiter, you can add a dependency on http://hamcrest.org/JavaHamcrest/[Hamcrest],
+http://joel-costigliola.github.io/assertj/[AssertJ] or https://github.com/google/truth[Truth]
+and statically import the necessary methods supported by your chosen framework.
+
+[source,java,indent=0]
+[subs="verbatim"]
+----
+import org.junit.jupiter.api.Test;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Is.is;
```
marcphilipp: IMHO this static import should rather use `org.hamcrest.Matchers` or `org.hamcrest.CoreMatchers`.

toniedzwiedz: @marcphilipp either will work. I'll change it to `org.hamcrest.Matchers`

sbrannen:
Please follow the [guidelines for commit messages](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#commit-messages) and squash the commits in your PR into a single commit.

Thanks

documentation/src/docs/asciidoc/writing-tests.adoc
```diff
@@ -87,6 +87,33 @@ methods in the `{Assertions}` class.
...
+public class MyHamcrestTest() {
+
+    @Test
+    public void simpleTestWithHamcrest() {
+        assertThat(2 + 1, is(3));
+    }
+}
+----
+
+Naturally, legacy tests using the JUnit 4 based Runner can continue using `org.junit.Assert#assertThat`
```
sbrannen: This sentence is missing a period (`.`).

toniedzwiedz:
@sbrannen will do. Thanks.

toniedzwiedz:
@sbrannen 

- re-formatted AsciiDoc to fit 90 columns per line
- squashed commits
- changed message

documentation/src/docs/asciidoc/writing-tests.adoc
```diff
@@ -87,6 +87,36 @@ methods in the `{Assertions}` class.
...
+
+[source,java,indent=0]
+[subs="verbatim"]
+----
+import org.junit.jupiter.api.Test;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+public class MyHamcrestTest() {
+
```
sbrannen: This obviously won't compile due to the parentheses declared after the class name, but fret not: I'll fix it after I merge. 😉

sbrannen:
Thanks for making the changes.

toniedzwiedz:
@sbrannen wow, that's embarrassing, guess I've used too many weird languages recently ^^ :)
Thanks

sbrannen:
We actually prefer to use compiled and tested code in the User Guide exactly to avoid such scenarios.

That's why I introduced a new `HamcrestAssertionDemo` class in the `documentation` project in commit 7c6aaa1230218987c3dcbec641b3fdc50cc0836c.

toniedzwiedz:
@sbrannen I did notice that the examples were actually compiled but wasn't sure about introducing new dependencies.

Speaking of which, I think the Mockito example is also inlined in the AsciiDoc. Might be worth looking at.

sbrannen:
> I did notice that the examples were actually compiled but wasn't sure about introducing new dependencies.

There was no need to introduce additional dependencies since the documentation module already contains examples using JUnit 4 which brings in Hamcrest anyway.

> Speaking of which, I think the Mockito example is also inlined in the AsciiDoc. Might be worth looking at.

Yeah, that's a bit unfortunate.

The `MockitoExtension` does not reside in the same repository, and the example for using it therefore cannot be _included_ like the examples local to the `documentation` module. At least, AFAIK, the `include` macro in AsciiDoc only supports the inclusion of files within the local file system.

Furthermore, the `MockitoExtension` is not published -- for example, to Maven Central -- so it's also not possible for the `documentation` module to _consume_ it.

However, once the Mockito team takes over the `MockitoExtension`, it should then be possible to _include_ a locally compiled and tested version of that example.

Cheers,

Sam

PR #446, opened 2016-08-03
Add JavaOne 2016 Session

smoyer64:
Updates to speakers.json and events.json to include our 2016 JavaOne presentation.

---


mmerdes:
@smoyer64 
Thanks for spreading the word

PR #482, opened 2016-08-27
Reduce configuration time impact of versioning plugin

### LINKED ISSUE #480: Versioning plugin slowing down your build
Hey folks,

I just noticed this line in your build:

```
'Build-Revision': versioning.info.commit,
```

This effectively calls `git status` once for *every* subproject on *every* Gradle invocation. At the very least you'll want to call it only once. This will speed up all your builds.

Ideally you could change it so that attribute is only computed on CI builds. On local builds you probably don't care so much about the "Build-Revision" in the manifest.

Cheers,
Stefan

oehme:
This partially fixes #480, only calling `git status´ once per build. That's already a great improvement from 1.6s down to 0.7s for `gradle help`.


sbrannen:

Thanks

Issue #454, opened 2016-08-11
Introduce query language for selecting tags


ondrejlerch:
Let's imagine that our tests are annotated with multiple tags such as fast / slow, priority1 / priority2, stable / unstable, integration / unit, subproject1 / subproject2.

If I am not mistaken:

``` java
tags {
        include 'fast', 'priority1', 'stable', 'integration', 'subproject1'
}
```

will run 'fast' OR 'priority1' OR 'stable' OR 'integration' OR 'subproject1'

I would like to be able to specify:

``` java
tags {
        include 'fast AND priority1 AND stable AND integration AND subproject1'
}
```

or even:

``` java
tags {
        include 'fast AND (priority1 OR priority2) AND NOT unstable AND integration AND (subproject1 OR subproject2)'
}
```

Syntax can be different but you get the idea.

Alternatively, it would be great if there is simple Java API which would help us create such queries. Benefit of Java API would be that queries could be typesafe, autocompleted by IDE and dynamic.

PS: let's make JUnit 5 great!

marcphilipp:
Regarding the Java API: It's already possible to add custom `PostDiscoveryFilters` to `LauncherDiscoveryRequests`. However, there's no way to specify such a filter through Gradle, Maven or the ConsoleLauncher, yet.

While I can see that such a query language might be useful in some cases, I honestly don't want to implement and maintain it.

@junit-team/junit-lambda What do you think?

smoyer64:
I was just sitting down to allow the junit-platform-surefire-provider to specify JUnit 5 tags through the <properties> configuration tag.  Expect a PR in the next couple days.  Should I consider other types of filters at the same time?

ondrejlerch:
Hi Steve,

Filters types that I described (AND, OR, NOT, brackets) are enough for our purposes. I don't know about any other types that you should consider.

More members of my team find "tag queries" feature very useful. We currently miss it in JUnit / TestNG & Maven Surefire Plugin.

To avoid misunderstanding (since I am not expert on junit-platform-surefire-provider) could you show me example how Maven "tag query" would look like if your PR is approved?

Would it be something like:
<plugin>
	<groupId>org.apache.maven.plugins</groupId>
	<artifactId>maven-surefire-plugin</artifactId>
	<configuration>
		<groups>fast AND (priority1 OR priority2) AND NOT stable</groups>
	</configuration>
</plugin>
or something else?

Thanks for your effort!

Ondrej

sbrannen:
[...]

PR #485, opened 2016-08-29
Introduce support for includeTags and excludeTags in the junit-platform-surefire-provider

smoyer64:
As noted in #454, I've been working on a small change to JUnit 5's SurefireProvider that allows the usage of `includeTags` and `excludeTags` within the plugin configuration section of the Maven POM.  An example is as follows:

``` xml
<plugin>
    <artifactId>maven-surefire-plugin</artifactId>
    <version>2.19.1</version>
    <configuration>
        <properties>
            <excludeTags>acceptance, integration</excludeTags>
        </properties>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.junit.platform</groupId>
            <artifactId>junit-platform-surefire-provider</artifactId>
            <version>${junit.platform.version}</version>
        </dependency>
    </dependencies>
</plugin>
```

I'll be submitting a related pull-request to the `junit5-samples` project that updates the POM and README files in addition to adding a couple of test classes that have tags.

---


britter:
I think you should be reusing the `groups` and `excludeGroups` element. See the [Surefire JUnit documentation](http://maven.apache.org/surefire/maven-surefire-plugin/examples/junit.html#Using_JUnit_Categories)

britter:
How about adding some tests to `JUnitPlatformProviderTests` ?

sbrannen:
> I think you should be reusing the `groups` and `excludeGroups` element. See the [Surefire JUnit documentation](http://maven.apache.org/surefire/maven-surefire-plugin/examples/junit.html#Using_JUnit_Categories)

As I stated in https://github.com/junit-team/junit5/issues/454#issuecomment-241394966, the Surefire support for the JUnit Platform should align with the Gradle support for the JUnit Platform -- which intentionally aligns with terminology used in the `ConsoleLauncher` as well as within the Platform itself.

So I don't think using the term _groups_ (even if it existed for JUnit 4 support) is such a good idea. The JUnit Platform provides first-class support for _tags_. So we shouldn't confuse users by calling them something different _only_ when using Maven Surefire.

Furthermore, the JUnit Platform may potentially support a feature called _groups_ at a later date. That's not foreseeable at the moment; however, we can not preclude the possibility.

smoyer64:
> As I stated in #454 (comment)

I realize that this PR doesn't satisfy the need for includeEngines and excludeEngines ... and it definitely needs tests.  Things have been crazy-busy at work so I wanted to at least submit the part that was done (and working in a local project).

britter:
@sbrannen:

> So I don't think using the term groups (even if it existed for JUnit 4 support) is such a good idea. The JUnit Platform provides first-class support for tags. So we shouldn't confuse users by calling them something different only when using Maven Surefire.

Problem with this is, that Maven uses `groups` and `excludedGroups` not only for JUnit, but also for TestNG and POJO (see http://maven.apache.org/surefire/maven-surefire-plugin/featurematrix.html). So from their point of view, it would be using the term _groups_ only when using Surefire with JUnit 5.

sbrannen:
@britter, I thought somebody might come back with that argument. 😉 

I can understand that the Maven Surefire team also wishes to maintain a level of consistency with regard to the names of features they support. So if the Maven Surefire team opts in favor of _groups_ over _tags_ for their JUnit Platform support, I will leave that up to them. 

And if they do choose _groups_ over _tags_, I agree that we should name the tag support correctly from the start in order to avoid inconveniencing early adopters of the provider.

sbrannen:
> So from their point of view, it would be using the term groups only when using Surefire with JUnit 5.

It's worth mentioning that referring to JUnit 5 (and potentially implying JUnit Jupiter with that) can quickly lead to confusion. So, for the sake of clarity, let's refer to either the JUnit Platform or JUnit Jupiter accordingly.

Cheers!

Tibor17:
groups and excludeGroups have the same meaning.
Imaging how confusing it would be to have groups and tags in plugin XSD with the same purpose.

Tibor17:
I mean if the documentation is written properly with example how the plugin is configured or an recommendation, then JUnit 5 may use tags in JUnit 5 API and Surefire project keep backwards compatibility. WDYT?

sbrannen:
Well, there will be confusion for certain sets of users no matter which route we go.

So I guess it's really just a choice of the lesser of two evils. 😉 

Plus, proper documentation will be necessary anyway.

smoyer64:
Is there a consensus on includeTags/excludeTags?  I guess it's also worth remembering we're eventually going to need includeEngines/excludeEngines configuration properties which won't align with any of the existing maven-surefire-plugin parameters.

marcphilipp:
I think we should support groups/excludeGroups as a synonym for includeTags/excludeTags for compatibility with Maven Surefire.

mmerdes:
@marcphilipp 
agreed

sbrannen:
> I think we should support groups/excludeGroups as a synonym for includeTags/excludeTags for compatibility with Maven Surefire.
>
> Edit: Only one of them should be present. Otherwise we should signal a validation error/throw an exception.

I fully agree!

We (and by "we" I mean the entire community) have to keep in mind that the JUnit Platform is **much more powerful** than JUnit 4 in terms of configurability. And that means there will be multiple configuration options that will have to be supported by build tools and IDEs.

So, while backwards compatibility is for sure "a good thing" (especially for people migrating from JUnit 4 to the JUnit Platform), we simultaneously have to keep an eye on the future.

In other words, the design decisions we make now must be _forward thinking_ in order to avoid having a mess of inconsistent configuration options/styles for the JUnit Platform as it evolves over time.

marcphilipp:
@smoyer64 Can you do the necessary changes to support groups/excludeGroups as synonyms to includeTags/excludeTags?

smoyer64:
@marcphilipp Yes, I can take care of this though it may be a couple days until I get to it.  I'd also like to (finally) dig into how platform tests work and I'll add those if possible.  I know we also want includeEngines/excludeEngines - should I delay this PR for that or should I submit a second one (probably at some point after JavaOne)?

marcphilipp:
Sure, no worries. I think includeEngines/excludeEngines should be done in a separate pull request. Thanks!

smoyer64:
One of these days I'll remember to apply Spotless **before** I commit ... I guess I'm still learning the Gradle/JUnit 5 workflow.

The latest commit allows one of includeTags or groups, and one of excludeTags or excludedGroups per the discussion above.  It does not check to make sure the same tag is implicitly or explicitly included and excluded at the same time.

Unit tests were added to cover the logic required to determine which filters should be created and the code to actually create them.

One other change that I made while in the code was to wrap the PreconditionViolationException that might be thrown when LauncherFactory.create() is called with a MojoFailureException.  I believe this will provide a nice message when the plugin is executed without a TestEngine, but will also expand the cause when the -X switch is included.  I couldn't figure out how to test the (simple) code that implements this wrapper, so I'd recommend inspection of lines 133 and 134 in JUnitPlatformProvider (or advice on how to create an appropriate test).

Since this provider is an surrogate of a Mojo and derives its configuration from the parent plugin, I also didn't write any tests that actually execute the maven-surefire-plugin - I've done that in the past and could write tests at that level if needed.  Since the filters are completely specified as properties in the ProviderParameters, I don't feel it's required.

Sorry it took so long for me to return to this task - JavaOne and these weeks afterwards have been busy.

smoyer64:
Please hold off on this merge - I'm seeing an intermittent failure resolve MojoFailureException.

smoyer64:
Whoever let the LauncherFactory.create() method throw a PreconditionViolationException was correct.  When the maven-surefire-plugin forks execution, MojoFailureException and MojoErrorException are not necessarily on the classpath.

I've reverted the change I made to the JUnitPlatformProvider's constructor and changed the new code to also throw a PreconditionViolationException.

I believe this is now ready to go!

marcphilipp:
Before merging this pull request we first need to complete #541.

smoyer64:
Will this make it into M3?  I have been waiting on this merge to complete the above referenced changes to the example project.

marcphilipp:
I don't see why not. I'll try to get #541 done today.

marcphilipp:
@smoyer64 #541 is done which means we've changed the license of junit-platform-surefire-provider and the header of all source files. GitHub does not show any conflicts but it might be good idea to rebase anyway. In addition, please explicitly agree to the changed Contributor License Agreement. Thanks!

smoyer64:
I'll be in the office in about an hour and will get this done immediately.

I do explicitly agree to providing this changed code under the ASL 2.0 license (I'm a committer on several Apache projects and have an iCLA on file).

smoyer64:
@marcphilipp 

The new License headers have been merged into the four files this PR changes.

marcphilipp:
@junit-team/junit-lambda I think this is ready to be merged. Do you want to take a look?

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
```diff
@@ -32,17 +36,33 @@
...
+	static final String EXCLUDE_GROUPS = "excludedGroups";
+	static final String EXCLUDE_TAGS = "excludeTags";
+	static final String INCLUDE_GROUPS = "groups";
+	static final String INCLUDE_TAGS = "includeTags";
+
+	static final String EXCEPTION_MESSAGE_BOTH_NOT_ALLOWED = "The " + INCLUDE_GROUPS + " and " + INCLUDE_TAGS
+			+ " parameters (or the " + EXCLUDE_GROUPS + " and " + EXCLUDE_TAGS + " pararameters) are synonyms - "
+			+ "only one of each is allowed (though neither is required).";
+
+	static final String EXCEPTION_MESSAGE_NO_ENGINE = "JUnit 5 Precondition Violation - No engines were specified.";
```
sbrannen: Please change this to "JUnit Platform" instead of "JUnit 5".

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
```diff
@@ -103,10 +124,55 @@ private void invokeSingleClass(Class<?> testClass, RunListener runListener) {
...
+			filters.add(TagFilter.includeTags(includes.get()));
+		}
+
+		Optional<List<String>> excludes = getGroupOrTags(getPropertiesList(EXCLUDE_GROUPS),
+			getPropertiesList(EXCLUDE_TAGS));
+		if (excludes.isPresent()) {
+			filters.add(TagFilter.excludeTags(excludes.get()));
+		}
+
+		return filters.toArray(new Filter<?>[0]);
```
sbrannen: Replace `0` with `filters.size()`.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
```diff
@@ -103,10 +124,55 @@ private void invokeSingleClass(Class<?> testClass, RunListener runListener) {
...
+	private Optional<List<String>> getPropertiesList(String key) {
+		List<String> compoundProperties = null;
+		String property = parameters.getProviderProperties().get(key);
+		if (property != null) {
+			compoundProperties = Arrays.asList(property.split("[, ]+"));
+		}
+		return Optional.ofNullable(compoundProperties);
+	}
+
+	private Optional<List<String>> getGroupOrTags(Optional<List<String>> groups, Optional<List<String>> tags) {
```
sbrannen: Rename to `getGroupsOrTags`.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
```diff
@@ -103,10 +124,55 @@ private void invokeSingleClass(Class<?> testClass, RunListener runListener) {
...
+		Optional<List<String>> elements = Optional.empty();
+
+		if (groups.isPresent() && tags.isPresent()) {
```
sbrannen: Use `Preconditions.condition(...)`.

marcphilipp:
Good catches, @sbrannen!

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
```diff
@@ -103,10 +125,53 @@ private void invokeSingleClass(Class<?> testClass, RunListener runListener) {
...
+		if (property != null) {
+			compoundProperties = Arrays.asList(property.split("[, ]+"));
+		}
+		return Optional.ofNullable(compoundProperties);
+	}
+
+	private Optional<List<String>> getGroupsOrTags(Optional<List<String>> groups, Optional<List<String>> tags) {
+		Optional<List<String>> elements = Optional.empty();
+
+		Preconditions.condition(groups.isPresent() && tags.isPresent(), EXCEPTION_MESSAGE_BOTH_NOT_ALLOWED);
```
marcphilipp: I guess the tests will pick this up, but this line should be

```java
Preconditions.condition(!groups.isPresent() || !tags.isPresent(), EXCEPTION_MESSAGE_BOTH_NOT_ALLOWED);
```

smoyer64:
I like the suggestions provided the code review with one exception - I'm not so sure how I feel about using the size of an array to provide the "prototype" for the type of array that the list should be converted to.

I've always used a zero size array during that conversion which seems to be in alignment with the Sun documentation: http://docs.oracle.com/javase/7/docs/api/java/util/List.html#toArray(T[]).

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
marcphilipp: Yes, `Preconditions.condition()` throw an exception if the first argument is `false`. `!groups.isPresent() || !tags.isPresent()` is only `false` when both `groups` and `tags` are present.

marcphilipp:
> I've always used a zero size array during that conversion which seems to be in alignment with the Sun documentation

Well, the Sun documentation says

> If the list fits in the specified array, it is returned therein. Otherwise, a new array is allocated with the runtime type of the specified array and the size of this list.

Thus, given a non-empty list, if you use zero, you will effectively allocate two arrays whereas when you pass an array with a length that equals the list's size you will only allocate one. However, the JVM or the JIT compiler may optimize that at runtime.

smoyer64:
I'm going to have to time that both ways ... but your argument sounds logical.

smoyer64:
No problem ... and sorry to be so high-maintenance today!

nicolaiparlog:
Since your discussing the performance aspect of this, you should check to [Aleksey Shipilev's _Arrays of Wisdom of the Ancients_](https://shipilev.net/blog/2016/arrays-wisdom-ancients/), which ends with:

> **Bottom line**: `toArray(new T[0])` seems faster, safer, and contractually cleaner, and therefore should be the default choice now. Future VM optimizations may close this performance gap for `toArray(new T[size])`, rendering the current "believed to be optimal" usages on par with an actually optimal one. Further improvements in toArray APIs would follow the same logic as `toArray(new T[0])` — the collection itself should create the appropriate storage.

But unless this is a hot code path (didn't check), the performance discussion is moot anyway. I'd do what's more readable and `0` makes it much shorter so I'd use that.

smoyer64:
@nicolaiparlog 

Thanks for the great link!

This particular code is only executed once when the Maven Surefire provider is reading its configuration, so I don't think it's an issue either way.  If the rest of the JUnit 5 project explicitly sets the size of the "template" array, I'd say consistency is a great reason to do this one the same way.

I'm going to pass the article you referenced around the shop though.

sbrannen:
Thanks for the link, @nicolaiparlog! Very interesting article.

As @smoyer64 mentioned, the performance hit in this particular use case is a moot point: neither variant will make a noticeable difference.

So let's leave it "as is" for the sake of consistency at the moment, and we can reassess our use of `toArray(...)` at a later date if it becomes pertinent.

sbrannen:
I think this PR looks good to go now!

_But..._ we may actually want to align the use of filters and selectors in the Surefire provider with the style used in the Launcher's request builder API and the Gradle plugin. See #554 for details.

@junit-team/junit-lambda what do you think? Shall we merge this PR _as is_ and address the configuration style consistency in a separate issue (potentially slotted for M4)?

marcphilipp:
I think we should merge this pull request _as is_ and open a new ticket for the other configuration changes.

marcphilipp:
I've created the follow-up issue #558 and will go ahead and merge this pull request.

marcphilipp:
Thanks, @smoyer64! 👍

sbrannen:
Thanks, @smoyer64!

Thanks, @marcphilipp.

smoyer64:
You're very welcome ... thanks for all the work you're putting into keeping the project moving forward in a consistent manner.

Issue #496, opened 2016-09-06
Provide consistent Precondition behavior for arrays and collections


britter:
After reading @sbrannen comment in #493 I had a look at the `Preconditions` class. Currently the implementation of `notNull(Object[])` checks that the array reference is not null and that the array does not contain any null values. The counterpart for Collections is `notEmpty(Collection)`, which checks that the given collection is not null and that it is not empty.

Looking at for example the`TagsFilter` class and how it checks its input parameters:

```java
public static PostDiscoveryFilter excludeTags(String... tags) {
	Preconditions.notNull(tags, "tags must not be null");
	return excludeTags(asList(tags));
}

public static PostDiscoveryFilter excludeTags(List<String> tags) {
	Preconditions.notEmpty(tags, "tags must not be null or empty");
	return descriptor -> FilterResult.includedIf(trimmedTagsOf(descriptor).noneMatch(tags::contains));
}
```

We can conclude that:

Method call | Result
--------------------|------------
 `excludeTags(null)` | Exception 
 `excludeTags((String[]) null)` | Exception 
 `excludeTags(new String[0]())` | okay 
 `excludeTags(Collections.singletonList(null))` | okay 
 `excludeTags((List<String>) null)` | Exception 
 `excludeTags(Collections.emptyList())` | Exceptions

My proposal here would be to introduce new methods:

* `Preconditions.noneNull(Collection)`
* `Preconditions.noneNull(Object[])`

These would check that the reference is not null, that the collection/array does not contain any null reference and that the collection/array is not empty. The last check does not really match the method name but I can't think of a better name right now...

britter:
Actually `excludeTags(String[])` with an empty array will throw an exception because it delegates to `excludeTags(Collection)`.

sbrannen:
Thanks for investigating and reporting back on your findings.

I'm slating this for M3 for the time being.

britter:
I'd like to supply a patch, but I'm currently not sure what would be the right approach for this. Here are some options I see:

* make `notNull(Object[], String)` also check that the array is not null
* introduce `notEmpty(Object[], String)` with the same behavior as `notEmpty(Collection, String)`
* make `notEmpty(Collection, String)` also check that the collection does not contain null values
* introduce `notNull(Collection, String)` with the same behavior as `notNull(Object[], String)`
* introduce new methods for example `noneNull(Object[], String)`, `noneNull(Collection, String)` which check for not null, not empty and no null references contained

WDYT?

[...]

PR #507, opened 2016-09-10
Make behavior of Precondition.notEmpty(Collection) consistent with

britter:
## Overview

- As discussed in #496

---

Precondition.notNull(Object[]).

Null elements in the collection will no longer be accepted. This
is part of #496.


marcphilipp:
Thanks! 👍

PR #522, opened 2016-09-25
Fixed typo in TestingAStackDemo

npathai:
## Overview

- AfterPushing context contained a typo, `isEmpty()` is changed to `isNotEmpty()`. After pushing an element in a stack it is no longer empty. The display name is correct but method name was not reflecting the intention. 

---

There was a typo in `AfterPushing` context, `isEmpty` is changed to `isNotEmpty`


marcphilipp:
Good catch, thanks!

npathai:
@marcphilipp it's the least I could do for u all. I will also start contributing with some low hanging ones.

marcphilipp:
👍

PR #525, opened 2016-09-28
Add overload of 'ExtensionContext::publishReportEntry'

nicolaiparlog:
This overload, just like the one on `TestReporter`, makes it easier to publish a single key-value pair.

There is also a small fix of a collection emptiness check in there.

---


marcphilipp:
Thanks! 👍

Issue #433, opened 2016-07-25
Introduce adapters for JUnit 4 Rules


sbrannen:
## Goals

Introduce _adapters_ for common JUnit 4 Rules in the `junit-jupiter-engine`.

The dependency from `junit-jupiter-engine` to JUnit 4 must be an _optional_ dependency.

This support should be automatically enabled if `junit-4.x.jar` is on the classpath.
## Proposed Candidates
- [x] `ExpectedException`
- [x] `ExternalResource`
- [x] `Verifier`
## Related Issues
- #169 
- #343 
## Related Blog Posts
- http://www.codeaffine.com/2016/04/06/replace-rules-in-junit5/
- http://blog.codefx.org/libraries/junit-5-conditions/

kcooney:
@sbrannen Is the proposal to allow these three Rules in JUnit Jupiter tests but not others?

sbrannen:
@kcooney, yes, that's correct.

The JUnit 5 team analyzed the existing rules in JUnit 4 and selected ones that can be supported with current extension APIs in JUnit Jupiter.

JUnit 4 rules not in the above list cannot currently be supported without some sort of `Statement`-like extension which is currently not planned for JUnit Jupiter.

mlevison:
Suggestion (from vacation again, excuse top posting).

Adapters will promote the continued use of the old rules.

Consider as an alternative providing mappings. For bonus points, automate
the mapping via regex (of simple enough) or a full blown test rewriter (ie
Java code).

NB I'm writing from my phone at a Gaelic football tournament. I've not
reflected on real code.

Cheers
Mark

kcooney:
Two points:

1. There is no need to support `ExpectedException` for JUnit Jupiter tests. The `expectThrows` and `assertThrows` methods are more powerful. In addition, `ExpectedException` is easy to use incorrectly so we shouldn't encourage it's use for projects that can use lambdas.

2. This doesn't solve the need for a Rule-like extension point that can be used by JUnit Jupiter tests and JUnit Vintage tests, so deveopers with a large code base can build simple reusable extension points that they can use in both styles of tests while they slowly migrate their tests to JUnit Jupiter. Did the JUnit Lambda team discuss my prototype for such an extension?

marcphilipp:
[...]

Issue #497, opened 2016-09-08
Support programmatic extension registration via fields


sbrannen:
## Status Quo

It is currently possible to register extensions declaratively via `@ExtendWith` on a class or method but not on a field.

Furthermore, it is currently impossible for a user of an extension to provide input to the extension programmatically. Rather, in order to make an extension configurable, current extension implementations typically provide a custom annotation, but such an annotation is hindered by the limitations of what can be declared as an annotation attribute in Java.

See the discussion in [Spring REST Docs](https://github.com/spring-projects/spring-restdocs/issues/296#issuecomment-244209553) for further background information.
## Related Issues
- #416

## Proposal for Programmatic Extension Registration

In general, it should be possible for an extension to be registered _programmatically_ -- for example, via a field or a factory method. This would allow a user to pass arguments to the extension's constructor (or `static` factory method) or to construct the extension programmatically, for example, via a method that returns the configured extension instance.

With regard to migrations from JUnit 4 to JUnit Jupiter, it should be possible to migrate a JUnit 4 rule from this:

``` java
@Rule
public JUnitRestDocumentation restDocumentation =
        new JUnitRestDocumentation("target/generated-snippets");
```

... to the following in JUnit Jupiter:

``` java
@RegisterExtension
RestDocumentationExtension restDocumentation =
        new RestDocumentationExtension("target/generated-snippets");
```

## Considerations

- Although the existing `@ExtendWith` annotation _could_ technically be repurposed for such use cases, the JUnit Team feels it would be better to introduce a new annotation dedicated to this purpose. The current proposed name is `@RegisterExtension`, but this may change.

## Deliverables

- [x] Support programmatic registration of a single extension via a field annotated with `@RegisterExtension` where the field's value is used as the extension instance.
- [x] Document in User Guide.
- [x] Document in Release Notes.

juergenzimmermann:
@sbrannen do you have a coarse estimation when M3 might be available?

sbrannen:
@juergenzimmermann, perhaps by the end of October. The core team is swamped with work, so we are unfortunately forced to push back the release dates.

SqAutoTestTeam:
We are migrating project from Juni4 to Junit5. 
Some of functionality written in Juni4 is based on usage of Rule.  The custom rule is used to 
send tests results to outer system - where they can be seen later . 

Rule provides me with three  callback methods 
   protected void succeeded(Description desc);
   protected void failed(Throwable e, Description desc);
[...]

Issue #169, opened 2016-03-02
Demonstrate migration path from JUnit 4 to JUnit 5


sbrannen:
## Overview

Based on discussions I've had with developers in the community, it is apparent that some people fear an upgrade from JUnit 4 to JUnit 5. Aside from the fact that it's totally feasible to run JUnit 4 tests alongside JUnit 5 tests, we should still attempt to migrate an existing JUnit 4 test suite just to see where the hurdles are.
## Proposal

Demonstrate how to migrate from JUnit 4's `ExpectedException` Rule to a JUnit 5 Extension-based alternative.
## Food for Thought

Can this be achieved with simple search-and-replace?

Can a JUnit 5 extension be implemented that honors JUnit 4's `ExpectedException` transparently, without the need to modify a single line within test methods -- for example, by removing `@Rule` from the `ExpectedException` field and adding `@ExtendWith(SomeExpectedExceptionExtension.class)`?
## Related Issues
- #343
## Related Blogs
- [How to Replace Rules in JUnit 5](http://www.codeaffine.com/2016/04/06/replace-rules-in-junit5/)

philwebb:
The other one I use quite often is `TemporaryFolder`

sormuras:
@philwebb Me, too! See https://github.com/junit-team/junit5-samples/issues/4

kcooney:
The lack of Rule support in JUnit5 might make it hard for large projects to incrementally migrate test classes, especially if they have custom Rules.

I wonder if we could have an optional jar that defines a simple extension point that can be used for JUnit5 and JUnit4. I find that most of our Rules only need the following:

* setUp()
* tearDown() - called only if setUp () called
* verify() - called only if the test method doesn't throw

If this interests anyone, I could write up a proposal.

akozlova:
Hi Kevin,

I thought about IDE-side migration tool to ease the transition. Of cause it
doesn't make the jar useless.

BTW If you have 'rules' for such tool in mind, it would help me a lot.

Thanks,
Anna

marcphilipp:
@kcooney: Can you please elaborate your proposal a bit more?

sbrannen:
Related blog post:

http://www.codeaffine.com/2016/04/06/replace-rules-in-junit5/

kcooney:
Thanks for the pointer

[...]

PR #526, opened 2016-09-28
Issue433 adapters for junit4 rules

mmerdes:
## Overview

This PR provides an initial implementation for #433.

Both, field and method rules are supported.
Implementations of `org.junit.rules.Verifier` and `org.junit.rules.ExternalResource` are supported. Currently, this supported is activated by an explicit `@ExtendWith`.
This might be replaced by a classpath-based activation (expensive) or better by some sort of meta-annotation, e.g. `@EnableJUnit4RuleSupport`.
Currently, there is no warning issued if an unsupported rule type is used.

---


bechte:
First of all, thanks @mmerdes for the effort. Do I understand your code right, that it will require a compile time dependency to the JUnit4 core to support the JUnit4 `@Rule` annotation and the proposed rules as of #433 (+comments)? If we go along the route, this should be at least only a runtime dependency. .o(This would only be a minor thing to tackle)

Reading the tests for your implementation is giving me a good look & feel, how this thing will be used in the future. It comes along with a advantages for test migration, i.e. the basic rules would require no change to the code. Still I feel, that we open the door for expectation far to wide. Allowing the basic rules will grow expectations for the other rules to follow. We might end up loosing the control of the test execution the same way as JUnit4 with `@Rule`s.

As @kcooney already mentioned on https://github.com/junit-team/junit5/issues/169#issuecomment-228498388, most rules only require to participate one or many of the three phases "setup", "teardown", and "afterEach". Our extension model allows exactly this (and that's the way to implemented your adapters). The only thing that is missing is, to allow people to register an extension at the field level. If we had `@ExtendWith` for fields, one could simple replace the `@Rule` annotation from JUnit4 with `@ExtendWith` annotation from JUnit5 during the migration. The tests have to been transformed anyway to reference the JUnit5 version of the `@Test` annotation, therefore, it wouldn't be much hassle to also replace the `@Rule`.

Now, as we already discussed, `@ExtendWith` on the field level doesn't read well and it somehow behaves different (it is not registering an extension for its own level and all tests below the test hierarchy, but it adds another extension to the class level). Therefore, I would vote for using a different annotation that communicates its intend better, and provides the functionality needed, i.e. registering extensions on the field level: `@UseExtension` or `@WithExtension`

The annotation will have no values as the Type of the field should resolve to an `Extension` class. With this mechanism in place a migration of the rules included in #433 (+comments) would be possible like this:

```
// code before migration
@Rule
public TemporaryFolder folder = new TemporaryFolder();

@Rule
public ExternalResource resource1 = new ExternalResource() {
   // ...
}

// code after migration
@UseExtension
public TemporaryFolder folder = new TemporaryFolder();

@UseExtension
public ExternalResource resource1 = new ExternalResource() {
   // ...
}
```

Note these 3 things on the code:
- `@Rule` will be replaced by `@UseExtension` (or `@WithExtension`, you-name-it)
- TemporaryFolder will not be resolved to the JUnit4 Rule `org.junit.rules.TemporaryFolder` but to a new class that we have to provide which basically does the same thing.
- ExternalResource will not be resolved to the JUnit4 Rule `org.junit.rules.ExternalResource` but to a new class that we have to provide which basically does the same thing.

Still, very simple and the migration could be done by a script, searching & replacing the parts. This way, I would have a much better feeling. If the wants me to provide this as an alternative approach, I would be happy to do so. But please let us take very careful steps with this topic. I have a really bad feeling about opening the gates to modify the test execution lifecycle.

mmerdes:
@bechte 
Thanks for your extended review. 
Yes, a runtime dependency to JUnit 4 is of course required. As stated in the description of this PR the exact mechanics are not yet fixed. I outlined tow options and stated my preference. So far so good. 
Concerning the main part of your comment:
If I understand you correctly, you are not comfortable with the _requirements_ of this issue. However, these requirements have been fixed long ago in the respective issue and specifically included the verbatim use of the JUnit 4 `@Rule` syntax without the need for a script transformation etc. 
As an additional approach field-based extensions are discussed here in #497.
I think we all agree that multiple migration options might be suitable for different companies/situations.

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/adapter/AbstractTestRuleAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+	public AbstractTestRuleAdapter(RuleAnnotatedMember annotatedMember, Class<? extends TestRule> adapteeClass) {
+		this.target = annotatedMember.getTestRuleInstance();
+		this.adapteeClass = adapteeClass;
+
+		this.failIfAdapteeClassIsNotAssignableFromTargetClass();
+	}
+
+	private void failIfAdapteeClassIsNotAssignableFromTargetClass() {
+		if (!this.adapteeClass.isAssignableFrom(this.target.getClass()))
+			throw new IllegalStateException(this.adapteeClass + " is not assignable from " + this.target.getClass());
```
marcphilipp: How about using `Preconditions`?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/adapter/AbstractTestRuleAdapter.java
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	protected void executeMethod(String name) {
+		try {
+			Method method = target.getClass().getDeclaredMethod(name);
+			method.setAccessible(true);
+			ReflectionUtils.invokeMethod(method, target);
+		}
+		catch (NoSuchMethodException | SecurityException exception) {
+			LOG.warning(exception.getMessage());
```
marcphilipp: I think this should throw an exception instead of logging it. Otherwise, it would only be visible in a log that something was silently ignored.

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ExternalResourceSupport.java
```diff
@@ -0,0 +1,70 @@
...
+
+	private final AbstractTestRuleSupport fieldSupport = new TestRuleFieldSupport(this.adapterGenerator,
+		ExternalResource.class);
+	private final AbstractTestRuleSupport methodSupport = new TestRuleMethodSupport(this.adapterGenerator,
+		ExternalResource.class);
+
+	@Override
+	public void beforeEach(TestExtensionContext context) throws Exception {
+		this.fieldSupport.beforeEach(context);
+		this.methodSupport.beforeEach(context);
```
marcphilipp: This should be methods first, then fields to maximize compatibility with Vintage.

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/TestRuleFieldSupport.java
```diff
@@ -0,0 +1,62 @@
...
+		return new RuleAnnotatedField(context.getTestInstance(), (Field) member);
+	}
+
+	@Override
+	protected List<Member> findRuleAnnotatedMembers(Object testInstance) {
+		return findAnnotatedFields(testInstance, super.getRuleType(), super.getAnnotationType()).collect(
+			Collectors.toList());
+	}
+
+	// TODO: decide whether this should be promoted to AnnotationUtils
```
marcphilipp: I think it should but not necessarily in this pull request. Can you please create an issue for it and delete this comment?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/member/RuleAnnotatedField.java
```diff
@@ -0,0 +1,36 @@
...
+
+	private static final Logger LOG = Logger.getLogger(RuleAnnotatedField.class.getName());
+
+	public RuleAnnotatedField(Object testInstance, Field testRuleField) {
+		try {
+			testRuleField.setAccessible(true);
+			this.testRuleInstance = (TestRule) testRuleField.get(testInstance);
+		}
+		catch (IllegalAccessException exception) {
+			LOG.warning(exception.getMessage());
```
marcphilipp: Same here.

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/member/RuleAnnotatedMethod.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.junit.platform.commons.meta.API;
+import org.junit.platform.commons.util.ReflectionUtils;
+import org.junit.rules.TestRule;
+
+@API(Internal)
+public class RuleAnnotatedMethod extends AbstractRuleAnnotatedMember {
+
+	public RuleAnnotatedMethod(Object testInstance, Method member) {
+		//no args
```
marcphilipp: ?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * Extensions which provide (limited) support for JUnit 4 rules within JUnit Jupiter.
```
marcphilipp: 4 = Vintage?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/adapter/AbstractTestRuleAdapter.java
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	protected Object executeMethod(String name, Class<?>[] parameterTypes, Object... arguments) {
+		try {
+			Method method = target.getClass().getDeclaredMethod(name, parameterTypes);
+			method.setAccessible(true);
+			return ReflectionUtils.invokeMethod(method, target, arguments);
+		}
+		catch (NoSuchMethodException | SecurityException exception) {
+			LOG.warning(exception.getMessage());
```
marcphilipp: I think this should throw an exception instead of logging it. Otherwise, it would only be visible in a log that something was silently ignored.

marcphilipp:
_in progress_

marcphilipp:
This pull request has been manually rebased onto master and merged.

@mmerdes Thanks for all your hard work! 👍

sbrannen:
+1: thanks, @mmerdes! 👍 

I'm looking forward to hearing what the community has to say about this migration support.

Issue #516, opened 2016-09-16
Cannot select default method by fully qualified method name


akozlova:
Given the code:

```java
interface ATest {
   @Test
    default void testSmthCommon() {}
}

class BTest implements ATest {}
```
I expect that passing `BTest#testSmthCommon` as the _fully qualified method name_ would discover test method in `BTest` class. 

Instead, the JUnit Platform discovers method in the interface and then the `JupiterTestEngine` fails to run it.

I believe the problem is due to the use of `method.getDeclaringClass()` in `DiscoverySelectors#selectJavaMethod(String)`.

This was noticed in the latest `SNAPSHOT` version.

sbrannen:
Thanks for raising the issue @akozlova!

I have confirmed that this is a bug.

sbrannen:
I have also reworded the title accordingly.

sbrannen:
FYI: I have introduced a failing test that reproduces this issue. See c8a589947f8073a00a35044dbb2c1138480a1f58.

sbrannen:
Tentatively assigned to @bechte in conjunction with his work on #489.

bechte:
This issue makes the difference of `JavaMethodSelector` and `JavaMethodSource` very obvious. A selector references to the _Thing Under Test_ (also requiring the explicit `Class` to run the tests with), while the source points to the place _where_ the code is defined. Therefore, for the selector we have to extract the class from the fully qualified name and cannot use the `getDeclaringClass()`. For the source, we really want to use `getDeclaringClass()` for its purpose.

I changed this accordingly in both classes (selector and source), see commit #a7c5808. This will be merged with #489 as it is part of the overall rework on this topic.

bechte:
Merged into master.

Issue #489, opened 2016-08-31
Store raw values in DiscoverySelectors & TestSources


sbrannen:
## Overview

In light of issues such as #477 and the discussion in #466,  the JUnit 5 Team has decided to store _raw_ values in all `DiscoverySelector` and `TestSource` implementations for the _entire Platform_.

In other words, the JUnit Platform should not _store_ anything in either selectors or sources using a _biased_ format (e.g., `File` vs. `Path` vs. `URI` vs. `String` _and_ `Method` vs. method name), since it is up to a specific `TestEngine` or IDE to interpret the _raw_ format.

## Proposal for file system sources

Input from @marcphilipp:

> We could change `FileSource` and `DirectorySource` to store a `URI` instead of a `File` and provide methods that return a `File`/`Path` from the `URI`.

## Deliverables

- [x] Ensure that each `DiscoverySelector` implementation stores the selected element in a _raw_ format.
  - Add _getter_ methods for retrieving the raw format as well as technology specific representations.
- [x] Ensure that each `TestSource` implementation stores the source in a _raw_ format.
  - Add _getter_ methods for retrieving the raw format as well as technology specific representations.

jbduncan:
Sounds pretty reasonable to me!

bechte:
First step: Ensure that each DiscoverySelector implementation stores the selected element in a raw format

- [ ] ClasspathResourceSelector
- [ ] ClasspathRootSelector
- [ ] DirectorySelector
- [ ] FileSelector
- [ ] JavaClassSelector
- [ ] JavaMethodSelector
- [ ] JavaPackageSelector
- [ ] UniqueIdSelector
- [ ] UriSelector

sbrannen:
@bechte, could you please take a look at #516 and keep that bug in mind as you address #489?

Cheers!

bechte:
Sure, thanks for pointing that. 👍

bechte:
Second Step: Ensure that each `TestSource` implementation stores the source in a raw format.
Add getter methods for retrieving the raw format as well as technology specific representations.

- [ ] JavaClassSource
- [ ] JavaMethodSource
- [ ] JavaPackageSource

bechte:
I did some rework on the classes to provide a good base for further discussion, see feature branch *sb_userawvaluesfordiscovery*. There are still some questions open that we will discuss in the team. Afterwards, I will prepare a pull request for the changes.

bechte:
Merged into master.

Issue #477, opened 2016-08-24
Non-standard method names are not supported for discovery selection


sbrannen:
## Status Quo

All `selectMethod()` variants in `DiscoverySelectors` attempt to eagerly load the selected method with the assumption that the supplied method name follows Java's standard convention for methods.

This is totally fine for standard Java methods; however, for testing frameworks such as Spock the method name might contain special characters.

For example, `"applying the plugin"` is a valid method name in a Spock specification (at least within the Groovy source code). In the decompiled byte code, however, we see that the actual Java method name is something completely different, such as `$spock_feature_0_0` in the following example.

```java
// Method descriptor #26 ()V
// Stack: 18, Locals: 3
@org.spockframework.runtime.model.FeatureMetadata(line=(int) 32,
  name="applying the plugin",
  ordinal=(int) 0,
  blocks={@org.spockframework.runtime.model.BlockMetadata(kind=org.spockframework.runtime.model.BlockKind.WHEN,
          texts={}),@org.spockframework.runtime.model.BlockMetadata(kind=org.spockframework.runtime.model.BlockKind.THEN,
          texts={})},
  parameterNames={})
public void $spock_feature_0_0();
```

For this reason, attempting to _select_ the `"applying the plugin"` method by name (e.g., via `selectMethod("org.junit.platform.gradle.plugin.JUnitPlatformPluginSpec#applying the plugin")`) results in an exception similar to the following.

```
org.junit.platform.commons.util.PreconditionViolationException: 
    'org.junit.platform.gradle.plugin.JUnitPlatformPluginSpec#applying the plugin'
    could not be resolved to a unique method
```

## Related Resources

- [IntelliJ IDEA](https://youtrack.jetbrains.com/issue/IDEA-159975) issue on the same topic.

## Deliverables

- [x] Introduce a `selectMethod()` variant that does not attempt to eagerly load the selected method.
  - The `MethodSelector` could then provide the input method name as a `String` and let the `TestEngine` decide how to load or reference the method.
- [x] Consider if all existing method selectors should be revised to avoid eager loading of methods using reflection.
- [x] Consider if all existing class selectors should be revised to avoid eager loading of classes using reflection.

sbrannen:
@junit-team/junit-lambda, we will need to discuss the ramifications of such use cases and come to a decision on how to proceed _across the board_ for all such selectors.

bechte:
[...]

PR #527, opened 2016-09-29
Pull request for Raw Value in DiscoverySelectors

bechte:
## Overview

This pull requests solves the following issues:

- Store raw values in DiscoverySelectors & TestSources #489
- Cannot select default method by fully qualified method name #516
- Non-standard method names are not supported for discovery selection #477

---

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
@@ -254,18 +254,18 @@ public final class DiscoverySelectors {
 	 */
 	public static JavaClassSelector selectJavaClass(String className) {
 		Preconditions.notBlank(className, "Class name must not be null or blank");
-		return new JavaClassSelector(loadClass(className));
+		return new JavaClassSelector(className);
 	}
 
 	/**
 	 * Create a {@code JavaMethodSelector} for the supplied <em>fully qualified
-	 * method name</em>.
+	 * method methodName</em>.
 	 *
 	 * <p>The following formats are supported.
 	 *
 	 * <ul>
-	 * <li>{@code [fully qualified class name]#[methodName]}</li>
-	 * <li>{@code [fully qualified class name]#[methodName](parameter type list)}
+	 * <li>{@code [fully qualified class methodName]#[methodName]}</li>
+	 * <li>{@code [fully qualified class methodName]#[methodName](parameter type list)}
 	 * <ul><li>The <em>parameter type list</em> is a comma-separated list of
 	 * fully qualified class names for the types of parameters accepted by
 	 * the method.</li></ul>
@@ -282,20 +282,15 @@ public final class DiscoverySelectors {
 	 * <tr><td>{@link String#substring(int, int)}</td><td>{@code java.lang.String#substring(int, int)}</td></tr>
 	 * </table>
 	 *
-	 * @param name the fully qualified name of the method to select; never
+	 * @param methodName the fully qualified methodName of the method to select; never
 	 * {@code null} or blank
-	 * @throws PreconditionViolationException if the supplied name is {@code null},
+	 * @throws PreconditionViolationException if the supplied methodName is {@code null},
 	 * blank, or does not specify a unique method
 	 * @see JavaMethodSelector
 	 */
-	public static JavaMethodSelector selectJavaMethod(String name) throws PreconditionViolationException {
-		Preconditions.notBlank(name, "Method name must not be null or blank");
-
-		Optional<Method> methodOptional = ReflectionUtils.loadMethod(name);
-		Method method = methodOptional.orElseThrow(() -> new PreconditionViolationException(
-			String.format("'%s' could not be resolved to a unique method", name)));
-
-		return selectJavaMethod(method.getDeclaringClass(), method);
+	public static JavaMethodSelector selectJavaMethod(String methodName) throws PreconditionViolationException {
+		Preconditions.notBlank(methodName, "Method name must not be null or blank");
+		return new JavaMethodSelector(methodName);
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/DiscoverySelectorResolver.java
```diff
@@ -24,11 +24,8 @@
 import org.junit.platform.commons.meta.API;
 import org.junit.platform.engine.EngineDiscoveryRequest;
 import org.junit.platform.engine.TestDescriptor;
-import org.junit.platform.engine.discovery.ClasspathRootSelector;
-import org.junit.platform.engine.discovery.JavaClassSelector;
-import org.junit.platform.engine.discovery.JavaMethodSelector;
-import org.junit.platform.engine.discovery.JavaPackageSelector;
-import org.junit.platform.engine.discovery.UniqueIdSelector;
+import org.junit.platform.engine.discovery.*;
```
jbduncan: I think this would be clearer as a bunch of explicit imports, rather than a wildcard.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/discovery/DiscoverySelectorResolverTests.java
```diff
@@ -50,11 +50,8 @@
 import org.junit.platform.engine.EngineDiscoveryRequest;
 import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.UniqueId;
-import org.junit.platform.engine.discovery.ClasspathRootSelector;
-import org.junit.platform.engine.discovery.JavaClassSelector;
-import org.junit.platform.engine.discovery.JavaMethodSelector;
-import org.junit.platform.engine.discovery.JavaPackageSelector;
-import org.junit.platform.engine.discovery.UniqueIdSelector;
+import org.junit.platform.engine.discovery.*;
```
jbduncan: Could this be a set of explicit imports as well?

platform-tests/src/test/java/org/junit/platform/engine/discovery/DiscoverySelectorsTests.java
```diff
@@ -14,20 +14,13 @@
...
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.platform.commons.util.CollectionUtils.getOnlyElement;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClasspathResource;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectDirectory;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectFile;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectJavaClass;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectJavaMethod;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectJavaPackage;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectUri;
+import static org.junit.platform.engine.discovery.DiscoverySelectors.*;
```
jbduncan: Could this be done as a set of explicit imports as well?

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherDiscoveryRequestBuilderTests.java
```diff
@@ -40,11 +40,7 @@
 import org.junit.platform.engine.DiscoveryFilter;
 import org.junit.platform.engine.TestEngine;
 import org.junit.platform.engine.UniqueId;
-import org.junit.platform.engine.discovery.ClasspathRootSelector;
-import org.junit.platform.engine.discovery.JavaClassSelector;
-import org.junit.platform.engine.discovery.JavaMethodSelector;
-import org.junit.platform.engine.discovery.JavaPackageSelector;
-import org.junit.platform.engine.discovery.UniqueIdSelector;
+import org.junit.platform.engine.discovery.*;
```
jbduncan: What about this line as well, regarding explicit imports?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/DiscoverySelectorResolver.java
bechte: fine with that one.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/discovery/DiscoverySelectorResolverTests.java
bechte: I would leave this one, as it is in a test class that uses mostly all.

platform-tests/src/test/java/org/junit/platform/engine/discovery/DiscoverySelectorsTests.java
bechte: see above

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherDiscoveryRequestBuilderTests.java
bechte: see above

junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineDiscoveryTests.java
```diff
@@ -676,9 +677,9 @@ private static void assertClassSource(Class<?> expectedClass, TestDescriptor tes
 	private static void assertMethodSource(Method expectedMethod, TestDescriptor testDescriptor) {
 		assertThat(testDescriptor.getSource()).containsInstanceOf(JavaMethodSource.class);
 		JavaMethodSource methodSource = (JavaMethodSource) testDescriptor.getSource().get();
-		assertThat(methodSource.getJavaClass()).isEqualTo(expectedMethod.getDeclaringClass());
```
marcphilipp: There should be a check for `getClassName()` now, shouldn't there?

bechte: True, missed that. Thanks. I will add it.

bechte:
I updated the branch with a `rebase` and a `push --force`. It seems that github will not detect changes this way. It does not resolve the change request by @marcphilipp even though I changed the code. Should we avoid to use `rebase & push --force` on PRs?

marcphilipp:
That's probably because I commented on a line that is still deleted. I've manually approved your changes now. Thanks!

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/JavaClassSelector.java
```diff
@@ -13,36 +13,60 @@
...
+	}
+
+	/**
 	 * Get the selected Java {@link Class}.
 	 */
 	public Class<?> getJavaClass() {
+		if (this.javaClass == null) {
+			this.javaClass = ReflectionUtils.loadClass(this.className).orElseThrow(
+				() -> new PreconditionViolationException("Could not load class with name: " + this.className));
+		}
```
mmerdes: Any danger we might run into thread safety problems here?

platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/JavaSourceTests.java
```diff
@@ -59,6 +59,26 @@ void packageSourceFromPackageReference() {
 	}
 
 	@Test
+	void classNameSource() {
+		String testClassName = "com.unknown.package.ClassByName";
```
mmerdes: Maybe we should avoid packages called 'package'?
:)

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -282,20 +287,24 @@ public static JavaClassSelector selectJavaClass(String className) {
...
+		Preconditions.notBlank(methodName, "Method name must not be null or blank");
 
-		Optional<Method> methodOptional = ReflectionUtils.loadMethod(name);
-		Method method = methodOptional.orElseThrow(() -> new PreconditionViolationException(
-			String.format("'%s' could not be resolved to a unique method", name)));
+		Matcher matcher = fullyQualifiedMethodNamePattern.matcher(methodName);
+		Preconditions.condition(matcher.matches(), "Method name is not a valid full qualified method name");
 
-		return selectJavaMethod(method.getDeclaringClass(), method);
+		if (StringUtils.isNotBlank(matcher.group(3))) {
```
mmerdes: I would prefer to have the various `matcher.group(index)`-calls assigned to variables with explanatory names for better readability

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/JavaClassSelector.java
bechte: Do you mean race conditions? It might but it's not expected that two threads will call it. And if so, it's lazy loading resulting in the same state. So no troubles if two threads calculate the result.

sbrannen:
Hi guys,

Regarding wildcards for imports, we actually agreed on using implicit imports (for both non-static and static imports) for up to 20 imports from the same package/class and only using `*` after 20.

I guess this may not actually be documented or enforced anywhere, _but_ we did agree on that at some point. 😉

bechte:
According imports:
I've changed my IDE to exactly the same numbers (20 for both, static and non-static) to avoid auto-replacements from my side.

According `javaClass`:
Running on the JVM, I dont feel so bad calling it `javaClass`, actually. ;-) It's personal taste (just like in many cases with naming), but I dont like misspelling things, e.g. clazz, paggage, etc. just because of restrictions. I rather have a prefix / suffix. I will rename the `javaClass` back to `clazz` during the renaming of `JavaClassSelector` to `ClassSelector` if that is fine with you.

platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/JavaSourceTests.java
bechte: I changed it to `mypackage`. ;)

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
bechte: I updated the PR.

PR #529, opened 2016-09-30
A couple of small refactorings

nicolaiparlog:

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
@@ -10,6 +10,8 @@
 
 package org.junit.jupiter.engine.extension;
 
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Stream.concat;
 import static org.junit.platform.commons.meta.API.Usage.Internal;
 
 import java.util.ArrayList;
@@ -88,6 +90,37 @@ public class ExtensionRegistry {
 	}
 
 	/**
+	 * Stream all {@code Extensions} in this registry or one of its ancestors of
+	 * the specified type.
+	 *
+	 * @param extensionType the type of {@link Extension} to get
+	 * @see #getReversedExtensions(Class)
+	 * @see #getExtensions(Class)
+	 */
+	public <E extends Extension> Stream<E> stream(Class<E> extensionType) {
+		if (this.parent == null) {
+			return streamLocal(extensionType);
+		}
+		else {
+			return concat(this.parent.stream(extensionType), streamLocal(extensionType));
+		}
+	}
+
+	/**
+	 * Stream all {@code Extensions} in this registry (ancestor are ignored!).
+	 *
+	 * @param extensionType the type of {@link Extension} to get
+	 * @see #getReversedExtensions(Class)
+	 */
+	private <E extends Extension> Stream<E> streamLocal(Class<E> extensionType) {
+		// @formatter:off
+		return registeredExtensions.stream()
+				.filter(extensionType::isInstance)
+				.map(extensionType::cast);
+		// @formatter:on
+	}
+
+	/**
 	 * Get all {@code Extensions} in this registry or one of its ancestors of
 	 * the specified type.
 	 *
@@ -96,20 +129,7 @@ public class ExtensionRegistry {
 	 * @see #stream(Class)
 	 */
 	public <E extends Extension> List<E> getExtensions(Class<E> extensionType) {
-		List<E> extensions = new ArrayList<>();
-
-		if (this.parent != null) {
-			extensions.addAll(this.parent.getExtensions(extensionType));
-		}
-
-		// @formatter:off
-		this.registeredExtensions.stream()
-				.filter(extension -> extensionType.isAssignableFrom(extension.getClass()))
-				.map(extensionType::cast)
-				.forEach(extensions::add);
-		// @formatter:on
-
-		return extensions;
+		return stream(extensionType).collect(toList());
 	}
 
 	/**
@@ -124,18 +144,6 @@ public class ExtensionRegistry {
 		List<E> extensions = getExtensions(extensionType);
 		Collections.reverse(extensions);
 		return extensions;
-	}
-
-	/**
-	 * Generate a stream for iterating over all registered extensions of the
-	 * specified type.
-	 *
-	 * @param extensionType the type of {@link Extension} to stream
-	 * @see #getExtensions(Class)
-	 * @see #getReversedExtensions(Class)
-	 */
-	public <E extends Extension> Stream<E> stream(Class<E> extensionType) {
-		return getExtensions(extensionType).stream();
 	}
 
 	/**
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
@@ -15,6 +15,7 @@ import static org.junit.platform.commons.meta.API.Usage.Experimental;
 import java.util.Optional;
 
 import org.junit.platform.commons.meta.API;
+import org.junit.platform.commons.util.ToStringBuilder;
 
 /**
  * A <em>node</em> within the execution hierarchy.
@@ -113,7 +114,7 @@ public interface Node<C extends EngineExecutionContext> {
 		private static final SkipResult alwaysExecuteSkipResult = new SkipResult(false, null);
 
 		private final boolean skipped;
-		private final String reason;
+		private final Optional<String> reason;
 
 		/**
 		 * Factory for creating <em>skipped</em> results.
@@ -140,7 +141,7 @@ public interface Node<C extends EngineExecutionContext> {
 
 		private SkipResult(boolean skipped, String reason) {
 			this.skipped = skipped;
-			this.reason = reason;
+			this.reason = Optional.ofNullable(reason);
 		}
 
 		/**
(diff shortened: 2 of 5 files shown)


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
```diff
@@ -88,28 +90,47 @@ private ExtensionRegistry(ExtensionRegistry parent) {
...
 	 * @see #getReversedExtensions(Class)
-	 * @see #stream(Class)
+	 * @see #getExtensions(Class)
 	 */
-	public <E extends Extension> List<E> getExtensions(Class<E> extensionType) {
-		List<E> extensions = new ArrayList<>();
-
-		if (this.parent != null) {
-			extensions.addAll(this.parent.getExtensions(extensionType));
+	public <E extends Extension> Stream<E> streamExtensions(Class<E> extensionType) {
```
jbduncan: Wouldn't this read better as a noun rather than a verb? E.g. `extensionsStream` rather than `streamExtensions`?

jbduncan: Or, if you want more consistency with other methods like `getExtensions`, how about `getExtensionsAsStream`? (I'm less eager to suggest this on though, as I think `extensionsStream` is pleasingly shorter and easier to read.)

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
```diff
@@ -88,28 +90,47 @@ private ExtensionRegistry(ExtensionRegistry parent) {
...
 		}
+	}
 
+	/**
+	 * Stream all {@code Extensions} in this registry (ancestor are ignored!).
+	 *
+	 * @param extensionType the type of {@link Extension} to get
+	 * @see #getReversedExtensions(Class)
+	 */
+	private <E extends Extension> Stream<E> streamLocalExtensions(Class<E> extensionType) {
```
jbduncan: See above.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
jbduncan: But then again, it seems to fit JUnit 5's code style more than `extensionsStream`, so I suppose I wouldn't mind either way.

nicolaiparlog: I actually prefer verbs for methods/functions - I'd maybe use nouns for properties if I ditch the get/set. It could be argued that this is such a case. But as you guessed, I wanted to keep this close to `getExtensions` and `getExtensionsAsStream` does that even better so I'm open to that as well. :smiley:

jbduncan: Okay, sounds good to me. :smiley:

Just so that you know where I'm coming from, the reason I disliked the idea of `streamExtensions` was because in OpenJDK 9 there is a forthcoming method `JarFile::versionedStream` (https://bugs.openjdk.java.net/browse/JDK-8163798) which to me looks and feels more like a noun than a verb and which uses "Stream" at the end of the phrase rather than the beginning, so I thought that, if you wanted to stick with something like `streamExtensions`, it would be less jarring and more consistent to use `extensionsStream`. But now that we've both decided on `getExtensionsAsStream`, which follows JUnit 5's coding style better, this is no longer a concern for me. :smile:

jbduncan: (Oops, I was a bit quick to say that we've "decided" on `getExtensionsAsStream`, as I've realised we haven't made a firm decision yet! So I suppose I just want to say that, if you're happy with `getExtensionsAsStream`, then I'm happy that it as well!)

sbrannen:
@nicolaiparlog, regarding the "Refactor extension accessors in 'ExtensionFactory'" commit, please change 'ExtensionFactory' to ExtensionRegistry.

And when you say "no delegation", I'm assuming you mean "now delegation...".

Also, in line with our [contributing guidelines](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#commit-messages), please wrap commit comments around 74 characters.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
sbrannen: Hi guys,

While I certainly appreciate your attention to detail, let's not go overboard here. This is _just_ an internal API. 😉 

The reason the method is currently called `stream` is for DSL-like conciseness. 

Although I can live with `streamExtensions`, I will personally veto `getExtensionsAsStream` since there is no need to be overly verbose here.

The _thing_ in question is an extension registry. Thus getting a stream from it will almost certainly _stream_ the extensions contained in that registry. Honestly, what else would you expect such a method to do? Plus, the method clearly documents what it does.

In summary, I welcome the other changes in this PR, but I don't see a concrete need for renaming the `stream` method.

@junit-team/junit-lambda, what do the rest of you think?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
```diff
@@ -159,6 +159,11 @@ public boolean isSkipped() {
 		public Optional<String> getReason() {
 			return Optional.ofNullable(this.reason);
 		}
+
+		@Override
+		public String toString() {
+			return skipped ? String.format("Skip [%s]", reason) : "Do not skip";
+		}
```
sbrannen: This is not how we implement `toString()`.

Please rework in line with `org.junit.jupiter.api.extension.ConditionEvaluationResult.toString()`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
jbduncan: > While I certainly appreciate your attention to detail, let's not go overboard here. This is just an internal API. :wink:

Haha, fair enough!

> Although I can live with streamExtensions, I will personally veto getExtensionsAsStream since there is no need to be overly verbose here.

Okay then. I can see that I was overly bikeshedding and letting myself be OCD here, so I'm happy for `streamExtensions` to be as-is. :smiley:

nicolaiparlog: > The _thing_ in question is an extension registry. Thus getting a stream from it will almost certainly stream the extensions contained in that registry.

Yes but the same argument could be made for `getExtensions`. I was just irked by the fact that their names differed more than their intent and think they should be unified. But whether it's `get`/`stream`, `getExtensions`/`streamExtensions`, or `getExtensions`/`getExtensionsAsStream` is a secondary concern (although I like the PR's variant best).

But let's not get distracted by bikeshedding. The much more "important" (well, relatively) change is that delegation order changed. I guess that part is ok?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
nicolaiparlog: Will do. Quick question on JUnit's stance on `Optional` fields - do you still do that? `ConditionEvaluationResult` uses one (`reason`) and `SkipResult` might for the same, err, `reason`. Personally I like that style - purposely ignoring the JDK team's recommendations - but I wanted to make sure that this is not just an old artifact.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
marcphilipp: > In summary, I welcome the other changes in this PR, but I don't see a concrete need for renaming the `stream` method.

👍

marcphilipp:
Team decisions:

- Please re-rename `streamExtensions` to `stream`
- Changing the delegation order is fine
- Please change Node.SkipResult to store an `Optional` to be consistent with `ConditionEvaluationResult`

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ExtensionRegistry.java
```diff
@@ -88,28 +90,46 @@ private ExtensionRegistry(ExtensionRegistry parent) {
...
 	 * @see #getReversedExtensions(Class)
-	 * @see #stream(Class)
+	 * @see #getExtensions(Class)
 	 */
-	public <E extends Extension> List<E> getExtensions(Class<E> extensionType) {
-		List<E> extensions = new ArrayList<>();
-
-		if (this.parent != null) {
-			extensions.addAll(this.parent.getExtensions(extensionType));
+	public <E extends Extension> Stream<E> streamExtensions(Class<E> extensionType) {
```
marcphilipp: Please re-rename `streamExtensions` to `stream`.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
```diff
@@ -159,6 +159,11 @@ public boolean isSkipped() {
 		public Optional<String> getReason() {
 			return Optional.ofNullable(this.reason);
```
marcphilipp: Please store an `Optional` like we do in `ConditionEvaluationResult`.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
sbrannen: Nope. That's not an old artifact. 

We _still do that_ if we feel it's appropriate.

nicolaiparlog:
All done. I still think that naming them `getExtensions` and `stream` is inconsistent but I'll capitulate. :crying_cat_face:

sbrannen:
@marcphilipp, can you take care of the "1 review requesting changes"?

sbrannen:
Thanks for the PR, @nicolaiparlog! 👍 

It has been merged into `master`.

Issue #483, opened 2016-08-27
Migrate to dedicated Degraph Gradle plugin


sbrannen:
## Overview

As mentioned by @oehme in #481:

> [He's] working on a [fix](https://github.com/oehme/junit5/commit/1228fd1fefec6623ef57e48364cde340f5b47e51), which is blocked on a DeGraph [issue](https://github.com/schauder/degraph/issues/67). Feel free to up-vote it :)
## Deliverables
- [ ] Migrate to a dedicated Gradle plugin for Degraph once such a plugin exists.

marcphilipp:
@oehme Now that degraph 0.1.4 has been released, do you have time to continue working on the fix?

oehme:
I will, when I come back from vacation in 2 weeks :)

Am 12.09.2016 04:49 schrieb "Marc Philipp" <notifications@github.com>:

schauder:
Minor thing: It's Degraph, not DeGraph

schauder:
If the plugin in agnostic of JUnit (except for it's own test) It would be cool to have it with the the other Degraph modules.  If interested, we can handle it via

1. PR to the Degraph repository
2. I'm willing to grant commit privileges to those doing most of the work
3. Migrate everything to a Degraph organisation and have second repo there

I'm open for alternatives, but would like to avoid the effect of Degraph artifacts coming from two places without an obvious connection.

marcphilipp:
> Minor thing: It's Degraph, not DeGraph

Fixed.

> If the plugin in agnostic of JUnit (except for it's own test) It would be cool to have it with the the other Degraph modules.

Agreed! Let's get a first version for use with JUnit and then figure out how to contribute it to Degraph, ok?

schauder:
> Let's get a first version for use with JUnit and then figure out how to contribute it to Degraph, ok?

Sure.

oehme:
I was initially planning to just write a script for the JUnit team, but
since there is broader interest, I'd be willing to start a real plugin.

You'll need to get an account on plugins.gradle.org so you can publish it.

Am 12.09.2016 12:23 schrieb "Jens Schauder" <notifications@github.com>:

schauder:
No pressure. There is no point in having a half done plugin (that's why I never started one so far)

If out of this comes a code snippet that I can post somewhere as an example how to have Degraph run separately it would help a lot.

oehme:
[...]

PR #532, opened 2016-10-02
Make DeGraph check faster and more user friendly

oehme:
## Overview

This is a first step towards a Degraph Gradle plugin as discussed in #483. It adds a `degraph` task to every project which ensures that the code is cycle-free. We no longer need to scan the whole classpath as each project is checked individually. This makes builds quite a bit faster.

As you can see the DegraphCheck task is still hardcoded with the configuration for the JUnit project. We'd need to discuss how to model these constraints. In the simplest case we could just have a dedicated configuration file, like Degraph [already has](http://blog.schauderhaft.de/degraph/documentation.html#configuration-file) for visulization.

---


gradle/degraph.gradle
```diff
@@ -0,0 +1,49 @@
...
+			.noJars()
+			.including("org.junit.platform.**")
+			.including("org.junit.vintage.**")
+			.including("org.junit.jupiter.**")
+			.withSlicing("module",
+				new NamedPattern("org.junit.vintage.engine.**", "junit-vintage-engine"),
+				new NamedPattern("org.junit.jupiter.api.**", "junit-jupiter-api"),
+				new NamedPattern("org.junit.jupiter.engine.**", "junit-jupiter-engine"),
+				"org.junit.platform.(*).**"
+			)
```
marcphilipp: Does definining the slicing still make sense?

oehme: As far as I can see it is there to provide nicer names in error messages/graphs.

marcphilipp: I'm fine with leaving it in there for now.

oehme:
> Thanks, @oehme! Have you measured how much this speeds up the build?

`gradle clean degraph` is down from 16s to 2.5s. So total build time is reduced by 13.5s.

> In theory, there's a difference between checking each subproject separately and everything at once. Multiple subprojects could contain classes in the same package and thereby introduce a package cycle that is not present when checking them individually. 

I'd consider split package a really bad practice in itself, but I'm sure we could make the plugin flexible enough to check for that too.

marcphilipp:
> I'd consider split package a really bad practice in itself.

Agreed!

marcphilipp:
Thanks again, @oehme! 👍

sbrannen:
Thanks, @oehme!

bechte:
Cool. Thank you. @oehme

PR #533, opened 2016-10-03
Remove 'clean' from build instructions

oehme:
## Overview

Gradle has excellent incremental build support. Invoking 'clean'
is a big waste of the developers time. As a rule of thumb, when
the developer has to invoke 'clean' to make the build work,
 the build is broken.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Code coverage using [Clover] for the latest build is available on the [Jenkins C
 We are thankful to [Atlassian] for providing the Clover license free of charge.
 
 A code coverage report can also be generated locally by executing
-`gradlew -PenableClover clean cloverHtmlReport` if you have a local Clover license file
+`gradlew -PenableClover cloverHtmlReport` if you have a local Clover license file
 on your computer. The results will be available in
 `junit-tests/build/reports/clover/html/index.html`.
 
@@ -62,13 +62,13 @@ project, click on "See console output" on the build scan page.
 All modules can be built with Gradle using the following command.
 
 ```
-gradlew clean assemble
+gradlew assemble
 ```
 
 All modules can be tested with Gradle using the following command.
 
 ```
-gradlew clean test
+gradlew test
 ```
 
 ## Installing in Local Maven Repository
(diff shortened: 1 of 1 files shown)


marcphilipp:
I agree with your point in general. IIRC you should use a `clean` after using Clover, though.

@junit-team/junit-lambda What are your opinions on this one?

sbrannen:
If you're running the tests frequently via Gradle, then I agree that `gradle test` is sufficient.

However, if you're using `cloverHtmlReport`, `assemble`, or `install`, I would then tend to prefer `gradle clean ...` just to be safe.

More importantly, the use of `clean` in the instructions is intended for the average user when executing _pre-commit checks_, again just to make sure things are totally in order before submitting a PR, etc.

It's assumed (albeit perhaps incorrectly) that experienced Gradle users know when it makes sense to clean and when it's ok to skip that.

In summary, I'd leave `clean` in most (if not all) examples and perhaps add a note regarding omitting `clean` for subsequent, frequent test runs.

oehme:
> However, if you're using cloverHtmlReport, assemble, or install, I would then tend to prefer gradle clean ... just to be safe.

This might be leftover fear from using Maven too long ;) Seriously, this is not necessary with Gradle. Gradle knows the inputs and outputs of your tasks, so there is no need to clean between invocations.

> It's assumed (albeit perhaps incorrectly) that experienced Gradle users know when it makes sense to clean and when it's ok to skip that.

You don't need to be experienced. If the build doesn't work and `clean` fixes it, then the build is broken. Usually due to a task not being implemented correctly.

> More importantly, the use of clean in the instructions is intended for the average user when executing pre-commit checks, again just to make sure things are totally in order before submitting a PR, etc.

I know I'm repeating myself, but: This will work just fine with incremental builds.

marcphilipp:
Team decision:

- Keep `clean` in README
- Document Gradle's incremental build support and that you can usually omit `clean`

oehme:
What were the reasons that lead to this decision? Is there something the Gradle team can do?

marcphilipp:
Maybe we're just old-fashioned developers who have used Maven for too long. ;-)

PR #538, opened 2016-10-06
Gradle plugin will honor the build dir, and making extension more usable

ethankhall:
## Overview
Before this change, there was an assumption that all users use the
project directory as the buildDir. In some cases, they are re-mapped to
a single buildDir at the root of the project to make management easier.
This change uses the builDir from the project to set where the output
will be.

Also to make the plugin nicer to use, users can set the reportsDir to be
a GString (or any object really) and rely on the behavior of
Project.file(..) to resolve it to the correct path.

---

### DIFF
--- a/junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/JUnitPlatformExtension.groovy
+++ b/junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/JUnitPlatformExtension.groovy
@@ -9,6 +9,7 @@
  */
 package org.junit.platform.gradle.plugin
 
+import org.gradle.api.Project
 import org.junit.platform.engine.discovery.ClassNameFilter
 
 /**
@@ -17,6 +18,12 @@ import org.junit.platform.engine.discovery.ClassNameFilter
  * @since 1.0
  */
 class JUnitPlatformExtension {
+
+	private Project project
+
+	JUnitPlatformExtension(Project project) {
+		this.project = project
+	}
 
 	/**
 	 * The version of the JUnit Platform to use.
@@ -37,9 +44,23 @@ class JUnitPlatformExtension {
 	/**
 	 * The directory for the test report files.
 	 *
-	 * <p>Defaults to {@code file('build/test-results/junit-platform')}.
+	 * <p>Defaults to {@code file("$buildDir/test-results/junit-platform")}.
 	 */
 	File reportsDir
+
+	/**
+	 * Accepts a path to the reportsDir. If the object is a {@link java.io.File) it
+	 * will be used as is. If the object is anything else, it will convert to File
+	 * automatically using {@link org.gradle.api.Project#file(Object)}
+	 */
+	void setReportsDir(Object reportsDir) {
+		// Work around for https://discuss.gradle.org/t/bug-in-project-file-on-windows/19917
+		if (reportsDir instanceof File) {
+			this.reportsDir = reportsDir
+		} else {
+			this.reportsDir = project.file(reportsDir)
+		}
+	}
 
 	/**
 	 * Whether or not the standard Gradle {@code test} task should be enabled.
--- a/junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPlugin.groovy
+++ b/junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPlugin.groovy
@@ -25,7 +25,7 @@ class JUnitPlatformPlugin implements Plugin<Project> {
 	private static final String TASK_NAME      = 'junitPlatformTest';
 
 	void apply(Project project) {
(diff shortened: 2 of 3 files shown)


sbrannen:
Thanks for the PR!

Note, however, that the build failed on Windows: https://ci.appveyor.com/project/marcphilipp/junit5/build/257

So please address that issue.

ethankhall:
Yep! I'll take a look.

build.gradle
```diff
@@ -77,6 +77,10 @@ allprojects { subproj ->
 	apply plugin: 'com.github.ben-manes.versions' // gradle dependencyUpdates
 	apply from: "$rootDir/gradle/degraph.gradle"
 
+	tasks.withType(Test) { task ->
+		task.testLogging.exceptionFormat = 'full'
+	}
+
```
sbrannen: FYI: the `junitPlatformTest` task created by the Gradle plugin is **not** a `Test` task but rather a `JavaExec` task.

So I don't see how changing the logging for `Test` tasks will help you here.

ethankhall: That's to debug the error. The CI job didn't have a whole lot of info about what happened so I was hoping to get more logging/stacktrace. I'll remove it before merging.

junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/JUnitPlatformExtension.groovy
```diff
@@ -37,11 +44,18 @@ class JUnitPlatformExtension {
...
+	 * <p>Defaults to {@code file("$buildDir/test-results/junit-platform")}.
 	 */
 	File reportsDir
 
 	/**
+	 * Accepts Object, and will convert to file automatically using {@link org.gradle.api.Project#file(Object)}
+	 */
+	void setReportsDir(Object reportsDir) {
+		this.reportsDir = project.file(reportsDir)
+	}
```
sbrannen: This appears to be the source of the failure on Windows.

The input path `"/any"` in the Spock test gets converted to a relative path instead of an absolute path, which is obviously incorrect.

sbrannen:
Here's an excerpt from the Javadoc for Gradle's `Project.file()` method:

> A File. If the file is an absolute file, it is returned as is. Otherwise, the file's path is interpreted relative to the project directory.

So I'd say it's a bug in Gradle, namely that Gradle does not interpret `/any` as an absolute path on Windows.

ethankhall:
Yep. I just opened up an issue on there tracker for this: https://discuss.gradle.org/t/bug-in-project-file-on-windows/19917

sbrannen:
Thanks for raising the issue with the Gradle team!

ethankhall:
I'm going to clean up the RB (remove prints). For the meantime until that can get resolved, I am shot-circuiting Gradle's resolution if the object is a File. I'll update the javadoc to note it.

Do you want me to leave the logging addition to the build.gradle file? It could be useful in the future  for debugging failed builds.

sbrannen:
> Do you want me to leave the logging addition to the build.gradle file? It could be useful in the future for debugging failed builds.

Sure. Why not?

We can always delete it at a later date if we find it annoying. 😉 

Actually, it would probably be better to move the logging change to a separate PR. Would you mind do that?

ethankhall:
Sure can do!

junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/JUnitPlatformExtension.groovy
```diff
@@ -37,11 +44,25 @@ class JUnitPlatformExtension {
...
 	File reportsDir
 
 	/**
+	 * Accepts a path to the reportsDir. If the object is a {@link java.io.File) it
+	 * will be used as is. If the object is anything else, it will convert to File
+	 * automatically using {@link org.gradle.api.Project#file(Object)}
+	 */
+	void setReportsDir(Object reportsDir) {
+		//Work around for https://discuss.gradle.org/t/bug-in-project-file-on-windows/19917
+		if(reportsDir instanceof File) {
```
sbrannen: Please add a space between `if` and `(...`.

sbrannen: Please add a space after `//`

sbrannen:
Per the contributing guidelines, please use the command form for commit message titles (e.g., "Honor buildDir in Gradle plugin" or similar) and keep the length to around 50 characters (max 70-ish).

ethankhall:
Let me know if you need anything else from me!

sbrannen:
Looks good now, and it's passing on Windows.

Thanks, @ethankhall!

sbrannen:

sbrannen:
@ethankhall,

> Let me know if you need anything else from me!

Well.... feel free to take a look at issues tagged with the "up-for-grabs" label! 😉 

Alternatively, you can look at any open issue, but if an issue isn't marked as "up-for-grabs" you should ask the team (via a comment) if it's OK to begin work on that issue.

PR #540, opened 2016-10-07
Test will now print the error to the console when they fail

ethankhall:
## Overview

This will make debugging CI build easier without having to keep track of
all the artifacts that get generated.

---


sbrannen:
Per the contributing guidelines, please use the command form for commit message titles and keep the length to around 50 characters (max 70-ish).

sbrannen:

thanks!

Issue #509, opened 2016-09-13
Split Assertions and AssertionsTests


bechte:
# Status Quo

The classes `Assertions` and `AssertionsTest` have been growing larger since the time they were created. The files are getting too large to keep them easy to maintain.
# Proposals
- Split `AssertionsTest` class into logical units, e.g. `AssertionsEqualsTests`, etc.
- Move private and package-private helper methods from `Assertions` into a package-private `AssertionsUtil` class

jbduncan:
This sounds like a good idea to me! I agree that they're getting too large - they're so much so that sometimes I get confused and I have to retrace my steps to find particular methods.

sbrannen:
SGTM 👍

signed:
I like the approach hamcrest is taking with there [CoreMatchers](https://github.com/hamcrest/JavaHamcrest/blob/master/hamcrest-core/src/main/java/org/hamcrest/CoreMatchers.java) and [Matchers](https://github.com/hamcrest/JavaHamcrest/blob/master/hamcrest-library/src/main/java/org/hamcrest/Matchers.java) classes. They are just a collection of method delegating to the classes actually implementing the logic.

For the Assertions class this would translate to classes like AssertEquals, AssertSame, AssertArrayEquals, ....
And Assertions than just delegating to this classes.

marcphilipp:
I think that sounds like a good idea!

sbrannen:
Resolved in #550 and #559.

PR #550, opened 2016-10-23
split assertions test into multiple smaller test classes

signed:
## Overview
-  
  Split up AssertionsTests class into multiple  Test classes that focus on particular topics (see #509)
---


jbduncan:
I'm not a JUnit team member, but I wanted to say that I feel uneasy about making the separate assertions classes implement `AssertionsHelper`. I suppose my uneasiness comes from the fact that the `Assertions*Tests` classes aren't really sorts of `AssertionsHelper`, whereas, for example, `ArrayList` is a sort of `List`.

I wonder if `AssertionsHelper` could be named better. Maybe `AssertionsTester`?

Or, I also wonder if it would be better implemented as a package-private utility class, so that

- the `default` methods would be translated into `static` methods
- the `interface` part would be changed to `final class`, with a private constructor at the top (to prevent instantiation)
- and the `Assertions*Tests` classes would import this class rather than implement it.

sbrannen:
I agree 100% with the proposal from @jbduncan.

We do not use the _implementation pattern_ proposed in the PR anywhere within the code base (i.e., implementing an interface to access utility methods). So I would not start doing that now.

Instead, `AssertionsHelper` should be a package-private utility class named `AssertionUtils` in line with the rest of the code base.

marcphilipp:
I felt a bit uneasy about that, too, but not enough to reject this pattern. But your arguments are sound, so I'm changing my mind... ;-)

@signed Can you make the requested change?

sbrannen:
Oops, I misspoke. Please name the class `AssertionTestUtils` instead of `AssertionUtils`.

Thanks!

signed:
Done.
If it's ok with you I'll start to work on the Assertions class as proposed [here](https://github.com/junit-team/junit5/issues/509#issuecomment-254938200) and open another pull request when done.

marcphilipp:
> Done.

Thanks! 👍 

> If it's ok with you I'll start to work on the Assertions class as proposed here and open another pull request when done.

Sounds good! 😄

PR #552, opened 2016-10-31
Extend TestExecutionSummary to allow a list of failures to be retrieved.

gdiegel:
## Overview

This change adds a method getFailures() to TestExecutionSummary and implements it. This makes the custom formatting and logging of failures possible instead of having to print to PrintWriter.

---


marcphilipp:
Before getting into details (like the failing build etc.) we should first decide if we want that API to be exposed. Personally, I think it makes sense. On the other hand, it could easily be achieved using a custom listener that collects all failure.

@junit-team/junit-lambda What do you think?

sbrannen:
I have no issues with making `Failure` public; however, it should be an interface local to `TestExecutionSummary` instead of a top-level class (which is currently in an inappropriate package).

marcphilipp:
> it should be an interface local to `TestExecutionSummary` instead of a top-level class

Just to clarify: You want it to be an nested interface in `TestExecutionSummary` like this?

```java 
interface TestExecutionSummary {
    […]
    interface Failure {
        TestIdentifier getTestIdentifier();
        Throwable getException();
    }
    […]
}
```

Right?

sbrannen:
Yes, that's correct.

marcphilipp:
Makes sense! 👍

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/TestExecutionSummary.java
```diff
@@ -127,4 +129,9 @@
 	 */
 	void printFailuresTo(PrintWriter writer);
 
+	/**
+	 * Get a list of the failures of this test execution.
+	 */
+	List<Failure> getFailures();
+
```
marcphilipp: Please create a nested interface for `Failure` inside this interface (see https://github.com/junit-team/junit5/pull/552#issuecomment-257602703).

junit-platform-launcher/src/main/java/org/junit/platform/launcher/Failure.java
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	public TestIdentifier getTestIdentifier() {
+		return testIdentifier;
+	}
+
+	public Throwable getException() {
+		return exception;
+	}
+}
```
marcphilipp: Please call this class `DefaultFailure`, let it implement `Failure`, move it back into `MutableTestExecutionSummary`, and make it private again.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/MutableTestExecutionSummary.java
```diff
@@ -190,6 +191,11 @@ public void printFailuresTo(PrintWriter writer) {
 		}
 	}
 
+	@Override
+	public List<Failure> getFailures() {
+		return failures.stream().collect(Collectors.toList());
```
marcphilipp: Please change this to

```java
return new ArrayList<>(failures);
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/MutableTestExecutionSummary.java
```diff
@@ -45,7 +45,7 @@
 	final AtomicLong testsFailed = new AtomicLong();
 
 	private final TestPlan testPlan;
-	private final List<Failure> failures = new ArrayList<>();
+	private final List<DefaultFailure> failures = new ArrayList<>();
```
sbrannen: AFAICT, there is no reason to use `DefaultFailure` as the generic type here.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/TestExecutionSummary.java
```diff
@@ -127,4 +129,16 @@
...
+	interface Failure {
```
sbrannen: Since this is now a public type, both the interface and its methods require proper Javadoc.

marcphilipp:
@gdiegel Thanks! 👍

gdiegel:
Thank you for your comments and accepting the PR!

PR #559, opened 2016-11-03
Move implementation of Assertions into separate classes

signed:
## Overview

- Split up the implementation of Assertions into multiple classes as proposed in the [comment](https://github.com/junit-team/junit5/issues/509#issuecomment-254938200) to #509

---


marcphilipp:
Thanks! I think we should make the individual classes and the now public methods in there package-private. 

@junit-team/junit-lambda Do you agree?

signed:
> since the fail(...) methods are more or less utilities and not actually assertions in their own right.

Following this reasoning shouldn't the two fail methods in `Assertions.fail(...)` be moved elsewhere e.g. `Fail.fail(...)`?

sbrannen:
> Should I mark the classes with @API(Internal) as well?

No. Non-public types do not need to be annotated with `@API`.

sbrannen:
> Following this reasoning shouldn't the two fail methods in `Assertions.fail(...)` be moved elsewhere e.g. `Fail.fail(...)`?

No, I don't think so.

We don't want to introduce a new class just for that purpose since that would only complicate things for users who actually need one of those two `fail(...)` methods. In other words, in terms of the public API, I think it's fine to retain those two methods in `Assertions`.

marcphilipp:
Despite my approval, Checkstyle found an unused import. :-/

signed:
3rd commit should fix it.
JavaDoc needed the import, but nobody else in the class. Used an absolute import in the java doc. Lets see if this fixes the problem.
While running the branch locally with ./gradlew clean spotlessApply build there where no checkstyle complaints :/

marcphilipp:
I also cannot reproduce this locally. Actually, our Checkstyle configuration set `processJavadoc` to `true` for the `UnusedImports` module which is supposed to collect references in Javadoc. Since we're using the same Checkstyle version on AppVeyor, Travis, as well as on your machine and mine, I don't have an explanation, though.

marcphilipp:
Thanks, @signed! 👍

PR #561, opened 2016-11-04
Updates references to AssertionFailedError to use its FQCN.

### LINKED ISSUE #560: Link to AssertionFailedError in the Javadoc
In the [Javadoc for the `Assertions` class](http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/Assertions.html), the following is mentioned:

> A *failed* assertion will always throw `AssertionFailedError` or a subclass thereof.

Junit doesn't declare that `AssertionFailedError` class, nor does the JDK (see [Junit's Javadoc](http://junit.org/junit5/docs/current/api/allclasses-frame.html) and [JDK's Javadoc](http://docs.oracle.com/javase/8/docs/api/allclasses-frame.html)).

So where is that class defined? Can you please add links in the Javadoc to the class you're actually referencing? I think that simply finding the missing link will be enough in the Javadoc build.

smoyer64:
Since javadocs are produced and published to Maven Central for both the JUnit 5 and OpenTest4J projects, Eclipse and other IDES don't have a problem linking imported Javadocs between the projects.  The JUnit 5 javadocs published at http://junit.org/junit5/docs/current/api/ cannot however resolve these links, so providing the FQCN provides readers with additional information to locate the related javadocs.

---

Resolves #560.

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
@@ -45,10 +45,10 @@ import org.opentest4j.MultipleFailuresError;
  * conditions in tests.
  *
  * <p>Unless otherwise noted, a <em>failed</em> assertion will throw an
- * {@link AssertionFailedError} or a subclass thereof.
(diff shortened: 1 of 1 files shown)


sbrannen:
merged. thanks!

PR #562, opened 2016-11-04
Documents the use of ExtensionContext.Store.

### LINKED ISSUE #287: Document ExtensionContext.Store in the User Guide
Section 5.8 -- "Keeping State in Extensions" -- in the User Guide currently states:

>   Usually, an extension is instantiated only once. So the question becomes relevant: How do you keep the state from one invocation of an extension point to the next? …​

The ParameterResolvers for upREST involve parameters that are related to each other, and regardless of the test, an HttpRequest and the subsequent HttpResponse have to be generated.  For parameters like an HTTP StatusLine or HTTP EntityBody, you don't want to make additional HTTP requests for each parameter that's being resolved

My current method is for each parameter type to understand it's dependent parameters (whether they're passed to the test method or not), and to retrieve those parameters from the `ExtensionContext.Store`.  If the dependent parameter isn't found in the store, then its parameter resolver is called and the results of that call are both added to the store and returned to the dependent parameter's resolver.

I should note that I've only found this to be necessary for ParameterResolvers called to execute a single test.  Since I'm a relatively big fan of keeping tests as decoupled as possible (so they can run concurrently), I can't foresee needing this technique in other types of extensions.

If you'd like, I could expand this section but even with a simplified scenario, it's going to be more than a few paragraphs.

smoyer64:
Using the existing TimingExtension as an example.

---

Resolves #287

### DIFF
--- a/documentation/src/docs/asciidoc/extensions.adoc
+++ b/documentation/src/docs/asciidoc/extensions.adoc
@@ -236,5 +236,13 @@ include::{testDir}/example/exception/IgnoreIOExceptionExtension.java[tags=user_g
 
 Usually, an extension is instantiated only once. So the question becomes relevant: How do
 you keep the state from one invocation of an extension to the next? The
-`ExtensionContext` API provides a `Store` exactly for this purpose. Consult the
-corresponding Javadoc for details.
+`ExtensionContext` API provides a `Store` exactly for this purpose.  Extensions may
+put values into a store for later retrieval.  See the `TimingExtension` in section
+5.6.1 of this document for an example of using the `ExtensionContext.Store` with a
+method-level scope.  It is important to remember that values stored during
+`TestExtensionContext` will not be available during the `ContainerExtensionContext`
+execution.  `ContainerExtensionContext`s may be nested, so the scope of inner contexts
+may also be limited.
+
+Consult the corresponding Javadoc for details on the methods available for storing and
+retrieving values via the `ExtensionContext.Store`.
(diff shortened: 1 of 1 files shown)


documentation/src/docs/asciidoc/extensions.adoc
```diff
@@ -236,5 +236,13 @@ include::{testDir}/example/exception/IgnoreIOExceptionExtension.java[tags=user_g
 
 Usually, an extension is instantiated only once. So the question becomes relevant: How do
 you keep the state from one invocation of an extension to the next? The
-`ExtensionContext` API provides a `Store` exactly for this purpose. Consult the
-corresponding Javadoc for details.
+`ExtensionContext` API provides a `Store` exactly for this purpose.  Extensions may
+put values into a store for later retrieval.  See the `TimingExtension` in section
+5.6.1 of this document for an example of using the `ExtensionContext.Store` with a
```
marcphilipp: Please replace this textual reference with a proper linked cross reference.

smoyer64:
Link added per your code review.

marcphilipp:
@smoyer64 Thanks for the pull request! I hope it's ok with you that I made a few changes to the wording. I was particularly concerned about having a hard-coded section number in there ("5.6.1") because those are bound to change.

smoyer64:
No offense taken ... in many ways, the documentation (both users guide and javadocs) is more important than the code.  I've also been doing this long enough to avoid feeling "ownership".  I'm sure someone will make this section better next month (next year, etc).

PR #564, opened 2016-11-07
Allows @API on ElementType.FIELD

smoyer64:
While working on #144 I noticed that public-scoped fields (presumably constants) can only inherit their @API Usage from the enclosing class.  This change would, at a minimum, allows constants to be marked with Usage.Deprecated or Usage.Internal while their enclosing class remains accessible.

---


marcphilipp:
Thanks!

Issue #518, opened 2016-09-18
Possibility to customize element resolving


aaschmid:
Hi @sbrannen,

according to your last comment on https://github.com/TNG/junit-dataprovider/issues/75, I looked deeper into the `Extension` API. I agree that I do not  want to provide a custom test engine. Maybe you can help me about implementing an extension for junit-dataprovider to run the tests with junit-jupiter rather than a complete test engine.

Correct me if I am wrong but as far as I understand the `Extension` API is currently not capable of registering additional test cases. However, extending the junit-jupiter framework to be able to register  another `ElementResolver` could discover additional (dynamic) test cases which fit the junit-dataprovider programming model. 

I found no issue / feature request concerning this topic, so I opend this one. Have you already thought about the possibility of register custom `ElementResolver` or does that fit into your vision of junit5? I would love to help out providing a pull request if you give me a little vision how you could think about a good solution for it.
At my first tries it felt natural to use the `Extension` API for this but while doing so, I saw that it is only available / used within `org.junit.platform.engine.TestEngine.execute(ExecutionRequest)`. 

Cheers,
Andreas

aaschmid:
As I haven't got an answer, I assume you either did not consider it yet as important enough or you have no clue how to do it (as I do / did).

Therefore I looked deeper into the code, played around a bit and got the following ideas how this could be solved (only ideas in arbitrary order):

* change / override the ```org.junit.jupiter.engine.discovery.TestMethodResolver``` via extension
* add some extension mechanism to ```org.junit.jupiter.engine.discovery.TestMethodResolver.resolveMethod(Method, ClassTestDescriptor, UniqueId)```
  * could maybe also used for ```@TestFactory``` annotated test methods to include dynamic tests into the lifecyle and get ```@BeforeEach``` / ```@AfterEach```support
* make it possible to register a new element resolver in ```org.junit.jupiter.engine.discovery.JavaElementsResolver```

Please let me now if you like any of the ideas (or if you dislike this issue at all). I would love to get feedback to start with a simple implementation to get a proof-of-concept and to be able to further discuss possibilities.

Thanks & regards,
Andreas

aaschmid:
I tried some things and like the following quite well at the momemt:
https://github.com/junit-team/junit5/compare/master...aaschmid:issue518-custom-element-resolver

[...]

Issue #506, opened 2016-09-10
Introduce mechanism for programmatic extension management


sbrannen:
## Background

This issue addresses some of the topics discussed in #112.

## Status Quo

It is currently possible to register extensions _declaratively_ via `@ExtendWith` on types and methods, and #497 will support programmatic extension registration via fields. However...
- Developers cannot _alter the order_ of registered extensions.
- Developers cannot _remove_ a registered extension.
- Developers cannot _insert_ an extension into the list of registered extensions -- for example, at the front or somewhere in the middle.

## Rationale

In order to avoid the introduction of an overly complex declarative model for sorting, appending, prepending, inserting, and removing extensions, we have opted for a programmatic means to achieve all of the above. Providing a programmatic means to achieve these goals frees developers to manage registered extensions as they see fit without any unnecessary restrictions imposed on them by the framework itself.

## Considerations

Such a mechanism could itself be an `Extension` registered declaratively via `@ExtendWith`; however, the current thinking in the team is that there should be one and only one such component registered at any given time. Since this is such a special case which affects all extensions which have been registered declaratively, it is therefore considered best to introduce a new declarative mechanism for registering this single _component_. Similarly, the API for such a component should therefore not extend `Extension` since doing so would allow users to incorrectly register it via `@ExtendWith` (in which case it would simply be silently ignored).

## Proposal

- Introduce an `ExtensionManager` API that receives a list of all registered extensions as input and returns a list of extensions.
- Introduce a `@ManageExtensionsWith` annotation that allows the user to declare a single class reference for an implementation of `ExtensionManager`.

## Related Issues

- #13 
- #416
- #497
- #864 
- #1707 

## Deliverables

- [ ] Introduce a mechanism for programmatic extension management.

sormuras:
Gist from https://stackoverflow.com/questions/47633819/injecting-test-classes-created-by-platform-launcher ... allow extension instances be part of a `LauncherDiscoveryRequest`.

        LauncherDiscoveryRequest request = LauncherDiscoveryRequestBuilder
                .request()
                .selectors(...)
                .extendWith(TestInstancePostProcessor.class, new Mutator("one else!"))
                .build();

sbrannen:
@sormuras, the proposed `extendWith()` method in the launcher API does not address the issue of extension _management_. Rather, it simply provides another means for programmatic extension _registration_.

So, unless I'm missing something, the proposal is not relevant to this GitHub issue.

[...]

Issue #14, opened 2015-11-18
Introduce support for parameterized tests


elygre:
## Deliverables

- [x] Introduce `@TestTemplate` and accompanying extension point.
- [x] Determine if #704 should be included in M4.
- [x] Determine if we need an `@InvocationIndex` annotation or `TestTemplateInfo` type for injecting the current _invocation index_, etc. into `@Test`, `@BeforeEach`, and `@AfterEach` methods.
  - See https://github.com/junit-team/junit5/commit/d2e3a4dfdc8cc4b4b8609e4297f8e00b753cb5b3
- [x] Introduce `junit-jupiter-params` module
  - [x] Merge initial contribution (#728)
  - [x] Javadoc
    - [x] `@since` tags
  - [x] API annotations
  - [x] Add unit tests where missing (see TODOs)
  - [x] Resolve TODOs
- [X] Document `@TestTemplate` and accompanying extension point ~~in User Guide and~~ Release Notes
  - [X] ~~Provide example(s) in the `documentation` module and include in the User Guide.~~ Will be addresses in #762
- [x] Document parameterized tests in User Guide and Release Notes
  - [x] Document the `junit-jupiter-params` module in the _Dependency Metadata_ section.
  - [x] Document the `junit-jupiter-params` module in the _Dependency Diagram_.
  - [x] Provide example(s) in the `documentation` module and include in the User Guide.

----

## Original Issue Description

At the time of writing, the https://github.com/junit-team/junit-lambda/wiki/Prototype-Test-Decorators page does not have all that much detail, so this might already be considered.

It would be good if the library shipped with ready-to-use simple parameter annotations that could be used inside a single test instance. It would in many ways look like JUnit 4's `@Parameterized` Runner, but for use inside a single class. When running the test, each `@Parameterized` method would appear many times, once for each data set.

``` java
class MyTest {

   // A simple test with a specified data set
   @Test @Parameterized("mod10")
   void testMod10(String value) { ... }

   @Parameterized("mod10") 
   Collection<Object[]> getMod10Values { ... }

   // Also including a method naming convention
   @Test @Parameterized("mod11", name="{0}")
   void testMod11(String value) { ... }

   @Parameterized("mod11")
   Collection<Object[]> getMod11Values { ... }

   // In the absense of a @Parameterized data provider, use reflection
   @Test @Parameterized("getMod12Values", name="{0}")
   void testMod12(String value) { ... }

   Collection<Object[]> getMod12Values{ ... }
}
```

----

[...]

PR #566, opened 2016-11-11
first draft for custom element resolver (#518)

aaschmid:
## Overview

To be able to provide a custom `ElementResolver` (#518), I tried some things and like this one quite well at the moment.

## Details

It fits quite nice (IMHO) into the existing code / API by resolving the `ExtensionRegistry` earlier within the affected test descriptions (problem of mutability should be solved, though). This could also be more generalized by moving the new property up to `JupiterTestDescription` and add the current default element resolvers from `DiscoverySelectorResolver` to default extensions. 

## Disclaimer

The junit-dataprovider tests would already run using some quick and dirty implementation, see https://github.com/TNG/junit-dataprovider/compare/issue75-custom-junit5-element-resolver?expand=1
Also the test fixes are a bit hacky at the moment

## Your opinion?
 
Does that go into the right direction? Or do you have something different in mind? What and why do you like / not like it (such that I can think of another solution)?

P.S.: This is just a first PoC / draft to discuss it more concrete ;-)

----


aaschmid:
If I understand #506 correctly, it could be an alternative of solving this. Namely, by replacing some the element resolver with a custom one.

aaschmid:
Seems that `@TestTemplate` (#14) is an alternative solution for this ...

aaschmid:
Closing this

Issue #565, opened 2016-11-10
Classpath scanner should not load JDK 9 module-info.class


akozlova:
```
java.lang.NoClassDefFoundError: java/jnlp/module-info (wrong name: module-info)
 at java.lang.ClassLoader.defineClass1(java.base@9-ea/Native Method)
 at java.lang.ClassLoader.defineClass(java.base@9-ea/ClassLoader.java:941)
 at java.security.SecureClassLoader.defineClass(java.base@9-ea/SecureClassLoader.java:150)
 at jdk.internal.loader.BuiltinClassLoader.defineClass(java.base@9-ea/BuiltinClassLoader.java:554)
 at jdk.internal.loader.BuiltinClassLoader.access$200(java.base@9-ea/BuiltinClassLoader.java:89)
 at jdk.internal.loader.BuiltinClassLoader$3.run(java.base@9-ea/BuiltinClassLoader.java:467)
 at jdk.internal.loader.BuiltinClassLoader$3.run(java.base@9-ea/BuiltinClassLoader.java:461)
 at java.security.AccessController.doPrivileged(java.base@9-ea/Native Method)
 at jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(java.base@9-ea/BuiltinClassLoader.java:460)
 at jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(java.base@9-ea/BuiltinClassLoader.java:407)
 at jdk.internal.loader.BuiltinClassLoader.loadClass(java.base@9-ea/BuiltinClassLoader.java:365)
 at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(java.base@9-ea/ClassLoaders.java:186)
 at java.lang.ClassLoader.loadClass(java.base@9-ea/ClassLoader.java:419)
 at org.junit.platform.commons.util.ReflectionUtils.loadClass(ReflectionUtils.java:330)
 at org.junit.platform.commons.util.ClasspathScanner.processClassFileSafely(ClasspathScanner.java:154)
 at org.junit.platform.commons.util.ClasspathScanner.lambda$findClassesForPath$3(ClasspathScanner.java:139)
 at org.junit.platform.commons.util.ClassFileVisitor.visitFile(ClassFileVisitor.java:40)
 at org.junit.platform.commons.util.ClassFileVisitor.visitFile(ClassFileVisitor.java:24)
 at java.nio.file.Files.walkFileTree(java.base@9-ea/Files.java:2695)
[...]

PR #567, opened 2016-11-13
Convert NoClassDefFound error to empty class optional

sormuras:
## Overview

Fix for #565 by converting `NoClassDefFoundError` (like `ClassNotFoundException`) into an empty `Optional<Class<?>>`. This prevents warnings when trying to load `module-info` classes. The warning can be seen here: https://youtrack.jetbrains.com/issue/IDEA-163795

---


sormuras:
Covered by https://github.com/junit-team/junit5/commit/d065ec688c2f442942215081eca780a2e5d384ca

sbrannen:
Thanks for the PR!

Please note, however, that the proposed solution has been superseded by the fix in d065ec688c2f442942215081eca780a2e5d384ca.

PR #524, opened 2016-09-28
Introduce Details, TreePrintingListener and VerboseTreePrintingListener

sormuras:
## Overview

Solves #369 and #546.

Removes `--hide-details` command line option and introduces an enum named **`Details`** in package `org.junit.platform.console.options` with the following constants:

``` java
enum Details {
	/** No test plan execution details are printed. */
	NONE,

	/** Test plan execution details are rendered in a flat, line-per-line mode. */
	FLAT,

	/** Test plan execution details are rendered as a simple tree. */
	TREE,

	/** Combines tree and flat mode. */
	VERBOSE
}
```

## Detailed description of `Details` constants

**NONE**

No test plan execution details are printed.

**FLAT**

The standard `ColoredPrintingTestListener` kicks in.

**TREE**

The `TreePrinter` renders the test plan execution tree in a compact manner.

- UTF-8 https://travis-ci.org/junit-team/junit5/jobs/180752503
- ASCII https://ci.appveyor.com/project/marcphilipp/junit5/build/424
- Monochrome (disabled ANSI colors) snippet below
```
├─ JUnit Vintage
│  ├─ example.JUnit4Tests
│  │  ├─ standardJUnit4Test ✔
├─ JUnit Jupiter
│  ├─ AssertionsDemo
│  │  ├─ timeoutNotExceeded() ✔
│  │  ├─ standardAssertions() ✔
│  │  ├─ groupedAssertions() ✔
│  │  ├─ exceptionTesting() ✔
│  │  ├─ timeoutExceeded() ✔
│  │  ├─ timeoutExceededWithPreemptiveTermination() ✔
│  ├─ AssumptionsDemo
│  │  ├─ testOnlyOnCiServer() ■ Assumption failed: assumption is not true
│  │  ├─ testOnlyOnDeveloperWorkstation() ■ Assumption failed: Aborting test: not on developer workstation
│  │  ├─ testInAllEnvironments() ✔
│  ├─ StringTests
│  │  ├─ returnsZeroWhenComparedToItself() ✔
│  │  ├─ returnsPositiveNumberComparedToSmallerValue() ✔
│  │  ├─ returnsNegativeNumberComparedToSmallerValue() ✔
│  │  ├─ valueEqualsItself() ✔
│  │  ├─ valueDoesNotEqualNull() ✔
│  │  ├─ valueDoesNotEqualDifferentValue() ✔
│  ├─ DisabledClassDemo class example.DisabledClassDemo is @Disabled
[...]

Issue #369, opened 2016-06-30
Introduce ASCII-based tree printing TestExecutionListener


sbrannen:
## Status Quo

Since build tools and IDEs do not yet support the JUnit Platform, test results cannot currently be viewed in their true tree based structure. This is especially true for _dynamic tests_.
## Deliverables
- [x] Introduce an ASCII-based `TestExecutionListener` that prints the test tree to STDOUT (or a configurable output destination) once the `TestPlan` has completed.
- [x] Allow the new listener to be automatically used by the `ConsoleLauncher`.
- [x] Allow the new listener to be enabled via the JUnit Platform Gradle plugin.

sormuras:
Although @mmerdes is assigned to this issue, I rolled my own tree printing test execution listener. See it's print out starting at line 275 here: https://travis-ci.org/sormuras/beethoven/builds/159539379 - avoiding ANSI colors and extended ASCII characters.
I'd be pleased to tidy up the [code](https://github.com/sormuras/beethoven/blob/master/src/test/java/org/junit/platform/console/tasks/ColoredPrintingTestListener.java) and file a PR as basis for further developement and tile layout tweaks.

sbrannen:
@sormuras, I checked out the output in your build.

I personally had something more in mind of how the standard `tree` command outputs directory structures, like the following for the JUnit 5 repository:

```
.
├── CONTRIBUTING.md
├── LICENSE.md
├── README.md
├── build.gradle
├── documentation
│   ├── README.md
│   ├── bin
│   │   ├── example
│   │   │   ├── Address.class
│   │   │   ├── AssertionsDemo.class
│   │   │   ├── AssumptionsDemo.class
│   │   │   ├── DisabledClassDemo.class
│   │   │   ├── DisabledTestsDemo.class
│   │   │   ├── DisplayNameDemo.class
│   │   │   ├── DocumentationTestSuite.class
│   │   │   ├── DynamicTestsDemo$1.class
│   │   │   ├── DynamicTestsDemo.class
│   │   │   ├── Fast.class
```

So unless @mmerdes states otherwise, feel free to put together a PR if you like! We can then fine tune the output later. However you decided to implement your tree-printing `TestExecutionListener`, please make sure that it can be used standalone and that is is not dependent on existing listeners. If you feel that is not feasible, please raise your concerns here.

Cheers!

sormuras:
[...]

PR #568, opened 2016-11-15
Upgrade to Gradle 3.2

sormuras:
## Overview

Upgrade to Gradle 3.2

See https://discuss.gradle.org/t/gradle-3-2-released/20441
Fixes UTF-8 emission bug described in the realm of #524 -- explicitly here https://github.com/junit-team/junit5/issues/369#issuecomment-246772931

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -545,7 +545,7 @@ configure(rootProject) {
 
 	task wrapper(type: Wrapper) {
 		description = 'Generates gradlew and gradlew.bat scripts'
-		gradleVersion = '3.1'
+		gradleVersion = '3.2'
 	}
 
 }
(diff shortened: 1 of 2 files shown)


sbrannen:

Thanks

sbrannen:
Next time, please make sure you actually run `gradle wrapper` after updating the wrapper version. 😉  (see 264f52bdd5714b9dccd393d45f46fb2cc35afc1f).

sormuras:
Pardon the inconvenience.

I just saw the two textual changes in https://github.com/junit-team/junit5/commit/224ac22aa27f65d5bebf7c284418c75605e65d15 and thought that'd be all. Missed a) the binary and b) the Gradle update instructions.

Next time.

sbrannen:
> Pardon the inconvenience.

No worries!

And well... it could be that I forgot to "do the right thing" that time. 😉 

It happens to the best of us.

In any case, I now typically run `gradle wrapper; gradle --version` twice... _just to be sure_... before committing. Plus, when you do it that way, there is no need to update the `gradle-wrapper.properties` file manually: it will get updated automatically.

Issue #354, opened 2016-06-24
RFC: Please critique this working PoC for an AfterDiscoveryCallback.


smoyer64:
### Rationale

JUnit 5 is trying to break the duopoly of Runners and @Rules that limits the vintage JUnit solution.  While JUnit 5 does eliminate the vintage system's inability to use more than one Runner through the use of Engines, the discovered Engines don't interact with each other.  Many types of Extensions enable functionality that can be changed during a test's execution, but there are no extensions that allow a TestDescriptor to be changed before, during or after test discovery.

The changes that provide the AfterDiscoveryCallback allow an Extension to alter the engine's (root) TestDescriptor before it becomes immutable as part of the TestPlan.  Classes of Extensions that benefit from this callback include:
-   ScenarioExtension:  Using the @Scenario({"predecessor1", "predecessor2"}) annotation, tests can be ordered based on a list of predecessor tests (by test method name).  The AfterDiscoveryCallback would find all the instances of @Scenario (which extends @Test) in a test class and, based on the dependency graph described by the predecessors, would replace the set of child MethodTestDescriptors in the parent ClassTestDescriptor with one of the sets that maintain order.  The child @Scenario annotated methods would be reordered to honor the requested dependency graph (assuming it's acyclic).
-   [JUnitParamsExtension](http://pragmatists.github.io/JUnitParams/):  This vintage JUnit Runner is a popular way of parameterizing tests.  There are multiple ways declaratively and programmatically providing a list of parameter sets to a method.  The size of the set must match number of parameters in the method declaration, but the size of the list determines how many times that method is executed.  In this case, the MethodTestDescriptor would be changed so that it was a test container, and new test methods would be added based on the size of the parameter set list.

If desired, test filtering could be a special case of this callback.  Care must be taken to make sure that filter conditions are executed first (since there's no point translating or expanding tests that are destined to be removed).

There are many more examples of these three classes of Extensions (reduce, translate or expand) that want to alter the TestDescriptor hierarchy. 
### Status

The PoC is available at https://github.com/selesy/junit5/tree/feature/discovery-callbacks.  Currently, there's a no-op AfterDiscoveryCallback installed as a default Extension - it doesn't adapt the TestDescriptor in any way but it does print TestDescriptor information to System.out.  All the existing unit and platform tests successfully execute (only one test was changed since there's now one more default extension).  No tests have been added for this code (so far).  While I've done a little clean-up and formatting, I didn't submit a pull request for this PoC since there are known issues.
[...]

PR #577, opened 2016-11-30
Adds DiscoveryCallbacks

smoyer64:
As described in #354, this PR provides access to the test plan before it becomes immutable and facilitates the creation of extensions to process parametric tests as well as to guarantee test order.  See #354 for a much longer rationale.

Like others, I had shelved this concept until M3 was out but I'd like to participate in the discussion since parametric tests (via JUnitParams) are one of my favorite patterns and I've been waiting for this ability in upREST.  This extension point is extremely powerful and therefore also dangerous.  I'd be happy to review other ways that parametric, ordered, scenario, theories tests might work and there's definitely value in only having one supported mechanism for each.

---


marcphilipp:
I don't think the rebase worked... You didn't really change 330 files, did you?

smoyer64:
No I didn't!  I'm not entirely sure how I produced that commit (I'll look later as it's on another laptop).

This PR is really for team discussion and should be labeled "don't-merge" since we don't really want the DiscoveryReportExtension to be added to the default ExtensionRegistry (it's only a demonstration of how the new callback works).

PeterWippermann:
I'd be really happy to see this proposal making progress, because I agree with @smoyer64 that this extension point would be very valuable. In my current understand this is the key feature to enable parameterized tests as we had them in JUnit 4.
@marcphilipp could you foster an internal discussion in your team please?

marcphilipp:
We are going to meet next week and will discuss this proposal. Thanks for your patience!

marcphilipp:
Please note that we've discussed parameterized tests in the team and they will be the main theme for M4. However, since we want to support use cases with a large number of parameters that may be read from a file etc. we will not create a `TestDescriptor` for each parameterized invocation during discovery. Instead, we will report additional `TestDescriptors` during execution, similar to how dynamic tests work under the hood. The programming model will probably not use `@TestFactory` methods, rather something that looks very similar to a regular `@Test` method. We'll let you know when we have a detailed proposal. You can track progress by subscribing to #14.

smoyer64:
I'm playing with the ```@TestTemplate``` and finding it very much to my liking and much more appropriate for parameterized tests.  As you noted, this doesn't cause the potential explosion of test descriptors for large sets of parameters.  This callback may still be useful for various types of sorted tests.

bechte:
I am wondering if it makes sense to have a single extension point at the end of the discovery phase or if it would make more sense to have a callback after each test descriptor being found, initialized and added to the test plan? Would that also help?

smoyer64:
@bechte - Somehow I missed your comment in September.  I think that a callback after each ```TestDescriptor``` was discovered would cover a the scenarios described in #354.  And since the root container is itself a ```TestDescriptor```, you would be passed the entire hierarchy just like the proposed ```AfterDiscoveryCallback```.  The difference is that, if I'm reading the code right, you'd get the root ```TestDescriptor```'s callback before it's hierarchy of children was established.  Otherwise your idea seems much better than my PoC since you wouldn't necessarily need to traverse the test tree yourself!

smoyer64:
@bechte - Somehow I missed your comment in September.  I think that a callback after each ```TestDescriptor``` was discovered would cover a the scenarios described in #354.  And since the root container is itself a ```TestDescriptor```, you would be passed the entire hierarchy just like the proposed ```AfterDiscoveryCallback```.  The difference is that, if I'm reading the code right, you'd get the root ```TestDescriptor```'s callback before it's hierarchy of children was established.  Otherwise your idea seems much better than my PoC since you wouldn't necessarily need to traverse the test tree yourself!

marcphilipp:

smoyer64:
I think ``TestTemplate`` is a safer way to go and accomplishes (so far) everything I've tried to do with test permutations.

Issue #583, opened 2016-12-04
Load and register custom TestExecutionListener implementations automatically


sormuras:
What about adding support for loading custom `TestExecutionListener` implementations via the Service API to the console launcher?

Enhance (and document) [ExecuteTestsTask#registerListeners](https://github.com/junit-team/junit5/blob/master/junit-platform-console/src/main/java/org/junit/platform/console/tasks/ExecuteTestsTask.java#L90) with:
```java
ServiceLoader.load(TestExecutionListener.class)
  .forEach(launcher::registerTestExecutionListeners);
```

Or add the load and register functionality to the platform `DefaultLauncher` itself?

marcphilipp:
I think we should add them to the DefaultLauncher not just when using ConsoleLauncher. I also proposed we do this in https://github.com/junit-team/junit5/issues/542#issuecomment-257308591.

marcphilipp:
@junit-team/junit-lambda Any objections?

sbrannen:
No objections. Go for it!

sbrannen:
Resolved in 3c52d958f1e3de36726724c62012d75f55db6d8b.

marcphilipp:
This still needs to be documented in the User Guide and Release Notes. We haven't really used a consistent strategy to address that need in the past. We could reopen this issue, create a new one, or procrastinate until some poor guy wants to do a release and has to go through all the issues.

I'm in favor of the first option.

@junit-team/junit-lambda What do you think?

sbrannen:
> This still needs to be documented in the User Guide and Release Notes. 

Oooooops. Yes... you are totally correct!

I'm in favor of reopening the original issue in such cases: no need to multiple issues for the _same issue_.

Since I prematurely closed it, I'll reopen it. 😉

sbrannen:
In retrospect, we need to (re)define our _Definition of Done_ for issues. Though perhaps we should discuss that amongst the team and then document it in the wiki (and of course actually adhere to the rules).

sbrannen:
Actually, I'll create an issue to address the DoD.

sbrannen:
See #593 regarding the _Definition of Done_.

PR #585, opened 2016-12-05
Load and register custom TestExecutionListener implementations automa…

sormuras:
## Overview

The `DefaultLauncher` now uses the `ServiceLoader` API to automatically load and register custom `TestExecutionListener` implementations discussed in #583 .

~~_If this is going in the right direction, I'll add unit tests for the new code._~~

---


sormuras:
The loading test case introduces two package-private getters to inspect the registered listeners by the default launcher factory. Could be a) promoted to public `Launcher` API or b) replaced by reflection. Both feels like too much, just to get the test report green.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
```diff
@@ -26,6 +27,12 @@
  * of the implementation class of the {@link org.junit.platform.engine.TestEngine}
  * interface is declared.
  *
+ * <p>Test execution listeners are discovered accordingly. For that purpose, a text
```
sbrannen: "accordingly" is insufficient here. 

Rather, we should mention that TELs which are discovered at runtime using the ServiceLoader mechanism will be automatically registered with the Launcher created by the factory but that clients of the Launcher may optionally register TELs manually in addition.

Not that you have to use my exact wording -- b/c I'd polish it myself -- but you get the gist.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
```diff
@@ -39,7 +46,11 @@
 	 * @throws PreconditionViolationException if no test engines are detected
```
sbrannen: The Javadoc must also mention that TELs which are dynamically detected will be automatically registered in the Launcher.

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
```diff
@@ -22,15 +22,36 @@
...
 import org.junit.platform.launcher.TestPlan;
+import org.junit.platform.launcher.listener.CustomTestExecutionListener;
 
 /**
  * @since 1.0
  */
 class LauncherFactoryTests {
 
 	@Test
+	void testCustomTestExecutionListenerServiceIsLoaded() {
```
sbrannen: Please don't use the "test" prefix for `@Test` methods, even if the class already has such naming in place. 😉

platform-tests/src/test/java/org/junit/platform/launcher/listener/CustomTestExecutionListener.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.engine.reporting.ReportEntry;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+import org.junit.platform.launcher.TestPlan;
+
+public class CustomTestExecutionListener implements TestExecutionListener {
+
+	public int testPlanExecutionStarted;
```
sbrannen: Please convert each `int` to an `AtomicInteger` to ensure that concurrent execution of listeners is supported from the start.

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
```diff
@@ -22,15 +22,36 @@
...
+		DefaultLauncher launcher = (DefaultLauncher) LauncherFactory.create();
+		TestExecutionListenerRegistry registry = launcher.getTestExecutionListenerRegistry();
+		List<TestExecutionListener> listeners = registry.getTestExecutionListeners();
+		assertThat(listeners).hasSize(2);
+		List<TestExecutionListener> one = listeners.stream().filter(
+			listener -> listener instanceof CustomTestExecutionListener).collect(toList());
+		assertThat(one).hasSize(1);
+		CustomTestExecutionListener listener = (CustomTestExecutionListener) one.get(0);
+		assertThat(listener.testPlanExecutionStarted).isEqualTo(0);
```
sbrannen: Asserting that each of these counts is zero is akin to asserting that a getter/setter method pair in a POJO work.

In other words, I don't see any added value in these assertions. However, if you were to assert the counters after an actual test run, that would be different. But I don't think the latter is truly necessary since we have other integration tests in place for such behavior in other TELs.
 
Now that I think about it, we don't really even need two different types of concrete TELs: they could both be a no-op TEL.

platform-tests/src/test/java/org/junit/platform/launcher/listener/CustomTestExecutionListener.java
sbrannen: Alternatively, you could just get rid of all of the counters since the current tests don't actually assert anything meaningful based on them.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
sormuras: Rewording the section _accordingly_ - but please, polish it. 😉

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
sormuras: Done.

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
sormuras: Renamed. Shall I rename the other test, too?

platform-tests/src/test/java/org/junit/platform/launcher/listener/CustomTestExecutionListener.java
sormuras: Removed the _counting_ listener.

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
sbrannen: sure

sbrannen:
This has been merged into `master`. Thanks!

PR #591, opened 2016-12-07
Add new TestDescriptors to existing parents

### LINKED ISSUE #589: Unable to discover 2 methods in the same class
IDEA allows users to run 2 test methods of the same test class. This is done now through `LauncherDiscoveryRequest`, like:

```java
        List<DiscoverySelector> selectors = new ArrayList<>();
          while ((line = reader.readLine()) != null) {
            selectors.add(createMethodSelector(line));
          }
          builder = builder.selectors(selectors);
          if (tagName != null && !tagName.isEmpty()) {
            builder = builder.filters(TagFilter.includeTags(tagName));
          }
          return builder.build();
```

Works well in most cases, but if both methods belong to the same class, then only the first one would be executed. After debugging it turns out that `org.junit.jupiter.engine.discovery.JavaElementsResolver#tryToResolveWithResolver` detects that there is already a corresponding class but instead of replacing newly created `ClassTestDescriptor` with old one, newly created descriptors are returned, and second method is added there and this subtree is lost.

marcphilipp:
This fixes #589.

@junit-team/junit-lambda Please take a look! I'm still astonished that we didn't discover (no pun intended) this bug until now.

---

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/JavaElementsResolver.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/JavaElementsResolver.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


sbrannen:
LGTM. Feel free to merge it.

Thanks for the quick fix!

And yeah... I imagine there are still a few of these types of _obvious_ bugs lurking about in the shadows. 😉

Issue #587, opened 2016-12-07
Rename test methods not to start with `test` literal


sormuras:
At the current state, there're 1.281 `@Test` annotation usages in JUnit 5 testing code.
70 of the annotated methods do start with `test` literal. Namely:
```
AssumptionsDemo
  testInAllEnvironments()
  testOnlyOnCiServer()
  testOnlyOnDeveloperWorkstation()

DisabledClassDemo
  testWillBeSkipped()

DisabledTestsDemo
  testWillBeExecuted()
  testWillBeSkipped()

DisplayNameDemo
  testWithDisplayNameContainingEmoji()
  testWithDisplayNameContainingSpaces()
  testWithDisplayNameContainingSpecialCharacters()

DynamicTestGenerationTests
  testFactoryMethodIsCorrectlyDiscoveredForMethodSelector()
  testFactoryMethodsAreCorrectlyDiscoveredForClassSelector()

ExceptionHandlingTests.FailureTestCase
  testWithCheckedException()
  testWithUncheckedException()

StandardTestClassTests
  testsFailWhenAfterEachFails()
  testsFailWhenBeforeEachFails()

StandardTestClassTests.TestCaseWithFailingAfter
  test1()

StandardTestClassTests.TestCaseWithFailingBefore
  test1()
  test2()

TestCaseWithInheritanceTests.TestCase3
  test3()

Class1WithTestCases
  test1()

Class2WithTestCases
  test2()

ClassWithStaticInnerTestCases.ShouldBeDiscovered
  test1()

ClassWithStaticInnerTestCases.ShouldNotBeDiscovered
  test2()

DiscoveryTests.LocalTestCase
  test1()
  test2()
  test3()
  test4()
  test4(TestInfo)

HerTestClass
    test7(String)

MyTestClass
    test1()
    test2()

OtherTestClass.NestedTestClass
   test5()
   test6()

TestCaseWithNesting
   testA()

TestCaseWithNesting.NestedTestCase
   testB()

TestCaseWithNesting.NestedTestCase.DoubleNestedTestCase
   testC()

YourTestClass
  test3()
  test4()

BeforeAndAfterEachTests
  testMethodThrowsAnException()

BeforeAndAfterEachTests.OuterTestCase
  testOuter()

BeforeAndAfterEachTests.OuterTestCase.InnerTestCase
  testInner()

BeforeAndAfterTestExecutionCallbackTests
  testMethodThrowsAnException()

[...]

PR #594, opened 2016-12-08
Remove test literal from method names.

sormuras:
## Overview

Solves #587 by renaming the 3 test methods.

---


sbrannen:
merged -- thanks!

Issue #297, opened 2016-06-07
Maintain list of third-party TestEngines and Extensions


marcphilipp:
For maintainers and users of the framework, we want to have a list of third-party engines and extensions in the wiki ~or on the website~.

## Test Engines
- [Specsy](http://specsy.org/)
- [Spek](https://github.com/JetBrains/spek)
- [Cucumber](https://github.com/signed/junit-cucumber-engine)

## Extensions
- [SpringExtension](https://github.com/sbrannen/spring-test-junit5)
- [upREST](https://github.com/selesy/uprest)
- [JUnit 5 Docker](https://faustxvi.github.io/junit5-docker/)
- [JGiven](https://github.com/TNG/JGiven/tree/junit5/jgiven-junit5)

sbrannen:
I think we should also maintain a list of third-party extensions.

Would you like to tack that onto this issue or create a separate one?

marcphilipp:
I've rephrased the title of the issue to also include extensions.

sbrannen:
Added Spek to the list of test engine.

jbduncan:
[JUnit Io](https://github.com/CodeFX-org/junit-io) should probably be added to this list. :)

FaustXVI:
I'd love to see the [docker extension](https://github.com/FaustXVI/junit5-docker) added to this list :)

sbrannen:
I added JUnit 5 Docker to this issue's description.

sbrannen:
Added link to JGiven

sbrannen:
Added link to junit-cucumber-engine

marcphilipp:
_in progress_

marcphilipp:
I've restructured the wiki a bit, added a custom sidebar, and made it editable by any GitHub user.

marcphilipp:
The new page documenting third-party extensions is available at https://github.com/junit-team/junit5/wiki/Third-party-Extensions.

jbduncan:
I'm not sure meeting if editable by any GitHub user is such a good idea.

When I went to try to edit the wiki just now, i found that I had permission to **delete** pages. That sounds rather dangerous to me.

Does GitHub have a mechanism to force wiki changes to be reviewed by someone on the JUnit team before they can be pushed through?

marcphilipp:
> I'm not sure if making the wiki editable by any GitHub user is such a good idea.
> 
[...]

PR #595, opened 2016-12-08
Add reference to docker extension

FaustXVI:
## Overview

Add a link to an extension that allow the user to start docker containers from their tests.

JUnit5-Docker has just been pushed to maven contral as a release candidate for version 1.0.0.
It only implements one feature : start one container from your test. The idea is to have something that works early. More features to come.

I think it will be a very useful extension and adding the reference to the documentation would show that an extension can do something else than integrate with another framework.

---

### DIFF
--- a/documentation/src/docs/asciidoc/extensions.adoc
+++ b/documentation/src/docs/asciidoc/extensions.adoc
@@ -140,8 +140,8 @@ process_ test instances.
 Common use cases include injecting dependencies into the test instance, invoking custom
 initialization methods on the test instance, etc.
 
-For concrete examples, consult the source code for the `{MockitoExtension}` and the
-`{SpringExtension}`.
+For concrete examples, consult the source code for the `{MockitoExtension}`, `{SpringExtension}` and the
+`{DockerExtension}`.
 
 [[extensions-parameter-resolution]]
 === Parameter Resolution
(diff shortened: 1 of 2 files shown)


jbduncan:
Have you considered making your proposal on [junit-io](https://github.com/CodeFX-org/junit-io)? :)

nicolaiparlog:
@FaustXVI The Docker extension looks pretty cool, great idea!

I started [JUnit Io](https://github.com/CodeFX-org/junit-io) to collect small(ish) extensions and breed ideas but didn't get very far yet. Your extension seems to be pretty well-defined and self-contained, so it makes sense to be its own library. But if you see value in merging into another project, I'm happy to discuss that.

FaustXVI:
@nicolaiparlog  That's a great initiative !
For the moment, I do agree that, since we have a quite well-defined scope it totally make sens to keep it separated.
Nonetheless, I keep the idea in mind, maybe I'll change my mind someday :)

smoyer64:
@FaustXVI 

This seems like one of those situations where @BeforeSuite/@AfterSuite would be ideal when you want the container to be used for more than one class.  In any case, it looks awesome.

FaustXVI:
Using `@BeforeSuite` and `@AfterSuite` is great idea ! I'll add it to my issues.
I'll definitely look at #297 , thanks !

sbrannen:
closing this issue since it will be addressed in #297.

PR #597, opened 2016-12-09
Use functional expressions for Optionals in JUnitPlatformProvider

kamilszymanski:

### DIFF
--- a/junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
+++ b/junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
@@ -134,15 +134,11 @@ public class JUnitPlatformProvider extends AbstractProvider {
 
 		Optional<List<String>> includes = getGroupsOrTags(getPropertiesList(INCLUDE_GROUPS),
 			getPropertiesList(INCLUDE_TAGS));
-		if (includes.isPresent()) {
-			filters.add(TagFilter.includeTags(includes.get()));
-		}
+		includes.map(TagFilter::includeTags).ifPresent(filters::add);
 
 		Optional<List<String>> excludes = getGroupsOrTags(getPropertiesList(EXCLUDE_GROUPS),
 			getPropertiesList(EXCLUDE_TAGS));
-		if (excludes.isPresent()) {
-			filters.add(TagFilter.excludeTags(excludes.get()));
-		}
+		excludes.map(TagFilter::excludeTags).ifPresent(filters::add);
 
 		return filters.toArray(new Filter<?>[filters.size()]);
 	}


junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
```diff
@@ -134,15 +134,11 @@ private void invokeSingleClass(Class<?> testClass, RunListener runListener) {
 
 		Optional<List<String>> includes = getGroupsOrTags(getPropertiesList(INCLUDE_GROUPS),
 			getPropertiesList(INCLUDE_TAGS));
-		if (includes.isPresent()) {
-			filters.add(TagFilter.includeTags(includes.get()));
-		}
+		includes.ifPresent(includeFilters -> filters.add(TagFilter.includeTags(includeFilters)));
```
sbrannen: Please rename `includeFilters` to `tags`.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
```diff
@@ -134,15 +134,11 @@ private void invokeSingleClass(Class<?> testClass, RunListener runListener) {
...
 
 		Optional<List<String>> excludes = getGroupsOrTags(getPropertiesList(EXCLUDE_GROUPS),
 			getPropertiesList(EXCLUDE_TAGS));
-		if (excludes.isPresent()) {
-			filters.add(TagFilter.excludeTags(excludes.get()));
-		}
+		excludes.ifPresent(excludeFilters -> filters.add(TagFilter.excludeTags(excludeFilters)));
```
sbrannen: Please rename `excludeFilters` to `tags`.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
smoyer64: When I originally called this variable `includeFilters` it was because the engine filters would be added at the same point.  Note that we decided that the code to include or exclude tags should be committed without waiting for the engine filter implementation.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
```diff
@@ -134,15 +134,11 @@ private void invokeSingleClass(Class<?> testClass, RunListener runListener) {
 
 		Optional<List<String>> includes = getGroupsOrTags(getPropertiesList(INCLUDE_GROUPS),
 			getPropertiesList(INCLUDE_TAGS));
-		if (includes.isPresent()) {
-			filters.add(TagFilter.includeTags(includes.get()));
-		}
+		includes.ifPresent(tags -> filters.add(TagFilter.includeTags(tags)));
```
marcphilipp: How about the following?

```java
includes.map(TagFilter::includeTags).ifPresent(filters::add);
```

sbrannen: even better 👍

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
sbrannen: @smoyer64, maybe I'm missing something, but it doesn't appear that `includeFilters` is from you. 😉

sbrannen:
@kamilszymanski, do you wish to make the latest changes recommended by @marcphilipp, or shall we merge your PR and make those minor changes afterwards?

kamilszymanski:
refactored to solution suggested by @marcphilipp

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java
kamilszymanski: :+1:

marcphilipp:
Thanks! 👍

PR #598, opened 2016-12-11
Prevent Java 9 on Travis CI from failing

sormuras:
## Overview

Assume that two tests using `Class::getPackage::getXyz()` are not executed on Java 9 -- until the reason why those methods return `null` is understood and fixed. Create new ticket to track this issue?

---

### DIFF
--- a/platform-tests/src/test/java/org/junit/platform/commons/util/PackageUtilsTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/commons/util/PackageUtilsTests.java
@@ -14,6 +14,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeFalse;
 
 import java.util.function.Supplier;
 
@@ -49,12 +50,14 @@ class PackageUtilsTests {
 
 	@Test
 	void vendorFromObjectClassIsPresent() {
+		assumeFalse(System.getProperty("java.version").startsWith("9"), "Java 9 not supported, yet");
 		assertTrue(PackageUtils.getAttribute(Object.class, Package::getSpecificationVendor).isPresent());
 		assertTrue(PackageUtils.getAttribute(Object.class, Package::getImplementationVendor).isPresent());
 	}
 
 	@Test
 	void versionSystemPropertyEqualsRuntimeClassImplementationVersion() {
+		assumeFalse(System.getProperty("java.version").startsWith("9"), "Java 9 not supported, yet");
 		Supplier<AssertionError> error = () -> new AssertionError("implementation version not available");
 		String actual = PackageUtils.getAttribute(Runtime.class, Package::getImplementationVersion).orElseThrow(error);
 		assertEquals(System.getProperty("java.version"), actual);


platform-tests/src/test/java/org/junit/platform/commons/util/PackageUtilsTests.java
```diff
@@ -49,12 +50,14 @@ void getAttributeWithFunctionReturningNullIsEmpty() {
 
 	@Test
 	void vendorFromObjectClassIsPresent() {
+		assumeFalse(System.getProperty("java.version").equals("9-ea"), "Java 9 not supported, yet");
```
sbrannen: Let's not check that it is equal to `9-ea` but rather that it starts with `9`.

platform-tests/src/test/java/org/junit/platform/commons/util/PackageUtilsTests.java
```diff
@@ -49,12 +50,14 @@ void getAttributeWithFunctionReturningNullIsEmpty() {
...
 		assertTrue(PackageUtils.getAttribute(Object.class, Package::getSpecificationVendor).isPresent());
 		assertTrue(PackageUtils.getAttribute(Object.class, Package::getImplementationVendor).isPresent());
 	}
 
 	@Test
 	void versionSystemPropertyEqualsRuntimeClassImplementationVersion() {
+		assumeFalse(System.getProperty("java.version").equals("9-ea"), "Java 9 not supported, yet");
```
sbrannen: ditto

sormuras:
Changes force pushed.

sormuras:
Back to three green CI lights.

sbrannen:
FYI: `getPackage()` is deprecated on JDK 9 in favor of [`getDefinedPackage()`](http://download.java.net/java/jdk9/docs/api/java/lang/ClassLoader.html#getDefinedPackage-java.lang.String-).

So perhaps that's the reason. 

But... `getDefinedPackage()` is not available until JDK 9.

sbrannen:
Here's an excerpt from the [JDK 9 mailing list](http://jigsaw-dev.openjdk.java.narkive.com/5zkdq3ON/re-examine-classloader-getpackage-s-methods):

> I considered deprecating `ClassLoader::getPackage(s)`. These methods are not doing anything inherently wrong or causing any risk although the spec and impl version, title, vendor, annotation might not be the expected value. The Package design and implementation has been flawed since 1.2. This patch leaves it with `@apiNote` to make it clear.

sbrannen:
Oops... I linked to documentation about `ClassLoader` instead of `Class`. So `Class::getPackage()` is not deprecated in JDK 9; however, based on the linked discussion, it appears that package metadata still might suffer from the same issues as with `ClassLoader`.

sbrannen:
And the Javadoc for `ClassLoader` states the following:

> A `Package` can be obtained with the `Package.getPackage(String)` and `ClassLoader.getDefinedPackage(String)` methods. Every `Package` defined by a class loader can be obtained with the `Package.getPackages()` and `ClassLoader.getDefinedPackages()` methods.

sbrannen:
And the following from the Javadoc for `Package` in JDK 9 seems to hit the nail on the head:

>  A Package automatically defined for classes in a named module has the following properties: ... The specification and implementation titles, versions, and vendors are unspecified.

So, since your test works with `Object` which is in the `java.base` module, I'm assuming the above applies, which means that the _implementation titles, versions, and vendors are unspecified_.

sbrannen:
> Create new ticket to track this issue?

Yes, we should create a new ticket to address the issue with package versioning metadata on JDK 9.

sbrannen:
Merged into `master`.

Thanks

sbrannen:
Created separate issue to address versioning metadata from JARs on Java 9 in #600.

PR #599, opened 2016-12-11
Regression: assertThrows should not use wildcard for expected exception type

DmitriyZaitsev:
For this case, instead of using a wildcard declared as `? extends Throwable`, we use a type T as the type argument of a given generic type.
It will help compiler to infer correct concrete return type for method instead of returning `Throwable` every time.

For example, in IDEA, press *Cmd+Option+V*
Before:
```java
Throwable t = assertThrows(AwesomeException.class, () -> { /* smth */ });
```

After:
```java
AwesomeException e = assertThrows(AwesomeException.class, () -> { /* smth */ });
```

---


sbrannen:
This is a regression I accidentally introduced in #531.

I have therefore updated the title of this issue accordingly.

sbrannen:

Thanks for bringing it to our attention.

PR #601, opened 2016-12-12
Use class from external JAR file for testing

### LINKED ISSUE #600: PackageUtils do not work with modules on Java 9
## Background

Invoking `clazz.getPackage().getImplementationVersion()` can currently return `null` on Java 9 while returning a valid implementation version on Java 8. See the discussion in #598 for details.

For example `PackageUtils.getAttribute(Object.class, Package::getImplementationVersion)` returns an empty `Optional` on Java 9 when passed a class that resides in a module (e.g., `java.base`).

## Analysis

Package versioning is **NOT** supported for _Modular JAR Files_ (i.e., JARs containing `module-info.class`).

Excerpt from the [JAR File Specification](http://download.java.net/java/jdk9/docs/technotes/guides/jar/jar.html#Modular) for Java 9:

> JAR files that are deployed as modules declare (explicitly or implicitly) their dependencies and may declare service providers too. The JAR Manifest is ignored. On the other hand, JAR files on the class path use the META-INF directory for security, versioning, or service configuration. 

## Deliverables

  - [x] Determine if classes such as `Object` which reside in the `java.base` module even have such metadata associated with their packages.
 - See _Analysis_ section above.
- [ ] Find an alternative to `PackageUtils.getAttribute(...)` for module versioning on Java 9.

sormuras:
## Overview

This PR replaces `vendorFromObjectClassIsPresent()` with `attributesFromValueWrapperClassArePresent()`  which works on Java 8 and 9, as the `ValueWrapper` class is loaded from an external JAR with all its metadata initialized from the encapsulated manifest file.

The second test, that was guarded by a _do not start on Java 9_ assumption, is removed as it didn't provide real additional test value.

Might resolve #600

---

### DIFF
--- a/platform-tests/src/test/java/org/junit/platform/commons/util/PackageUtilsTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/commons/util/PackageUtilsTests.java
@@ -14,11 +14,17 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assumptions.assumeFalse;
+import static org.junit.jupiter.api.DynamicTest.dynamicTest;
 
-import java.util.function.Supplier;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Function;
 
+import org.junit.jupiter.api.DynamicTest;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


sbrannen:
@sormuras, thanks for the PR.

I like this version of the tests better!

However, it doesn't resolve #600. The reason this works is that the `opentest4j.jar` is loaded via the class-path, not the module-path. Thus, the JAR manifest versioning information is available for `opentest4j.jar`. So, we still have the issue with not being able to read such versioning information from _Modular JAR Files_ loaded via the module-path on Java 9.

In any case, I'll go ahead and merge this PR.

PR #626, opened 2017-01-10
Updated to Spotless 3.0.0.

nedtwigg:
- added removeUnusedImports() rule
- removed lamda fix (no longer needed)
- changed groovy target to `src/**/*.groovy`,
  because it is faster to resolve than `**/*.groovy`.
- renamed `customReplaceRegex` to its new name
  `replaceRegex`.

Time for a `spotlessCheck` to execute before this commit: 8.4s config, 5s execution (total 13.4s)
Time for a `spotlessCheck` to execute after this commit: 5.5s config, 0.3s execution (total 5.8s).
So about twice as fast.

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
 		maven { url 'https://oss.sonatype.org/content/repositories/snapshots' }
 	}
 	dependencies {
-		classpath 'com.diffplug.gradle.spotless:spotless:2.4.0'
+		classpath 'com.diffplug.spotless:spotless-plugin-gradle:3.0.0'
 		classpath 'org.ajoberstar:gradle-git:1.6.0'
 		classpath 'org.junit.platform:junit-platform-gradle-plugin:1.0.0-SNAPSHOT'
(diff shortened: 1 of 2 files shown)


jbduncan:
Whoops! Appveyor does _not_ like that.

Looks like it's the same bug as what we tried to workaround [here](https://github.com/diffplug/spotless/commit/a64e831138ce9045dc94d182e7afe29d0073af4f) @nedtwigg. -_-

I'll get around to submitting a bug report to google-java-format at some point...

nedtwigg:
In the meantime, if we get rid of the "removeUnusedImports()", we keep all the same functionality as before, but we don't bump into this bug.

sbrannen:
@nedtwigg, thanks for taking care of the upgrade to Spotless 3.0!

Too bad about the new `removeUnusedImports` feature not working on Windows. That's a pity, but it will come with time...

sbrannen:
This has been merged into`master`.

sormuras:
As soon as GJF 1.2 is released - or switch to GJF-SNAPSHOT, it already contains the fix: https://github.com/google/google-java-format/issues/112#issuecomment-271660475

Issue #314, opened 2016-06-18
Introduce support for excluding classes via filters


sbrannen:
## Status Quo

It is currently possible to _filter_ a class for _inclusion_ but **not** for _exclusion_.

## Related Issues
- #301
- #312 
- #313 
- #474

## Deliverables
- [x] Introduce _exclude_ support in `ClassNameFilter` (e.g., `excludeClassNamePatterns()`)
- [x] Introduce _exclude_ filter support for classes in the `ConsoleLauncher` (e.g., `-N` and `--exclude-classname`).
- [x] Introduce _exclude_ filter support for classes in the Gradle JUnit Platform plugin (e.g., `excludeClassNamePattern()` and `excludeClassNamePatterns()`).
- [x] Introduce _exclude_ filter support for classes in the `JUnitPlatform` runner (e.g., `@ExcludeClassNamePatterns`).

mmerdes:
in progress

sbrannen:
FYI: I updated the deliverables with regard to plurality.

marcphilipp:
Remainder of work is addressed in #644.

PR #630, opened 2017-01-18
Issue314 exclude classname patterns

mmerdes:
## Overview

This PR addresses issue #314.  
It adds support for filtering classes for exclusion in the ConsoleLauncher, the Gradle JUnit 5 plugin, and the JUnitPlatform runner.

---

### DIFF
--- a/junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/FiltersExtension.groovy
+++ b/junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/FiltersExtension.groovy
@@ -30,8 +30,22 @@ class FiltersExtension {
 	 */
 	List<String> includeClassNamePatterns
 
+	/**
+	 * List of class name patterns in the form of regular expressions for
+	 * classes that should be <em>excluded</em> from the test plan.
+	 *
+	 * <p>The patterns are combined using OR semantics, i.e. if the fully
+	 * qualified name of a class matches against at least one of the patterns,
+	 * the class will be excluded from the test plan.
+	 */
+	List<String> excludeClassNamePatterns
+
 	protected List<String> getIncludeClassNamePatterns() {
 		return includeClassNamePatterns ?: [ ClassNameFilter.STANDARD_INCLUDE_PATTERN ];
+	}
+
+	protected List<String> getExcludeClassNamePatterns() {
+		return excludeClassNamePatterns
 	}
 
 	/**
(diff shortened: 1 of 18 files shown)


sormuras:
@mmerdes The "running tests" documentation of this new [-N option](https://github.com/junit-team/junit5/blob/master/documentation/src/docs/asciidoc/running-tests.adoc#options) is missing, afaics.

junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/FiltersExtension.groovy
```diff
@@ -51,4 +65,21 @@ class FiltersExtension {
...
 
+	/**
+	 * Add a pattern to the list of <em>excluded</em> patterns.
+	 */
+	void excludeClassNamePattern(String pattern) {
+		excludeClassNamePatterns(pattern)
+	}
+
+	/**
+	 * Add patterns to the list of <em>included</em> patterns.
```
sbrannen: That should say _excluded_. 😉

junit-platform-runner/src/main/java/org/junit/platform/runner/JUnitPlatform.java
```diff
@@ -255,6 +266,19 @@ private void addExcludedEnginesFilter(LauncherDiscoveryRequestBuilder requestBui
 		return patterns;
 	}
 
+	private String[] getExcludeClassNamePatterns() {
+		String[] patterns = getValueFromAnnotation(ExcludeClassNamePatterns.class, ExcludeClassNamePatterns::value,
+			new String[0]);
+
+		if (patterns.length == 0) {
+			return new String[] {};
```
sbrannen: No need to create a new array here. That was done in the previous line. So just return `patterns` if its length is `0`.

junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/FiltersExtension.groovy
sbrannen: Though... TBH, you did a great job of finding all other occurrences of _included_ and replacing them. 👍

junit-platform-runner/src/main/java/org/junit/platform/runner/JUnitPlatform.java
mmerdes: True enough, thanks for pointing out. On second thoughts I decided to drop the check for zero length altogether. The following code is robust with respect to zero length arrays and the resulting code is simpler.

mmerdes:
@sormuras 
Yes, indeed. Thanks for the hint. Will be fixed.

marcphilipp:
Looks good! Please remember to update the documentation of `ConsoleLauncher` in the User Guide and add an entry to the Release Notes. 🙂

sormuras:
> Looks good! Please remember to update the documentation of ConsoleLauncher in the User Guide and add an entry to the Release Notes. 🙂

🎉

PR #634, opened 2017-01-19
Add details parsing fix and tests.

sormuras:
## Overview

PR #524 pretended to parse option details case insensitive -- but it did not. This PR fixes that issue and provides tests ensuring all of case-wise permutations from `NONE` to `none` are allowed and valid for `--details`.

---


sbrannen:
Thanks!

PR #639, opened 2017-01-25
Make Gradle incremental build work

### LINKED ISSUE #631: Task inputs for junitPlatformTest are broken
## Status Quo

Executing `gradlew clean test; gradlew test; gradlew test` results in **all** `junitPlatformTest` tasks being executed **thrice** instead of just once. In other words, the _task inputs_ for the task created by the `JUnitPlatformPlugin` no longer work, thus rendering incremental builds with Gradle not so very useful.

## Deliverables

- [ ] Revise task inputs for `junitPlatformTest` so that incremental builds work once again.

sormuras:
## Overview

Executing `gradlew clean test; gradlew test; gradlew test` resulted in
all tests being executed thrice instead of just once. By generating the
shadowed `junit-platform-console-x.y.z.jar` only if code changes are
made, Gradles' UP-TO-DATE checks succeed again.

Copied from `build.gradle` file:
```
 // Generate JAR manifest only if code was compiled or recompiled;
 // otherwise the junitPlatformTest task will always be executed even if
 // no code changes were made. The reason is that the generation of
 // the buildDate and buildTime causes JAR manifests to be modified
 // which triggers unnecessary rebuilding of the dependent JARs.
```

Fixes #631

---

### DIFF
--- a/junit-platform-console/junit-platform-console.gradle
+++ b/junit-platform-console/junit-platform-console.gradle
@@ -42,6 +42,11 @@ javadoc {
 jar.enabled = false
 
 shadowJar {
+	// Generate shadow jar only if the underlying manifest was regenerated.
+	// See https://github.com/junit-team/junit5/issues/631
+	onlyIf {
+		project.generateManifest
+	}
 	classifier = null
 	configurations = [project.configurations.shadowed]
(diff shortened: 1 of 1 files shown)


sbrannen:

Thanks!

Issue #416, opened 2016-07-21
Support @ExtendWith on fields


slu-it:
## Overview

**Is it planned for JUnit 5 to discover composed `@ExtendWith` annotations on fields?**

As far as I can see (from Milestone 1) they are only discovered in case they are used on types or methods.

For projects like Mockito it would be useful to also discover extension by analyzing field annotations.

**Example:**

``` java
// The following explicit declaration would become unnecessary.
// @ExtendWith(MockitoExtension.class)
public class MyServiceTest {

    @Mock
    FooService fooService;

    @InjectMocks
    MyService cut;

    @Test
    void testSomething() {
        ....
    }

}
```

In case `@Mock` and `@InjectMocks` were to be changed to composed annotations declaring `@ExtendWith(MockitoExtension.class)`, they could simply be used without having to declare the extension for each test class.

## Related Issues

- #497
- #506 
- #864 

## Deliverables

- [ ] Support registration of extensions on fields with the desired extension classes supplied via `@ExtendWith(...)`.
  - This is analogous to the current support for declaring `@ExtendWith` on classes and methods.

sbrannen:
Discovery of extensions on fields is something we have in fact considered, but the feature is not currently planned. So I'll put it in the backlog now.

Thanks for raising the issue.

sbrannen:
Overhauled this issue's description by introducing _Status Quo_, _Proposal_, and _Deliverables_ sections with a new focus on _programmatic_ extension registration via fields.

sbrannen:
Currently slated for M3.

sbrannen:
**Update**

The description of this issue has been returned to its original focus.

Support for registering extensions programmatically will be addressed in #497.

sbrannen:
@slu-it, in light of #497, is this issue still relevant?

slu-it:
Hi @sbrannen,

I think it is still relevant (as in not resolved) because #497 does not address the described usecase (`@Mock` without having to declare the corresponding `MockitoExtension`). In cases like that, the annotated field is not itself an extension implementation but rather something the extension should act upon.

But since I did not feel much pain regarding this in the past years, I can't speak to how important it really is. I just would find it useful.

sbrannen:
> In cases like that, the annotated field is not itself an extension implementation but rather something the extension should act upon.

Indeed, that is something different. Sorry for overlooking that.

In any case, I am closing this as superseded by #1836.

If you feel that is inappropriate, please provide further feedback.

[...]

PR #642, opened 2017-01-28
Introduce @TestTemplate and accompanying extension point

marcphilipp:
## Overview

This pull request introduces the concept of `@TestTemplate` methods. In contrast to `@Test` methods, a test template is not itself a test case but rather a template for test cases. As such, it is designed to be invoked multiple times depending on the number of `TestTemplateInvocationContexts` returned by the registered `TestTemplateInvocationContextProviders`.

Each invocation of a test template method, behaves like the execution of a regular @Test method, i.e. it supports the same lifecycle callbacks and extensions. Each context may provide a custom display name and provide additional extensions that will only be registered for the current invocation, e.g. a `ParameterResolver`, a `TestInstancePostProcessor` or both.

This extension point is a prerequisite for implementing support for parameterized tests as well as repeated tests.

---

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestTemplate.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestTemplate.java
@@ -33,6 +33,10 @@ import org.junit.platform.commons.meta.API;
  * org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider
  * providers}.
  *
+ * <p>Each invocation of a test template method, behaves like the execution of
+ * a regular {@link Test @Test} method, i.e. it supports the same lifecycle
+ * callbacks and extensions.
+ *
  * <p>{@code @TestTemplate} methods must not be {@code private} or {@code static}
  * and must return {@code void}.
  *
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestTemplate.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestTemplate.java
@@ -31,7 +31,8 @@ import org.junit.platform.commons.meta.API;
  * org.junit.jupiter.api.extension.TestTemplateInvocationContext invocation
  * contexts} returned by the registered {@linkplain
  * org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider
- * providers}.
+ * providers}. Must be used together with at least one provider. Otherwise,
+ * execution will fail.
  *
  * <p>Each invocation of a test template method, behaves like the execution of
  * a regular {@link Test @Test} method, i.e. it supports the same lifecycle
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestTemplate.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestTemplate.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.jupiter.api;
+
+import static org.junit.platform.commons.meta.API.Usage.Experimental;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.platform.commons.annotation.Testable;
+import org.junit.platform.commons.meta.API;
+
+/**
+ * {@code @TestTemplate} is used to signal that the annotated method is a
+ * <em>test template</em> method.
+ *
+ * <p>In contrast to {@link Test @Test} methods, a test template is not itself
+ * a test case but rather a template for test cases.
+ *
+ * <p>{@code @TestTemplate} methods must not be {@code private} or {@code static}
+ * and must return {@code void}.
+ *
+ * <p>{@code @TestTemplate} methods may optionally declare parameters to be
+ * resolved by {@link org.junit.jupiter.api.extension.ParameterResolver
+ * ParameterResolvers}.
+ *
+ * @since 5.0
+ * @see Test
+ */
+@Target({ ElementType.ANNOTATION_TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@API(Experimental)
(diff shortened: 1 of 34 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestTemplate.java
```diff
@@ -0,0 +1,58 @@
...
+ * <em>test template</em> method.
+ *
+ * <p>In contrast to {@link Test @Test} methods, a test template is not itself
+ * a test case but rather a template for test cases. As such, it is designed to
+ * be invoked multiple times depending on the number of {@linkplain
+ * org.junit.jupiter.api.extension.TestTemplateInvocationContext invocation
+ * contexts} returned by the registered {@linkplain
+ * org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider
+ * providers}.
+ *
```
mmerdes: I would add something like:
Must be used together with at least one TestTemplateInvocationContextProvider

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/discovery/DiscoveryTests.java
```diff
@@ -105,6 +107,24 @@ public void discoverCompositeSpec() {
...
+		LauncherDiscoveryRequest spec = request().selectors(
+			selectUniqueId(uniqueIdForTestTemplateMethod(TestTemplateClass.class, "testTemplate()"))).build();
+
+		TestDescriptor engineDescriptor = discoverTests(spec);
+		assertEquals(2, engineDescriptor.getDescendants().size(), "# resolved test descriptors");
+	}
+
+	@Test
+	public void discoverTestTemplateMethodByMethodReference() {
+		LauncherDiscoveryRequest spec = request().selectors(
```
mmerdes: Maybe discoverTestTemplateMethodByClassReferenceAndMethodName() would be a better name?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java
```diff
@@ -0,0 +1,162 @@
...
+import org.opentest4j.TestAbortedException;
+
+/**
+ * {@link TestDescriptor} for {@link org.junit.jupiter.api.TestTemplate @TestTemplate}
+ * methods.
+ *
+ * @since 5.0
+ */
+@API(Internal)
+public class TestTemplateTestDescriptor extends JupiterTestDescriptor {
```
mmerdes: There is a little of redundancy with class `MethodTestDescriptor`. Should we introduce a helper class or common super class?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java
```diff
@@ -0,0 +1,162 @@
...
+
+	private final Class<?> testClass;
+	private final Method templateMethod;
+
+	public TestTemplateTestDescriptor(UniqueId uniqueId, Class<?> testClass, Method templateMethod) {
+		super(uniqueId, determineDisplayName(Preconditions.notNull(templateMethod, "Method must not be null"),
+			MethodTestDescriptor::generateDefaultDisplayName));
+
```
mmerdes: Does the method reference to `generateDefaultDisplayName` show slight symptoms of 'feature envy'? 
Again, a common helper structure could be useful.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContext.java
```diff
@@ -0,0 +1,65 @@
...
+	 *
+	 * <p>The extensions provided by this method will only be used for this
+	 * invocation of the test template. Thus, it does not make sense to return
+	 * an extension that acts solely on the container level (e.g.
+	 * {@link BeforeAllCallback}).
+	 *
+	 * @return the additional extensions for this invocation; never {@code null}
+	 * but potentially empty
+	 */
+	default List<Extension> getAdditionalExtensions() {
```
mmerdes: Here the list of additional extensions must be provided programmatically.
While this is very powerful and flexible it differs a lot from the usual declarative application via `ExtendWith`.
Any ideas whether these additional extensions could be provided declaratively for TestTemplates as well? Would this be desirable or even feasible at all? Would would be the pros and cons?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestTemplate.java
marcphilipp: Point taken and addressed in 0784682c1bfc6fd9350f21a3fe780071fbb1092c.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/discovery/DiscoveryTests.java
marcphilipp: You're right. Done in 4edfd37cfa33b9082e948ea9c8d2bf16872e37d2.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContext.java
marcphilipp: I think all use cases need to share some state which is done easiest via extensions that store that state, either as instance variable or via a closure. In addition, we've already been discussing adding support for extensions declared in fields in #416. So, I think it's okay this way in this case.

marcphilipp:
> Conceptually a TestTemplate executed once and a normal @Test method are similar;
In other words a normal test could be viewed just as a special case of a TestTemplate.
Should this be reflected more fully in the code?

I think it's already strongly reflected in the code: `TestTemplateInvocationTestDescriptor` extends `MethodTestDescriptor` and inherits all the execution logic that we use for test methods.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java
marcphilipp: Feature envy and duplication have been addressed in 628f14b03b5382055eace6e79e790e24396e80cf.

marcphilipp:
@junit-team/junit-lambda I've followed most of @mmerdes' suggestions and added a few comments. Please take another look!

smoyer64:
There's no way to simply "thumbs-up" a commit on Github ... here's a manual one:

THUMBS UP

marcphilipp:
Feels good to finally have merged this. 😊

PR #645, opened 2017-02-07
Replace sout instruction with assertion

sormuras:
## Overview

Replace `System.out.println` instruction with actual assertion.

---


marcphilipp:
Thanks!

PR #647, opened 2017-02-11
Support bridged methods

sormuras:
## Overview
In preparation of fixing #333, convert `testBridge()` https://github.com/junit-team/junit4/pull/1413 test from JUnit 4 and disable it because the current Jupiter implementation produces this call sequence:
```
() -> assertEquals("static parent.beforeAll()", bridgeMethodNameSequence.get(0)),
() -> assertEquals("parent.beforeEach()", bridgeMethodNameSequence.get(1)),
() -> assertEquals("child.anotherBeforeEach()", bridgeMethodNameSequence.get(2)),
() -> assertEquals("child.test()", bridgeMethodNameSequence.get(3)),
() -> assertEquals("parent.afterEach()", bridgeMethodNameSequence.get(4)),  // should be 5
() -> assertEquals("child.anotherAfterEach()", bridgeMethodNameSequence.get(5)), // should be 4
() -> assertEquals("static parent.afterAll()", bridgeMethodNameSequence.get(6))
```
instead of the expected:
```
() -> assertEquals("static parent.beforeAll()", bridgeMethodNameSequence.get(0)),
() -> assertEquals("parent.beforeEach()", bridgeMethodNameSequence.get(1)),
() -> assertEquals("child.anotherBeforeEach()", bridgeMethodNameSequence.get(2)),
() -> assertEquals("child.test()", bridgeMethodNameSequence.get(3)),
() -> assertEquals("child.anotherAfterEach()", bridgeMethodNameSequence.get(4)),
() -> assertEquals("parent.afterEach()", bridgeMethodNameSequence.get(5)),
() -> assertEquals("static parent.afterAll()", bridgeMethodNameSequence.get(6))
```

- [x] Analyze current implementation regarding bridge methods - **ignore** them in the result set of `ReflectionUtils.findAllMethodsInHierarchy(Class<?>, MethodSortOrder)`
- [x] Ensure the Jupiter engine behaves correctly for bridge methods

---

I hereby agree to the terms of the JUnit Contributor License Agreement.

PR #649, opened 2017-02-12
Add integration test

### LINKED ISSUE #648: Add integration test to ensure Jupiter engine correctly runs tests
Most of the tests are written in JUnit5-style. This is good in the sense that we want to try out our APIs, but bad because regressions in the framework can cause failing tests to pass or can filter tests entirely. One example was in the initial version of #647 where an incorrect usage of a Stream's `filter()` method removed all tests from being run.

We should have some kind of integration test that runs some sample tests (which includes failing and passing tests), reads the output, and ensures that we get the expected result. This integration test should not be written as a JUnit5-style test (for obvious reasons).

Some initial discussion for this is in the comments for #647

sormuras:
## Overview

Fixes #648 by running some sample tests (which includes failing and passing tests)
reads the output, and ensures that we get the expected result. This integration test
relies on external test plan execution using the standalone jar.

When the something similar to the incorrect usage of a Stream's filter() method
that removed all tests from being run happens, this integration test will fail.

---

### DIFF
--- a/junit-platform-console/junit-platform-console.gradle
+++ b/junit-platform-console/junit-platform-console.gradle
@@ -17,6 +17,9 @@ configurations {
 
 dependencies {
 	compile(project(':junit-platform-launcher'))
+	testCompile(project(':junit-jupiter-api'))
+	testCompile("junit:junit:${junit4Version}")
+
 	shadowed('net.sf.jopt-simple:jopt-simple:5.0.3')
 
 	standalone(project(':junit-platform-console'))
@@ -51,6 +54,7 @@ javadoc {
 }
 
 jar.enabled = false
+test.enabled = false // prevent supposed-to-fail integration tests from failing the build
 
 shadowJar {
 	// Generate shadow jar only if the underlying manifest was regenerated.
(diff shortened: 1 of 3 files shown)


sormuras:
And yes, this integration yields 🔴 when you remove all test methods via https://github.com/junit-team/junit5/pull/647#discussion_r100697450

sormuras:
When you add a failing test, the program error code results in a broken build. May need to fork the VM within the standaloneExec task. Or something else.

marcphilipp:
Yep, `ignoreExecValue` is also available for `JavaExec`:
https://docs.gradle.org/current/dsl/org.gradle.api.tasks.JavaExec.html#org.gradle.api.tasks.JavaExec:ignoreExitValue

sormuras:
Works perfectly for failing Jupiter test, but a failing JUnit 4 test breaks the build. See next forced push.

marcphilipp:
Ahh, the standard test task discovers and executes JUnit 4 tests...

Adding `test.enabled = false` to the Gradle file should help.

sormuras:
> Adding test.enabled = false to the Gradle file should help.

It does.

Issue #352, opened 2016-06-24
Make reflection, annotation, and scanning utilities available to TestEngine authors


jlink:
Most 3rd party engines will have to handle reflection and annotations and do classpath scanning. If you want engines to be consisten in how they do it and how they handle corner cases, make the important stuff of junit-platform-commons, where everything is API(Internal), available, e.g. in `org.junit.platform.engine.support`

raniejade:
I found myself depending on those internal APIs as well. I'm just being lazy and don't want to reinvent the wheel. :)

sbrannen:
This is closely related to #246.

These issues should therefore be addressed together in order to come up with a consistent, maintainable means to provide such support to third-party developers.

sormuras:
PR #651 adds an interface named `Util` to `ExtensionContext` supplying **extension** authors access to utility methods. That can easily be widend to a general first-class citizen utility API that can be used by both, extension and engine authors.

sormuras:
Two possible solutions:
1. Mark parts of `org.junit.platform.commons.util.*Utils` with`@API(Maintained)` and compile an overview of stable and maintained assets: https://github.com/junit-team/junit5/issues/144
2. Implement an indirection layer exposing maintained parts of utilities, like https://github.com/junit-team/junit5/compare/issue-246-352-utilities

sormuras:
Skimming http://openjdk.java.net/projects/jigsaw/spec/sotms/ - especially:
> _Strong encapsulation_, to allow a component to declare which of its **public types** are accessible to other components, and which are not.
makes me wonder, if i could solve this issue ... in the future.

nicolaiparlog:
> makes me wonder, if it could solve this issue ... in the future.

This does not really change anything. It just means that you have to pick the packages whose public types are to be accessible but it still requires you to put the feature into a specific class in a specific package.

One thing Jigsaw can do, though, is make `@API(Maintained)` superfluous as, ideally, exactly those types should be accessible to code outside of the Jigsaw modules.

jlink:
As far as I know Jigsaw can only provide/deny visibility. The `API` annotation is more explicit about the level of support.

jbduncan:
> Two possible solutions:
>
> 1. Mark parts of org.junit.platform.commons.util.*Utils with@API(Maintained) and compile an overview of stable and maintained assets: #144
[...]

PR #651, opened 2017-02-15
Add utility API to ExtensionContext.Util

### LINKED ISSUE #246: Make core features of AnnotationUtils available to extension authors
While implementing extensions it is common to look for annotations. The easiest way to do this properly (i.e. including repeatable and meta-annotations) seems to be `AnnotationUtils`. It is currently marked as internal API.

Annotations and especially meta-annotations are a pillar of JUnit 5's extensibility. Being able to freely compose existing annotations according to ones own requirements will be a great boost for customizations. It would be counter-intuitive if third-party extensions would not compose as well. They are hence required to use the same look-up logic as official annotations.

The easiest way to do this is to _encourage_ using the same utility to find annotations.

sormuras:
## Overview

Fixes #246 by publishing useful internal helper methods via an API facade. At the moment, only annotation-related helpers are included. The `ExtensionContext.Util` interface might grow in the future.

## Discussion

1. Omit nested `Util` interface and pull helper methods up to `ExtensionContext`?
2. If nested interface is wanted, find a better name then `Util`?
3. Expose more helper methods, if yes, which?

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -12,11 +12,14 @@ package org.junit.jupiter.api.extension;
 
 import static org.junit.platform.commons.meta.API.Usage.Experimental;
 
+import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
@@ -35,6 +38,7 @@ import org.junit.platform.commons.util.Preconditions;
  * @since 5.0
  * @see Store
  * @see Namespace
+ * @see Util
  */
 @API(Experimental)
 public interface ExtensionContext {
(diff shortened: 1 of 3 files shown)


jbduncan:
It's not clear to me if having a nested interface `Util` is the way to go, but if it is, then I'd suggest renaming it to `Utilities` or `Utility(Functions|Methods)`.

jlink:
The methods in Util will also be needed by other test engines. Making them available in ExtensionContext does not really help here.

sormuras:
Preparing a new PR which exposes the same "Util" API to both: `ExtensionContext` and `TestEngine` and potential more base interfaces. See https://github.com/junit-team/junit5/pull/651#ref-issue-162102546

sormuras:
@jbduncan @junit-team/junit-lambda What do think of this https://github.com/junit-team/junit5/compare/issue-246-352-utilities approach? Using interface extension feels a bit ... wrong. May be a dedicated `default Utilities getUtilities() { return singletonUtilies; }` feels better.

jlink:
IMO parts of *Utils org.junit.platform.commons should be made available in org.junit.platform.engine.support. 

Alternatively: Just start marking parts of org.junit.platform.commons with the appropriate @API annotations. It's just a matter of what the team wants to commit to. No indirection through additional classes will get rid of the fundamental problem.

sormuras:
True, @jlink - one advantage of having an indirection is, that if an internal change is needed, the published API could possibly offer the old behaviour through its API w/o forcing changes to code from extension/engine authors.

Personally, I like the idea of tagging the already `public` helper methods with `@API(Maintained)`

jbduncan:
I agree with @jlink that annotating the relevant methods in the commons utilities with `@API(Maintained)` (or the relevant API annotations) is simplest, and that's what I'd personally be inclined to do.

sormuras:
Closing this PR. Discussion continues over at #352 -- which should include solution needed for #246

Issue #616, opened 2016-12-24
Running Cucumber with JUnit5 and Surefire exposes base64 ids.


avandeursen:
The classname set for Cucumber tests when run with Surefire in JUnit5 exposes the base64 encoding used internally in the generated XML test reports.

Original outcome when running junit4:

    grep classname target/surefire-reports/TEST-nl.tudelft.cucumber64.CucumberTest.xml

Output: 

```xml
    <testcase name="Given I have chosen to sign up" classname="Scenario: Successful sign up." time="0.001"/>
    <testcase name="When I sign up with valid details" classname="Scenario: Successful sign up." time="0"/>
    <testcase name="Then I should receive a confirmation email" classname="Scenario: Successful sign up." time="0"/> 
    <testcase name="And I should see a personalized greeting message" classname="Scenario: Successful sign up." time="0.001"/>
    <testcase name="Scenario: Successful sign up." classname="Scenario: Successful sign up." time="0.001"/>    
```

Note that the classname is not a real class, but the name of the scenario executed extracted from a `.feature` file. The underlying magic is done by the Cucumber Junit4 test runner.

Doing the same when using the JUnit 5 Surefire yields as output:

```xml
    <testcase 
      name="Given I have chosen to sign up" 
      classname=
        "[engine:junit-vintage]
        /[runner:nl.tudelft.cucumber64.CucumberTest]
[...]

PR #654, opened 2017-02-16
Legacy reporting in surefire provider

LiamClark:
## Overview
This pr implements @marcphilipp solution for subclass names in legacy reporting in #616 

Open questions:

1 To make the existing ListenerAdapter tests pass with the new legacy reporting. I changed the implementation of generateDefaultDisplayName in ClassTestDescriptor. Before it used the classname but stripped the package prefix. 

However there were other tests that expected just the classname, these pass now because I changed the expected value to the canocicalName instead of the simpleClassName. Since this is a display name and as mentioned in the javadoc this string is only used for display and should never be parsed, I think this is not a breaking change, however It might not be the behaviour you want for showing the name of the just the test to a user instead of a reporting tool.

The name including the packages, is however required for the default implementation to make any sense at all. Should we instead split this up into two different methods?

2 There is an override of getLegacyReportingInfo in ClassTestDescriptor, that juggles around the class name and the method name, this feels a bit hacky. Would it be possible to solve this behaviour in the default implementation by checking if the descriptor is a container?

3 What would you like to see in the javadoc for the legacy info?

---

---

### Definition of Done


junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -182,6 +183,24 @@ default void accept(Visitor visitor) {
...
 	/**
+	 * Does a logical level display name resolution for legacy reporting formats, instead of a source level name resolution.
+	 *
+	 * @return A LegacyReportingInfo containing the logical method and class display names.
+	 */
+	default LegacyReportingInfo getLegacyReportingInfo() {
+		final String displayName = getDisplayName();
+		final Optional<String> className = getParent().map(TestDescriptor::getDisplayName);
+
+		return new DefaultLegacyReportingInfo(displayName, className.orElse(null));
```
LiamClark: This makes the degraph task fail in the ci, because org.junit.platform.engine now also references org.junit.platform.engine.support.descriptor, creating a cyclic dependency

I could just return an anomynous implementation here, but what is the preferred solution?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/DefaultLegacyReportingInfo.java
```diff
@@ -0,0 +1,33 @@
+package org.junit.platform.engine.support.descriptor;
```
jbduncan: Hi @LiamClark, would you kindly run `./gradlew spotlessApply` as soon you're ready, as it will (or, at least should) put an Eclipse Public License header just above this line. :)

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/DefaultLegacyReportingInfo.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class DefaultLegacyReportingInfo implements TestDescriptor.LegacyReportingInfo, Serializable {
+    private static final long serialVersionUID = 2896941058543948169L;
+    private final String methodName;
+    private final String className;
+
+
+    public static DefaultLegacyReportingInfo from(TestDescriptor.LegacyReportingInfo info) {
+        final String methodName = info.getMethodName().orElse(null);
+        final String className = info.getClassName().orElse(null);
```
jbduncan: I feel a bit iffy about having `final` keywords on the two lines.

IMO, they feel unnecessary as the method is so short that it's reasonably easy to tell that they're never modified, and because of this, I think they only make the code a bit harder to read and maintain.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestIdentifier.java
```diff
@@ -61,18 +63,20 @@ public static TestIdentifier from(TestDescriptor testDescriptor) {
 		boolean container = !test || !testDescriptor.getChildren().isEmpty();
 		Optional<String> parentId = testDescriptor.getParent().map(
 			parentDescriptor -> parentDescriptor.getUniqueId().toString());
-		return new TestIdentifier(uniqueId, displayName, source, tags, test, container, parentId);
+		final DefaultLegacyReportingInfo legacyReportingInfo = DefaultLegacyReportingInfo.from(testDescriptor.getLegacyReportingInfo());
```
jbduncan: I don't think the `final` keyword is need here either, as the variable is used immediately on the next line.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java
```diff
@@ -83,36 +81,19 @@ else if (testIdentifier.isTest()) {
 	}
 
 	private SimpleReportEntry createReportEntry(TestIdentifier testIdentifier, Optional<Throwable> throwable) {
-		Optional<String> className = getClassName(testIdentifier);
+		final TestDescriptor.LegacyReportingInfo legacyReportingInfo = testIdentifier.getLegacyReportingInfo();
```
jbduncan: Again I don't think the `final` is needed, because the method is so short.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java
```diff
@@ -83,36 +81,19 @@ else if (testIdentifier.isTest()) {
...
+		Optional<String> className = legacyReportingInfo.getClassName();
+		final String methodName = legacyReportingInfo.getMethodName().orElse("");
```
jbduncan: Ditto.

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/JUnitPlatformProviderTests.java
```diff
@@ -191,7 +191,7 @@ private void verifyPreconditionViolationException(Map<String, String> properties
 		assertEquals(JUnitPlatformProvider.EXCEPTION_MESSAGE_BOTH_NOT_ALLOWED, throwable.getMessage());
 	}
 
-	private static ProviderParameters providerParametersMock(Class<?>... testClasses) {
+	static ProviderParameters providerParametersMock(Class<?>... testClasses) {
```
jbduncan: Considering that this method is now used in both `JupiterPlatformProviderTests` and `LegacyReportingTests`, I suggest either creating a superclass containing this method which both classes will extend from, or to create a tiny utilities class which both classes will import.

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/JUnitPlatformProviderTests.java
```diff
@@ -212,7 +212,7 @@ private static ProviderParameters providerParametersMock(Class<?>... testClasses
 		return providerParameters;
 	}
 
-	private static TestsToRun newTestsToRun(Class<?>... testClasses) {
+	static TestsToRun newTestsToRun(Class<?>... testClasses) {
```
jbduncan: Ditto.

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/LegacyReportingTests.java
```diff
@@ -0,0 +1,55 @@
...
+public class LegacyReportingTests {
+	private static final String[] classNames = {"org.junit.platform.surefire.provider.LegacyReportingTests$Sub1Tests",
+			"org.junit.platform.surefire.provider.LegacyReportingTests$Sub2Tests"};
+
+
+	@Test
+	public void testLegacyXmlReport() throws TestSetFailedException, InvocationTargetException {
+		ProviderParameters providerParameters = providerParametersMock(Sub1Tests.class, Sub2Tests.class);
+
+		final JUnitPlatformProvider jUnitPlatformProvider = new JUnitPlatformProvider(providerParameters);
```
jbduncan: Again `final` not needed AFAICT. :)

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/LegacyReportingTests.java
```diff
@@ -0,0 +1,55 @@
...
+		TestsToRun testsToRun = newTestsToRun(Sub1Tests.class, Sub2Tests.class);
+
+		jUnitPlatformProvider.invoke(testsToRun);
+		final RunListener reporter = providerParameters.getReporterFactory().createReporter();
```
jbduncan: Ditto.

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/LegacyReportingTests.java
```diff
@@ -0,0 +1,55 @@
...
+
+		ArgumentCaptor<ReportEntry> reportEntryArgumentCaptor = ArgumentCaptor.forClass(ReportEntry.class);
+		verify(reporter, times(2)).testSucceeded(reportEntryArgumentCaptor.capture());
+
+		final List<ReportEntry> allValues = reportEntryArgumentCaptor.getAllValues();
```
jbduncan: Ditto.

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
```diff
@@ -180,9 +183,11 @@ private static TestIdentifier newClassIdentifier() {
 
 	private static TestIdentifier newSourcelessIdentifierWithParent(TestPlan testPlan, String parentDisplay) {
 		// A parent test identifier with a name.
+		final DefaultLegacyReportingInfo emptyLegacyReportingInfo = new DefaultLegacyReportingInfo("", parentDisplay);
```
jbduncan: Again final not needed.

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
```diff
@@ -170,6 +171,8 @@ void notifiedWithParentDisplayNameWhenTestClassUnknown() throws Exception {
 	private static TestIdentifier newMethodIdentifier() throws Exception {
 		TestDescriptor testDescriptor = new MethodTestDescriptor(newId(), MyTestClass.class,
 			MyTestClass.class.getDeclaredMethod(MY_TEST_METHOD_NAME));
+
```
jbduncan: I think the blank line here is more confusing than not, as I consider the 4 lines in this method to be part of the same "logical" block of code/functionality.

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -182,6 +182,24 @@ default void accept(Visitor visitor) {
...
+	}
+
+	interface LegacyReportingInfo {
```
marcphilipp: Please convert this interface into a top-level interface, add Javadoc and an `@API` annotation.

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
marcphilipp: I think an anonymous implementation should be fine here (see https://github.com/junit-team/junit5/issues/616#issuecomment-270918178).

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -182,6 +182,24 @@ default void accept(Visitor visitor) {
...
+		Optional<String> getMethodName();
```
marcphilipp: Please rename this to `getName()` and make it non-optional.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestIdentifier.java
```diff
@@ -61,18 +63,21 @@ public static TestIdentifier from(TestDescriptor testDescriptor) {
 		boolean container = !test || !testDescriptor.getChildren().isEmpty();
 		Optional<String> parentId = testDescriptor.getParent().map(
 			parentDescriptor -> parentDescriptor.getUniqueId().toString());
-		return new TestIdentifier(uniqueId, displayName, source, tags, test, container, parentId);
+		final DefaultLegacyReportingInfo legacyReportingInfo = DefaultLegacyReportingInfo.from(
```
marcphilipp: Please remove the `final` keyword.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java
```diff
@@ -83,36 +82,19 @@ else if (testIdentifier.isTest()) {
...
```
marcphilipp: If you do the rename I requested above, this local variable should then be called `name`.

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/LegacyReportingTests.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.maven.surefire.util.TestsToRun;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+
+public class LegacyReportingTests {
+	private static final String[] classNames = { "org.junit.platform.surefire.provider.LegacyReportingTests$Sub1Tests",
+			"org.junit.platform.surefire.provider.LegacyReportingTests$Sub2Tests" };
+
+	@Test
+	public void testLegacyXmlReport() throws TestSetFailedException, InvocationTargetException {
```
marcphilipp: Please move this test to `JUnitPlatformProviderTests` and give it a more meaningful name, e.g. `usesClassNamesForXmlReport`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -114,10 +115,13 @@ public final boolean isContainer() {
...
+	@Override
+	public LegacyReportingInfo getLegacyReportingInfo() {
+		return new DefaultLegacyReportingInfo(null, this.getDisplayName());
+	}
+
 	private static String generateDefaultDisplayName(Class<?> testClass) {
-		String name = testClass.getName();
-		int index = name.lastIndexOf('.');
-		return name.substring(index + 1);
+		return testClass.getName();
```
marcphilipp: Please don't change the display name of the test descriptor.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -114,10 +115,13 @@ public final boolean isContainer() {
 		return true;
 	}
 
+	@Override
+	public LegacyReportingInfo getLegacyReportingInfo() {
+		return new DefaultLegacyReportingInfo(null, this.getDisplayName());
```
marcphilipp: I think this should be:
```
return new DefaultLegacyReportingInfo(testClass.getName(), getParent().map(TestDescriptor::LegacyReportingInfo).map(LegacyReportingInfo::getName));
```

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
LiamClark: What API level should it be? i'm assuming expirmental

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
LiamClark: I was indeed dubious about this (also in the open questions), but some tests for example: https://github.com/junit-team/junit5/blob/master/junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java#L70

expect the name with the package prefix 
org.junit.platform.surefire.provider.RunListenerAdapterTests$MyTestClass

and not:RunListenerAdapterTests$MyTestClass.

so what would be the clean way to fix this?

marcphilipp:
@LiamClark Please put work on this pull request on hold until we've discussed my previous concern:

> However, now that I see it, I think it would be sufficient if we added a `getLegacyReportingName() : String` to `TestDescriptor` and `TestIdentifier`. The `classname` we report in the XML would then simply be `testPlan.getParent(identifier).map(TestIdentifier::getLegacyReportingName).orElse("")`.

@junit-team/junit-lambda What do you think?

LiamClark:
@marcphilipp if you don't mind me chiming in,

I simply tried out your proposal, (I can share the implementation if it helps with the discussion).
It seems cleaner since a TestIdentifier only has to identify itself and not deal with the awkward name and classname. 

This is now still taken care of in the RunListenerAdapter, but this remains an awkward instanceof check for the ClassTestDescriptor, but keeps this (what I assume to be) detail local to the surefire code and doesn't spread it throughout the core model.

marcphilipp:
> I can share the implementation if it helps with the discussion

Please do, e.g. via an additional pull request. Thanks! 👍

marcphilipp:
Closed in favor of #662.

PR #662, opened 2017-02-18
Implementation of legacyreporting for the surefire provider using getLegacyName

LiamClark:
## Overview

An alternative Implementation of #616 corresponding to the suggestion in the discussion in #654 
---

---

### Definition of Done

### DIFF
--- a/junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java
+++ b/junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java
@@ -65,8 +65,8 @@ final class RunListenerAdapter implements TestExecutionListener {
 
 	@Override
 	public void executionSkipped(TestIdentifier testIdentifier, String reason) {
-		String source = getClassName(testIdentifier).orElseGet(() -> parentDisplayName(testIdentifier));
-		runListener.testSkipped(ignored(source, testIdentifier.getDisplayName(), reason));
+		String source = sourceLegacyReportingName(testIdentifier);
+		runListener.testSkipped(ignored(source, testIdentifier.getLegacyReportingName(), reason));
 	}
 
 	@Override
@@ -84,14 +84,18 @@ final class RunListenerAdapter implements TestExecutionListener {
 
 	private SimpleReportEntry createReportEntry(TestIdentifier testIdentifier, Optional<Throwable> throwable) {
 		Optional<String> className = getClassName(testIdentifier);
-		if (className.isPresent()) {
-			StackTraceWriter traceWriter = new PojoStackTraceWriter(className.get(),
-				getMethodName(testIdentifier).orElse(""), throwable.orElse(null));
-			return new SimpleReportEntry(className.get(), testIdentifier.getDisplayName(), traceWriter, null);
-		}
-		else {
-			return new SimpleReportEntry(parentDisplayName(testIdentifier), testIdentifier.getDisplayName(), null);
-		}
+
+		Optional<StackTraceWriter> stackTraceWriter = throwable.flatMap(
+			t -> className.map(name -> new PojoStackTraceWriter(name, getMethodName(testIdentifier).orElse(""), t)));
+
+		String source = sourceLegacyReportingName(testIdentifier);
+		return new SimpleReportEntry(source, testIdentifier.getLegacyReportingName(), stackTraceWriter.orElse(null),
+			null);
+	}
+
+	private String sourceLegacyReportingName(TestIdentifier testIdentifier) {
+		return testPlan.flatMap(plan -> plan.getParent(testIdentifier)).map(
+			TestIdentifier::getLegacyReportingName).orElse("<unrooted>");
 	}
 
 	private Optional<String> getClassName(TestIdentifier testIdentifier) {
@@ -117,7 +121,7 @@ final class RunListenerAdapter implements TestExecutionListener {
 		// @formatter:off
 		return testPlan
 			.flatMap(plan -> plan.getParent(testIdentifier))
-			.map(TestIdentifier::getDisplayName)
+			.map(TestIdentifier::getLegacyReportingName)
 			.orElseGet(testIdentifier::getUniqueId);
 		// @formatter:on
 	}
--- a/junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/JUnitPlatformProviderTests.java
+++ b/junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/JUnitPlatformProviderTests.java
@@ -21,9 +21,9 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -32,8 +32,10 @@ import java.util.List;
 import java.util.Map;
 
 import org.apache.maven.surefire.providerapi.ProviderParameters;
+import org.apache.maven.surefire.report.ReportEntry;
 import org.apache.maven.surefire.report.ReporterFactory;
 import org.apache.maven.surefire.report.RunListener;
+import org.apache.maven.surefire.testset.TestSetFailedException;
 import org.apache.maven.surefire.util.RunOrderCalculator;
 import org.apache.maven.surefire.util.ScanResult;
 import org.apache.maven.surefire.util.TestsToRun;
@@ -45,6 +47,7 @@ import org.junit.platform.launcher.TestPlan;
 import org.junit.platform.launcher.core.LauncherFactory;
 import org.junit.platform.launcher.listeners.SummaryGeneratingListener;
 import org.junit.platform.launcher.listeners.TestExecutionSummary;
+import org.mockito.ArgumentCaptor;
 
 /**
  * Unit tests for {@link JUnitPlatformProvider}.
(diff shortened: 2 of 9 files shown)


junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java
```diff
@@ -83,41 +80,27 @@ else if (testIdentifier.isTest()) {
...
-		}
-		else {
-			return new SimpleReportEntry(parentDisplayName(testIdentifier), testIdentifier.getDisplayName(), null);
-		}
-	}
+		String className = getClassName(testIdentifier);
+
+		StackTraceWriter traceWriter = new PojoStackTraceWriter(className, testIdentifier.getLegacyReportingName(),
+			throwable.orElse(null));
+		return new SimpleReportEntry(className, testIdentifier.getDisplayName(), traceWriter, null);
```
marcphilipp: SimpleReportEntry should _always_ use `testIdentifier.getLegacyReportingName()` as `name` and `testPlan.flatMap(plan -> plan.getParent(testIdentifier)).map(TestIdentifier::getLegacyReportingName)
  			.orElse("<unrooted>")` as `source`.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java
```diff
@@ -83,41 +80,27 @@ else if (testIdentifier.isTest()) {
...
-		if (testSource instanceof MethodSource) {
-			return Optional.of(((MethodSource) testSource).getMethodName());
+	private String getClassName(TestIdentifier testIdentifier) {
+		if (testIdentifier.isContainer()) {
+			return testIdentifier.getLegacyReportingName();
 		}
-		return Optional.empty();
+
+		return parentDisplayName(testIdentifier);
 	}
```
marcphilipp: I think we should leave `getClassName()` and `getMethodName()` as is but _only_ pass them to `PojoStackTraceWriter` to "optimize" the stacktrace.

junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java
marcphilipp: In addition, we should only create a `StackTraceWriter` when there's a `Throwable`.

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
```diff
@@ -92,13 +95,17 @@ void notifiedWhenMethodExecutionSkipped() throws Exception {
...
+		TestIdentifier classIdentifier = newClassIdentifierOnTestPlan(testPlan);
 
-		adapter.executionSkipped(newClassIdentifier(), "test");
+		adapter.executionSkipped(classIdentifier, "test");
 		verify(listener).testSkipped(entryCaptor.capture());
 
 		ReportEntry entry = entryCaptor.getValue();
 		assertTrue(MyTestClass.class.getTypeName().contains(entry.getName()));
-		assertEquals(MyTestClass.class.getName(), entry.getSourceName());
+		assertEquals("<unrooted>", entry.getSourceName());
```
LiamClark: This test's behaviour is now wrong, the result is "<unrooted>", because the class identifier has no parent in the test fixture. However for the right name to appear here the class' parent would need to be itself.

@marcphilipp how do you propose we fix this, or was the test wrong in the first place?

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
```diff
@@ -128,12 +135,15 @@ void notifiedWhenMethodExecutionFailed() throws Exception {
...
+		adapter.testPlanExecutionStarted(testPlan);
 
-		adapter.executionFinished(newClassIdentifier(), TestExecutionResult.failed(new RuntimeException()));
+		adapter.executionFinished(newClassIdentifierOnTestPlan(testPlan),
+			TestExecutionResult.failed(new RuntimeException()));
 		verify(listener).testFailed(entryCaptor.capture());
 
 		ReportEntry entry = entryCaptor.getValue();
-		assertEquals(MyTestClass.class.getName(), entry.getSourceName());
+		assertEquals("<unrooted>", entry.getSourceName());
```
LiamClark: ditto

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/JUnitPlatformProviderTests.java
```diff
@@ -21,9 +21,9 @@
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
jbduncan: Hi @LiamClark, did you mean to collapse the Mockito imports into a single wildcard?

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
```diff
@@ -16,9 +16,7 @@
 
 package org.junit.platform.surefire.provider;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.*;
```
jbduncan: Did you also mean to collapse these assertion imports into one?

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
LiamClark: Checking the old behaviour of surefire reporting an entire class being skipped it gives: 
<testcase name="org.test.SureFireTest" classname="org.test.SureFireTest" time="0">
    <skipped/>
</testcase>

And running this pr through an integration test we report:
name= "org.test.SureFireTest" and classname= "JUnit Jupiter"

marcphilipp: I think in this case it would be best to omit the `classname` completely. Does the schema even allow that?

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/JUnitPlatformProviderTests.java
LiamClark: @jbduncan I did not, still had a conflicting intelij setting, oops

jbduncan: @LiamClark Great, looks good in your latest commit now. :)

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
LiamClark: Every implementation of report entry in surefire does require it as a constructor argument, which would suggest it is required rather than optional.

marcphilipp: Ok, I was afraid so. I think the old behavior is kind of strange and more of an edge case. Would you agree?

LiamClark: Yes, reporting the same name twice in different slots seems redundant and like a quirk in surefire's model

marcphilipp: Maybe reporting "JUnit Jupiter" isn't so bad after all. This is really a tricky issue. I don't think we need 100% compatibility with the old output, though.

LiamClark: I was planning on digging through surefire a bit more to see if I could find something usefull, if not ill change the test to atleast correctly output that :)

marcphilipp: Sure, makes sense!

LiamClark:
@marcphilipp That's in there now, but just to be sure this is so that the `@DisplayName` doesn't get into the report?

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
```diff
@@ -31,8 +31,7 @@
 
 import org.apache.maven.surefire.report.ReportEntry;
 import org.apache.maven.surefire.report.RunListener;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.*;
```
marcphilipp: Please don't use wildcard imports.

marcphilipp:
> […] just to be sure this is so that the `@DisplayName` doesn't get into the report?

Yes, because tools that process the legacy XML report expect `methodName` to point to a method name, right?

junit-platform-surefire-provider/src/test/java/org/junit/platform/surefire/provider/RunListenerAdapterTests.java
LiamClark: oops, I worked on a different machine sorry

LiamClark:
exactly, that's what I thought

marcphilipp:
I'd like to wait for Travis and Gradle to recover from the S3 outage and restart the builds before merging this PR.

LiamClark:
@marcphilipp seems the build passes now, anything else preventing a merge? Should I perhaps do a squash?

marcphilipp:
@LiamClark Looks good! Can you add an entry to the Release Notes?

LiamClark:
I gave it my best shot, but sadly I have little experience writing release notes.

I put it as a bugfix under JUnit platform, is this correct? 
Are there other things you would you like to see?

Feedback would be very much appreciated!

marcphilipp:
Thanks, @LiamClark! 👍

PR #684, opened 2017-02-23
Close dynamic test streams, fixes #683

### LINKED ISSUE #683: Close the stream in org.junit.jupiter.engine.descriptor.TestFactoryTestDescriptor#invokeTestMethod()
## Overview

The stream received by a @TestFactory method should be closed after all the tests have been run.

## Motivation

Currently, if you wish to create dynamic tests based on the data from file, you cannot call the following:

```
@TestFactory
Stream<DynamicText> pointTests() {
    return
        Files.lines(fileName).map(Point::parse).map(
            point -> dynamicTest("Testing " + p, () -> pointTest(p))
        );
}

```
because the stream represented by `Files.lines()` is a file-system-based resource and must be closed.

Unfortunately we cannot just use
```
@TestFactory
Stream<DynamicText> pointTests() {
{
    try (final Stream<String> lines = Files.lines(fileName)) {
    return
        lines.map(Point::parse).map(
            point -> dynamicTest("Testing " + p, () -> pointTest(p)))
    );
}
```
as the source stream is closed **before** JUnit5 even starts reading from it (the termination operation is lazy).

The only solution currently is probably only using an interim collection and collect the lines to it, which is awkward:

```
@TestFactory
Stream<DynamicText> pointTests() {
    final List<Point> data;
    try (final Stream<String> lines = Files.lines(fileName)) {
        data = lines.map(Point::parse).collect(Collectors.toList());
    );
    return data.map(point -> dynamicTest("Testing " + p, () -> pointTest(p));
}
```

**It would be most convenient if the contract of @TestFactory guaranteed calling Stream#close().**

## Documentation

It is crucial that all the related documentation states that the stream is closed, so it becomes part of the contract and the clients may rely on it.

LiamClark:
## Overview
Added a call to close for dynamic test streams #683 .

I decided to put the test in it's own test file, due to the considerable amount of setup.
Is there any other place this test might belong to?

### Documentation

I'm not sure where the right place would be to document this, My guess is in: https://github.com/junit-team/junit5/blob/master/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestFactory.java#L32

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/writing-tests.adoc
@@ -394,6 +394,9 @@ Technically speaking, a `@TestFactory` method must return a `Stream`,
 These `DynamicTest` instances will then be executed lazily,
 enabling dynamic and even non-deterministic generation of test cases.
 
+any Stream returned by a {@code @TestFactory} will be properly closed,
+making it safe to use a resource for example Files.lines().
+
 As with `@Test` methods, `@TestFactory` methods must not be `private` or `static`
 and may optionally declare parameters to be resolved by `ParameterResolvers`.
 
--- a/documentation/src/docs/asciidoc/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/writing-tests.adoc
@@ -394,8 +394,8 @@ Technically speaking, a `@TestFactory` method must return a `Stream`,
 These `DynamicTest` instances will then be executed lazily,
 enabling dynamic and even non-deterministic generation of test cases.
 
-any Stream returned by a {@code @TestFactory} will be properly closed,
-making it safe to use a resource for example Files.lines().
+Any `Stream` returned by a `@TestFactory` will be properly closed by calling `stream.close()`,
+making it safe to use a resource such as `Files.lines()`.
 
 As with `@Test` methods, `@TestFactory` methods must not be `private` or `static`
 and may optionally declare parameters to be resolved by `ParameterResolvers`.
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestFactory.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestFactory.java
@@ -34,6 +34,9 @@ import org.junit.platform.commons.meta.API;
  * will then be executed lazily, enabling dynamic and even non-deterministic
  * generation of test cases.
  *
+ * any Stream returned by a {@code @TestFactory} will be properly closed,
+ * making it safe to use a resource for example {@code Files.lines()}.
+ *
  * <p>{@code @TestFactory} methods may optionally declare parameters to be
  * resolved by {@link org.junit.jupiter.api.extension.ParameterResolver
  * ParameterResolvers}.
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestFactory.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestFactory.java
@@ -34,8 +34,8 @@ import org.junit.platform.commons.meta.API;
  * will then be executed lazily, enabling dynamic and even non-deterministic
  * generation of test cases.
  *
- * any Stream returned by a {@code @TestFactory} will be properly closed,
- * making it safe to use a resource for example {@code Files.lines()}.
+ * <p>Any {@code Stream} returned by a {@code @TestFactory} will be properly closed by calling {@code stream.close()},
+ * making it safe to use a resource such as {@code Files.lines()}.
  *
  * <p>{@code @TestFactory} methods may optionally declare parameters to be
  * resolved by {@link org.junit.jupiter.api.extension.ParameterResolver
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
@@ -81,9 +81,10 @@ public class TestFactoryTestDescriptor extends MethodTestDescriptor {
 			Object testFactoryMethodResult = executableInvoker.invoke(method, instance, testExtensionContext,
 				context.getExtensionRegistry());
 
-			try {
+			try (Stream<DynamicTest> dynamicTestStream = toDynamicTestStream(testExtensionContext,
+				testFactoryMethodResult)) {
 				AtomicInteger index = new AtomicInteger();
-				toDynamicTestStream(testExtensionContext, testFactoryMethodResult).forEach(
+				dynamicTestStream.forEach(
 					dynamicTest -> registerAndExecute(dynamicTest, index.incrementAndGet(), dynamicTestExecutor));
 			}
 			catch (ClassCastException ex) {
--- a/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
+++ b/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.jupiter.engine.descriptor;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.lang.reflect.Method;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.TestFactory;
+import org.junit.jupiter.api.extension.TestExtensionContext;
+import org.junit.jupiter.engine.execution.JupiterEngineExecutionContext;
+import org.junit.jupiter.engine.execution.ThrowableCollector;
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.hierarchical.Node;
+
+public class DynamicTestStreamClosedTests {
+	@SuppressWarnings("unchecked")
+	private static final Stream<DynamicTest> mockStream = mock(Stream.class);
+	private JupiterEngineExecutionContext context;
+	private Method testMethod;
+
+	@Before
+	public void before() throws NoSuchMethodException {
+		TestExtensionContext testExtensionContext = mock(TestExtensionContext.class);
+
+		testMethod = DynamicCloseHookedStreamTest.class.getMethod("customStream");
+		when(testExtensionContext.getTestMethod()).thenReturn(Optional.of(testMethod));
+		when(testExtensionContext.getTestInstance()).thenReturn(new DynamicCloseHookedStreamTest());
+
+		context = new JupiterEngineExecutionContext(null, null).extend().withThrowableCollector(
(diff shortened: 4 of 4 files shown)


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
```diff
@@ -88,8 +88,12 @@ protected void invokeTestMethod(JupiterEngineExecutionContext context) {
 
 			try {
```
sormuras: Should be using try-with-resource syntax, like:
```
AtomicInteger index = new AtomicInteger();
try (Stream<DynamicTest> stream = toDynamicTestStream(testExtensionContext, testFactoryMethodResult)) {
    stream.forEach(dynamicTest -> registerAndExecute(dynamicTest, index.incrementAndGet(), listener));
}
```
No need to close the stream manually, if exception are thrown.

LiamClark: ah yes ofcourse, for some reason my brain failed to associate the close in stream method with autoclosable, thanks!

sormuras:
The documentation target guess is good and the same text should go into the `=== Dynamic Tests` section in `documentation/src/docs/asciidoc/writing-tests.adoc` below line 377 somewhere as well.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.lang.reflect.Method;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.junit.Before;
+import org.junit.Test;
```
sormuras: Convert to JUnit 5 test by using JUnit 5 annotations, please.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestFactory.java
```diff
@@ -34,6 +34,9 @@
  * will then be executed lazily, enabling dynamic and even non-deterministic
  * generation of test cases.
  *
+ * <p>Any Stream returned by a {@code @TestFactory} will be properly closed by calling {@code stream.close()},
```
jbduncan: Minor nitpick - please consider surrounding `Stream` with `{@code ... }` for consistency with all the other class references. :)

documentation/src/docs/asciidoc/writing-tests.adoc
```diff
@@ -394,6 +394,9 @@ Technically speaking, a `@TestFactory` method must return a `Stream`,
 These `DynamicTest` instances will then be executed lazily,
 enabling dynamic and even non-deterministic generation of test cases.
 
+Any Stream returned by a `@TestFactory` will be properly closed by calling `stream.close()`,
```
jbduncan: Ditto.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestFactory;
+import org.junit.jupiter.api.extension.TestExtensionContext;
+import org.junit.jupiter.engine.execution.JupiterEngineExecutionContext;
+import org.junit.jupiter.engine.execution.ThrowableCollector;
+import org.junit.platform.engine.UniqueId;
+
+public class DynamicTestStreamClosedTests {
```
jbduncan: Consider making this test class package-private for brevity's sake. :)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,103 @@
...
+	private JupiterEngineExecutionContext context;
+	private Method testMethod;
+	private TestExtensionContext testExtensionContext;
+	private boolean isClosed;
+
+	@BeforeEach
+	public void before() {
```
jbduncan: And also consider making this method package-private for brevity.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,103 @@
...
+		testExtensionContext = mock(TestExtensionContext.class);
+		isClosed = false;
+
+		context = new JupiterEngineExecutionContext(null, null).extend().withThrowableCollector(
+			new ThrowableCollector()).withExtensionContext(testExtensionContext).build();
+	}
+
+	@Test
+	public void streamsFromTestFactoriesShouldBeClosed() throws NoSuchMethodException {
```
jbduncan: Ditto.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,103 @@
...
+		testMethod = DynamicCloseHookedStreamTest.class.getMethod("customStream");
+		when(testExtensionContext.getTestMethod()).thenReturn(Optional.of(testMethod));
+
+		Stream<DynamicTest> stream = Stream.<DynamicTest> of().onClose(() -> isClosed = true);
```
jbduncan: Is this type argument `<DynamicTest>` really needed? If so, I'd suggest removing the space between the `>` and `of` for readability/brevity (assuming that passes the `gradlew spotlessCheck`!)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,103 @@
...
+
+		TestFactoryTestDescriptor descriptor = new TestFactoryTestDescriptor(UniqueId.forEngine("engine"),
+			DynamicCloseHookedStreamTest.class, testMethod);
+		descriptor.invokeTestMethod(context);
+
+		assertTrue(isClosed);
+	}
+
+	@Test
+	public void streamsFromTestFactoriesShouldBeClosedWhenTheyThrow() throws NoSuchMethodException {
```
jbduncan: Ditto.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,103 @@
...
+
+		TestFactoryTestDescriptor descriptor = new TestFactoryTestDescriptor(UniqueId.forEngine("engine"),
+			StreamOfIntClosedTest.class, testMethod);
+		descriptor.invokeTestMethod(context);
+
+		assertTrue(isClosed);
+	}
+
+	@Disabled
+	public static class DynamicCloseHookedStreamTest {
```
jbduncan: Ditto.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,103 @@
...
+		when(testExtensionContext.getTestInstance()).thenReturn(new StreamOfIntClosedTest(mockStream));
+
+		TestFactoryTestDescriptor descriptor = new TestFactoryTestDescriptor(UniqueId.forEngine("engine"),
+			StreamOfIntClosedTest.class, testMethod);
+		descriptor.invokeTestMethod(context);
+
+		assertTrue(isClosed);
+	}
+
+	@Disabled
```
jbduncan: I'm struggling to understand why `DynamicCloseHookedStreamTest` is annotated with `@Disabled`.

Could you explain to me what your intent was here?

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
```diff
@@ -0,0 +1,103 @@
...
+			this.mockStream = mockStream;
+		}
+
+		@TestFactory
+		public Stream<DynamicTest> customStream() {
+			return mockStream;
+		}
+	}
+
+	@Disabled
```
jbduncan: Could you also explain your intent for annotating `StreamOfIntClosedTest` with `@Disabled`?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestFactory.java
```diff
@@ -34,6 +34,9 @@
...
+ * making it safe to use a resource, for example {@code Files.lines()}.
```
jbduncan: I think it would read better if this line were to be replaced with something like,

> making it safe to use a resource **such as** {@code Files.lines()}.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
LiamClark: sadly yes it is needed in this form, gradle spotlessApply also put the space in there. however using two variables would get around this, would you prefer that?

I'll also change it to `Stream.empty()`

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
LiamClark: When I changed the annotations used in DynamicTestStreamClosedTests to junit.jupiter.api.Test, these classes started getting picked up for running by them selves. This was the quick fix to stop that, but seems dubious indeed. Lowering the visibility fixes this problem as well so i'll change it to that. :)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/DynamicTestStreamClosedTests.java
LiamClark: Or perhaps just add an no-op dynamic test into the of() call?

jbduncan: I am fine with having it changed to `Stream.{<DynamicTest> }empty()` for explicitness's sake. :)

And if using 2 or more variables would allow the `<DynamicTest>` type arg to be removed, then yes I would personally prefer that.

It's not clear to be if adding a no-op dynamic test in the `of()` would be any more readable than using 2+ vars (I'm writing this with my phone, so I don't have access to IntelliJ IDEA, ATM), but if you think that it is more readable, then I'm happy to trust your judgement. :)

marcphilipp:
Thanks! 👍

jbduncan:
LGTM! Thanks @LiamClark. :)

jbduncan:
And @marcphilipp too. :)

LiamClark:
@sormuras @jbduncan my, pleasure thanks for the great review!

@marcphilipp I love how much you simplified the test, that Stream<?> does look scary though

marcphilipp:
A bit, yes. However, I found the generics don't provide any runtime benefit in this case.

PR #697, opened 2017-02-28
Change ConsoleLauncher execution result from int to summary

sormuras:
Before this commit, the result of `ConsoleLauncher#execute(...)` was reduced to an `int` primitive, the exit code used by `ConsoleLauncher#main`. That reduction limited the use of the `ConsoleLauncher` for integration tests, as the `TestExecutionSummary` created while executing wasn't passed to the calling test.

Now standard and error stream parsing tests may use of all methods defined in `TestExecutionSummary`, like:

    assertEquals(2, new ConsoleLauncherWrapper().execute(args).getTestsFoundCount())

instead of former:

    assertTrue("captured out stream as string".contains("2 tests found"))

## Overview

### TODO

- Rename `ExecuteTestsTask` ... there is no `ConsoleTask` anymore. Could stay for good.
- Re-introduce `ConsoleTaskExecutor` to manage `PrintStream`s _out_ and _err_.
- Move more integration tests from parsing text snippet to new feature.
---


junit-platform-console/src/main/java/org/junit/platform/console/tasks/ExecuteTestsTask.java
```diff
@@ -37,7 +37,7 @@
  * @since 1.0
  */
 @API(Internal)
-public class ExecuteTestsTask implements ConsoleTask {
+public class ExecuteTestsTask {
```
marcphilipp: Maybe `ConsoleTestExecutor` would be a better name now?

Issue #708, opened 2017-03-06
'junitPlatformTest' task can not be retrieved at configuration time by Gradle plugin


mkobit:
## Overview

Currently, the `JUnitPlatformPlugin` has logic to be applied `afterEvaluate` which makes it so users must also do some lazy evaluation logic to be able to also configure it.

For example, is a user wants to apply configuraiton to the `'junitPlatformTest'` task, they might try this by:

```groovy
tasks.getByName('junitPlatformTest').dependsOn(myTask)
```

This will fail at configuration time with an NPE.

A user will have to work around this by making the configuration more lazy. One workaround for a user would be something like:

```groovy
tasks.whenTaskAdded {
  if (it.name == "junitPlatformTest") {
    it.dependsOn(myTask)
  }
}
```

## Deliverables

- [x] Make the `JUnitPlatformPlugin` applied at configuration time rather than `afterEvaluate`
- [x] Change is documented in the User Guide and Release Notes

sormuras:
Seems like an easy fix. Want to send a PR?

Mid-term, an official plugin will be provided by Gradle itself: https://github.com/gradle/gradle/issues/1037

mkobit:
@sormuras Sure! I just opened https://github.com/junit-team/junit5/pull/709 which _I think_ should be merged in first.

sbrannen:
Reopening since this has not been documented in the release notes for M5.

sbrannen:
In other words, #751 was merged without adhering to the Definition of Done.

mkobit:
Sorry @sbrannen , I'll follow up with a PR today or tomorrow with the change.

sbrannen:
Thanks, @mkobit!

We are all in the process of adjusting to our _Definition of Done_. 😉

sbrannen:
Closing this issue since the release notes were updated in #800.

PR #709, opened 2017-03-06
Apply the 'java' plugin when the `JunitPlatformPlugin` is applied

mkobit:
Currently, the `JUnitPlatformPlugin` makes assumptions that the `'java'` plugin has been applied by looking at `sourceSets` and tasks that are created by it.
If the `'java'` plugin is not applied, the configuration will fail during the `afterEvaluate`

This change makes it so that the `'java'` plugin is also applied.

## Overview

* I'm not sure how this should be documented in the release notes

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
+++ b/documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
@@ -19,6 +19,8 @@ on GitHub.
 * The JUnit Platform Gradle plugin now adds its dependencies with a fixed version (same as plugin
   version) instead of a dynamic versioning scheme (was `1.+`) by default to ensure reproducible
   builds.
+* The JUnit Platform Gradle plugin explicitly applies the Gradle built-in `java` plugin as
+  it has an implicit dependency on it being applied.
 * All `findMethods()` implementations in `ReflectionUtils` no longer return synthetic methods.
   Shadowed https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2[override-equal]
   methods are also no longer included in the result collection.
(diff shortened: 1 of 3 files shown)


marcphilipp:
How about the new `java-library` plugin? Does it also depend on the `java` plugin or is it a separate plugin?

mkobit:
@marcphilipp the `'java-library'` does depend on the `'java'` plugin, so that could be an option as well.

I believe that would introduce a requirement that consumers use at least Gradle 3.4, which would be up to you guys.

One thing to think about as well is that consumers may not be producing libraries if they are consuming the `JUnitPlatformPlugin`. They may be testing other things like a a service like Spring Boot, or in my case a [Ratpack](https://github.com/ratpack/ratpack) service.

I'm not sure if always using the `'java-library'` is the right way to move forward as I didn't see anything like this mentioned on the [blog post](https://blog.gradle.org/incremental-compiler-avoidance).

marcphilipp:
I didn't mean to suggest that we should always use the `java-library` plugin. On the contrary, I wanted to ensure that our plugin works with both `java` and `java-library`.

mkobit:
@marcphilipp Ahh, yes. It will work with the `'java-library'` plugin based upon my understanding (see this [comment](http://disq.us/p/1gjn0iw) and this part of the [Gradle source](https://github.com/gradle/gradle/blob/v3.4.1/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaLibraryPlugin.java#L44)). It could help to add a couple [GradleTestKit](https://docs.gradle.org/current/userguide/test_kit.html) test cases to the plugin to ensure some stability until the Gradle team can prioritize built-in support.I could add a couple of these as part of https://github.com/junit-team/junit5/issues/708 or as part of this PR.

marcphilipp:
Sounds good! Please proceed... 🙂

sormuras:
Upgraded master to Gradle 3.4.1 with https://github.com/junit-team/junit5/commit/aa56e995beb451d2965182c90e18f170b1b1ff53.

> It could help to add a couple GradleTestKit test cases to the plugin to ensure some stability [...]

That'd be nice. I tend to merge this PR as-is -- then the GradleTestKist tests would be parts of #708.

sormuras:
Only thing missing is a note in M4 release notes.

sormuras:
Thanks @mkobit !

PR #719, opened 2017-03-10
Fix regression for class hierarchies without synthetic methods

sormuras:
## Overview

Code introduced by https://github.com/junit-team/junit5/commit/82174c15aba026a52a2566ee8643af89917fc1a1 broke standard support for test class hierarchies without bridge or synthetic methods present.

This commit fixes the regression by comparing the default and generic string representation of the upper method. If they are equal, the upper method does not shadow the lower one.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
@@ -21,10 +21,13 @@ import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
+import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Member;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
+import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
 import java.net.URI;
 import java.nio.file.Files;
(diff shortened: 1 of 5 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -716,7 +719,11 @@ private static boolean isMethodShadowedBy(Method upper, Method lower) {
 		if (lower.getParameterCount() != upper.getParameterCount()) {
 			return false;
 		}
-		// Check for method sub-signatures.
+		// trivial case: parameter types exactly match
+		if (Arrays.equals(lower.getParameterTypes(), upper.getParameterTypes())) {
```
sbrannen: This is definitely a good (i.e., required) check to have in place.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -725,7 +732,27 @@ private static boolean isMethodShadowedBy(Method upper, Method lower) {
 				return false;
 			}
 		}
-		return true;
+		// lower is sub-signature of upper: check for generics in upper method
+		if (isGeneric(upper)) {
```
sbrannen: At a glance, it appears that this _may_ suffice, especially if the tests are all passing now. However, I can't be certain that it covers all use cases without a thorough investigation on my part. In any case, it appears that it might well be "_good enough_" for the time being.

Issue #520, opened 2016-09-21
TestIdentifier methods: isTest() and isContainer()


noopur2507:
TestIdentifier has these two methods: `isTest()` and `isContainer()`. As per my understanding, `isTest()` is used to identify a test case that is executed and `isContainer()` is used to identify a container of these test cases. Hence, a `@TestFactory` method is a container and not a test.

Is there anything else to the meaning of `isContainer()`? If not, these two are exactly opposite of each other and one of them can be removed.

Also, it would be useful to have another method called as `isTestFactory()`, which can be used to identify a container method. Or, is there any other way to identify a test factory method?

orfjackal:
Related: #407

bechte:
This issue is related with #508 and we might find a common solution. Therefore, I set it to blocked, until #508 is solved or closed.

marcphilipp:
Resolved in #720.

Issue #407, opened 2016-07-14
Does it make sense for a TestDescriptor to be both a test and a container?


orfjackal:
## Overview

In [Specsy](http://specsy.org/), it's not known before executing a test whether its TestDescriptor has children or not. The nested tests are discovered dynamically while the test is executed. This holds true also for the top-level test, which is the test class itself. Thus conceptually all TestDescriptors are both tests and containers. This caused a problem with IntelliJ IDEA test results (see [this comment](https://youtrack.jetbrains.com/issue/IDEA-148288#comment=27-1518123) and [its reply](https://youtrack.jetbrains.com/issue/IDEA-148288#comment=27-1518474)) and I had to do [a workaround](https://youtrack.jetbrains.com/issue/IDEA-148288#comment=27-1518757).

## Deliverables

This issue should be not closed until the following issues have been resolved.

- [ ] What does JUnit mean by the concepts of "test" and "container"?
- [ ] How should IDEs and other tools change their behavior based on those variables?
   - If they shouldn't change behavior, then why do the variables exist?
- [ ] What about the case of isTest()=false and isContainer()=false?
- [ ] What about defining them dynamically based on whether there are children? (I tried it. IntelliJ IDEA doesn't notice that the values change during test execution.)
- [ ] Would it be possible to remove both concepts, and just check whether a TestDescriptor has children?

marcphilipp:
Thanks for reaching out! These are all valid questions that we've to discuss in the team. 

`isTest()` is intended to be used for reporting, e.g. the number of tests executed. A descriptor should either be a test or a container for tests.

Since you can't decide a priori: Have you tried adding the descriptors dynamically?

orfjackal:
The descriptors are already added dynamically. That's the only way for Specsy to work, because all tests are discovered during the execution of their enclosing test. The "Syntax" and "Isolated Execution Model" chapters in http://specsy.org/documentation demonstrate nested tests (the Java version shows best how test registration is implemented as method calls).

orfjackal:
If isTest and isContainer should be mutually exclusive, then how about making it an enum? Or just remove one of them. In that case all descriptors in Specsy would be tests and at the same time they would have 0 to many children. The test class would also be a test and not a container, because the root spec (the `run` method or constructor) is conceptually no different from the child specs (calls to `spec` method).

bechte:
This issue is related with #508 and we might find a common solution. Therefore, I set it to blocked, until #508 is solved or closed.

[...]

PR #720, opened 2017-03-10
Redesign isTest() and isContainer() in TestDescriptor

### LINKED ISSUE #508: Allow TestDescriptor to state it will eventually register dynamic tests
## Overview

See the discussion in #492 for details.

## Related Issues

- #367
- #407 
- #492 
- #732 

## Deliverables
- [x] Investigate whether we can get rid of `isTest()` and `isContainer()`.
- [ ]  Introduce support for allowing a `TestDescriptor` to state that it will _eventually_ register dynamic tests, if still desirable.

## Initial Team Brainstorming

- For reporting, we somehow need to figure out whether or not a `TestDescriptor` points to a test. Otherwise, we won't be able to tell how many tests have been executed. Therefore, it might not be possible to get rid of `isTest()` or at least the concept of a "test".
- Allowing the dynamic registration of `TestDescriptor`s at any time (also within a test itself) makes the concept of a "container" superfluous. Every `TestDescriptor` may possibly be a container or at least become one (see also #407). Therefore, removing `isContainer()` seems to be reasonable.

_Concerning eventually registering dynamic tests:_

- While we might spawn new tests during runtime, it may be difficult to state whether or not additional tests may be registered. Only after test execution of the current `TestDescriptor` can we give a guaranteed answer that no or all children have been found.
- This is closely related to the extension mechanism and its implementation by a `TestEngine`. One might expect that a `TestDescriptor` `CHILD` below a `TestDescriptor` `PARENT` would inherit the context and state of `PARENT`, i.e. all `BeforeAll/Each` and `AfterAll/Each` statements will get executed. This might be difficult to guarantee (at least for the Jupiter Engine).
  - Having said that, a `hasBeenExecuted()` method might be a good alternative, but it doesn't help us with pruning the `TestPlan` up-front.

sormuras:
## Overview

Fixes #508 by replacing logic behind `isTest()` and `isContainer()` from two independent boolean properties into an enumeration based one:

```java
enum TestDescriptor.Type {

	ENGINE, CONTAINER, TEST, CONTAINER_AND_TEST;

	public boolean isContainer() {
		return this == ENGINE || this == CONTAINER || this == CONTAINER_AND_TEST;
	}

	public boolean isTest() {
		return this == TEST || this == CONTAINER_AND_TEST;
	}
}
```

Convenient and code duplicate remover methods `TestDescriptor.prune()` and `TestDescriptor.pruneTree()` are introduced.

---

---

### Definition of Done

- [x]  All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/DiscoverySelectorResolver.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/DiscoverySelectorResolver.java
@@ -64,7 +64,7 @@ public class DiscoverySelectorResolver {
 		request.getSelectorsByType(UniqueIdSelector.class).forEach(selector -> {
 			javaElementsResolver.resolveUniqueId(selector.getUniqueId());
 		});
-		pruneTree(engineDescriptor);
+		engineDescriptor.pruneTree();
 	}
 
 	private JavaElementsResolver createJavaElementsResolver(TestDescriptor engineDescriptor) {
@@ -75,15 +75,6 @@ public class DiscoverySelectorResolver {
 		resolvers.add(new TestFactoryMethodResolver());
 		resolvers.add(new TestTemplateMethodResolver());
 		return new JavaElementsResolver(engineDescriptor, resolvers);
-	}
-
-	private void pruneTree(TestDescriptor root) {
-		TestDescriptor.Visitor removeChildrenWithoutTests = (descriptor) -> {
-			if (!descriptor.isRoot() && !descriptor.hasTests()) {
-				descriptor.removeFromHierarchy();
-			}
-		};
-		root.accept(removeChildrenWithoutTests);
 	}
 
 }
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
@@ -149,20 +149,70 @@ public interface TestDescriptor {
 	}
 
 	/**
-	 * Determine if this descriptor describes a container.
+	 * Determine this descriptor's {@link Type}.
+	 *
+	 * @return the descriptor type; never {@code null}.
+	 * @see #isContainer()
+	 * @see #isTest()
 	 */
-	boolean isContainer();
+	Type getType();
+
+	/**
+	 * Determine if this descriptor describes a container.
+	 *
+	 * <p>This default implementation delegates to {@link Type#isContainer()}.
+	 */
+	default boolean isContainer() {
+		return getType().isContainer();
+	}
 
 	/**
 	 * Determine if this descriptor describes a test.
+	 *
+	 * <p>This default implementation delegates to {@link Type#isTest()}.
 	 */
-	boolean isTest();
+	default boolean isTest() {
+		return getType().isTest();
+	}
 
 	/**
 	 * Determine if this descriptor or any of its descendants describes a test.
+	 *
+	 * <p>This default implementation recursively calls itself until a descriptor
+	 * returns {@code true} to {@link #isTest()}; starting with this descriptor
+	 * instance.
 	 */
 	default boolean hasTests() {
-		return (isTest() || getChildren().stream().anyMatch(TestDescriptor::hasTests));
+		return isTest() || getChildren().stream().anyMatch(TestDescriptor::hasTests);
+	}
+
+	/**
+	 * Remove this descriptor from hierarchy unless it is a root or has tests.
+	 *
+	 * <p>An engine implementation may override this method and do it differently
+	 * or skip pruning altogether.
+	 *
+	 * @see #isRoot()
+	 * @see #hasTests()
+	 * @see #removeFromHierarchy()
+	 */
+	default void prune() {
+		if (isRoot() || hasTests()) {
+			return;
+		}
+		removeFromHierarchy();
+	}
+
+	/**
+	 * Remove this and descendant descriptors from hierarchy.
+	 *
+	 * <p>The default implementation passes the potentially overridden
+	 * {@link #prune()} as a {@link Visitor} to this instance. I.e. it removes
+	 * itself and descendants unless it is a root descriptor or itself is test
+	 * or any of its children describes a test.
+	 */
+	default void pruneTree() {
+		accept(TestDescriptor::prune);
 	}
 
 	/**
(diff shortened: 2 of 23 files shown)


bechte:
Are we really sure that we won't to have this exclusive or relationship? I can think of tests that contain tests themselves and therefore are test and container at the same time (e.g. spawning dyn. tests or running some kind of parameterized tests... - just as examples). 

I was also spending time in this topic and came to the conclusion it would be better to have test and container as separate things.

Not sure if this optimization is leading into the right direction.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -81,7 +81,7 @@
 
 	@Override
 	public boolean isLeaf() {
-		return !isContainer();
+		return isTest();
```
bechte: I doubt this assumption is really true for all tree leaves. Their might be leaves that are by definition containers, e.g. dynamic tests.

sormuras: That's why they, namely `TestFactoryTestDescriptor` and `TestTemplateTestDescriptor`, do override that method with `return true;`

sormuras:
>  [...] it would be better to have test and container as separate things.

I didn't find a single use-case within the affected source paths that used all 4 possible states of the double-boolean-property. Actually, all relevant parts relied on the xor-relation ship. Including test paths: no unit test had to be touched by this optimization. Another point is, that users are confused by the independent test/container relation ship: see #407 and #520.

gaganis:
I must admit that the approach @sormuras took for this feels more natural to me. When I was trying to reason on this myself I was more confused by the previous code.

bechte:
Well, I am happy to go that route if all are fine with it. 😄

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -159,9 +162,31 @@ default boolean isRoot() {
...
+	@API(Internal)
 	default boolean hasTests() {
-		return (isTest() || getChildren().stream().anyMatch(TestDescriptor::hasTests));
+		return isTest() || getChildren().stream().anyMatch(TestDescriptor::hasTests);
+	}
+
+	/**
+	 * Remove this descriptor from hierarchy unless it is a root descriptor or
+	 * itself is a test or any of its children describes a test.
+	 */
```
marcphilipp: This method actually opens up an interesting use case: An engine might override `prune` and do it differently or skip it altogether. I think that's actually a good idea! 👍 

However, I think the Javadoc should be more clear on this use case and should clarify that this is what the default implementation does.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/DiscoverySelectorResolver.java
```diff
@@ -64,7 +64,7 @@ public void resolveSelectors(EngineDiscoveryRequest request, TestDescriptor engi
 		request.getSelectorsByType(UniqueIdSelector.class).forEach(selector -> {
 			javaElementsResolver.resolveUniqueId(selector.getUniqueId());
 		});
-		pruneTree(engineDescriptor);
+		engineDescriptor.pruneTree();
```
marcphilipp: Do we really need to prune here (since DefaultLauncher will do it later anyway)?

gaganis: Wouldn't _If you you can do something today, why leave it  for tomorrow_ apply to this?

If DiscoverySelectorResolver can positively determine entries to be pruned why not do it?

sormuras: When the line is deleted, three `DiscoverySelectorResolverTests` fail:

- `resolvingSelectorOfNonTestMethodResolvesNothing()`
- `resolvingUniqueIdWithWrongParamsResolvesNothing()`
- `resolvingUniqueIdOfNonTestMethodResolvesNothing()`

What was the purpose to write these? Who is depending on the same assumption?

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
sormuras: Updating the Javadoc.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/DiscoverySelectorResolver.java
marcphilipp: I don't think anyone is depending on this assumption but let's move this investigation into a separate issue. Can you create one, please?

sormuras: Create https://github.com/junit-team/junit5/issues/732 -- and will leave the prune here alone.

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
sormuras: Are the updated Javadoc good enough?

sormuras:
@junit-team/junit-lambda  Writing the M4 notes I searched for usages of `TestDescriptor.isContainer()` and only found a single one that is not located in the test sources: `org.junit.vintage.engine.execution.RunListenerAdapter#handleFailure(...)` in line 93. Replace that with `!isTest()`, mark `TestDescriptor.isContainer()` as deprecated and remove it in M5?

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -148,28 +148,26 @@ default boolean isRoot() {
 	}
 
 	/**
-	 * Determine if this descriptor describes a container.
-	 *
-	 * <p>This default implementation negates the boolean result returned by
-	 * {@link #isTest()}. This makes being a test descriptor or a container
-	 * descriptor mutual exclusive.
+	 * Determine this descriptors' {@link Type}.
```
sbrannen: "this descriptor's"

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -148,28 +148,26 @@ default boolean isRoot() {
...
 	 *
-	 * @return {@code true} if this descriptor is a container; if it is a
-	 * test {@code false} is returned.
```
sbrannen: The `@return` needs to be reinstated pointing out that it can never be `null`.

sormuras:
Description updated.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
sormuras: @bechte I reverted that change.

sormuras:
Squashed and rebased onto master.

PR #723, opened 2017-03-11
JUnit 4 parameterized support as an extension (#715)

### LINKED ISSUE #715: Introduce migration support for JUnit 4's Parameterized Runner
## Overview

Introduce migration support for JUnit 4's `@RunWith(Parameterized.class)`.

@marcphilipp mentioned through email that it would be nice to have an extension in `junit-jupiter-migration-support` that allows tests to be run that were originally written with JUnit 4 using the `Parameterized` runner. 

What this would look like:

```java
import static org.junit.Assert.assertEquals;

import java.util.Arrays;
import java.util.Collection;
import org.junit.jupiter.api.TestTemplate;
import org.junit.jupiter.api.extension.ExtendWith;
import org.junit.jupiter.migrationsupport.extensions.Parameterized;
import org.junit.runners.Parameterized.Parameters;

@ExtendWith(ParameterizedExtension.class)
public class FibonacciTest {

    @Parameters
    public static Collection<Object[]> data() {
        return Arrays.asList(new Object[][] {     
                 { 0, 0 }, { 1, 1 }, { 2, 1 }, { 3, 2 }, { 4, 3 }, { 5, 5 }, { 6, 8 }  
           });
    }

    private int input;

    private int expected;

    public FibonacciTest(int input, int expected) {
        this.input= input;
        this.expected= expected;
    }

    @TestTemplate
    public void test() {
        assertEquals(expected, Fibonacci.compute(input));
    }
}
```

```java
public class Fibonacci {
    public static int compute(int n) {
    	int result = 0;
    	
        if (n <= 1) { 
        	result = n; 
        } else { 
        	result = compute(n - 1) + compute(n - 2); 
        }
        
        return result;
    }
}
```

If everybody agrees this is a good idea, I would like to tackle this issue.

## Deliverables

- [ ] Introduce `ParameterizedExtension` extension in `junit-jupiter-migration-support`.
- [ ] An updated section in the user guide.
- [ ] Document in release notes.

LiamClark:
## Overview
This allows running parameterized tests from JUnit 4 with Field and Constructor injection, fixes #715.
Fail silently when asked whether we can support the test case,
but fail explicitly with an exception when parameters can't be resolved.

It does this by combining a TestTemplateContextProvider that adds one context for every Object[] in the parameters Collection. These contexts contain an additional extension that performs the field injection if required.

However, when constructor injection is used, there needs to be a parameter resolver in place before the TestTemplateContextProvider is even triggered, therefore the parameter resolver is not an additional extension. Having the parameter resolver on the top level causes us to lose the context of which TemplateContext we were in and this information has to be regained by carefully counting the parameter resolutions. Is there a cleaner way to solve this?

I Store the results of the parameters factory function In the ClassContext store,
so we only call the parameters function once.

The tests are all integration tests, since doing unit tests resulted in unwieldy mocking, however, the amount of 'TestSubjectClasses' needed seems a bit much any idea's to clear this up?

This also ties in with the production code, I tried to keep everything as private as possible, maybe the tests could be simpler if I open up the access modifiers a bit, what would your design goals be here?

I tried to follow the behavior of JUnit 4 as closely as possible are there perhaps any behaviors missing?

Lastly, there is a lot of passing around of ExtensionContexts perhaps extracting a class to hold this could simplify the code a bit. 

---

### Definition of Done

### DIFF
--- a/junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/parameterized/ParameterizedExtension.java
+++ b/junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/parameterized/ParameterizedExtension.java
@@ -8,7 +8,7 @@
  * http://www.eclipse.org/legal/epl-v10.html
  */
 
-package org.junit.jupiter.migrationsupport.rules;
+package org.junit.jupiter.migrationsupport.parameterized;
 
 import static java.util.Collections.singletonList;
 import static java.util.function.Function.identity;
@@ -27,14 +27,13 @@ import java.util.Optional;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
-import org.junit.jupiter.api.extension.BeforeTestExecutionCallback;
 import org.junit.jupiter.api.extension.ContainerExtensionContext;
 import org.junit.jupiter.api.extension.Extension;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.api.extension.ParameterResolutionException;
 import org.junit.jupiter.api.extension.ParameterResolver;
-import org.junit.jupiter.api.extension.TestExtensionContext;
+import org.junit.jupiter.api.extension.TestInstancePostProcessor;
 import org.junit.jupiter.api.extension.TestTemplateInvocationContext;
 import org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider;
 import org.junit.platform.commons.meta.API;
@@ -215,7 +214,7 @@ public class ParameterizedExtension implements TestTemplateInvocationContextProv
 		};
 	}
 
-	private static class InjectionExtension implements BeforeTestExecutionCallback {
+	private static class InjectionExtension implements TestInstancePostProcessor {
 		private final Object[] parameters;
 
 		public InjectionExtension(Object[] parameters) {
@@ -223,7 +222,7 @@ public class ParameterizedExtension implements TestTemplateInvocationContextProv
 		}
 
 		@Override
-		public void beforeTestExecution(TestExtensionContext context) throws Exception {
+		public void postProcessTestInstance(Object testInstance, ExtensionContext context) throws Exception {
 			List<Field> parameters = parametersFields(context);
 
 			if (!parameters.isEmpty() && parameters.size() != this.parameters.length) {
@@ -233,7 +232,7 @@ public class ParameterizedExtension implements TestTemplateInvocationContextProv
 			for (Field param : parameters) {
 				Parameterized.Parameter annotation = param.getAnnotation(Parameterized.Parameter.class);
 				int paramIndex = annotation.value();
-				param.set(context.getTestInstance(), this.parameters[paramIndex]);
+				param.set(testInstance, this.parameters[paramIndex]);
 			}
 		}
 	}
--- a/junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/parameterized/ParameterizedExtension.java
+++ b/junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/parameterized/ParameterizedExtension.java
@@ -17,9 +17,9 @@ import static java.util.stream.Collectors.toList;
 import static org.junit.platform.commons.meta.API.Usage.Experimental;
 
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Executable;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -41,10 +41,9 @@ import org.junit.platform.commons.util.ReflectionUtils;
 import org.junit.runners.Parameterized;
 
 @API(Experimental)
-public class ParameterizedExtension implements TestTemplateInvocationContextProvider, ParameterResolver {
+public class ParameterizedExtension implements TestTemplateInvocationContextProvider {
 	private static ExtensionContext.Namespace parameters = ExtensionContext.Namespace.create(
 		ParameterizedExtension.class);
-	private int parametersCollectionIndex = 0;
 
 	/**
 	 * Indicate whether we can provide parameterized support.
@@ -54,86 +53,33 @@ public class ParameterizedExtension implements TestTemplateInvocationContextProv
 	 */
 	@Override
 	public boolean supports(ContainerExtensionContext context) {
-		return hasParametersMethod(context) && hasCorrectParameterFields(context);
+		return hasParametersMethod(context) && validInjectionMix(context);
+	}
+
+	private static boolean validInjectionMix(ExtensionContext context) {
+		List<Field> fields = parametersFields(context);
+		boolean hasParameterFields = !fields.isEmpty();
+		boolean hasCorrectParameterFields = areParametersFormedCorrectly(fields);
+		boolean hasArgsConstructor = hasArgsConstructor(context);
+
+		if (hasArgsConstructor) {
+			return !hasParameterFields;
+		}
+		else {
+			return !hasParameterFields || hasCorrectParameterFields;
+		}
 	}
 
 	@Override
 	public Stream<TestTemplateInvocationContext> provide(ContainerExtensionContext context) {
-		// grabbing the parent ensures the paremeters are stored in the same store.
+		// grabbing the parent ensures the parameters are stored in the same store across multiple TestTemplates.
 		return context.getParent().flatMap(ParameterizedExtension::parameters).map(
-			ParameterizedExtension::testTemplateContextsFromParameters).orElse(Stream.empty());
-	}
-
-	/**
-	 * Since the parameterized runner in JUnit 4 could only resolve constructor parameters
-	 * this extension once again here only support them on the constructor and require an {@code @Parameters} method
-	 *
-	 * @param parameterContext the context for the parameter to be resolved; never
-	 * {@code null}
-	 * @param extensionContext the extension context for the {@code Executable}
-	 * about to be invoked; never {@code null}
-	 * @return true if the above is met otherwise false.
-	 */
-	@Override
-	public boolean supports(ParameterContext parameterContext, ExtensionContext extensionContext) {
-		return hasParametersMethod(extensionContext)
-				&& parameterContext.getDeclaringExecutable() instanceof Constructor;
-	}
-
-	@Override
-	public Object resolve(ParameterContext parameterContext, ExtensionContext extensionContext)
-			throws ParameterResolutionException {
-		int parameterCount = parameterContext.getDeclaringExecutable().getParameterCount();
-		Object[] parameters = resolveParametersForConstructor(extensionContext, parameterCount);
-
-		int parameterIndex = parameterContext.getIndex();
-		// move to the next set of parametersFields
-		if (lastParameterToBeResolved(parameterContext)) {
-			this.parametersCollectionIndex++;
-		}
-
-		return parameters[parameterIndex];
-	}
-
-	/**
-	 * Retrieves the Object[] of the current iteration we are working on.
-	 *
-	 * @param extensionContext the extensionContext
-	 * @param parameterCount the amount of parameters of the constructor.
-	 *
-	 * @return the Object[] for this parameter iteration.
-	 * @throws ParameterResolutionException If the amount of arguments of the constructor doesn't match the amount
-	 * of arguments of the currently resolved object[]
-	 */
-	private Object[] resolveParametersForConstructor(ExtensionContext extensionContext, int parameterCount)
-			throws ParameterResolutionException {
-		// @formatter:off
-		return parameters(extensionContext).map(ArrayList::new)
-				.map(l -> l.get(this.parametersCollectionIndex))
-				.filter(params -> params.length == parameterCount)
-				.orElseThrow(ParameterizedExtension::unMatchedAmountOfParametersException);
-		// @formatter:on
-	}
-
-	private static boolean hasCorrectParameterFields(ExtensionContext context) {
-		List<Field> fields = parametersFields(context);
-		boolean hasFieldInjection = !fields.isEmpty();
-
-		if (hasArgsConstructor(context) && hasFieldInjection) {
-			return false;
-		}
-		else if (hasFieldInjection) {
-			return areParametersFormedCorrectly(fields);
-		}
-
-		return true;
+			o -> testTemplateContextsFromParameters(o, context)).orElse(Stream.empty());
 	}
 
 	private static boolean areParametersFormedCorrectly(List<Field> fields) {
 		List<Integer> parameterValues = parameterIndexes(fields);
-
 		List<Integer> duplicateIndexes = duplicatedIndexes(parameterValues);
-
 		boolean hasAllIndexes = indexRangeComplete(parameterValues);
 
 		return hasAllIndexes && duplicateIndexes.isEmpty();
@@ -164,10 +110,6 @@ public class ParameterizedExtension implements TestTemplateInvocationContextProv
 				.map(i -> parameterValues.containsAll(IntStream.range(0, i).boxed().collect(toList())))
 				.orElse(false);
 		// @formatter:on
-	}
-
-	private static boolean lastParameterToBeResolved(ParameterContext parameterContext) {
-		return parameterContext.getIndex() == parameterContext.getDeclaringExecutable().getParameterCount() - 1;
 	}
 
 	private static Optional<Collection<Object[]>> parameters(ExtensionContext context) {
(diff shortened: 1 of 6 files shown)


junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,266 @@
...
+	 *
+	 * @return the object[] for this parameter iteration.
+	 * @throws ParameterResolutionException If the amount of arguments of the constructor doesn't match the amount
+	 * of arguments of the currently resolved object[]
+	 */
+	private Object[] resolveParametersForConstructor(ExtensionContext extensionContext, int parameterCount)
+			throws ParameterResolutionException {
+		return parameters(extensionContext).map(ArrayList::new).map(l -> l.get(this.parameterCount.get())).filter(
+			params -> params.length == parameterCount).orElseThrow(
+				ParameterizedExtension::unMatchedAmountOfParametersException);
```
gaganis: I've had to put quite some effort to understand the code in this class because of the extensive use of lambdas. It is probably me... I am not smart enough or young enough to understand these(... well definitely both) :)

I was especially troubled by the above lambda chain. I think it would have helped me if `this.parameterCount` was named differently. In this case it has the same name as the `int parameterCount`. I think a more descriptive name would be something like `parametersCollectionSize` or `parameterArraysCount`

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,266 @@
...
+import org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider;
+import org.junit.platform.commons.meta.API;
+import org.junit.platform.commons.util.ReflectionUtils;
+import org.junit.runners.Parameterized;
+
+@API(Experimental)
+public class ParameterizedExtension implements TestTemplateInvocationContextProvider, ParameterResolver {
+	private static ExtensionContext.Namespace parameters = ExtensionContext.Namespace.create(
+		ParameterizedExtension.class);;
+	private AtomicInteger parameterCount = new AtomicInteger();
```
gaganis: Why did you use an AtomicInteger here? Do reads and writes need to be atomic for this?

Thanks

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
LiamClark: @gaganis ah yes definitely, that would be a much clearer name.

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
LiamClark: It's mainly because I had this piece of code in the back of my head :
https://github.com/junit-team/junit5/blob/master/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java#L93

but it seems that the AtomicInteger there is used to get around the effectively final rule rather than atomicity. 
It seems that resolve is called sequentially so it could just be a normal int.

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,265 @@
...
+import org.junit.jupiter.api.extension.TestTemplateInvocationContext;
+import org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider;
+import org.junit.platform.commons.meta.API;
+import org.junit.platform.commons.util.ReflectionUtils;
+import org.junit.runners.Parameterized;
+
+@API(Experimental)
+public class ParameterizedExtension implements TestTemplateInvocationContextProvider, ParameterResolver {
+	private static ExtensionContext.Namespace parameters = ExtensionContext.Namespace.create(
+		ParameterizedExtension.class);;
```
jbduncan: Two semi-colons? :)

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,265 @@
...
+	 * or to have a constructor that could be injected.
+	 */
+	@Override
+	public boolean supports(ContainerExtensionContext context) {
+		return hasParametersMethod(context) && hasCorrectParameterFields(context);
+	}
+
+	@Override
+	public Iterator<TestTemplateInvocationContext> provide(ContainerExtensionContext context) {
+		//grabbing the parent ensures the paremeters are stored in the same store.
```
jbduncan: No space between comment start and "grabbing"?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	@Override
+	public Object resolve(ParameterContext parameterContext, ExtensionContext extensionContext)
+			throws ParameterResolutionException {
+		int parameterCount = parameterContext.getDeclaringExecutable().getParameterCount();
+		Object[] parameters = resolveParametersForConstructor(extensionContext, parameterCount);
+
+		int parameterIndex = parameterContext.getIndex();
+		//move to the next set of parametersFields
```
jbduncan: No space between comment start and "move"?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,265 @@
...
+	 *
+	 * @return the object[] for this parameter iteration.
+	 * @throws ParameterResolutionException If the amount of arguments of the constructor doesn't match the amount
+	 * of arguments of the currently resolved object[]
+	 */
+	private Object[] resolveParametersForConstructor(ExtensionContext extensionContext, int parameterCount)
+			throws ParameterResolutionException {
+		return parameters(extensionContext).map(ArrayList::new).map(l -> l.get(this.parametersCollectionIndex)).filter(
+			params -> params.length == parameterCount).orElseThrow(
+				ParameterizedExtension::unMatchedAmountOfParametersException);
```
jbduncan: Consider using `// @formatter:{off|on}` to make this more readable:
```java
// @formatter:off
return parameters(extensionContext).map(ArrayList::new)
    .map(l -> l.get(this.parametersCollectionIndex))
    .filter(params -> params.length == parameterCount)
    .orElseThrow(ParameterizedExtension::unMatchedAmountOfParametersException);
// @formatter:on
```

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,265 @@
...
+		List<Integer> duplicateIndexes = duplicatedIndexes(parameterValues);
+
+		boolean hasAllIndexes = indexRangeComplete(parameterValues);
+
+		return hasAllIndexes && duplicateIndexes.isEmpty();
+	}
+
+	private static List<Integer> parameterIndexes(List<Field> fields) {
+		return fields.stream().map(f -> f.getAnnotation(Parameterized.Parameter.class)).map(
+			Parameterized.Parameter::value).collect(toList());
```
jbduncan: Consider using `// @formatter:{off|on}` here and all the other stream-using methods below.

junit-jupiter-migration-support/src/test/java/org/junit/jupiter/migrationsupport/ParametrizedExtensionTests.java
```diff
@@ -0,0 +1,308 @@
...
+		engine.execute(new ExecutionRequest(testDescriptor, eventRecorder, request.getConfigurationParameters()));
+		return eventRecorder;
+	}
+
+	private static List<Throwable> exceptionsThrown(ExecutionEventRecorder executionEventRecorder) {
+		return executionEventRecorder.getFailedTestFinishedEvents().stream().map(
+			it -> it.getPayload(TestExecutionResult.class)).map(
+				o -> o.flatMap(TestExecutionResult::getThrowable)).filter(Optional::isPresent).map(
+					Optional::get).collect(Collectors.toList());
+	}
```
jbduncan: Another case where `// @formatter:{off|on}` would help readability.

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,265 @@
...
+
+	/**
+	 * Retrieves the Object[] of the current iteration we are working on.
+	 *
+	 * @param extensionContext the extensionContext
+	 * @param parameterCount the amount of parameters of the constructor.
+	 *
+	 * @return the object[] for this parameter iteration.
+	 * @throws ParameterResolutionException If the amount of arguments of the constructor doesn't match the amount
+	 * of arguments of the currently resolved object[]
```
jbduncan: `object[]` instead of `Object[]`?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
gaganis: Hey, I rewrote TestTemplateTestDescriptor in #725 to remove the AtomicInteger. :)

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,294 @@
...
+	 * and correct {@code @Parameter} and their corresponding values
+	 * or to have a constructor that could be injected.
+	 */
+	@Override
+	public boolean supports(ContainerExtensionContext context) {
+		return hasParametersMethod(context) && hasCorrectParameterFields(context);
+	}
+
+	@Override
+	public Iterator<TestTemplateInvocationContext> provide(ContainerExtensionContext context) {
```
marcphilipp: I'm sorry, but I've changed `provide` to return a `Stream` in 410b47d68a199e371fc066f5274bb707938e5a79 which has just been merged into `master`. Please rebase this pull request and update this method.

LiamClark: @marcphilipp no problem at all, the iterator was being created from a stream anyways. So it actually simplified the code!

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,292 @@
...
+import org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider;
+import org.junit.platform.commons.meta.API;
+import org.junit.platform.commons.util.ReflectionUtils;
+import org.junit.runners.Parameterized;
+
+@API(Experimental)
+public class ParameterizedExtension implements TestTemplateInvocationContextProvider, ParameterResolver {
+	private static ExtensionContext.Namespace parameters = ExtensionContext.Namespace.create(
+		ParameterizedExtension.class);
+	private int parametersCollectionIndex = 0;
```
LiamClark: So this index should actually be saved in the store, since the extension is required to be stateless, however before changing that. I'd like to know if there is any way that might avoid this index.

marcphilipp: Hmm, instead of implementing `ParameterResolver` in the extension itself, can you return a `ParameterResolver` from `TestTemplateInvocationContext.getAdditionalExtensions()`?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,292 @@
...
+	private static TestTemplateInvocationContext contextFactory(Object[] parameters) {
+		return new TestTemplateInvocationContext() {
+			@Override
+			public List<Extension> getAdditionalExtensions() {
+				return singletonList(new InjectionExtension(parameters));
+			}
+		};
+	}
+
+	private static class InjectionExtension implements BeforeTestExecutionCallback {
```
marcphilipp: Shouldn't this be a `TestInstancePostProcessor` instead?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,292 @@
...
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.jupiter.migrationsupport.rules;
```
marcphilipp: This should reside in a different package. How about `org.junit.jupiter.migrationsupport.parameterized`?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,292 @@
...
+		// @formatter:on
+	}
+
+	private static boolean lastParameterToBeResolved(ParameterContext parameterContext) {
+		return parameterContext.getIndex() == parameterContext.getDeclaringExecutable().getParameterCount() - 1;
+	}
+
+	private static Optional<Collection<Object[]>> parameters(ExtensionContext context) {
+		return context.getStore(parameters).getOrComputeIfAbsent("parameterMethod",
+			k -> new ParameterWrapper(callParameters(context)), ParameterWrapper.class).getValue();
```
marcphilipp: There should be no need to store the parameters if we make all extensions "local" and handle them in `provide()`.

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
```diff
@@ -0,0 +1,292 @@
...
+		return new ParameterResolutionException("The @Parameters returns the wrong type");
+	}
+
+	@SuppressWarnings("unchecked")
+	private static Collection<Object[]> convertParametersMethodReturnType(Object o) {
+		if (o instanceof Collection) {
+			return (Collection<Object[]>) o;
+		}
+		else {
+			throw wrongParametersReturnType();
```
marcphilipp: `Parameters` methods may also return an `Iterable` or an `Object[]` (cf. https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/runners/Parameterized.java#L288-L297). You may want to use `CollectionUtils.toStream(o)` even though it supports additional types.

LiamClark:
@marcphilipp The reason the parameter resolver is not provided through the additional extensions, is because:

>However, when constructor injection is used, there needs to be a parameter resolver in place before the TestTemplateContextProvider is even triggered, therefore the parameter resolver is not an additional extension. Having the parameter resolver on the top level causes us to lose the context of which TemplateContext we were in and this information has to be regained by carefully counting the parameter resolutions. Is there a cleaner way to solve this?

If no parameter resolver is present for the first constructor the test fails before the test template can even be considered. Perhaps we can resolve only the first constructor invocation from the top level parameter resolver and supply the rest of the parameters through the additional extensions?

marcphilipp:
@LiamClark Sorry for not re-reading your comment properly. I now remember reading your comment when you opened the pull request. 😳

It sounds like you've actually tried returning the `ParameterResolver` from `getAdditionalExtensions()`, right?

LiamClark:
@marcphilipp No problem!

I didn't exactly try to return the parameter resolver from `getAdditionalExtensions()`. It was indeed my original idea, but this problem showed it self simply trying to get the extension callback to trigger.

It of course would be the nicest spot to resolve the parameters in, but it seems like the mismatch between Junit4 and Jupiter is slightly to big.

marcphilipp:
I'll look into it and get back to you.

marcphilipp:
Although I was quite sure that this should already be supported, I've now added a test that verifies that in bae2768d7f34bbc2310834393bd212ea1dd343b8: A `ParameterResolver` returned by a `TestTemplateInvocationContext` may be used to resolve constructor arguments.

@LiamClark Can you try to change the code in this PR accordingly?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
LiamClark: @marcphilipp It is moved into provide now :), but if there are multiple test templates in the class, it would still trigger the parameters method twice. 

However is calling the parameters method once something we care about?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
LiamClark: @marcphilipp is this only for the [single parameter case](https://github.com/junit-team/junit4/wiki/parameterized-tests#tests-with-single-parameter) or also in general?

marcphilipp: Both.

LiamClark: Done!

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
marcphilipp: You sure? You already store it in the parent context's `Store` but only if it's not already there, don't you? Can you add a test for that?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/parameterized/ParameterizedExtension.java
```diff
@@ -0,0 +1,268 @@
...
+import org.junit.jupiter.api.extension.TestTemplateInvocationContext;
+import org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider;
+import org.junit.platform.commons.meta.API;
+import org.junit.platform.commons.util.CollectionUtils;
+import org.junit.platform.commons.util.ReflectionUtils;
+import org.junit.runners.Parameterized;
+
+@API(Experimental)
+public class ParameterizedExtension implements TestTemplateInvocationContextProvider {
+	private static ExtensionContext.Namespace parameters = ExtensionContext.Namespace.create(
```
marcphilipp: This can also be `final` and then follow the naming convention for constants, right?

junit-jupiter-migration-support/src/main/java/org/junit/jupiter/migrationsupport/rules/ParameterizedExtension.java
LiamClark: @marcphilipp Yes, I already did here https://github.com/junit-team/junit5/pull/723/files#diff-c8fe8d61d6bba040fc43c22d9db3f16aR291 (Looking at it now it could definitely have a better name) 

So in the current implementation the parameters method is indeed only called once, but it does still require storing them in the store. My comment was intended to show that even though the extensions are now local, I still need to store them for this behaviour.

marcphilipp: Yeah, ok, I think that's acceptable in order to preserve the old behavior. 👍

sbrannen:
Hi @LiamClark,

Just wondering what the status of this effort is.

Can you shed any light on the subject?

marcphilipp:
I think this would be a perfect use case for container templates.

sbrannen:
> I think this would be a perfect use case for container templates.

Yep... could well be!

LiamClark:
So I believe all feedback in the last set of requested changes was actually implemented.

So mainly it would be adding it to the user guide, if you guys still want it.
It is however based on a now somewhat dated version of master. 

I'd be willing to do the rebase and fix the code or if you want to explore the possibilities to use container templates for this, I have some time the next few weeks to do that as well.

let me know!

sbrannen:
> So I believe all feedback in the last set of requested changes was actually implemented.

OK. Thanks for the feedback.

> I'd be willing to do the rebase and fix the code or if you want to explore the possibilities to use container templates for this, I have some time the next few weeks to do that as well.

Well, _container templates_ don't exist yet. So let's hold off on rebasing/reworking until we make progress on _container templates_.

marcphilipp:

PR #724, opened 2017-03-11
Explicitly catch RuntimeExceptions

JAdeLeeuw:
## Overview 
I noticed several instances where `Exception` was caught, but was not sure if this was done intentionally to also catch `RuntimeException` or not.

I decided to dig a bit deeper and found that this was indeed intentional. Since this was not clear (to me) from the start I thought it might be a good idea to make it explicit.

In `platform.commons.util.ClasspathScanner.java` I was unsure whether or not `RuntimeException` was a possibility, and therefore I included it to prevent accidentally changing behaviour.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ConditionEvaluator.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ConditionEvaluator.java
@@ -116,7 +116,7 @@ public class ConditionEvaluator {
 			logResult(condition.getClass(), result);
 			return result;
 		}
-		catch (Exception ex) {
+		catch (RuntimeException ex) {
 			throw evaluationException(condition.getClass(), ex);
 		}
 	}
@@ -127,7 +127,7 @@ public class ConditionEvaluator {
 			logResult(condition.getClass(), result);
 			return result;
 		}
-		catch (Exception ex) {
+		catch (RuntimeException ex) {
 			throw evaluationException(condition.getClass(), ex);
 		}
 	}
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java
@@ -20,6 +20,7 @@ import static org.junit.platform.commons.util.ClassFileVisitor.CLASS_FILE_SUFFIX
 
 import java.io.IOException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -78,7 +79,7 @@ class ClasspathScanner {
 			return packageName.isEmpty() // default package
 					|| getClassLoader().getResources(packagePath(packageName.trim())).hasMoreElements();
 		}
-		catch (Exception ex) {
+		catch (IOException | RuntimeException ex) {
 			return false;
 		}
 	}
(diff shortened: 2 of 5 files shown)


sbrannen:
`RuntimeException` is a subclass of `Exception`. Thus catching `Exception` allows us to catch all checked exceptions as well as unchecked exceptions (i.e., `RuntimeException`).

Even if there is not a test in place that fails as a result of your changes, your proposed changes will actually break JUnit since we are no longer catching exceptions properly.

I encourage you to read a tutorial on exception handling in Java, especially checked vs. unchecked exceptions.

I am closing this PR as "works as designed".

marcphilipp:
I think there is a valid reason not to catch `Exception`: `InterruptedException` isn't a `RuntimeException`. I guess that's a different issue, though.

gaganis:
When I saw this PR it more or less it seemed to not making much difference and then I read @sbrannen's comment below:

> Even if there is not a test in place that fails as a result of your changes, your proposed changes will actually break JUnit since we are no longer catching exceptions properly.

I was surprised reading this, and I still don't understand how it would eventually break. I would be immensely greatful @sbrannen, if you could elaborate a bit more.

The only thing that I can think of is that at the point where we catch `Exception` it created a clear statement about where we put a barrier for all errors between called and caller code. But if that was the case then we would need to catch `Throwable` so as to also catch `Errors` eg `OOME` 

Thanks.

sbrannen:
> I was surprised reading this, and I still don't understand how it would eventually break. I would be immensely greatful @sbrannen, if you could elaborate a bit more.

Obviously my choice or words was less than stellar. So I apologize for that.

What I actually _meant_ is that the intended exception policy of JUnit _might_ break by making the proposed changes.

In general, we do not catch specific types of exceptions simply because there is nothing we could or should do based on that additional information. The fact that an exception is an `Exception` or `RuntimeException` is generally sufficient for the needs of a testing framework.

> The only thing that I can think of is that at the point where we catch `Exception` it created a clear statement about where we put a barrier for all errors between called and caller code.

Yes, that's the point: we are catching `Exception` in those cases where we think it makes sense -- where we could potentially do something about it (like log and swallow it).

Normally we just let exceptions propagate up the call stack. That may seem counter intuitive to many, but for a testing framework like JUnit this behavior is often quintessential.

For example, we even have a special _trick_ in place to assist us in achieving exactly that: `ExceptionUtils.throwAsUncheckedException()`.

> But if that was the case then we would need to catch `Throwable` so as to also catch `Errors` eg `OOME` 

No. We explicitly do _not_ want to catch such errors. That's another use case for `ExceptionUtils.throwAsUncheckedException()`, combined with `BlacklistedExceptions.rethrowIfBlacklisted()`. Note that `OutOfMemoryError` is blacklisted.

Does that clarify things?

sbrannen:
Furthermore, just because a method signature does not indicate that it throws a checked `Exception`, it may in fact do so simply because of our transparent use of `ExceptionUtils.throwAsUncheckedException()`. Thus, catching `Exception` instead of `RuntimeException` is essential in such use cases.

sbrannen:
Long story, short: exception handling in JUnit 5 is not comparable with exception handling in a standard Java application; we have special needs.

I'm not saying that the exception handling in place is perfect: I'm sure there are _bugs_ lurking around in the code base. But... inspecting JUnit code as if it were _normal_ Java code is insufficient when determining the _right thing to do_. 😉 

So with all of this in mind, I prefer to err on the side of caution and not change the current exception handling unless there is a known bug that needs to be addressed.

gaganis:
@sbrannen Thank you very much for your extensive and very informative response!

>Does that clarify things?

Yes, absolutely!

>But... inspecting JUnit code as if it were normal Java code is insufficient when determining the right thing to do.

The above is what I missed when thinking about the OOME. JUnit can definitely not make an informed decision when seeing an error like this.

JAdeLeeuw:
@sbrannen I would also like to thank you for the additional explanation. Your initial explanation left me puzzeled but this clarifies everything.

sbrannen:
@JAdeLeeuw, you're very welcome! And I apologize for being a bit terse in my initial response. 😉

Issue #378, opened 2016-07-05
Enhance support for dynamic tests


sbrannen:
## Status Quo

Since 5.0 M1, dynamic tests can be registered as lambda expressions, but there are a few limitations with the current feature set.

## Open Issues

1. Lifecycle callbacks and extensions are not applied _around_ the invocation of a dynamic test.
1. A dynamic test cannot directly benefit from parameter resolution.

## Deliverables

Address each _open issue_.

mkobit:
I think it could also be useful/interesting for extensions in this case to do be able to compute "test matrix" by using multiple different parameter resolver extensions and different declarative parameterized values.

For example:

```java
@Test
@ForAValue({3, 4})
@ForBValue({5, 6})
void runMyTest(AClass aValue, BClass bValue) { ... }
```

I would like to have the ability to have an extension that can register additional tests by looking at the context of a method or class.

smoyer64:
@mkobit Take a look at PR #577 and the associated proposal in issue #354 ... the idea is to be able to alter the test plan before it's made immutable.

nicolaiparlog:
Regarding lifecycle integration for dynamic tests... What exactly should that look like? Would the whole extension shebang (test instance processing, test parameter injection, before/after, ...) be run for test factory methods and then again for each individual dynamic test?

Mmmh, now that I'm writing this I don't actually see a problem with that anymore. But I'm sure I saw one when creating #530... Damn brain! :tired_face:

sbrannen:
> Would the whole extension shebang (test instance processing, test parameter injection, before/after, ...) be run for test factory methods and then again for each individual dynamic test?

That's certainly what I have in mind!

ledoyen:
Here is a proposition for lifecycle hooks invocations around dynamic tests : #735

sormuras:
Related issues updated.

paulmiddelkoop:
At the moment my team is using JUnit 4 in all our tests except when we we need data driven tests. We use Spock there and were very happy with table based testing using `where`. Writing Groovy tests inside a Java project worked well for us. But since we switched to Kotlin this does not work very well (for instance using Kotlin data classes in Groovy is kind of weak). 

So we switched to JUnit 5 and see if we could use the experimental parameterized testing functionality. Personally I don't like if the parameterized data of your test is hidden away in a separate method, class or CSV file. I think one the powers of Spock is that the table is part of the test method.
[...]

Issue #60, opened 2015-12-11
Opt-in support for parallel execution


marcphilipp:
Related to #90.

ptahchiev:
+1 for this one. TestNG has had this feature for a long time and the maven surefire plugin supports parallel tests out of the box:

http://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html

radai-rosenblatt:
+1. would definitely help in shortening build times

leonard84:
Will this be part of the platform, or the engine, or both?

While thinking about better support for parallel execution in Spock, I identified some problems that should be handled by the testing framework.

* Parallel (integration) testing in conjunction with global state (system properties/db/...) can leak into other tests
  * e.g., Spock has the `@RestoreSystemProperties` extension, which will save the `SystemProperties` before the execution and restore it after the test. This works fine for single threaded execution, but in concurrent test execution this will lead to race conditions.
  * The bad solution would be to disable concurrent execution

My idea is that each test (class/method) could declare its use of global resources, e.g. `@UseResource(value="SystemProperties", mode=ReadWrite)`. The engine would keep a list of `ReentrantReadWriteLock` for each of the resources and acquire/release those locks accordingly.  Care would have to be taken to avoid deadlocks, by ordering the lock acquisition accordingly. 

This way the framework could offer the maximum concurrency while still offering providing tools to isolate problematic tests.

There would be a list of predefined resources, such as `Global` (all tests would implicitly acquire a ReadLock for this) or `SystemProperties`, but users can define their own to match their application.

Thoughts? Of course all names are open for change.

marcphilipp:
I think an engine needs to decide on which level tests can be executed in parallel. However, there are quite a few similarities between different testing frameworks where the platform could provide support, e.g. via a base implementation that multiple test engines can use.

I really like your reentrant lock idea! Is there already a `TestEngine` implementation for Spock? If so, do you use the `HierarchicalTestEngine` abstraction? I wrote a `ForkJoinPool` based prototype for it some time ago.

leonard84:
@marcphilipp no Spock doesn't have it's own `TestEngine` yet. As better support for parallelism would be a major part in it, I'd wanted to check in on this first before running in the wrong direction/doing redundant work. Switching to a own `TestEngine` will be done with Spock 2.0

wlsc:
Any news on that topic?

marcphilipp:
[...]

PR #735, opened 2017-03-13
Invoke lifecycle hooks around each dynamic test

ledoyen:
## Overview

`@BeforeEach`, `@AfterEach` and associated callback extensions run once around `@TestFactory` methods.

The proposition here is to run these hooks **around each dynamic tests** created by the factory method.

Hooks are not invoked around the factory method, because it is a container and type (container or test) cannot be determined from a hook implementation.

---

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodTestDescriptor.java
@@ -16,6 +16,7 @@ import java.lang.reflect.Method;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
@@ -108,6 +109,12 @@ public class MethodTestDescriptor extends MethodBasedTestDescriptor {
 	@Override
 	public JupiterEngineExecutionContext execute(JupiterEngineExecutionContext context,
 			DynamicTestExecutor dynamicTestExecutor) throws Exception {
+		executeTestBetweenBeforeAndAfterHooks(context, c -> invokeTestMethod(c, dynamicTestExecutor));
+		return context;
+	}
+
+	protected void executeTestBetweenBeforeAndAfterHooks(JupiterEngineExecutionContext context,
+			Consumer<JupiterEngineExecutionContext> testExecutor) {
 		ThrowableCollector throwableCollector = context.getThrowableCollector();
 
 		// @formatter:off
@@ -117,7 +124,7 @@ public class MethodTestDescriptor extends MethodBasedTestDescriptor {
 				if (throwableCollector.isEmpty()) {
 					invokeBeforeTestExecutionCallbacks(context);
 					if (throwableCollector.isEmpty()) {
-						invokeTestMethod(context, dynamicTestExecutor);
+						testExecutor.accept(context);
 					}
 					invokeAfterTestExecutionCallbacks(context);
 				}
@@ -127,8 +134,6 @@ public class MethodTestDescriptor extends MethodBasedTestDescriptor {
 		// @formatter:on
 
 		throwableCollector.assertEmpty();
-
-		return context;
 	}
 
 	private void invokeBeforeEachCallbacks(JupiterEngineExecutionContext context) {
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
@@ -20,6 +20,7 @@ import org.junit.jupiter.api.DynamicTest;
 import org.junit.jupiter.api.extension.TestExtensionContext;
 import org.junit.jupiter.engine.execution.ExecutableInvoker;
 import org.junit.jupiter.engine.execution.JupiterEngineExecutionContext;
+import org.junit.jupiter.engine.execution.ThrowableCollector;
 import org.junit.platform.commons.JUnitException;
 import org.junit.platform.commons.meta.API;
 import org.junit.platform.commons.util.CollectionUtils;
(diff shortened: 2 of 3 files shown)


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodTestDescriptor.java
```diff
@@ -108,27 +109,31 @@ public SkipResult shouldBeSkipped(JupiterEngineExecutionContext context) throws
...
+		invokeBetweenBeforeAndAfter(context, c -> invokeTestMethod(c, dynamicTestExecutor));
+		return context;
+	}
+
+	protected void invokeBetweenBeforeAndAfter(JupiterEngineExecutionContext context,
+			Consumer<JupiterEngineExecutionContext> action) {
 		ThrowableCollector throwableCollector = context.getThrowableCollector();
 
 		// @formatter:off
 		invokeBeforeEachCallbacks(context);
```
sbrannen: Please don't mess with the _indentation art_.

In other words, please retain the original indentation so that diffs are easier to follow.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodTestDescriptor.java
```diff
@@ -108,27 +109,31 @@ public SkipResult shouldBeSkipped(JupiterEngineExecutionContext context) throws
 	@Override
 	public JupiterEngineExecutionContext execute(JupiterEngineExecutionContext context,
 			DynamicTestExecutor dynamicTestExecutor) throws Exception {
+		invokeBetweenBeforeAndAfter(context, c -> invokeTestMethod(c, dynamicTestExecutor));
+		return context;
+	}
+
+	protected void invokeBetweenBeforeAndAfter(JupiterEngineExecutionContext context,
```
sbrannen: I don't like this method name: it does not align nicely with the existing methods in that class.

Something like `invokeTest()` or `executeTest()` would make more sense here.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodTestDescriptor.java
```diff
@@ -108,27 +109,31 @@ public SkipResult shouldBeSkipped(JupiterEngineExecutionContext context) throws
...
+			Consumer<JupiterEngineExecutionContext> action) {
```
sbrannen: The term `action` here is insufficient to convey its purpose.

Something like `testInvoker` or `testExecutor` would be more suitable.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
```diff
@@ -69,6 +70,18 @@ public boolean isLeaf() {
 	}
 
 	@Override
+	public JupiterEngineExecutionContext execute(JupiterEngineExecutionContext context,
```
sbrannen: Please explain *in great detail* the need to override the `execute()` method, not only here but also in the commit message.

sbrannen:
Thanks for the PR! 👍 

I only had a few minutes to briefly review it within the browser. So if my review comments are vague, please ask for clarification.

sbrannen:
I have some serious concerns about management of the `JupiterEngineExecutionContext`.

Specifically:

What happens if a dynamic test results in a failure (i.e., exception thrown)? Does the `ThrowableCollector` get reset?

What happens if an extension stores state in the `ExtensionContext.Store`? Does the `Store` get cleared before the next dynamic test is executed?

In general, in JUnit Jupiter it is required that each execution of a node in the test tree gets a new execution context.

So without explicit tests in place that confirm proper behavior for the above requirement, we will not be able to merge this PR.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
ledoyen: I choose to override the `execute()` method here to prevent calls to _before_ and _after_ hooks around a container execution.
This is because a hook implementation can not know on which (of a container or a test) it is executed.
I assumed that `@BeforeEach` meant before each test, in this particular case, unlike what's said in the annotations javadoc.


A cleaner way can be to declare a new annotation `@BeforeEachTest` and let callback implementations know about the type of test (container or not).

If this is something you prefer, I can give it a shot.

ledoyen:
I understand.

To answer to these questions and provide a solution, can you tell be me if dynamic tests are supposed to be executable in parallel (as the executor is stream-based).

I was not able to find in the user guide or in the javadoc, if there is any concerns about parallelism and how this should be handled in `TestDescriptor`s.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
sbrannen: Unless I'm mistaken, lifecycle callbacks currently apply to `@TestFactory` methods.

So it appears that you have removed this functionality, which some people may depend on.

sbrannen: At the moment, no, I don't think we want to introduce any additional lifecycle annotations.

sbrannen:
> To answer to these questions and provide a solution, can you tell be me if dynamic tests are supposed to be executable in parallel (as the executor is stream-based).

At the moment, no.

There is currently zero support for parallel test execution with JUnit Jupiter. See #60.

So, please do not add support related to parallel test execution in conjunction this this PR.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
ledoyen: What do you think about invoking lifecycle callbacks around `@TestFactory` methods _and_ around each dynamic tests and making implementations aware of the type (container or test) they are invoked on ?
This could be done by adding an `isContainer:boolean` method to the `TestExtensionContext` like on the `TestDescriptor` interface.

ledoyen:
It's been a while since I have time to spend on this.

After considerations, it may not be such a good idea to try micro-management around `DynamicTest`s.
GIven the other issues pointed-out in #378 , I think I will focus on the use `@TestTemplate`.

I suggest that we close this PR if that is ok with you.

marcphilipp:
Sure, thanks for getting back to us!

PR #736, opened 2017-03-14
Assertions.assertLinesMatch(List<String>, List<String>) proposal

sormuras:
## Overview

This is PR promotes an assertion utility developed for internal `List<String>` comparsion to the public Jupiter API as `org.junit.jupiter.api.Assertions.assertLinesMatch(List<String> expected, List<String> actual)`

Simple example:
```
List<String> expected = Arrays.asList("^first.+line", ".[[oO)", "last line$");
List<String> actual = Arrays.asList("first line", ".[[oO)", "last line");
assertLinesMatch(expected, actual);
```
File/System.out-based [example](https://github.com/junit-team/junit5/blob/assert_lines_match/platform-tests/src/test/java/org/junit/platform/console/ConsoleDetailsTests.java#L207):
```
List<String> expectedLines = Files.readAllLines(path, UTF_8);
List<String> actualLines = asList(result.out.split("\\R"));

assertLinesMatch(expectedLines, actualLines);
```

## Motivation

The new assertion enhances the already existing `assertIterableEquals(Iterable<?> expected, Iterable<?> actual)` by supporting the following "match outline" in sequence:

1. test simple `String` equality fur current expected and actual line
2. treat the expected line as a regular expression pattern and match the actual line
3. interpret fast-forward marker `>> [limit | or some arbitrary comment ] >>` to skip a (limited) block of actual lines. If the fast-forward limit is not given, the skipped block grows until the next expected line is found.

See [AssertionsAssertLinesMatchTests.java](https://github.com/junit-team/junit5/blob/assert_lines_match/junit-jupiter-engine/src/test/java/org/junit/jupiter/api/AssertionsAssertLinesMatchTests.java) for more usage examples.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
+++ b/documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
@@ -104,7 +104,10 @@ on GitHub.
 * `TestInstancePostProcessors` registered on test methods are now invoked.
 * There are two new signatures for `Assertions.fail`: `Assertions.fail(Throwable cause)` and
   `Assertions.fail(String message, Throwable cause)`.
-
+* New `Assertions.assertLinesMatch()` comparing lists of `String` featuring `Object::equals` and
+  regular expression checks. It also provides a fast-forward mechanism to skip actual lines that
+  are expected to change in each invocation, e.g. duration, timestamps and stack traces.
+  Consult the Javadoc for `{Assertions}` for details.
 
 [[release-notes-5.0.0-m4-junit-vintage]]
 ==== JUnit Vintage
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
@@ -12,8 +12,9 @@ package org.junit.jupiter.api;
 
 import static java.lang.String.format;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.platform.commons.util.Preconditions.condition;
+import static org.junit.platform.commons.util.Preconditions.notNull;
 
 import java.util.Deque;
 import java.util.LinkedList;
@@ -21,9 +22,6 @@ import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
-
-import org.junit.platform.commons.util.Preconditions;
-import org.opentest4j.AssertionFailedError;
 
 /**
  * {@code AssertLinesMatch} is a collection of utility methods that support asserting
@@ -34,8 +32,8 @@ import org.opentest4j.AssertionFailedError;
 class AssertLinesMatch {
 
 	static void assertLinesMatch(List<String> expectedLines, List<String> actualLines) {
-		Preconditions.notNull(expectedLines, "expectedLines must not be null");
-		Preconditions.notNull(actualLines, "actualLines must not be null");
+		notNull(expectedLines, "expectedLines must not be null");
+		notNull(actualLines, "actualLines must not be null");
 
 		// trivial case: same list instance
 		if (expectedLines == actualLines) {
@@ -56,14 +54,16 @@ class AssertLinesMatch {
 
 		// simple case: both list are equally sized, compare them line-by-line
 		if (expectedSize == actualSize) {
-			try {
-				for (int i = 0; i < expectedSize; i++) {
-					assertMatches(expectedLines.get(i), actualLines.get(i), i, i);
+			boolean allOk = true;
+			for (int i = 0; i < expectedSize; i++) {
+				if (matches(expectedLines.get(i), actualLines.get(i))) {
+					continue;
 				}
-				return;
+				allOk = false;
+				break;
 			}
-			catch (AssertionFailedError ignore) {
-				// fall through and try with fast-forward support
+			if (allOk) {
+				return;
 			}
 		}
 
@@ -73,27 +73,37 @@ class AssertLinesMatch {
 	private static void assertLinesMatchWithFastForward(List<String> expectedLines, List<String> actualLines) {
 		Deque<String> expectedDeque = new LinkedList<>(expectedLines);
 		Deque<String> actualDeque = new LinkedList<>(actualLines);
+
 		while (!expectedDeque.isEmpty()) {
 			String expectedLine = expectedDeque.pop();
 			String actualLine = actualDeque.peek();
+
 			// trivial case: take the fast path when they simply match
-			if (matches(expectedLine, actualLine, false)) {
+			if (matches(expectedLine, actualLine)) {
 				actualDeque.pop();
 				continue;
 			}
-			// fast-forward markers found in expected line: fast-forward actual line...
+
+			// fast-forward marker found in expected line: fast-forward actual line...
 			if (isFastForwardLine(expectedLine)) {
 				int fastForwardLimit = parseFastForwardLimit(expectedLine);
-				// value was given
+
+				// trivial case: fast-forward marker was in last expected line
+				if (expectedDeque.isEmpty()) {
+					int actualRemaining = actualDeque.size();
+					// no limit given or perfect match? we're done.
+					if (fastForwardLimit == Integer.MAX_VALUE || fastForwardLimit == actualRemaining) {
+						return;
+					}
+					fail(format("terminal fast-forward(%d) error: fast-forward(%d) expected", fastForwardLimit,
+						actualRemaining));
+				}
+
+				// peek next expected line
+				expectedLine = expectedDeque.peek();
+
+				// fast-forward limit was given: use it
 				if (fastForwardLimit != Integer.MAX_VALUE) {
-					if (fastForwardLimit > actualDeque.size()) {
-						fail(format("fast-forward %d lines failed, only %d lines left", fastForwardLimit,
-							actualDeque.size()));
-					}
-					if (expectedDeque.isEmpty()) {
-						assertEquals(fastForwardLimit, actualDeque.size(), "wrong number of actual lines remaining");
-						return; // perfect match
-					}
 					// fast-forward now: actualDeque.pop(fastForwardLimit)
 					for (int i = 0; i < fastForwardLimit; i++) {
 						actualDeque.pop();
@@ -101,43 +111,37 @@ class AssertLinesMatch {
 					if (actualDeque.isEmpty()) {
 						fail(format("%d more lines expected, actual lines is empty", expectedDeque.size()));
 					}
+					continue;
 				}
-				else {
-					if (expectedDeque.isEmpty()) {
-						return; // ignore all remaining actual lines
+
+				// fast-forward "unlimited": until next match
+				while (true) {
+					if (actualDeque.isEmpty()) {
+						fail(format("no match for `%s` line fast-forwarding: %n%s", expectedLine, actualLines));
 					}
-					// scan actual lines deque for next match
-					while (!actualDeque.isEmpty()) {
-						actualLine = actualDeque.pop();
-						if (matches(expectedDeque.peek(), actualLine, false)) {
-							actualDeque.push(actualLine);
-							break;
-						}
+					if (matches(expectedLine, actualDeque.pop())) {
+						break;
 					}
 				}
-				expectedLine = expectedDeque.pop();
 			}
-			// now, assert equality of current expected and actual line
-			int expectedIndex = expectedLines.size() - expectedDeque.size();
-			int actualIndex = actualLines.size() - actualDeque.size();
-			assertMatches(expectedLine, actualDeque.pop(), expectedIndex, actualIndex);
 		}
-		if (actualDeque.isEmpty()) {
-			return;
+
+		// after math
+		if (!actualDeque.isEmpty()) {
+			fail("more actual lines than expected: " + actualDeque.size());
 		}
-		fail("more actual lines than expected: " + actualDeque.size());
 	}
 
 	private static boolean isFastForwardLine(String line) {
 		line = line.trim();
-		return line.equals("S T A C K T R A C E") || line.startsWith(">>") && line.endsWith(">>");
+		return line.startsWith(">>") && line.endsWith(">>");
 	}
 
 	private static int parseFastForwardLimit(String fastForwardLine) {
 		String text = fastForwardLine.trim().substring(2, fastForwardLine.length() - 2).trim();
 		try {
 			int limit = Integer.parseInt(text);
-			Preconditions.condition(limit > 0, "fast-forward must greater than zero, it is: " + limit);
+			condition(limit > 0, "fast-forward must greater than zero, it is: " + limit);
 			return limit;
 		}
 		catch (NumberFormatException e) {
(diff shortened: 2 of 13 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
```diff
@@ -0,0 +1,165 @@
...
+
+		// trivial case: same list instance
+		if (expectedLines == actualLines) {
+			return;
+		}
+
+		int expectedSize = expectedLines.size();
+		int actualSize = actualLines.size();
+
+		// trivial case: when expecting more then actual lines available, something is wrong
```
jbduncan: Minor typo at "then", which I think should be "than".

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	private static boolean matches(String expectedLine, String actualLine) {
+		if (expectedLine.equals(actualLine)) {
+			return true;
+		}
+		try {
+			Pattern pattern = Pattern.compile(expectedLine);
+			Matcher matcher = pattern.matcher(actualLine);
+			return matcher.matches();
```
jbduncan: Have you considered using the shorthand `return expectedLine.matches(actualLine);` for brevity?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
sormuras: Is there a similar site like: http://www.seidseit.de for German? Ah, there is: http://thenthan.com ;)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
sormuras: I split the calls to cache the compiled patterns... but didn't implement that.

sormuras:
> On a sort of side note, have you considered allowing assertLinesMatch to accept Iterable<String> or Iterable<?> instead of just List<String>?

Ja. Tried that and reverted back to `List<String>`. The `Collections.size()` getter is used heavily by the implementation and I wanted to distinguish its API against the `assertIterableEquals(...)` one(s).

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
jbduncan: ;)

Looking at how http://thenthan.com describes the difference between the two words, I think that "than" is the correct word to use here, since AFAICT we're talking about a comparison between "expected lines" and "actual lines", rather than specifying that something about "expected lines" happens or should happen before "actual lines".

Unless, have I misunderstood the intention of this comment?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
jbduncan: Okay, cool. I'm happy either way (that is, either having the `Pattern`s compiled and cached, or using `String::matches` for brevity).

jbduncan:
> Ja. Tried that and reverted back to List<String>. The Collections.size() getter is used heavily by the implementation and I wanted to distinguish its API against the assertIterableEquals(...) one(s).

Sounds fair. I think wanting to distinguish its API against `assertIterableEquals(...)` is a good reason for wanting to accept `List<String>` rather than plain iterables. :+1:

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
sormuras: I change it to `return actualLine.matches(expectedLine);` (the argument is the regex) and will split/cache if it is a performance gain later.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
```diff
@@ -0,0 +1,161 @@
...
+				return;
+			}
+		}
+
+		assertLinesMatchWithFastForward(expectedLines, actualLines);
+	}
+
+	private static void assertLinesMatchWithFastForward(List<String> expectedLines, List<String> actualLines) {
+		Deque<String> expectedDeque = new LinkedList<>(expectedLines);
+		Deque<String> actualDeque = new LinkedList<>(actualLines);
```
jbduncan: Is there any particular reason that `LinkedList`s were used instead of `ArrayDeque`s (which I believe usually have better performance)?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,38 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
 		AssertIterableEquals.assertIterableEquals(expected, actual, messageSupplier);
 	}
 
+	// --- assertLinesMatch ----------------------------------------------------
+
+	/**
+	 * <em>Asserts</em> that {@code expected} list of {@linkplain String}s match {@code actual}
```
jbduncan: "match" -> "matches"

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,38 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
...
+	 * list.
+	 *
+	 * <p>This methods differs from other assertions, that effectively only check
```
jbduncan: "methods" -> "method"

jbduncan: I'm not sure that the comma here is needed. :/

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,38 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
...
+	 * {@link String#equals(Object)}, by using following staged match-making algorithm:
```
jbduncan: "following" -> "the following"

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,38 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
...
+	 *
+	 * <p>For each pair of expected and actual lines do
+	 * <ol>
+	 *   <li>check if {@code expected.equals(actual)} - if yes, continue with next pair</li>
+	 *   <li>treat {@code expected} as a regular expression and check via {@link String#matches(String)}
```
jbduncan: I think this should probably say, "_otherwise_ check...".

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,38 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
...
+	 *   - if yes, continue with next pair</li>
+	 *   <li>check if {@code expected} is a fast-forward marker, if yes apply fast-forward actual lines
```
jbduncan: Ditto regarding "_otherwise_ check...".

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,38 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
...
+	 * </ol>
+	 *
+	 * <p>A valid fast-forward marker is an expected line that starts and ends with the literal
+	 * {@code >>} and contains at least 4 characters. Examples:
+	 * <ul>
+	 *   <li>{@code >>>>}<br>{@code >> stacktrace >>}<br>{@code >> single line, non Integer.parse()-able comment >>}
+	 *   <br>Skip arbitrary number of actual lines, until first matching subsequent expected line is found. Any
+	 *   character between the fast-forward literals are discarded.</li>
+	 *   <li>{@code ">> 21 >>"}
+	 *   <br>Skip strictly 21 lines. If they can't be skipped for any reason, an error is raised.</li>
```
jbduncan: What sort of error is raised? :)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,38 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
...
+	 * </ul>
+	 */
```
jbduncan: I think it's worth giving some examples of how fast-forwarding works (or to provide a link to the tests for this assertion, which themselves provide nice examples IMO).

It's not really clear just from the description what fast-forwarding does, but as I'm not a master of the English language (despite being a native speaker!), I'm struggling to give useful feedback to improve the description itself, hence why I think examples would be useful.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -873,8 +873,34 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
 
 	// --- assertLinesMatch ----------------------------------------------------
 
-	public static void assertLinesMatch(List<String> expected, List<String> actual) {
-		AssertLinesMatch.assertLinesMatch(expected, actual);
+	/**
+	 * <em>Asserts</em> that {@code expected} list of {@linkplain String}s match {@code actual}
```
jbduncan: Sorry, this was a duplicate of another comment!

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
sormuras: Thanks for the review! Applied all your proposed changes. (-:

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,51 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
...
 	}
 
+	// --- assertLinesMatch ----------------------------------------------------
+
+	/**
+	 * <em>Asserts</em> that {@code expected} list of {@linkplain String}s matches {@code actual}
+	 * list.
+	 *
+	 * <p>This method differs from other assertions that effectively only check
+	 * {@link String#equals(Object)} by using the following staged match-making algorithm:
```
jbduncan: Sorry @sormuras, I realise that I was unclear; I thought that only the comma on line 880 needed to be removed, not also the one on line 881.

Hope about we change the middle of the sentence so it's clearer - something like:
> This method differs from other assertions that effectively only check {@code ...}, in that it uses the following staged match-making algorithm:

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
jbduncan: YW @sormuras. :-)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
```diff
@@ -0,0 +1,178 @@
...
+					}
+					if (matches(expectedLine, actualDeque.peek())) {
+						continue main;
+					}
+					actualDeque.pop();
+				}
+			}
+
+			int number = expectedLines.size() - expectedDeque.size(); // 1-based line number
+			fail(expectedLines, actualLines, "expected line #%d:`%s` don't match", number, snippet(expectedLine));
```
sbrannen: don't --> doesn't

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
```diff
@@ -0,0 +1,178 @@
...
+	static boolean isFastForwardLine(String line) {
+		line = line.trim();
+		return line.length() >= 4 && line.startsWith(">>") && line.endsWith(">>");
+	}
+
+	static int parseFastForwardLimit(String fastForwardLine) {
+		String text = fastForwardLine.trim().substring(2, fastForwardLine.length() - 2).trim();
+		try {
+			int limit = Integer.parseInt(text);
+			condition(limit > 0, "fast-forward limit must be greater than zero, it is: " + limit);
```
sbrannen: put actual limit _inline_: "fast-forward limit [%s] must be greater than zero".

sbrannen: Also, when building a string you are required to use the `Supplier<String>` variant of `condition()`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -870,6 +871,51 @@ public static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual
...
 	}
 
+	// --- assertLinesMatch ----------------------------------------------------
+
+	/**
+	 * <em>Asserts</em> that {@code expected} list of {@linkplain String}s matches {@code actual}
+	 * list.
+	 *
+	 * <p>This method differs from other assertions that effectively only check {@link String#equals(Object)},
+	 * in that it uses the following staged match-making algorithm:
```
sbrannen: "match making" is for a dating web site.

You mean simply "matching". 😛

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
sormuras: Check.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
sormuras: Check. Changed to same pattern used above. It reads now:

`condition(limit > 0, () -> format("fast-forward(%d) limit must be greater than zero", limit));`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
sormuras: Check. Borrowed the word from [here](https://us.battle.net/forums/en/overwatch/topic/20745144780#post-11).

documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
```diff
@@ -104,7 +104,9 @@ on GitHub.
 * `TestInstancePostProcessors` registered on test methods are now invoked.
 * There are two new signatures for `Assertions.fail`: `Assertions.fail(Throwable cause)` and
   `Assertions.fail(String message, Throwable cause)`.
-
+* New `Assertions.assertLinesMatch()` comparing lists of `String` featuring `Object::equals` and
+  regular expression checks. It also provides a fast-forward mechanism to skip actual lines that
+  are expected to change in each invocation, e.g. duration, timestamps and stack traces.
```
jbduncan: @sormuras I know at a basic level what you're trying to say here, as I've looked at the Javadoc and tests for this new assertion, but I don't think newcomers would understand from this rather sparse description what the assertion does or what its purpose is.

How about something like this?

> Introduced a new `Assertions.assertLinesMatch()` assertion, which is similar to `Assertions.assertIterableEquals()`, except that it executes on `List<String>`s rather than `Iterable<?>`s, and the expected lines can be a combination of literal lines to match against; regular expressions; and representations of a new concept called "fast-forward", which allows some of the actual lines to be exempted from checking. See the Javadoc for more details.

sormuras: Mh, that seems a bit too detailed for a short note. I'll add the _See the Javadoc for more details._ with a link to it, though.

jbduncan: Sounds good to me!

PR #739, opened 2017-03-18
Add Launcher.execute(Request, Listener…) overload

### LINKED ISSUE #628: Register TestExecutionListener specific to a LauncherDiscoveryRequest
I am using a single instance of `Launcher` to execute multiple requests. For each request, I want to register a separate instance of `TestExecutionListener` with the launcher so that a listener is notified of the events only for that particular request/test plan. The listener should also provide access to the `TestPlan` to which it is listening.  

I am using the existing APIs in Eclipse in the following way:
- Registering a single static listener with the launcher for all the requests being executed. This is not correct as we don't want to use a single static listener.
- Accessing the test plan in the listener via `TestExecutionListener.testPlanExecutionStarted(TestPlan testPlan)` API. This could be a problem if test plans are being run in parallel.

You can have a look at these files for the current implementation: 
- [JUnit5TestReference.java](https://github.com/noopur2507/eclipse-junit5/blob/master/org.eclipse.jdt.junit5.runtime/src/org/eclipse/jdt/internal/junit5/runner/JUnit5TestReference.java)
- [JUnit5TestListener.java](https://github.com/noopur2507/eclipse-junit5/blob/master/org.eclipse.jdt.junit5.runtime/src/org/eclipse/jdt/internal/junit5/runner/JUnit5TestListener.java)

Please let me know if I am missing something or not using these APIs in the intended way.

sormuras:
## Overview

Addresses #628 by adding an overloaded execute(LauncherDiscoveryRequest, TestExecutionListener) method that uses the supplied TestExecutionListeners in addition to already registered listeners but only for the supplied LauncherDiscoveryRequest.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
+++ b/documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
@@ -48,6 +48,11 @@ on GitHub.
   by an executable standalone JAR distribution. For details, see the "New Features" section below.
 * The `MethodSortOrder` enum in `ReflectionUtils` has been renamed to `HierarchyTraversalMode`. Those
   affected should now use `ReflectionSupport` instead of `ReflectionUtils`.
+* The method `execute(LauncherDiscoveryRequest launcherDiscoveryRequest)` in `Launcher` has been
+  deprecated and will be removed in milestone M5. Instead use the following new method that registers
+  supplied ``TestExecutionListener``s in addition to already registered listeners but only for the
+  supplied `LauncherDiscoveryRequest`:
+  `execute(LauncherDiscoveryRequest launcherDiscoveryRequest, TestExecutionListener... listeners)`
 
 ===== New Features
 
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
@@ -81,8 +81,29 @@ public interface Launcher {
 	 * collecting their results, and notify {@linkplain #registerTestExecutionListeners
 	 * registered listeners} about the progress and results of the execution.
 	 *
-	 * @param launcherDiscoveryRequest the launcher discovery request; never {@code null}
+	 * <p>The <em>deprecated</em> default implementation delegates to
+	 * {@link #execute(LauncherDiscoveryRequest, TestExecutionListener...)} with no
+	 * additional listeners.
+	 *
+	 * <p>The method will be removed in milestone M5.
+	 * @deprecated Use {@link #execute(LauncherDiscoveryRequest, TestExecutionListener...)} instead.
 	 */
-	void execute(LauncherDiscoveryRequest launcherDiscoveryRequest);
+	@Deprecated
+	default void execute(LauncherDiscoveryRequest launcherDiscoveryRequest) {
+		execute(launcherDiscoveryRequest, new TestExecutionListener[0]);
+	}
 
+	/**
+	 * Execute a {@link TestPlan} which is built according to the supplied
+	 * {@link LauncherDiscoveryRequest} by querying all registered engines and
+	 * collecting their results, and notify {@linkplain #registerTestExecutionListeners
+	 * registered listeners} about the progress and results of the execution.
+	 *
+	 * <p>Supplied test execution listeners are registered in addition to already
+	 * registered listeners but only for the supplied launcher discovery request.
+	 *
+	 * @param launcherDiscoveryRequest the launcher discovery request; never {@code null}
+	 * @param listeners additional test execution listeners; never {@code null}
+	 */
+	void execute(LauncherDiscoveryRequest launcherDiscoveryRequest, TestExecutionListener... listeners);
 }
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
@@ -80,9 +80,11 @@ class DefaultLauncher implements Launcher {
 	}
 
 	@Override
-	public void execute(LauncherDiscoveryRequest discoveryRequest) {
+	public void execute(LauncherDiscoveryRequest discoveryRequest, TestExecutionListener... listeners) {
 		Preconditions.notNull(discoveryRequest, "LauncherDiscoveryRequest must not be null");
-		execute(discoverRoot(discoveryRequest, "execution"), discoveryRequest.getConfigurationParameters());
+		Preconditions.notNull(listeners, "TestExecutionListener array must not be null");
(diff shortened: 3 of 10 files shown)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
```diff
@@ -81,8 +81,24 @@
...
 	 * registered listeners} about the progress and results of the execution.
 	 *
+	 * <p>The default implementation delegates to
+	 * {@link #execute(LauncherDiscoveryRequest, TestExecutionListener...)} with no
+	 * additional listeners.
+	 *
 	 * @param launcherDiscoveryRequest the launcher discovery request; never {@code null}
 	 */
-	void execute(LauncherDiscoveryRequest launcherDiscoveryRequest);
+	default void execute(LauncherDiscoveryRequest launcherDiscoveryRequest) {
```
sormuras: If this (default) method was removed (as the new variable parameter method overload suggests) external tools will break. Leave it as-is, mark as @Deprecated, remove it anyway?

sbrannen: Deprecate it in M4 and remove it in M5 (and create an issue assigned to M5 so that the removal is not forgotten).

sbrannen: And of course document the deprecation and pending removal clearly in the M4 release notes.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -80,9 +80,12 @@ public TestPlan discover(LauncherDiscoveryRequest discoveryRequest) {
 	}
 
 	@Override
-	public void execute(LauncherDiscoveryRequest discoveryRequest) {
+	public void execute(LauncherDiscoveryRequest discoveryRequest, TestExecutionListener... listeners) {
 		Preconditions.notNull(discoveryRequest, "LauncherDiscoveryRequest must not be null");
+		Preconditions.notNull(listeners, "TestExecutionListener array must not be null");
+		this.listenerRegistry.registerListeners(listeners);
```
sbrannen: Although we don't state that the `execute()` method can be launched in parallel from different threads for the same `Launcher` instance, it still makes me a bit uneasy to _mutate_ the listener registry in this manner.

I think I would prefer to introduce a composite wrapper in such cases and leave the original listener registry untouched.

sbrannen:
As a side note, please use the word "addresses" (or similar) instead of "fixes" when a PR is not a bug fix.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
sormuras: Adding `@Deprected` to the method breaks the build, because of `error: warnings found and -Werror specified`. Will temporarely modify all no-arg calls to pass an extra `new TEListener[0]` argument to make the compiler happy. Or do you know a better way around this?

documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
```diff
@@ -48,6 +48,11 @@ on GitHub.
   by an executable standalone JAR distribution. For details, see the "New Features" section below.
 * The `MethodSortOrder` enum in `ReflectionUtils` has been renamed to `HierarchyTraversalMode`. Those
   affected should now use `ReflectionSupport` instead of `ReflectionUtils`.
+* The method `execute(LauncherDiscoveryRequest launcherDiscoveryRequest)` in `Launcher` was
```
sbrannen: "was" --> "has been"

documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
```diff
@@ -48,6 +48,11 @@ on GitHub.
...
+  deprecated and will be removed in milestone M5. Instead use new method that registers supplied
+  `TestExecutionListener`s in addition to already registered listeners but only for the
```
sbrannen: This formatting doesn't work AsciiDoc. See my recent commit addressing that issue. 😉 

Basically, you have to wrap the _code_ in double back-ticks if you place any character directly adjacent to the formatted code: \`\`TestExecutionListener\`\`s

documentation/src/docs/asciidoc/release-notes-5.0.0-M4.adoc
```diff
@@ -48,6 +48,11 @@ on GitHub.
   by an executable standalone JAR distribution. For details, see the "New Features" section below.
 * The `MethodSortOrder` enum in `ReflectionUtils` has been renamed to `HierarchyTraversalMode`. Those
   affected should now use `ReflectionSupport` instead of `ReflectionUtils`.
+* The method `execute(LauncherDiscoveryRequest launcherDiscoveryRequest)` in `Launcher` was
+  deprecated and will be removed in milestone M5. Instead use new method that registers supplied
```
sbrannen: "use new" --> "use the following new"

junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
```diff
@@ -81,8 +81,28 @@
...
 	 * registered listeners} about the progress and results of the execution.
 	 *
-	 * @param launcherDiscoveryRequest the launcher discovery request; never {@code null}
+	 * <p>The <em>deprecated</em> default implementation delegates to
+	 * {@link #execute(LauncherDiscoveryRequest, TestExecutionListener...)} with no
+	 * additional listeners.
+	 *
+	 * <p>The method will be removed in milestone M5.
 	 */
-	void execute(LauncherDiscoveryRequest launcherDiscoveryRequest);
```
sbrannen: When deprecating something, we not only annotate it with `@Deprecated`, but we also used the `@deprecated` JavaDoc tag so that the information shows up in the published API documentation. See other usages in the framework for examples.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -80,9 +80,10 @@ public TestPlan discover(LauncherDiscoveryRequest discoveryRequest) {
 	}
 
 	@Override
-	public void execute(LauncherDiscoveryRequest discoveryRequest) {
+	public void execute(LauncherDiscoveryRequest discoveryRequest, TestExecutionListener... listeners) {
 		Preconditions.notNull(discoveryRequest, "LauncherDiscoveryRequest must not be null");
-		execute(discoverRoot(discoveryRequest, "execution"), discoveryRequest.getConfigurationParameters());
+		Preconditions.notNull(listeners, "TestExecutionListener array must not be null");
```
sbrannen: You'll want to use `Preconditions.containsNoNullElements(T[], String)` here as well.

Issue #672, opened 2017-02-22
Introduce extension API for creating the test instance


smoyer64:
## Overview

The javadocs for the JUnit 4 [```TestRule```](http://junit.org/junit4/javadoc/latest/org/junit/rules/TestRule.html) states:

>   The Statement that executes the method or suite is passed to each annotated Rule in turn,
>   and each may return a substitute or modified Statement, which is passed to the next Rule, if any.

I currently have a JUnit 4 rule that uses this ability to replace the framework's provided instance with one that has been initialized via JavaEE CDI (Weld) but this isn't currently possible with JUnit 5.  The ```MockitoExtension``` post-processes the instance provided by the engine - doing that in a ```WeldExtension``` wouldn't be terribly hard if we're simply talking about processing the ```@Inject``` annotations but fully initializing the instance as a proper CDI managed object would be non-trivial and would require duplicating much of the functionality already provided by the CDI container.

## Feature request

Provide an ```Callback``` that allows an ```Extension``` to produce or replace the test instance.  In the case of CDI, you'd replace:

```
    Object testInstance = new ClassUnderTest();
```

with something like:

```
    Object testInstance = beanManager.select(ClassUnderTest.class).get();
```

## Deliverables

- [x] Provide the ability to replace or alter the engine's provided test instance.

smoyer64:
The weld team has responded:

> (08:45:02 AM) mkouba: smoyer1: hi
> (08:45:14 AM) mkouba: what do you mean by "initialize"?
> (08:45:20 AM) mkouba: to perform injection?
> (08:49:20 AM) smarlow_ [~smarlow@redhat/jboss/smarlow] entered the room.
> (08:49:43 AM) smoyer1: well ... injection is part of it right?
> (08:50:12 AM) smoyer1: if it's @Observing events those need to be registered too ... same with Interceptors, etc
> (08:50:47 AM) smoyer1: basically ... instead of doing beanManager.select(ClassUnderTest.class).get
> (08:51:44 AM) smoyer1: I need that beanManager to perform the same functions on a test instance that was originally created with new ClassUnderTest()
> (08:54:57 AM) mkouba: that's not possible, you can only perform injection on a non-contextual instance
> (08:55:09 AM) mkouba: observers are initialized during bootstrap
> (08:55:39 AM) mkouba: interceptors and decorators are only applied to instances created by the container
[...]

Issue #203, opened 2016-03-14
InstancePostProcessor should return test instance


jlink:
```
public interface InstancePostProcessor extends ExtensionPoint {
        /** Return the original or a wrapped instance of the test object */
    Object postProcessTestInstance(TestExtensionContext context) throws Exception;
}
```
This seems to be necessary for PowerMock extension (see #201). 

@junit-team Any objections?

sbrannen:
I don't think it is a wise idea to repurpose `InstancePostProcessor` for extensions that wish to replace the test instance. 

For example, let's assume we have two registered extensions that implement `InstancePostProcessor`. For the sake of argument, let's call them `S` and `P` (for Spring and PowerMock).

If `S` injects dependencies into the test instance and `P` then completely replaces the test instance, any tests relying on dependency injection from Spring will fail even though Spring's log message will state that dependency injection in fact occurred.

At the very least, there is an issue with regard to ordering of such extensions, but I would argue that is only an indicator of the true problem:

_When does it ever make sense for more than one extension to replace the test instance?_

In my opinion, the answer to the above question is __never__.

With that in mind, I would propose a new extension -- call it `InstanceFactory` or similar -- that can only be registered once per test container. Any attempt for a second extension to register itself as such a factory would then result in an exception.

sbrannen:
Of course, the potential problem I outlined above does not come into effect if instances of `InstancePostProcessor

thekingnothing:
For PowerMock main goal- it's to be able to load test class and all other classes with PowerMock class loader. 

Approach with `InstanceFactory` very attractive for us, it's how TestNG works now.  And of course, only one `InstanceFactory` could be declared. 

But, it also could be a point of fail, for example assertion exception or any other exception witch test could throw and jUnit engine is expected. In this class casting issues moved forward. I mean that trying to cast exception to expected class could throw `ClassCastException' or instanceOf would return incorrect result.

jlink:
@sbrannen It's impossible to enforce wrapping, or is it. It might not even be practical for PowerMock's purposes when using a different class loader to create basically the same object.

mepeisen:
In my Showcase I got another Problem. Playing around with class loader results in other extensions not working as expected.
[...]

Issue #201, opened 2016-03-13
Introduce extension API for providing a custom ClassLoader (e.g., for Powermock)


mepeisen:
## Overview

I tried to build a Powermock `Extension` similar to the Mockito example. I have to test some classes that create new objects within their constructors. In JUnit 4 I simply use the `PowerkmockRunner` and the `whenNew` from Powermock.

At least I ended with many problems and gave up. However in JUnit 4 Powermock creates an instrumented class loader and even duplicates everything (the test instance etc.) by using the new class loader. It is simple because the loader/rule are aware of executing the test itself. In JUnit Jupiter an `Extension` is not able to execute the test by itself.

I am wondering how this will be done in JUnit Jupiter. Are you planning to introduce some `CreateTestInstance` extension API? Something that is aware of manipulating class loaders. Or at least an `AroundAll` extension API to instrument the whole test class?

## Related Issues

- #157
- #805
- #806 
- #1680
- #1799

jlink:
What we have is the `InstancePostProcessor` extension point. Currently, it does not allow to return a different test instance but we have been discussing to allow that, e.g.:

```
public interface InstancePostProcessor extends ExtensionPoint {
        /** Return the original or a wrapped instance of the test object */
	Object postProcessTestInstance(TestExtensionContext context) throws Exception;
}

```
Would that be sufficien for your purposes? I am curious though, why PowerMock has to change the test instance (which can easily lead to incompatibilities with other extensions). Maybe there's another way to go about it?

As for an extension point that wraps test execution itself, we are currently shying away from it for reasons discussed (among other places) in #157.

mepeisen:
As far as I can see returning a different instance should fit the needs.
I do not know why powermock needs it but I will ask the authors. All I can see is that mocking objects and classes already works.

Actually in my Scenario i need to mock things happening in constructors.

`class ToTest {
public ToTest() {
    this.aObj = new A();
    doSomeThingWithAObj();
}
}`

And I need to inject a mock for "aObj". Actually aObj is a console reader to receive commands.

Mabye exactly this scenario need playing around with class loaders and replacing the test instance. I will invite the powermock guys to discuss this in this issue :-)

jlink:
2016-03-14 8:43 GMT+01:00 Martin Eisengardt <notifications@github.com>:

> Mabye exactly this scenario need playing around with class loaders and
> replacing the test instance. I will invite the powermock guys to discuss
> this in this issue :-)


[...]

PR #748, opened 2017-03-24
Introduce extension point before test instantiation

codecholeric:
To migrate our existing framework to JUnit 5 I'd like to hook into the lifecycle of JUnit 5 test execution before any test instance is created.

We want to configure the test environment for each test right before the test instance is created.

 * TestInstancePostProcessor is too late, as instance fields have already been initialized.
 * BeforeAll is not appropriate, as we need to configure the environment for each test run.

 ---

## Overview

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/BeforeTestInstantiationCallback.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/BeforeTestInstantiationCallback.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.jupiter.api.extension;
+
+import static org.junit.platform.commons.meta.API.Usage.Experimental;
+
+import org.junit.platform.commons.meta.API;
+
+/**
+ * {@code BeforeTestInstantiationCallback} defines the API for {@link Extension
+ * Extensions} that wish to provide additional behavior to tests immediately
+ * before each test instance is created.
+ *
+ * <p>Implementations must provide a no-args constructor.
+ *
+ * @since 5.0
+ * @see org.junit.jupiter.api.Test
+ * @see BeforeTestExecutionCallback
+ * @see AfterTestExecutionCallback
+ * @see BeforeEachCallback
+ * @see AfterEachCallback
+ * @see BeforeAllCallback
+ * @see AfterAllCallback
+ */
+@FunctionalInterface
+@API(Experimental)
+public interface BeforeTestInstantiationCallback extends Extension {
+
+	/**
+	 * Callback that is invoked <em>immediately before</em> each test instance is created.
+	 *
+	 * @param context the current extension context; never {@code null}
+	 */
+	void beforeTestInstantiation(ContainerExtensionContext context) throws Exception;
+
+}
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
@@ -26,6 +26,7 @@ import java.util.function.Function;
 
 import org.junit.jupiter.api.extension.AfterAllCallback;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
+import org.junit.jupiter.api.extension.BeforeTestInstantiationCallback;
 import org.junit.jupiter.api.extension.ContainerExtensionContext;
 import org.junit.jupiter.api.extension.Extension;
 import org.junit.jupiter.api.extension.ExtensionContext;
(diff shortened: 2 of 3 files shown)


sbrannen:
> TestInstancePostProcessor is too late, as instance fields have already been initialized.

That's not necessarily true. The primary use case for a `TestInstancePostProcessor` is in fact to set fields (i.e., dependency injection).

> TestInstancePostProcessor is too late

Can you please expound on what you mean by that statement?

It's too late to achieve _what_?

codecholeric:
## Short story:

For our test infrastructure to create test data, we must distinguish calls during the static initialization of each test class from calls during test instantiation for each test method.
This could be solved by the proposed `BeforeTestInstantiationCallback`.

## Long story:

When we write our tests, we want to restrict the data that have to be set explicitly to those that the specific test requires. All other properties are set randomly, but valid. Naturally using random data may introduce randomly failing tests, so we need the means to easily reproduce test failures that e.g. happen in the CI environment.
Reproducing such failures requires knowing the seed values for our random data when the specific test was executed. This is where we want to take advantage of a JUnit 5 extension. It has to store one seed value at the class level, for example if static constants are initialized using our random mechanism. In addition, it has to store the seed value at the instance level, for each test run. If a test should fail, the respective seed values must be reported in the failure message. If the failure is related to the use of random data, all this together allows us to reproduce it in any other environment. Example:

```
private static final Address staticAddress = 
    createRandomAddress(); // determined by the class seed
 
private Address instanceAddress = 
    createRandomAddressBuilder() // determined by the instance seed
        .withStreet("some street")
        .build();
 
@Test
void someTest() {
    Address anotherAddress = 
        createRandomAddress(); // also determined by the instance seed
    // ...
}
```

We note the class seed in a `BeforeAllCallback`, which is the perfect place. The switch to using the instance seed however, has to be right before the test instance is created, to cover random data used for instance field initialization, which is a rather common use case in our tests.
Obviously, the `TestInstancePostProcessor` callback is too late for switching from the class seed to the instance seed. Currently we have an ugly workaround, where we check the stacktrace to switch to the instance seed, as soon as we detect a call from the test's constructor. If there are no such calls we switch to the instance seed in the `TestInstancePostProcessor` callback.
The proposed `BeforeTestInstantiationCallback` would be the natural place to make the required switch to the instance seed trivial.
Do you think, adding the `BeforeTestInstantiationCallback` would be harmful?

marcphilipp:
IIRC there has also been some discussion about providing an extension point for test class instantiation, i.e. it would return the test class instance, but I cannot remember the issue at the moment. That would be more powerful and I think we'd need only one.

@smoyer64 @junit-team/junit-lambda Do you remember the issue?

marcphilipp:
I believe the issue I was looking for is #672.

jonashoef:
I fully agree. An extension point for test instance creation would certainly solve our problem and provide more flexibility than the one we proposed.

smoyer64:
@marcphilipp - Yes, I think #672 (a Weld use-case) is the closest match.  In that issue's discussion, you reference #203 (now closed) and following the chain of issues, @jlink references #201 (a PowerMock use-case).

Looking back at that issue, the @sbrannen quote:

>   With that in mind, I would propose a new extension -- call it InstanceFactory or similar -- that can only be registered once per test container. Any attempt for a second extension to register itself as such a factory would then result in an exception.

should probably be clarified.  Does "test container" refer to the top-level test container or does it refer to (at a minimum) the test's immediate parent container?  I can certainly imagine scenarios where you might use different a InstanceFactory for different tests.  It definitely makes sense that for each test, only one InstanceFactory can provide the actual instance.  While my original question to the Weld team (performing CDI bootstrap and initialization for a provided instance) couldn't be done, I think the idea of an InstanceDecorator might also provide value ... since I have no use-cases to provide for this concept, I won't propose it (at least not yet).

In general, the JUnit 4 rules that I've written have predominately been doing some form of decorating/initializing but many of those cases can be handled by TestInstancePostProcessor extensions (and perhaps all).

Thanks for keeping this "on the radar", I'd love to be able to use Weld in JUnit 5 tests!

jonashoef:
I fully understand that the proposed InstanceFactory extension may exists only once per test (container). 
So, to that respect the additional flexibility does impose restrictions that our original proposal doesn't: Several extensions implementing the BeforeTestInstantiationCallback may be freely combined, but only a single extension per test (container) may implement the InstanceFactory. I'm afraid this restriction is inherent, or do you have an idea how it could be avoided?

nicolaiparlog:
I typed a longer answer with different follow-up questions and then realized that this is a PR. I don't know how the JUnit team stands on this but I would prefer discussing the general feasibility of such an extension point on an issue and use the PR to discuss the concrete code change.

@codecholeric Could you open a new issue with the feature request and the short and long story as background?

codecholeric:
Done -> #839

marcphilipp:
I think this use case should be solvable using  `TestInstanceFactory` introduced in 5.3. Feel free to comment if you feel that this one should not have been closed.

codecholeric:
@marcphilipp I finally had a possibility to check it out. `TestInstanceFactory` does solve our problem in a certain context, but it still is not as clean as it should be.
As @jonashoef mentioned, we really only care about a hook before the test instance is created. Yes, we can implement `TestInstanceFactory`, execute the hook and return `testClass.newInstance()`. But due to the nature of `TestInstanceFactory`, this can now not be combined anymore with any other extension that wants to create the test instance, even though for our use case we would not need to control how the test instance is created at all.
Nevertheless I see your point if our project is the only project in the world that could use such a hook :wink:

marcphilipp:
@codecholeric Thanks for checking! Feel free to open a new issue should this lack of composability become a problem.

PR #751, opened 2017-03-27
Gradle plugin improvements

mkobit:
## Overview

Fixes https://github.com/junit-team/junit5/issues/708 and adds functional tests for validating Gradle plugin until the Gradle team can start implementing built-in support.

- No changelog updated yet, wasn't sure if this would get in.
- Testing Uses locally built components by providing the classpath of those components as dependencies
- Does not currently use the [`java-gradle-plugin`](https://docs.gradle.org/current/userguide/javaGradle_plugin.html) - I wasn't sure how that would fare with the current signing and other components used right now, so I didn't add that in (can be done later)

*Questions*

- Should we run the functional tests on every build?
- Should we run the functional tests with multiple Gradle versions (very easy to add)?
- Is the formatting for writing build files, java files, etc. in the functional test acceptable? 

---

---

### Definition of Done

### DIFF
--- a/junit-platform-gradle-plugin/src/test/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPluginFunctionalSpec.groovy
+++ b/junit-platform-gradle-plugin/src/test/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPluginFunctionalSpec.groovy
@@ -1,0 +1,250 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package org.junit.platform.gradle.plugin
+
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import spock.lang.Specification
+
+class JUnitPlatformPluginFunctionalSpec extends Specification {
+
+	@Rule
+	final TemporaryFolder testProjectDir = new TemporaryFolder()
+	private File buildFile
+	private List<File> pluginClasspath
+	private List<File> testCompileClasspath
+	private List<File> testRuntimeClasspath
+
+	def setup() {
+		pluginClasspath = loadClassPathManifestResource('plugin-classpath.txt')
+		testCompileClasspath = loadClassPathManifestResource('test-compile-classpath.txt')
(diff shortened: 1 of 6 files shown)


junit-platform-gradle-plugin/src/test/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPluginFunctionalSpec.groovy
```diff
@@ -0,0 +1,272 @@
...
+			.withPluginClasspath(pluginClasspath)
+			.withArguments('test')
+			.build()
+
+		then:
+		result.task(':junitPlatformTest').outcome == TaskOutcome.SUCCESS
+
+		when:
+		result = GradleRunner.create()
+												.withProjectDir(testProjectDir.root)
```
sbrannen: What happened to the indentation here?

junit-platform-gradle-plugin/src/test/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPluginFunctionalSpec.groovy
```diff
@@ -0,0 +1,272 @@
...
+}
+"""
+	}
+
+	private void javaFile() {
+		Path javaFile = Paths.get(testProjectDir.root.toString(), 'src', 'main', 'java', 'org', 'junit', 'gradletest', 'Adder.java')
+		Files.createDirectories(javaFile.parent)
+		javaFile.withWriter {
+			it.write(
+				'''
```
sbrannen: I think it's more consistent if the `'''` are on the previous line immediately following `write(`.

mkobit: Good idea, will do that.

junit-platform-gradle-plugin/src/test/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPluginFunctionalSpec.groovy
mkobit: Good catch, I missed this one.

I typically use spaces and the "chop down and align" settings for continuations.
That is the default for me in IntelliJ.

In this case with tabs the many tab characters lead to something... that looks like this.
I removed a not of the indentation on the others, but missed this one. I'll reduce it to be in the same style as the others.

sormuras:
Thanks @mkobit !

> Should we run the functional tests on every build?
> Should we run the functional tests with multiple Gradle versions (very easy to add)?

Only build duration time will increase here, am I right?

On the other hand, I'm fine with the current situation -- as the Gradle team will soon take over the plugin developement.

> Is the formatting for writing build files, java files, etc. in the functional test acceptable?

Sure.

mkobit:
Yeah, build duration should be the only increase, but I'm with you that this will hopefully be enough until the Gradle team takes over the plugin development.

PR #773, opened 2017-04-06
Add a plugin to detect out of date plugins

grimreaper:


build.gradle
```diff
@@ -17,6 +17,7 @@ buildscript {
 plugins {
 	id 'com.gradle.build-scan' version '1.6'
 	id 'net.nemerosa.versioning' version '2.6.0'
+	id 'com.github.ben-manes.versions' version '0.14.0'
```
marcphilipp: Already present in line 13?

PR #776, opened 2017-04-06
Do not execute abstract tests in VintageTestEngine

LiamClark:
## Overview

The Junit4DiscoveryRequestResolver did not take the isPotentialJunit4TestClass
predicate in consideration. This is now added in the TestClassCollector, so
Abstract classes are properly filtered.

Multiple tests however used non-JUnit 4 compliant classes as test input.
These are now moved to make them proper JUnit 4 test classes.

Perhaps some of these test classes should be given better names, since they are now on the top level.

The logging follows the output maven generated for the same class when it was being
run from the jupiter engine.

---

---

### Definition of Done

### DIFF
--- a/junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/JUnit4DiscoveryRequestResolver.java
+++ b/junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/JUnit4DiscoveryRequestResolver.java
@@ -35,6 +35,7 @@ import org.junit.platform.engine.support.filter.ExclusionReasonConsumingFilter;
 @API(Internal)
 public class JUnit4DiscoveryRequestResolver {
 
+	private static final IsPotentialJUnit4TestClass isPotentialJUnit4TestClass = new IsPotentialJUnit4TestClass();
 	private final EngineDescriptor engineDescriptor;
 	private final Logger logger;
 
@@ -81,7 +82,19 @@ public class JUnit4DiscoveryRequestResolver {
 		Filter<Class<?>> classFilter = new ExclusionReasonConsumingFilter<>(adaptedFilter,
 			(testClass, reason) -> logger.fine(() -> String.format("Class %s was excluded by a class filter: %s",
 				testClass.getName(), reason.orElse("<unknown reason>"))));
-		return collector.toRequests(classFilter.toPredicate());
+		return collector.toRequests(classFilter.toPredicate().and(loggingPotentialJUnit4TestClassPredicate()));
+	}
+
+	private Predicate<Class<?>> loggingPotentialJUnit4TestClassPredicate() {
+		return testClass -> {
+			boolean isPotentialTestClass = isPotentialJUnit4TestClass.test(testClass);
+
+			if (!isPotentialTestClass) {
+				logger.warning(() -> String.format("Class %s could not be resolved", testClass.getName()));
+			}
+
+			return isPotentialTestClass;
+		};
 	}
 
 	private void populateEngineDescriptor(Set<TestClassRequest> requests) {
--- a/junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineExecutionTests.java
+++ b/junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineExecutionTests.java
@@ -410,13 +410,14 @@ class VintageTestEngineExecutionTests {
 
 	}
 
+	@RunWith(MisbehavingSuiteRunner.class)
+	public static class MisBehavingSuiteTestClass {
+
+	}
+
 	@Test
 	void ignoreEventsForUnknownDescriptionsByMisbehavingSuiteRunner() {
-		@RunWith(MisbehavingSuiteRunner.class)
-		class TestClass {
-
-		}
-		Class<?> testClass = TestClass.class;
+		Class<?> testClass = MisBehavingSuiteTestClass.class;
 
 		List<ExecutionEvent> executionEvents = execute(testClass);
 
@@ -447,13 +448,14 @@ class VintageTestEngineExecutionTests {
 
 	}
 
+	@RunWith(MisbehavingChildlessRunner.class)
+	public static class MisbehavingChildTestClass {
+
+	}
+
 	@Test
 	void ignoreEventsForUnknownDescriptionsByMisbehavingChildlessRunner() {
-		@RunWith(MisbehavingChildlessRunner.class)
-		class TestClass {
-
-		}
-		Class<?> testClass = TestClass.class;
+		Class<?> testClass = MisbehavingChildTestClass.class;
 
 		List<ExecutionEvent> executionEvents = execute(testClass);
 
--- a/junit-vintage-engine/src/test/java/org/junit/vintage/engine/discovery/VintageDiscoveryRequestResolverTests.java
+++ b/junit-vintage-engine/src/test/java/org/junit/vintage/engine/discovery/VintageDiscoveryRequestResolverTests.java
@@ -25,7 +25,9 @@ import org.junit.jupiter.api.Test;
 import org.junit.platform.engine.EngineDiscoveryRequest;
 import org.junit.platform.engine.discovery.ClassNameFilter;
 import org.junit.platform.engine.support.descriptor.EngineDescriptor;
+import org.junit.platform.launcher.LauncherDiscoveryRequest;
 import org.junit.vintage.engine.RecordCollectingLogger;
+import org.junit.vintage.engine.samples.junit4.AbstractJunit4TestCaseWithConstructorParameter;
 
 /**
  * @since 4.12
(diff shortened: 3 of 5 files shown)


junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/JUnit4DiscoveryRequestResolver.java
```diff
@@ -81,7 +81,23 @@ private TestClassCollector collectTestClasses(EngineDiscoveryRequest discoveryRe
 		Filter<Class<?>> classFilter = new ExclusionReasonConsumingFilter<>(adaptedFilter,
 			(testClass, reason) -> logger.fine(() -> String.format("Class %s was excluded by a class filter: %s",
 				testClass.getName(), reason.orElse("<unknown reason>"))));
-		return collector.toRequests(classFilter.toPredicate());
+
+		Predicate<Class<?>> loggingTestClass = isPotentialJunit4Predicate();
```
sbrannen: It's not a test class that logs; rather, it's a predicate that logs the test class.

So please rename to `loggingPotentialJUnit4TestClassPredicate` or similar.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/JUnit4DiscoveryRequestResolver.java
```diff
@@ -81,7 +81,23 @@ private TestClassCollector collectTestClasses(EngineDiscoveryRequest discoveryRe
...
+
+		return collector.toRequests(classFilter.toPredicate().and(loggingTestClass));
+	}
+
+	private Predicate<Class<?>> isPotentialJunit4Predicate() {
```
sbrannen: ...is a potential JUnit 4.... what?

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/JUnit4DiscoveryRequestResolver.java
```diff
@@ -81,7 +81,23 @@ private TestClassCollector collectTestClasses(EngineDiscoveryRequest discoveryRe
...
+		IsPotentialJUnit4TestClass potentialTestClass = new IsPotentialJUnit4TestClass();
```
sbrannen: No need to instantiate the predicate every time.

Please instantiate it once as a `static final` field as is done elsewhere in the code base for such predicates.

Also, please name the field consistent with the rest of the code base: `isPotentialJUnit4TestClass`

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/JUnit4DiscoveryRequestResolver.java
```diff
@@ -81,7 +81,23 @@ private TestClassCollector collectTestClasses(EngineDiscoveryRequest discoveryRe
...
+		return testClass -> {
+			boolean isTestClass = potentialTestClass.test(testClass);
```
sbrannen: `isTestClass` --> `isPotentialTestClass`

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/TestClassCollector.java
```diff
@@ -48,7 +48,11 @@ void addFiltered(Class<?> testClass, RunnerTestDescriptorAwareFilter filter) {
 	}
 
 	private Stream<TestClassRequest> completeRequests(Predicate<? super Class<?>> predicate) {
-		return completeTestClasses.stream().filter(predicate).map(TestClassRequest::new);
+		// @formatter:off
```
sbrannen: No need to reformat this.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineExecutionTests.java
```diff
@@ -66,12 +66,13 @@
...
 import org.junit.vintage.engine.samples.junit4.PlainJUnit4TestCaseWithSingleTestWhichIsIgnored;
 import org.junit.vintage.engine.samples.junit4.PlainJUnit4TestCaseWithTwoTestMethods;
+import org.junit.vintage.engine.samples.junit4.SuiteTestClass;
 import org.opentest4j.MultipleFailuresError;
 
 /**
  * @since 4.12
  */
-class VintageTestEngineExecutionTests {
+public class VintageTestEngineExecutionTests {
```
sbrannen: Why did you make this public?

junit-vintage-engine/src/test/java/org/junit/vintage/engine/discovery/VintageDiscoveryRequestResolverTests.java
```diff
@@ -58,19 +60,21 @@ void logsWarningWhenFilterExcludesClass() {
 			logRecord.getMessage());
 	}
 
-	public static class Foo {
```
sbrannen: Why did you extract `Foo` and `Bar` into top-level classes?

junit-vintage-engine/src/test/java/org/junit/vintage/engine/samples/junit4/ChildlessTestClass.java
```diff
@@ -0,0 +1,19 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.vintage.engine.samples.junit4;
+
+import org.junit.runner.RunWith;
+import org.junit.vintage.engine.VintageTestEngineExecutionTests;
+
+@RunWith(VintageTestEngineExecutionTests.MisbehavingChildlessRunner.class)
+public class ChildlessTestClass {
```
sbrannen: Why did you make this a top-level class instead of simply a nested static class in the test class that actually uses it?

junit-vintage-engine/src/test/java/org/junit/vintage/engine/samples/junit4/SuiteTestClass.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.vintage.engine.samples.junit4;
+
+import org.junit.runner.RunWith;
+import org.junit.vintage.engine.VintageTestEngineExecutionTests;
+
+@RunWith(VintageTestEngineExecutionTests.MisbehavingSuiteRunner.class)
```
sbrannen: Why did you make this a top-level class instead of simply a nested static class in the test class that actually uses it?

sbrannen:
Commit message must follow the guidelines laid out here: https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#commit-messages

In addition, the commit message must contain a reference to the issue it addresses -- for example:

```
Issue: #999
```

junit-vintage-engine/src/test/java/org/junit/vintage/engine/discovery/VintageDiscoveryRequestResolverTests.java
LiamClark: @sbrannen Otherwise they would not be discovered by the vintage engine. 
The isPotentialJunit4TestClass predicate contains a check: !candidate.isMemberClass();
So foo as a nested class would be filtered out making the tests fail.

sbrannen:
Although the title "Fix Junit4DiscoveryRequestResolver" is technically correct, it would be preferable to describe what is being fixed in plain English.

Something like the following would be better: _Do not execute abstract tests in VintageTestEngine_

sbrannen:
Hi @LiamClark,

Thanks for the PR!

I've left some comments and performed a quick review.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/TestClassCollector.java
LiamClark: Ah yes my bad, I added an extra filter here in an initial try, but forgot to remove the formatting

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/JUnit4DiscoveryRequestResolver.java
jbduncan: Consider making the predicate class an enum like the following, since then it cannot be instantiated more than once.
```java
public enum IsPotentialJUnit4TestClass {
  INSTANCE;
  . . .
}
```

LiamClark: @jbduncan @sbrannen doesn't it need to be instantiated for every instance of JUnit4DiscoveryRequestResolver since it accesses the logger field?

jbduncan: Actually, this _may_ be a bad idea if other predicate classes exist that follow the same convention as `IsPotentialJUnit4TestClass` currently does. (I can't quite check ATM since I'm a bit busy...)

junit-vintage-engine/src/test/java/org/junit/vintage/engine/discovery/VintageDiscoveryRequestResolverTests.java
sbrannen: No. That should not be the case. A static nested class is perfectly valid as a test class in the eyes of JUnit 4. And a "member class" is non-static. 

But... if you believe otherwise, please submit a separate PR demonstrating the bug you think is lurking here.

In the interim, please revert your changes to the `Foo` and `Bar` classes.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/JUnit4DiscoveryRequestResolver.java
sbrannen: No, `IsPotentialJUnit4TestClass` does _not_ access any logger whatsoever.

sbrannen: Please do not make the predicate an enum. We don't do that anywhere, and if we were to start implementing our predicates as _singletons_, we would do that in a separate issue.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/discovery/VintageDiscoveryRequestResolverTests.java
marcphilipp: Since #787 has now been merged this PR should be rebased onto master and this particular change can be reverted.

LiamClark:
@sbrannen I've rebased and undid the move for the test classes to the top level.
The tests also used two local classes. These are now public static member classes.

is this ok?

sbrannen:
Yep, looks good now.

Thanks! 👍

PR #779, opened 2017-04-07
revise Asciidoctor task configuration

mojavelinux:
## Overview

Revise the Asciidoctor build task configuration / setup in the following ways:

- remove dedicated task to copy images as this is built into the plugin
- use AsciidoctorJ Diagram instead of low-level asciidoctor-diagram gem dependency
- don't set imagesoutdir as it defaults to destination dir + imagesdir
- set imagesdir in document instead of build configuration to avoid breaking HTML
- no need to activate Gradle plugin defined using plugins clause
- correct a few bad AsciiDoc attribute declarations

---

---

### DIFF
--- a/documentation/documentation.gradle
+++ b/documentation/documentation.gradle
@@ -2,21 +2,15 @@ buildscript {
 	dependencies {
 		// classpath 'org.asciidoctor:asciidoctorj-epub3:1.5.0-alpha.6'
 		classpath 'org.asciidoctor:asciidoctorj-pdf:1.5.0-alpha.14'
+		classpath 'org.asciidoctor:asciidoctorj-diagram:1.5.4'
 	}
 }
 
 plugins {
 	id 'org.asciidoctor.convert' version '1.5.3'
-	// for asciidoctor-diagram support
-	id 'com.github.jruby-gradle.base' version '1.4.0'
 }
 
-apply plugin: 'org.asciidoctor.convert'
 apply plugin: 'org.junit.platform.gradle.plugin'
-
-dependencies {
-	gems 'rubygems:asciidoctor-diagram:1.5.4'
-}
 
 junitPlatform {
 	details 'tree'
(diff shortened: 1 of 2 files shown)


mojavelinux:
cc: @sbrannen

mojavelinux:
Here's why you don't need a dedicated task to copy images:

https://github.com/asciidoctor/asciidoctor-gradle-plugin/tree/release_1_5_3#handling-additional-files

mojavelinux:
> set imagesdir in document instead of build configuration to avoid breaking HTML

This is where using separate Gradle tasks would be more elegant. You want to set imagesdir to an absolute path for the pdf converter, but not for the html5 converter. However, since the configuration is shared, any attribute you set applies to all converters. The workaround is to set the attribute in the document using the following:

```
ifdef::backend-pdf[:imagesdir: {docdir}/{imagesdir}]
```

It's just not as elegant.

Btw, `docdir` contains the absolute path of the source document.

mojavelinux:
Here's an example of how to define multiple tasks so you can pass different attributes per converter:

https://github.com/mraible/infoq-mini-book/blob/master/build.gradle

mojavelinux:
I removed the docinfo1 attribute since you are not using docinfo files.

The sectanchors attributes was misspelled as setanchors.

mojavelinux:
The other issue, which I didn't fix here, is that you have blank lines in your AsciiDoc document header. The document header must not contain blank lines or else the attributes will not be scoped to the document.

See http://asciidoctor.org/docs/user-manual/#doc-header

If you like to have some breathing room, the recommendation is to insert a comment line. For example:

```
= JUnit 5 User Guide
Author 1; Author 2
:imagesdir: images
//
sectnums:
//
:javadoc-root: link:../api
//
:snapshot-repo: https://oss.sonatype.org/content/repositories/snapshots
```

sbrannen:
Thanks for the PR, @mojavelinux! 👍

Looks like there is a _lot_ of room for improvement.

Don't have time for a full review right now, so I'll investigate more thoroughly in the coming week or so and merge your PR then.

mojavelinux:
I'm ready to answer any questions you might have.

Keep in mind, however, that if you publish without this pull request, the images in the HTML will be broken as they currently reference an absolute system path.

mojavelinux:
> Looks like there is a lot of room for improvement.

Actually, I'd say the changes are pretty minor. Switching to AsciidoctorJ Diagram and using the built-in resource copying just eliminated a lot of lines. You got it right in a lot of ways.

sbrannen:
It all looks great, @mojavelinux!

Just one minor issue: the `component-diagram.svg` is now generated in the "source" folder meaning that Git asks if we want to add it to the repo.

But I have a work-around for that. So I'll go ahead and merge your PR in _as is_.

Cheers!

sbrannen:
@mojavelinux, my _workaround_ for the aforementioned issue can be seen in 442e5d3aa9d2dc86f3b6c702a4e0d1587f3760ba.

Do you have any feedback on that technique?

sbrannen:
@mojavelinux, I addressed the blank lines in the doc-header here: db0e44aac1aeb6e8df65bd7345ed326507c035b5

Thanks for the tip!

mojavelinux:
> Just one minor issue: the component-diagram.svg is now generated in the "source" folder meaning that Git asks if we want to add it to the repo.

That shouldn't be happening. Could be a bug in the Gradle plugin. I'll investigate and get back to you. This is supposed to "just work".

mojavelinux:
Your workaround is what I would have done as well. It shouldn't be necessary to set imagesoutdir explicitly, but necessary due to a bug somewhere.

mojavelinux:
However, imagesoutdir should be a String, not a File object. It's best to resolve that.

mojavelinux:
Since buildDir is already absolute, you can simply use:

```
def buildImagesDir = new File(buildDir, 'asciidoc/images').path
```

mojavelinux:
Here's a follow-up PR that's slightly cleaner and more stable.

https://github.com/junit-team/junit5/pull/784

The "bug" I referred to in the Gradle plugin is that it does not set the outdir (and outfile) AsciiDoc attributes, even though it should. But it's easy enough to define it explicitly based on the information in the plugin.

I had mistakenly used docdir in my original PR when I meant to use outdir.

PR #787, opened 2017-04-09
Consider static nested classes as potential JUnit4 tests

LiamClark:
## Overview

The IsPotentialJunit4TestClass now checks if a memberclass is declared static.
If so it lets it through the filter.

Also includes unit tests for the IsPotentialJUnit4TestClass. It might be nice to use the
parameters support for these tests.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
+++ b/documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
@@ -54,4 +54,5 @@ on GitHub.
 
 ===== Bug Fixes
 
-* ❓
+* Fixed a bug where static member classes were incorrectly being filtered out.
+  They are now considered valid JUnit Vintage test classes.
--- a/junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/IsPotentialJUnit4TestClass.java
+++ b/junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/IsPotentialJUnit4TestClass.java
@@ -12,6 +12,7 @@ package org.junit.vintage.engine.discovery;
 
 import static org.junit.platform.commons.util.ReflectionUtils.isAbstract;
 import static org.junit.platform.commons.util.ReflectionUtils.isPublic;
+import static org.junit.platform.commons.util.ReflectionUtils.isStatic;
 
 import java.util.function.Predicate;
 
@@ -27,7 +28,14 @@ class IsPotentialJUnit4TestClass implements Predicate<Class<?>> {
 			return false;
 		if (!isPublic(candidate))
 			return false;
-		return !candidate.isMemberClass();
+		if (isNonStaticMemberClass(candidate))
+			return false;
+
+		return true;
+	}
+
+	private boolean isNonStaticMemberClass(Class<?> candidate) {
+		return candidate.isMemberClass() && !isStatic(candidate);
 	}
 
 }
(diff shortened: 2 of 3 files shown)


junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/IsPotentialJUnit4TestClass.java
```diff
@@ -22,12 +23,20 @@
 
 	@Override
 	public boolean test(Class<?> candidate) {
-		// Do not collapse into single return.
 		if (isAbstract(candidate))
 			return false;
 		if (!isPublic(candidate))
 			return false;
-		return !candidate.isMemberClass();
+		// Do not collapse into single return.
```
marcphilipp: I think this comment should stay at the top.

documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
```diff
@@ -54,4 +54,5 @@ on GitHub.
 
 ===== Bug Fixes
 
-* ❓
+* The `IsPotentialJUnit4TestClass` predicate incorrectly filtered out static nested classes.
+  It now let's them through properly
```
marcphilipp: I think that's a bit too technical. We should try to write this from the point of view of a Vintage user. It should be "static member classes" because that excludes local and anonymous classes. Moreover, the last sentence is missing a full stop/period.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/IsPotentialJUnit4TestClass.java
LiamClark: I moved it here because this is where intellij generates the warning / quick fix for collapsing it. But if you want it at the top i'll move it.

marcphilipp: Yeah, I think that was the original intent.

LiamClark: So the warning moved here.
warning-moved2

So I should keep the comment here?

documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
LiamClark: Would this be better?

>Fixed a bug where static member classes were incorrectly being filtered out.
  They are now considered valid JUnit Vintage test classes.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/IsPotentialJUnit4TestClass.java
marcphilipp: Well, before the warning was on the second `if`. My point is: You could write this method body as

```java
return !isAbstract(candidate) && isPublic(candidate) && !isNonStaticMemberClass(candidate);
```

but we prefer the `if` statements in this case. Thus, the comment is valid for the whole method body and therefore should be at the very top. Does that make sense?

LiamClark: yes!

documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
marcphilipp: 👍

marcphilipp:
Thanks, @LiamClark! :+1:

Issue #775, opened 2017-04-06
Compile and test on Java 9


sormuras:
## Moved "Lessons learned..." to Wiki

[Lessons learned "JUnit 5 compile and test on Java 9 b162"](https://github.com/junit-team/junit5/wiki/Lessons-learned-%22JUnit-5---compile-and-test-on-Java-9-b162%22)

Here, only the remaining technical steps resulting from the lessons learned are retained.

## Related Issues

- #425

## Deliverables

- [x] Upgrade to Gradle 3.5 - done in master with https://github.com/junit-team/junit5/commit/5ef538f7b7e953f89053de6986e254ee737cf918
- [x] Add information tag stating location from where an engine was loaded from - see #788 
- [x] Merge main Java 8 and this Java 9 build into a single `.travis.yml` script -- via https://github.com/junit-team/junit5/commit/a313a075353a573d97b443ba57dd83db0f00f8fd
- [x] Include [AssertJ 3.7.0](https://github.com/joel-costigliola/assertj-core/milestone/28) - remove jitpack from repositories
- [x] Include [Gradle Shadow 2.0.0](https://github.com/johnrengelman/shadow/milestone/18) - remove jfrog.org/artifactory from repositories
- [ ] ~~Report and~~ remove workaround for ServiceLoader "bug" when "everybody" has access to build 163

sbrannen:
@sormuras, thanks for compiling all of this information! 👍

sbrannen:
FYI: the Shadow plugin has already been upgraded to ASM 5.2: https://github.com/johnrengelman/shadow/issues/285

sbrannen:
Why not just suppress deprecation warnings on the call for now?

Or... would that mess up the JDK 8 build? 😜

sbrannen:
It'll also be nice to be able to switch to [trySetAccessible()](http://download.java.net/java/jdk9/docs/api/java/lang/reflect/AccessibleObject.html#trySetAccessible--), but we'll have to wait until JUnit 6 for that probably.

jbduncan:
Alternatively to waiting for JUnit 6, we might be able to use MR-JAR support coming with Java 9, once we have a clear picture of how MR-JARs fit into Gradle. :)

sbrannen:
> Alternatively to waiting for JUnit 6, we might be able to use MR-JAR support coming with Java 9, once we have a clear picture of how MR-JARs fit into Gradle. :)

[...]

PR #788, opened 2017-04-10
Add ClassLoaderUtils and log location from which an engine was loaded

sormuras:
## Overview

This commit enhances the configuration-level log statement with the location URL from where the classloader loaded the engine class. This commit also introduces dedicated `ClassLoaderUtils` class that encapsulates static class-related helper methods.

Addresses part of #775 

---

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
+++ b/documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
@@ -20,9 +20,6 @@ on GitHub.
 
 ===== Deprecations and Breaking Changes
 
-* The deprecated `execute(LauncherDiscoveryRequest)` method has been removed from the
-  `org.junit.platform.launcher.Launcher` API; use
-  `execute(LauncherDiscoveryRequest, TestExecutionListener...)` instead.
 * The deprecated `--hide-details` option of the `ConsoleLauncher` has been removed; use
   `--details none` instead.
 
@@ -47,7 +44,9 @@ on GitHub.
 
 ===== New Features and Improvements
 
-* ❓
+* Added internal utility method to provide information about the location from which
+  the engine class was loaded. When the location URL is available, it is logged at
+  configuration-level by the service loading mechanism.
 
 
 [[release-notes-5.0.0-m5-junit-vintage]]
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.platform.commons.util;
+
+import static org.junit.platform.commons.meta.API.Usage.Internal;
+
+import java.net.URL;
+import java.security.CodeSource;
+import java.util.Optional;
+
+import org.junit.platform.commons.meta.API;
+
+/**
+ * Collection of utilities for working with {@linkplain ClassLoader} and associated tasks.
+ *
+ * <h3>DISCLAIMER</h3>
+ *
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.0
+ */
+@API(Internal)
+public final class ClassLoaderUtils {
+
+	///CLOVER:OFF
+	private ClassLoaderUtils() {
+		/* no-op */
+	}
+	///CLOVER:ON
+
+	public static ClassLoader getDefaultClassLoader() {
+		try {
+			ClassLoader contextClassLoader = Thread.currentThread().getContextClassLoader();
+			if (contextClassLoader != null) {
+				return contextClassLoader;
+			}
+		}
+		catch (Throwable ex) {
+			/* ignore */
+		}
+		return ClassLoader.getSystemClassLoader();
+	}
+
+	/**
+	 * Get the location from which the supplied object's class was loaded.
+	 *
+	 * @param object the object for whose class the location should be retrieved
+	 * @return an {@code Optional} containing the URL of the class' location; never
+	 * {@code null} but potentially empty
+	 */
+	public static Optional<URL> getLocation(Object object) {
+		Preconditions.notNull(object, "object must not be null");
+		// determine class loader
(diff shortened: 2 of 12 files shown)


documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
```diff
@@ -47,7 +47,9 @@ on GitHub.
 
 ===== New Features and Improvements
 
-* ❓
+* Added `getSourceLocation()` method in `TestEngine` interface to provide information
+  about the location from where the engine class was loaded from. When location URL is
```
sbrannen: "the location from which the engine class was loaded."

sbrannen: "When the location URL is available..."

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
```diff
@@ -0,0 +1,64 @@
...
+public final class ClassLoaderUtils {
+
+	///CLOVER:OFF
+	private ClassLoaderUtils() {
+		/* no-op */
+	}
+	///CLOVER:ON
+
+	/**
+	 * Get the location from where this objects underlying class was loaded from.
```
sbrannen: "Get the location from which the supplied object's class was loaded."

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
```diff
@@ -0,0 +1,64 @@
...
+	 *
+	 * @param object to find the location its class was loaded from for
```
sbrannen: This is a tough one, grammatically speaking, but something like the following is more suitable:

```
@param object the object for whose class the location should be retrieved
```

In general, we always follow the pattern `@param object the object...`.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
```diff
@@ -0,0 +1,64 @@
...
+		ClassLoader loader = object.getClass().getClassLoader();
+		if (loader == null) {
+			loader = ClassLoader.getSystemClassLoader();
+			while (loader != null && loader.getParent() != null) {
+				loader = loader.getParent();
+			}
+		}
+		if (loader != null) {
+			String name = object.getClass().getCanonicalName();
+			name = name.replace(".", "/") + ".class";
```
sbrannen: Don't we have this kind of "replace" call in multiple places in the code base now?

If so, I think that would warrant the introduction of a new `ClassUtils`.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.0
+ */
+@API(Internal)
+public final class ClassLoaderUtils {
+
```
sbrannen: Please relocate `org.junit.platform.commons.util.ReflectionUtils.getDefaultClassLoader()` to this class as well.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
```diff
@@ -0,0 +1,64 @@
...
+		}
+		if (loader != null) {
+			String name = object.getClass().getCanonicalName();
+			name = name.replace(".", "/") + ".class";
+			URL resource = loader.getResource(name);
+			return Optional.ofNullable(resource);
+		}
+		return Optional.empty();
+	}
+}
```
sbrannen: I'm wondering if we shouldn't wrap the meat of `getLocation()` in a try-catch block _just in case..._.

junit-platform-engine/src/main/java/org/junit/platform/engine/TestEngine.java
```diff
@@ -175,4 +177,16 @@
 			PackageUtils.getAttribute(getClass(), Package::getImplementationVersion).orElse("DEVELOPMENT"));
 	}
 
+	/**
+	 * Get the location from where this engine was loaded from.
```
sbrannen: "Get the location from which this engine was loaded."

platform-tests/src/test/java/org/junit/platform/commons/util/ClassLoaderUtilsTests.java
```diff
@@ -0,0 +1,42 @@
...
+package org.junit.platform.commons.util;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for {@link PackageUtils}.
```
sbrannen: `PackageUtils`? 😜

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
sormuras: "Find in path..." revealed only this occurance of `replace(".", "/")` -- so no need for a `ClassUtils` collection for this specific use-case.

On the other hand, what happens when the class is a nested one, that is separated with a `$`-sign? Going to write a test for that.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
sormuras: Makes sense.

We could also try the `.getProtectionDomain().getCodeSource().getLocation()` as a last resort here.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
sormuras: Relocating in a separate commit to gather all 14 related file changes in a single place.

sormuras: Done via https://github.com/junit-team/junit5/pull/788/commits/89c98340d1e7aff0cb587c1113692a5183704308

sbrannen: 👍

sormuras:
Failing "pr" build: https://travis-ci.org/junit-team/junit5/builds/221362686
But working "push" build: https://travis-ci.org/junit-team/junit5/builds/221362671

?

Re-creating this PR.

Issue #750, opened 2017-03-26
DefaultLauncher should catch exceptions thrown by test engines


jlink:
## Overview

If a test engine throws an exception during discovery (see https://github.com/junit-team/junit5/blob/master/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java#L115) other engines won't be queried any more.

I think this is a bug because one test engine can thereby prevent others from working. 

## Deliverables

- [x] `DefaultLauncher` should catch and report exceptions from engines and proceed with other engines.
  - This applies to both the discovery phase and execution phase.

## Additional Proposals

- `DefaultLauncher` could even collect discovery results concurrently.

sbrannen:
Thanks for raising the issue.

I totally agree!

The same applies to the execution phase as well.

sbrannen:
> `DefaultLauncher` could even collect discovery results concurrently.

It could do that, but I consider that a separate topic. I have therefore updated this issue's description to reflect the deliverables.

Feel free to create a new issue to propose concurrent test discovery across registered engines.

sbrannen:
Slated for M5 since we are trying to get M4 out the door as soon as possible.

sormuras:
With #792 merged, an erroneous engine implementation doesn't any longer break the test plan execution. 

We should decide if a "simple warning" is enough for the end-user to take note of the failed engine. Perhaps the execution summary needs an update, like:

```
Test run finished after 3091 ms
[         2 engines found         ]
[         1 engines successful    ]
[         1 engines aborted       ]
[         0 containers skipped    ]
[        70 containers started    ]
[         0 containers aborted    ]
[        70 containers successful ]
[         0 containers failed     ]
[       602 tests found           ]
[         0 tests skipped         ]
[       602 tests started         ]
[         1 tests aborted         ]
[       601 tests successful      ]
[         0 tests failed          ]
```

Updated issue description accordingly.

sbrannen:
> We should decide if a "simple warning" is enough for the end-user to take note of the failed engine. Perhaps the execution summary needs an update, like:

I think we should go ahead and _close_ this issue and address the topic of _exception handling vs. logging_ in conjunction with #242 (thereby broadening the scope of #242 to cover the Platform as well).

marcphilipp:
👍

sbrannen:
I have moved the second deliverable to #242 and am therefore closing this issue.

PR #792, opened 2017-04-13
Catch exceptions in DefaultLauncher and proceed

sormuras:
## Overview

`DefaultLauncher` now catches and reports exceptions from engines and proceeds with other engines.
This applies to both the discovery phase and execution phase.

Addresses #750 

---


junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -111,17 +114,28 @@ private Root discoverRoot(LauncherDiscoveryRequest discoveryRequest, String phas
...
-			UniqueId uniqueEngineId = UniqueId.forEngine(testEngine.getId());
+			Optional<TestDescriptor> engineRoot = discoverRoot(testEngine, discoveryRequest);
+			engineRoot.ifPresent(rootDescriptor -> root.add(testEngine, rootDescriptor));
+		}
+		root.applyPostDiscoveryFilters(discoveryRequest);
+		root.prune();
+		return root;
+	}
+
+	private Optional<TestDescriptor> discoverRoot(TestEngine testEngine, LauncherDiscoveryRequest discoveryRequest) {
```
sormuras: Renamed method to `discoverEngineRoot`.

PR #796, opened 2017-04-18
Remove semicolons at the end of enums

mureinik:
## Overview

This patch removes redundant semicolons in the end of enum constants
lists.
While these semicolons aren't technically wrong, they're redundant.

---

---

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/HierarchyTraversalMode.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/HierarchyTraversalMode.java
@@ -33,6 +33,6 @@ public enum HierarchyTraversalMode {
 	/**
 	 * Traverse the hierarchy using bottom-up semantics.
 	 */
-	BOTTOM_UP;
+	BOTTOM_UP
 
 }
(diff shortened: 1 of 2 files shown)


marcphilipp:
Thanks! 👍

smoyer64:
This is **not** a recommended practice for JUnit 5 - Leaving the unnecessary semi-colon in place has another consequence that I've seen people use for robustness - you can put a comma after the last enum value.  With the prevalence of IDEs I'm not entirely sure how valuable this is as a missing comma is easily caught by the syntax checker.

sbrannen:
I agree with @smoyer64.

I do not approve of this being merged into `master`.

The presence of those semicolons is no accident.

I **intentionally** typed those semicolons.

Just because Java syntax _allows_ one to omit such a semicolon does _not_ mean that one should.

1. There are zero negative side effects due to the presence of such a semicolon.
2. The presence of such a semicolon makes it easier to refactor the code by adding constructors or methods to the enum at a later date without having to needlessly insert a semicolon.
3. The presence of such a semicolon makes it clear that the author completed the task at hand.

I strongly vote to revert this PR!

p.s.

>  While these semicolons aren't technically wrong, they're redundant.

They are not "redundant" by definition. You may argue that they are _superfluous_ in the current context of the code, but that is not forward thinking, and a framework as ubiquitous as JUnit must be forward thinking.

marcphilipp:
I wasn't aware you felt so strongly about those. I'm fine with keeping the semicolons and have reverted this change in a0c85d0c6f6226abdcebe5f614315fa8d9dd02d6.

PR #798, opened 2017-04-19
Added favicon to userguide and javadocs.

Andrei94:
This pull request adds the favicon to the userguide and the javadocs.


marcphilipp:
This change would add the favicon to the generated output of the Javadoc and AsciiDoctor tasks. Instead, the favicon should be added when those artifacts are generated.

Andrei94:
Any hints of where the source could be? I thought that this was the place where I should edit.

PR #841, opened 2017-05-07
Introduce engine-level ExtensionContext

### LINKED ISSUE #671: Introduce a top-level ExtensionContext for the JupiterTestEngine
## Feature Request

When using the available extension contexts in test runs with extensions like `BeforeAllCallback` and `BeforeTestExecutionCallback`, it only seems possible to use these contexts during one test and lose all context information in following test runs.

Trying to use the parent context of an instance of `ContainerExtensionContext` in a `beforeAllCallback` leads to an empty `Optional` object, thus no information can be used or set.

```java
import java.util.logging.Logger;

import org.junit.jupiter.api.extension.BeforeAllCallback;
import org.junit.jupiter.api.extension.BeforeTestExecutionCallback;
import org.junit.jupiter.api.extension.ContainerExtensionContext;
import org.junit.jupiter.api.extension.TestExtensionContext;

/**
 * @author Simon Ebner
 */
public class ParentAwareExtension implements BeforeAllCallback, BeforeTestExecutionCallback {

	@Override
	public void beforeAll(ContainerExtensionContext context) throws Exception {
		/**
		 * Issue - no parent context during test execution available to act on
		 * or use any global context information
		 */
		context.getParent().ifPresent(parent -> {
			Logger.getLogger("JUnit5-Issue").info("'before all' has parent context " + parent.getDisplayName());
		});
	}

	@Override
	public void beforeTestExecution(TestExtensionContext context) throws Exception {
		context.getParent().ifPresent(parent -> {
			Logger.getLogger("JUnit5-Issue").info("'before test' has parent context " + parent.getDisplayName());
		});
	}
}
```

It would be great to be able to set and use context information for an entire test run, especially when needing to execute time and resource expensive operations, that could be reduced to one call per entire test run.

## Suite Considerations

I am not sure as to how it would be necessary to look also at `@SelectClasses` and `@SelectPackages` in this case, because they also may create a distinct series of tests and thus a logical unit, that might need its own context to be able to separate tests that run in a suite and others. I would like to know your opinion in this respect.

marcphilipp:
Fixes #671.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/execution/ExtensionContextTests.java
+++ b/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/execution/ExtensionContextTests.java
@@ -10,20 +10,21 @@
 
 package org.junit.jupiter.engine.execution;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Collections;
 import java.util.Map;
-import java.util.Optional;
 
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.engine.descriptor.ClassBasedContainerExtensionContext;
 import org.junit.jupiter.engine.descriptor.ClassTestDescriptor;
+import org.junit.jupiter.engine.descriptor.JupiterEngineDescriptor;
+import org.junit.jupiter.engine.descriptor.JupiterEngineExtensionContext;
 import org.junit.jupiter.engine.descriptor.MethodBasedTestExtensionContext;
 import org.junit.jupiter.engine.descriptor.MethodTestDescriptor;
 import org.junit.jupiter.engine.descriptor.NestedClassTestDescriptor;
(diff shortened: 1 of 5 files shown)


junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/execution/ExtensionContextTests.java
```diff
@@ -36,22 +40,39 @@
...
+			UniqueId.root("engine", "junit-jupiter"));
+
+		JupiterEngineExtensionContext engineContext = new JupiterEngineExtensionContext(null, engineTestDescriptor);
+
+		assertAll("engineContext", //
+			() -> assertThat(engineContext.getTestClass()).isEmpty(), //
+			() -> assertThat(engineContext.getTestMethod()).isEmpty(), //
+			() -> assertThat(engineContext.getElement()).isEmpty(), //
+			() -> assertEquals(engineTestDescriptor.getDisplayName(), engineContext.getDisplayName()), //
+			() -> assertEquals(Optional.empty(), engineContext.getParent()) //
```
sbrannen: I'd simply use `assertThat()` here as well.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ExtensionContextExecutionTests.java
```diff
@@ -0,0 +1,96 @@
...
+				throws ParameterResolutionException {
+			return extensionContext;
+		}
+	}
+
+	@Test
+	void twoTestClassesCanShareStateViaEngineExtensionContext() {
+		Parent.counter.set(0);
+
+		executeTests(request().selectors(selectClass(A.class), selectClass(B.class)).build());
```
sbrannen: I know we assume tests from both test classes get executed, but maybe it would be good to assert that assumption as well. 😉

LymanEfunktion:
thank your for the work and new feature

frnusmartobjects:
If I understand this feature correctly it will allow to pass information between two test classes during the same test run/execution session, right? If so are there any examples on how to use this feature?

We have a base class (`BaseClass`) annotated with `@ExtendWith(MyExtension.class)`, where `MyExtension` is a class that implements `TestTemplateInvocationContextProvider` that allows to pass different databases connections to the method annotated with `@BeforeEach` of `BaseClass` using templates. `BaseClass` is inherited by multiple test classes and we would like to avoid recreating the database on each test class/test method and only create each database only once per the whole test execution.

Is it possible to do that using this feature? We are experimenting with implementing `BeforeAllCallback` in `MyExtension` and storing in `context.getStore(ExtensionContext.Namespace.GLOBAL)` the different databases but we aren't sure if this is the right approach

Thank you in advance.

marcphilipp:
You'll have to store it in the root `ExtensionContext` in order for other classes to see it, i.e. `context.getRoot().getStore(GLOBAL)`.

stickfigure:
It looks like this mechanism will allow me to start up the resource once for the entire test run, which is great. Is there a way to shut down the resource once at the end?

marcphilipp:
It will, once #1174 is merged. 😉

Issue #845, opened 2017-05-09
Make Assertions.fail usable as an expression


sormuras:
## Overview

Change return value of `Assertions.fail` from `void` to `<V> V`, in order to make `fail` directly usable in places where an expression is needed. The original request can be found [here](https://github.com/forax/pro/commit/eb5f1b7b6df63c75ffa7a14a61d8d4c596b627bc#commitcomment-22058065).

Like:
```
public static <V> V fail(String message) {
  AssertionUtils.fail(message);
  return null; // will never return
}
```
which would allow end-users to write:
```
  @Test
  void failUsableAsAnExpression() {
    List<Number> list = new ArrayList<>();
    long count = list.stream()
        .peek(number -> fail("peek should never be called"))
        .filter(number -> fail("filter should never be called"))
        .map(number -> fail("map should never be called"))
        .sorted((n, m) -> fail("sorted should never be called"))
        .count();
    assertEquals(0L, count);
  }
```
without creating a lambda on their own.

## Deliverables

- [ ] Implement and test `<V> V fail()` for all 4 overloads of `Assertions.fail`

PR #847, opened 2017-05-09
Allow Assertions.fail methods to be used as expressions

sormuras:
## Overview

Prior to this commit the return type `void` prevented the use of a `fail`-method as an expression. For example, the user has to wrap the `fail` call in a lambda like:

```java
  Stream.of().map(entry -> { fail("should not be called"); return ""; })
```

Now, the line can be written more clearly as:

```java
  Stream.of().map(entry -> fail("should not be called"))
```

Addresses #845

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
@@ -50,32 +50,56 @@ public final class Assertions {
 
 	/**
 	 * <em>Fails</em> a test with the given failure {@code message}.
+	 *
+	 * <p>The generic return type {@code V} allows this method to be used directly
+	 * as an expression value. As an {@link org.opentest4j.AssertionFailedError}
+	 * is created and thrown before the final return statement, this method never
+	 * returns a value to its caller.
 	 */
-	public static void fail(String message) {
+	public static <V> V fail(String message) {
 		AssertionUtils.fail(message);
+		return null; // appeasing the compiler: this line will never be executed.
 	}
 
 	/**
 	 * <em>Fails</em> a test with the given failure {@code message} as well
 	 * as the underlying {@code cause}.
+	 *
+	 * <p>The generic return type {@code V} allows this method to be used directly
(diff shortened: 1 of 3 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -51,31 +51,35 @@ private Assertions() {
 	/**
 	 * <em>Fails</em> a test with the given failure {@code message}.
 	 */
-	public static void fail(String message) {
+	public static <V> V fail(String message) {
```
nicolaiparlog: Shouldn't this be `Optional<V>`? 🤣

sormuras: No. That won't work. The (**never**) returned type is unknown until compilation. And it has to adopt to every possible type.

`boolean` here: https://github.com/forax/pro/commit/eb5f1b7b6df63c75ffa7a14a61d8d4c596b627bc#diff-3cb01c85a2fef586b2a43f61d42e2f3aR97
`String` there: https://github.com/forax/pro/commit/eb5f1b7b6df63c75ffa7a14a61d8d4c596b627bc#diff-3cb01c85a2fef586b2a43f61d42e2f3aR107

nicolaiparlog: Apparently the 🤣 did not suffice to mark this as a joke. :wink:

sormuras: ;)

sormuras:
This feature is also neat to have, when you test methods supporting callbacks that should not be called everytime. See https://github.com/forax/pro/pull/44/commits/91b695d697c075bc6554c6b775fb677fc177651b#diff-27a5653516a2a1477f33b4cc6c331988R57

sormuras:
Rebased and add javadoc about the generic return type to each `fail` method. Awaiting M6 file to appear...

Issue #765, opened 2017-04-03
Rename ObjectArrayArguments.create() to arguments()


sbrannen:
## Overview

In order to improve the user experience and allow for the static factory method in `ObjectArrayArguments` to be used via a static import, the name should be changed to something more DSL-like.

## Deliverables

- [x] Rename `ObjectArrayArguments.create()` to `arguments()`.

sbrannen:
Resolved in 30438b4dec1de88a1e6d7737b35449486f56f015.

sbrannen:
Further addressed in c810d13142fc74b8f5b63864c3451a4eed97e322.

PR #863, opened 2017-05-23
Make Arguments creation simpler

nicolaiparlog:
I know nobody asked for this but it makes things a lot easier. If you think this is a good idea, I will eliminate `ObjectArrayArguments` entirely.

---

---

### Definition of Done


junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -31,4 +31,8 @@
 
 	Object[] get();
 
+	static Arguments create(Object... args) {
```
marcphilipp: I think we should call it `arguments`.

nicolaiparlog:
Assuming the build goes through I have nothing more to add here.

smoyer64:
I'm not trying to be argumentative but ... I read @marcphilipp code review comment as changing:

```
static Arguments create(Object... args) {
```

into:

```
static Arguments create(Object... arguments) {
```

In use, isn't ```create``` less repetitive?

```
Arguments arguments = Arguments.arguments(myArguments);
```

sormuras:
@smoyer64 see #765

nicolaiparlog:
@smoyer64 I went with the flow but I absolutely agree. It is used as a method reference almost as often as with a static import and `Arguments::create` would look much better then `Arguments::arguments`. Short of declaring both methods, you can't really win here, though...

sbrannen:
Another option would be to name the static factory method `of` or `from`, resulting in `Arguments.of(...)` & `Arguments::of` or `Arguments.from(...)` & `Arguments::from`.

If we go that route it would then be awkward as a static import to just see `of` or `from`, but I could live with that, especially since it's the trend for APIs in the JDK (a la `Stream.of(...)`).

sormuras:
> If we go that route it would then be awkward as a static import to just see of or from, but I could live with that, especially since it's the trend for APIs in the JDK (a la Stream.of(...)).

Good point. Let's change the factory name to `Arguments.of(...)` merge this.

nicolaiparlog:
As you command. :smiley:

documentation/src/test/java/example/ParameterizedTestDemo.java
```diff
@@ -15,7 +15,7 @@
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.params.provider.Arguments.arguments;
+import static org.junit.jupiter.params.provider.Arguments.of;
```
sormuras: Is this static import still used?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileArgumentsProvider.java
```diff
@@ -105,7 +105,7 @@ public boolean hasNext() {
 
 		@Override
 		public Arguments next() {
-			Arguments result = arguments(this.nextCsvRecord);
+			Arguments result = of(this.nextCsvRecord);
```
sormuras: Here, I like to see the "long" form: `Arguments.of(this.nextCsvRecord)`

`of` is really (too) short. Even when "result" is an `Arguments` type.

What happens, when the next line reads (in Java 9):
```
  List<String> list = of("1", "2"); // import static java.util.List.of;
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
```diff
@@ -10,7 +10,7 @@
 
 package org.junit.jupiter.params.provider;
 
-import static org.junit.jupiter.params.provider.Arguments.arguments;
+import static org.junit.jupiter.params.provider.Arguments.of;
```
sormuras: Unused, I guess.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileArgumentsProvider.java
nicolaiparlog: It won't because you can't statically import two methods of the same name. :wink:

I can still change it but I prefer it this way, precisely because the return type is at the beginning of the line.

sormuras: Fine. As long **javac** and **spotless** are happy...

Hey! What about calling the method `$`? ;)

nicolaiparlog: [Don't tempt me!](https://github.com/junit-pioneer/junit-pioneer/pull/45)

documentation/src/docs/asciidoc/release-notes-5.0.0-M5.adoc
```diff
@@ -72,8 +72,9 @@ on GitHub.
 ** See <<release-notes-5.0.0-m5-migration-extension-api>> for details.
 * The `arguments()` method in the `ArgumentsProvider` API in the `junit-jupiter-params`
   module has been renamed to `provideArguments()`.
-* The `create()` factory method in `ObjectArrayArguments` in the `junit-jupiter-params`
-  module has been renamed to `arguments()`.
+* The type `ObjectArrayArguments` in the `junit-jupiter-params` module has been removed;
+  its functionality to create an `Argument` instance with `create()` is fulfilled by the
+  static factory method `arguments()` on the interface `Arguments`.
```
sormuras: static factory method `Arguments.of()` ~~on the interface `Arguments`~~.

documentation/src/docs/asciidoc/writing-tests.adoc
```diff
@@ -601,8 +601,7 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=primitive_MethodSourc
 ----
 
 In case you need multiple parameters, you need to return `Argument` instances as shown below.
-Note that `arguments(Object...)` is a static factory method defined in
-`org.junit.jupiter.params.provider.ObjectArrayArguments`.
+Note that `arguments(Object...)` is a static factory method defined in the interface itself.
```
sormuras: Note that `Argumentsof(Object...)`...

sormuras:
Cleanup of the M5 notes will happen on master

sormuras:
And kudos to @nicolaiparlog

Issue #896, opened 2017-06-22
Assertions dependency on another assertion


FilipJirsak:
## Overview

It would be nice to have possibility to execute some assertions only if another assertion was successful. For example I test if some object is not null and then I test object's properties. It is useless to test properties when non-null check failed.

It is different  from assumes, because test should fail if condition is not fulfiled.

## Deliverables
There can be new methods in Assertions:
```java
public static void assertFirstThenOthers(Executable contolExecutable, Executable... dependentExecutables);
public static void assertFirstThenOthers(Executable contolExecutable, Stream<Executable> dependentExecutables);
//plus variants with heading
```
Then I can write test like this:
```java
class BeanTest {
  void testBean() {
    Bean bean = new Bean();
    //…
    assertAll(
      () -> assertEquals("A", bean.getPropertyA()),
      () -> assertEquals("B", bean.getPropertyB()),
      () -> assertFirstThenOthers(
        () -> assertNotNull(bean.getNestedBean()),
        () -> assertEquals("C", bean.getNestedBean().getPropertyC()),
        () -> assertEquals("D", bean.getNestedBean().getPropertyD())
      )
    );
  }
}
```

Or even smarter one:
```java
class BeanTest {
  void testBean() {
    Bean bean = new Bean();
    //…
    assertAll(
      () -> assertEquals("A", bean.getPropertyA()),
      () -> assertEquals("B", bean.getPropertyB()),
      () -> assertFirstThenOthers(
        () -> assertNotNull(bean.getNestedBean()), //inject result of first test as param to others tests
        (nested) -> assertEquals("C", nested.getPropertyC()),
        (nested) -> assertEquals("D", nested.getPropertyD())
      )
    );
  }
}
```

It can be written with assumptions, but it needs code duplication:
```java
class BeanTest {
  void testBean() {
    Bean bean = new Bean();
    //…
    assertAll(
      () -> assertEquals("A", bean.getPropertyA()),
      () -> assertEquals("B", bean.getPropertyB()),
      () -> assertNotNull(bean.getNestedBean()),
      () -> assumingThat(bean.getNestedBean() != null,
        () -> assertAll(
          () -> assertEquals("C", bean.getNestedBean().getPropertyC()),
          () -> assertEquals("D", bean.getNestedBean().getPropertyD())
        )
      )
    );
  }
}
```

sbrannen:
This is already possible with the following construct:

```java
assertAll(
[...]

PR #897, opened 2017-06-23
example of multiple assertions in one codeblock

FilipJirsak:
Add example of multiple assertions in one code block. Can be used for dependent assertions.
Based on issue #896.


sbrannen:
Thanks for the PR.

PR #918, opened 2017-07-05
Check that the discovery request is not null

stefanbirkner:
## Overview

The JupiterTestEngine checks this already. I don't see any reason why the VintageTestEngine should behave differently.

---

I hereby agree to the terms of the JUnit Contributor License Agreement.

---

### Definition of Done

- [x] There are no TODOs left in the code
- [x] Method [preconditions](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [x] [Coding conventions](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [x] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#tests)
- [ ] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/meta/API.html)
- [ ] Change is documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [x] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

Issue #358, opened 2016-06-25
Clarify when returned values might be null


smoyer64:
When the `ParameterResolver`'s `supports()` and `resolve()` methods were changed to use `ParameterContext`, the statement that the `Parameter` was never `null` was replaced by text that stated `ParamterResolver` is never null.  `ParameterResolver` itself makes no assertion that [getParameter()](https://junit.ci.cloudbees.com/job/JUnit5/javadoc/org/junit/jupiter/api/extension/ParameterContext.html#getParameter) always returns a non-null value.  Thus, I'm now checking for null parameters as follows:

```java
    boolean supported = false;
    Parameter parameter = parameterContext.getParameter();
    if (parameter != null && Performance.class.equals(parameter.getType())) {
      supported = true;
    }
    return supported;
```

For the `ParameterContext`, it wouldn't make sense to ever have a non-null parameter (otherwise it would be a `SomethingElseContext`), so this code is probably completely unnecessary.  As a defensive coder, I put it in anyway.  Here are my recommendations in order of priority/complexity:

-   [ ] For every returned value, explicitly state in the Javadoc whether or not the return type may be null.
-   [ ] Globally use `Optional` when a return type might be `null`.
-   [ ] State in the User Guide that no public APIs ever return `null`.
-   [ ] Add the JSR-305 annotation library with a provided scope and add `@CheckForNull`, `@Nonnull` and `@Nullable` to all public APIs.

I know the JUnit team's policy on avoiding dependencies - adding JSR-305 as provided allows those of us who chose to perform static analysis on our Engine, Extension and test code to choose to add the dependency to our classpath.

## Related Resources

- There is additional detail in #741.
- [SPR-15540](https://jira.spring.io/browse/SPR-15540) and the [related commit](https://github.com/spring-projects/spring-framework/commit/f813712f5b413b354560cd7cc006352e9defa9a3).

sbrannen:
Thanks for raising the issue.

The omission of "never null" from the Javadoc for `ParameterContext` was simply an oversight that I have now addressed, and we will look into the rest of your proposals during the M2 time frame.

smoyer64:
Thanks!

I thought that might be the case ... and for ParameterResolvers, just that little sentence makes a huge difference since (almost) every supports() and resolve() call wants access to the parameter.

(Plus I love it when my code base shrinks)

marcphilipp:
See #741 for additional motivation to use annotations to support static analysis and Kotlin interoperability.

[...]

PR #919, opened 2017-07-06
Don't check that the discovery request is not null

stefanbirkner:
## Overview

Engines are called by a launcher. The DefaultLauncher never calls an engine with a null request. We trust all callers of the `discover` method to use the engine properly. Hence there is no need to check for null. (This change was proposed by @marcphilipp in #918.)

---

---

### Definition of Done


stefanbirkner:
@marcphilipp Do you think we should add "Must not be null." to the `@param` JavaDoc in the interface?

marcphilipp:
Yep, I think we should add `; never {@code null}` to all `@param` tags of all methods in the interface.

stefanbirkner:
I do that in a separate commit because it also affects other parameters that are not logically part of that change.

marcphilipp:
Thanks! 👍

smoyer64:
Related to #358

Issue #581, opened 2016-12-02
Is it possible to have test filters use full boolean semantics?


littleclay:
I see in the User Guide that filters for tests use the same OR semantics as JUnit 4's categories. Unfortunately, we've run into multiple instances where the OR semantics were not powerful enough. Take the following setup:

```
public interface Foo { }
public interface Bar { }

public class SomeTest {

    @Test
    @Category({ Foo.class })
    public void testFoo() { }

    @Test
    @Category({ Bar.class })
    public void testBar() { }

    @Test
    @Category({ Foo.class, Bar.class })
    public void testBoth() { }
}
```

If you want to run all tests with just `Foo` or `Bar` that's easy. But what if you **only** want to run tests with **both** `Foo` and `Bar` (e.g. just the method `testBoth`)? This is only a simple example, but it would be really nice if the filters for tags in JUnit 5 allowed for the execution of tests that have full boolean expressions. Something like:

```
mvn test --groups=Foo&Bar
```

to denote that both `Foo` and `Bar` need to be present.

smoyer64:
@littleclay There have been other discussions about how complex the include/exclude logic should be.  Since I've just completed adding the filtering to the maven-surefire-plugin, I can tell you that the semantics of the filter specifications in the Maven configuration file (POM) will still be a limiting factor (groups and excludedGroups are synonyms for includeTags and excludeTags for JUnit 5):

-   http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#groups
-   http://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#excludedGroups

marcphilipp:
I'm closing this as a duplicate of #454.

@signed is already working on an implementation that will allow to specify full boolean expressions like `Foo & Bar`.

PR #927, opened 2017-07-09
[RFC] tag expressions

signed:
## Overview
A proof of concept implementation of a tag expression language (#454 #581).
The implementation is based on [tag expressions in cucumber][cucumber-source-code].
The basic idea is documented [here](https://docs.cucumber.io/tag-expressions/).

The language consists of the reserved case sensitive characters/words `(`, `)`, `not`, `or`, `and`.

I implemented a new tokenizer to support tags with white space characters.
Due to `@Tag("my tag")` in `TestInfoDemo` I assume this is required.
If white space in tags is not required the cucumber code could be used as is.

If this is a feature you consider to merge I'll work on the code until it complies with the DoD and open a clean pull request.
These are the open Todo's I see at the moment.

## Todo's
- [ ] proper attribution to the cucumber project (MIT license)
- [ ] documentation in the user guide
- [ ] fix test in AutomaticModuleNameTests.java
- [ ] add more tests
- [ ] extend gradle plugin ?
- [ ] extend surefire provider ?

---

---

### Definition of Done

[cucumber-source-code]: https://github.com/cucumber/cucumber/tree/master/tag-expressions/java
[cucumber-documentation]: https://docs.cucumber.io/tag-expressions/


jbduncan:
There's a few low-level areas where the code could be improved, but it's not clear to me yet if this feature is something that the JUnit 5 team would want, so I'll hold back until there's confirmation. :)

marcphilipp:
@signed Thanks for submitting this RFC! Overall, I think the complexity looks manageable.

Regarding the syntax, `foo bar and baz` yields `hasTag("foo bar") && hasTag("baz")`, right? I think this would be more readable if we forced people to use `'` for tags with whitespace, e.g. `'foo bar' and baz qux`. Thoughts?

Regarding the license: I don't think we should mix EPL and MIT licensed code within a single module. So, either we put it into a separate module (which would make it harder to use), or we write it from scratch (which we might need to do anyway, depending on the changes we want from the query language).

signed:
> Regarding the syntax, foo bar and baz yields hasTag("foo bar") && hasTag("baz"), right?

Correct.

>`'foo bar' and baz qux`

How about using `( )` like `(foo bar) and baz qux` and leave it to personal preference?

What about this expression `' foo bar '`. Is this the tag "foo bar" or " foo bar "? Can tags have leading/trailing white spaces?

marcphilipp:
> What about this expression `' foo bar '`. Is this the tag "foo bar" or " foo bar "? Can tags have leading/trailing white spaces?

Currently, the only restriction is that they may not be blank, i.e. only contain whitespace. So, yes, leading/trailing whitespace is currently allowed. However, I don't think that's a real requirement. We might as well trim them.

@junit-team/junit-lambda Thought?

sbrannen:
If we are not already trimming user-supplied tags, I would personally consider that a **bug**.

marcphilipp:
Let's create a separate issue for that and fix it right away. Any takers?

sbrannen:
> Regarding the syntax, `foo bar and baz` yields `hasTag("foo bar") && hasTag("baz")`, right? I think this would be more readable if we forced people to use `'` for tags with whitespace, e.g. `'foo bar' and baz qux`. Thoughts?

Agreed: tags containing multiple tokens separated by whitespace must be wrapped.

> Regarding the license: I don't think we should mix EPL and MIT licensed code within a single module. 

Agreed.

> So, either we put it into a separate module (which would make it harder to use), or we write it from scratch (which we might need to do anyway, depending on the changes we want from the query language).

Agreed.

sbrannen:
> Let's create a separate issue for that and fix it right away. Any takers?

I'll create the issue (and potentially tackle it as well).

marcphilipp:
I'm closing this PR as it will be superseded by a new one by @signed pretty soon. 🙂

PR #929, opened 2017-07-09
Allow setSource to be called with null

stefanbirkner:
## Overview

The field source already can be `null`. This is the case as long as `setSource` is not called. Existing code always checked the provided source and didn't call `setSource` if it was `null`. This check is not needed anymore.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/AbstractTestDescriptor.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/AbstractTestDescriptor.java
@@ -126,7 +126,7 @@ public abstract class AbstractTestDescriptor implements TestDescriptor {
 	}
 
 	protected final void setSource(TestSource source) {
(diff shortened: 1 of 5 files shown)


sbrannen:
> This check is not needed anymore.

Hmmmm.... I'm not sure about that. I'd have to look at the code in greater detail, but if I recall correctly: the current code prevents a previously non-null value from being overridden with null.

Do your changes exhibit the same behavior?

stefanbirkner:
It depends. `setSource` is always called in constructors. This means for code in this repository it exhibits the same behaviour but external code may violate that constraint.

I created an alternative pull request #936. In this pull request `setSource` is replaced by an additional constructor argument. I favour the new pull request over this one.

sbrannen:
> but external code may violate that "constraint".

Yep, that was my hunch.

sbrannen:
> I created an alternative pull request #936. 

OK. We'll check it out.

marcphilipp:
Closing in favor of #936.

PR #931, opened 2017-07-10
Simplify `Map.entrySet().forEach()` usage to `Map.forEach()`

jbduncan:
## Overview

This PR simplifies a usage of `Map.entrySet().forEach()` in `XmlReportWriter` to the equivalent `Map.forEach()`, which removes a number of unneeded characters.

---

I hereby agree to the terms of the JUnit Contributor License Agreement.

---

### Definition of Done

- [x] There are no TODOs left in the code
- [ ] Method [preconditions](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [x] [Coding conventions](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [ ] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#tests)
- [ ] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/meta/API.html)
- [ ] Change is documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [ ] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #932, opened 2017-07-10
Prevent subclassing of a utilities class

jbduncan:
## Overview

Although `ClasspathScanningSupport` is an `@API(Experimental)` utilities class, it can be subclassed, which doesn't seem to make sense to allow. This PR ensures that it cannot be subclassed.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/filter/ClasspathScanningSupport.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/filter/ClasspathScanningSupport.java
@@ -29,7 +29,13 @@ import org.junit.platform.engine.discovery.PackageNameFilter;
  * @since 1.0
  */
 @API(Experimental)
-public class ClasspathScanningSupport {
+public final class ClasspathScanningSupport {
+
+	///CLOVER:OFF
+	private ClasspathScanningSupport() {
+		/* no-op */
+	}
+	///CLOVER:ON
 
 	/**
 	 * Build a {@link Predicate} for fully qualified class names to be used for


jbduncan:
Similarly to https://github.com/junit-team/junit5/pull/931, it's not clear to me Travis running against JDK 9 fails. It seems to be something JDK-9-specific, rather than related to this PR. But then again maybe coming back to this in the near future will allow me to figure out what's wrong. :)

Andrei94:
Hi @jbduncan the Travis JDK 9 seems to be a bit buggy. It keeps running out of memory. I also had this problem :disappointed: . The solution was to restart the server and then it passed :smile:

Andrei94:
Regarding the PR, are you sure it's a good idea to prevent the subclassing? It might hinderthe mocking and the it's use in tests.

sormuras:
> Hi @jbduncan the Travis JDK 9 seems to be a bit buggy. It keeps running out of memory. I also had this problem 😞 . The solution was to restart the server and then it passed 😄

About to fix that issue (hopefully) with https://github.com/junit-team/junit5/pull/933

sbrannen:
> Regarding the PR, are you sure it's a good idea to prevent the subclassing? It might hinderthe mocking and the it's use in tests.

I don't think that's a concern here. We actually always aim to make our "static utility method" classes final.

jbduncan:
@Andrei94

> Regarding the PR, are you sure it's a good idea to prevent the subclassing? It might hinderthe mocking and the it's use in tests.

I hadn't actually considered mocking, but I made a half-educated, half-biased guess (based on my personal experience with Guava and understanding of Effective Java 2nd edition) that any potential uses of subclassing a static-utilities-methods class we so rare or limited that it was worth suggesting making `ClassScanningSupport` final.

jbduncan:
Thanks for pulling this in @sbrannen! :+1:

Andrei94:
I was thinking that it might add a constraint to the class that we now think it's helpful, but should someone would need to mock the class, there might be problems.

sbrannen:
@Andrei94, I understand where you're coming from; however, none of JUnit 5's utility classes are meant to be extended (or easily mocked via extension). 

Thus, this PR addressed an inconsistency within the code base. All other such classes are `final`, and now this one is as well.

Andrei94:
Oh, okay. Thanks @sbrannen and good job @jbduncan

PR #941, opened 2017-07-12
Execute test methods in a stable order

### LINKED ISSUE #934: Execute test methods in a stable order
## Overview

Related issue #13 

## Deliverables

- [ ] Internally "sort" (stable but not predictable) test methods before executing them

sormuras:
## Overview

Prior to this commit the execution order of reflected test methods was not well-defined. With this commit, the execution order is stabilized by using same default order as JUnit 4 applies.

Note copied from org/junit/internal/MethodSorter.java

>   Using the JVM order is unwise since the Java platform does not
>   specify any particular order, and in fact JDK 7 returns a more or less
>   random order; well-written test code would not assume any order, but
>   some does, and a predictable failure is better than a random failure
>   on certain platforms. By default, uses an unspecified but
>   deterministic order.

Closes #934 

This PR does **not** provide a way to customize the method sort order. This will be addressed by #13 

---

---

### Definition of Done


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -840,6 +842,22 @@ private static void findNestedClasses(Class<?> clazz, Set<Class<?>> candidates)
...
 
 	/**
+	 * Method comparator based upon JUnit4 org.junit.internal.MethodSorter implementation.
+	 */
+	private static int defaultMethodSorter(Method m1, Method m2) {
+		int i1 = m1.getName().hashCode();
+		int i2 = m2.getName().hashCode();
+		if (i1 != i2) {
+			return i1 < i2 ? -1 : 1;
+		}
```
marcphilipp: Since we don't have to stay backwards compatible with Java 5 (unlike Vintage), we can use `Integer.compare`.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -826,6 +826,8 @@ private static void findNestedClasses(Class<?> clazz, Set<Class<?>> candidates)
 				.collect(toList());
 		// @formatter:on
 
+		localMethods.sort(ReflectionUtils::defaultMethodSorter);
+
```
marcphilipp: We could move this up to where `localMethods` is computed an use the Stream API: `.sorted(ReflectionUtils::defaultMethodSorter).collect(...)`.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
sormuras: Sure.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
sormuras: With respect to #13 I chose a more visible location. And that location offers the possibility to insert something like `Collections.shuffle(localMethods);` more easily.

marcphilipp: Fair enough.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
sormuras: Just replace `return i1 < i2 ? -1 : 1;` with `return Integer.compare(i1, i2);`?

Or/and drop the entire duplicate hash value handling?

marcphilipp: How about the following?

```java
int i1 = m1.getName().hashCode();
int i2 = m2.getName().hashCode();
int comparison = Integer.compare(i1, i2);
if (comparison == 0) {
    comparison = m1.getName().compareTo(m2.getName());
    if (comparison == 0) {
    	comparison = m1.toString().compareTo(m2.toString());
    }
}
return comparison;
```

sormuras: Perfect.

sormuras:
Add short-cut when the same method instance is passed twice:

    if (method1 == method2) {
	    return 0;
    }

sbrannen:
> Added short-cut when the same method instance is passed twice:

How could that ever happen?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -840,6 +842,22 @@ private static void findNestedClasses(Class<?> clazz, Set<Class<?>> candidates)
...
+		int i1 = m1.getName().hashCode();
+		int i2 = m2.getName().hashCode();
+		if (i1 != i2) {
+			return i1 < i2 ? -1 : 1;
+		}
+		int comparison = m1.getName().compareTo(m2.getName());
+		if (comparison != 0) {
+			return comparison;
+		}
+		return m1.toString().compareTo(m2.toString());
```
sormuras: Could compute slightly slower than just "toString()" ... but also lead to more stable results. Not completely sure if that assumption holds.

   Set<?> a() { ... }
   Set<String> a(byte b) { ... }

Does that compile?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -840,6 +842,25 @@ private static void findNestedClasses(Class<?> clazz, Set<Class<?>> candidates)
...
+		String name2 = method2.getName();
+		int comparison = Integer.compare(name1.hashCode(), name2.hashCode());
+		if (comparison == 0) {
+			comparison = name1.compareTo(name2);
+			if (comparison == 0) {
+				comparison = method1.toString().compareTo(method2.toString());
+			}
+		}
+		return comparison;
+	}
```
jbduncan: Could we even go as far as using the default methods on `Comparator` directly in some way here?

sormuras:
>> Added short-cut when the same method instance is passed twice:
>
> How could that ever happen?

Mh, true. It should not happen with result of `clazz.getDeclaredMethods()`. A general-purpose comparator could use that short-cut when applied to a collection that permits duplicates.

Will remove it.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
sormuras: I don't see any that would fit here. Do you?

jbduncan: First idea that comes to mind:

```java
return Comparator.comparingInt(method -> method.getName().hashCode())
    .comparing(Method::getName)
    .comparing(Method::toString)
    .compare(method1, method2);
```

This potential implementation assumes that `Method::getName` and `Method::toString` have O(1) time complexity. Maybe not a reasonable assumption here?

sormuras: Does not compile as-is here: `method -> method.getName().hashCode() // getName() can not be resolved` and many "static member accessed via instance reference" warnings in IDEA...

jbduncan: Oh whoops, I hadn't tested it in my IDE!

How about this; does this compile for you?

```java
return Comparator.comparingInt((Method method) -> method.getName().hashCode())
        .thenComparing(Method::getName)
        .thenComparing(Method::toString)
        .compare(method1, method2);
```

jbduncan: I think this reads a little better.
```java
ToIntFunction<Method> toNameHashCode = method -> method.getName().hashCode();
return Comparator.comparingInt(toNameHashCode)
    .thenComparing(Method::getName)
    .thenComparing(Method::toString)
    .compare(method1, method2);
```

sormuras: Here you can compare both implementations side-by-side:
https://github.com/junit-team/junit5/compare/defaultmethodsorter_uses_stream_api

Thoughts? Pros, cons? Performance impact? Just a matter of taste? ;)

jbduncan: @sormuras Thanks!

I would make an educated guess that the original implementation would be faster than the Stream-like implementation I suggested. How faster? I don't really know.

IMO the Stream-like implementation is easier to read, since it's easier to see which aspects of the methods are chosen for sorting, and in which order those aspects apply. So if performance is no big concern for this method, I'd go for this implementation.

But how do we know if performance is a big problem here or not? Well, it would only be a big performance problem if it was a _bottleneck_ for JUnit 5, which would require profiling to find out. So unless we're willing to profile things now, again I'd go for the Stream-like implementation for clarity.

sbrannen: Well, the Stream-like implementation results in 4 lambda expressions and 2 method references; whereas, the original implementation uses simple nested if-statements.

So I can only assume that the original implementation performs better (even if only marginally).

Consequently, since looking up methods is an often used, core, low-level feature of the JUnit Platform, I would vote for keeping the original implementation _as is_.

The readability or clarity of such an implementation does not play a big role here, especially since typically only core committers would ever even look at that private utility method.

jbduncan: @sbrannen Good point! I'm now convinced to stay with the original implementation.

Issue #835, opened 2017-05-04
Forbid return values from test and lifecycle methods


sbrannen:
## Status Quo

In the JUnit Jupiter programming model it is currently not forbidden for `@Test` and lifecycle methods (i.e., `@BeforeEach`, `@BeforeAll`, etc.) to have a return value, even if it's _assumed_ that no one declares anything other than `void` as a return type for such methods.

For example, the following currently execute.

```java
@Test
int test() {
	fail("Boom!");
	return 42;
}
```

```java
@BeforeEach
int before() {
	fail("Boom!");
	return 42;
}
```

## Related Issues

- #242 

## Deliverables

- [ ] Forbid return values from test and lifecycle methods via test predicates that exclude such methods.

marcphilipp:
**Team decision:** Forbid via test predicates that exclude such methods.

mmerdes:
in progress

Issue #242, opened 2016-05-06
Introduce error handling mechanism for validation and discovery errors


marcphilipp:
## Overview

Validation errors (e.g., for invalid `@BeforeEach` method declarations) should not abort the entire discovery phase. Instead the _discovery phase_ should continue, with the error tracked and reported during the _execution phase_. 

## Areas of Applicability

- [x] exceptions thrown by test engines, as discussed in #750
- [x] unresolvable unique IDs, as discussed in #210
  - note, however, that a TestEngine should not report a unique ID as unresolvable if the unique ID does not apply to the particular TestEngine (see #1026 ).
- [x] invalid `@BeforeAll`, `@AfterAll`, `@BeforeEach`, and `@AfterEach` method declarations
- [x] [Competing method-level annotations](https://github.com/marcphilipp/junit5/blob/202d939953a2fa61f54a1b0c70e74e122740a427/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodSelectorResolver.java#L99-L108) (e.g. `@Test` _and_ `@TestTemplate` on the same method)
- [x] invalid `@Test`, `@TestFactory`, `@RepeatedTest`, and `@ParameterizedTest` method declarations (see #2244)
- [x] invalid `@Suite` class declarations
- [x] invalid `@Nested` test class declarations (see #1223 and #2717)
- [x] non-static test class declaration (see #2311)
- [x] missing `@Nested` annotation (see #1736)
- [x] _blank_ display names supplied by the user, as discussed in #743
- [x] `@Order` annotations on test classes and methods when class/method orderer is not `OrderAnnotation`

## Proposals

1. Allow engines to track errors by creating a special type of `TestDescriptor` such as an `AlwaysFailingTestDescriptor`, `DeadOnArrivalTestDescriptor`, or `ErrorReportingTestDescriptor`.
    - Tracked errors for such a corresponding `TestDescriptor` could then be thrown as an exception during the execution phase instead of executing the corresponding container or test.
2. Introduce a new property in `TestDescriptor` that signals an error that was encountered during the _discovery phase_.
3. Pass a _reporting object_ from the `Launcher` into each `TestEngine` to report errors.
    - See https://github.com/junit-team/junit5/issues/242#issuecomment-294301875

## Related Issues

- #121 
- #210 
- #743 
- #750 
- #835
- #876  
- #971 
- #1026 
- #1074 
- #1223
- #1944
- #2244
- #2311
- #2717
- #4125

[...]

PR #943, opened 2017-07-13
Issues/835 forbid return values from test and lifecycle methods

mmerdes:
## Overview

This PR implements #835.

Following the existing styles in the respective parts of the codebase this is realised via predicates for `@Test` methods, but via `JUnitException`s for lifecycle methods. This might be subject to change in the context of the somewhat related issue #242.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterAll.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterAll.java
@@ -36,6 +36,7 @@ import org.junit.platform.commons.meta.API;
  * {@link TestInstance @TestInstance(Lifecycle.PER_CLASS)}. {@code @AfterAll}
  * methods may optionally declare parameters to be resolved by
  * {@link org.junit.jupiter.api.extension.ParameterResolver ParameterResolvers}.
+ * {@code @AfterAll} methods must not return a value.
  *
  * <h3>Inheritance</h3>
  *
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterAll.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterAll.java
@@ -29,14 +29,14 @@ import org.junit.platform.commons.meta.API;
  *
  * <h3>Method Signatures</h3>
  *
- * <p>{@code @AfterAll} methods must not be {@code private} and must be
- * {@code static} by default. Consequently, {@code @AfterAll} methods are not
+ * {@code @AfterAll} methods must have a {@code void} return type,
+ * must not be {@code private}, and must be {@code static} by default.
+ * Consequently, {@code @AfterAll} methods are not
  * supported in {@link Nested @Nested} test classes or as <em>interface default
  * methods</em> unless the test class is annotated with
  * {@link TestInstance @TestInstance(Lifecycle.PER_CLASS)}. {@code @AfterAll}
  * methods may optionally declare parameters to be resolved by
  * {@link org.junit.jupiter.api.extension.ParameterResolver ParameterResolvers}.
- * {@code @AfterAll} methods must not return a value.
  *
  * <h3>Inheritance</h3>
  *
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/predicates/IsTestableMethod.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/predicates/IsTestableMethod.java
@@ -26,12 +26,14 @@ import org.junit.platform.commons.meta.API;
  * @since 5.0
  */
 @API(Internal)
-class IsTestableMethod implements Predicate<Method> {
+abstract class IsTestableMethod implements Predicate<Method> {
 
 	private final Class<? extends Annotation> annotationType;
+	private final boolean mustReturnVoid;
 
-	IsTestableMethod(Class<? extends Annotation> annotationType) {
+	IsTestableMethod(Class<? extends Annotation> annotationType, boolean mustReturnVoid) {
 		this.annotationType = annotationType;
+		this.mustReturnVoid = mustReturnVoid;
 	}
 
 	@Override
@@ -46,7 +48,16 @@ class IsTestableMethod implements Predicate<Method> {
 		if (isAbstract(candidate)) {
 			return false;
 		}
+		if ((this.returnsVoid(candidate) != mustReturnVoid)) {
+			return false;
+		}
+
 		return isAnnotated(candidate, this.annotationType);
+	}
+
+	private boolean returnsVoid(Method method) {
+		//checking the class itself yields problems with bridge methods
+		return method.getReturnType().toString().equals("void");
 	}
 
 }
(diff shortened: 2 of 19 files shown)


junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/LifecycleMethodUtilsTest.java
```diff
@@ -0,0 +1,223 @@
...
+		assertTrue(contains(methods, "eight"));
+	}
+
+	private boolean contains(List<Method> methods, String methodName) {
+		// @formatter:off
+        return methods.stream()
+                .map(Method::getName)
+                .anyMatch(name -> name.equals(methodName));
+        // @formatter:on
+	}
```
jbduncan: @mmerdes I think the formatting is a bit off here. :)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/LifecycleMethodUtilsTest.java
```diff
@@ -0,0 +1,223 @@
...
+		assertEquals(2, methods.size());
+		assertTrue(contains(methods, "seven"));
+		assertTrue(contains(methods, "eight"));
+	}
+
+	private boolean contains(List<Method> methods, String methodName) {
+		// @formatter:off
+        return methods.stream()
+                .map(Method::getName)
+                .anyMatch(name -> name.equals(methodName));
```
jbduncan: @mmerdes I think it's possible to shorten this to `.anyMatch(methodName::equals)` if `methodName` is never null.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -244,6 +244,11 @@ public static boolean isStatic(Member member) {
 		return Modifier.isStatic(member.getModifiers());
 	}
 
+	public static boolean returnsVoid(Method method) {
+		//checking the class itself yields problems with bridge methods
+		return method.getReturnType().toString().equals("void");
```
marcphilipp: `method.getReturnType().equals(void.class)`?

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/LifecycleMethodUtilsTest.java
mmerdes: @jbduncan 
Thanks for the hint. Looks slightly better indeed.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/descriptor/LifecycleMethodUtilsTest.java
mmerdes: @jbduncan Good call! It was a tabs vs spaces issue which got obscured by turning off the formatter...

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
mmerdes: Works like a charm. Opted for the `Void.TYPE` variant which I find slightly less ugly.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterAll.java
```diff
@@ -36,6 +36,7 @@
  * {@link TestInstance @TestInstance(Lifecycle.PER_CLASS)}. {@code @AfterAll}
  * methods may optionally declare parameters to be resolved by
  * {@link org.junit.jupiter.api.extension.ParameterResolver ParameterResolvers}.
+ * {@code @AfterAll} methods must not return a value.
```
sbrannen: I would move this new requirement to the first sentence and reword it something along the lines of:

```
{@code @AfterAll} methods must have a {@code void} return type, must not be {@code private}, and must be {@code static} by default.
```

And I would do the same for the rest of the related JavaDoc updates.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/predicates/IsTestableMethod.java
```diff
@@ -46,6 +49,10 @@ public boolean test(Method candidate) {
 		if (isAbstract(candidate)) {
 			return false;
 		}
+		if (returnsVoid(candidate) != mustReturnVoid) {
```
sbrannen: I'd recommend:

```java
if (mustReturnVoid && !returnsVoid(candidate))
```

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/NonVoidTestableMethodIntegrationTests.java
```diff
@@ -0,0 +1,45 @@
...
+class NonVoidTestableMethodIntegrationTests {
+
+	@Test
+	void valid() {
+	}
+
+	//this method must never be called
+	@Test
+	int invalidMethodReturningPrimitive() {
+		assertTrue(false);
```
sbrannen: Instead of `assertTrue(false)`, I'd recommend invoking `fail()` with an appropriate failure message, and I'd do that for the rest of this test class.

mmerdes: Good call; saves the comments as well.

sbrannen: exactly 😉

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/predicates/IsTestableMethod.java
sbrannen: BTW, this requested change is not only about readability but more importantly about performance... since the suggested construct benefits from boolean short-circuiting.

mmerdes: True, that's more readable.

mmerdes: I guess the performance improvements would be hard to measure but the readability gain is substantial :)

sbrannen: true! 😜

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterAll.java
mmerdes: ok

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/predicates/IsTestableMethod.java
mmerdes: mmh, the more readable version creates a regression here:
`org.junit.jupiter.engine.discovery.predicates.IsTestFactoryMethodTests#factoryMethodReturningVoidEvaluatesToFalse`
which seems to be a valid test case to me...

mmerdes: In essence, the new behavior creates false-positives when we do have a void return type but need a collection or stream return type for factory methods. So will revert to the original behavior.

sbrannen: hmmmm..... well, the "more readable version" seemed _logical_ to me, but if it's wrong... then it's wrong. 

Maybe I need to retake a basic course in boolean logic. 😉

PR #945, opened 2017-07-14
Make UniqueId segment list immutable

sormuras:
## Overview

Make UniqueId segment list immutable and turn `getSegments()` into a simple getter, i.e. it does no longer create a copy of the internal list on each invocation. The single place where a modifiable list was expected was in `JavaElementsResolver`. There a copy is created explicitly.

This commit also removes the defensive copy creation of the segment list from the package-private constructor. All two internal callers do not modify the underlying list, they actually drop their list reference.

---

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
@@ -10,6 +10,8 @@
 
 package org.junit.platform.engine;
 
+import static java.util.Collections.singletonList;
+import static java.util.Collections.unmodifiableList;
 import static org.junit.platform.commons.meta.API.Usage.Experimental;
 
 import java.io.Serializable;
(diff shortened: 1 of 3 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
```diff
@@ -81,16 +83,23 @@ public static UniqueId root(String segmentType, String value) {
...
+	private final List<Segment> segments;
 
 	private UniqueId(UniqueIdFormat uniqueIdFormat, Segment segment) {
 		this.uniqueIdFormat = uniqueIdFormat;
-		this.segments.add(segment);
+		this.segments = singletonList(segment);
 	}
 
+	/**
+	 * Initialize an {@code UniqueId} instance.
```
sbrannen: "a unique thing", not "an unique thing". 😉

junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
```diff
@@ -81,16 +83,23 @@ public static UniqueId root(String segmentType, String value) {
...
+	 *
+	 * @implNote A defensive copy of the segment list is <b>not</b> created by
+	 * this implementation. All callers should drop the list instance immediately
+	 * they pass into this constructor.
```
sbrannen: --> All callers should immediately drop the reference to the list instance that they pass into this constructor.

sbrannen:
I made some recommendations for two minor changes in wording, but it otherwise LGTM.

PR #962, opened 2017-07-18
Add Java 9 compatibility documentation

sormuras:
## Overview

Add Java 9 compatibility documentation

---

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes-5.0.0-M6.adoc
+++ b/documentation/src/docs/asciidoc/release-notes-5.0.0-M6.adoc
@@ -14,6 +14,37 @@ For a complete list of all _closed_ issues and pull requests for this release, c
 link:{junit5-repo}+/milestone/11?closed=1+[5.0 M6] milestone page in the JUnit repository
 on GitHub.
 
+==== Java 9 compatibility
+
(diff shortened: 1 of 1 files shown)


documentation/src/docs/asciidoc/release-notes-5.0.0-M6.adoc
```diff
@@ -14,6 +14,35 @@ For a complete list of all _closed_ issues and pull requests for this release, c
...
+module foo.bar {
+  requires org.junit.jupiter.api;
+}
+```
+
+* Usually it's sufficient to run tests on the class-path. There's no change between
+  Java 8 and 9. All command line based tools and IDEs continue to work with JUnit 5
+  _out-of-the-box_ -- if they support the JUnit platform. If your tool of choice
+  does not support JUnit 5 yet, you may always resort to the `ConsoleLauncher` or
+  even to executable `junit-platform-console-standalone` all-in-one jar.
```
sormuras: Or "to the"?

jbduncan: Yes, I think that sounds just as good. :)

Issue #978, opened 2017-07-24
Overridden interface default methods are not considered overridden


sbrannen:
## Overview

This is a follow-up to #976.

Whether or not the bug described in _this_ issue existed at a previous point in time is currently unknown (simply because it has not been researched).

In any case, we now have a situation where an _overridden_ interface default method is not considered to be overridden by the `findMethods()` search algorithm in `ReflectionUtils`.

Consequently, if we execute all tests in the following `GenericTestCaseWithOverriddenDefaultMethod` JUnit Jupiter based test class, we see that 3 tests are executed instead of 2. Specifically, we expect that only `test(Long)` and `test(Double)` will be executed; however, `test(Number)` is additionally _discovered_ resulting in `test(Long)` being executed twice.

```java
@ExtendWith({ LongParameterResolver.class, DoubleParameterResolver.class })
interface GenericTestInterface<N extends Number> {

	@Test
	default void test(N number) {
	}
}

class GenericTestCaseWithOverriddenDefaultMethod implements GenericTestInterface<Long> {

	@Test
	@Override
	public void test(Long number) {
	}

	@Test
	void test(Double number) {
	}
}
```

## Analysis

The recent series of [Whac-A-Mole](https://en.wikipedia.org/wiki/Whac-A-Mole) bug fixes in this area leads to me believe that the only _real_ solution to this set of problems is one that takes into account the actual generic and parameterized type information instead of relying on a best-guess approach involving method sub-signatures and blind checks for the presence of generics.

Depending on how we _take into account the actual generic and parameterized type information_, we might potentially wish to reverse a subset of changes in fae5b6a4f001e7427fba13b13fb844074d5ff099 that explicitly treat default methods as locally declared methods (while leaving the other changes in that commit in tact).

## Further Considerations

This issue currently only focuses on overridden interface default methods; however, it may well be the case that our `findMethods()` search algorithm contains other bugs related to generics.

## Related Issues

- #333 
- #976 

## Deliverables

- [x] Ensure that overridden interface default methods are not returned in the results of `ReflectionUtils.findMethods()`.
- [x] Investigate possible additional bugs hinted at in _Further Considerations_.
  - to be addressed in #981

sbrannen:
Added _Further Considerations_ section.

jbduncan:
My understanding of the interplay between generics and reflection is a bit limited, but I would think that https://github.com/jhalterman/typetools or something similar would be helpful here.

sbrannen:
[...]

PR #979, opened 2017-07-24
Return only visible default methods

sormuras:
Return only visible default methods

Addresses #978

---


sbrannen:
Thanks! 👍

Issue #740, opened 2017-03-18
Remove deprecated Launcher.execute() method


sormuras:
## Overview

Addressing issue #628 with PR #739 marked the old single-parameter `Launcher.execute(LauncherDiscoveryRequest launcherDiscoveryRequest)` as deprecated.

Remove this method and revert all explicit `Launcher.execute(request, new TestExecutionListener[0])` invocations that were introduced by #739.

## Deliverables

- [ ] Revert https://github.com/junit-team/junit5/commit/cb6ca6823ac854165246caebd6dcafe2e085d981 before RC1

sormuras:
in progress...

sormuras:
Partly reverting this commit to make current IDE(A)s running local tests again.

This new code will be removed before GA!

sbrannen:
> This new code will be removed before GA!

Then we should remove the "in progress" label for now and assign this issue to RC1.

Or... are you doing work on it right now?

sormuras:
I am. ;-)

sormuras:
Revert https://github.com/junit-team/junit5/commit/cb6ca6823ac854165246caebd6dcafe2e085d981 before RC1

sormuras:
Updated deliverables.

PR #988, opened 2017-07-25
Fixed "NoSuchMethodError" when running tests from IntelliJ

Andrei94:
## Overview

Upon running any test from the `documentation` project or from `org.junit.vintage.engine` IntelliJ throws `NoSuchMethodError`. It seems that the problem is the missing `execute(LauncherDiscoveryRequest)` from `org.junit.platform.launcher`. Upon adding it, the tests run as expected.

---

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
@@ -89,4 +89,10 @@ public interface Launcher {
 	 */
 	void execute(LauncherDiscoveryRequest launcherDiscoveryRequest, TestExecutionListener... listeners);
 
+	// This method is temporarily available to restore binary compatibility.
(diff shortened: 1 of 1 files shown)


Andrei94:
@sormuras thoughts?

sormuras:
It was removed on purpose via https://github.com/junit-team/junit5/commit/b243753b785477172411cf989186876b1d8998e5 triggered by https://github.com/junit-team/junit5/issues/740

We don't want that method to prevail in a RC nor the GA version of JUnit 5. You may add it locally and move that change to a custom change list within IDEA until the IDE is updated to include a more recent version of JUnit 5.

Closing this issue as "works-as-designed".

Andrei94:
Okay. I saw it was removed in a previous commit. Since IntelliJ wasn't running the tests, I thought it was a bug. Thank you for the clarification! 😄

PR #995, opened 2017-07-30
Add Store.Key, to allow easier type-safe access to values in a store.

### LINKED ISSUE #361: Update Store to use type-safe keys.
Currently the APIs for updating and removing values from `Store` was not type-safe. In 55175604801ada @sbrannen added methods so that you can avoid casts when getting the values, but the programmer still has to ensure that all of the places were a key is accessed uses the same type.

Instead, we can make the keys generic. In other words, we would create a key class that encapsulates two things: a key for getting the value and the type for the values.

One proposal would be to have a `Key` interface that just looks like this:

```
public interface Key<V> {}
```

Keys could be constructed like this:

```
private static final Key<Long> KEY = Key.create();
```

Then used like this:

```
Long value = store.remove(KEY);
```

This works if you need a key that could be represented as a compile time constant. If you wanted the key to depend on a value type (for example, `Method`) we could provide a different API to create Keys:

```
private static final KeySpace<Method, Long> METHOD_START_TIMES = new KeySpace<>() {
  @Override protected Long initialValue(Method key) {
    return System.currentTimeMillis;
  }
};   
```

(named "KeySpace" because it is a namespace for keys)

A `KeySpace` would be used like this:

```
Key<Long> key = METHOD_START_TIMES.keyFor(method);
Long startTime = store.remove(key);
```

We could add methods to `Store` to make using a `KeySpace` even easier:

```
Long startTime = store.remove(METHOD_START_TIMES, method);
```

kcooney:
Fixes #361

## Overview

This provides a much-less error-prone way to access a Store in a type-safe way.

Open question: Can we remove the old methods, or is it too late for that?
---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes-5.0.0-RC1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes-5.0.0-RC1.adoc
@@ -64,7 +64,8 @@ on GitHub.
   class is annotated with `@TestInstance(Lifecycle.PER_CLASS)`.
 * Test classes written in the Kotlin programming language are now executed with
   `@TestInstance(Lifecycle.PER_CLASS)` semantics by default.
-
+* `ExtensionContext.Store` now supports variants with a type-safe key (see
+  `ExtensionContext.Store.Key`).
 
 [[release-notes-5.0.0-rc1-junit-vintage]]
 ==== JUnit Vintage
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -267,6 +267,21 @@ public interface ExtensionContext {
 	interface Store {
 
 		/**
+		 * Get the value that is stored under the supplied {@link Key}.
+		 *
+		 * <p>If no value is stored in the current {@link ExtensionContext}
+		 * for the supplied {@code key}, ancestors of the context will be queried
+		 * for a value with the same {@code key} in the {@code Namespace} used
+		 * to create this store.
+		 *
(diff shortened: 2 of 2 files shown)


marcphilipp:
How would one use this in an extension? Since `Key` relies on object identity, an extension now has to save it somewhere in order to access it at a later point in time. This effectively forces an extension to use constants for `Key` instances, right?

sbrannen:
If an extension is forced to save the key, it's a no-go!

A key must be computable, typically based on information from the current extension context. 

Something like a `KeyHolder` (that holds the key and the required type) would be required if we want to go this route.

kcooney:
didn't realize a key needs to be computable. There was no mention of that in the bug, and I proposed a similar solution. The JavaDoc for Store doesn't this use case either.

kcooney:
(woops! Didn't mean to close)

I'll have to give this more thought.

Could you answer my earlier question?  Can we deprecate or remove methods or is it too late for that?

kcooney:
Replacing with #996

sbrannen:
> didn't realize a key needs to be computable. 

That's always been a primary use case for the team, even if not explicitly stated somewhere.

In any case, several examples demonstrate the use of custom computed keys.

> There was no mention of that in the bug, and I proposed a similar solution. 

What "bug"?

> The JavaDoc for Store doesn't this use case either.

I suppose the JavaDoc could make that more explicit. Care to raise an issue for that?

kcooney:
The issue was #361  (see the "Fixes` line).

I don't personally find it useful to file bugs for improving a JavaDoc for a single class. If someone feels strongly enough to fix the JavaDoc for a class, IMHO they should just fix it. If you have specific wording in mind, I would be happy to do include it in my pull.

PR #996, opened 2017-07-30
Introduce StoredValues

### LINKED ISSUE #361: Update Store to use type-safe keys.

kcooney:
## Overview

This provides more type-safe access to values in a `Store`.

Fixes #361

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -275,7 +275,7 @@ public interface ExtensionContext {
 		 * to create this store.
 		 *
 		 * <p>For greater type safety, consider using {@link #get(Object, Class)}
-		 * instead.
+		 * or {@link StoredValues#get(Store, Object) instead.
 		 *
 		 * @param key the key; never {@code null}
 		 * @return the value; potentially {@code null}
@@ -291,6 +291,9 @@ public interface ExtensionContext {
 		 * for the supplied {@code key}, ancestors of the context will be queried
 		 * for a value with the same {@code key} in the {@code Namespace} used
 		 * to create this store.
+		 *
+		 * <p>For greater type safety, consider using
+		 * {@link StoredValues#get(Store, Object) instead.
 		 *
 		 * @param key the key; never {@code null}
 		 * @param requiredType the required type of the value; never {@code null}
@@ -311,7 +314,9 @@ public interface ExtensionContext {
 		 * the {@code key} as input), stored, and returned.
 		 *
 		 * <p>For greater type safety, consider using
-		 * {@link #getOrComputeIfAbsent(Object, Function, Class)} instead.
+		 * {@link #getOrComputeIfAbsent(Object, Function, Class)} or
+		 * {@link StoredValues#getOrComputeIfAbsent(Store, Object, Function)}
+		 * instead.
 		 *
 		 * @param key the key; never {@code null}
 		 * @param defaultCreator the function called with the supplied {@code key}
@@ -334,6 +339,10 @@ public interface ExtensionContext {
 		 * a new value will be computed by the {@code defaultCreator} (given
 		 * the {@code key} as input), stored, and returned.
 		 *
+		 * <p>For greater type safety, consider using
+		 * {@link StoredValues#getOrComputeIfAbsent(Store, Object, Function)}
+		 * instead.
+		 *
 		 * @param key the key; never {@code null}
 		 * @param defaultCreator the function called with the supplied {@code key}
 		 * to create a new value; never {@code null}
@@ -352,6 +361,9 @@ public interface ExtensionContext {
 		 * ExtensionContexts} for the store's {@code Namespace} unless they
 		 * overwrite it.
 		 *
+		 * <p>For greater type safety, consider using
+		 * {@link StoredValues#put(Store, Object, Object) instead.
+		 *
 		 * @param key the key under which the value should be stored; never
 		 * {@code null}
 		 * @param value the value to store; may be {@code null}
@@ -365,7 +377,7 @@ public interface ExtensionContext {
 		 * not in ancestors.
 		 *
 		 * <p>For greater type safety, consider using {@link #remove(Object, Class)}
-		 * instead.
+		 * or {@link StoredValues#remove(Store, Object)} instead.
 		 *
 		 * @param key the key; never {@code null}
 		 * @return the previous value or {@code null} if no value was present
@@ -381,7 +393,10 @@ public interface ExtensionContext {
 		 * <p>The value will only be removed in the current {@link ExtensionContext},
 		 * not in ancestors.
 		 *
-		 * @param key the key; never {@code null}
+		 * <p>For greater type safety, consider using
+		 * {@link StoredValues#remove(Store, Object)} instead.
+		 *
+		 * @@param key the key; never {@code null}
 		 * @param requiredType the required type of the value; never {@code null}
 		 * @param <V> the value type
 		 * @return the previous value or {@code null} if no value was present
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -280,7 +280,7 @@ public interface ExtensionContext {
 		 * to create this store.
 		 *
 		 * <p>For greater type safety, consider using {@link #get(Object, Class)}
-		 * or {@link StoredValues#get(Store, Object) instead.
+		 * or {@link StoredValues#get(Store, Object)} instead.
 		 *
 		 * @param key the key; never {@code null}
 		 * @return the value; potentially {@code null}
@@ -298,7 +298,7 @@ public interface ExtensionContext {
 		 * to create this store.
 		 *
 		 * <p>For greater type safety, consider using
-		 * {@link StoredValues#get(Store, Object) instead.
+		 * {@link StoredValues#get(Store, Object)} instead.
 		 *
 		 * @param key the key; never {@code null}
 		 * @param requiredType the required type of the value; never {@code null}
@@ -368,7 +368,7 @@ public interface ExtensionContext {
 		 * overwrite it.
 		 *
 		 * <p>For greater type safety, consider using
-		 * {@link StoredValues#put(Store, Object, Object) instead.
+		 * {@link StoredValues#put(Store, Object, Object)} instead.
 		 *
 		 * @param key the key under which the value should be stored; never
 		 * {@code null}
@@ -402,7 +402,7 @@ public interface ExtensionContext {
 		 * <p>For greater type safety, consider using
 		 * {@link StoredValues#remove(Store, Object)} instead.
 		 *
-		 * @@param key the key; never {@code null}
+		 * @param key the key; never {@code null}
 		 * @param requiredType the required type of the value; never {@code null}
 		 * @param <V> the value type
 		 * @return the previous value or {@code null} if no value was present
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -262,12 +262,17 @@ public interface ExtensionContext {
 	Store getStore(Namespace namespace);
 
 	/**
-	 * {@code Store} provides methods for extensions to save and retrieve data.
+	 * Provides methods for extensions to save and retrieve data.
+	 *
+	 * <p>For better isolation and type-safety, consider accessing the
+	 * {@code Store} via a {@link StoredValues} instance.
+	 *
+	 * @see StoredValues
 	 */
 	interface Store {
 
 		/**
-		 * Get the value that is stored under the supplied {@code key}.
+		 * Returns the value that is stored under the supplied {@code key}.
 		 *
 		 * <p>If no value is stored in the current {@link ExtensionContext}
 		 * for the supplied {@code key}, ancestors of the context will be queried
@@ -284,7 +289,7 @@ public interface ExtensionContext {
 		Object get(Object key);
 
 		/**
-		 * Get the value of the specified required type that is stored under
+		 * Returns the value of the specified required type that is stored under
 		 * the supplied {@code key}.
 		 *
 		 * <p>If no value is stored in the current {@link ExtensionContext}
@@ -304,7 +309,8 @@ public interface ExtensionContext {
 		<V> V get(Object key, Class<V> requiredType);
 
 		/**
-		 * Get the value that is stored under the supplied {@code key}.
+		 * Returns the value that is stored under the supplied {@code key},
+		 * computing a value if there is no stored value.
 		 *
 		 * <p>If no value is stored in the current {@link ExtensionContext}
 		 * for the supplied {@code key}, ancestors of the context will be queried
@@ -329,8 +335,8 @@ public interface ExtensionContext {
 		<K, V> Object getOrComputeIfAbsent(K key, Function<K, V> defaultCreator);
 
 		/**
-		 * Get the value of the specified required type that is stored under the
-		 * supplied {@code key}.
+		 * Returns the value of the provided type that is stored under the
+		 * supplied {@code key}, computing a value if there is no stored value.
 		 *
 		 * <p>If no value is stored in the current {@link ExtensionContext}
 		 * for the supplied {@code key}, ancestors of the context will be queried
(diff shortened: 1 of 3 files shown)


marcphilipp:
> Can we deprecate or remove methods or is it too late for that?

To answer your question from #995: I'm going ro release RC1 today which – according to the current plan – is the last chance to make breaking changes. If we see fit, we could add additional release candidates to the plan, though.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -262,20 +262,25 @@ default Store getStore() {
...
+	 * <p>For better isolation and type-safety, consider accessing the
+	 * {@code Store} via a {@link StoredValues} instance.
+	 *
+	 * @see StoredValues
 	 */
 	interface Store {
 
 		/**
-		 * Get the value that is stored under the supplied {@code key}.
+		 * Returns the value that is stored under the supplied {@code key}.
```
marcphilipp: We always use "Get" instead if "Gets" etc. because we want to describe the action carried out by the method not the method itself.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/StoredValues.java
```diff
@@ -0,0 +1,153 @@
...
+ *   private static final StoredValues<Method, Instant> START_TIMES
+ *       = new StoredValues<>();
+ * 
+ *   @Override
+ *   public void beforeTestExecution(ExtensionContext context) {
+ *     START_TIMES.put(
+ *         getStore(context), context.getRequiredTestMethod(), Instant.now());
+ *   }
+ * }
+ * }</pre>
```
marcphilipp: To be honest, I'm not sure this is a big improvement over the current API. Need to ponder upon it for awhile. 🤔

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -262,20 +262,25 @@ default Store getStore() {
 	Store getStore(Namespace namespace);
 
 	/**
-	 * {@code Store} provides methods for extensions to save and retrieve data.
+	 * Provides methods for extensions to save and retrieve data.
```
sbrannen: Please don't change our Javadoc style.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
sbrannen: Ditto.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -284,14 +289,17 @@ default Store getStore() {
 		Object get(Object key);
 
 		/**
-		 * Get the value of the specified required type that is stored under
+		 * Returns the value of the specified required type that is stored under
```
sbrannen: Ditto

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -301,7 +309,8 @@ default Store getStore() {
 		<V> V get(Object key, Class<V> requiredType);
 
 		/**
-		 * Get the value that is stored under the supplied {@code key}.
+		 * Returns the value that is stored under the supplied {@code key},
```
sbrannen: ditto

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -324,8 +335,8 @@ default Store getStore() {
 		<K, V> Object getOrComputeIfAbsent(K key, Function<K, V> defaultCreator);
 
 		/**
-		 * Get the value of the specified required type that is stored under the
-		 * supplied {@code key}.
+		 * Returns the value of the provided type that is stored under the
```
sbrannen: ditto

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -346,26 +361,29 @@ default Store getStore() {
 		<K, V> V getOrComputeIfAbsent(K key, Function<K, V> defaultCreator, Class<V> requiredType);
 
 		/**
-		 * Store a {@code value} for later retrieval under the supplied {@code key}.
+		 * Stores a {@code value} for later retrieval under the supplied {@code key}.
```
sbrannen: And.... "ditto" for all such changes.

I'll stop marking each one now. 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/StoredValues.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.Objects;
+import java.util.function.Function;
+
+import org.junit.jupiter.api.extension.ExtensionContext.Store;
+import org.junit.platform.commons.meta.API;
+
+/**
+ * Provides isolated, type-safe access to values in a {@link Store}.
+ *
+ * <p>Example use:
```
sbrannen: Please make this an H3 HTML header.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/StoredValues.java
sbrannen: I feel the same as @marcphilipp: I don't (yet) see how this is an improvement.

It appears to be just a _different_ set of hoops to jump through.

Plus, with this proposal, the static `StoredValues` is limited to storing only instances of `Instant`. Thus if the extension needed to have additional types of data stored between callbacks, the extension would then have to have multiple instances of `StoredValues`, one for each concrete type of thing to store. So that honestly seems like a step back to me.

Or am I missing something?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
kcooney: That contradicts the Sun/Oracle JavaDoc style, and the style for mist JavaDoc in the JDK. See, for example, the JavaDoc for `Map.get()`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/StoredValues.java
kcooney: With the current API, each time to you access a Store for a key, you have to specify the expected value, which is error-prone.

I see no problem with having multiple `StoredValues` values. It's like having multiple fields. Or a `ThreadLocal`.

Plus you are not required to use the new API.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
kcooney: Again, what was here contradicts the JavaDoc practices, so I was cleaning up while making other changes.

If you want to require a different style , I suggest you document it.

kcooney:
re: the Javadoc changes. From my job, I'm in the habit of doing minor improvements when I touch code ("always leave code in a better state"). While updating the JavaDoc to mention the new class, I updated it to match the Oracle style (see [this section](http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#styleguide), specifically "Method descriptions begin with a verb phrase") and the Javadoc of `Map`. If you strongly feel that the JUnit 5 Javadoc should not match these conventions, I can revert those changes, but I prefer to wait until we get agreement about whether I should make this change before taking the time to do that.

I had initially filed the bug (#361) and worked on these pulls because I thought that the type-safety protections of `Store` are too easy to use incorrectly (you have to specify the type each time you access the key, which is error-prone, particularly on `put()`). After the discussion on keys being "computable" I looked at the examples in the User Guide, and I'm now much more concerned about isolation

I looked at [the example](http://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks-before-after-execution) in the User Guide, and unless I'm misunderstanding something, the call to `Store.put()` in `beforeTestExecution()` would conflict with any other extension that stores a value using a  `Method`as a key. The code code, of course, use a `Namespace` for isolation, but that seems like a strange requirement to get encapsulation.

Even if the code used a `Namespace`, it could store the end time using the `Method` as a key without overwriting the start time (yes, this particular class doesn't need to, but if an extension wanted to record start and end times and write it to a file at the end of a test run, it would).

The class I add here solves both problems.

marcphilipp:
> I had initially filed the bug (#361) and worked on these pulls because I thought that the type-safety protections of `Store` are too easy to use incorrectly (you have to specify the type each time you access the key, which is error-prone, particularly on `put()`). After the discussion on keys being "computable" I looked at the examples in the User Guide, and I'm now much more concerned about isolation.

How does this proposal help with isolation? Storing the same key in the same namespace multiple times would still "conflict", wouldn't it?

> I looked at [the example](http://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks-before-after-execution) in the User Guide, and unless I'm misunderstanding something, the call to `Store.put()` in `beforeTestExecution()` would conflict with any other extension that stores a value using a  `Method`as a key. The code code, of course, use a `Namespace` for isolation, but that seems like a strange requirement to get encapsulation.

Isolation/encapsulation is the sole purpose of `Namespace`. The `Store` API is designed in this way to enable different extensions to access each other's properties, given they use the same namespace. There is the potential for accidentally doing that, but it's fairly minimal IMHO (cf. the Javadoc of `Namespace`).

> Even if the code used a `Namespace`, it could store the end time using the `Method` as a key without overwriting the start time (yes, this particular class doesn't need to, but if an extension wanted to record start and end times and write it to a file at the end of a test run, it would).

Such an extension could use different keys for method start/end, or store an object that contains start and end using the `Method` as a key.

marcphilipp:
> I had initially filed the bug (#361) and worked on these pulls because I thought that the type-safety protections of `Store` are too easy to use incorrectly (you have to specify the type each time you access the key, which is error-prone, particularly on `put()`). After the discussion on keys being "computable" I looked at the examples in the User Guide, and I'm now much more concerned about isolation.

How does this proposal help with isolation? Storing the same key in the same namespace multiple times would still "conflict", wouldn't it?

> I looked at [the example](http://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks-before-after-execution) in the User Guide, and unless I'm misunderstanding something, the call to `Store.put()` in `beforeTestExecution()` would conflict with any other extension that stores a value using a  `Method`as a key. The code code, of course, use a `Namespace` for isolation, but that seems like a strange requirement to get encapsulation.

Isolation/encapsulation is the sole purpose of `Namespace`. The `Store` API is designed in this way to enable different extensions to access each other's properties, given they use the same namespace. There is the potential for accidentally doing that, but it's fairly minimal IMHO (cf. the Javadoc of `Namespace`).

> Even if the code used a `Namespace`, it could store the end time using the `Method` as a key without overwriting the start time (yes, this particular class doesn't need to, but if an extension wanted to record start and end times and write it to a file at the end of a test run, it would).

Such an extension could use different keys for method start/end, or store an object that contains start and end using the `Method` as a key.

kcooney:
> How does this proposal help with isolation? 

This new class wraps the passed-in keys with a class that will not be equal with the key or any other code that wraps the key (other than the same instance of `StoredValues`).

After seeing your question, I realized that `StoredValues.KeyWrapper.equals()` had a bug. I'll add tests if we agree on the general approach in this pull.

> The Store API is designed in this way to enable different extensions to access each other's properties

While that is a great goal, having extensions being able to access values stored by other extensions is dangerous as a default behavior. If someone wrote an extension using [the code from the example](http://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks-before-after-execution) in the User Guide, it would work fine on it's own, but could fail if any other extension stores value using a `Method`. That behavior isn't clear from the `getStore()` method.

I think at the very least we should either rename `getStore()` to `getSharedStore()` or remove the method completely and force users to use `getStore(Namespace)` instead. If we do that, then developers would be much more likely to read the Javadoc for `Namespace` and understand the isolation issues with using the default namespace. If we delete `getStore()` then I recommend renaming `Namespace.DEFAULT` to `Namespace.GLOBAL`.

>> Even if the code used a Namespace, it could store the end time using the Method as a key without
>> overwriting the start time (yes, this particular class doesn't need to, but if an extension wanted to
>> record start and end times and write it to a file at the end of a test run, it would).

> Such an extension could use different keys for method start/end, or store an object that contains start
> and end using the Method as a key.

That means that an extension developer needs to be aware of every type that they use as a key whenever they decide to introduce code that Stores values a new key. Don't you think that's error-prone?

It also still means that if another extension stores any value using `Method` as a key, there will be a conflict (unless you use namespaces, but developers doing the simple thing will not likely use namespaces, and not realize they get conflicts).

kcooney:
If it wasn't so close to release, I would suggest changing the `Store` API to look like this:

    <K, V> V get(Key<?, V> key);
    <K, V> V getOrComputeIfAbsent(Key<K, V> key, Function<K, V> defaultCreator);
    <K, V> void put(Key<?, V> key, V value);
    <K, V> V remove(Key<?, V> key);

As suggested by @sbrannen in #995  "Something like a `KeyHolder` (that holds the key and the required type) would be required if we want to go this route."

So that might look like this:

    private static final KeyHolder<Method, Instant> START_TIMES = KeyHolder.create();

    // save start time for this method. is write is isolated from other extensions
    Instant startTime = Instant.now();
    store.put(START_TIMES.keyFor(method), instant);

    // save earliest start time so any extension can access it
    store.getOrComputeIfAbsent(
        Key.<Method,Instant>shared(), // admittedly ugly syntax
        (Method m) -> startTime);

This way, when you want to use shared keys, you are forced to make an explicit decision to do so when you make the method call to the store

marcphilipp:
> I think at the very least we should either rename `getStore()` to `getSharedStore()` or remove the method completely and force users to use `getStore(Namespace)` instead. If we do that, then developers would be much more likely to read the Javadoc for `Namespace` and understand the isolation issues with using the default namespace. If we delete `getStore()` then I recommend renaming `Namespace.DEFAULT` to `Namespace.GLOBAL`.

Again, I think `Namespace` is all we need for encapsulation. Making it more discoverable makes sense, though. So, I'd support removing `getStore()` and renaming `Namespace.DEFAULT` to `Namespace.GLOBAL`.

sbrannen:
> Again, I think Namespace is all we need for encapsulation. Making it more discoverable makes sense, though. So, I'd support removing getStore() and renaming Namespace.DEFAULT to Namespace.GLOBAL.

:+1:

sbrannen:
Regarding JavaDoc style, we have been consistent across the entire code base and therefore will not be changing every class (or individual classes) just before the GA release. 

Documenting _our style_, however, is something we should likely do in the contributors' guide.

sbrannen:
JavaDoc style will be better documented in conjunction with #1007.

sbrannen:
@kcooney, would you be willing to remove `getStore()` and rename `Namespace.DEFAULT` to `Namespace.GLOBAL` in a separate PR?

If not, please let us know so that we can address that ourselves.

sbrannen:
Currently slated for 5.0 RC3 in order to ensure that any necessary action items are taken care of before GA.

sormuras:
> ...  remove getStore() and rename Namespace.DEFAULT to Namespace.GLOBAL in a separate PR?

Addressed in #1013

marcphilipp:
@kcooney Thanks for bringing up the potential for collisions! The major concern (having to use t "global" explictly) has been addressed in #1013. I'm closing this issue because we're not convinced by the type-safe API proposal.

PR #998, opened 2017-07-30
Fix thread-safety issue in ExtensionValuesStore

kcooney:
## Overview

getOrComputeIfAbsent() was accessing the parent store's map without
acquiring the parent's lock.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
@@ -47,10 +47,8 @@ public class ExtensionValuesStore {
 	}
 
 	Object get(Namespace namespace, Object key) {
-		synchronized (this.monitor) {
-			StoredValue storedValue = getStoredValue(namespace, key);
-			return storedValue == null ? null : storedValue.value;
-		}
+		StoredValue storedValue = getStoredValue(namespace, key);
+		return storedValue == null ? null : storedValue.value;
 	}
 
 	<T> T get(Namespace namespace, Object key, Class<T> requiredType) {
@@ -96,15 +94,23 @@ public class ExtensionValuesStore {
 	}
 
 	private StoredValue getStoredValue(Namespace namespace, Object key) {
-		StoredValue storedValue = this.storedValues.get(new CompositeKey(namespace, key));
-		if (storedValue != null) {
-			return storedValue;
-		}
-		else if (this.parentStore != null) {
-			return this.parentStore.getStoredValue(namespace, key);
-		}
-		else {
-			return null;
+		/*
+		 * This method can be called from another ExtensionValuesStore instance,
+		 * so we must synchronized on the monitor.
+		 *
+		 * TODO: Consider rewriting with ConcurrentHashMap.
+		 */
+		synchronized (this.monitor) {
+			StoredValue storedValue = this.storedValues.get(new CompositeKey(namespace, key));
+			if (storedValue != null) {
+				return storedValue;
+			}
+			else if (this.parentStore != null) {
+				return this.parentStore.getStoredValue(namespace, key);
+			}
+			else {
+				return null;
+			}
 		}
 	}
 
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
@@ -49,15 +49,7 @@ public class ExtensionValuesStore {
 	Object get(Namespace namespace, Object key) {
 		synchronized (this.monitor) {
 			StoredValue storedValue = getStoredValue(namespace, key);
-			if (storedValue != null) {
-				return storedValue.value;
-			}
-			else if (this.parentStore != null) {
-				return this.parentStore.get(namespace, key);
-			}
-			else {
-				return null;
-			}
+			return storedValue == null ? null : storedValue.value;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


kcooney:
@jbduncan I actually would prefer ReentrantLock because it is more efficient, but I believe some others on the JUnit team didn't like the extra boilerplate of the `try` ... `finally`

Note there are two commits (my first attempt failed; oh well). Don't forget to squash them, or let me know if you want me to squash.

jbduncan:
I actually just remembered Guava's `Monitor` as well... shame we can't import Guava. :P

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
```diff
@@ -110,7 +110,16 @@ Object remove(Namespace namespace, Object key) {
 
 	private StoredValue getStoredValue(Namespace namespace, Object key) {
 		CompositeKey compositeKey = new CompositeKey(namespace, key);
-		return this.storedValues.get(compositeKey);
+
+		/*
+		 * This method can be called from another ExtensionValuesStore instance,
+		 * so we must synchronized on the monitor.
```
sbrannen: It _can_ be, but is there currently anywhere in the code base where that occurs from outside a synchronized block?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
```diff
@@ -110,7 +110,16 @@ Object remove(Namespace namespace, Object key) {
...
+		 *
+		 * TODO: Consider rewriting with ConcurrentHashMap.
```
sbrannen: We've already discussed this once before in a previous thread, and the conclusion was that a `ConcurrentHashMap` would not help since there are multiple maps involved within a hierarchy of stores.

Do you perceive this differently?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
sbrannen: "synchronized" --> "synchronize

kcooney: > It can be, but is there currently anywhere
> In the code 

Yes. From `getOrComputeIfAbsent()` where the code accesses the parent store's map without obtained the parent's lock.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
kcooney: I actually do. I found this bug while rewriting the class to use fine-grained locks. I wanted to send out the fix first.

kcooney:
@sbrannen if would be practically impossible to write a test that would consistently fail. The test would have to try to updating the parent store at the same time as updating the child.

If.you think filing a bug only to fix it in the next few days would be useful, feel free to file the bug.

I also found a second bug in the same part of the code. I didn't see it at first. When you call `gerOrComputeIfAbsent` it will insert into the store even if the parent's parent `Store` has a value for the key.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
sbrannen: > Yes. From `getOrComputeIfAbsent()` where the code accesses the parent store's map without obtained the parent's lock.

Ahhhh. OK. Good catch! 👍

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
sbrannen: > I actually do. I found this bug while rewriting the class to use fine-grained locks. I wanted to send out the fix first.

So, are you saying we currently have a bug that would be fixed if we switched to using `ConcurrentHashMap` as the implementation for `ExtensionValuesStore.storedValues`?

kcooney: No, I'm saying that when I was working on the rewrite, I realized that the existing code had a thread-visibility issue and a functional bug.

The thread-visibility issue would be less likely to happen if we used `ConcurrentMap`.

I added a commit to fix the bug.

kcooney:
I changed this to have two commits.

1. fix the functional bug: `getOrComputeIfAbsent()` doesn't check the grandparent `Store` before computing the value
2. fix for the thread-safety issue

I suggest not squashing the commits when you merge this PR

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
sbrannen: OK

sbrannen:
> If.you think filing a bug only to fix it in the next few days would be useful, feel free to file the bug. 😊

We often file bugs first.

But I'm fine with foregoing that, so I've added the "bug" label to this PR. 😉

sbrannen:
> I also found a second bug in the same part of the code. I didn't see it at first. When you call gerOrComputeIfAbsent it will insert into the store even if the parent's parent Store has a value for the key.

Oh wow... Good catch! 👍 

Thank you!

sbrannen:
> it would be practically impossible to write a test that would consistently fail. The test would have to try to updating the parent store at the same time as updating the child.

I agree that would be _challenging_ to test in a robust fashion. So let's just leave it without the test.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java
sbrannen: @kcooney, FYI: I pushed commit ac16d703171a414a8a253875c29144c0a76bacb9 based on your recommendation.

Cheers!

PR #1016, opened 2017-08-13
CSV parser returns an empty String instead of null

### LINKED ISSUE #1014: CsvSource with empty strings instead of null
## Overview

When I add the following `CsvSource` to a test

```java
@CsvSource({",''", "'',''", "text1,text2"})
void teaserText(String source, String expected) {
}
```

I would expect, that 

- in the first step `source` is `null`, and `expected` is an empty string.
- in the second step both are empty strings.
- in the third step `source` is `text1`, and `expected` is `test2`.

but two single quotes without any content inside are parsed as `null`.

How can one create an empty string?

sormuras:
## Overview

When using `@CsvSource` or `@CsvFileSource` in a parameterized test, if the CSV parser does not read any character from the input, and the input is within quotes, an empty string `""` is returned instead of `null`.

Closes #1014

---

---

### Definition of Done

- [x]  Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/meta/API.html)


smoyer64:
In the situation above, how would I purposely pass an empty String to my tests?  In JUnitParams, you can't pass a null to a parameter when your parameters are being converted from a CSV and I've gradually learned to believe that returning an empty String is the correct behavior.

sormuras:
See second example at https://github.com/junit-team/junit5/issues/1014#issuecomment-321756169

smoyer64:
Okay ... that looks correct.  [RFC4180](https://tools.ietf.org/html/rfc4180) doesn't really have anything to add since the duality of null versus empty Strings is language dependent.  What does the team think about: ```@CsvSource("null , , empty , \"\"")``` - it seems to me that this should be a valid empty String as well.

sormuras:
Added an example show-casing an empty and a `null` value to the documentation: https://github.com/junit-team/junit5/pull/1016/commits/826742b2c0374c8f8f73beaac352e76b69ae6e93

Needs some written words...

sormuras:
> What does the team think about: `@CsvSource("null , , empty , \"\"")` - it seems to me that this should be a valid empty String as well.

The parser we use internally does not support this, IIRC. You may only set a single escape character. See https://github.com/junit-team/junit5/blob/master/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java#L40 and https://github.com/junit-team/junit5/blob/master/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileArgumentsProvider.java#L55 for the internal parser settings.

I'll add a section to the CSV section in the user-guide -- "including the rationale for the differences" between `@CsvSource` and `@CsvFileSource`. Generally, it boils down to:

`@CvsSource` using ' (single) is logical due to " (double) are used by Java String represantation.

- `"a,b,c"` escaped value
- `"''"` empty value (escaped empty string)
- `""` _null_

`@CvsFileSource` using " (double) feels more natural when reading text files.

- `"a,b,c"` escaped value
- `""` empty value (escaped empty string)
- `` (no character) _null_

documentation/src/test/java/example/ParameterizedTestDemo.java
```diff
@@ -135,9 +136,9 @@ void testWithMultiArgMethodSource(String first, int second) {
 
 	// tag::CsvSource_example[]
 	@ParameterizedTest
-	@CsvSource({ "foo, 1", "bar, 2", "'baz, qux', 3" })
+	@CsvSource({ "foo, 1", "bar, 2", "'baz, qux', 3", "'', 4", ",5" })
 	void testWithCsvSource(String first, int second) {
-		assertNotNull(first);
+		assertNotNull(Objects.toString(first, "<null>"));
```
sbrannen: Unless I'm missing something, `Objects.toString(first, "<null>")` can never return `null`.

So, were you perhaps intending to assert something else?

documentation/src/test/java/example/ParameterizedTestDemo.java
```diff
@@ -146,7 +147,7 @@ void testWithCsvSource(String first, int second) {
 	@ParameterizedTest
 	@CsvFileSource(resources = "/two-column.csv")
 	void testWithCsvFileSource(String first, int second) {
-		assertNotNull(first);
+		assertNotNull(Objects.toString(first, "<null>"));
```
sbrannen: same here

documentation/src/test/java/example/ParameterizedTestDemo.java
sormuras: What about:

    assertEquals(String.class, Objects.toString(first, "<null>").getClass());

?

sormuras: The difficulty here is we have:

- `null`
- `""` (empty String)
- `"foo"`
- `"bar"`
- `"baz, qux"`

smoyer64:
Looking at the section of the users guide for parameterized tests, there's a lot of great information which is exactly the sort of thing I'd like to see when my IDE pops up the javadocs for ```@CsvSource```.  Unfortunately, those javadocs simply link back to the users guide for the important stuff.

I do think it's important to have a good users guide for those who are jumping into junit5 but I believe that those who are heavy users will appreciate the javadocs being complete *and* stand-alone.  It's a lot more work to keep up-to-date if the information is in both places but I believe for things like this, it's important - kind of like date and regex format descriptions in the Java SE javadocs.

documentation/src/test/java/example/ParameterizedTestDemo.java
sbrannen: Sure. It's _difficult_ because it's a meaningless demo (like many of our demos). 😉 

So perhaps it would be better and clearer to the user to move the empty and null use cases to separate test methods.

If we don't do that, I'd recommend making use of some conditional logic for the assertion performed for the first parameter, but... I think that makes the example overly complicated. Thus, I still prefer separate test methods.

sbrannen:
@smoyer64, would you mind opening a separate issue to address your concerns about the level of documentation in the JavaDoc?

documentation/src/test/java/example/ParameterizedTestDemo.java
sormuras: > So perhaps it would be better and clearer to the user to move the empty and null use cases to separate test methods.

This.

I'll revert the tests and describe the special cases entirely in the text. Like done here: https://github.com/junit-team/junit5/pull/1016#issuecomment-322039840

sormuras:
The user guide is updated: http://junit.org/junit5/docs/snapshot/user-guide/#writing-tests-parameterized-tests-sources-CsvSource 

Created #1021 -- "Level of documentation in the JavaDoc"

smoyer64:
Not at all ... I was a bit concerned about writing the comment above because if you generalize it, you end up with all the documentation in javadocs like Mockito or you end up with no javadocs or placeholder javadocs.  Clearly the users guide (with examples) and the javadocs serve two different purposes - I still hate that there's not an easy way to include content in the javadocs so that it's always in sync.

novityFS:
I maybe found a bug with the CsvSource parser for empty strings. Consider the following example:

```java
@ParameterizedTest
    @CsvSource({
        "  '', Content",
        "  '',      ''",
        "  '', ''     ",
    })
void contentTest(final String one, final String two) {
  Assertions.assertNotNull(two);
}
```

Where in the second test the string two is, as expected an empty string, in the third test the string two is ```null```. Obviously the parser is confused by the trailing blanks of the second parameter.

I'm currently using junit jupiter version 5.0.0

sormuras:
Good catch. Smells like a bug in https://github.com/uniVocity/univocity-parsers because if you switch the content inside the input, everything works as expected:

```java
	@ParameterizedTest
	@CsvSource(value = { 
		"'', 1",
		" '', 2",
		"'' , 3",
		" '' , 4" })
	void contentTestTwo(String one, String two) {
		System.out.printf("`%s`, `%s`\n", one, two);
		assertNotNull(one);
		assertNotNull(two);
	}
[...]
``, `1`
``, `2`
``, `3`
``, `4`
```
So, yes, please file a new bug -- even if the real issue lies potentially in an external library.

PR #1018, opened 2017-08-14
Fix selection of method from JUnit 4 parameterized test

### LINKED ISSUE #549: Selecting one method in JUnit 4 parameterized test has no effect - test method is not executed
## Example test code

``` java
import org.junit.Assert;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.junit.runners.Parameterized;

import java.util.Arrays;
import java.util.Collection;

/**
 * Created 20/10/16 16:51
 *
 * @author Vladimir Bogodukhov
 **/
@RunWith(Parameterized.class)
public class JUnit4ParametrizedExample {

    final boolean b;
    final long l;
    final int i;

    @Parameterized.Parameters(name = "{0} ---> : Test")
    public static Collection<Object[]> data() {
        return Arrays.asList(new Object[][]{
                {false, 1L, 21},
                {true, 21L, 11},
                {false, 10L, 21},
        });
    }

    public JUnit4ParametrizedExample(boolean b, long l, int i) {
        this.b = b;
        this.l = l;
        this.i = i;
    }


    @Test
    public void test1() {
        throw new IllegalStateException(); /* always fails */
    }

    @Test
    public void test2() {
        /* always succeed */
    }
}
```

## Test that reproduces the bug

``` java
import org.junit.Assert;
import org.junit.jupiter.api.Tag;
import org.junit.platform.engine.TestExecutionResult;
import org.junit.platform.engine.discovery.DiscoverySelectors;
import org.junit.platform.launcher.*;
import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
import org.junit.platform.launcher.core.LauncherFactory;

import java.util.HashMap;
import java.util.Map;

/**
 * @author Vladimir Bogodukhov
 */
@Tag("DebugRun")
public class DynamicTestsHelperRun {

    enum State {
        skipped,
        failed,
        success,
        aborted
    }

    @Test
    void test1() {
        Launcher launcher = LauncherFactory.create();

        Map<State, Integer> callCounts = new HashMap<>();
        launcher.registerTestExecutionListeners(new TestExecutionListener() {
            @Override
            public void executionSkipped(TestIdentifier identifier, String reason) {

                if (identifier.isTest()) {
                    callCounts.putIfAbsent(State.skipped, 0);
                    callCounts.put(State.skipped, callCounts.get(State.skipped) + 1);
                }
            }

            @Override
            public void executionFinished(TestIdentifier identifier, TestExecutionResult result) {
                if (identifier.isTest()) {
                    State state = calcState(result);
                    callCounts.putIfAbsent(state, 0);
                    callCounts.put(state, callCounts.get(state) + 1);
                }
            }
        });


        final LauncherDiscoveryRequestBuilder builder = LauncherDiscoveryRequestBuilder.request();
        builder.selectors(DiscoverySelectors.selectJavaClass(JUnit4ParametrizedExample.class));
        LauncherDiscoveryRequest discoveryRequest = builder.build();

        TestPlan discover = launcher.discover(discoveryRequest);
        launcher.execute(discoveryRequest); 
        /* We call class with two parametrized methods - each method is called three times.  One test always fails another always succeed 
        And we get right result
        */

        Assert.assertEquals(new HashMap<State, Integer>() {
            {
                put(State.success, 3); /* Three succedd */
                put(State.failed, 3);/* Three failed */
            }
        }, callCounts);
    }


    @Test
        // Test that fails but certainly mustn't
    void test2() {
        Launcher launcher = LauncherFactory.create();

        Map<State, Integer> callCounts = new HashMap<>();
        launcher.registerTestExecutionListeners(new TestExecutionListener() {
            @Override
            public void executionSkipped(TestIdentifier identifier, String reason) {

                if (identifier.isTest()) {
                    callCounts.putIfAbsent(State.skipped, 0);
                    callCounts.put(State.skipped, callCounts.get(State.skipped) + 1);
                }
            }

            @Override
            public void executionFinished(TestIdentifier identifier, TestExecutionResult result) {
                if (identifier.isTest()) {
                    State state = calcState(result);
                    callCounts.putIfAbsent(state, 0);
                    callCounts.put(state, callCounts.get(state) + 1);
                }
            }
        });

        final LauncherDiscoveryRequestBuilder builder = LauncherDiscoveryRequestBuilder.request();
        builder.selectors(DiscoverySelectors.selectJavaMethod(JUnit4ParametrizedExample.class, "test1")); 
        LauncherDiscoveryRequest discoveryRequest = builder.build();

        TestPlan discover = launcher.discover(discoveryRequest);
        launcher.execute(discoveryRequest);
        /* we select  test1() method that always fails */



        /* if we call to test method that fails three times it is logical to get three failures as result */
        Assert.assertEquals(new HashMap<State, Integer>() {
            {
                put(State.failed, 3);/* Three failed */
            }
        }, callCounts);
        /* However it is not so -  we get actually only one call.  and we do not stop at the breakpoint  in test code  */
    }

    private static State calcState(TestExecutionResult result) {
        TestExecutionResult.Status status = result.getStatus();
        switch (status) {
            case ABORTED: {
                return State.aborted;
            }

            case FAILED: {
                return State.failed;
            }

            case SUCCESSFUL: {
                return State.success;
            }

            default: {
                throw new IllegalStateException("unknown state");
            }
        }
    }

}
```

Not that using `DiscoverySelectors.selectJavaMethod(JUnit4ParametrizedExample.class, JUnit4ParametrizedExample.class.getMethod("test1"))` instead of `DiscoverySelectors.selectJavaMethod(JUnit4ParametrizedExample.class, "test1")` has the same effect.

Andrei94:
## Overview

When selecting a test method from a JUnit 4 parameterized test it, the method would run only once, without considering the parameters it should receive.

Resolves #549 

---

---

### DIFF
--- a/junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/MethodSelectorResolver.java
+++ b/junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/MethodSelectorResolver.java
@@ -10,7 +10,6 @@
 
 package org.junit.vintage.engine.discovery;
 
-import static org.junit.runner.manipulation.Filter.matchMethodDescription;
 import static org.junit.vintage.engine.discovery.RunnerTestDescriptorAwareFilter.adapter;
 
 import java.lang.reflect.Method;
@@ -18,6 +17,7 @@ import java.lang.reflect.Method;
 import org.junit.platform.engine.EngineDiscoveryRequest;
 import org.junit.platform.engine.discovery.MethodSelector;
 import org.junit.runner.Description;
+import org.junit.runner.manipulation.Filter;
 
 /**
  * @since 4.12
@@ -36,4 +36,35 @@ class MethodSelectorResolver implements DiscoverySelectorResolver {
 		collector.addFiltered(testClass, adapter(matchMethodDescription(methodDescription)));
 	}
 
+	/**
+	 * The method {@link Filter#matchMethodDescription(Description)} returns a filter that does not account for
+	 * the case when the description is for a Parametrized runner.
+	 */
+	private static Filter matchMethodDescription(final Description desiredDescription) {
+		return new Filter() {
+			@Override
+			public boolean shouldRun(Description description) {
+				if (description.isTest()) {
+					return desiredDescription.equals(description) || isMethodWithParameters(description);
+				}
+
+				// explicitly check if any children want to run
+				for (Description each : description.getChildren()) {
+					if (shouldRun(each)) {
+						return true;
+					}
+				}
+				return false;
+			}
+
+			private boolean isMethodWithParameters(Description description) {
+				return description.getMethodName().startsWith(desiredDescription.getMethodName() + "[");
+			}
+
+			@Override
+			public String describe() {
+				return String.format("Method %s", desiredDescription.getDisplayName());
+			}
+		};
+	}
 }
--- a/junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParameterizedTestCase.java
+++ b/junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParameterizedTestCase.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.vintage.engine;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Parameterized.class)
+public class JUnit4ParameterizedTestCase {
+
+	private final boolean b;
+	private final long l;
+	private final int i;
+
+	@Parameterized.Parameters(name = "{0} ---> : Test")
+	public static Collection<Object[]> data() {
+		return Arrays.asList(new Object[][] { { 1L, false, 21 }, { 2L, true, 11 }, { 3L, false, 21 }, });
+	}
+
+	public JUnit4ParameterizedTestCase(long l, boolean b, int i) {
+		this.b = b;
+		this.l = l;
+		this.i = i;
+	}
+
+	@Test
+	public void test1() {
+		fail("this test should fail");
+	}
+
+	@Test
+	public void endingIn_test1() {
+		fail("this test should fail");
+	}
+
+	@Test
+	public void test1_atTheBeginning() {
+		fail("this test should fail");
+	}
+
+	@Test
+	public void test2() {
+		/* always succeed */
+	}
+}
--- a/junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParameterizedTests.java
+++ b/junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParameterizedTests.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v1.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.vintage.engine;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
+import static org.junit.platform.engine.discovery.DiscoverySelectors.selectMethod;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.launcher.Launcher;
+import org.junit.platform.launcher.LauncherDiscoveryRequest;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
+import org.junit.platform.launcher.core.LauncherFactory;
+
+class JUnit4ParameterizedTests {
+	private final Launcher launcher = LauncherFactory.create();
+	private final Map<TestExecutionResult.Status, Integer> callCounts = new HashMap<>();
+	private final LauncherDiscoveryRequestBuilder requestBuilder = LauncherDiscoveryRequestBuilder.request();
+
+	@BeforeEach
+	void setUpLauncher() {
+		launcher.registerTestExecutionListeners(new TestListenerMock());
+	}
+
+	@Test
+	void selectingWholeParameterizedClassRunsTestsWithAllValues() {
+		requestBuilder.selectors(selectClass(JUnit4ParameterizedTestCase.class));
+		LauncherDiscoveryRequest discoveryRequest = requestBuilder.build();
+
+		launcher.execute(discoveryRequest);
+
+		HashMap<TestExecutionResult.Status, Integer> expectedCallCounts = new HashMap<>();
+		expectedCallCounts.put(TestExecutionResult.Status.SUCCESSFUL, 3);
+		expectedCallCounts.put(TestExecutionResult.Status.FAILED, 9);
+		assertEquals(expectedCallCounts, callCounts);
+	}
+
+	@Test
+	void selectingOneTestFromParameterizedClassRunsWithAllValues() {
+		requestBuilder.selectors(selectMethod(JUnit4ParameterizedTestCase.class, "test1"));
+		LauncherDiscoveryRequest discoveryRequest = requestBuilder.build();
+
+		launcher.execute(discoveryRequest);
+
+		HashMap<TestExecutionResult.Status, Integer> expectedCallCounts = new HashMap<>();
+		expectedCallCounts.put(TestExecutionResult.Status.FAILED, 3);
+		assertEquals(expectedCallCounts, callCounts);
+	}
+
+	private class TestListenerMock implements TestExecutionListener {
+		@Override
+		public void executionFinished(TestIdentifier identifier, TestExecutionResult result) {
+			if (identifier.isTest()) {
+				final TestExecutionResult.Status status = result.getStatus();
+				callCounts.merge(status, 1, Integer::sum);
+			}
+		}
+	}
+}


junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+
+package org.junit.vintage.engine;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.Assert;
```
sbrannen: We do not use `org.junit.Assert`. 

Please use static imports for methods in `org.junit.jupiter.api.Assertions`.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+	private LauncherDiscoveryRequestBuilder requestBuilder;
+	enum State {
+		failed, success,
+	}
+
+	@BeforeEach
+	void setUpLauncher() {
+		this.launcher = LauncherFactory.create();
+		this.callCounts = new HashMap<>();
+		this.requestBuilder = LauncherDiscoveryRequestBuilder.request();
```
sbrannen: Please move the previous three initialization lines to the corresponding fields and make the fields `final`.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/MethodSelectorResolver.java
```diff
@@ -36,4 +36,28 @@ private void resolve(MethodSelector selector, TestClassCollector collector) {
 		collector.addFiltered(testClass, adapter(matchMethodDescription(methodDescription)));
 	}
 
+	private static Filter matchMethodDescription(final Description desiredDescription) {
+		return new Filter() {
+			@Override
+			public boolean shouldRun(Description description) {
+				if (description.isTest()) {
+					return desiredDescription.equals(description)
+							|| description.getMethodName().startsWith(desiredDescription.getMethodName() + "[");
```
sbrannen: Please add an inline comment explaining the additional logic that accounts for parameterized tests.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/MethodSelectorResolver.java
```diff
@@ -36,4 +36,28 @@ private void resolve(MethodSelector selector, TestClassCollector collector) {
 		collector.addFiltered(testClass, adapter(matchMethodDescription(methodDescription)));
 	}
 
+	private static Filter matchMethodDescription(final Description desiredDescription) {
```
sbrannen: Please document where the code for `matchMethodDescription()` was copied from, including the rationale for why it was copied and modified.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+		launcher.registerTestExecutionListeners(new TestListenerMock(callCounts));
+	}
+
+	@Test
+	void succeeding() {
+		requestBuilder.selectors(DiscoverySelectors.selectClass(JUnit4ParametrizedExample.class));
+		LauncherDiscoveryRequest discoveryRequest = requestBuilder.build();
+
+		launcher.discover(discoveryRequest);
```
sbrannen: You can delete this line as it is superfluous.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+		launcher.execute(discoveryRequest);
+
+		final HashMap<State, Integer> expectedCallCounts = new HashMap<>();
```
sbrannen: Please remove the `final` declaration.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+
+		Assert.assertEquals(expectedCallCounts, callCounts);
+	}
+
+	@Test
+	void failing() {
+		requestBuilder.selectors(DiscoverySelectors.selectMethod(JUnit4ParametrizedExample.class, "test1"));
+		LauncherDiscoveryRequest discoveryRequest = requestBuilder.build();
+
+		launcher.discover(discoveryRequest);
```
sbrannen: ditto

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+package org.junit.vintage.engine;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.Assert;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.engine.discovery.DiscoverySelectors;
```
sbrannen: Please use static imports for methods in `DiscoverySelectors`.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.platform.launcher.TestIdentifier;
+import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
+import org.junit.platform.launcher.core.LauncherFactory;
+
+class DynamicTestsHelperRun {
+
+	private Launcher launcher;
+	private Map<State, Integer> callCounts;
+	private LauncherDiscoveryRequestBuilder requestBuilder;
+	enum State {
```
sbrannen: Why did you create a new `State` enum?

Is there something wrong with simply using `org.junit.platform.engine.TestExecutionResult.Status`?

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+		TestListenerMock(Map<State, Integer> callCounts) {
+			this.callCounts = callCounts;
+		}
+
+		@Override
+		public void executionFinished(TestIdentifier identifier, TestExecutionResult result) {
+			if (identifier.isTest()) {
+				State state = calcState(result);
+				callCounts.putIfAbsent(state, 0);
+				callCounts.put(state, callCounts.get(state) + 1);
```
sbrannen: Can't you combine the previous two lines into a single call to `java.util.Map.merge(K, V, BiFunction<? super V, ? super V, ? extends V>)`?

junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParametrizedExample.java
```diff
@@ -0,0 +1,57 @@
...
+		this.i = i;
+	}
+
+	@Test
+	public void test1() {
+		throw new IllegalArgumentException();
+	}
+
+	@Test
+	public void endingInTest1() {
```
sbrannen: Wouldn't this method need to be named `endingIn_test1()` to make it useful?

junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParametrizedExample.java
```diff
@@ -0,0 +1,57 @@
...
+
+	public JUnit4ParametrizedExample(long l, boolean b, int i) {
+		this.b = b;
+		this.l = l;
+		this.i = i;
+	}
+
+	@Test
+	public void test1() {
+		throw new IllegalArgumentException();
```
sbrannen: Please replace all `throw new IllegalArgumentException();` statements with an invocation of `fail()` with a meaningful failure message.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParametrizedExample.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.Arrays;
+import java.util.Collection;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Parameterized.class)
+public class JUnit4ParametrizedExample {
```
sbrannen: Please rename the class to `JUnit4ParameterizedTestCase` to align with our coding standards.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+	@BeforeEach
+	void setUpLauncher() {
+		this.launcher = LauncherFactory.create();
+		this.callCounts = new HashMap<>();
+		this.requestBuilder = LauncherDiscoveryRequestBuilder.request();
+		launcher.registerTestExecutionListeners(new TestListenerMock(callCounts));
+	}
+
+	@Test
+	void succeeding() {
```
sbrannen: Please give the test method a more meaningful name.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+
+		final HashMap<State, Integer> expectedCallCounts = new HashMap<>();
+		expectedCallCounts.put(State.success, 3);
+		expectedCallCounts.put(State.failed, 9);
+
+		Assert.assertEquals(expectedCallCounts, callCounts);
+	}
+
+	@Test
+	void failing() {
```
sbrannen: Please give the test method a more meaningful name.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.engine.discovery.DiscoverySelectors;
+import org.junit.platform.launcher.Launcher;
+import org.junit.platform.launcher.LauncherDiscoveryRequest;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
+import org.junit.platform.launcher.core.LauncherFactory;
+
+class DynamicTestsHelperRun {
```
sbrannen: This test call will never be included in the JUnit 5 build since it violates our naming conventions.

Furthermore, it has nothing to do with "dynamic tests".

Please rename it to `JUnit4ParameterizedTests`.

sbrannen:
In addition to the requested changes in the review, please rename the title of your commit (i.e., the first line of your commit message) to " Fix selection of method from JUnit 4 parameterized test".

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
Andrei94: This is how the test was written in order to produce the bug.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
Andrei94: This is how the test was written in order to produce the bug.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
Andrei94: ditto

junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParametrizedExample.java
Andrei94: The rationale behind it was to check whether methods ending in the string from the selection (i.e. `test1`) are being run or not, so I don't believe it is necessary to add the underscore.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
Andrei94: Didn't know about this method. Nice 😄

junit-vintage-engine/src/test/java/org/junit/vintage/engine/DynamicTestsHelperRun.java
sbrannen: I don't see how executing the `discover` method could possibly help to reproduce the bug since the `execute` method will perform discovery in the same manner.

In other words, "less is more". If a line of code serves no purpose it must be deleted.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/JUnit4ParametrizedExample.java
sbrannen: Well, it depends on whether the string comparison is case sensitive.

If it's case sensitive, then I'm right (and my suggested change is necessary); otherwise, you're right. 😉

Andrei94: Didn't think about this difference. Thanks! 😉

Andrei94:
@sbrannen changes made.

sbrannen:
Looks good now.

Thanks, @Andrei94!

Andrei94:
ditto @sbrannen :wink:

sbrannen:
Added further polishing in 18abf38a10dea91ac65a8789fb0702589f298de1 and documented the fix in the release notes in f66eb9b66732b3bdf98319055426c62b3cda6190.

PR #961, opened 2017-07-18
Enable error-prone static analysis checks

jbduncan:
## Overview

Following issue #955, this is a work-in-progress PR to introduce error-prone checks to JUnit 5.

Feedback is welcome!

---

I hereby agree to the terms of the JUnit Contributor License Agreement.

---

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Method [preconditions](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [x] [Coding conventions](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [x] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#tests)
- [ ] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/meta/API.html)
- [ ] Change is documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [x] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #1028, opened 2017-08-21
Polishing

jbduncan:
## Overview

As requested by @sbrannen in https://github.com/junit-team/junit5/pull/961, I have moved some "Polishing" changes to this PR, whilst adding other changes that I think improve readability.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
@@ -81,9 +81,7 @@ class AssertionUtils {
 			return String.format("expected: %s but was: %s", formatClassAndValue(expected, expectedString),
 				formatClassAndValue(actual, actualString));
 		}
-		else {
-			return String.format("expected: <%s> but was: <%s>", expectedString, actualString);
-		}
+		return String.format("expected: <%s> but was: <%s>", expectedString, actualString);
 	}
 
 	private static String formatClassAndValue(Object value, String valueString) {
@@ -147,9 +145,7 @@ class AssertionUtils {
 		if (obj1 == null) {
 			return (obj2 == null);
 		}
-		else {
-			return obj1.equals(obj2);
-		}
+		return obj1.equals(obj2);
 	}
 
 	private static void failIllegalDelta(String delta) {
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
@@ -70,17 +70,8 @@ public final class StringUtils {
 	 * @see Character#isWhitespace(int)
 	 */
 	public static boolean containsWhitespace(String str) {
-		if (str == null) {
-			return false;
-		}
+		return str != null && str.codePoints().anyMatch(Character::isWhitespace);
 
-		for (int i = 0; i < str.length(); i++) {
-			if (Character.isWhitespace(str.codePointAt(i))) {
-				return true;
-			}
-		}
-
-		return false;
 	}
 
 	/**
@@ -106,17 +97,7 @@ public final class StringUtils {
 	 * @see Character#isISOControl(int)
 	 */
 	public static boolean containsIsoControlCharacter(String str) {
-		if (str == null) {
-			return false;
-		}
-
-		for (int i = 0; i < str.length(); i++) {
-			if (Character.isISOControl(str.codePointAt(i))) {
-				return true;
-			}
-		}
-
-		return false;
+		return str != null && str.codePoints().anyMatch(Character::isISOControl);
 	}
 
 	/**
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
@@ -71,7 +71,6 @@ public final class StringUtils {
 	 */
 	public static boolean containsWhitespace(String str) {
 		return str != null && str.codePoints().anyMatch(Character::isWhitespace);
-
 	}
 
 	/**
--- a/junit-platform-console/src/main/java/org/junit/platform/console/options/Theme.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/options/Theme.java
@@ -79,43 +79,43 @@ public enum Theme {
 		this.blank = new String(new char[vertical().length()]).replace('\0', ' ');
 	}
 
-	public String root() {
+	public final String root() {
 		return tiles[0];
 	}
 
-	public String vertical() {
+	public final String vertical() {
 		return tiles[1];
 	}
 
-	public String blank() {
+	public final String blank() {
 		return blank;
 	}
 
-	public String entry() {
+	public final String entry() {
 		return tiles[2];
 	}
 
-	public String end() {
+	public final String end() {
 		return tiles[3];
 	}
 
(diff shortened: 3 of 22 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
```diff
@@ -147,9 +145,7 @@ static boolean objectsAreEqual(Object obj1, Object obj2) {
 		if (obj1 == null) {
 			return (obj2 == null);
 		}
-		else {
-			return obj1.equals(obj2);
-		}
+		return obj1.equals(obj2);
```
marcphilipp: Isn't this equivalent to `Objects.equals()`?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
```diff
@@ -70,17 +70,8 @@ public static boolean isNotBlank(String str) {
 	 * @see Character#isWhitespace(int)
 	 */
 	public static boolean containsWhitespace(String str) {
-		if (str == null) {
-			return false;
-		}
+		return str != null && str.codePoints().anyMatch(Character::isWhitespace);
```
marcphilipp: blank line before closing parenthesis?

junit-platform-console/src/main/java/org/junit/platform/console/options/Theme.java
```diff
@@ -79,43 +79,43 @@ public static Theme valueOf(Charset charset) {
 		this.blank = new String(new char[vertical().length()]).replace('\0', ' ');
 	}
 
-	public String root() {
+	public final String root() {
```
marcphilipp: Why these? `Theme` is final anyway, isn't it?

jbduncan: > `Theme` is final anyway, isn't it?

Yeah, I wondered about that...

> Why these?

The reason is a bit complicated, so please bear with me. :)

Before we can [successfully adopt error-prone](https://github.com/junit-team/junit5/pull/961), we need to make sure that all error-prone checks that we have enabled pass, because error-prone is an OpenJDK-derived compiler developed by Google that makes additional compile-time guarantees over the other javac compilers out there, and so if we want things to compile successfully, error-prone needs to be appeased.

We don't strictly need the `final`s for most of the methods in `Theme`, but for whatever reason, we need it at least on `Theme.vertical()`, because error-prone sees it as an overridable method that is being called by a constructor - `Theme`'s constructor specifically. In most cases, i.e. with non-final classes, this would be a real security bug, because then malicious users could subclass and change the behaviour of `vertical()`, thus changing the behaviour of the constructor and causing to be in an invalid state.

However, as you realised, since `Theme` is an enum, it's effectively final.

It seems that error-prone isn't intelligent enough yet to realise that, when an _enum_ constructor calls an overridable method, it can be safely ignored since enums cannot be subclassed. I suspect that error-prone takes a liberal approach where it considers _every_ class or class-like structure to be non-final, and for this reason it mandates at compile-time that `Theme.vertical()` _must_ be `final`.

I should really submit it as a bug report to the error-prone guys, but in the short-term at least, I thought it sensible to make not _only_ `Theme.vertical()` `final`, but every other method as well, to make things look consistent.

It's not ideal, yes, but I currently cannot see a way forward other than disabling the error-prone rule that mandates the `final` keyword's existence (where the rule is http://errorprone.info/bugpattern/ConstructorInvokesOverridable), which I consider to be the worse option of the two.

I hope this clarifies things for you @marcphilipp.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
jbduncan: Oh, perhaps!

I do seem to vaguely remember though that, some time ago, somewhere in the JUnit 5 codebase, that `Objects.equals()` was used somewhere in the JUnit 5 code base, but it was realised that it was the cause of a bug, and so it was swapped out for correctly-functioning equivalence code.

I cannot remember if it was this method where the bug was found. So I'll do a quick search of the GitHub history, and if I can't find anything to do with this bug, then I'll try swapping the code out for `Objects.equals()` and see what happens.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
jbduncan: Sorry @marcphilipp, it's not clear to me where you'd like to me insert this blank line. Could you be more specific for me (e.g. before/after a certain line number)?

junit-platform-console/src/main/java/org/junit/platform/console/options/Theme.java
jbduncan: @marcphilipp As it turns out, `veritical()` _can_ be overridden! (Thanks for @cushon over at https://github.com/google/error-prone/issues/709#issuecomment-323873109 for pointing this out to me.)

If one writes the following code:
```java
public enum Theme {
  ASCII(".", "| ", "+--", "'--", "[OK]", "[A]", "[X]", "[S]") {
    @Override public String vertical() { ... }
  } ,
  ...
}
```

Then one is effectively overriding `Theme.vertical()` from within the `ASCII` constant. So it seems to me that ("ConstructorInvokesOverridable")[http://errorprone.info/bugpattern/ConstructorInvokesOverridable] is working correctly.

Thus we at the very least must either make `Theme.vertical()` final or disable the rule (which seems to be the least wise option to me).

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
jbduncan: @marcphilipp Okay, I think I was right, because when I replace the contents of `AssertionUtils.objectsAreEqual()` with `return Objects.equals(obj1, obj2); `, it causes a few tests to fail.

So, to answer your question: no, it seems that apparently the code here _isn't_ equivalent to `Objects.equals()`. :)

junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
marcphilipp: Sorry for not being clear. I want you to _remove_ the blank line.

junit-platform-console/src/main/java/org/junit/platform/console/options/Theme.java
marcphilipp: Right, forgot about that possibility. In that case, I'm fine with declaring all methods `final`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
marcphilipp: Ok, thanks for checking! 👍

junit-platform-commons/src/main/java/org/junit/platform/commons/util/StringUtils.java
jbduncan: Oh, cool, I'd accidentally missed that blank line! Thanks for pointing it out to me. :)

jbduncan:
@marcphilipp I believe I've addressed all of your comments here as well. Please do let me know if there's anything I missed. :)

sbrannen:

Thanks!

jbduncan:
Many thanks @sbrannen! :)

Issue #856, opened 2017-05-18
Revise declarations of all @API annotations


marcphilipp:
## Overview

Right now, most of our API is marked as "experimental".

## Deliverables

Before the GA release...

- [x] Remove `@API` declarations on non-public types.
- [x] Go through all public APIs and decide on the correct `value` of all `@API` annotations.
- [ ] Apply `@API` annotations to Groovy source code for the Gradle plugin.
- [x] New section/warnings in User Guide for _experimental_ APIs

**`@API(Stable)`**

- [x] jupiter test annotations, including currently maintained ones (except TestFactory)
- [x] JUnitPlatform (runner)
- [x] Assertions/Assumptions
- [x] RepetitionInfo, TestInfo, TestReporter
- [x] org.junit.jupiter.api.extension
- [x] org.junit.platform.engine (except hierarchical)
- [x] org.junit.platform.launcher & .core
- [x] Constants

**`@API(Maintained)`**

- [x] suite api
- [x] org.junit.platform.engine.support.hierarchical
- [x] org.junit.platform.launcher.listeners

**`@API(Experimental)`**

- [x] DynamicContainer/… and TestFactory
- [x] params

**`@API(Internal)`**

- [x] API
- [x] JupiterTestEngine, VintageTestEngine (final)
- [x] ClasspathScanningSupport

sbrannen:
Update: I've already made one pass through the code base removing `@API` declarations on non-public types.

sormuras:
**224** usages of `API` found by IDEA in all project files.
**190** usages of `API` found  in the scope of `junit-platform-console-standalone`:

```
### Internal (56) ###

  org.junit.jupiter.engine.descriptor.ClassExtensionContext
  org.junit.jupiter.engine.descriptor.ClassTestDescriptor
  org.junit.jupiter.engine.descriptor.JupiterEngineDescriptor
  org.junit.jupiter.engine.descriptor.JupiterEngineExtensionContext
  org.junit.jupiter.engine.descriptor.JupiterTestDescriptor
  org.junit.jupiter.engine.descriptor.MethodExtensionContext
  org.junit.jupiter.engine.descriptor.NestedClassTestDescriptor
  org.junit.jupiter.engine.descriptor.TestFactoryTestDescriptor
  org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor
  org.junit.jupiter.engine.descriptor.TestTemplateExtensionContext
[...]

PR #1034, opened 2017-08-28
Add program that prints usages of @API annotation

sormuras:
## Overview

Add program that prints usages of @API annotation.

See current result at https://gist.github.com/sormuras/2f3df0f4d35af83b625eff66d8a9f130

Addresses #856 

---


sormuras:
Prints list(s) sorted by package names now.

PR #1042, opened 2017-09-03
Change PostDiscoveryFilter to filter on TestIdentifier

kcooney:
## Overview

`TestDescriptor` is mutable, so a `PostDiscoveryFilter` could add tests
or directly remove tests, which is problematic. Since `TestIdentifier` is
immutable, changing `PostDiscoveryFilter` to use it solves this problem.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/PostDiscoveryFilter.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/PostDiscoveryFilter.java
@@ -14,11 +14,10 @@ import static org.junit.platform.commons.meta.API.Usage.Stable;
 
 import org.junit.platform.commons.meta.API;
 import org.junit.platform.engine.Filter;
-import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.TestEngine;
 
 /**
- * A {@code PostDiscoveryFilter} is applied to {@link TestDescriptor TestDescriptors}
+ * A {@code PostDiscoveryFilter} is applied to {@link TestIdentifier TestIdentifiers}
  * after test discovery.
  *
  * <p>{@link TestEngine TestEngines} must <strong>not</strong> apply
@@ -29,5 +28,5 @@ import org.junit.platform.engine.TestEngine;
  * @see TestEngine
  */
 @API(Stable)
-public interface PostDiscoveryFilter extends Filter<TestDescriptor> {
+public interface PostDiscoveryFilter extends Filter<TestIdentifier> {
 }
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
@@ -20,6 +20,7 @@ import org.junit.platform.engine.Filter;
 import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.TestEngine;
 import org.junit.platform.launcher.LauncherDiscoveryRequest;
+import org.junit.platform.launcher.TestIdentifier;
 
 /**
  * Represents the root of all discovered {@link TestEngine TestEngines} and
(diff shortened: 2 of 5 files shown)


sbrannen:
This sounds like a reasonable change, even if exceedingly late in the game. 😉 

Tentatively slated for 5.0 GA in order for the team to discuss it in time.

kcooney:
The other option would be to keep the signature of the filter the same, but wrap the `TestDescription` instances before passing them into the filter with a wrapper that makes them Immutable (see https://github.com/junit-team/junit5/commit/7e2555edaeab873c1525df2a0cc06dd9213a7779). The problem with that approach is that the person writing the filter doesn't know that some of the methods in the passed-in description will throw exceptions. But, of course, that approach wouldn't change any method signatures.

One problem with the approach in this pull is that inspecting parents and children requires having access to the `TestPlan`

sbrannen:
When you say `TestDescription` do you mean `TestDescriptor`? 😉

sbrannen:
> The problem with that approach is that the person writing the filter doesn't know that some of the methods in the passed-in description will throw exceptions. But, of course, that approach wouldn't change any method signatures.

Well, this behavior could/should in fact be documented, in which case the filter author would be _informed_.

> One problem with the approach in this pull is that inspecting parents and children requires having access to the `TestPlan`

Hmmmm.... will have to ponder it a bit.

kcooney:
@sbrannen Yes, I meant `TestDescriptor`

> Well, this behavior could/should in fact be documented, in which case the filter author would be informed.

Certainly. If we went with the approach in 7e2555e we could even make a `PostDiscoveryFilter` a `Filter<UnmodifiableTestDescriptor>` so that it's obvious that the descriptor cannot be modified.

marcphilipp:
> One problem with the approach in this pull is that inspecting parents and children requires having access to the `TestPlan`.

I share this concern and therefore prefer making `TestDescriptor` unmodifiable for `PostDiscoveryFilters`.

> If we went with the approach in 7e2555e we could even make a `PostDiscoveryFilter` a `Filter<UnmodifiableTestDescriptor>` so that it's obvious that the descriptor cannot be modified.

How about we split `TestDescriptor` into a interface with only getters ("TestDescriptor") and a sub-interface ("MutableTestDescriptor") that adds methods to mutate it, e.g. `addChild()` etc. That way we wouldn't have to wrap it.

Both approaches share one flaw: What would we do with the `Visitor` interface?

marcphilipp:
BTW I've now documented that `PostDiscoveryFilters` must not mutate `TestDescriptors` in 371df45f6dea3d581f4577f64ce7139a1db72731.

On second thought, I think splitting the interface is too much just for this one use case of filtering. I'm fine with having the documentation in place for now and maybe later preventing or detecting mutation. That wouldn't require a breaking change (as long as we don't change the signature of `PostDiscoveryFilter`), so we can do it in 5.1+.

bechte:
> Well, this behavior could/should in fact be documented, in which case the filter author would be informed

IMHO, one of the worst things in the Java Collections API is the fact that it does not guarantee that the underlying Collection is mutable. This results in a pseudo mutability. If the source (aka the place where it was instanciated at) of a Collection is unknown, one cannot rely on the mutability resulting in a de facto requirement to copy the Collection where it is necessary to modify. I'd rather have the Collection interface not having modifiers but only "getter" methods.

Having said that, I could also live with the documentation here, but I'd rather go along with the identifier or a separate interface for mutable TestDescriptors. As making use of the TestIdentifier has the downside of missing the TestPlan as stated above, splitting the interface is my preferred solution.

> What would we do with the Visitor interface?

I think we can come up with another way of doing it. I would like to tackle this issue during the work on #242 and provide a proposal if that is fine with you.un the meantime I think the documentation done by @marcphilipp is fine for the time being.

kcooney:
I did some experimenting and splitting the interface does indeed become problematic due to the visitor. You would need two visitor classes. You will likely also need two differently-named accept methods if you want to allow users of `accept()` to use function references.

The only use case where we would not want to.allow mutation is in the filter and wanting mutable operations in a filter is so unusual that I am personally fine with the caller getting an unexpected `UnsupportedOperationException` I they try to mutate the passed I description. So I recommend keeping the API as-is and silently wrap with an `UnmodifiableTestDescriptor` and documents on the filter API that the filter should modify the passed in value. We can make  `UnmodifiableTestDescriptor` a package-scope class.

bechte:
The "problem" with the ```TestDescriptor.Visitor``` is that it mixes responsibilities between the ```Launcher``` and the ```Engine```. This becomes now obvious as we would like to have a mutable and immutable "version" of the ```TestDescriptor``` interface.

I was working on a solution for it and pushed my first proposal into a feature branch called "issues/1042-remove-visitor". It is not yet perfect, but it points the direction in which we should go, at least in my oppinion, i.e. having the ```TestEngine`` being _responsible_ for performing the pruning and the actual removal, while the ```Launcher``` _initialized_ the process.

marcphilipp:
I'm closing this PR because we've agreed that using `TestIdentifier` with `TestPlan` is too restrictive.

Issue #939, opened 2017-07-11
Consider static factory methods instead of constructors


jbduncan:
## Overview

Currently, there aren't many classes (if any) in JUnit 5 that are instantiated via _static factory methods_, as opposed to constructors. _Effective Java 2nd Edition_, Item 1 gives a number of compelling reasons for preferring static factory methods. Therefore we should consider adopting them more, skipping over those classes where there is a more compelling reason for them to have public constructors instead.

## Deliverables

- [x] Evaluate which classes should have public _static factory methods_.
- [x] For the following classes, make their constructors private or protected and introduce static factory methods (SFMs) in their place.
  - [x] `ClasspathResourceSource`
  - [x] `ClassSource`
  - [x] `CompositeTestSource`
  - [x] `DirectorySource`
  - [x] `FilePosition`
  - [x] `FileSource`
  - [x] `MethodSource`
  - [x] `PackageSource`
  - [x] `LoggingListener`: make the existing constructor private and introduce a `forBiConsumer(...)` SFM.
- ❌ Introduce a new `TestSources` utility class (analogous to `DiscoverySelectors`) that will house the SFMs for `TestSource` implementations.
  - The team decided to introduce `from(...)` SFMs in each `TestSource` implementation instead.

sbrannen:
slated for RC1 for team discussion

jbduncan:
A possible regex to use to search for such classes (at least within IntelliJ IDEA): 
```regex
"public\s*(?!.*final|.*abstract)\s*class"`.
```

sbrannen:
That RegEx also works in Eclipse.

sbrannen:
@jbduncan, do you have any particular classes in mind?

sbrannen:
> Currently, there aren't many classes (if any) in JUnit 5 that are instantiated via static factory methods, as opposed to constructors.

That's actually not true.

We have several classes with static factory methods, and in other situations the constructor is package private so that no third party could invoke them anyway.

jbduncan:
@sbrannen

> That's actually not true.
>
> We have several classes with static factory methods, and in other situations the constructor is package private so that no third party could invoke them anyway.

Oh yes, you're right! I admit I didn't search thoroughly for offending classes when I opened the issue, so apologies.

> @jbduncan, do you have any particular classes in mind?

I've identified the following list of classes which I think should be considered for the static factory method treatment.

[...]

PR #1046, opened 2017-09-07
Replace constructors with static factory methods

sbrannen:
## Overview

This pull request includes changes discussed in #939.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
@@ -14,13 +14,15 @@ import static org.junit.platform.commons.meta.API.Usage.Stable;
 
 import java.io.Serializable;
 import java.util.Objects;
+import java.util.Optional;
 
 import org.junit.platform.commons.meta.API;
+import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.commons.util.ToStringBuilder;
 
 /**
  * Position inside a file represented by {@linkplain #getLine line} and
- * {@linkplain #getColumn column}.
+ * {@linkplain #getColumn column} numbers.
  *
  * @since 1.0
  */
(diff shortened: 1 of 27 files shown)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/LoggingListener.java
```diff
@@ -73,7 +74,13 @@ public static LoggingListener forJavaUtilLogging(Level logLevel) {
 	 * @see #forJavaUtilLogging()
 	 * @see #forJavaUtilLogging(Level)
 	 */
-	public LoggingListener(BiConsumer<Throwable, Supplier<String>> logger) {
+	public static LoggingListener forBiConsumer(BiConsumer<Throwable, Supplier<String>> logger) {
```
jbduncan: @sbrannen I realise  that it's not clear to me from this javadoc what the BiConsumer parameter consumes, so I wonder if having a `@param` description for the BiConsumer, with additional notes on the two effective parameters of the BiConsumer itself, would improve clarity here.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
```diff
@@ -29,17 +29,21 @@
...
-
 	/**
 	 * Create a new {@code FilePosition} using the supplied {@code line} and
 	 * {@code column}.
 	 *
 	 * @param line the line (1-based)
 	 * @param column the column (1-based)
 	 */
-	public FilePosition(int line, int column) {
+	public static FilePosition from(int line, int column) {
```
jbduncan: @sbrannen I wonder if this SFM would be better named something like `fromLineAndColumn(int, int)`, as looking at the usages of the SFM in the tests below, they confuse me I cannot tell what the two `int` parameters mean without looking at the javadoc.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
```diff
@@ -29,17 +29,21 @@
...
+		return new FilePosition(line, column);
+	}
+
+	private final int line;
+	private final int column;
+
+	private FilePosition(int line, int column) {
 		this.line = line;
 		this.column = column;
```
jbduncan: @sbrannen This would probably warrant another issue, but I think that `line` and `column` may need precondition guards to ensure they are positive (> 0) numbers, as it personally makes little sense to me for line and column numbers in source files to be anything less than 1.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/LoggingListener.java
marcphilipp: Good point! Addressed in 838cb09.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
marcphilipp: What about the case where a `TestEngine` can only identify a line but not a column? Would it pass 1 as `column` then?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
marcphilipp: That's a valid point. However, if we make an exception here, do we also need to use more specific names in all `TestSource` implementations?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
jbduncan: @marcphilipp Good point. Perhaps we could enforce a precondition of = 0 or -1, where 0 or -1 would mean that there is no meaningful line or column? Or how about accepting `@Nullable Integer`s instead, where `null` would mean the same thing?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
jbduncan: I personally don't see a need to do the same for `TestSource` implementations, since they only accept one argument, and those arguments are self-descriptive AFAICT.

jbduncan: My concern is when we e.g. have two consecutive `int` parameters, which (1) are easy to confuse with one another, and (2) don't have any inherent meaning to their type apart from being whole numbers.

jbduncan: This may warrant using the Builder pattern instead of SFMs for `FilePosition` specifically, so rather than having the following:

```java
FilePosition f = FilePosition.from(10, 10);
```

we'd have this instead, which IMO is easier to understand.

```java
FilePosition f = FilePosition.builder()
    .line(10)
    .column(10)
    .build():
```

jbduncan: This would allow us to have poor-man's named parameters a la Python.

marcphilipp: Looks a bit over-engineered for two ints, doesn't it?

jbduncan: Yes, looking back over this, I agree with you that using Builder here would over-engineer things a bit.

I still advocate for at least an SFM like `fromLineAndColumn`, but I can see that this is starting to fall into bikeshedding territory, so if you and @sbrannen disagree with that name and prefer to stick with `from`, then I'll relent. :)

marcphilipp:
Team decision:
- Introduce second SFM `from(int line)` which will validate `line > 0`.
- SFM `from(int line, int column)` will validate `line > 0` and `column > 0`.
- Change `getColumn()` to return `Optional<Integer>`.

PR #1049, opened 2017-09-08
Polishing

jbduncan:
## Overview

This PR introduces some polishing changes, including:
- Moving parentheses to make the precedence of certain operations clearer.
- Replacing a complicated usage of the ternary operator (?:) with if/else statements.
- Shortening very simple if/else statements to ternary operators.
- Applying `private` and `final` for class fields where applicable.
- Using `Stream` more for brevity and clarity.
- Using AssertJ's assertions over JUnit 5's own assertions more consistently.
- Renaming `URIConverter` to `UriConverter` for consistency with `UriSelector`.

---

I hereby agree to the terms of the JUnit Contributor License Agreement.

---

### Definition of Done

- [x] There are no TODOs left in the code
- [x] [Coding conventions](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [x] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#tests)(http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/meta/API.html)
- [ ] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #1052, opened 2017-09-08
Polishing

jbduncan:
## Overview

This PR is a smaller separated part of #1049, as requested by @sbrannen. It does the following:
- Rename `URIConverter` to `UriConverter` for consistency with `UriSelector`.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-console/src/main/java/org/junit/platform/console/options/UriConverter.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/options/UriConverter.java
@@ -20,7 +20,7 @@ import joptsimple.ValueConverter;
 /**
  * @since 1.0
  */
-class URIConverter implements ValueConverter<URI> {
+class UriConverter implements ValueConverter<URI> {
 
 	@Override
 	public URI convert(String value) {
--- a/junit-platform-console/src/main/java/org/junit/platform/console/options/AvailableOptions.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/options/AvailableOptions.java
@@ -124,7 +124,7 @@ class AvailableOptions {
 		selectedUris = parser.acceptsAll(asList("u", "select-uri"), //
 			"Select a URI for test discovery. This option can be repeated.") //
 				.withRequiredArg() //
-				.withValuesConvertedBy(new URIConverter());
+				.withValuesConvertedBy(new UriConverter());
 
 		selectedFiles = parser.acceptsAll(asList("f", "select-file"), //
 			"Select a file for test discovery. This option can be repeated.") //


junit-platform-console/src/main/java/org/junit/platform/console/options/UriConverter.java
```diff
@@ -20,7 +20,7 @@
 /**
  * @since 1.0
  */
-class URIConverter implements ValueConverter<URI> {
+class UriConverter implements ValueConverter<URI> {
```
JLLeitschuh: Shouldn't this match the datatype name? The class you are converting from is called `java.net.URI`. Doesn't it make sense to be consistent with that?

jbduncan: I think it would be better for it to be consistent with the other class `UriSelector` that uses this Capital-Letter-Then-Lowercase format, but that's probably because I'm biased and prefer Google's Java Style Guide for class naming, which advocates this style. :)

If the JUnit 5 team agree with you on this, then I'd be more than happy to make `UriSelector` consistent with this class instead (i.e. naming it to `URIConverter`).

JLLeitschuh: I think renaming `UriSelector` to `URISelector` seems like the more reasonable approach but I'm just a random contributor so it is best to wait until the team gives you feedback.

marcphilipp: > Shouldn't this match the datatype name? The class you are converting from is called `java.net.URI`. Doesn't it make sense to be consistent with that?

Yes, that was my original intent.

> I think renaming `UriSelector` to `URISelector` seems like the more reasonable approach but I'm just a random contributor so it is best to wait until the team gives you feedback.

That would be a breaking change now that 5.0 is released.

I think we should leave it as-is.

sormuras: "leave it as-is" +1

jbduncan: @marcphilipp

> That would be a breaking change now that 5.0 is released.

Oh, really? I'm sure I'm brainfarting ATM (for a lack of a better term), but I can't think of why this would be the case, as by my understanding `UriSelector` is package-private and thus inaccessible by JUnit 5 users, so changing its name shouldn't be a breaking change.

marcphilipp: `UriSelector` is public and marked as "stable".

marcphilipp:
Closing because we want to keep `URISelector`.

junit-platform-console/src/main/java/org/junit/platform/console/options/UriConverter.java
jbduncan: Ahhh, makes sense to me. Ta, @marcphilipp, for clarifying this for me. :)

jbduncan:
Okey dokey. 👍

PR #1055, opened 2017-09-09
Collecting launcher

kcooney:
## Overview

Add CollectingLauncher which provides access to TestCollection.
    
The TestCollection allows users of Launcher to review the test plan,
do additional filtering, and execute the tests, without going through
a potentially expensive discovery stage more than once.

This pull contains three commits. Please don't squash them.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-runner/src/main/java/org/junit/platform/runner/JUnitPlatformTestTree.java
+++ b/junit-platform-runner/src/main/java/org/junit/platform/runner/JUnitPlatformTestTree.java
@@ -23,16 +23,23 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 
 import org.junit.platform.commons.util.AnnotationUtils;
+import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.commons.util.StringUtils;
+import org.junit.platform.engine.FilterResult;
+import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.TestSource;
 import org.junit.platform.engine.support.descriptor.ClassSource;
 import org.junit.platform.engine.support.descriptor.MethodSource;
+import org.junit.platform.launcher.PostDiscoveryFilter;
+import org.junit.platform.launcher.TestCollection;
 import org.junit.platform.launcher.TestIdentifier;
 import org.junit.platform.launcher.TestPlan;
 import org.junit.platform.suite.api.SuiteDisplayName;
 import org.junit.platform.suite.api.UseTechnicalNames;
 import org.junit.runner.Description;
 import org.junit.runner.manipulation.Filter;
+import org.junit.runner.manipulation.NoTestsRemainException;
+import org.junit.runner.notification.RunNotifier;
 
 /**
  * @since 1.0
@@ -40,14 +47,22 @@ import org.junit.runner.manipulation.Filter;
 class JUnitPlatformTestTree {
 
 	private final Map<TestIdentifier, Description> descriptions = new HashMap<>();
-	private final TestPlan plan;
+	private final TestCollection testCollection;
 	private final Function<TestIdentifier, String> nameExtractor;
-	private final Description suiteDescription;
+	private final String suiteDisplayName;
+	private Description suiteDescription;
 
-	JUnitPlatformTestTree(TestPlan plan, Class<?> testClass) {
-		this.plan = plan;
-		this.nameExtractor = useTechnicalNames(testClass) ? this::getTechnicalName : TestIdentifier::getDisplayName;
-		this.suiteDescription = generateSuiteDescription(plan, testClass);
+	JUnitPlatformTestTree(TestCollection testCollection, Class<?> testClass) {
+		this.testCollection = testCollection;
+		if (useTechnicalNames(testClass)) {
+			this.nameExtractor = JUnitPlatformTestTree::getTechnicalName;
+			this.suiteDisplayName = testClass.getName();
+		}
+		else {
+			this.nameExtractor = TestIdentifier::getDisplayName;
+			this.suiteDisplayName = getSuiteDisplayName(testClass);
+		}
+		this.suiteDescription = generateSuiteDescription(testCollection);
 	}
 
 	private static boolean useTechnicalNames(Class<?> testClass) {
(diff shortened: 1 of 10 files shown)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultTestCollector.java
```diff
@@ -0,0 +1,104 @@
...
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestPlan;
+
+/**
+ * Default implementation of the {@link TestCollection} API.
+ *
+ * @since 1.0
+ * @see Launcher
+ */
+class DefaultTestCollection implements TestCollection {
```
jbduncan: @kcooney I think the class name here is different from the file's name `DefaultTestCollector`. :)

kcooney: Woops! Fixed. Thanks!

kcooney:
Friendly ping

marcphilipp:
Looks interesting for sure! Can you outline how this would be used by tools like Bazel?

I'm wondering if we could store `Root` and `ConfigurationParameters` in `TestPlan` and get away without introducing another method in `Launcher`/`CollectionLauncher`.

junit-platform-runner/src/main/java/org/junit/platform/runner/JUnitPlatformTestTree.java
```diff
@@ -60,28 +67,64 @@ Description getDescription(TestIdentifier identifier) {
...
 
 	void addDynamicDescription(TestIdentifier newIdentifier, String parentId) {
-		Description parent = getDescription(this.plan.getTestIdentifier(parentId));
-		this.plan.add(newIdentifier);
-		buildDescription(newIdentifier, parent, this.plan);
+		TestPlan plan = this.testCollection.testPlan();
+		TestIdentifier parentTestIdentifier = plan.getTestIdentifier(parentId);
+		Description parent = getDescription(parentTestIdentifier);
+
+		if (parent == null) {
```
kcooney: BTW, I'm not entirely sure how the current code works without this `if` block (see the comment on line 113 for more context)

kcooney:
> Can you outline how this would be used by tools like Bazel?

@marcphilipp For sharded tests, Bazel passes the number of shards and the current shard as environment variables. So if there are 3 shards, the test will run 3 times in parallel in different processes, each time with `NUM_SHARDS=3` but each with a different value of `CURRENT_SHARD`.

Bazel's JUnit test runner would:

1. Call `CollectingLauncher.collect()` to get a `TestCollection`
1. Call `TestCollection.testPlan()`
1. Iterate over the leaf nodes and create a sorted list of tests (this is for correctness in determining the shard map, in case the ordering is different between test runs; Bazel won't run the tests in this order)
1. Assign each test in that list to a shard by iterating through the list and assigning the shard to `index % numShards`(producing a `Map<TestIdentifier, Index>` that maps a test to a shard)
1. Call `applyPostDiscoveryFilters()` to filter out tests that are not on the "current" shard 
1. Call `execute()` to run the tests

Without this change. Bazel would have to go through the discovery phase twice. For large projects with many classes and jars, scanning classes on the classpath can take several seconds. 

> I'm wondering if we could store `Root` and `ConfigurationParameters` in `TestPlan` and get
> away without introducing another method in `Launcher`/`CollectionLauncher`

That would only help if there was a way to filter and execute a `TestPlan` (which is what `TestCollection` does in my pull).

kcooney:
Friendly ping

marcphilipp:
@kcooney Thanks for the ping! I'll take a look on the weekend.

marcphilipp:
> Bazel's JUnit test runner would:
> 
> 1. Call `CollectingLauncher.collect()` to get a `TestCollection`
> 1. Call `TestCollection.testPlan()`
> 1. Iterate over the leaf nodes and create a sorted list of tests (this is for correctness in determining the shard map, in case the ordering is different between test runs; Bazel won't run the tests in this order)
> 1. Assign each test in that list to a shard by iterating through the list and assigning the shard to `index % numShards`(producing a `Map<TestIdentifier, Index>` that maps a test to a shard)
> 1. Call `applyPostDiscoveryFilters()` to filter out tests that are not on the "current" shard 
> 1. Call `execute()` to run the tests

If I understood you correctly, Bazel will still have to go through this process for each shard, right?

Wouldn't a `PostDiscoveryFilter` be able to assign the shard index on the fly and filter at the same time?

kcooney:
Last I checked I could not see a way for a `PostDiscoveryFilter` to see the entire test plan. My plan was to use the API here to collect tests (once, because for large suites it can be expensive), build the shard map (again only once), build the `PostDiscoveryFilter`  using the shard map, the run the previously-collected tests with the `PostDiscoveryFilter` .

Is there some existing API that would allow this?

marcphilipp:
Well, when a `PostDiscoveryFilter` is called, it gets a reference to a `TestDescriptor` which provides access to the test tree for the entire engine it was created by. So it could build a shard map for each engine. Would that be feasible?

kcooney:
@marcphilipp it's my understanding that if I walk up the `TestDescriptor` tree (by repeatedly calling `getParent()`) it will stop at a root node. If there are multiple test engines, there could be multiple root nodes. So if I go with your proposal, it sounds like my `TestDescriptor` would have to build a shard map based on an incomplete view of the tests included in the `TestPlan`, and later update the shard map as JUnit calls repeatedly calls my filter function. That would be a lot more complicated than traversing the `TestPlan`. It's also wouldn't be deterministic unless JUnit provides some guarantees about ordering of calls to the filter function. 

It would be much simpler if I could examine all of the leaf nodes in the test plan before the start of the post-discovery-filter phase. JUnit 4 supports this (by having a root Runner that allows me to traverse the `Description` tree independently of running the tests). I'm just proposing a way to support the same thing in JUnit5.

marcphilipp:
@kcooney I'm hesitant to add another concept to the `Launcher` interface.

How about we add a `initialize(Map<String, TestDescriptor> engineTestDescriptorById)` method to `PostDiscoveryFilter`? That would give it a chance to prepare the shard map and apply it in subsequent calls to `apply(TestDescriptor)`.

If the shard map should be garbage collected or other resources need to be cleared, we could add another `cleanUp()` method or similar. Both could have empty implementations by default.

kcooney:
@marcphilipp I assume by "another concept" you mean `TestCollection`? Although I understand the desire to not have one-off interfaces, here I am extracting out an API for a concept that already exists in the code: the collection of tests that is being filtered during the discovery process. If you look at the third commit, this interface comes in useful for the `JUnitPlatform` where the filtering code now reuses the classes used by the `Launcher` implementation.

Your idea of an initialize method in `PostDiscoveryFilter` is an interesting option, and should work for Bazel. Unfortunately, I don't expect to have much time for JUnit5 in 2018. If after my response above you still think that updating `PostDiscoveryFilter` is your preferred option, someone else will likely have to implement it.

marcphilipp:
Maybe it isn't "another concept", maybe it's just the wording that puts me off. I'm not excited about `TestCollection collect(...)`. It actually does perform discovery, so I'd like to keep that wording. Maybe `PreliminaryDiscoveryResult discoverPreliminarily(...)` or `FilterableDiscoveryResult discoverForFiltering(...)` or something along those lines. 🤔

Moreover, I'd like to get rid of the separate interface. `Launcher` isn't really intended to be implemented by clients. I think we could add the additional method with a default implementation that throws an exception to formally stay binary compatible.

marcphilipp:

PR #1057, opened 2017-09-11
Part II - Add module org.junit.platform.commons.jpms

sormuras:
## Overview

This PR introduces `org.junit.platform.commons.jpms` as the default `ModuleClassFinder` implementation. See #1061 for details.

- [x] Revise directory layout according to https://guides.gradle.org/building-java-9-modules/
- [ ] Get a Gradle guru to review changes made to the build process - especially conditional `include`

---

I hereby agree to the terms of the JUnit Contributor License Agreement.

---

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Method [preconditions](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [ ] [Coding conventions](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [ ] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#tests)
- [ ] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/meta/API.html)
- [ ] Change is documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [ ] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #1061, opened 2017-09-13
Part I - Introduce ModuleUtils with ClassFinder SPI

sormuras:
## Overview

This PR prepares the support for selecting Java 9 modules for test discovery.

Basic idea: publish a `String`-based service providing interface for naming module(s) of interest and let an explicit module register a service implementation for actually finding classes in the module graph at runtime. This service implementation should be hosted in a different (sub-) project, like `junit-platform-commons-jpms`. See #1057 

At runtime, using Java 9, the user can add the `org.junit.platform.commons.jpms` module to the module-path and specify which module he wants to test. The `mods` directory contains the JUnit 5 artifacts (loaded as automatic modules), the `org.junit.platform.commons.jpms` module, the application modules (tested objects), and the test modules (classes with `@Test` annotated methods):

```
java
--module-path
  mods
--add-modules
  ALL-MODULE-PATH
--module
  org.junit.platform.console
--select-module
  jpms.integration
```

## New Console Launcher Options

- `--scan-module-path`
Scan all modules on the runtime boot module-path for test discovery.
See http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-September/013180.html for details.

- `-o` or `--select-module` with single argument `<module name>`
Select single module for test discovery. This option can be repeated.
Same runtime boot module-path as above, but only the test classes of the named modules are selected.

## TODO

- [ ] Add an option handle to (console) launcher that parses an list of paths like `--scan-module-path [entry1[:entry_n]]` and delegates to the `ModuleFinderSelector`
See http://download.java.net/java/jdk9/docs/api/java/lang/module/ModuleFinder.html

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ModuleClassFinder.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ModuleClassFinder.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.util;
+
+import static org.apiguardian.api.API.Status.INTERNAL;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+import org.apiguardian.api.API;
+
+/**
+ * Class finder service providing interface.
+ *
+ * <h3>DISCLAIMER</h3>
+ *
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.1
+ */
+@API(status = INTERNAL, since = "1.1")
+public interface ModuleClassFinder {
+
+	/**
+	 * Special name indicating a search for all classes in all modules on the module-path.
+	 */
+	String ALL_MODULE_PATH = "ALL-MODULE-PATH";
+
+	/**
+	 * Return list of classes found in the named module that may contain testable methods.
+	 *
+	 * @param moduleName name of the module to inspect or {@code ALL-MODULE-PATH}
+	 * @param classTester filter to apply to each class instance
+	 * @param classNameFilter filter to apply to the fully qualified class name
+	 * @return list of test classes
+	 *
+	 * @see #ALL_MODULE_PATH
+	 */
+	List<Class<?>> findAllClassesInModule(String moduleName, Predicate<Class<?>> classTester,
+			Predicate<String> classNameFilter);
+}
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ModuleClassFinder.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ModuleClassFinder.java
@@ -12,8 +12,8 @@ package org.junit.platform.commons.util;
 
 import static org.apiguardian.api.API.Status.INTERNAL;
 
+import java.nio.file.Path;
 import java.util.List;
-import java.util.function.Predicate;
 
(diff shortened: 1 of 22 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ModuleClassFinder.java
```diff
@@ -0,0 +1,51 @@
...
+
+	/**
+	 * Special name indicating a search for all classes in all modules on the module-path.
+	 */
+	String ALL_MODULE_PATH = "ALL-MODULE-PATH";
+
+	/**
+	 * Return list of classes found in the named module that may contain testable methods.
+	 *
+	 * @param moduleName name of the module to inspect or {@code ALL-MODULE-PATH}
```
marcphilipp: I think having two separate methods, e.g. `findAllClassesInModule()` and `findAllClassesInModulePath()`, would be a cleaner. Thoughts?

junit-platform-console/src/main/java/org/junit/platform/console/options/AvailableOptions.java
```diff
@@ -150,6 +152,16 @@
...
+		// --- Java Platform Module System -------------------------------------
+
+		scanModulePath = parser.acceptsAll(asList("scan-modulepath", "scan-module-path"), //
+			"EXPERIMENTAL: Scan all modules on the module-path for test discovery.");
+
+		selectedModules = parser.acceptsAll(asList("o", "select-module"), //
+			"EXPERIMENTAL: Select single module for test discovery. This option can be repeated.") //
+				.withRequiredArg() //
+				.describedAs("module name");
+
```
marcphilipp: Should we also add support in the Vintage engine?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ModuleClassFinder.java
sormuras: Having two methods makes sense here, because we have two different selectors in the launcher part. The blue-print where I took the special module name `"ALL-MODULE-PATH"` from only has a single command line option, namely: `--add-modules`. There is no `--add-all-module-path` or in our world: `--scan-module-path`.

tl;dr: Two separate methods are good - and remove the special module name `"ALL-MODULE-PATH"`.

junit-platform-console/src/main/java/org/junit/platform/console/options/AvailableOptions.java
sormuras: Mh? Vintage should support both new selectors. If it doesn't then something was lost in rebasing to master.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ModuleUtils.java
```diff
@@ -0,0 +1,84 @@
...
+			Predicate<String> classNameFilter) {
+		Preconditions.notBlank(moduleName, "module name must not be null or blank");
+		Preconditions.notNull(classTester, "class tester must not be null");
+		Preconditions.notNull(classNameFilter, "class name filter must not be null");
+
+		ClassLoader classLoader = ClassLoaderUtils.getDefaultClassLoader();
+		List<Class<?>> classes = new ArrayList<>();
+
+		logger.config(() -> "Loading auto-detected class finders...");
+		int serviceCounter = 0;
```
sormuras: We could limit the single service implementation to be our own ... by comparing the class name or some other "magic ID". Or is it okay to allow third-party implementations here? Thoughts?

junit-platform-console/src/main/java/org/junit/platform/console/options/AvailableOptions.java
sormuras: Find vintage-affecting changes below, around https://github.com/junit-team/junit5/pull/1061/files#diff-5e437a25c16596592cf66086916f6fed

marcphilipp: Sorry, missed that! LGTM!

sormuras:
Superseded by #1087

PR #1078, opened 2017-09-27
Introduce ClassFilter, an internal predicate wrapper

### LINKED ISSUE #1077: Introduce ClassFilter for use in reflection utilities
## Overview

Many methods in classes and interfaces of package `org.junit.platform.commons.util` are using the two parameters `classNameFilter `and `classTester` in their signature their like:
```
T method(... Predicate<Class<?>> classTester, Predicate<String> classNameFilter, ...)
```
In most cases the actual values are just passed down the call-stack.

This issue proposes to introduce a wrapper named `ClassFilter` or `ClassTester` that could look like:

```java
@API(status = INTERNAL, since = "1.1")
public class ClassFilter {

	public static ClassFilter of(Predicate<Class<?>> classPredicate, Predicate<String> namePredicate) {
		Preconditions.notNull(classPredicate, "class predicate must not be null");
		Preconditions.notNull(namePredicate, "name predicate must not be null");

		return new ClassFilter(classPredicate, namePredicate);
	}

	private final Predicate<Class<?>> classPredicate;
	private final Predicate<String> namePredicate;

	private ClassFilter(Predicate<Class<?>> classPredicate, Predicate<String> namePredicate) {
		this.classPredicate = classPredicate;
		this.namePredicate = namePredicate;
	}

	public boolean test(Class<?> type) {
		return classPredicate.test(type);
	}

	public boolean test(String name) {
		return namePredicate.test(name);
	}
}
```

With such a type, the internal method signature will reduce to:
```
T method(... ClassTester classTester, ...)
```
and if we ever need to add another predicate, it can be done in the wrapper type w/o changing the signature of the internal methods.

## Deliverables

- [ ] Refactor internal methods to use the new class tester wrapper type

sormuras:
## Overview

Prior to this PR many methods in classes and interfaces of package `org.junit.platform.commons.util` were using the two parameters `classNameFilter` and `classTester` in their signatures their like:

    ..., Predicate<Class<?>> classTester, Predicate<String> classNameFilter, ...

These two parameters can now be put into a ClassFilter instance. This will reduce many method signature parts down to

    ..., ClassFilter classFilter, ...

and provides a future-proof way of extending the class filter implementation without changing the method signatures again.

Closes #1077 

---

---

### Definition of Done

- Non-public change is **not** documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassFilter.java
```diff
@@ -0,0 +1,64 @@
...
+@API(status = INTERNAL, since = "1.1")
+public class ClassFilter implements Predicate<Class<?>> {
+
+	public static ClassFilter of(Predicate<Class<?>> classPredicate) {
+		return of(name -> true, classPredicate);
+	}
+
+	public static ClassFilter of(Predicate<String> namePredicate, Predicate<Class<?>> classPredicate) {
+		Preconditions.notNull(namePredicate, "name predicate must not be null");
+		Preconditions.notNull(classPredicate, "class predicate must not be null");
```
marcphilipp: I think these precondition checks should be moved to the constructor.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassFilter.java
```diff
@@ -0,0 +1,64 @@
...
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.1
+ */
+@API(status = INTERNAL, since = "1.1")
+public class ClassFilter implements Predicate<Class<?>> {
+
+	public static ClassFilter of(Predicate<Class<?>> classPredicate) {
```
marcphilipp: Can we make the static factory methods more descriptive, e.g. `acceptAllNamesButFilterClasses()` and `filterNamesAndClasses()` or similar?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassFilter.java
sormuras: Moved via https://github.com/junit-
team/junit5/pull/1078/commits/1a6623b6d27595ebbc5105a869554ca8d896298a

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassFilter.java
sormuras: Would a javadoc comment (stating the obvious) suffice? ;)

junit-vintage-engine/src/test/java/org/junit/vintage/engine/discovery/VintageDiscovererTests.java
```diff
@@ -59,7 +59,7 @@ void logsDebugMessageWhenFilterExcludesClass(LogRecordListener listener) {
 			.map(LogRecord::getMessage)
 			.filter(m -> m.equals("Class " + Bar.class.getName() + " was excluded by a class filter: no match"))
 			.count()
-		).isEqualTo(1);
+		).isEqualTo(0); // TODO Should be 1 -- no logging is done or recorded anymore?!
```
sormuras: Expected log message is not here ... the assertion before this check is met.

- Either the log message is not created, because the logging-filter is not applied.
- Or the log is at a different level.
- Or the log message is different.
- Or the tracker is not working here.

sormuras: The issue can be _fixed_ by replacing the real name predicate with `name -> true` in `VintageDiscoverer`:

	private ClassFilter createClassFilter(EngineDiscoveryRequest request) {
		return ClassFilter.of(name -> true, createTestClassPredicate(request));
	}

That would bypass `ClasspathScanningSupport.buildClassNamePredicate` though...

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/VintageDiscoverer.java
```diff
@@ -56,31 +47,24 @@ public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId
...
-			new UniqueIdSelectorResolver()//
+	private List<DiscoverySelectorResolver> getAllDiscoverySelectorResolvers() {
+		return asList(
+		// @formatter:off
+			new ClasspathRootSelectorResolver(),
+			new PackageNameSelectorResolver(),
+			new ClassSelectorResolver(),
+			new MethodSelectorResolver(),
+			new UniqueIdSelectorResolver()
+		// @formatter:on
```
marcphilipp: Since these are now all stateless, we could change the list into a constant, right?

sormuras: Sure. Stored in final immutable field?

sbrannen: > Since these are now all stateless, we could change the list into a constant, right?

Good catch!

> Sure. Stored in final immutable field?

Yep

sbrannen:
Regarding your latest "Remove debug level logging from Vintage Engine for excluded classes" commit, is this information no longer being logged at all?

sormuras:
Yes, Marc proposed to remove it for good. It was part of "debugging" the initial Vintage Engine implementation.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/discovery/VintageDiscoverer.java
sormuras: Where is `List.of(...)` when you need it?! ;)

Issue #1080, opened 2017-09-28
Add --scan-module-path support


sormuras:
## Overview

Umbrella issue for adding `--scan-module-path` support to the JUnit Platform.

## Related Issues

- #1087 - [Multi-release JAR](http://openjdk.java.net/jeps/238) with `JigsawUtils` and module-path scanning support
- https://github.com/junit-team/junit5-samples/pull/36 - Sample running on the module-path

## Outdated Pull Requests

- Part I #1061 - SPI `ModuleClassFinder`
- Part II #1057  - Default (single) implementation

## Deliverables

- [x] Make `--scan-module-path` work
- [X] ~Add `--scan-module-path` and `--select-module` support to Gradle Plugin~ will be done in #1119
- [X] ~Add `--scan-module-path` and `--select-module` support to Maven Surefire Plugin~

marcphilipp:
Closed in favor of #1119.

PR #1087, opened 2017-09-30
Add "--scan-module-path" support on Java 9

sormuras:
## Overview

Enhance `junit-platform-commons` to a multi-release jar project: http://openjdk.java.net/jeps/238

Addresses #1080 

## Tasks
- [x] `--select-module <name>` selector and console option
- [x] `--scan-module-path` console option
- [x] Jupiter supports new selector
- [x] Include all `META-INF/versions/x/...` binaries in standalone JAR 
- [x] Add tests for multi-release feature
- [x] Vintage supports new selector
- [x] `--release 8/9` for all subprojects (party addressed by #1097)
- [x] after merging #1097, remove the `if (JavaVersion.current().isJava9Compatible())` condition from `settings.gradle`.
- [x] exclude `junit-platform-commons-java-9` from javadoc
- [x] configure/test build on Jenkins (cloudbees)
- [x] add functional test for `ModuleUtils.findAllClassesInModule("java.base")` that actually finds and loads a class
---

---

### Definition of Done


junit-platform-commons-java-9/junit-platform-commons-java-9.gradle
```diff
@@ -0,0 +1,70 @@
...
+// mounted as an automatic jar into the integration test module-path.
+task generateIntegrationTestsJar(type: Jar, dependsOn: compileTestJava) {
+	archiveName 'junit-commons-integration-tests.jar'
+	from compileTestJava.destinationDir
+	include("integration/**/*")
+}
+
+// Copy runtime dependencies to a single directory.
+task generateDependenciesDirectory(type: Copy) {
+	into buildDir.toPath().resolve('deps').toString()
```
marcphilipp: `into "$buildDir/deps"`?

junit-platform-commons-java-9/src/main/java/org/junit/platform/commons/util/JigsawUtils.java
```diff
@@ -0,0 +1,159 @@
...
+ * <h3>DISCLAIMER</h3>
+ *
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.1
+ */
+@API(status = INTERNAL, since = "1.1")
+public class JigsawUtils {
```
marcphilipp: This class is the only mention of "Jigsaw" in the codebase. I'm wondering whether `ModuleUtils` would be a better name... Thoughts? 🤔

settings.gradle
```diff
@@ -6,6 +6,7 @@ include 'junit-jupiter-engine'
 include 'junit-jupiter-migrationsupport'
 include 'junit-jupiter-params'
 include 'junit-platform-commons'
+if (JavaVersion.current().isJava9Compatible()) include 'junit-platform-commons-java-9'
```
marcphilipp: Shall we make this mandatory right away?

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -241,6 +242,21 @@ public static ClasspathResourceSelector selectClasspathResource(String classpath
 	}
 
 	/**
+	 * Create a {@code ModuleSelector} for the supplied module name.
+	 *
+	 * <p>The unnamed module is not supported.
+	 *
+	 * @param moduleName the module name to select; never {@code null} and
+	 * never blank
```
marcphilipp: I think we usually use `never {@code null} or blank`.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
```diff
@@ -55,6 +56,12 @@ LauncherDiscoveryRequest toDiscoveryRequest(CommandLineOptions options) {
 				"Scanning the classpath and using explicit selectors at the same time is not supported");
 			return createClasspathRootSelectors(options);
 		}
+		if (options.isScanModulepath()) {
+			Preconditions.condition(!options.hasExplicitSelectors(),
+				"Scanning the module-path and using explicit selectors at the same time is not supported");
+			return Collections.singletonList(DiscoverySelectors.selectModule("ALL-MODULES"));
+
```
marcphilipp: Can we reference the constant here? There's an extra blank line after the return.

junit-platform-commons-java-9/junit-platform-commons-java-9.gradle
sormuras: Well. Yes.

junit-platform-commons-java-9/src/main/java/org/junit/platform/commons/util/JigsawUtils.java
sormuras: I intend to refactor the `eval()/invokeGetters()` logic used in `TestEngine` to a call to `JigsawUtils`. Something like `JigsawUtils.getModuleName(Class<?> type)` and `JigsawUtils.getModuleVersion(Class<?> type)` returning `Optional<String>` of course.

settings.gradle
sormuras: And ditch JDK 8?

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
sormuras: Copy and paste error by me.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
sormuras: The constant is declared "private" in `JigsawUtils (9)`. For it to be used is must use `public` and present in both `JigsawUtils` versions.

junit-platform-commons-java-9/src/main/java/org/junit/platform/commons/util/JigsawUtils.java
marcphilipp: Does that speak against `ModuleUtils`?

settings.gradle
marcphilipp: Well, we'll have to build on JDK 9 once we merge this PR to build correct artifacts, don't we?

junit-platform-commons-java-9/src/main/java/org/junit/platform/commons/util/JigsawUtils.java
sormuras: Looking at the proposed names ... no. Not yet. :)

settings.gradle
sormuras: True.

The major source and target compatibility stays on `1.8`. Only selected subprojects need `1.9`. Mh, should work.

sormuras: Jenkins CI builds need to be suspended until they support JDK 9.

sormuras: Travis CI build matrix can be reduced to JDK 9 only.

sormuras: `junit5-samples` should stay as they are, meaning most of the sample run on JDK 8 and only some on JDK 9.

marcphilipp: I just opened a support request with CloudBees to get JDK 9 support. Let's hope they respond soon.

sormuras: Nit: current IDE support for JDK-9 is still moving. E.g. IDEA sometimes fails to "Refresh all Gradle projects" with a "MODULE" not supported error. Eclipse might have similar ... challenges.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
sormuras: FYI, the `jar --update ... --release X` checks for some (all?) binaries API differences of types that override type in base. It fails to update the artifact when there's an issue.

sormuras:
The only "real" `MethodSelector` test is integrated into the Gradle build file:
```
// Execute console launcher on the module-path.
task testScanModulepath(type: Exec, dependsOn: [generateDependenciesDirectory, generateIntegrationTestsJar]) {
	executable = 'java'
	args = [
			'--module-path', files(
					generateDependenciesDirectory.destinationDir,
					generateIntegrationTestsJar.archivePath
				).asPath,
			'--add-modules', 'ALL-MODULE-PATH',
			'--module', 'org.junit.platform.console',
			'--scan-module-path'
	]
}

[...]

:junit-platform-commons-java-9:generateDependenciesDirectory UP-TO-DATE
:junit-platform-commons-java-9:compileTestKotlin
:junit-platform-commons-java-9:compileTestJava UP-TO-DATE
:junit-platform-commons-java-9:generateIntegrationTestsJar UP-TO-DATE
:junit-platform-commons-java-9:testScanModulepath
.
+-- JUnit Vintage [OK]
| '-- integration.VintageIntegrationTest [OK]
|   '-- successfulTest [OK]
'-- JUnit Jupiter [OK]
  '-- JupiterIntegrationTests [OK]
    +-- version() [OK]
    +-- moduleIsNamed() [OK]
    '-- packageName() [OK]
```

I could add assertions that test for strings in the produced tree.

settings.gradle
sormuras: See https://stackoverflow.com/questions/45899027/errorbug-exception-in-phase-class-generation-in-source-unit-buildscript

sormuras: When we do switch to JDK-9 as our build platform, we should also provide _real_ `module-info.class` as `--release 9` assets. At least for `org.junit.platform.commons` to export only those packages we want to used by external code.

sormuras: Opened https://github.com/junit-team/junit5/issues/1091 to address the _real_ Module Descriptor.

sormuras:
Compare Gradle "multi-release" build configuration to the one presented here: https://melix.github.io/javaone-2017-jigsaw/#/multi-release-jars

sormuras:
The solution presented by @melix is not well suited for our multi-project build.
- Having a separate subproject for the Java-9 based code allows IDEs to treat that subproject with the source/target level as 9. All others stay on 8.
- Patching/updating the base/shared JAR artifact in a post-step using the "[jar](https://docs.oracle.com/javase/9/tools/jar.htm#JSWOR614)" tool allows API cross-checks to be applied. Those checks assure that the overloaded types are API-compatible with the versioned ones.
- Testing the integration of the MRJAR against various runtimes is easier to maintain in separate project -- as it doesn't interfer with the basic build process.

melix:
> Having a separate subproject for the Java-9 based code allows IDEs to treat that subproject with the source/target level as 9. All others stay on 8.

IntelliJ IDEA supports module per source-set. I need to check with @oehme but I think Eclipse also. Using a separate subproject is overkill and will never scale. A separate source-set is lighter and more correct modeling-wise (it's the *same* project). I would never encourage using a different subproject because of this.

> Patching/updating the base/shared JAR artifact in a post-step using the "jar" tool allows API cross-checks to be applied. Those checks assure that the overloaded types are API-compatible with the versioned ones.

I reckon that the API check that the `jar` task is doing is useful, but that's one of the reasons MRJars are no good idea: there's absolutely *nothing* that enforces that the APIs *must* be the same. This means that tools like Maven, Gradle, Ant, or whatever can generate wrong jars and they would still be valid. To me it means Gradle needs to implement the check, but it's yet not a good reason for patching the jar using `jar`.

The second reason is that you **must never** patch a jar generated by Gradle, or, any output generated by a task. This kills up-to-date checking, and breaks caching. You're compromising the performance of your build there by making the task always out-of-date for no reason. Either create a separate jar, or do this in the `doLast` block of the `jar` task itself so that it's a single unit of work.

> Testing the integration of the MRJAR against various runtimes is easier to maintain in separate project -- as it doesn't interfer with the basic build process.

This is a different issue, that is also going to be solved differently. In a nutshell, it's not a new problem. When you use reflection instead of mrjars, you still have the problem: how to you test that all versions of your class are properly tested? We want to model this properly by allowing the tests to be executed on various target platforms. By separating your work in a distinct project, you will make this unapplicable. I'm going to demonstrate how to properly implement cross-platform testing soon in a blog post.

Eventually, one of the issues with multi-release jars is that they force you to create a jar. Before, there was no need to create a jar to run tests, or run the application: using class directories and resource-directories was enough. Now, the semantics of your class depend on whether they are packaged in a jar or not. This is horrible. And it's also a killer for performance, because we're forced to produce a jar where we don't need: this means that dependents will need to wait for a jar to be available, instead of having to just wait for the _classes_ to be available. This may sound strange, but for larger projects it's a performance hog: this means that we cannot build the jar and run tests concurrently, for example.

Anyway, I still think the separate source-set approach is the way to go.

sormuras:
Being forced to create a jar before depending on the project is good thing imho. That jar can now serve as the only and single point of thruth for the build and for deployment. If your multi-threaded tasks wait for "classes" or "classes+jar" to finish should not be a major impact. I guess. Am I wrong here?

sormuras:
I did hook into the jar task with a `doLast` closure. The build stays up-to-date with it.

See https://github.com/junit-team/junit5/pull/1087/files#diff-c197962302397baf3a4cc36463dce5eaR440 and below.

melix:
> Being forced to create a jar before depending on the project is good thing imho. That jar can now serve as the only and single point of truth for the build and for deployment. If your multi-threaded tasks wait for "classes" or "classes+jar" to finish should not be a major impact. I guess. Am I wrong here?

This is not just true when you have inter-project dependencies. A consumer only requires the class files (and, in fact, only the ABI of those class files). So if you build the jar, you're just blocking downstream consumers from compiling. This may be not an issue for you, it is for larger projects.

> I did hook into the jar task with a doLast closure. The build stays up-to-date with it.

I see. Nevertheless you're invoking `jar` from an external process here, which will have an impact on performance (not reusable). Again this may be ok for your project, but we cannot assume a general rule from this.

settings.gradle
sormuras: When #1097 is merged, I'll rebase this PR and remove the `if (JavaVersion.current().isJava9Compatible())` condition.

sormuras:
> Nevertheless you're invoking jar from an external process here, which will have an impact on performance (not reusable).

A local version used the http://download.java.net/java/jdk9/docs/api/java/util/spi/ToolProvider.html to avoid launching an external process. But as you said, in our project it's a small impact to spawn another VM. Might change if we introduce a `-java-9` overlay project for each artifact. ;)

PR #1090, opened 2017-10-03
Display name for @ParameterizedTest should only include consumed arguments

karollewandowski:
Redundant arguments are filtered out in `ParameterizedTestExtension`, so only consumed arguments are passed to invocation context and used in name formatter. Filtering is applied only when number of arguments is greater than number of test method parameters. It should work with other extensions like parameter resolvers - when there are eg. 3 columns provided and only 2 test method parameters consumes them and the 3rd one is resolved by another extension, then exception about incorrect configuration will be thrown.

Please notice that some unchanged code in `ParameterizedTestIntegrationTests.java` is reformatted. I used formatter from `junit5/src/eclipse` and assume it was formatted incorrectly before. If it was correct, please let me know and I'll revert those changes.


marcphilipp:
Please run `./gradlew spotlessApply` and add another commit. 🙂

karollewandowski:
Thanks. Formatting fixed.

marcphilipp:
Thanks, @karollewandowski! 👍

Issue #1126, opened 2017-10-25
surefire provider should accept empty string for includeTags and excludeTags


tmurakam:
## Overview

The junit-platform-surefire-provider does not accept empty string for includeTags and excludeTags properties.

It is inconvenient, because I can't  use default value in pom.xml and overwrite it with -D option. 

## Detail

When I pass empty string to includeTags/excludeTags, following exception occurs:

```
Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test failed: There was an error in the forked process
[ERROR] org.apache.maven.surefire.util.SurefireReflectionException: org.junit.platform.commons.util.PreconditionViolationException: Tag name [] must be syntactically valid
[ERROR] 	at org.apache.maven.surefire.util.ReflectionUtils.instantiateOneArg(ReflectionUtils.java:135)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.createProviderInCurrentClassloader(ForkedBooter.java:321)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:290)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:242)
[ERROR] 	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:121)
[ERROR] Caused by: org.junit.platform.commons.util.PreconditionViolationException: Tag name [] must be syntactically valid
[ERROR] 	at org.junit.platform.commons.util.Preconditions.condition(Preconditions.java:285)
[ERROR] 	at org.junit.platform.engine.TestTag.<init>(TestTag.java:120)
[ERROR] 	at org.junit.platform.engine.TestTag.create(TestTag.java:116)
[ERROR] 	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
[ERROR] 	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
[ERROR] 	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
[ERROR] 	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
[ERROR] 	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
[...]

PR #1136, opened 2017-11-02
IncludeTags and excludeTags accept empty strings

VikasSherawat:
Issue: #1126

## Overview

IncludeTags and ExcludeTags are modified to accept empty strings as input parameter.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes-5.1.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes-5.1.0-M1.adoc
@@ -41,6 +41,7 @@ on GitHub.
 * New console launcher option `--scan-module-path` for selecting all resolved modules
   available on the boot layer configuration.
 * Maven Surefire: Add support for the `redirectTestOutputToFile` feature.
+* Maven Surefire: Add support to accept empty strings in `@IncludeTags` and `@ExcludeTags`.
 
 [[release-notes-5.1.0-junit-jupiter]]
 ==== JUnit Jupiter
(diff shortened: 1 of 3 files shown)


sbrannen:
I'm not convinced that we should remove the "not empty" validation entirely from the Platform.

Rather, I think it may be better to "ignore" (and not register) empty tags within the Surefire Provider and Gradle plugin.

sbrannen:
In any case, it does **not** make sense to actually register an empty tag.

So we cannot accept the PR _as is_.

sbrannen:
If an empty tag is supplied and ignored, the Platform should log such cases.

marcphilipp:
> Rather, I think it may be better to "ignore" (and not register) empty tags within the Surefire Provider and Gradle plugin.

I quite agree. I think this should be done in `JUnitPlatformProvider` instead.

VikasSherawat:
@sbrannen @marcphilipp 
Thanks for the valuable comments.!! I'll try to understand JUnitPlatformProvider class and make changes accordingly.

marcphilipp:
I think it boils down to changing the following line to use `StringUtils.isNotBlank()` and writing tests for it.

https://github.com/junit-team/junit5/blob/master/junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/JUnitPlatformProvider.java#L185

VikasSherawat:
@marcphilipp @sbrannen did i address the issue in right way?

VikasSherawat:
@marcphilipp should i document these changes in the Release notes?

marcphilipp:
@VikasSherawat Yes, that would be great!

marcphilipp:
Thanks, @VikasSherawat!

Issue #456, opened 2016-08-11
Introduce `@BeforeSuite` and `@AfterSuite` annotations


ondrejlerch:
I suggest that `@BeforeSuite` and `@AfterSuite` annotations should be added to JUnit 5.

Obviously they should execute before and after the whole suite.

Or is there some other mechanism to accomplish that?

I suggest that there should not be static restriction, i.e. also non-static methods could be annotated with `@BeforeSuite` and `@AfterSuite`.

I personally prefer non-static as we use Spring in our tests and want to use injected beans in `@BeforeSuite` method.

smoyer64:
Related to #163.

ttddyy:
Related to #19 
(e.g.: _@BeforeAllTests/@AfterAllTests_)

ondrejlerch:
In my consideration "Suite" can be defined via:
- Maven/Gradle plugin inclusion/exclusion
- JUnitPlatform declarative declaration
- IDE when I click Run on the whole package of tests

In all those cases I would expect that @BeforeSuite/@AfterSuite would be executed just once before/after the whole bunch of tests, for example to perform time consuming operation such as database cleanup.

sebersole:
I agree with everything @ondrejlerch brings up

marcphilipp:
@ondrejlerch How would JUnit find `@BeforeSuite`/`@AfterSuite` annotated methods? For example, when you run a test class from an IDE would it only look at the test class or someplace else?

The `Store` abstraction along with the root `ExtensionContext` provides a means to prepare something once and share it between multiple test classes. However, there's no way to clean it up at the moment (that depends on #742).

sebersole:
@marcphilipp you could consider a solution such as Jandex, which provides you an index over all annotations

sormuras:
https://github.com/wildfly/jandex ?

marcphilipp:
Well, in any case, it has to be obvious for test authors where JUnit will actually look for these annotations, i.e. if they will be applied when they execute a test class.

sormuras:
```
environment (shell/build tool/IDE/...)  (1)
  platform/launcher (DefaultLauncher)   (2)
    engine a (Jupiter)                 (3a)
    ...
    engine x                           (3x)
    engine z                           (3z)
```

I see an easy way to implement an extension point for **3a** -- which frames the all tests executed by the Jupiter engine with a "before the first" and "after the last" callback.

[...]

PR #1143, opened 2017-11-05
Introduce before and after callbacks at the Jupiter engine-level

sormuras:
## Overview

Addresses #456 

This PR is a basis for discussion. Error handling and documentation are left out on-purpose.

### Usage
```java
@BeforeEngineExecution
static void before0() {
	countBefore0++;
}

@BeforeEngineExecution
static void before0(ExtensionContext context) {
	context.getStore(ExtensionContext.Namespace.GLOBAL).put("alpha", "omega");
	countBefore0++;
}

@AfterEngineExecution
static void after0(ExtensionContext context) {
	assertEquals("omega", context.getStore(ExtensionContext.Namespace.GLOBAL).get("alpha"));
	countAfter0++;
}

@AfterEngineExecution
void after0NonStatic(ExtensionContext context) {
	assertEquals("omega", context.getStore(ExtensionContext.Namespace.GLOBAL).get("alpha"));
	countAfter0++;
}
```

### Implementation notes

* If a single parameter `ExtensionContext` is declared in an annotated method, the root `ExtensionContext` of `JupiterEngineExecutionContext` is injected at runtime.
* `@BeforeEngineExecution`-annotated methods must be `static`.
* `@AfterEngineExecution`-annotated methods must be `static`. If you're using `@TestInstance(TestInstance.Lifecycle.PER_CLASS)`, `@AfterEngineExecution`-annotated methods don't have be `static`.
* ~~`@ExtendWith(AlphaOmega.class)` does **not** work at the moment.~~

---

---

### Definition of Done

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,4 +1,4 @@
 jdk: oraclejdk9
 
 install:
-  - ./gradlew install -x test
+  - ./gradlew install -x test --no-daemon --stacktrace
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterEngineExecution.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AfterEngineExecution.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2015-2017 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api;
+
+import static org.apiguardian.api.API.Status.STABLE;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+
+/**
+ * TODO ...
+ */
+@Target({ ElementType.ANNOTATION_TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@API(status = STABLE, since = "5.1")
+public @interface AfterEngineExecution {
+}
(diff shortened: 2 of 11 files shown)


sebersole:
Sorry if I missed it, but how are these picked up?  Can they exist anywhere (i.e. across multiple test classes)?

sormuras:
As implemented in this PR, they are discovered in every class that "contains" tests.

But we might drop the entire idea and this PR in particular in favor for a broader solution like discussed in #744

sebersole:
I should point out though that I am personally more interested in the corresponding extension corollary to these annotations - although I guess how they registered there would be questionable (service?).

sormuras:
If you want to play with this PR in your sample projects - direct your dependency resolver to jitpack, that host a compiled and packed version of this branch: 
https://jitpack.io/com/github/junit-team/junit5/issues~456-jupiter-engine-level-callbacks-SNAPSHOT/build.log

...almost. Jitpack (and others) seems to suffer from `server certificate change is restricted during renegotiation` issues.

sormuras:
Superseded by #1174

luolong:
Actually, I am rather keen on the idea of before- and after callbacks on the (Jupiter) test engine level.

My use-case is that the code base we are writing tests over is sometimes heavily time-dependent. We are using JodaTime for accessing current time in production code and JodaTime has a concept of a `MillisProvider` which is a global state that all methods that return current time use under the covers to get current milliseconds value.

To be able to improve stability with time dependent tests, we need to override this `MillisProvider` so that we could control time in our tests. That on the other hand is also problematic because if every test now overrides global time, we cannot run these tests in parallel as they may well start overriding the global state.

A solution would be to initialize custom `MillisProvider` with the _`TestEngine`_ that keeps separate time per thread using `ThreadLocal` context and only manipulate time through this test double API.

As a bonus, I could clean up and reset the global `MillisProvider` to the default one when the test engine has finished running all the tests.

luolong:
As for the discovery mechanism. 

Because this is such a deep integration with test engine that there is no good natural context where to define such classes, why not use _ServiceLoader_ for initializing test engine extensions?

sormuras:
If I understand your use-case correctly, it can be implemented similar to the example in #1174 -- on first access, create and store your `MillisProvider` at the engine level context and use that singleton instance throughout your tests. When the context store is closed, restore the global `MillisProvider` to the default.

You won't get the exact system instant millis the engine "started" with this approach, but you may refer to the instant your resource is created and stored.

luolong:
Yes, that could work.

Issue #955, opened 2017-07-17
Consider using static analysis and formal verification tools


jbduncan:
## Overview

Currently, AFAICT, no static analysis or formal verification tools are being used to catch bugs in the JUnit 5 code base. IMO it would be a sensible thing to do to start using a combination of various tools now, so that when we reach GA and proceed beyond GA, we are less likely to be inundated by preventable bug reports and/or suffer from potential security issues.

I'd personally suggest a combination of the following tools, but I'm more than happy to discuss the merits of the tools listed and to discuss other tools which aren't listed or that I may have not thought of:
- [error-prone](http://errorprone.info/)
- [SpotBugs](https://spotbugs.github.io/) or FindBugs
- SpotBugs/FindBugs plugins:
    - [fb-contrib](http://fb-contrib.sourceforge.net/)
    - [find-sec-bugs](http://find-sec-bugs.github.io/)
- PMD
- [Checker Framework](https://checkerframework.org/) (specifically for [nullness formal verification](https://checkerframework.org/manual/#nullness-checker))

## Deliverables

- [ ] A number of agreed-upon static analysis and/or formal verification tools are adopted.

sormuras:
Would you mind filing a PR with the tool(s) of your choice?

alixwar:
SonarQube exists in a cloud version, free for use by Open Source projects: https://about.sonarcloud.io/

SonarQube includes support for PMD, FindBugs and more

jbduncan:
@alixwar We could indeed use SonarQube in the future, but I'd feel leery about us primarily depending on it considering that Gradle already comes with first- and third-party support for SpotBugs (FindBugs' current successor), PMD, and other great tools like error-prone.

jbduncan:
...and also considering that SonarQube's cloud version is yet another thing which we don't have a lot of control over.

alixwar:
@jbduncan Maybe one solution does not exclude the other? My thinking is that bootstrapping SonarQube is probably very quick and easy (based on my experience setting it up locally)

jbduncan:
You make a good point. I'm most likely dismissing SonarQube out of fear too quickly. :)

I personally have no experience with SonarQube, so it's not something I could setup unless I get a lot of free time and develop the inclination to work on it in the future, so I wouldn't mind at all if you wanted to have a stab at it yourself at some point. :)

alixwar:
[...]

PR #1214, opened 2017-12-21
A POC for replacing explicit parameter null-check by an automated solution

denis-zhdanov:
## Overview

I'd like to contribute to *junit* by replacing existing *Preconditions.notNull()* checks with an automated solution based on *NotNull* method parameters annotations.  

The job is done by the [Traute javac plugin](http://traute.oss.harmonysoft.tech/) (I'm its author) and current *PR* illustrates how it works:  

1. *Traute* is configured for the *junit* project
2. Target method parameters are marked by *NotNull* annotation (a number of annotations are [supported by default](http://traute.oss.harmonysoft.tech/core/javac/#71-notnull-annotations), I used jsr305's *Nonnull* in this *PR)
3. Existing *Preconditions.notNull()* checks for the annotated method parameters are removed  

E.g. here we apply the steps above to the [AbstractExtensionContext](https://github.com/denis-zhdanov/junit5/commit/c4b77d03cea8570ebf6f3c77e58b69cf3f7ee96d#diff-47d6e0008ed4935f2c5d0811c80b4e01) class. Result:  

```
javap -c ./junit-jupiter-engine/build/classes/java/main/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.class
...
  public org.junit.jupiter.api.extension.ExtensionContext$Store getStore(org.junit.jupiter.api.extension.ExtensionContext$Namespace);
    Code:
       0: aload_1
       1: ifnonnull     14
       4: new           #18                 // class org/junit/platform/commons/util/PreconditionViolationException
       7: dup
       8: ldc           #19                 // String Namespace must not be null
      10: invokespecial #20                 // Method org/junit/platform/commons/util/PreconditionViolationException."<init>":(Ljava/lang/String;)V
      13: athrow
```  
Benefits:
* the code is better documented - method parameters constraints are clear either in javadocs or source
* IDE support - there are nullability inspections in eclipse and intellij which highlight possible problems during development (like an attempt to give a *potentially null* reference as a *NotNull* method parameter)
* source code is cleaner - explicit checks are removed

Also note *Traute* offers two more feature which might be useful for *junit*:
* method return type might also be marked as *NotNull*. The plugin inserts a check for the return value then
* it's possible not only mark all target method parameters by *NotNull*, but define *NotNullByDefault* and mark excludes by *Nullable* (if any) - see [NotNullByDefault](http://traute.oss.harmonysoft.tech/core/javac/#72-notnullbydefault-annotations) and [Nullable](http://traute.oss.harmonysoft.tech/core/javac/#73-nullable-annotations)  

Please let me know if the team is interested in this change, I'm ok to apply the change to the whole project then.
---

---

### Definition of Done


jbduncan:
I cannot speak for the JUnit 5 team, but I'm personally interested to hear why Traute would be a better option than something like the [Checker Framework](https://github.com/typetools/checker-framework) or [uber/NullAway](https://github.com/uber/NullAway).

Please correct me if I've misunderstood something, but by my understanding so far, Traute auto-generates runtime null-pointer checks from `@NotNull` and `@Nullable` annotations, but the Checker Framework (which is thoroughly exhaustive in its checking) and NullAway (which is far less thorough but is faster) both do null checking at compile-time, which I believe make them better alternatives.

@denis-zhdanov I'd be interested your thoughts.

denis-zhdanov:
@jbduncan, *NullAway* and *Checker* are intended to be applied to 'application' projects to catch possible *NPE*s early.  

However, that's not the case for *junit* which is a library used in other projects, i.e. that 'host' projects should be configured to perform *null*-checks but we can't force users to do that.  

So, *Traute* enforces **library** contracts even when the library itself is used in another projects.   

(Not directly related to the original question) I've made a [sample project](https://github.com/denis-zhdanov/samples/tree/master/java/checker) which uses *Checker* and ran a couple of tests. It's not all the shiny:  
1. *Checker* didn't detect a possible *NPE* [here](https://github.com/denis-zhdanov/samples/blob/master/java/checker/src/main/java/org/denis/sample/java/checker/jdk/JdkClasses.java#L11)
2. False-positive error [occurred](https://github.com/denis-zhdanov/samples/blob/master/java/checker/src/main/java/org/denis/sample/java/checker/library/LibraryUsage.java#L14) during working with library files (we can't expect that all third-party libraries properly configure *NotNull*/*Nullable* for method parameters

smoyer64:
Related to #358 and #955.

sbrannen:
@denis-zhdanov, I have not looked at your plugin in detail; however, based on a quick glance at the documentation, it appears to me that Traute always throws a `NullPointerException` which **strictly violates** JUnit Platform policies.

We _never_ intentionally throw anything other than a `JUnitException` (or subclass thereof).

So, if Traute cannot be configured to throw a `PreconditionViolationException`, that's basically a showstopper for us.

jbduncan:
@denis-zhdanov Ah, I hadn't considered the distinction between applications and libraries in this case. Your argument makes a good deal of sense to me, so thanks for clearing things up for me.

I now believe that it would be a good idea to use both Traute and the Checker Framework in JUnit 5. The way I imagine it: Traute would be used to automatically throw `PreconditionViolationException`s at runtime for JUnit 5 users who themselves don't use Checker Framework/NullAway, whereas Checker Framework would be used to do compile-time null checks for internal JUnit 5 classes, and Checker Framework annotations would be used instead of `javax.annotation.Nullable` et al for Java 9 compatibility.

jbduncan:
@denis-zhdanov Regarding the [possible NPE](https://github.com/denis-zhdanov/samples/blob/master/java/checker/src/main/java/org/denis/sample/java/checker/jdk/JdkClasses.java#L11) you found in your Checker Framework example, I wonder if it's not detecting the NPE in that case because your project is not configured to use the Checker Framework's annotated version of the JDK (https://checkerframework.org/manual/#gradle).

By my understanding, the Checker Framework needs a version of the JDK pre-annotated with Checker Framework annotations to do its job effectively, because otherwise it assumes that all return values are non-null (which I believe to otherwise be a sane default). I believe that's why your example doesn't catch the possible NPE from `File.listFiles()`, since `File.listFiles()` isn't annotated with `@Nullable` in normal distributions of the JDK, and thus Checker Framework deduces incorrectly that `File.listFiles()` can never return `null`.

jbduncan:
@denis-zhdanov As for the [false-positive error](https://github.com/denis-zhdanov/samples/blob/master/java/checker/src/main/java/org/denis/sample/java/checker/library/LibraryUsage.java#L14) you found, I think it errors there because the Checker Framework follows what I believe is a sane default rule, where if a parameter, variable or field is not annotated with `@Nullable` then it's assumed that `null` is not a legal value. In other words, everything is treated by the Checker Framework as if it were annotated with `@NonNull` by default.

I'm sure that there's a way in the [manual](https://checkerframework.org/manual/) to get Checker Framework working with third-party APIs like `StringUtils.capitalize()` that accept or return `null`s but do not indicate so with `@Nullable`.

denis-zhdanov:
@sbrannen, thanks for looking into this. Regarding your concern - the plugin generates checks with *NullPointerException* by default but any other exception can be [configured](http://traute.oss.harmonysoft.tech/core/javac/#75-exception-to-throw).  

Moreover, current *POC PR* already [configures](https://github.com/denis-zhdanov/junit5/blob/traute-null-checks/build.gradle#L193) *Traute* to throw *org.junit.platform.commons.util.PreconditionViolationException* (in the same way as *Preconditions.notNull()*). Also note that custom exception text is also [configured](https://github.com/denis-zhdanov/junit5/blob/traute-null-checks/build.gradle#L198) to match *Preconditions.notNull()*

denis-zhdanov:
@jbduncan, yeah, I believe it's worth to have both static *null* protection or runtime *null*-checks.  

Thanks, will check the annotated *JDK*.  

I'm not so sure about the *not-null by default* processing in case of third-party libraries - there are many of them which do not use *NotNull*/*Nullable* annotations at all and also many which use other annotations (IntelliJ, Eclipse, jsr-305 etc). Current behavior (an error is reported in a situation when there is no error) doesn't look too attractive. I didn't dig too much into *Checker*'s documentation in order to find out if it's possible to configure that though.

jbduncan:
@denis-zhdanov I believe that Checker Framework recognises and type-checks against a number of third-party implementations of `@NotNull`/`@Nullable`-like annotations. A current list can apparently be found in the manual [here](https://checkerframework.org/manual/#nullness-related-work).

As for dealing with false-positives like Apache Commons Lang's `StringUtils.capitalize()`, I just found again the [chapter](https://checkerframework.org/manual/#annotating-libraries) in the manual which describes how to deal with unannotated third-party libraries.

Hope this helps! :)

marcphilipp:
There may be another advantage of this: null safety when used from Kotlin.

@sbrannen I know Spring introduced its own set of composed annotations for this. Can you elaborate on why that strategy was chosen?

sbrannen:
> There may be another advantage of this: null safety when used from Kotlin.
> 
> @sbrannen I know Spring introduced its own set of composed annotations for this. Can you elaborate on why that strategy was chosen?

@sdeleuze can best elaborate on that! 😉

sbrannen:
> Moreover, current *POC PR* already [configures](https://github.com/denis-zhdanov/junit5/blob/traute-null-checks/build.gradle#L193) *Traute* to throw *org.junit.platform.commons.util.PreconditionViolationException* (in the same way as *Preconditions.notNull()*). Also note that custom exception text is also [configured](https://github.com/denis-zhdanov/junit5/blob/traute-null-checks/build.gradle#L198) to match *Preconditions.notNull()*

@denis-zhdanov, ahhhh, OK. Thanks for the pointers!

sdeleuze:
@sbrannen @marcphilipp In a nutshell, we leverage JSR 305 capabilities to define null safety semantics without exposing it directly since it is a dormant and non official JSR (even if widely used in the ecosystem) which could be replaced at some point by a more official initiative. To achieve that, we annotate Spring null-safety annotations with JSR 305 meta-annotations. This is the best trade-off we found. This also allows us to define the scope not available in JSR 305 default annotations we need (for example `@NonNullApi` or `@NonNullFields`).

Current use cases are IDE checks for Java developers in IDEA/Eclipse and full null-safety for Kotlin developers, see more details in [this Kotlin KEEP documentation](https://github.com/Kotlin/KEEP/blob/jsr-305/proposals/jsr-305-custom-nullability-qualifiers.md).

Notice that automatic checks are not possible in Spring because we need to keep our API as it it, so we sometimes handle `null` values in `@NonNull` annotated parameters, or the other way around.

marcphilipp:
@sdeleuze Thanks for the explanation! Sounds like that's the way to go for us, too.

marcphilipp:

PR #1219, opened 2017-12-28
Introduce support for tag expressions

signed:
Tag expressions provide fine grained control over which tests to run,
as long as they are tagged. This is based on Boolean expressions.

## Overview
Up until now junit provides two parameters to select tagged tests for execution
`-t, --include-tag <String>`
`-T, --exclude-tag <String> `

`junit -t foo -t bar`  will run all test tagged with `foo OR bar`.
There is no way to only run tests that are tagged with `foo AND bar`.

This is now possible by extending the existing parameter to accept tag expressions.
The most simple tag expression being just a tag and therefore maintaining backward compatibility.

`junit -t "foo & bar"`
`junit -t "foo & bar" -t baz`     `((foo AND bar) OR baz)`
`junit -T "foo & !bar" -T baz`    `NOT ((foo AND NOT bar) | baz )`

For more details please have a look at the added documentation.

I'll force push required changes to keep this feature within a single commit.

---

---

### Definition of Done


sormuras:
Related issue: #454

marcphilipp:
@signed Thanks a lot for the PR! I'll review it in the next few days.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
```diff
@@ -117,12 +115,11 @@ public static PostDiscoveryFilter excludeTags(String... tags) throws Preconditio
 	 */
 	public static PostDiscoveryFilter excludeTags(List<String> tags) throws PreconditionViolationException {
 		Preconditions.notEmpty(tags, "tags list must not be null or empty");
-		List<TestTag> activeTags = toTestTags(tags);
-		return descriptor -> FilterResult.includedIf(descriptor.getTags().stream().noneMatch(activeTags::contains));
+		return TagExpressionFilter.includeMatching("! (" + orExpressionFor(tags) + ")");
```
marcphilipp: The Javadoc for these two methods doesn't make it clear that `tags` may now be a list of tag expressions instead of plain tags. Maybe we should keep this class as-is and add `includeTagExpressions` and `excludeTagExpressions` to `TagExpressionFilter`.

Thoughts?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/TagExpression.java
```diff
@@ -0,0 +1,35 @@
...
+
+	///CLOVER:OFF
+	private TagExpression() {
+		/* no-op */
+	}
+	///CLOVER:ON
+
+	public static ParseResult parseFrom(String infixTagExpression) {
+		return new Parser().parse(infixTagExpression);
+	}
```
marcphilipp: I think we should merge `Expression` and `TagExpression` (I prefer the latter name). Moreover, I think `parseFrom()` should return a `TagExpression`. Then, we can make `ParseResult` package-private.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ParseResults.java
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	static ParseResult error(String errorMessage) {
+		return new ParseResult() {
+			@Override
+			public Optional<String> errorMessage() {
+				return Optional.of(errorMessage);
+			}
+		};
+	}
```
marcphilipp: Any reason not to move these into `ParseResult`?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/Associativity.java
```diff
@@ -0,0 +1,15 @@
...
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.launcher.tagexpression;
+
+enum Associativity {
+	Left, Right
+}
```
marcphilipp: Can we move this into `Operator`?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ExpressionCreator.java
```diff
@@ -0,0 +1,15 @@
...
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.launcher.tagexpression;
+
+interface ExpressionCreator {
+	ParseStatus createExpressionAndAddTo(Stack<Position<Expression>> expressions, int position);
+}
```
marcphilipp: Can we move this into `Operator`?

platform-tests/src/test/java/org/junit/platform/launcher/tagexpression/ParserErrorTests.java
```diff
@@ -0,0 +1,104 @@
...
+                Arguments.of("foo & (bar |baz) &", "& at <7> missing rhs operand"),
+
+                Arguments.of("foo | (bar baz) &", "& at <6> missing rhs operand"),
+                Arguments.of("foo | (bar baz) &quux", "missing operator between bar <3> and (baz & quux) <6>"),
+
+                Arguments.of("foo & |", "& at <1> missing rhs operand"),
+                Arguments.of("foo !& bar", "! at <1> missing rhs operand"),
+                Arguments.of("foo !| bar", "! at <1> missing rhs operand")
+        );
+        // @formatter:on
```
marcphilipp: Ditto.

platform-tests/src/test/java/org/junit/platform/launcher/tagexpression/ParserTests.java
```diff
@@ -0,0 +1,77 @@
...
+				Arguments.of("foo", "foo"),
+				Arguments.of("! foo", "!foo"),
+				Arguments.of("foo & bar", "(foo & bar)"),
+				Arguments.of("foo | bar", "(foo | bar)"),
+				Arguments.of("( ! foo & bar | baz)", "((!foo & bar) | baz)"),
+				Arguments.of("(foo & bar ) | baz & quux", "((foo & bar) | (baz & quux))"),
+				Arguments.of("! foo | bar & ! baz | ! quux | quuz & corge", "(((!foo | (bar & !baz)) | !quux) | (quuz & corge))"),
+                Arguments.of("(foo & bar ) | baz & quux", "((foo & bar) | (baz & quux))"),
+                Arguments.of("foo | bar & baz|quux", "((foo | (bar & baz)) | quux)")
+        );
```
marcphilipp: Ditto.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ShuntingYard.java
```diff
@@ -0,0 +1,146 @@
...
+		this.tokens = tokens;
+		pushOperatorAt(-1, Sentinel);
+	}
+
+	public ParseResult execute() {
+		// @formatter:off
+		ParseStatus parseStatus = processTokens()
+                .process(this::consumeRemainingOperators)
+                .process(this::ensureOnlySingleExpressionRemains);
+        // @formatter:on
```
marcphilipp: Indentation looks wrong here. Spaces instead of tabs?

platform-tests/src/test/java/org/junit/platform/runner/JUnitPlatformRunnerTests.java
```diff
@@ -695,9 +738,10 @@ private static Description testDescription(String uniqueId) {
 		return createTestDescription(uniqueId, uniqueId, uniqueId);
 	}
 
-	private TestDescriptor testDescriptorWithTag(String tag) {
+	private TestDescriptor testDescriptorWithTag(String... tag) {
```
marcphilipp: `testDescriptorWithTags`?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/Operator.java
```diff
@@ -0,0 +1,121 @@
...
+	static Operator nullaryOperator(String representation, int precedence) {
+		return new Operator(representation, precedence, 0, null, (expressions, position) -> success());
+	}
+
+	static Operator unaryOperator(String representation, int precedence, Associativity associativity,
+			Function<Expression, Expression> unaryExpression) {
+		return new Operator(representation, precedence, 1, associativity, (expressions, position) -> {
+			Position<Expression> rhs = expressions.pop();
+			if (position < rhs.position) {
+				Expression not = unaryExpression.apply(rhs.element);
```
marcphilipp: Since this is more general, it should *not* be called `not`... 😉

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/package-info.java
```diff
@@ -0,0 +1,6 @@
+/**
+ * The tag expression language {@link org.junit.platform.launcher.tagexpression.Parser}
+ * and related support classes for.
```
marcphilipp: Since `Parser` is package-private it won't make it into the generated Javadoc. Thus, we should not link to it. I think `The tag expression language parser and related support classes.` is sufficient.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ShuntingYard.java
```diff
@@ -0,0 +1,146 @@
...
+import static org.junit.platform.launcher.tagexpression.ParseStatus.missingClosingParenthesis;
+import static org.junit.platform.launcher.tagexpression.ParseStatus.missingOpeningParenthesis;
+import static org.junit.platform.launcher.tagexpression.ParseStatus.missingOperator;
+import static org.junit.platform.launcher.tagexpression.ParseStatus.success;
+
+import java.util.List;
+
+/**
+ * This is based on a modified version of the <a href="https://en.wikipedia.org/wiki/Shunting-yard_algor">Shunting-yard algorithm</a>
+ */
```
marcphilipp: The link is broken. Please wrap Javadoc comments at 80 chars, if possible.

platform-tests/src/test/java/org/junit/platform/launcher/tagexpression/ParserErrorTests.java
```diff
@@ -0,0 +1,104 @@
...
+
+	@Test
+	void partialUnaryOperator() {
+		assertThat(parseErrorFromParsing("!")).contains("! at <0> missing rhs operand");
+	}
+
+	@Test
+	void partialBinaryOperator() {
+		assertThat(parseErrorFromParsing("& foo")).contains("& at <0> missing lhs operand");
+		assertThat(parseErrorFromParsing("foo |")).contains("| at <1> missing rhs operand");
```
marcphilipp: These error messages are already pretty good! I'm wondering whether we could store the character index for each token and emit it as part of the error message instead of the "token index". Thoughts?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
signed: And divert the callers in `JUnitPlatform` and `DiscoveryRequestCreator` to the new methods? Otherwise tag expressions will not work in gradle/maven.
Am I missing something?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ParseResults.java
signed: If they are in `ParseResult`they become accessible outside the package and become part of the public API. I want to avoid this, because nobody outside the package needs those methods

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ShuntingYard.java
signed: replaced with tabs

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ShuntingYard.java
signed: Fixed the link

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/Operator.java
signed: inlined not to keep make it more similar to binaryOperator

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ExpressionCreator.java
signed: done

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/Associativity.java
signed: done

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/package-info.java
signed: done.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/TagExpression.java
signed: If `parseFrom()` returns a TagExpression/Expression how do you then communicate a parse error? In this case I do not like to throw an Exception, because it is not an exceptional thing for a parser to get invalid input. Thats why I prefer to keep `ParseResult` as return type of `parseFrom`.
This makes the method easier to re-use for example in the IDEA run dialogue if they start supporting tag expressions.

How about this:
1. move `parseFrom()` into Expression
1. delete `TagExpression`
1. rename `Expression` -> `TagExpression`

platform-tests/src/test/java/org/junit/platform/runner/JUnitPlatformRunnerTests.java
signed: Done

platform-tests/src/test/java/org/junit/platform/launcher/tagexpression/ParserErrorTests.java
signed: I'll try to tweak the Lexer and see what I can do.

signed:
> I'll force push required changes to keep this feature within a single commit.
Had a look at the [github documentation](https://help.github.com/articles/about-pull-requests/) and they think it is a bad idea ...

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
marcphilipp: You're right. There's at least the following callers: 

- Gradle plugin / `ConsoleLauncher` (`-t`/`-T`)
- `JUnitPlatform` runner (`@IncludeTags`/`@ExcludeTags`)
- Surefire provider (`includeTags`/`excludeTags`)

All of them currently allow to specify a list of tags to include or exclude. I think it doesn't make sense to duplicate these APIs to handle tag expressions separately. In that case, I think we should then also get rid of `TagExpressionFilter` and move `includeMatching` as a private method into `TagExpression`. Moreover, we'd have to change the documentation for the above APIs to mention that tag expressions are now allowed.

However, before we go down that road, I'd like to discuss this the @junit-team/junit-lambda first. I'll bring it up in the team call next week and let you know afterwards.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/ParseResults.java
marcphilipp: Makes sense! Thanks for the explanation!

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/TagExpression.java
marcphilipp: You're right, I forgot about that use case. This plan sounds good. 👍

platform-tests/src/test/java/org/junit/platform/launcher/tagexpression/ParserErrorTests.java
marcphilipp: 👍

marcphilipp:
> Had a look at the [github documentation](https://help.github.com/articles/about-pull-requests/) and they think it is a bad idea ...

Yeah, don't worry about that. GitHub now even offers the functionality to do it [when merging the PR](https://github.com/blog/2141-squash-your-commits).

platform-tests/src/test/java/org/junit/platform/launcher/tagexpression/ParserErrorTests.java
signed: Done

junit-platform-launcher/src/main/java/org/junit/platform/launcher/tagexpression/TagExpression.java
signed: done

marcphilipp:
Team decision: We will allow tag expressions in all places tags are allowed right now.

Action items:

- [ ] Delete `TagExpressionFilter` (move method into `TagFilter`)
- [ ] Change parameter names, Javadoc, and documentation for Gradle plugin, Surefire provider, and `JUnitPlatform` runner/annotations

@signed We'd like to release 5.1 M2 next weekend. Do you have time to work on these changes within the next week? I really want to include this PR. So, if you don't have time, please let us know, so we can take over.

signed:
I moved the factory method. I started with the documentation, but I'm not 100% sure what you want to have changed. Can you pitch in with the documentation, as you already have a plan in mind?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
```diff
@@ -117,12 +122,27 @@ public static PostDiscoveryFilter excludeTags(String... tags) throws Preconditio
...
+	/**
+	 * Create an <em>include</em> filter based on the supplied {@code infixTagExpression}.
+	 *
+	 * <p>Containers and tests will only be executed if their tags match the supplied <em>infixTagExpression</em>.
+	 *
+	 * @param infixTagExpression to parse and evaluate against a {@link TestDescriptor}; never {@code null} or empty
+	 * @throws PreconditionViolationException if the supplied infixTagExpression can not be parsed.
+	 * @since 1.1
+	 */
+	public static PostDiscoveryFilter includeMatching(String infixTagExpression) {
```
sbrannen: Unless I'm mistaken, the idea was to make `includeMatching` a `private` method -- right?

sbrannen:
> I moved the factory method.

Thanks.

I left a comment on the visibility of the new method.

sbrannen:
> I started with the documentation, but I'm not 100% sure what you want to have changed. Can you pitch in with the documentation, as you already have a plan in mind?

If you feel uncomfortable with taking on the documentation task, I'm sure we can handle it after merging in the code changes.

Just let us know if that's the route you prefer.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
signed: Made it private

signed:
> Just let us know if that's the route you prefer.

Please take over the documentation part. I have pushed the documentation changes I have made to the [tag-expressions-documentation](https://github.com/signed/junit5/commits/tag-expressions-documentation) branch in my fork. As far as I can see `ExcludeTags` and `IncludeTags` are the ones still missing. Feel free to use/ignore what ever part of my documentation changes you want.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
sbrannen: thanks

sbrannen:
> Please take over the documentation part. I have pushed the documentation changes I have made to the tag-expressions-documentation branch in my fork. As far as I can see ExcludeTags and IncludeTags are the ones still missing. Feel free to use/ignore what ever part of my documentation changes you want.

OK. Thanks!

We'll look into it...

marcphilipp:
I'm starting to look now. 🙂

marcphilipp:
Since I can't push to signed/tag-expressions, I'm continuing on a separate branch: [issues/454-tag-expressions](https://github.com/junit-team/junit5/tree/issues/454-tag-expressions).

I've added the documenation I had in mind in 142728b88d456334f2dd71d158a4688559308b17.

@sbrannen Would be great if you could polish wrt. language. 🙂

marcphilipp:
Thanks for the great work, @signed! 👍 

I've squashed your commits, added a few additional ones, and submitted #1253.

Issue #1025, opened 2017-08-18
Cannot select @TestTemplate or @TestFactory invocations by Unique ID


noopur2507:
## Overview

With reference to [bug request 520923](https://bugs.eclipse.org/bugs/show_bug.cgi?id=520923) in Eclipse, I am trying to rerun a single invocation of a parameterized test (and similarly, a repeated and a dynamic test) using its unique ID like this:

``` java
LauncherDiscoveryRequest request = LauncherDiscoveryRequestBuilder.request().selectors(DiscoverySelectors.selectUniqueId(uniqueId)).build();
```

A new `Launcher` instance is created to discover the above request.

This method works for a normal test with ID:

``` java
[engine:junit-jupiter]/[class:test.FirstTest]/[method:myFirstTest(org.junit.jupiter.api.TestInfo)]
``` 

But it gives this error for a dynamic test:

```
org.junit.jupiter.engine.discovery.JavaElementsResolver resolveUniqueId
WARNING: Unique ID '[engine:junit-jupiter]/[class:test.dynamic.DynamicTestsDemo]/[test-factory:dynamicTestsFromIntStream()]/[dynamic-test:#2]' could not be resolved
```

or a parameterized test:

```
org.junit.jupiter.engine.discovery.JavaElementsResolver resolveUniqueId
WARNING: Unique ID '[engine:junit-jupiter]/[class:jb.Junit5SimpleParamTest]/[test-template:dummy(java.lang.String)]/[test-template-invocation:#1]' could not be resolved
```

In JUnit 4, it was done using `org.junit.runner.Request.filterWith(Filter filter)`. 

## Question

What is the recommended way to do this in JUnit 5?

## Related Issues

- #1026 
- #1031

## Analysis

Given a _Unique ID_ such as `[engine:junit-jupiter]/[class:test.dynamic.DynamicTestsDemo]/[test-factory:dynamicTestsFromIntStream()]/[dynamic-test:#2]`,
`JavaElementsResolver.resolveUniqueId(TestDescriptor, List<Segment>)` returns `true` if **all** segments of the supplied Unique ID could be resolved.

So, in such a case, the test class and test factory method do in fact get resolved, but the dynamic test itself does not get resolved.

That means that the test class and test factory method will be executed in the test plan even though there is a warning (which actually only applies to the last segment of the Unique ID).

## Proposals

1. Selection of any invocation of a test template or test factory results in selection of the test template or test factory as a _container_ and consequently the _selection_ of all invocations.
[...]

PR #1228, opened 2018-01-04
selectively execute arguments from method source using only

signed:
## Overview

Debugging a single case of a parameterized test is quiet cumbersome. To make this easier this pull requests adds a new factory method `only` on `Arguments`.

As soon as at least a single Argument in a `MethodSource` is flagged as `only` only those are executed. If no Argument is flagged with `only` all elements of the MethodSource are executed.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
@@ -38,6 +38,10 @@ public interface Arguments {
 	 */
 	Object[] get();
 
+	default boolean only() {
+		return false;
+	}
+
 	/**
 	 * Factory method for creating an instance of {@code Arguments} based on
 	 * the supplied {@code arguments}.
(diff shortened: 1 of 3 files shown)


marcphilipp:
Interesting idea, thanks for the PR!

To be honest, I think we should rather fix #1025 instead of introducing a workaround that only works for `@MethodSource`.

signed:
The linked issue seems to be about re-running failed tests.
In my case I want to run a single Argument without having to run all Arguments first.

Or is there a way for the IDE to derive the unique id without running all Arguments first?

marcphilipp:
> In my case I want to run a single Argument without having to run all Arguments first.

Understood. Any ideas how to make this more generally useful besides `@MethodSource`?

> Or is there a way for the IDE to derive the unique id without running all Arguments first?

Nope, since these are dynamic descriptors, the IDE cannot discover them before execution.

sbrannen:
> To be honest, I think we should rather fix #1025 instead of introducing a workaround that only works for `@MethodSource`.

I agree with @marcphilipp on this one.

The typical developer follows the following steps:

1. run parameterized test.
2. it fails for a given invocation.
3. re-run the single invocation (covered by #1025) in _debug_ mode.

Thus, there is no need to pick a particular invocation in advance (i.e., before having seen that it even fails).

I therefore propose that we close this issue.

sormuras:
> I therefore propose that we close this issue.

+1

sbrannen:
**Team Decision**:

Since the typical developer workflow follows the steps outlined in https://github.com/junit-team/junit5/pull/1228#issuecomment-356294389, we have decided to respectfully decline this PR.

Issue #219, opened 2016-03-31
Introduce predefined @Disabled*/@Enabled* variants


ttddyy:
## Overview

I just happened to have a test only failing on MS Windows machines and thought about JUnit Jupiter.

I could write a custom extension implementing `ContainerExecutionCondition` and `TestExecutionCondition`, but having a predefined `@Disabled` annotation such as `@DisabledIfWindows` or `@Disabled(condition = IsWindowsEvaluator.class)` to extend disabling feature might be handy.

## Related Issues

- #1247

## Deliverables

- [x] system property matches a pattern
- [x] environment variable matches a pattern
- ❌  current day of week matches one of the supplied enums
  - The team decided not to publish `@EnabledOnDayOfWeek` and `@DisabledOnDayOfWeek`.
- [x] current OS matches one of the supplied enums
- [x] current JRE version matches one of the supplied enums
- [x] Check for user configuration errors. For example, if a user declares `@EnabledOnOs({})`, that should result in an exception instead of a disabled result.
- [x] Introduce integration tests for each condition.
- [x] Write JavaDoc for each annotation.
- [x] Document in User Guide.
- [x] Document in Release Notes.

ttddyy:
I just realized I should write like this to skip the test on windows machines for my use case.
 `assumeFalse(System.getProperty("os.name").toLowerCase().contains("win"))`

However, I still think *annotating* such condition would be easier to detect than writing assumptions in body of the test methods.

sbrannen:
@ttddyy, that's an interesting idea, and I in fact had a similar idea (hence the new subject).

I'm not sure that we will introduce such mechanisms in core JUnit 5, but I do imagine they will end up somewhere (perhaps in an add-on module or in an example project).

sbrannen:
We have recently received a recommendation for introducing an annotation like `@DisabledUntil( /* date */)`.

smoyer64:
I wanted to add a use-case to this issue that was described during the "Metrics-Driven Continuous Delivery in Practice [TUT5941]" session at JavaOne.

The presenter's company (Dynatrace) has developed a custom @Ignore annotation which allows tests to be ignored only if the issue (in this case in Jira) it's testing against is still open.  This seems like a pretty clever TDD trick too.

sbrannen:
Yep, that's also a clever use case.

But something like that is very specific to a particular technology. So I don't foresee that being implemented as a generic solution within JUnit itself.

smoyer64:
[...]

PR #1257, opened 2018-01-22
Introduce @EnabledIf execution condition

sormuras:
# Evaluate predicate to en- or disable test execution

Addresses #219

## Examples

`JavaScript` examples

```java
@EnabledIf("true")
@EnabledIf("false")
@EnabledIf("1 == 2")
@EnabledIf("java.lang.Boolean.getBoolean('is-not-set')")
@EnabledIf(imports = "java.nio.file", value = "Files.exists(Paths.get('dir', 'file.txt'))")
@EnabledIf("java.lang.System.getProperty('os.name').toLowerCase().contains('win')")
@EnabledIf("/64/.test(systemProperties.get('os.arch'))")
```

More `Groovy`

```java
@Conditional(engine = "groovy", // Select Groovy as language.
	delimiter = "\n", // Set delimiter to Unix-style line separator.
	bindExtensionContext = false, // Not needed here.
	bindSystemProperties = false, // Not needed here, Groovy auto-imports "java.lang.*".
	value = { //
			"System.properties['jsr'] = '233'", //
			"'233' == System.properties['jsr']" }, //
	reason = "Groovy as script language, multiple lines, self-fulfilling.")
@Test
void groovy() {
	assertEquals("233", System.getProperty("jsr"));
}
```

### Supported script languages

All JSR-223 ScriptEngine compatible languages, like:

- `javascript` http://openjdk.java.net/projects/nashorn/

### Just mount the script engine at runtime...

- `groovy` http://docs.groovy-lang.org/latest/html/api/org/codehaus/groovy/jsr223/GroovyScriptEngineFactory.html
- `bsh` http://beanshell.org

### Other script languages

- `jshell` https://github.com/dmac100/JShellScriptEngine
- `SpEL` https://jira.spring.io/browse/SPR-7651

---

---

### Definition of Done

### DIFF
--- a/documentation/documentation.gradle
+++ b/documentation/documentation.gradle
@@ -35,6 +35,8 @@ test {
 */
 
 dependencies {
+	runtimeOnly('org.codehaus.groovy:groovy-jsr223:2.4.12')
+
 	testImplementation(project(':junit-jupiter-api'))
 	testImplementation(project(path: ':junit-jupiter-params', configuration: 'shadow'))
 	testImplementation(project(':junit-platform-runner'))
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.1.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.1.0-M2.adoc
@@ -147,7 +147,8 @@ _@API Guardian_ JAR _mandatory_ again.
    example, `fail<Nothing>("Some message")`. These new top-level functions remove this
    requirement by returning
    https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-nothing.html[`Nothing`].
-
+* New `@EnabledIf` execution condition added. Supply a script, default language is
+  `JavaScript`, that controls whether the container or test is executed.
 
 [[release-notes-5.1.0-M2-junit-vintage]]
 === JUnit Vintage
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -164,6 +164,17 @@ And here's a test case with a disabled test method.
 include::{testDir}/example/DisabledTestsDemo.java[tags=user_guide]
 ----
 
+[[writing-tests-conditional]]
+=== Conditional Test Execution
+
+Here are some tests that are enabled or disabled depending
+on the evaluation of the script passed to the `@EnabledIf` annotation.
+
+[source,java,indent=0]
+----
+include::{testDir}/example/EnabledIfTestsDemo.java[tags=user_guide]
+----
+
 [[writing-tests-tagging-and-filtering]]
 === Tagging and Filtering
 
--- a/documentation/src/test/java/example/EnabledIfTestsDemo.java
+++ b/documentation/src/test/java/example/EnabledIfTestsDemo.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package example;
+
+// tag::user_guide[]
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.junit.jupiter.api.EnabledIf;
+import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.Test;
+
+class EnabledIfTestsDemo {
+
+	@Test // Static JavaScript expression
+	@EnabledIf("1 == 1")
+	void testWillBeExecuted() {
+		assertTrue(1 == 1);
+	}
+
+	@RepeatedTest(10) // Dynamic JavaScript expression
+	@EnabledIf("Math.random() >= 0.314159")
+	void testWillNeverOrSometimesBeExecuted() {
+	}
+
+	@Test // Regular expression testing bound system property.
+	@EnabledIf("/64/.test(systemProperties.get('os.arch'))")
+	void testWillBeExecutedIfOsArchitectureContains64() {
+		assertTrue(System.getProperty("os.arch").contains("64"));
+	}
+
+	@Test // Multi-line script and import Java package names.
+	@EnabledIf(imports = "java.nio.file", //
+			value = { //
+					"path = Files.createTempFile('volatile-', '.temp')", //
+					"systemProperties.put('volatile', path)", //
+					"Files.exists(path)" })
+	void importJavaPackages() {
+		assertTrue(Files.exists((Path) System.getProperties().get("volatile")));
+	}
+
+	@Test // Select Groovy as ScriptEngine.
+	@EnabledIf(engine = "groovy", //
+			imports = { "java.nio.file.Files", "java.nio.file.Paths" }, //
+			value = "Files.isWritable(Paths.get(System.properties['user.home']))")
+	void groovy() {
+		assertTrue(Files.isWritable(Paths.get(System.getProperty("user.home"))));
+	}
+}
+// end::user_guide[]
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/EnabledIf.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/EnabledIf.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api;
+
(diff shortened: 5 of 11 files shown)


mkobit:
A few personal thoughts from a quick glance at the user-facing API:

- One downside to this is that you can't really have a statically typed or checked `value`
- Managing imports separately looks like it would feel clunky 
- `@Enabled` is asymmetrical to `@Disabled` - `@Disabled` always implied test disabled with a reason, while `@Enabled` takes in various arguments to determine if it is enabled.
- Script evaluation cannot provide a reason itself. The reason must be statically defined

This is cool and definitely useful. NiFI has something similar with its [`ExecuteScript`](https://nifi.apache.org/docs/nifi-docs/components/org.apache.nifi/nifi-scripting-nar/1.5.0/org.apache.nifi.processors.script.ExecuteScript/index.html) processor.

sbrannen:
First and foremost, thanks to @sormuras for having the idea and putting together this PR for discussion!

@mkobit, it's just a proof concept at the moment, but we still appreciate early feedback.

> - One downside to this is that you can't really have a statically typed or checked `value`

Well, that's life with Java annotations, and we unfortunately cannot change that. If we want to support scripting languages for conditional test execution (where the script is supplied via an annotation), I don't see any way around this.

Got any better ideas?

> - Managing imports separately looks like it would feel clunky 

True. Any better ideas?

> - `@Enabled` is asymmetrical to `@Disabled` - `@Disabled` always implied test disabled with a reason, while `@Enabled` takes in various arguments to determine if it is enabled.

I agree that the name would have to be changed. I think something along the lines of `@EnabledIf` would be better (even if that clashes with the name of the annotation I introduced in Spring Framework 5).

> - Script evaluation cannot provide a reason itself. The reason must be statically defined

Another good point. Any suggestions on how to achieve that in a simple and clean manner?

sormuras:
@mkobit Thanks for the feedback!

As Sam said, it's an early draft and nothing's set in stone, yet.

In general, I see this annotation as a lightweight (read: quick'n dirty) but flexible band-aid that can never replace a dedicated, strongly typed, custom implementation of `ExecutionCondition`. For example, there should be an extension that solely cares for the underlying operating system or another one testing the Java runtime version. Many more ideas are collected here: https://github.com/junit-pioneer/junit-pioneer/issues/1

If you find yourself repeating the same script in various places, you should a) implement a dedicated `ExecutionCondition` and b) think about the testability of your business modell.

### Other names
- `@Iff` _if and only if_
- `@EnabledIf`
- `@Predicate`
- `@Script[Barrier|Guard|Test]`

### "imports()"

It is just a short-cut for the following wrapper code:

```javascript
var javaImporter = new JavaImporter(" + imports() + ")
with (javaImporter) {
  // your script code here
}
```
Leave the array empty and nothing will be inserted.

### reason() is static

a) If you need to inform the user in a more detailed way, implemented a dedicated `ExecutionCondition`.
b) The reason is explicitly given and determined by the script itself. See the console output for some scripts that evaluated to `false` or contain syntax errors:

```javascript
// @Enabled("java.lang.Boolean.getBoolean('is-not-set')")
Script `java.lang.Boolean.getBoolean('is-not-set')` evaluated to: false

// @Enabled("1 == 2")
Script `1 == 2` evaluated to: false

// @Enabled(value = "junit$context.publishReportEntry('foo', 'bar')", reason = "no result, no execution")
timestamp = 2018-01-23T18:48:31.555349500, foo = bar
no result, no execution

// @Enabled("false")
Script `false` evaluated to: false

// @Enabled(value = "syntactically, some thing is not right")
<eval>:1:20 Expected ; but found thing
syntactically, some thing is not right
                    ^ in <eval> at line number 1 at column number 20
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Enabled.java
```diff
@@ -0,0 +1,92 @@
...
+@API(status = STABLE, since = "5.1")
+public @interface Enabled {
+
+	/**
+	 * Short name of the {@link javax.script.ScriptEngine ScriptEngine} to use.
+	 *
+	 * @return script engine name, defaults to {@code "javascript"}
+	 * @see javax.script.ScriptEngineManager#getEngineByName(String)
+	 */
+	String engine() default "javascript";
```
marcphilipp: I think we should make the default engine a configuration parameter because it's very likely that you'd want to use the same engine for all your `@Enabled` annotations.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/EnabledCondition.java
```diff
@@ -0,0 +1,152 @@
...
+	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+		Optional<AnnotatedElement> element = context.getElement();
+		Optional<Enabled> annotation = findAnnotation(element, Enabled.class);
+		if (!annotation.isPresent()) {
+			return enabled("@Enabled is not present");
+		}
+
+		// Prepare actual script text from annotation properties
+		Enabled enabled = annotation.get();
+		Preconditions.condition(enabled.value().length > 0, "@Enable#value() array must not be empty");
```
marcphilipp: FWIW there's a `d` missing.

marcphilipp: How about `Preconditions.notBlank()`?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/EnabledCondition.java
```diff
@@ -0,0 +1,152 @@
...
+			lines.add("}");
+			lines.add("");
+		}
+
+		return joinLines(enabled.delimiter(), lines);
+	}
+
+	private String joinLines(String delimiter, Iterable<? extends CharSequence> elements) {
+		if (delimiter.isEmpty()) {
+			delimiter = System.lineSeparator();
```
marcphilipp: Not sure we should make this dependent on the current platform. That way, test execution might yield different results on Windows and on the CI server. How about just using `\n`?

mkobit:
@sbrannen 

I don't have any concrete suggestions regarding making the script content static. I think going down that route for users would probably end up in the territory of "you should probably just write your own `ExecutionCondition`".

On the imports, I don't know if there is a different way other than putting them into the script body themselves, which doesn't seem any better. Another option may be to compose multiple different piece together such as script engine creation, script lookup, execution environment preparation, execution, etc, so that the individual pieces may be reused. Some use cases there might be "the default imports for script execution for Groovy should always include my list and the JUnit API". I'm just thinking out loud here as that increases the complexity and API surface.

Regarding scripts providing a reason - it looks like `true` or `false` are the expected results right now. Maybe scripts could also be allowed to return the `ConditionEvaluationResult`, or just always expect to return `ConditionEvaluationResult`? But then again, maybe that should be reserved more for strongly typed `ExecutionCondition` as @sormuras brought up in his comment. Allowing scripts to return multiple different types could be more API surface and the "dynamicness" of it could be confusing.

----

@sormuras 

> In general, I see this annotation as a lightweight (read: quick'n dirty) but flexible band-aid that can never replace a dedicated, strongly typed, custom implementation of `ExecutionCondition`

Totally agree with that. This implementation will enable a bunch of use cases with (what appears to be) little overhead and API surface, while making easier to programmatically enable or disable tests.

Naming is obviously difficult, and I don't want to harp on too much, but I did want to point out the differences.

Working code and examples is always better than hypotheticals! Looking forward to this!

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Enabled.java
sormuras: Good point. If available, I'll prefer the configuration parameter value over the default return value defined here.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/EnabledCondition.java
sormuras: Not so good. Using `Preconditions.notEmpty` instead. It's an array of `String` (to support multi-line scripts).

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/EnabledCondition.java
sormuras: Hm, what-if a script parser is implemented in platform-specific manner? I think the default is okay here, the user may override with `\n` or ` ` or _what-ever_ within the `@Conditional` usage.

sormuras:
Force-pushed -- now using `@Conditional` as name and allowing a custom `ConditionEvaluationResult` instance.

sormuras:
`@Conditional` is still not the final name...

- `@Condition`
- `@EnabledIf`
- `@Assume[That]`
- `@Iff` _if and only if_
- `@Predicate`
- `@Script[Barrier|Guard|Test]`

Or invert the predicate logic and "extend" the unconditional `@Disabled` with an evaluating `@DisabledIf`:
```java
@RepeatedTest(10)
@DisabledIf("Math.random() >= 0.5")
void test() {}
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/EnabledCondition.java
marcphilipp: Ah, sorry, I missed that. If only Java had multiline strings...

sormuras:
Team-decision: `@EnabledIf` it is.

sormuras:
Superseded by https://github.com/junit-team/junit5/pull/1267

Issue #1258, opened 2018-01-23
Certain JUnit Jupiter assertions perform slower than JUnit 4.12's


karayv:
## Overview

Accidentally I found out that JUnit 5.0.3 assertion (assertTrue) worked significantly slower comparing to JUnit 4.12.

In my 2 tests I repeat the same assertion many times. One test uses old junit, other - junit5.

https://github.com/karayv/junit-perf-demo/blob/master/src/test/java/my/perftest/JUnit5Test.java
```java
public class JUnit5Test {
    TestRunner testRunner = new TestRunner();

    @Test
    void test() {
        testRunner.go(Assertions::assertTrue);
    }
}
```
I was able to pass **288230376151711744 assertions in 1.515 sec with JUnit 4** 
and only **17179869184 assertions in 3.015 sec with JUnit 5**

I created a sample repo https://github.com/karayv/junit-perf-demo where the issue can be reproduced and possibly investigated. It contains four classes: _JUnit5Test_, _OldJUnitTest_, _TestRunner_ and a imaginary super fast prod service - _Library_.

Full console output for both tests:

_JUnit5Test:_
```
Level of load: 15, duration: PT0.195S, number of operations: 1073741824
Level of load: 16, duration: PT0.771S, number of operations: 4294967296
Level of load: 17, duration: PT3.015S, number of operations: 17179869184
```

 _OldJUnitTest:_
```
Level of load: 26, duration: PT0.185S, number of operations: 4503599627370496
Level of load: 27, duration: PT0.383S, number of operations: 18014398509481984
Level of load: 28, duration: PT0.758S, number of operations: 72057594037927936
Level of load: 29, duration: PT1.515S, number of operations: 288230376151711744
```

jbduncan:
I believe this issue is related to https://github.com/junit-team/junit5/issues/880. :)

karayv:
Oops...
:)

Okay, I took a look myself

It looks like the issue is in creation of _java.util.function.BooleanSupplier_

If I change my JUnit4 test to 
```java
public class OldJUnitTest {
    LoadRunner testRunner = new TestRunner();

    @Test
    public void test() {
        //testRunner.go(Assert::assertTrue);
        
        testRunner.go((cond) -> {
            BooleanSupplier bs = () -> cond;
            Assert.assertTrue(bs.getAsBoolean());
        });
    }
}
```
performance degrades to the same level.

karayv:
A quick pull-request
[...]

PR #1259, opened 2018-01-24
Issue-1258 Issue-880 Removed unnecessary creation of BooleanSupplier

karayv:
## Overview

BooleanSupplier is an object that is created every time assertion is called. This affects performance. My pull request changes that. If condition comes as a boolean, that specific boolean is checked.

It's non-functional change, therefore I didn't put any tests for that. Also no other assertion classes except of AssertTrue were checked against performance.

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTrue.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTrue.java
@@ -32,11 +32,13 @@ class AssertTrue {
 	///CLOVER:ON
 
 	static void assertTrue(boolean condition) {
-		assertTrue(() -> condition, () -> null);
+		assertTrue(condition, () -> null);
 	}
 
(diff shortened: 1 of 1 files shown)


karayv:
Okay, this pull-request is too quick. More work needs to be done in order to properly test the fix. Most likely the issue will persist because of  `Supplier<String> messageSupplier`.

sbrannen:
Hi @karayv,

If performance is truly an issue here (i.e., you've benchmarked it), then I'd prefer that you open an issue to address this concern across the board for both `BooleanSupplier` **and** `Supplier<String>` in **all** assertions in JUnit Jupiter.

Then the team can discuss the topic (based on your benchmarks) and decide whether or not to go forward with such changes.

In light of that, I'm declining this PR.

Looking forward to your "issue".

Cheers,

Sam

sbrannen:
@karayv, I just noticed you already opened #1258.

I somehow missed that previously.

In any case, let's continue the discussion there.

Issue #880, opened 2017-06-10
JUnit 5 is 10x slower than JUnit 4


ctapobep:
# Code to reproduce

## JUnit Jupiter

```java
    static long nOfTimes, start = System.nanoTime();

    @org.junit.jupiter.api.Test
    void testSpeed() {
        Assertions.assertTrue(true);
        if(++nOfTimes == 1000)
            System.out.println(TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start));//10085
    }
```

## JUnit 4

```java
    static long nOfTimes, start = System.nanoTime();

    @org.junit.Test
    public void testSpeed() {
        org.junit.Assert.assertTrue(true);
        if (++nOfTimes == 1000)
            System.out.println(TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start));//1041
    }
```

# Results

**Results** on my machine (i5): ~10sec (JUnit5) and ~1sec (JUnit4).

# Environment

* IntelliJ 2017.2 EAP (see screenshot on how to setup 1K repeats)
* JUnit 4.11
* Jupiter 5.0.0-20170610.145620-613
* Platform Launcher 1.0.0-20170610.145632-612
* Platform Engine 1.0.0-20170610.145629-613
* Surefire Provider 1.0.0-20170610.145635-612
* MacOS

![intellij-repeat-tests-n-times](https://user-images.githubusercontent.com/702510/27004661-8e23804a-4e15-11e7-96df-b2bfa66010d3.png)

# Justification

The speed will be important for randomized testing since it's a common practice to repeat such tests many times. Especially if someone implements a Property Based Testing framework like QuickCheck based on JUnit5.

sormuras:
Thanks for the report, @ctapobep -- what happens to the timings when you leave IDEA out off the setup?

Maybe you want to setup a [JMH](http://openjdk.java.net/projects/code-tools/jmh/) suite to compare JUnit 4 and JUnit Platform + JUnit Jupiter again?

ctapobep:
Um.. I don't know how relevant that would be:

1. You need to run the test many times - JUnit4 doesn't support this. Though I assume it's possible to write a rule for this, but this is not the same as testing multiple test launches. So such a test would check different performance.
2. I'm particularly interested in the speed of tests in IDE because that is usually the environment where I run them many times.

[...]

PR #1260, opened 2018-01-25
Avoid unnecessary lambda creation in assertions and preconditions

karayv:
## Overview

Lambda objects are created on assertions. This affects performance. 

This pull request removed unnecessary object creation.

Issues: #1258, #880


marcphilipp:
Thanks for the PR! Before we merge this, I'd like to see some micro-benchmarks that verify that this is a real issue. Can you use JMH to do that?

sbrannen:
FYI: @sormuras set up JMH for us in the Gradle build.

Latest results:

```
Benchmark                                                              Mode  Cnt           Score           Error  Units
AssertionBenchmarks.junit4_assertTrue_String_boolean                  thrpt   20  3032294853,832 ±  86272460,759  ops/s
AssertionBenchmarks.junit4_assertTrue_boolean                         thrpt   20  2712835302,130 ± 301584020,424  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_BooleanSupplier_String    thrpt   20  2981161203,797 ±  81629519,280  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_BooleanSupplier_Supplier  thrpt   20  2311813113,566 ± 529503426,823  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean                   thrpt   20  2614259276,714 ± 520176160,930  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean_String            thrpt   20  2959380217,598 ± 149093095,096  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean_Supplier          thrpt   20  2574678336,447 ± 382469807,611  ops/s
```

sbrannen:
FYI: I revised the title of this PR in order to conform with the project's commit guidelines.

marcphilipp:
When using the recommended fork option (fork = 1), there is literally no difference between the different assertions:

```
AssertionBenchmarks.junit4_assertTrue_String_boolean                  thrpt   10  3211005206,690 ± 160833563,096  ops/s
AssertionBenchmarks.junit4_assertTrue_boolean                         thrpt   10  3283039382,681 ± 109724472,251  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_BooleanSupplier_String    thrpt   10  3280715074,446 ± 111796460,389  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_BooleanSupplier_Supplier  thrpt   10  3270766322,493 ± 114843136,087  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean                   thrpt   10  3262371952,329 ± 137961027,785  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean_String            thrpt   10  3283744956,156 ± 109271283,420  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean_Supplier          thrpt   10  3260943468,363 ± 147164523,018  ops/s
```

karayv:
@marcphilipp @sbrannen 
Probably JMH is not able to generate such load. My original issue stated 5,698,132,399 op/sec for JUnit5, and 190,251,073,367,466,500 op/sec for JUnit4. JHM reports much less ops - 3,260,943,468

I ran this code against my branch and master
```java
	@RepeatedTest(10)
	void testName(RepetitionInfo repetitionInfo) throws Exception {
		
		long start = System.nanoTime();

		for (int i = 0; i < Integer.MAX_VALUE; i++) {
			org.junit.jupiter.api.Assertions.assertTrue0(true);
		}

		long end = System.nanoTime();
		long oneSecond = 1_000_000_000;
		System.out.printf("Assertions.assertTrue(true) call #%2d: %d op/sec\n", repetitionInfo.getCurrentRepetition(),
				oneSecond / (end - start) * (long) Integer.MAX_VALUE);
	}
```

**Results**
Current master
```
Assertions.assertTrue(true) call # 1: 45097156587 op/sec
Assertions.assertTrue(true) call # 2: 45097156587 op/sec
Assertions.assertTrue(true) call # 3: 55834574822 op/sec
Assertions.assertTrue(true) call # 4: 55834574822 op/sec
Assertions.assertTrue(true) call # 5: 49392123881 op/sec
Assertions.assertTrue(true) call # 6: 53687091175 op/sec
Assertions.assertTrue(true) call # 7: 51539607528 op/sec
Assertions.assertTrue(true) call # 8: 47244640234 op/sec
Assertions.assertTrue(true) call # 9: 53687091175 op/sec
Assertions.assertTrue(true) call #10: 51539607528 op/sec
```

This branch
```
Assertions.assertTrue(true) call # 1: 474593885987 op/sec
Assertions.assertTrue(true) call # 2: 644245094100 op/sec
Assertions.assertTrue(true) call # 3: 30137785501998 op/sec
Assertions.assertTrue(true) call # 4: 26188822920686287 op/sec
Assertions.assertTrue(true) call # 5: 37025579454229213 op/sec
Assertions.assertTrue(true) call # 6: 26188822920686287 op/sec
Assertions.assertTrue(true) call # 7: 46684426175007110 op/sec
Assertions.assertTrue(true) call # 8: 23860929172501817 op/sec
Assertions.assertTrue(true) call # 9: 42107522195443421 op/sec
Assertions.assertTrue(true) call #10: 31580641109711654 op/sec
```

karayv:
@marcphilipp @sbrannen 
Probably JMH is not able to generate such load. My original issue stated 5,698,132,399 op/sec for JUnit5, and 190,251,073,367,466,500 op/sec for JUnit4. JHM reports much less ops - 3,260,943,468

I ran this code against my branch and master
```java
	@RepeatedTest(10)
	void testName(RepetitionInfo repetitionInfo) throws Exception {
		int numOfOps = Integer.MAX_VALUE;

		long start = System.nanoTime();

		for (int i = 0; i < numOfOps; i++) {
			org.junit.jupiter.api.Assertions.assertTrue(true);
		}

		long durationNanos = System.nanoTime() - start;
		long oneSecondInNanos = 1_000_000_000;

		double durationSeconds = ((double) durationNanos) / oneSecondInNanos;
		System.out.printf("Assertions.assertTrue(true) run #%2d: %.3f op/sec\n", repetitionInfo.getCurrentRepetition(),
				numOfOps / durationSeconds);
	}
```

**Results**
Current master
```
Assertions.assertTrue(true) run # 1: 46400355056.708 op/sec
Assertions.assertTrue(true) run # 2: 55528328648.706 op/sec
Assertions.assertTrue(true) run # 3: 45096466791.444 op/sec
Assertions.assertTrue(true) run # 4: 54668461121.124 op/sec
Assertions.assertTrue(true) run # 5: 54805350507.832 op/sec
Assertions.assertTrue(true) run # 6: 54677508630.837 op/sec
Assertions.assertTrue(true) run # 7: 56783579248.428 op/sec
Assertions.assertTrue(true) run # 8: 55612348987.692 op/sec
Assertions.assertTrue(true) run # 9: 51762031461.769 op/sec
Assertions.assertTrue(true) run #10: 56836239427.453 op/sec
```

This branch
```
Assertions.assertTrue(true) run # 1: 471236709432.684 op/sec
Assertions.assertTrue(true) run # 2: 681453548154.937 op/sec
Assertions.assertTrue(true) run # 3: 31375776503418.855 op/sec
Assertions.assertTrue(true) run # 4: 37675151701754384.000 op/sec
Assertions.assertTrue(true) run # 5: 32537631015151516.000 op/sec
Assertions.assertTrue(true) run # 6: 24129029741573032.000 op/sec
Assertions.assertTrue(true) run # 7: 29020049283783784.000 op/sec
Assertions.assertTrue(true) run # 8: 53687091175000000.000 op/sec
Assertions.assertTrue(true) run # 9: 31122951405797104.000 op/sec
Assertions.assertTrue(true) run #10: 35204649950819672.000 op/sec
```

Possible reason JMH gives ~3 * 10^9 op/sec is that they are using 64 bit counters. I.e. if I’m changing counter in the for loop from int to long this my benchmark degrades to the same ~3 * 10^9 op/sec.

sormuras:
> I ran this code against my branch and master [...]

I'm 99% sure that your simple loop benchmark does not generate valid (comparable) numbers. I guess the JIT compiler/optimizer eliminates the `assertTrue` call within the loop. See http://daniel.mitterdorfer.name/articles/2014/microbenchmarking-environment/ and other articles why JMH is doing a better job producing reliable numbers.

karayv:
@sormuras 
Yes, I'm sure that, as you wrote, my simple loop benchmark does not generate 100% valid numbers. However, assuming you right and the numbers are not comparable since JIT optimized the runtime, the fact that JIT is able to optimize better without creating lambdas makes sense to remove the ones that are not required to be created. 

I'm sure that numbers does not mean much for this specific merge request. I'm able to write code that never finish execution on JUnit5 and works half a second on JUnit4.

sbrannen:
> I'm able to write code that never finish execution on JUnit5 and works half a second on JUnit4.

You're comparing apples to oranges here -- actually (in your words) "billion square times" apples to 1 single orange.

See my comments here: https://github.com/junit-team/junit5/issues/1258#issuecomment-360743214

karayv:
@sbrannen 
```java
	@org.junit.jupiter.api.Test
	void testName() throws Exception {
		for (int i = 0; i < 1_000_000_000; i++) {
			for (int j = 0; j < 1_000_000_000; j++) {
				org.junit.jupiter.api.Assertions.assertTrue(() -> true);
			}
		}
	}
```
I think either I should explain better or understand better. 
The code above looks similar to my very first test. There was not lambda, it was in the guts. On my machine it never finishes, and now, with my branch if I remove lambda it finishes immediately.

sbrannen:
@karayv, thanks for all of the input!

We're looking into it...

karayv:
Regarding that test coverage decrease, I don't know how to fix it fast. There suppose to be a race condition simulated in order to cover those lines. I don't understand how they were covered before...

karayv:
@sbrannen @sormuras 
Regarding JMH. I still believe it's not suitable for your benchmarking. Take a look on a generated class `AssertionBenchmarks_junitJupiter_assertTrue_boolean_jmhTest`. Every call there is an `long` increment.

Line 181:
```java
    public static void junitJupiter_assertTrue_boolean_avgt_jmhStub(InfraControl control, RawResults result, BenchmarkParams benchmarkParams, IterationParams iterationParams, ThreadParams threadParams, Blackhole blackhole, Control notifyControl, int startRndMask, AssertionBenchmarks_jmhType l_assertionbenchmarks0_0) throws Throwable {
        long operations = 0;
//skipped
        do {
            l_assertionbenchmarks0_0.junitJupiter_assertTrue_boolean();
            operations++;
        } while(!control.isDone);
// skipped
    }
```

I actually never checked that but `long ++` should be much slower than `boolean ==`, and it looks like it's ever slower than `() -> boolean`.

I did check that `long++` is significantly slower than `int++`

sbrannen:
@karayv, it's not about `long` vs.`int` in the JMH. 

Rather, it's about the JIT being able to optimize hot paths in the code, and the JIT in our JDKs apparently does not optimize the use of lambda expressions in the current implementation of JUnit Jupiter's `assertTrue(boolean)`.

So, keep in mind that your proposed changes will not help in cases where the user actually supplies a `BooleanSupplier` or `Supplier<String>`.

documentation/src/test/java/example/AssertionsDemo.java
```diff
@@ -32,7 +32,7 @@
 	// tag::user_guide[]
 	@Test
 	void standardAssertions() {
-		assertEquals(2, 2);
+		assertEquals(2, 2, () -> "should not fail");
```
sbrannen: Please undo **all** changes made to the `documentation` project.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertEquals.java
```diff
@@ -20,161 +20,185 @@
...
- * equality on objects and primitives in tests.
+ * {@code AssertEquals} is a collection of utility methods that support
+ * asserting equality on objects and primitives in tests.
  *
  * @since 5.0
  */
 class AssertEquals {
 
-	///CLOVER:OFF
+	/// CLOVER:OFF
```
sbrannen: Please undo **all** changes made to `CLOVER` control comments.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertEquals.java
```diff
@@ -20,161 +20,185 @@
 import java.util.function.Supplier;
 
 /**
- * {@code AssertEquals} is a collection of utility methods that support asserting
- * equality on objects and primitives in tests.
+ * {@code AssertEquals} is a collection of utility methods that support
+ * asserting equality on objects and primitives in tests.
```
sbrannen: Please undo **all** changes to formatting in Javadoc.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
```diff
@@ -52,8 +52,11 @@ static void fail(String message, Object expected, Object actual) {
 		throw new AssertionFailedError(message, expected, actual);
 	}
 
-	static String nullSafeGet(Supplier<String> messageSupplier) {
-		return (messageSupplier != null ? messageSupplier.get() : null);
+	static String nullSafeGet(Object messageContainer) {
```
sbrannen: Unless I'm missing something, we should retain the original `nullSafeGet(Supplier<String>)` method and simply introduce an overloaded version that accepts a `String`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
```diff
@@ -24,38 +24,38 @@
...
+		assertIterableEquals(expected, actual, new ArrayDeque<>(), message);
 	}
 
 	static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual, Supplier<String> messageSupplier) {
 		assertIterableEquals(expected, actual, new ArrayDeque<>(), messageSupplier);
 	}
 
 	private static void assertIterableEquals(Iterable<?> expected, Iterable<?> actual, Deque<Integer> indexes,
-			Supplier<String> messageSupplier) {
+			Object messageContainer) {
```
sbrannen: Please undo **all** renamings of `messageSupplier` to `messageContainer`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
karayv: I'm adding back `nullSafeGet(Supplier<String> messageSupplier)` and keeping the overloaded version `String nullSafeGet(Object messageContainer)`. Not a `String` as you mentioned since if I make it string I'll need to copy/paste more than 11 methods. I.e. `AssertIterableEquals#assertIteratorsAreEmpty`. Please note that `Object messageContainer` is used only in private methods. The only place it's package private is `AssertionUtils.nullSafeGet(Object messageContainer)`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
```diff
@@ -71,45 +71,45 @@ private static void assertIterableEquals(Iterable<?> expected, Iterable<?> actua
...
+		fail(buildPrefix(nullSafeGet(messageContainer)) + "expected iterable was <null>" + formatIndexes(indexes));
 	}
 
-	private static void failActualIterableIsNull(Deque<Integer> indexes, Supplier<String> messageSupplier) {
-		fail(buildPrefix(nullSafeGet(messageSupplier)) + "actual iterable was <null>" + formatIndexes(indexes));
+	private static void failActualIterableIsNull(Deque<Integer> indexes, Object messageContainer) {
+		fail(buildPrefix(nullSafeGet(messageContainer)) + "actual iterable was <null>" + formatIndexes(indexes));
 	}
 
 	private static void assertIteratorsAreEmpty(Iterator<?> expected, Iterator<?> actual, int processed,
```
karayv: @sbrannen this method and some others require to be copy/pasted if there is no `nullSafeGet(Object)` utility method. What do you think?

karayv:
@sbrannen thanks for the review! 
Is there an eclipse formatter config file for your formatting style anywhere?

sormuras:
> For future, is there an eclipse formatter config file for your formatting style anywhere?

https://github.com/junit-team/junit5/tree/master/src/eclipse

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
```diff
@@ -56,6 +56,15 @@ static String nullSafeGet(Supplier<String> messageSupplier) {
...
 	}
 
+	static String nullSafeGet(Object messageContainer) {
+		// messageContainer suppose to contain a message string or a message supplier
+		// added to avoid the need to wrap message into lambda (() -> message)
+		return (messageContainer != null
+				? (messageContainer instanceof Supplier ? nullSafeGet(((Supplier<?>) messageContainer).get())
+						: messageContainer.toString())
+				: null);
+	}
```
karayv: @sbrannen  My previous comment lost. More than 11 methods need to be copy/pasted without this method with `Object` parameter. Except of this method `Object messageContainer` is present only in private methods.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
karayv: It's not just renaming but changing the parameter. I've done this in several methods. All they except of one are `private`. I would prefer to keep this change in order to avoid massive code duplication in the API module.

karayv:
All this benchmarking thing seems not so easy. I added empty method 
```java
	@Benchmark
	public void emptyMethodIndicatingMaximumPossileRateByJMH() {
		// this empty method produces best possible rate by JMH
	}
```
The results show that there is no difference:
```
Benchmark                                                              Mode  Cnt           Score           Error  Units
AssertionBenchmarks.emptyMethodIndicatingMaximumPossileRateByJMH      thrpt    7  3389434201.329 ± 101831259.659  ops/s
AssertionBenchmarks.junit4_assertTrue_String_boolean                  thrpt    7  3338075590.844 ± 562238557.565  ops/s
AssertionBenchmarks.junit4_assertTrue_boolean                         thrpt    7  3428572898.539 ±  76466904.791  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_BooleanSupplier_String    thrpt    7  3394945654.191 ±  71936443.842  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_BooleanSupplier_Supplier  thrpt    7  3419692198.527 ±  92191379.260  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean                   thrpt    7  3357105207.916 ± 570820495.330  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean_String            thrpt    7  3490230115.578 ±  46526306.056  ops/s
AssertionBenchmarks.junitJupiter_assertTrue_boolean_Supplier          thrpt    7  3467741056.969 ±  36875110.862  ops/s
```
I was trying figure out how to improve that, no ideas.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
sbrannen: OK.

I'll have to review it in detail on my machine.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionUtils.java
sbrannen: OK. I'll have a look at this as well when I review on my machine.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
sbrannen: Same as above... I'll have to review it all together on my machine to come to a conclusion.

So just leave it "as is" for now.

sbrannen:
> The results show that there is no difference:

Hmmmm... interesting. 😕

sbrannen:
Thanks for contribution, @karayv!

I've manually merged this into `master` with some additional polishing in the following commits:

- 1ac8c653305c608a99f5e8055896ae262aeeb680
- c83d6c1ba49857bb2e99b8a480ed5f854b65bb0b
- 0deced1e8248a17fc47c547a41f642aefef1464e

karayv:
@sbrannen very nice, good luck

PR #1261, opened 2018-01-25
Introduce JMH-based benchmarks

sormuras:
## Overview

Addresses https://github.com/junit-team/junit5/pull/1260#issuecomment-360400653 _"[...] I'd like to see some micro-benchmarks that verify that this is a real issue. Can you use JMH to do that?"_

---

---


sbrannen:
Awesome! 👍 

Thanks for setting up JMH in the build.

PR #1266, opened 2018-01-25
Enable error-prone static analysis checks

jbduncan:
## Overview

Following issue #955 and closed PR #961, this is a PR to introduce error-prone checks to JUnit 5.

Feedback is welcome!

---

---

### Definition of Done


jbduncan:
Hmm. My understanding of Vintage Test Engine is very limited, to say the least, so I'm stumped as to why one of its corresponding tests is failing.

Any thoughts would be appreciated. :)

sormuras:
You only changed a single file in "vintage": https://github.com/junit-team/junit5/pull/1266/files#diff-fbff14085969788bb222d28b4e767dee

Does reverting the comparison help?

jbduncan:
@sormuras It does, actually! I'll push a new commit with that section reverted.

sormuras:
See the javadoc of `lookupTestDescriptor(Description description)`:

```
<p>There are edge cases where multiple {@link Description Descriptions}
with the same {@code uniqueId} exist, e.g. when using overloaded methods
to define {@linkplain org.junit.experimental.theories.Theory theories}.
In this case, we try to find the correct {@link TestDescriptor} by
checking for object identity on the {@link Description} it represents.
```

jbduncan:
Oh, doh! Thanks @sormuras. :P 

I must've made that change thinking that error-prone was going to complain about the usage of `==`. Reverting soon. :)

jbduncan:
Welp, I'm thoroughly stumped once again. For some reason, task `compileTestJavaWithErrorProne` is apparently not able to find classes written in Kotlin, and I don't know where I can find help to resolve the error.

@sormuras Do you have any ideas as to where I could seek help, or any other thoughts?

JLLeitschuh:
@jbduncan There are two things that I've discovered when working with Kotlin that might help you.

  1. The Kotlin compiler depends upon the Java AST to resolve the symbols. The Java compiler depends upon the compiled Kotlin bytecode to compile the Java code. So, the Java compiler needs to have the Kotlin bytecode added as an input so it can compile everything.
 2. If this project is using the `java-library` plugin, then there was a time (this might not be a problem anymore) where Kotlin Code wasn't always included as an output of a project.
  You had to add this hack to make it actually work (this uses Gradle Kotlin DSL but you can do the same in Groovy).
    ```kotlin
    val compileKotlin: KotlinCompile by tasks
    afterEvaluate {
        /*
         * Needed to configure kotlin to work correctly with the "java-library" plugin.
         * See:
         * https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_known_issues
         */
        pluginManager.withPlugin("java-library") {
            configurations {
                "apiElements" {
                    outgoing
                        .variants
                        .getByName("classes")
                        .artifact(mapOf(
                            "file" to compileKotlin.destinationDir,
                            "type" to "java-classes-directory",
                            "builtBy" to compileKotlin
                        ))
                }
            }
        }
    }
    ```

jbduncan:
@JLLeitschuh Wow, thanks a lot for the pointers! :+1:  Both of the approaches you listed look promising to me, so I'll look into them when I'm next free.

jbduncan:
@JLLeitschuh Many thanks for your help! As you suspected, the error-prone Java compiler needed the .class files produced by the Kotlin compiler, so after doing some digging around in the sources of the Gradle Kotlin Plugin, I discovered that adding the .class files to the classpath of the error-prone Java compiler fixes things.

Cheers! :tada:

jbduncan:
@sormuras The JDK 10 build on Travis is complaining about being unable to find `annotation method 'value()' in type 'Profile+Annotation': class file for jdk.Profile+Annotation not found` when building `:junit-platform-commons-java-9:compileJava`.

I've done a Google search of `jdk.Profile+Annotation`, but I could only find this GitHub link: https://github.com/ceylon/ceylon-compiler/issues/1347, which doesn't seem to help.

It's not clear to me if this is happening because of the changes I made to `build.gradle` or if it's a very obscure bug in Gradle or JDK 10. Do you have any insight or other thoughts on this?

sormuras:
I have no idea.

jbduncan:
Unless if anyone else in the JUnit 5 team has encountered this sort of error before (https://travis-ci.org/junit-team/junit5/jobs/345033831), I wonder if the best course of action in the meantime would be to make it optional for the JDK 10 build to pass.

WDYT @sormuras?

jbduncan:
@sormuras I believe that the error is caused by error-prone issue https://github.com/google/error-prone/issues/860.

error-prone uses a pretty recent Java compiler, but not the absolute bleeding-edge one, so it doesn't yet recognise Java 10 classfiles. I think that's the reason why Travis's JDK 10 build is producing strange messages. For that reason, I'm opting to only enable error-prone when running on JDK 9, and I'll work on making that change next.

jbduncan:
I'm ready for another review whenever the JUnit team's ready. :)

marcphilipp:
@jbduncan Thanks for keeping this PR up-to-date and sorry if we've been stringing you along. My gut feeling is that using error-prone for this project might be more trouble than it's worth. We have to make sure that JUnit always works with the latest JDK, even EA releases. Naturally, error-prone lacks behind in that regard. Thus, I think we should abandon this effort, at least for now.

jbduncan:
@marcphilipp No worries! I was aware that JUnit 5 is tested against the latest JDK releases, so I knew that there was risk that this PR wouldn't be accepted; I was keeping it up-to-date as I find this sort of thing enjoyable.

(And I also secretly hoped that the fact that error-prone cannot keep up with the EA releases would not have been that big a problem. :wink:)

If it is decided that this effort should be abandoned, then I'll accept that decision with grace. :)

JLLeitschuh:
Isn't error prone a fork of the eclipse java compiler?

@jbduncan I'm sorry to hear this might get canned, I know you put a significant amount of effort into it.

jbduncan:
@JLLeitschuh IIRC error-prone is actually a fork of OpenJDK's javac.

jbduncan:
@JLLeitschuh I'd be interested to understand why you ask, though. :)

mmerdes:
@marcphilipp
I agree with you.

sbrannen:
Unfortunately _("unfortunate" because of the amount of effort invested in this by @jbduncan)_, I have to agree with @marcphilipp as well: it appears that staying on top of issues with error prone will be... well... _error prone_ for the maintainers of the build.

In any case, thanks for all of the hard work, @jbduncan!

And... I'm glad you at least found the work "enjoyable". 😉

marcphilipp:
I'm closing this PR since the team seems to be in consensus on this topic.

@jbduncan Sorry again and thanks for trying to convince us! I'm sure we can find another issue for you to work on if you want. 😉

jbduncan:
> And... I'm glad you at least found the work "enjoyable". :wink:

Well, there were times where I got maddeningly stuck, as I think you already knew... :wink:

...but overall I found the process very enjoyable, even though it was just a dead end all along. :grin:

jbduncan:
@marcphilipp Thanks! I'll likely decide to go back to https://github.com/junit-team/junit5/issues/955 in the foreseeable future, either to choose another static analysis tool if any exist that follow the latest stable releases and/or EA releases, or to close the issue altogether if no such tools apparently exist. :)

JLLeitschuh:
PMD as an alternative?

> I'd be interested to understand why you ask, though.

Because I've found that the eclipse compiler is a PITA as well. It can't handle generic inference correctly in many cases. Wondering if I was going to deal with these same issues moving some of my projects forward to JDK 9 support.

PR #1272, opened 2018-01-30
Introduce @DisabledIf annotation

sormuras:
## Overview

Introduce `@DisabledIf` annotation.

- Merge logic of `@EnabledIf` and `@DisabledIf` execution condition into a single implementation.
- Compile and cache compiled script instances if the selected script engine implements `javax.script.Compilable`

#### Examples

```java
@Test
@DisabledIf("java.lang.System.getProperty('os.name').toLowerCase().contains('win')")
void notOnMicrosoftWindows() {
	assertTrue(!System.getProperty("os.name").toLowerCase().contains("win"));
}

@Test
@DisabledIf("java.lang.System.getProperty('os.name').toLowerCase().contains('mac')")
void notOnMacOs() {
	assertTrue(!System.getProperty("os.name").toLowerCase().contains("mac"));
}

```
---

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/Constants.java
```diff
@@ -89,4 +92,106 @@ private Constants() {
...
 	}
 
+	/**
+	 * Script-related constants used by {@link DisabledIf} and {@link EnabledIf}.
+	 *
+	 * <p>Declares names used for script {@link javax.script.Bindings bindings}
+	 * and also placeholder tags usable in custom reason message.
+	 */
+	@API(status = EXPERIMENTAL, since = "5.1")
+	public interface Script {
```
jbduncan: Hey @sormuras. I wonder if you can explain to me why you made `Script`, `Bind` and `Reason in this class sub-interfaces? I ask because AFAICT they are just being used to store constants and not being extended anywhere, which I understand to be an anti-pattern according to Effective Java 2nd Edition, Item 19.

If they are indeed being used just to store constants, I would recommend changing them to `public static final class`es , or even better package-private `static final class`es, and prepending the constants with `public static final` or just `static final` as needed.

jbduncan: ...and implementing empty private constructors for each class like `private Script() {}`.

sormuras: Bad old habits prevail. :)

I moved the generic script-related types into `org.junit.jupiter.engine.script` package. The constants are now part of the `Script` class.

jbduncan: Okey dokey! Thank you very much for addressing this. :)

sbrannen: Good catch, @jbduncan... I was going to make comments along the same lines.

Issue #1282, opened 2018-02-04
Document module path requirements for running junit-jupiter-engine


marcphilipp:
Document that junit-jupiter-engine does not ship with a module descriptor but requires java.scripting and can be run using `--add-modules ALL-MODULE-PATH,ALL-DEFAULT`.

sormuras:
This is the stacktrace produced on OpenJDK 10-ea+42:
```
JUnit Jupiter
    => java.lang.NoClassDefFoundError: javax/script/ScriptException
       org.junit.jupiter.engine@5.1.0-M2/org.junit.jupiter.engine.extension.ExtensionRegistry.<clinit>(ExtensionRegistry.java:54)
       org.junit.jupiter.engine@5.1.0-M2/org.junit.jupiter.engine.descriptor.JupiterEngineDescriptor.prepare(JupiterEngineDescriptor.java:37)
       org.junit.jupiter.engine@5.1.0-M2/org.junit.jupiter.engine.descriptor.JupiterEngineDescriptor.prepare(JupiterEngineDescriptor.java:28)
       org.junit.platform.engine@1.1.0-M2/org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor$NodeExecutor.prepare(HierarchicalTestExecutor.java:89)
       org.junit.platform.engine@1.1.0-M2/org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor$NodeExecutor.execute(HierarchicalTestExecutor.java:74)
       org.junit.platform.engine@1.1.0-M2/org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:55)
       org.junit.platform.engine@1.1.0-M2/org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:43)
       org.junit.platform.launcher@1.1.0-M2/org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:170)
       org.junit.platform.launcher@1.1.0-M2/org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:154)
       org.junit.platform.launcher@1.1.0-M2/org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:90)
```

sormuras:
[...]

PR #1286, opened 2018-02-06
Make dependency on "javax.script" optional

sormuras:
## Overview

Copied from http://mail.openjdk.java.net/pipermail/jigsaw-dev/2018-February/013581.html
> [...] The more general issue is of course that you are 
deploying a module that does not know what modules it depends. In these 
cases you have to help the module system by specifying --add-modules to 
force the modules that it depends on to be resolved. You got lucky with 
the Oracle JDK build because jdk.deploy pulled in the java.scripting module.

Addresses #1282

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
@@ -22,8 +22,8 @@ import org.junit.jupiter.api.condition.DisabledIf;
 import org.junit.jupiter.api.condition.EnabledIf;
 import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junit.jupiter.api.extension.ExtensionContext;
-import org.junit.jupiter.api.extension.ScriptEvaluationException;
 import org.junit.jupiter.engine.script.Script;
 
 /**
@@ -160,14 +160,14 @@ class ScriptExecutionCondition implements ExecutionCondition {
 	}
 
 	/**
-	 * Evaluator implementation that always throws an {@link ScriptEvaluationException}.
+	 * Evaluator implementation that always throws an {@link ExtensionConfigurationException}.
 	 */
 	static class ThrowingEvaluator implements Evaluator {
 
-		final ScriptEvaluationException exception;
+		final ExtensionConfigurationException exception;
 
 		ThrowingEvaluator(String message, Throwable cause) {
-			this.exception = new ScriptEvaluationException(message, cause);
+			this.exception = new ExtensionConfigurationException(message, cause);
 		}
 
 		@Override
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
@@ -10,7 +10,6 @@
 
 package org.junit.jupiter.engine.extension;
 
-import static org.junit.jupiter.api.extension.ConditionEvaluationResult.disabled;
 import static org.junit.jupiter.api.extension.ConditionEvaluationResult.enabled;
 import static org.junit.platform.commons.util.AnnotationUtils.findAnnotation;
 
@@ -19,20 +18,13 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
-import javax.script.Bindings;
-import javax.script.ScriptException;
-import javax.script.SimpleBindings;
-
 import org.junit.jupiter.api.condition.DisabledIf;
 import org.junit.jupiter.api.condition.EnabledIf;
 import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
-import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
 import org.junit.jupiter.api.extension.ScriptEvaluationException;
 import org.junit.jupiter.engine.script.Script;
-import org.junit.jupiter.engine.script.ScriptAccessor;
-import org.junit.jupiter.engine.script.ScriptExecutionManager;
 
 /**
  * {@link ExecutionCondition} that supports the {@link DisabledIf} and {@link EnabledIf} annotation.
@@ -48,9 +40,19 @@ class ScriptExecutionCondition implements ExecutionCondition {
 
 	private static final ConditionEvaluationResult ENABLED_NO_ANNOTATION = enabled("Annotation not present");
 
-	private static final ConditionEvaluationResult ENABLED_ALL = enabled("All results are enabled");
+	private static final String EVALUATOR_CLASS_NAME = "org.junit.jupiter.engine.extension.ScriptExecutionEvaluator";
 
-	private static final Namespace NAMESPACE = Namespace.create(ScriptExecutionCondition.class);
+	private final Evaluator evaluator;
+
+	// Used by the ExtensionRegistry.
+	ScriptExecutionCondition() {
+		this(EVALUATOR_CLASS_NAME);
+	}
+
+	// Used by tests.
+	ScriptExecutionCondition(String evaluatorImplementationName) {
+		this.evaluator = Evaluator.forName(evaluatorImplementationName);
+	}
 
 	@Override
 	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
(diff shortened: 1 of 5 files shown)


sbrannen:
> WORK-IN-PROGRESS

OK... just let us know when we are allowed to comment.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
```diff
@@ -111,66 +91,23 @@ private String createSource(String[] lines) {
...
+		// Create worker via reflection to hide the `java.scripting` dependencies.
 		try {
-			Object result = manager.evaluate(script, bindings);
-			return computeConditionEvaluationResult(script, result);
+			return (Worker) Class.forName(
+				"org.junit.jupiter.engine.extension.ScriptExecutionWorker").getDeclaredConstructor().newInstance();
 		}
-		catch (ScriptException e) {
-			throw new ScriptEvaluationException("Script evaluation failed for: " + script.getAnnotationAsString(), e);
+		catch (ReflectiveOperationException e) {
```
marcphilipp: I think we should make this more explicit, i.e. check whether a class from `javax.script` is present. Moreover, we should provide a hint to the user, possibly a link to the User Guide, of how to fix the setup.

sormuras: > WORK-IN-PROGRESS

:)

marcphilipp: Well, if you don't want "helpful suggestions", don't open a PR... 😜

sbrannen: *+1* 😝

sormuras: 🤓🤓

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
```diff
@@ -111,66 +98,70 @@ private String createSource(String[] lines) {
...
 	}
 
-	ConditionEvaluationResult computeConditionEvaluationResult(Script script, Object result) {
-		// Treat "null" result as an error.
-		if (result == null) {
-			throw new ScriptEvaluationException("Script returned `null`: " + script.getAnnotationAsString());
+	// This method may fail on JREs that don't provide the "javax.script" package at all
+	// or it fails on JREs launched with an active module system using insufficient module
+	// graphs, i.e. the application does not read "java.scripting" module.
+	Optional<Throwable> assumeScriptEngineIsLoadableByClassForName() {
```
sbrannen: Everything related to the "worker detection" that uses reflection needs to be performed once in a static initialization block.

Rationale: if we can't load a class from `javax.script` when the condition class is loaded, we won't be able to load it later either. Plus the workers should be stateless, so creating a single one eagerly as a static field should not be a problem.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
```diff
@@ -111,66 +98,70 @@ private String createSource(String[] lines) {
...
+			return Optional.of(cause);
 		}
+	}
 
-		String resultAsString = String.valueOf(result);
-		String reason = script.toReasonString(resultAsString);
+	/**
+	 * Evaluates scripts and returns a conditional evaluation result.
+	 */
+	interface Worker {
```
sbrannen: From my perspective, the term `Worker` implies that we are doing something with forked processes or threads; whereas, as in your JavaDoc... this is something that "Evaluates scripts". 

So, I'd simply call it a `ScriptEvaluator` analogous to the existing `ConditionEvaluator`. 😉

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
```diff
@@ -98,57 +103,53 @@ private String createSource(String[] lines) {
...
+		/**
+		 * Create evaluator via reflection to hide the `javax.script` dependency.
+		 */
+		static Evaluator forName(String name) {
+			Optional<Throwable> optionalThrowable = assumeScriptEngineIsLoadableByClassForName();
+			if (optionalThrowable.isPresent()) {
+				return new ThrowingEvaluator(optionalThrowable.get());
+			}
+			try {
+				return (Evaluator) Class.forName(name).getDeclaredConstructor().newInstance();
```
sbrannen: I think it might be better to use `getClass().getClassLoader().loadClass(name)...`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
```diff
@@ -98,57 +103,53 @@ private String createSource(String[] lines) {
...
+		}
+
+		/**
+		 * This method may fail on JREs that don't provide the "javax.script" package at all
+		 * or it fails on JREs launched with an active module system using insufficient module
+		 * graphs, i.e. the application does not read "java.scripting" module.
+		 */
+		static Optional<Throwable> assumeScriptEngineIsLoadableByClassForName() {
+			try {
+				Class.forName("javax.script.ScriptEngine");
```
sbrannen: ditto

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
```diff
@@ -98,57 +103,53 @@ private String createSource(String[] lines) {
...
+				return new ThrowingEvaluator(cause);
+			}
+		}
+
+		/**
+		 * This method may fail on JREs that don't provide the "javax.script" package at all
+		 * or it fails on JREs launched with an active module system using insufficient module
+		 * graphs, i.e. the application does not read "java.scripting" module.
+		 */
+		static Optional<Throwable> assumeScriptEngineIsLoadableByClassForName() {
```
sbrannen: maybe change the method name to `...ViaReflection` instead of `ByClassForName`...

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
sormuras: Using the following new _private_ method for both cases:

		static Class<?> load(String name) throws ClassNotFoundException {
			return Evaluator.class.getClassLoader().loadClass(name);
		}

sormuras: Or revert to the initial code as it says in [Class.forName(String)](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#forName-java.lang.String-):
```
Returns the Class object associated with the class or interface with the given string name. Invoking this method is equivalent to:
  Class.forName(className, true, currentLoader)
where currentLoader denotes the defining class loader of the current class.
```

sbrannen: sure 😛

sbrannen: Or switch to `org.junit.platform.commons.util.ReflectionUtils.loadClass(String)` for consistency. 😇

sbrannen: You pick!

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
```diff
@@ -62,115 +64,116 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
...
+		final ScriptEvaluationException exception;
 
-		// Flip enabled/disabled result based on the associated annotation type.
-		if (script.getAnnotationType() == EnabledIf.class) {
-			return isTrue ? enabled(reason) : disabled(reason);
-		}
-		if (script.getAnnotationType() == DisabledIf.class) {
-			return isTrue ? disabled(reason) : enabled(reason);
+		ThrowingEvaluator(String message, Throwable cause) {
+			this.exception = new ScriptEvaluationException(message, cause);
```
sormuras: Pondering the idea to introduce a new `JUnitException` specialization here. Something like `ScriptingConfigurationException`. As there is factually no "evaluation" going on.

sbrannen: Hmmmm... if evaluation is not the cause, maybe it's best just to throw an `ExtensionConfigurationException` and get rid of the `ScriptEvaluationException` subclass (if it then becomes unused).

sbrannen: Or are you saying you want `ScriptingConfigurationException` to replace `ScriptEvaluationException`?

sormuras: Going to use `ExtensionConfigurationException` in the `ThrowingEvaluator`. The `ScriptEvaluationException` is still used for when a user-defined script contains (syntax) errors.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
```diff
@@ -62,115 +64,116 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
...
-	ConditionEvaluationResult computeConditionEvaluationResult(Script script, Object result) {
-		// Treat "null" result as an error.
-		if (result == null) {
-			throw new ScriptEvaluationException("Script returned `null`: " + script.getAnnotationAsString());
+		static Evaluator forName(String nameOfScriptEngine, String name) {
+			// Assert that precondition name is loadable via basic reflection.
+			try {
+				Class.forName(nameOfScriptEngine);
+			}
+			catch (Throwable cause) {
```
marcphilipp: Do we really want to catch `Throwable` here? How about `ClassNotFoundException`?

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ScriptExecutionConditionTests.java
```diff
@@ -143,23 +83,62 @@ void annotationDefaultValues(TestInfo info) {
...
+	@Test
+	void throwingEvaluatorIsCreatedWhenScriptEngineIsNotAvailable() {
+		String nameOfScriptEngine = "javax.script.DoesNotExist";
+		String name = "org.junit.jupiter.engine.extension.ScriptExecutionEvaluator";
+		ScriptExecutionCondition.Evaluator evaluator = ScriptExecutionCondition.Evaluator.forName(nameOfScriptEngine,
+			name);
+		Exception e = assertThrows(Exception.class, () -> evaluator.evaluate(null, null));
+		assertTrue(e instanceof ExtensionConfigurationException);
+		String message = e.getMessage();
+		System.out.println(message);
```
marcphilipp: Please remove the `println`.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ScriptExecutionConditionTests.java
```diff
@@ -143,23 +83,62 @@ void annotationDefaultValues(TestInfo info) {
...
+		assertTrue(message.startsWith("Class `" + nameOfScriptEngine + "` is not loadable"));
+		assertTrue(message.endsWith("`--add-modules ...,java.scripting`"));
```
marcphilipp:
```java
assertThat(e).isInstanceOf(ExtensionConfigurationException.class)
		.hasMessageStartingWith("Class `" + nameOfScriptEngine + "` is not loadable")
		.hasMessageEndingWith("`--add-modules ...,java.scripting`")
```

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ScriptExecutionConditionTests.java
```diff
@@ -143,23 +83,62 @@ void annotationDefaultValues(TestInfo info) {
...
+	@Test
+	void throwingEvaluatorIsCreatedWhenDefaultEvaluatorClassNameIsIllegal() throws ReflectiveOperationException {
+		String name = "illegal class name";
+		ScriptExecutionCondition condition = new ScriptExecutionCondition(name);
+		ExtensionContext context = Mockito.mock(ExtensionContext.class);
+		AnnotatedElement element = SimpleTestCases.class.getDeclaredMethod("testIsEnabled");
+		Mockito.when(context.getElement()).thenReturn(Optional.of(element));
+		Exception e = assertThrows(Exception.class, () -> condition.evaluateExecutionCondition(context));
+		assertTrue(e instanceof ExtensionConfigurationException);
+		assertTrue(e.getMessage().startsWith("Creating instance of class `" + name + "` failed"));
```
marcphilipp: Ditto.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
sormuras: Yes we do. The throwable that raised the issue was an instance of "java.lang.NoClassDefFoundError: javax/script/ScriptException" -- `extends LinkageError extends Error extends Throwable`.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ScriptExecutionConditionTests.java
sormuras: How did that survive? Consider it gone.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ScriptExecutionConditionTests.java
sormuras: I like the one assertion per statement style. Will change it. :)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/ScriptExecutionConditionTests.java
sormuras: Here too.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java
marcphilipp: I never understood the rationale for having `NoClassDefFoundError` _and_ `ClassNotFoundException`...

Just to be safe, we should call `BlacklistedExceptions.rethrowIfBlacklisted(cause)` to not accidentally swallow an `OutOfMemoryError`.

sormuras: Aye. Good catch!

sbrannen: Yeah... `NoClassDefFoundError` and `ClassNotFoundException` are confusing and have horribly ambiguous names. (sigh)

`ClassNotFoundException` would be thrown if we were using reflection to load the `ScriptException` class; whereas, in our case we actually compiled against it without errors. So `NoClassDefFoundError` is thrown to signal that the dependency (in the class that we are loading without reflection) cannot be found at runtime.

In any case, I usually forget the difference as well and have to look it up... like [here](http://www.baeldung.com/java-classnotfoundexception-and-noclassdeffounderror). 😉

sbrannen: +1 for using `BlacklistedExceptions` though!

PR #1291, opened 2018-02-10
allow to disable reporting

robstoll:
## Overview

This is only a hack, but I hope it shows what the intention is. I want to be able to disable the report entirely. Currently I modify `junitPlatformTest` as follows which is even a bigger hack:

```groovy
def reportIndex = junitPlatformTestTask.args.findIndexOf { it=='--reports-dir' }
def keep = (0..(junitPlatformTestTask.args.size()-1)) - [reportIndex, reportIndex + 1]
junitPlatformTestTask.args = junitPlatformTestTask.args[keep]
```

---

---

### Definition of Done

### DIFF
--- a/junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPlugin.groovy
+++ b/junit-platform-gradle-plugin/src/main/groovy/org/junit/platform/gradle/plugin/JUnitPlatformPlugin.groovy
@@ -187,8 +187,10 @@ class JUnitPlatformPlugin implements Plugin<Project> {
 			args.addAll('--config', "${key}=${value}")
 		}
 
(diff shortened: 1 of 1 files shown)


sbrannen:
Hi @robstoll,

Thanks for raising this issue.

I agree: that's definitely a hack.

If we were to introduce support for disabling report generation, we would want to do it via an explicit flag similar to how the `test` task in core Gradle works -- for example, something like the following.

```groovy
reports.junitXml.enabled = false
```

However, due to the fact that Gradle will provide first-class support for the JUnit Platform in Gradle 4.6 (see this merged [PR](https://github.com/gradle/gradle/pull/4116)), I don't think we will implement such a "disable report generation" feature.

Regards,

Sam

Issue #1309, opened 2018-02-24
Provide support of parameterized test arguments descriptions


dmitry-timofeev:
## Overview

This is a *feature request* and a follow-up to the discussion in #1306. Currently if a user needs to add a description of a set of arguments (e.g., because their string representation is not enough, or the differences between adjacent arguments are subtle), they need to (ab)use the `Arguments` by adding an additional one. This has a couple of drawbacks:
- That argument must be specified in the test method signature. Although it may be given a descriptive name to clarify its purpose (e.g., "testDescription"), it will still be highlighted by an IDE as _unused_ (see #1306).
- A proper workaround is ugly: `@SuppressWarnings("unused")  // Used in the test name above`
- Such argument is misleading, as it isn't a test _parameter_, but a meta-data about test parameters. It shall not affect the test results.

### Example
```java
import static org.junit.jupiter.api.Assertions.assertThrows;

import java.util.stream.Stream;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.Arguments;
import org.junit.jupiter.params.provider.MethodSource;

class SampleTest {

  @ParameterizedTest(name = "{1}: {0}")
  @MethodSource("testParameters")
  void test(String s,
            @SuppressWarnings("unused")  // Used in the test name above
            String testDescription) {
    assertThrows(NumberFormatException.class,
        () -> Integer.parseInt(s)
    );
  }

  static Stream<Arguments> testParameters() {
    return Stream.of(
        Arguments.of("1a", "illegal character"),
        Arguments.of("10000000000000", "out of bounds")
    );
  }
}
```

## Possible solutions
As suggested by @sormuras  in the [discussion](https://github.com/junit-team/junit5/issues/1306#issuecomment-368229340), an optional attribute may be added to `Arguments`.

## Alternatives
- #1306, but it will enable users to write confusing code (= reference some arguments that do not appear in the method parameters by index).

## Questions
**TBD**

## Deliverables

- [ ] **TBD**

marcphilipp:
According to the discussion in #1306, the new API would look like `Arguments.of(...).describedAs("")` or something similar. How would you use it? Via a special `{description}` placeholder for `@ParameterizedTest(name="{description}: ...")`? If so, what do we use for `Arguments` instances without a description?

sormuras:
Default to `lorem ipsum dolor sit amet`... ;)

[...]

PR #1310, opened 2018-02-24
WIP of #1309: Add support for a description of Arguments.

dmitry-timofeev:
## Overview

This is a WIP PR for #1309. Some documentation is _deliberately_ missed, because we haven't yet agreed on the approach to go with, and on some implementation details.

Feedback is welcome! Perhaps some of TODOs are better to be discussed in the [issue](https://github.com/junit-team/junit5/issues/1309), not in the PR.

---

---
### TODO
- [ ] CsvSource & CsvFileSource (need to agree on requirements, see the issue).
- [ ] Consider lazy `Arguments#get` (a decorator) and more factory methods (depends on ^, see the [thread](https://github.com/junit-team/junit5/pull/1310#discussion_r170445795))
- [ ] Test description is preserved when arguments are stripped.
- [ ] Default `ParameterizedTest#name` and how it is expanded (need to agree on requirements, see the issue).

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
@@ -72,6 +72,8 @@ public @interface ParameterizedTest {
 	 * <ul>
 	 * <li><code>{index}</code>: the current invocation index (1-based)</li>
 	 * <li><code>{arguments}</code>: the complete, comma-separated arguments list</li>
+	 * // todo: link Arguments.getDescription?
+	 * <li><code>{arguments.description}</code>: a description of the arguments</li>
 	 * <li><code>{0}</code>, <code>{1}</code>, etc.: an individual argument</li>
 	 * </ul>
 	 *
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
@@ -73,7 +73,6 @@ public @interface ParameterizedTest {
 	 * <ul>
 	 * <li><code>{index}</code>: the current invocation index (1-based)</li>
 	 * <li><code>{arguments}</code>: the complete, comma-separated arguments list</li>
-	 * // todo: link Arguments.getDescription?
 	 * <li><code>{arguments.description}</code>: a description of the arguments</li>
 	 * <li><code>{0}</code>, <code>{1}</code>, etc.: an individual argument</li>
 	 * </ul>
@@ -83,6 +82,7 @@ public @interface ParameterizedTest {
 	 *
 	 * @see java.text.MessageFormat
 	 */
+	// todo: link Arguments.getDescription to {arguments.description}?
 	//todo: consider either changing the default name or its interpretation:
 	// arguments = all arguments + arguments.description
 	String name() default "[{index}] {arguments}";
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
@@ -17,7 +17,6 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-
 import org.apiguardian.api.API;
 import org.junit.jupiter.api.TestTemplate;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -81,6 +80,8 @@ public @interface ParameterizedTest {
 	 *
 	 * @see java.text.MessageFormat
 	 */
+	//todo: consider either changing the default name or its interpretation:
+	// arguments = all arguments + arguments.description
 	String name() default "[{index}] {arguments}";
 
 }
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
@@ -17,6 +17,7 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+
 import org.apiguardian.api.API;
 import org.junit.jupiter.api.TestTemplate;
 import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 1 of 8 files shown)


junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
```diff
@@ -30,17 +30,19 @@
...
 
-	private String prepareMessageFormatPattern(int invocationIndex, Object[] arguments) {
+	private String prepareMessageFormatPattern(int invocationIndex, Arguments arguments) {
+		// todo: shall it be split into three methods for clarity?
 		String result = namePattern.replace("{index}", String.valueOf(invocationIndex));
+		result = result.replace("{arguments.description}", arguments.description());
 		if (result.contains("{arguments}")) {
 			// @formatter:off
-			String replacement = IntStream.range(0, arguments.length)
+			String replacement = IntStream.range(0, arguments.get().length)
```
marcphilipp: I'd like to avoid calling `get()` multiple times since it may be costly to compute the array.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -38,13 +38,47 @@
...
+//	default boolean hasDescription() {
+//		return !description().isEmpty();
+//	}
+
+	// todo: @MustCheckReturnValue
+	default Arguments description(String testCaseDescription) {
+		Preconditions.notBlank(testCaseDescription,
+				() -> "Test case description must not be blank: '" + testCaseDescription + "'!");
+
+		Object[] arguments = get();
```
marcphilipp: I don't think we should call `get()` here. We could return a decorator instead.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
```diff
@@ -81,6 +80,8 @@
 	 *
 	 * @see java.text.MessageFormat
 	 */
+	//todo: consider either changing the default name or its interpretation:
+	// arguments = all arguments + arguments.description
 	String name() default "[{index}] {arguments}";
```
marcphilipp: How about `"[{index}] {arguments} {description}"` and trimming the name so an empty description would be ignored?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -38,13 +38,47 @@
 	 */
 	Object[] get();
 
+	// todo: document properly this and below.
+	// todo: The fact there used to be `hasDescription` indicates we might be better of with Optional<String>
+	default String description() {
```
marcphilipp: Let's go with `Optional<String>`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
dmitry-timofeev: Sure can do, although I'm not sure I understand the reasoning. `Arguments.of` captures its variable arity parameter (= a reference to an `Object[]`), and `Arguments.get` return the same reference to the same object. The current code does not perform any defensive copies because they seem to be redundant in this case (it is never OK to modify an underlying varargs parameter at call-site; regarding internal, package-private code -- we can trust it).

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
dmitry-timofeev: May I ask why? Do you think it would be simpler, more readable, faster? `get()` simply returns a reference to the underlying array.

dmitry-timofeev:
@marcphilipp , Thank you for the early feedback! I've replied to code-related issues here, and to UX-related -- in the issue.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
dmitry-timofeev: I've replied in the issue, I currently believe it's better to expand `{arguments}` into two patterns, depending on whether an instance of `Arguments` has a description.

sormuras:
Please run `gradlew spotlessApply` before committing. That'll will fix the CI builds.

dmitry-timofeev:
@sormuras , sure, thanks. I have a question though: it usually not a goal to write a minimum number of lines, but to write a readable code (Google Style Guide even says that explicitly). Spotless tends to condense chain calls into a single line, resulting in an overly complicated expression (e.g., `new Builder().setFoo("foo").setBar(5).map(/* … */).build();`. How do you handle this?

marcphilipp:
Spotless uses the Eclipse formatter which unfortunately does not support chained method calls very well.

Here's what we usually do:
https://github.com/junit-team/junit5/blob/6befdacb3abd44759bf3a3948d1b8a5c601e20c3/junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java#L268-L273

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
marcphilipp: I may be a bit too defensive here, but since it's an interface people might implement it and compute the array lazily. It's a bit like a `Supplier`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
marcphilipp: See above comment.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
dmitry-timofeev: I see, thank you, I'll think how to express it better.

Just to confirm, `Arguments` contract does not allow implementers to return _different_ arrays in subsequent calls, does it? That is, I can preserve whatever the first call returns? If not, can we restrict it in the spec?

dmitry-timofeev: And a couple of things to consider:
 1. Does it ever happen (if all tests are correct) that JUnit does not consume all of the arguments? I might recall a recent Run-test-by-unique-id feature :thinking: 
 2. Does it matter in terms of prospective parallel execution _where_ we actually consume the arguments?

marcphilipp:
Closing for the time being since #1309 is "waiting-for-interest".

Issue #1191, opened 2017-11-29
Support conversion of multiple @CsvFileSource parameters into one object


klassounski-daitan:
## Overview

Hello there Java testers.
I started working with JUnit 5 a few days ago because I loved the new way of creating parameterized tests.

There's the `@ParameterizedTest` that allows a test to run in a parameterized fashion and `@CsvFileSource` that loads the parameters from a CSV file.

The thing is that I have too much columns in my CSV and don't want to have a huge method signature in my unit test. Let me give you an example:

    @ParameterizedTest
    @CsvFileSource(resources = "/test-data.csv")
    void myTest(String p1, String p2, String p3, String p4, String p5, String p6) {
        //test using parameters
    }

It would be awesome to have some kind of annotation that would allow us to do something like this:

    @ParameterizedTest
    @CsvFileSource(resources = "/test-data.csv")
    void myTest(@ConvertWith(TestDataConverter.class) TestData testData) {
        //test using parameters
    }
    
    static class TestDataConverter implements TestConverter{
        public TestData convert(Object ... params){
           //a simple method that creates the TestData object and inserts the params in it
        }
    }

I believe that this improvement would be much welcomed by our community.

Happy coding and stay awesome!

sbrannen:
Assigned to 5.1 backlog for team discussion.

sbrannen:
FYI: I updated the title of this issue to reflect its more generic nature.

sbrannen:
Hi @hglasgow,

Thanks for submitting PR #1290.

Unfortunately that is not the direction we would want to go.

If we do anything along these lines, it would have to be a general-purpose `ArgumentConverter` that can be applied to the arguments for any `@ParameterizedTest`. In other words, we would not limit the support to just `@CsvFileSource`. Similarly, we would not bake the support into a single `ArgumentsProvider` implementation. 

For example, I could imagine introducing a sub-interface of `ArgumentConverter` (perhaps something like an `AggregatingArgumentConverter`) that is handled consistently across all `ArgumentsProvider` implementations (i.e., something more along the lines of what @klassounski-daitan proposed in this issue's description).

Regards,

Sam

geo-m:
Greetings @sbrannen 

[...]

PR #1325, opened 2018-03-11
Support aggregation of multiple arguments into a single object for parameterized tests

geo-m:
## Overview

Draft implementation of the proposed API for #1191 

Issue: #1191

---

---

### Definition of Done


sbrannen:
Thanks for the PR!

Unfortunately, I don't think this is the route we want to go.

So, in order to clarify things, I'll post more information to #1191.

It'll then be easier to discuss the overall feature set there in case you scrap or totally rework this PR.

geo-m:
@sbrannen Great, i will make an implementation attempt according to your specifications in #1191 and if that still doesnt meet your requirements you can close this PR.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/AggregateWith.java
```diff
@@ -0,0 +1,24 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.PARAMETER)
+public @interface AggregateWith {
+	Class<? extends ArgumentsAggregator> value();
+
+	int[] range() default {};
```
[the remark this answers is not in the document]
geo-m: Do you mean the Range attribute? That was accidental sorry, i implemented a range feature but decided to wait for your input

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
```diff
@@ -0,0 +1,80 @@
...
+
+	public Double getDouble(int index) {
+		return (Double) getAs(Double.class, index);
+	}
+
+	public Object getAs(Class<?> clazz, int index) {
+		Preconditions.condition(index >= 0 && index < arguments.length,
+			String.format("Index must be between 0 and %d", arguments.length));
+		try {
+			return clazz.cast(arguments[index]);
```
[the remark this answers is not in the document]
geo-m: Initially did so but was aware that it could only convert from String. Just for clarity, are you referring to the getAs method or the entire class? Thank you

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/AggregateWith.java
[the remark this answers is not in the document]
marcphilipp: Yep, I was referring to the `range` attribute.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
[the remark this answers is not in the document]
marcphilipp: The entire class.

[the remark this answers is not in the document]
geo-m: So any conversion methods should delegate to DefaultArgumentConverter#convert, with the get{TYPE} methods being shortcuts for 

``
getString( i ) => DefaultArgumentConverter#convert(arguments[i] , String.class)
``

and for now we can assume that conversion from other types will be implemented in the future?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+
+public interface ArgumentsAccessor {
+
+	Object[] toArray();
+
+	List<Object> toList();
+
+	int getSize();
```
marcphilipp: Let's just call it `size()` like in `java.util.Collection`, shall we?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+	Object get(int index);
+
+	Object getAs(Class<?> clazz, int index);
```
marcphilipp: Just `get(Class, int)`? Otherwise, we'd have to rename the other ones, e.g. `getAsInt(int)`, too.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAggregator.java
```diff
@@ -0,0 +1,15 @@
...
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.params.aggregator;
+
+public interface ArgumentsAggregator {
+	Object aggregateArguments(ArgumentsAccessor accessor);
```
marcphilipp: Please add `ParameterContext` as the second parameter.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+	Character getChar(int index);
+
+	Boolean getBoolean(int index);
+
+	Byte getByte(int index);
+
+	Short getShort(int index);
+
+	Integer getInt(int index);
```
marcphilipp: Since we return the "wrapper" types, we should probably stick with the longer names, e.g. `getInteger(int)`.

geo-m:
Would like to thank @marcphilipp  for taking the time to review the seemingly silly errors. As my first PR and first time working with Java ever (Helloworld.java excluded), the awareness of such details is invaluable to me.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
```diff
@@ -0,0 +1,88 @@
...
+
+	public Double getDouble(int index) {
+		return (Double) get(Double.class, index);
+	}
+
+	public String getString(int index) {
+		return (String) get(String.class, index);
+	}
+
+	public Object get(Class<?> clazz, int index) {
```
marcphilipp: Please make this method generic:

```java
public <T> T get(Class<T> clazz, int index) {
    ...
    return clazz.cast(DefaultArgumentConverter.INSTANCE.convert(arguments[index], clazz));
}
```

This way you don't have cast in any of the more specific `get*`  methods and neither will code using the method.

geo-m: Thanks, looks great!

I am currently doubtful about how to proceed on the following 2 issues

1. Since the ArgumentAccessor is now an interface, instead of providing the DefaultArgumentAccessor i believe there is a need for a way to provide another implementation? In case of DI its straightforwar, just declare the implementation as type. 
In case of `@AggregateWith` , it must be passed somehow to the ArgumentsAggregator

2. Is there a need for an ArgumentAggregationException thrown by ArgumentAggregators and Accessors?

marcphilipp: > Since the ArgumentAccessor is now an interface, instead of providing the DefaultArgumentAccessor i believe there is a need for a way to provide another implementation?

I don't think that's necessary. The interface in this case is only intended to make it harder to depend on implementation details and give us more clarity and possibly flexibility to change things in the future. In the end, `DefaultArgumentsAccessor` should be annotated with `@API(Internal)`.

> Is there a need for an ArgumentAggregationException thrown by ArgumentAggregators and Accessors?

I was also wondering about that. Let's do it for the sake of consistency with `ArgumentConverter` but I think it should only be used by `ArgumentsAggregator` because an `ArgumentsAccessor` does not aggregate.

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.IntStream;
+
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class AggregatorTests {
+	static class Person {
```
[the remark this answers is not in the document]
geo-m: Not aware of this convention, thanks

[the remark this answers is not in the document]
sbrannen: Sure. No problem.

For inspiration, just look at other test classes within the code base. 😉

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+	int size();
+
+	Object get(int index);
+
+	<T> T get(Class<T> clazz, int index);
+
+	Character getChar(int index);
+
+	Boolean getBoolean(int index);
```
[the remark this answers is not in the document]
geo-m: No reason at all, just assumed this was your decision for visual consistency. Correcting now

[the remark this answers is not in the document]
sbrannen: It's not about visual consistency. Rather, it's about runtime semantics.

If it would ever make sense for the returned value to be `null`, we would then have to use wrapper types.

Otherwise, if the returned value can never be `null`, we should then use primitive types where possible.

In other words, I was basically asking if you had come up with a scenario where returning `null` would make sense.

Make sense? 😉

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
```diff
@@ -0,0 +1,77 @@
...
+	@CsvSource({ "John, Doe", "Jack, Smith" })
+	public void personAggregationTest(@AggregateWith(PersonAggregator.class) Person person) {
+		String fullName = person.name + " " + person.surname;
+		assertTrue(fullName.equals("John Doe") || fullName.equals("Jack Smith"));
+	}
+
+	@ParameterizedTest
+	@CsvSource({ "Foo,Bar,Baz", "Baz,Bar,Foo" })
+	public void hashMapAggregationTest(@AggregateWith(TestHashMapAggregator.class) Map map) {
+		assertEquals(map.size(), 3);
```
jbduncan: I believe that for `assertEquals` calls, it's idiomatic to make the _expected_ value `3` first, and the _actual_ calculated value `map.size()` second.

So something like:
```java
assertEquals(3, map.size());
```

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
```diff
@@ -0,0 +1,77 @@
...
+			() -> assertTrue(map.containsValue("Bar")), () -> assertTrue(map.containsValue("Foo")));
+
+	}
+
+	@ParameterizedTest
+	@CsvSource({ "1,2,3,4,5,6,7,8,9,10" })
+	public void directConversionToAccessorTest(ArgumentsAccessor accessor) {
+		assertTrue(accessor.size() == 10);
+		int sum = IntStream.range(0, 10).map(i -> accessor.getInteger(i)).sum();
+		assertEquals(sum, 55);
```
jbduncan: Ditto regarding my "assertEquals" comment above: I believe that `55` should be first, and `sum` should the second, i.e.:

```java
assertEquals(55, sum);
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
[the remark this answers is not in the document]
geo-m: Well the converter does "convert" null to any reference type ,as in, does not throw an exception. So if there is any possibility of a provided argument being null, aggregators could batch convert arguments without crashing. Waiting for your instructions on this

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
```diff
@@ -0,0 +1,77 @@
...
+			return new Person(accessor.getString(0), accessor.getString(1));
+		}
+	}
+
+	static class TestHashMapAggregator implements ArgumentsAggregator {
+		@Override
+		public Object aggregateArguments(ArgumentsAccessor accessor, ParameterContext context) {
+			Map<String, Object> m = new HashMap<>();
+			IntStream.range(0, accessor.size()).forEach(i -> m.put(Integer.toString(i), accessor.get(i)));
+			return m;
```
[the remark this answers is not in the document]
geo-m: Any feedback is greatly appreciated at that point! I think i have reached the point of having the most corrected PR in the entire repo so try to take my last sentence not too seriously haha

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.jupiter.params.converter.DefaultArgumentConverter;
+import org.junit.platform.commons.util.Preconditions;
+
+public class DefaultArgumentsAccessor implements ArgumentsAccessor {
+
+	private final Object[] arguments;
+
+	public DefaultArgumentsAccessor(Object[] arguments) {
+		Preconditions.notNull(arguments, "DefaultArgumentsAccessor initialized with null arguments");
+		this.arguments = arguments;
```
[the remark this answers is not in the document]
geo-m: Was thinking the same. Pretty sure the team will agree on this, as these arguments might eventually be passed to some other class in the future, and bugs will follow.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
```diff
@@ -0,0 +1,88 @@
...
+
+package org.junit.jupiter.params.aggregator;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.jupiter.params.converter.DefaultArgumentConverter;
+import org.junit.platform.commons.util.Preconditions;
+
+public class DefaultArgumentsAccessor implements ArgumentsAccessor {
```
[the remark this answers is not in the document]
geo-m: For the current implementation it is needed for instantation in ParameterizedTestParameterResolver, where i figured it was the most appropriate place to place it at the time.

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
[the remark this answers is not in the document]
jbduncan: @geo-m I'll do my best. 😉

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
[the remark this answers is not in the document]
sbrannen: OK

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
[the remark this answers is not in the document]
sbrannen: Well, it's possible for a user to specify a `null` value in the source for a given index and at other times provide a non-null value for the same index.

So I guess your original proposal is correct to allow `null` converted values.

In other words, I'm thinking we should stick with wrappers instead of primitive types for the greatest flexibility in the API. Consumers of the API can then decide on their own what to do about required vs. optional values.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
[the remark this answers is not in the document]
geo-m: Great, so i believe there are no pending changes, correct? When the team has decided on the final form of the API or has more improvements to suggest please inform me, thank you.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
```diff
@@ -86,8 +85,9 @@ private ParameterizedTestNameFormatter createNameFormatter(Method templateMethod
 	}
 
 	private Object[] consumedArguments(Object[] arguments, Method templateMethod) {
-		int parametersCount = templateMethod.getParameterCount();
-		return arguments.length > parametersCount ? Arrays.copyOf(arguments, parametersCount) : arguments;
+		/*int parametersCount = templateMethod.getParameterCount();
+		return arguments.length > parametersCount ? Arrays.copyOf(arguments, parametersCount) : arguments;*/
```
marcphilipp: This should still be in place unless one of the parameters is an `ArgumentsAccessor` or annotated with `@AggregateWith`, i.e. unless `isAggregate(parameter)` is true for any method parameter.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
```diff
@@ -41,20 +44,29 @@
...
+
+	}
+
+	private Object convert(ParameterContext parameterContext, ExtensionContext extensionContext) {
 		Parameter parameter = parameterContext.getParameter();
+		Object argument = arguments[parameterContext.getIndex()];
 		Optional<ConvertWith> annotation = AnnotationUtils.findAnnotation(parameter, ConvertWith.class);
-		// @formatter:off
+		// @formatter:off;
+
```
marcphilipp: Please revert this formatting change.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
```diff
@@ -41,20 +44,29 @@
...
 		ArgumentConverter argumentConverter = annotation.map(ConvertWith::value)
 				.map(clazz -> (ArgumentConverter) ReflectionUtils.newInstance(clazz))
 				.map(converter -> AnnotationConsumerInitializer.initialize(parameter, converter))
-				.orElse(DefaultArgumentConverter.INSTANCE);
+				.orElse	(DefaultArgumentConverter.INSTANCE
+				);
```
marcphilipp: Please revert this formatting change.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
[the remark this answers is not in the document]
marcphilipp: I don't think we need a copy here, but in `toArray()`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
geo-m: Should i make isAggregate static, or move it to an Util class then (or something else)? Instantiating a specific extension at that point seemed wrong semantically which is why i temporarily commented it out until i find a better way. Unless you disagree.

marcphilipp: You can make it `static` for now. I plan to go over the code in an IDE before we merge it and will potentially refactor a few things.

sormuras:
[12 commits](https://github.com/junit-team/junit5/pull/1325/commits) and not a single green check mark. Please run `gradlew spotlessApply build` locally and fix the issues reported before committing. If you need help to figure out what the issues are, please let us know.

geo-m:
That fix is already on its way. Unfortunately for some reason i dont get any exceptions locally but im currently looking into it.

As for the green marks i was under the impression that the Junit team will judge whether conventions have been followed, or the tests are adequate etc. I apologize and will try to get it right at once on my next PR.

sormuras:
The "green marks" are a good first indicator that our "automated source style rules" are met and nothing old is broken by a PR.

No need to apologize! Your fix and my comment met half way on their way to github. IIRC, there're some failing unit tests in your PR. Didn't investigate deeper what causes them to fail, yet.

Keep up the good work, @geo-m!

geo-m:
Thank you! We are aware of the issue from the first commit, and it is noted in my description, which why all builds "failed" so far (excluding the unused import), until @marcphilipp came up with a compromise at his last review.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
```diff
@@ -86,8 +86,12 @@ private ParameterizedTestNameFormatter createNameFormatter(Method templateMethod
 	}
 
 	private Object[] consumedArguments(Object[] arguments, Method templateMethod) {
+		boolean hasAggregate = Arrays.stream(templateMethod.getParameters()).filter(
+			p -> ParameterizedTestParameterResolver.isAggregate(p)).count() > 0;
```
marcphilipp: Please use:

```java
// @formatter:off
boolean hasAggregate = Arrays.stream(templateMethod.getParameters())
        .anyMatch(ParameterizedTestParameterResolver::isAggregate);
// @formatter:on
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
```diff
@@ -41,20 +44,28 @@
 	public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
 		Executable declaringExecutable = parameterContext.getParameter().getDeclaringExecutable();
 		Method testMethod = extensionContext.getTestMethod().orElse(null);
-		return declaringExecutable.equals(testMethod) && parameterContext.getIndex() < arguments.length;
+		return declaringExecutable.equals(testMethod)
+				&& ((parameterContext.getIndex() < arguments.length || isAggregate(parameterContext.getParameter())));
```
marcphilipp: Please remove the extra pair of parentheses.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
```diff
@@ -41,20 +44,28 @@
...
+				: convert(parameterContext, extensionContext);
+
+	}
+
+	private Object convert(ParameterContext parameterContext, ExtensionContext extensionContext) {
 		Parameter parameter = parameterContext.getParameter();
+		Object argument = arguments[parameterContext.getIndex()];
 		Optional<ConvertWith> annotation = AnnotationUtils.findAnnotation(parameter, ConvertWith.class);
 		// @formatter:off
+
```
marcphilipp: Please remove this blank line.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
```diff
@@ -41,20 +44,28 @@
...
 		ArgumentConverter argumentConverter = annotation.map(ConvertWith::value)
 				.map(clazz -> (ArgumentConverter) ReflectionUtils.newInstance(clazz))
 				.map(converter -> AnnotationConsumerInitializer.initialize(parameter, converter))
-				.orElse(DefaultArgumentConverter.INSTANCE);
+				.orElse	(DefaultArgumentConverter.INSTANCE);
```
marcphilipp: Please remove the space in front of the parenthesis.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
```diff
@@ -64,4 +75,23 @@ public Object resolveParameter(ParameterContext parameterContext, ExtensionConte
 				ex);
 		}
 	}
+
+	// An aggregate parameter resolution requires passing all method arguments
```
marcphilipp: Does this comment belong here? Do we need it at all?

geo-m: Seemed a bit ambiguous to me for a potentially unknowing contributor reading the code. But i guess javadoc will do a much better job at it. Speaking of which, should i be adding documentation, or wait for the junit5 team to finalize the api?

geo-m:
@marcphilipp Should i add documentation or wait for junit team to finalize the api first?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
```diff
@@ -0,0 +1,88 @@
...
+
+	private final Object[] arguments;
+
+	public DefaultArgumentsAccessor(Object[] arguments) {
+		Preconditions.notNull(arguments, "DefaultArgumentsAccessor initialized with null arguments");
```
sbrannen: Please change message to "arguments array must not be null" for consistency.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
```diff
@@ -0,0 +1,88 @@
...
+	public Double getDouble(int index) {
+		return get(Double.class, index);
+	}
+
+	public String getString(int index) {
+		return get(String.class, index);
+	}
+
+	public <T> T get(Class<T> clazz, int index) {
+		Preconditions.condition(index >= 0 && index < arguments.length,
```
sbrannen: No need to copy-n-paste the precondition here. Simply invoke `get(index)` instead of `arguments[index]`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
```diff
@@ -64,4 +74,22 @@ public Object resolveParameter(ParameterContext parameterContext, ExtensionConte
...
+
+	private Object aggregate(ParameterContext parameterContext, ExtensionContext extensionContext) {
+
+		Parameter parameter = parameterContext.getParameter();
+		Optional<AggregateWith> annotation = AnnotationUtils.findAnnotation(parameter, AggregateWith.class);
+		ArgumentsAccessor accessor = new DefaultArgumentsAccessor(arguments);
+		// @formatter:off
+		return annotation.map(AggregateWith::value)
+				.map(clazz -> ReflectionUtils.newInstance(clazz))
+				.map(aggregator -> aggregator.aggregateArguments(accessor,parameterContext))
```
sbrannen: This needs exception handling analogous that found in the `convert()` method.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
```diff
@@ -0,0 +1,88 @@
...
+			String.format("Index must be between 0 and %d", arguments.length));
+		return clazz.cast(DefaultArgumentConverter.INSTANCE.convert(arguments[index], clazz));
```
sbrannen: What happens if the cast fails?

It looks like this needs proper exception handling since JUnit code is not allowed to throw a `ClassCastException` directly.

For inspiration, have a look at `org.junit.jupiter.engine.execution.ExtensionValuesStore.castToRequiredType(Object, Object, Class<T>)`

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
```diff
@@ -0,0 +1,93 @@
...
+		assertEquals(3, map.size());
+
+		// @formatter:off
+		assertAll("Map contents",
+				() -> assertTrue(map.containsValue("Baz")),
+				() -> assertTrue(map.containsValue("Bar")),
+				() -> assertTrue(map.containsValue("Foo"))
+		);
+		// @formatter:on
+
```
sbrannen: delete blank line

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	@ParameterizedTest
+	@CsvSource({ "1,2,3,4,5,6,7,8,9,10" })
+	public void directConversionToAccessorTest(ArgumentsAccessor accessor) {
+		assertTrue(accessor.size() == 10);
```
sbrannen: We never use `assertTrue` for equality checks.

Instead, please use `assertEquals`.

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
```diff
@@ -0,0 +1,93 @@
...
+			this.name = name;
+			this.surname = surname;
+		}
+	}
+
+	static class PersonAggregator implements ArgumentsAggregator {
+
+		@Override
+		public Object aggregateArguments(ArgumentsAccessor accessor, ParameterContext context) {
+			assertTrue(accessor.size() >= 2);
```
sbrannen: ditto

junit-jupiter-params/src/test/java/org/junit/jupiter/params/aggregator/AggregatorTests.java
```diff
@@ -0,0 +1,93 @@
...
+			return IntStream
+					.range(0, accessor.size())
+					.boxed()
+					.collect(
+							toMap(
+									i -> Integer.toString(i),
+									i -> accessor.get(i)
+							)
+					);
+			// @formatter:on`
```
sbrannen: please delete the trailing backtick

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAccessor.java
```diff
@@ -0,0 +1,44 @@
...
+
+	Object[] toArray();
+
+	List<Object> toList();
+
+	int size();
+
+	Object get(int index);
+
+	<T> T get(Class<T> clazz, int index);
```
sbrannen: Please swap the order of the formal parameters: the required type is an optional argument and should therefore come after `index`.

sbrannen: Please rename `clazz` to `requireType`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
geo-m: If i understand correctly, ParameterResolution semantically encompasses any operation that results to the final value? Which means that i could move the Exception to the "resolveParameter" method to account for anything that goes wrong in the process?

sbrannen: Well, it depends on how much _contextual information_ needs to be provided to the exception when you create it.

If it's generic, it could be at a higher level; however, if meaningful information is only available at a lower level, the exception needs to be created there with the contextual information.

In other words, when we throw an exception we do our best to provide meaningful information to the user to help with diagnosing the problem.

geo-m:
All great general (and JUnit specific) practices to follow, thanks for your time.

I decided to make 2 new Exceptions, ArgumentsAccessorException and ArgumentsAggregationException, which i found help specify the error cause. I have not thrown any ArgumentsAggregationException, but it could be useful for Aggregator implementations to throw (since you mentioned general ones are not allowed)

sbrannen:
@marcphilipp, can you please either approve your requested changes or dismiss your review?

TIA

sbrannen:
> All great general (and JUnit specific) practices to follow, thanks for your time.

No problem.

> I decided to make 2 new Exceptions, ArgumentsAccessorException and ArgumentsAggregationException, which i found help specify the error cause. I have not thrown any ArgumentsAggregationException, but it could be useful for Aggregator implementations to throw (since you mentioned general ones are not allowed). I can remove if you feel they are redundant, and should throw conversion / resolution exceptions instead.

We'll analyze that in greater detail when we merge into master.

sbrannen:
@geo-m, can you please add the missing documentation (JavaDoc, User Guide, Release Notes)?

Thanks in advance!

geo-m:
Tried to follow the existing conventions. Targeting at least 110 comments after the upcoming reviews!

sbrannen:
> Tried to follow the existing conventions.

Thanks for drafting up the documentation. 👍 

> Targeting at least 110 comments after the upcoming reviews!

Nah. I think we'll forgo any further rounds of reviews and just polish after we merge your work. 😉 

Cheers, and enjoy your weekend!

sbrannen:
**Update**: I am merging this PR locally.

sbrannen:
The corresponding work on this issue has been pushed to `master` in commits e2295abd0069604852567a042acd5bc6fd3d576f and 3ec77993d46efe04cc5a3b62086bb3d5fd924e3d.

Thanks again for all of your hard work on these features, @geo-m!

Issue #1216, opened 2017-12-22
Revise layout and building of multi-release JARs


sormuras:
## Overview

Current "solution" uses external `jar` tool:
https://github.com/junit-team/junit5/blob/master/build.gradle#L489-L519

Better Gradle intergration presented by @melix at:
https://blog.gradle.org/mrjars

## Deliverables

- [ ] Revise layout and building of multi-release-JAR(s). Keep in mind, that the `jar` tool offers some API/ABI sanity checks when it packages the version-aware types.

sormuras:
Evaluate https://github.com/zyxist/chainsaw

PR #1326, opened 2018-03-14
Build should respect JAVA_HOME when running executables

tsachev:
Currently when java is not on the path or the java/jar executable is on path, but the version is not 9+, junit5 cannot be build.

It would be nice if project can be build using command something like 

```
JAVA_HOME=/usr/lib/jvm/java-9-jdk ./gradlew  clean build
```

The proposed change uses gradle's internal api which is probably not ok, but couldn't find better solution.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,3 @@
-import org.gradle.internal.jvm.Jvm
-
 import java.time.OffsetDateTime
 import java.time.format.DateTimeFormatter
 
@@ -562,11 +560,11 @@ subprojects { subproj ->
 			dependsOn = [versionedProject.classes]
 			doLast {
 				project.exec {
-					executable "${Jvm.current().javaHome}/bin/jar"
+					executable "${System.properties['java.home']}/bin/jar"
 					args = ['--version']
 				}
 				project.exec {
-					executable "${Jvm.current().javaHome}/bin/jar"
+					executable "${System.properties['java.home']}/bin/jar"
 					args = ['--update','--file', archivePath,
 							'--release', version,
 							'-C', versionedProject.compileJava.destinationDir,
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import org.gradle.internal.jvm.Jvm
+
 import java.time.OffsetDateTime
 import java.time.format.DateTimeFormatter
 
@@ -560,11 +562,11 @@ subprojects { subproj ->
 			dependsOn = [versionedProject.classes]
 			doLast {
 				project.exec {
-					executable 'jar'
+					executable "${Jvm.current().javaHome}/bin/jar"
 					args = ['--version']
 				}
 				project.exec {
-					executable 'jar'
+					executable "${Jvm.current().javaHome}/bin/jar"
 					args = ['--update','--file', archivePath,
 							'--release', version,
 							'-C', versionedProject.compileJava.destinationDir,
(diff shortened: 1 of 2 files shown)


sormuras:
```
ToolProvider jar = ToolProvider.findFirst('jar').get()
jar.run(System.out, System.err, '--version')
jar.run(System.out, System.err, '--update',
			'--file', archivePath.toString(),
			'--release', version,
			'-C', versionedProject.compileJava.destinationDir.toString(),
			'.')
```

fails with

```

* Where:
Build file 'R:\dev\github\junit-team\junit5\build.gradle' line: 563

* What went wrong:
Execution failed for task ':junit-platform-commons:jar'.
> Method java.util.spi.ToolProvider.findFirst(Ljava/lang/String;)Ljava/util/Optional; must be InterfaceMethodref constant

```

build.gradle
```diff
@@ -560,11 +562,11 @@ subprojects { subproj ->
 			dependsOn = [versionedProject.classes]
 			doLast {
 				project.exec {
-					executable 'jar'
+					executable "${Jvm.current().javaHome}/bin/jar"
```
sormuras: `executable "${System.properties['java.home']}/bin/jar"`

build.gradle
```diff
@@ -560,11 +562,11 @@ subprojects { subproj ->
...
 					args = ['--version']
 				}
 				project.exec {
-					executable 'jar'
+					executable "${Jvm.current().javaHome}/bin/jar"
```
sormuras: `executable "${System.properties['java.home']}/bin/jar"`

junit-platform-commons-java-9/junit-platform-commons-java-9.gradle
```diff
@@ -48,7 +50,7 @@ task generateDependenciesDirectory(type: Copy) {
 
 // Execute console launcher on the module-path.
 task execScanModulepath(type: Exec, dependsOn: [generateDependenciesDirectory, generateIntegrationTestsJar]) {
-	executable = 'java'
+	executable = "${Jvm.current().javaHome}/bin/java"
```
sormuras: `executable "${System.properties['java.home']}/bin/java"`

junit-platform-commons-java-9/junit-platform-commons-java-9.gradle
```diff
@@ -99,7 +101,7 @@ task execNoJavaScripting(type: Exec, dependsOn: [generateDependenciesDirectory,
 	ignoreExitValue = true
 	standardOutput = new ByteArrayOutputStream()
 	errorOutput = new ByteArrayOutputStream()
-	executable = 'java'
+	executable = "${Jvm.current().javaHome}/bin/java"
```
sormuras: `executable = "${System.properties['java.home']}/bin/java"`

marcphilipp:
@sormuras #1216 would be another option to get rid of the path dependency to `jar`, right?

sormuras:
Partly. We'd lose the built-in checks that the `jar` performs regarding same API (ABI?) of the multi-versioned types.

1. Replace plain `ja[va|r]` calls with `${System.properties['java.home']}/bin/ja[va|r]` now.
2. Use `ToolProvider.findFirst('jar')` when Groovy/Gradle supports it.
3. Wait for s/o writing a plugin (or native support) for MR-JAR support with similar checks that `jar` provides

Prefer 3 over 2, when it is released before 2.

sormuras:
4. Get `ExtensionValuesStore.java` race-condition simulation stable ... for stable code coverage reports. 😠

sormuras:
Thanks, @tsachev

PR #1347, opened 2018-03-28
Ensure failures in lifecycle methods take precedence over earlier TestAbortedExceptions

### LINKED ISSUE #1313: Failures in lifecycle methods should take precedence over TestAbortedException
## Overview

In the following example, `abortedTest` will be reported as _aborted_, while `successfulTest` will be reported as _failed_ due to the exception in `afterEach()`.

Expected behavior: `abortedTest` should be reported as _failed_ with a suppressed `TestAbortedException`.

```java
class TestCase {

	@Test
	void abortedTest() {
		assumeFalse(true, "abortedTest");
	}

	@Test
	void successfulTest() {
	}

	@AfterEach
	void afterEach() {
		throw new RuntimeException("afterEach");
	}
}
```

## Deliverables

- [ ] Let other exceptions take precedence over `TestAbortedException`.

marcphilipp:
Fixes #1313.

---

### Definition of Done

- [X] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/SingleTestExecutor.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/SingleTestExecutor.java
@@ -10,7 +10,7 @@
 
 package org.junit.platform.engine.support.hierarchical;
 
-import static org.apiguardian.api.API.Status.MAINTAINED;
+import static org.apiguardian.api.API.Status.DEPRECATED;
 import static org.junit.platform.commons.util.BlacklistedExceptions.rethrowIfBlacklisted;
 import static org.junit.platform.engine.TestExecutionResult.aborted;
 import static org.junit.platform.engine.TestExecutionResult.failed;
@@ -26,8 +26,11 @@ import org.opentest4j.TestAbortedException;
  *
  * @since 1.0
  * @see #executeSafely(Executable)
+ * @deprecated Please use {@link ThrowableCollector#execute} and
+ * {@link ThrowableCollector#toTestExecutionResult} instead.
  */
-@API(status = MAINTAINED, since = "1.0")
+@Deprecated
+@API(status = DEPRECATED, since = "1.2")
 public class SingleTestExecutor {
 
 	/**
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ThrowableCollector.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ThrowableCollector.java
@@ -16,22 +16,45 @@ import static org.junit.platform.engine.TestExecutionResult.aborted;
 import static org.junit.platform.engine.TestExecutionResult.failed;
 import static org.junit.platform.engine.TestExecutionResult.successful;
 
+import java.util.function.Predicate;
+
 import org.apiguardian.api.API;
 import org.junit.platform.commons.util.ExceptionUtils;
 import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.engine.TestExecutionResult;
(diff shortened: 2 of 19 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/SingleTestExecutor.java
```diff
@@ -1,78 +0,0 @@
...
-import org.opentest4j.TestAbortedException;
-
-/**
- * {@code SingleTestExecutor} encapsulates the execution of a single test
- * wrapped in an {@link Executable}.
- *
- * @since 1.0
- * @see #executeSafely(Executable)
- */
-@API(status = MAINTAINED, since = "1.0")
```
marcphilipp: Just realized this has status _maintained_... In that case, I'll resurrect and deprecate it.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ThrowableCollector.java
```diff
@@ -8,22 +8,27 @@
...
+import static org.junit.platform.engine.TestExecutionResult.aborted;
+import static org.junit.platform.engine.TestExecutionResult.failed;
+import static org.junit.platform.engine.TestExecutionResult.successful;
 
 import org.apiguardian.api.API;
-import org.junit.jupiter.api.function.Executable;
 import org.junit.platform.commons.util.ExceptionUtils;
 import org.junit.platform.commons.util.Preconditions;
+import org.junit.platform.engine.TestExecutionResult;
+import org.opentest4j.TestAbortedException;
```
sbrannen: This has the consequence of introducing a mandatory dependency on the OTA for any implementation of `HierarchicalTestExecutor` even if such a `TestEngine` does not use or even support exceptions from the OTA.

Thus, I'm not so sure we would want to do that.

Perhaps an alternative would be to allow a caller of the relocated `ThrowableCollector` to specify what the caller considers to be an "overridable exception".

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ThrowableCollector.java
```diff
@@ -57,7 +67,11 @@ private void add(Throwable t) {
 		if (this.throwable == null) {
 			this.throwable = t;
 		}
-		else {
+		else if (this.throwable instanceof TestAbortedException && !(t instanceof TestAbortedException)) {
+			t.addSuppressed(this.throwable);
+			this.throwable = t;
+		}
+		else if (throwable != t) {
```
sbrannen: What's the use case for this last `else if` block?

I'm not seeing it in my browser. 😉

marcphilipp:
@sbrannen I addressed your concerns in b68ee5a660788fc9525f6c558cd6966b9a5fffe5. PTAL! 🙂

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/SingleTestExecutor.java
```diff
@@ -26,8 +26,11 @@
  *
  * @since 1.0
  * @see #executeSafely(Executable)
+ * @deprecated Please use {@link ThrowableCollector#execute} and
+ * {@link ThrowableCollector#toTestExecutionResult} instead.
  */
-@API(status = MAINTAINED, since = "1.0")
+@Deprecated
+@API(status = DEPRECATED, since = "1.2")
```
marcphilipp: This should be 1.3.

sbrannen:
> I addressed your concerns in b68ee5a. PTAL! 🙂

Looks better now. Thanks.

PR #1365, opened 2018-04-12
Introduce ReflectiveFieldArguments for parameterized tests

nikowitt:
Hi there,

Maybe this pull request is considered useful as I'm probably going to use this quite often with Lombok's `@AllArgsConstructor` in our code base.

The basic idea is just to have a class with a field definition that automatically produces arguments that can be then used e.g via `@MethodSource`.

Any comments or feedback is greatly appreciated.

If you think if that is useful, I will clean up the code and complete the documentation.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/ReflectiveFieldArguments.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/ReflectiveFieldArguments.java
@@ -1,0 +1,21 @@
+package org.junit.jupiter.params.provider;
+
+import java.util.Arrays;
+
+/**
+ * helper class that can produce arguments based on the object field order. 
+ */
+public abstract class ReflectiveFieldArguments implements Arguments {
+
+	@Override
+	public Object[] get() {
+		return Arrays.stream(this.getClass().getDeclaredFields()).filter(f -> !f.isSynthetic()).map(f -> {
+			f.setAccessible(true);
+			try {
+				return f.get(this);
+			} catch (IllegalAccessException e) {
+				throw new RuntimeException("Error while getting value for field '" + f.getName() + "' in " + this, e);
+			}
+		}).toArray(Object[]::new);
+	}
+}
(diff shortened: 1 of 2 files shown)


sbrannen:
I'm having a bit of a hard time imagining how this is useful based on the PR.

Can you please provide a real-life test case that demonstrates its utility?

nikowitt:
Sure. In one of our scenarios, we test if grids work as expected, filtering & sorting every column of every grid of every entity for all of our products. These tests have to automatically scale with new grids. Of course there is an exception to the rule to some grid in some product, so we want to exclude particular automatically created tests based on condition. So this means there is one big method the effectively creates all possible test cases in a product (say, returning a Stream<Arguments>), but post processing just Arguments is not very nice. Instead, I'd like to return Steam<ReflectiveFieldArguments> that can be filtered based on fields, but does not need any boilerplate to implement get().

On a second thought, it probably might be better if getters instead of fields are used for the get() implementation to make filtering easier.

nikowitt:
Another use case I've just encountered:

```
public void suppress_exception_logs(String description, Supplier<ServletParameters> params, Throwable exception, boolean loggedAsError) {
```

This is my test signature and I need an arguments with no boilerplate class that gives me type safety so I can just supply the supplier that cannot be just put on an array without additional explicit casting.

marcphilipp:
@nikowitt Can you please provide an example on how you would use this? A method referenced by `@MethodSource` can already simply return a `Stream<YourObject>` instead of `Stream<Argument>` so you can have your custom domain object injected directly. Isn't that clearer than a detour via reflection?

nikowitt:
In my use case, we don't want to inject a custom domain object directly, we indeed have to rely on arguments to be injected into the test case as we rely on a test template with a provided signature. Otherwise, I'd totally agree.

marcphilipp:

nikowitt:
@marcphilipp, sorry for being off topic this time, but since yesterday evening, your name sounds even more familiar 🤪

marcphilipp:
@nikowitt I had a similar feeling when I opened this PR yesterday... 😉

Issue #542, opened 2016-10-11
[Extension model] - Test result extension.


SqAutoTestTeam:
Some of functionality written in Juni4 is based on usage of Rule. 
There is a custom rule is used to send tests results to outer system - where they can be seen later.

instance of subclass TestWatcher(Junit 4) provides me with three callback methods 
protected void succeeded(Description desc);
protected void failed(Throwable e, Description desc);
protected void skipped(AssumptionViolatedException e, Description desc);

So I can be aware what is the test result - failed, success or skipped -
(AssumptionViolatedException exception is thrown while test execution)

@mmerdes has made a good work of adapting existing junit4 rules to extensions of junit5
but it does support method result callbacks unfortunately.

I've researched this topic and came to realise that for now there is no strait reliable way to implement this feature using only Junit5 extension model.

The closest extension is AfterTestExecutionCallback but its callback method 
....
public void afterTestExecution(TestExtensionContext context) throws Exception
....
does not pass necessary info - test execution result. So none can tell for sure 
if test is ok, failed or ignored.
Indirect ways to solve this problem do not look reliable.

Suggestion: 
Possible solution is to add another Extension sub-interface 
(TestFinished for instance) that passes test result into its callback method. 

public interface TestFinished extends Extension {
    void testExecutionFinished(TestIdentifier testIdentifier, TestExtensionContext context, 
                           TestExecutionResult testExecutionResult) throws Exception;
}

Each extension class that supports  TestFinished interface gets notification of test execution result.
Thank you.

marcphilipp:
Does this need to be a Jupiter extension or would a Platform TestExecutionListener work for you (if there was a more dynamic way to register it as proposed in https://github.com/junit-team/junit5/pull/19#issuecomment-166610979)?

SqAutoTestTeam:
@marcphilipp

From my point of view :
Does this need to be a Jupiter extension ?
Yes this needs to be a Jupiter extension - because I need to know test result from extension.
For now there is no way to do so.

Or would a Platform TestExecutionListener work for you ?
No. Not at the moment
Typical use of TestExecutionListener is 
  
Launcher launcher = LauncherFactory.create();
launcher.registerTestExecutionListeners(new TestExecutionListener() {
...
})

launcher.discover( <---- to discover tests, get TestPlan ...
launcher.execute( <----  actual tests execution

launcher object is not accessible from extensions, so for now there is no way yo use it from within an extension object.

But if you mean something like this


/**
[...]

PR #1393, opened 2018-04-25
Introduce TestWatcher extension API for processing test execution results

geo-m:
## Overview

Draft implementation of a `TestWatcher`-like extension point for discussion, with a sample extension and a test case. Only supports `@Test` methods so far.

Issue: #542

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
@@ -10,11 +10,11 @@
 
 package org.junit.jupiter.api.extension;
 
-import org.apiguardian.api.API;
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
 import java.util.Optional;
 
-import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+import org.apiguardian.api.API;
 
 /**
  * The {@code TestWatcher} interface defines an extension API for processing test results.
@@ -38,8 +38,7 @@ public interface TestWatcher extends Extension {
 	 *
 	 * @param context the current extension context; never {@code null}
 	 */
-	default void testSuccessful(ExtensionContext context) {
-	}
+	void testSuccessful(ExtensionContext context);
 
 	/**
 	 * Invoked after a test was aborted.
@@ -47,8 +46,7 @@ public interface TestWatcher extends Extension {
 	 * @param context the current extension context; never {@code null}
 	 * @param cause the throwable responsible for the test being aborted; may be {@code null}
 	 */
-	default void testAborted(ExtensionContext context, Throwable cause) {
-	}
+	void testAborted(ExtensionContext context, Throwable cause);
 
 	/**
 	 * Invoked after a test has failed.
@@ -56,8 +54,7 @@ public interface TestWatcher extends Extension {
 	 * @param context the current extension context; never {@code null}
 	 * @param cause the throwable that caused test failure; may be {@code null}
 	 */
-	default void testFailed(ExtensionContext context, Throwable cause) {
-	}
+	void testFailed(ExtensionContext context, Throwable cause);
 
 	/**
 	 * Invoked after skipping a disabled test.
@@ -65,6 +62,5 @@ public interface TestWatcher extends Extension {
 	 * @param context the current extension context; never {@code null}
 	 * @param reason the reason for skipping the test; never {@code null}
 	 */
-	default void testDisabled(ExtensionContext context, Optional<String> reason) {
-	}
+	void testDisabled(ExtensionContext context, Optional<String> reason);
 }
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api.extension;
+
+import java.util.Optional;
+
+public interface TestWatcher extends Extension {
+
+	default void testSuccessful(String identifier, ExtensionContext context) {
+	}
+
+	default void testAborted(String identifier, Throwable cause, ExtensionContext context) {
+	}
+
+	default void testFailed(String identifier, Throwable cause, ExtensionContext context) {
+	}
+
+	default void testSkipped(String identifier, Optional<String> Reason, ExtensionContext context) {
+	}
+}
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
@@ -10,19 +10,61 @@
 
 package org.junit.jupiter.api.extension;
 
+import org.apiguardian.api.API;
+
 import java.util.Optional;
 
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+/**
+ * The {@code TestWatcher} interface defines an extension API for processing test results.
+ *
+ * <p> Interface methods are called after a test has been processed. Currently,
+ * only method based tests will be reported.
+ *
+ * <p>Extensions implementing this API can be registered at any level, reporting subsequent
+ * tests down the chain.
+ *
+ * <p>Any {@code ClosableResource} objects stored in the injected {@link ExtensionContext}
+ * have already been <strong>closed</strong> at invocation time.
+ *
+ * @since 5.4
+ */
+@API(status = EXPERIMENTAL, since = "5.4")
 public interface TestWatcher extends Extension {
 
+	/**
+	 * Invoked after a test has completed successfully.
+	 *
+	 * @param context the current extension context; never {@code null}
+	 */
 	default void testSuccessful(ExtensionContext context) {
 	}
 
+	/**
+	 * Invoked after a test was aborted.
+	 *
+	 * @param context the current extension context; never {@code null}
+	 * @param cause the throwable responsible for the test being aborted; may be {@code null}
+	 */
 	default void testAborted(ExtensionContext context, Throwable cause) {
 	}
 
+	/**
+	 * Invoked after a test has failed.
+	 *
+	 * @param context the current extension context; never {@code null}
+	 * @param cause the throwable that caused test failure; may be {@code null}
+	 */
 	default void testFailed(ExtensionContext context, Throwable cause) {
 	}
 
+	/**
+	 * Invoked after skipping a disabled test.
+	 *
+	 * @param context the current extension context; never {@code null}
+	 * @param reason the reason for skipping the test; never {@code null}
+	 */
 	default void testDisabled(ExtensionContext context, Optional<String> reason) {
 	}
 }
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
@@ -14,15 +14,15 @@ import java.util.Optional;
 
 public interface TestWatcher extends Extension {
 
-	default void testSuccessful(String identifier, ExtensionContext context) {
+	default void testSuccessful(ExtensionContext context) {
 	}
 
-	default void testAborted(String identifier, Throwable cause, ExtensionContext context) {
+	default void testAborted(ExtensionContext context, Throwable cause) {
 	}
 
-	default void testFailed(String identifier, Throwable cause, ExtensionContext context) {
+	default void testFailed(ExtensionContext context, Throwable cause) {
 	}
 
-	default void testSkipped(String identifier, Optional<String> Reason, ExtensionContext context) {
+	default void testDisabled(ExtensionContext context, Optional<String> reason) {
 	}
 }
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
@@ -35,6 +35,8 @@ import org.junit.jupiter.engine.execution.BeforeEachMethodAdapter;
 import org.junit.jupiter.engine.execution.ExecutableInvoker;
 import org.junit.jupiter.engine.execution.JupiterEngineExecutionContext;
 import org.junit.jupiter.engine.extension.ExtensionRegistry;
+import org.junit.platform.commons.logging.Logger;
+import org.junit.platform.commons.logging.LoggerFactory;
 import org.junit.platform.commons.util.ExceptionUtils;
 import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.TestExecutionResult;
@@ -64,6 +66,7 @@ import org.junit.platform.engine.support.hierarchical.ThrowableCollector.Executa
 public class TestMethodTestDescriptor extends MethodBasedTestDescriptor {
 
 	private static final ExecutableInvoker executableInvoker = new ExecutableInvoker();
+	private static final Logger logger = LoggerFactory.getLogger(TestMethodTestDescriptor.class);
 
 	public TestMethodTestDescriptor(UniqueId uniqueId, Class<?> testClass, Method testMethod) {
 		super(uniqueId, testClass, testMethod);
@@ -238,9 +241,9 @@ public class TestMethodTestDescriptor extends MethodBasedTestDescriptor {
 	@Override
 	public void nodeSkipped(JupiterEngineExecutionContext context, TestDescriptor descriptor, SkipResult result) {
 		if (context != null) {
-			invokeTestWatchers(context, t -> {
-				t.testSkipped(createTestIdentifier(), result.getReason(), context.getExtensionContext());
-			});
+			invokeTestWatchers(context, watcher -> {
+				watcher.testDisabled(context.getExtensionContext(), result.getReason());
+			}, false);
 		}
 	}
 
@@ -248,39 +251,39 @@ public class TestMethodTestDescriptor extends MethodBasedTestDescriptor {
 	public void nodeFinished(JupiterEngineExecutionContext context, TestDescriptor descriptor,
 			TestExecutionResult result) {
 		if (context != null) {
-			invokeTestWatchers(context, t -> {
+			invokeTestWatchers(context, watcher -> {
 				ExtensionContext extensionContext = context.getExtensionContext();
 				TestExecutionResult.Status status = result.getStatus();
-				String identifier = createTestIdentifier();
 				switch (status) {
 					case SUCCESSFUL:
-						t.testSuccessful(identifier, extensionContext);
+						watcher.testSuccessful(extensionContext);
 						break;
 					case FAILED:
-						t.testFailed(identifier, result.getThrowable().get(), extensionContext);
+						watcher.testFailed(extensionContext, result.getThrowable().get());
 						break;
 					case ABORTED:
-						t.testAborted(identifier, result.getThrowable().get(), extensionContext);
+						watcher.testAborted(extensionContext, result.getThrowable().get());
 						break;
 				}
-			});
+			}, true);
 		}
 	}
 
-	private void invokeTestWatchers(JupiterEngineExecutionContext context, Consumer<TestWatcher> callback) {
+	private void invokeTestWatchers(JupiterEngineExecutionContext context, Consumer<TestWatcher> callback,
+			boolean reverseOrder) {
 		ExtensionRegistry registry = context.getExtensionRegistry();
-		registry.getReversedExtensions(TestWatcher.class).forEach(t -> {
+		List<? extends TestWatcher> watchers = reverseOrder ? registry.getReversedExtensions(TestWatcher.class)
+				: registry.getExtensions(TestWatcher.class);
+		watchers.forEach(watcher -> {
 			try {
-				callback.accept(t);
+				callback.accept(watcher);
 			}
 			catch (Exception e) {
-				//swallow
+				logger.warn(() -> {
+					return String.format("Error invoking TestWatcher %s for test %s : %s", watcher.getClass().getName(),
+						e);
+				});
 			}
 		});
-	}
-
-	private String createTestIdentifier() {
-		Method testMethod = this.getTestMethod();
-		return String.format("%s#%s", testMethod.getDeclaringClass().getName(), testMethod.getName());
 	}
 }
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
@@ -237,30 +237,34 @@ public class TestMethodTestDescriptor extends MethodBasedTestDescriptor {
 
 	@Override
 	public void nodeSkipped(JupiterEngineExecutionContext context, TestDescriptor descriptor, SkipResult result) {
-		invokeTestWatchers(context, t -> {
-			t.testSkipped(createTestIdentifier(), result.getReason(), context.getExtensionContext());
-		});
+		if (context != null) {
+			invokeTestWatchers(context, t -> {
+				t.testSkipped(createTestIdentifier(), result.getReason(), context.getExtensionContext());
+			});
+		}
 	}
 
 	@Override
 	public void nodeFinished(JupiterEngineExecutionContext context, TestDescriptor descriptor,
 			TestExecutionResult result) {
-		invokeTestWatchers(context, t -> {
-			ExtensionContext extensionContext = context.getExtensionContext();
-			Status status = result.getStatus();
-			String identifier = createTestIdentifier();
-			switch (status) {
-				case SUCCESSFUL:
-					t.testSuccessful(identifier, extensionContext);
-					break;
-				case FAILED:
-					t.testFailed(identifier, result.getThrowable().get(), extensionContext);
-					break;
-				case ABORTED:
-					t.testAborted(identifier, result.getThrowable().get(), extensionContext);
-					break;
-			}
-		});
+		if (context != null) {
+			invokeTestWatchers(context, t -> {
+				ExtensionContext extensionContext = context.getExtensionContext();
+				TestExecutionResult.Status status = result.getStatus();
+				String identifier = createTestIdentifier();
+				switch (status) {
+					case SUCCESSFUL:
+						t.testSuccessful(identifier, extensionContext);
+						break;
+					case FAILED:
+						t.testFailed(identifier, result.getThrowable().get(), extensionContext);
+						break;
+					case ABORTED:
+						t.testAborted(identifier, result.getThrowable().get(), extensionContext);
+						break;
+				}
+			});
+		}
 	}
 
 	private void invokeTestWatchers(JupiterEngineExecutionContext context, Consumer<TestWatcher> callback) {
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
@@ -13,7 +13,6 @@ package org.junit.jupiter.engine.descriptor;
 import static org.apiguardian.api.API.Status.INTERNAL;
 import static org.junit.jupiter.engine.descriptor.ExtensionUtils.populateNewExtensionRegistryFromExtendWithAnnotation;
 import static org.junit.jupiter.engine.support.JupiterThrowableCollectorFactory.createThrowableCollector;
-import static org.junit.platform.engine.TestExecutionResult.Status;
 
 import java.lang.reflect.Method;
 import java.util.List;
@@ -37,6 +36,7 @@ import org.junit.jupiter.engine.execution.JupiterEngineExecutionContext;
 import org.junit.jupiter.engine.extension.ExtensionRegistry;
 import org.junit.platform.commons.logging.Logger;
 import org.junit.platform.commons.logging.LoggerFactory;
+import org.junit.platform.commons.util.BlacklistedExceptions;
 import org.junit.platform.commons.util.ExceptionUtils;
 import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.TestExecutionResult;
@@ -278,11 +278,10 @@ public class TestMethodTestDescriptor extends MethodBasedTestDescriptor {
 			try {
 				callback.accept(watcher);
 			}
-			catch (Exception e) {
-				logger.warn(() -> {
-					return String.format("Error invoking TestWatcher %s for test %s : %s", watcher.getClass().getName(),
-						e);
-				});
+			catch (Throwable throwable) {
+				BlacklistedExceptions.rethrowIfBlacklisted(throwable);
+				logger.warn(throwable, () -> String.format("Failed to invoke TestWatcher %s for test %s",
+					watcher.getClass().getName(), context.getExtensionContext().getUniqueId()));
 			}
 		});
 	}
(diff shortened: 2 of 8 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
```diff
@@ -0,0 +1,28 @@
...
+	}
+
+	default void testAborted(String identifier, Throwable cause, ExtensionContext context) {
+	}
+
+	default void testFailed(String identifier, Throwable cause, ExtensionContext context) {
+	}
+
+	default void testSkipped(String identifier, Optional<String> Reason, ExtensionContext context) {
+	}
```
marcphilipp: I think we should remove the `String identifier` and put `ExtensionContext` first in the argument list. Moreover, `Reason` should be lower case.

marcphilipp:
@geo-m Do you have time to make the suggested changes?

geo-m:
@marcphillip it might take a while due to limited time. If you're on a tight schedule feel free to take over the PR

marcphilipp:
There's no rush, I probably won't get to in the next few days either.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
sbrannen: I agree.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
```diff
@@ -230,4 +234,48 @@ private void invokeAfterEachCallbacks(JupiterEngineExecutionContext context) {
 		});
 	}
 
+	@Override
+	public void nodeSkipped(JupiterEngineExecutionContext context, TestDescriptor descriptor, SkipResult result) {
+		invokeTestWatchers(context, t -> {
```
sbrannen: Please rename `t` to `watcher`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
```diff
@@ -230,4 +234,48 @@ private void invokeAfterEachCallbacks(JupiterEngineExecutionContext context) {
...
+			t.testSkipped(createTestIdentifier(), result.getReason(), context.getExtensionContext());
+		});
+	}
+
+	@Override
+	public void nodeFinished(JupiterEngineExecutionContext context, TestDescriptor descriptor,
+			TestExecutionResult result) {
+		invokeTestWatchers(context, t -> {
```
sbrannen: Please rename `t` to `watcher`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
```diff
@@ -230,4 +234,48 @@ private void invokeAfterEachCallbacks(JupiterEngineExecutionContext context) {
...
+			}
+			catch (Exception e) {
+				//swallow
+			}
+		});
+	}
+
+	private String createTestIdentifier() {
+		Method testMethod = this.getTestMethod();
+		return String.format("%s#%s", testMethod.getDeclaringClass().getName(), testMethod.getName());
```
sbrannen: This is inappropriate for a few reasons.

1. The declaring class may not be the current test class, which would lead to conflicts for IDs for the same method executed multiple times in a class hierarchy.
2. The method parameters are missing, which would lead to conflicts for overloaded test methods.

If we were to do this, we'd use `ReflectionUtils.getFullyQualifiedMethodName(Class<?>, String, Class<?>...)`, but as @marcphilipp already suggested, we should actually remove this "ID" since the `ExtensionContext` provides sufficient contextual information.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
```diff
@@ -0,0 +1,131 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestWatcher;
+import org.junit.jupiter.engine.AbstractJupiterTestEngineTests;
+import org.junit.platform.engine.test.event.ExecutionEventRecorder;
+
+@Disabled
+public class TestWatcherTests extends AbstractJupiterTestEngineTests {
```
sbrannen: Does not need to be `public` -- right?

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
```diff
@@ -0,0 +1,131 @@
...
+
+	@Test
+	void testWatcherValidityIncludingNestedTest() {
+		ExecutionEventRecorder recorder = executeTestsForClass(nestedTestWatcherTestCase.class);
+		assertEquals(0, recorder.getContainerFailedCount());
+	}
+
+	@ExtendWith(TestResultAggregator.class)
+	static class nestedTestWatcherTestCase {
```
sbrannen: All types begin with a capital letter: n --> N

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
```diff
@@ -0,0 +1,131 @@
...
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestWatcher;
+import org.junit.jupiter.engine.AbstractJupiterTestEngineTests;
+import org.junit.platform.engine.test.event.ExecutionEventRecorder;
+
+@Disabled
```
sbrannen: Please document why it's disabled via the implicit `value` attribute.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		@Test
+		@Disabled
+		public void skippedTest() {
+			//no-op
+		}
+
+		@Nested
+		class secondLevelTestWatcherTestCase {
```
sbrannen: s --> S

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
```diff
@@ -0,0 +1,131 @@
...
+		public void testSkipped(String descriptor, Optional<String> Reason, ExtensionContext context) {
+			storeResult("SKIPPED", descriptor);
+		}
+
+		private void storeResult(String status, String method) {
+			List<String> l = results.get(status);
+			if (l == null) {
+				l = new ArrayList<>();
+			}
+			l.add(method);
```
sbrannen: You can achieve this in one line of code using `computeIfAbsent(...)`.

From the JavaDoc for `java.util.Map#computeIfAbsent`:

> Or to implement a multi-value map, Map<K,Collection<V>>, supporting multiple values per key:
> 
> `map.computeIfAbsent(key, k -> new HashSet<V>()).add(v);`

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
```diff
@@ -0,0 +1,131 @@
...
+
+	}
+
+	public static class TestResultAggregator implements TestWatcher, AfterAllCallback {
+
+		private Map<String, List<String>> results = new HashMap<>();
+
+		@Override
+		public void testSuccessful(String descriptor, ExtensionContext context) {
+			storeResult("SUCCESSFUL", descriptor);
```
sbrannen: Once you remove the `descriptor` argument from the event methods, you'll need to switch to `context.getUniqueId()`.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
```diff
@@ -0,0 +1,131 @@
...
+			if (l == null) {
+				l = new ArrayList<>();
+			}
+			l.add(method);
+			results.put(status, l);
+		}
+
+		@Override
+		public void afterAll(ExtensionContext context) throws Exception {
+			this.results.entrySet().stream().forEach(resultType -> assertEquals(2, resultType.getValue().size()));
```
sbrannen: You should be able to replace that with `this.results.values().forEach(idList -> assertEquals(2, idList.size()))` -- right?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/HierarchicalTestExecutor.java
```diff
@@ -158,16 +158,21 @@ private void cleanUp() {
 		private void reportDone() {
 			if (executionResult != null) {
 				addExecutionErrorsToTestExecutionResult();
+				node.nodeFinished(context, testDescriptor, executionResult);
```
sbrannen: Must be wrapped in a try-catch block in order to ensure that a `TestWatcher` that throws an exception cannot abort the execution and reporting.

Any exception should be logged (e.g., at `WARNING` log level) and swallowed.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/HierarchicalTestExecutor.java
```diff
@@ -158,16 +158,21 @@ private void cleanUp() {
...
 				listener.executionFinished(testDescriptor, executionResult);
 			}
 			else if (executionErrors.isEmpty() && skipResult.isSkipped()) {
+				node.nodeSkipped(context, testDescriptor, skipResult);
```
sbrannen: Add exception handling as described above.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/HierarchicalTestExecutor.java
```diff
@@ -158,16 +158,21 @@ private void cleanUp() {
...
 				listener.executionSkipped(testDescriptor, skipResult.getReason().orElse("<unknown>"));
 			}
 			else {
 				// Call executionStarted first to comply with the contract of EngineExecutionListener.
 				listener.executionStarted(testDescriptor);
-				listener.executionFinished(testDescriptor, createTestExecutionResultFromExecutionErrors());
+				TestExecutionResult result = createTestExecutionResultFromExecutionErrors();
+				node.nodeFinished(context, testDescriptor, result);
```
sbrannen: Add exception handling as described above.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
sbrannen: I think test watchers should be invoked in "registration order" for skipped events since a node is skipped _before_ it would have been executed (analogous to before-each lifecycle callbacks); however, your current implementation of `invokeTestWatchers()` retrieves the watchers via `registry.getReversedExtensions()` which will always invoke the watchers in "reverse registration order".

sbrannen:
@geo-m, to address your feedback request:

> - Whether this is the correct info to be finally passed to extension methods. E.g an alternative of the method identifier string is using a `TestIdentifier` object but that will require a new dependency.

@marcphilipp and I have voiced our desire to omit an additional, custom identifier from this extension API. See review comments for details.

> - Whether the node notification should be moved before `cleanup()` in order to avoid operating on a closed `ExtensionContext`. I am not aware of the full consequences of doing this (tests show no change).

That's a good question.

The way you've done it now it by far the easiest way to do it, since you do it alongside events for Platform `TestExecutionListener`s.

However, if a `TestWatcher` event/callback method is invoked after `cleanUp()` is invoked, that means that any `CloseableResource` that is stored in the `ExtensionContext.Store` will already have been closed. 

So... if a `TestWatcher` needs to interact with a `CloseableResource`, that interaction could potentially fail.

Note, however, that the contents of the `Store` are not actually cleared in `cleanUp()`, so interaction with objects in the `Store` that do not implement `CloseableResource` should be fine.

sbrannen:
To follow up on my last comment, we basically need to decide if it's acceptable that a `TestWatcher` cannot reliably interact with a `CloseableResource` in the `Store` for aborted, failed, and successful events. _(note that skipped events are not affected.)_

If that's not acceptable, the solution would require a potentially extensive overhaul to the internals of (what is now in `master` called) `NodeTestTask`.

I say "potentially", because I have not investigated it fully.

marcphilipp:
I agree that `TestWatchers` should be called before `cleanUp()`. Thus, you might have to change the implementation of `NodeTestTask` to split `TestWatcher` from `ExecutionListener` notification but only compute the `TestExecutionResult` once.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
sbrannen: Please rework the implementation so that watchers are invoked in _registration order_ for skipped events and otherwise in _reverse registration order_.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/HierarchicalTestExecutor.java
```diff
@@ -158,16 +158,21 @@ private void cleanUp() {
 		private void reportDone() {
```
sbrannen: Please rework the implementation so that the new node event callbacks are invoked before `cleanUp()`.

For implementation hints, please see the comment from @marcphilipp: https://github.com/junit-team/junit5/pull/1393#issuecomment-403209962

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
```diff
@@ -0,0 +1,28 @@
...
+	default void testSuccessful(String identifier, ExtensionContext context) {
+	}
+
+	default void testAborted(String identifier, Throwable cause, ExtensionContext context) {
+	}
+
+	default void testFailed(String identifier, Throwable cause, ExtensionContext context) {
+	}
+
+	default void testSkipped(String identifier, Optional<String> Reason, ExtensionContext context) {
```
sbrannen: After putting a bit more thought into this, I am inclined to say that we should name this `testDisabled()` instead of `testSkipped()`.

Although "skipped" is technically correct with regard to the `Node` abstraction at the engine level, the actual terminology used within JUnit Jupiter is "disabled" due to the `org.junit.jupiter.api.extension.ConditionEvaluationResult.isDisabled()` API.

Thus, I think it is best to be consistent regarding the terminology used within the Jupiter extension model.

marcphilipp:
Please disregard [my previous comment](https://github.com/junit-team/junit5/pull/1393#issuecomment-403209962). I put some more thought into this and came to a different conclusion. Since an exception in `cleanUp()` would cause the test to fail, we cannot call the `TestWatcher` extensions before. If we did, we would potentially tell them a test passed when it might fail at a later point in time. Thus, I think we'll just have to document in `TestWatcher` that `CloseableResources` in the `ExtensionContext.Store` have already been closed. I think that's acceptable since the intent of `TestWatcher` is to do additional reporting, not implementing cleanup etc.

> One thing I find odd about the current proposal is that a node already has a `Node.shouldBeSkipped()` method.
> 
> So a node technically already knows if got _skipped_, and yet the current proposal introduces a `nodeSkipped()` method to inform the node of something that it already knows.

With the same rationale as above, the node does in fact not know whether it will be reported as skipped or failed so I think we need the additional `nodeSkipped()` callback.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/HierarchicalTestExecutor.java
marcphilipp: I [changed my mind](https://github.com/junit-team/junit5/pull/1393#issuecomment-403237102)... sorry! 😬

sbrannen:
Thanks for putting more thought into it, @marcphilipp.

I agree with your most recent analysis.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestWatcher.java
sbrannen: Based on the 👍 from @marcphilipp, please do rename that to `testDisabled()`.

marcphilipp:
@geo-m Do you have time to continue work on this one?

geo-m:
@marcphilipp In a couple of days i will be able to access a capable machine and start the review changes. For now i just brought the original changes up to date with the parallel processing structure. Congrats on that feature by the way!

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
```diff
@@ -231,4 +234,53 @@ private void invokeAfterEachCallbacks(JupiterEngineExecutionContext context) {
 		});
 	}
 
+	@Override
+	public void nodeSkipped(JupiterEngineExecutionContext context, TestDescriptor descriptor, SkipResult result) {
+		if (context != null) {
```
marcphilipp: When would `context` be `null` here and below?

geo-m: Context was null in the tests launched by the ExceptionHandlingTests and that caused them to fail with a NullPointerException on context. I have not looked into the exact cause yet(will do), but there is another check for a null context in NodeTestTask.

marcphilipp: I checked and `context` is indeed `null` when `prepare()` throws an exception which happens when an exception occurs creating the test instance.

If we want to report such failed tests to extensions as well and I think we should, we need to change `prepare` to add the exception to `ThrowableCollector` instead and add `context.getThrowableCollector().assertEmpty()` to `JupiterTestDescriptor.shouldBeSkipped()` (otherwise a test with a failing constructor could be reported as skipped).

One more thing I noticed: When something fails on the test class level, e.g. an exception a `@BeforeAll` method, `TestMethodTestDescriptor` never gets called. I think we should clearly document this limitation in the Javadoc of `TestWatcher`.

sbrannen: Sounds reasonable to me, @marcphilipp

sbrannen: And... thanks for investigating!

geo-m:
@sbrannen @marcphilipp 

Working on fixing the extension composability test that causes the build to fail under certain parameters, plus the null context issue that marc investigated.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
```diff
@@ -165,13 +165,25 @@ private void cleanUp() {
 
 	private void reportCompletion() {
 		if (throwableCollector.isEmpty() && skipResult.isSkipped()) {
+			try {
```
joerg1985: What is this try / finally block good for, it doesn't swallow the exception like the comment indicates.
Shouldn't this be a try / catch(Exception e) block to really swallow the exception?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
```diff
@@ -165,13 +165,25 @@ private void cleanUp() {
...
+				//swallow
+			}
 			listener.executionSkipped(testDescriptor, skipResult.getReason().orElse("<unknown>"));
 			return;
 		}
 		if (!started) {
 			// Call executionStarted first to comply with the contract of EngineExecutionListener.
 			listener.executionStarted(testDescriptor);
 		}
+		try {
```
joerg1985: What is this try / finally block good for, it doesn't swallow the exception like the comment indicates.
Shouldn't this be a try / catch(Exception e) block to really swallow the exception?

geo-m: Fixed it, typo from some resource handling habits. Thanks!

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
```diff
@@ -165,13 +165,25 @@ private void cleanUp() {
 
 	private void reportCompletion() {
 		if (throwableCollector.isEmpty() && skipResult.isSkipped()) {
+			try {
+				node.nodeSkipped(context, testDescriptor, skipResult);
+			}
+			catch(Throwable t) {
+				//swallow
```
sbrannen: If `nodeSkipped()` throws an exception, let's at least log it a `DEBUG` level.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
```diff
@@ -165,13 +165,25 @@ private void cleanUp() {
...
 		}
 		if (!started) {
 			// Call executionStarted first to comply with the contract of EngineExecutionListener.
 			listener.executionStarted(testDescriptor);
 		}
+		try {
+			node.nodeFinished(context, testDescriptor, throwableCollector.toTestExecutionResult());
+		}
+		catch(Throwable t) {
+			//swallow
```
sbrannen: If `nodeFinished()` throws an exception, let's at least log it a `DEBUG` level.

geo-m: Would you suggest that its an additional logging (as there is already in the method implementation), or should i remove that one? Since Node is a general abstraction but testwatchers are limited to jupiter extension api, i will need to know when the error comes from a watcher. Maybe a new exception type?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
```diff
@@ -231,4 +237,53 @@ private void invokeAfterEachCallbacks(JupiterEngineExecutionContext context) {
...
+		ExtensionRegistry registry = context.getExtensionRegistry();
+		List<? extends TestWatcher> watchers = reverseOrder ? registry.getReversedExtensions(TestWatcher.class)
+				: registry.getExtensions(TestWatcher.class);
+		watchers.forEach(watcher -> {
+			try {
+				callback.accept(watcher);
+			}
+			catch (Exception e) {
+				logger.warn(() -> {
+					return String.format("Error invoking TestWatcher %s for test %s : %s", watcher.getClass().getName(),
```
sbrannen: This is broken. There are 3 occurrences of `%s`. Please pass the exception as an argument to the `warn()` method, and include the "test" in the message.

sbrannen: Also, please introduce tests that verify the expected exception handling.

sbrannen: Furthermore, we should likely catch `Throwable`, rethrow it if it's _blacklisted_, and otherwise swallow it (but log it, as it is now).

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
sbrannen: Although it's overkill for the specific case of `TestWatcher` with the `JupiterTestEngine`, I think we should still log any exceptions thrown here for the sake of use with other third-party engine implementations.

@junit-team/junit-lambda, do you agree?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
sbrannen: Swallowing `Throwable` is not allowed in JUnit 5. We need to rethrow it if it's _blacklisted_.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
sbrannen: Swallowing `Throwable` is not allowed in JUnit 5. We need to rethrow it if it's _blacklisted_.

geo-m:
Not sure what causes the composability test to fail, why is a $jacocoInit() method expected from the extensions? Will look into it a bit more.

sbrannen:
> EDIT: Apparently its possible that jacoco automatically inserts some synthetic fields in classes, and actualMethods picks them up?

Could be. 

`master` seems to be OK. So, what happens if you rebase on the current `master`?

sbrannen:
FYI: I approved the first round of changes.

Once the PR is passing on the CI servers again, I think we'll be ready for another round of reviews.

Though, at this point, it looks like this will go into JUnit Jupiter 5.4.

geo-m:
Rebasing on master locally works fine, and the jacoco issue was solved with the last commit until a better solution is implemented. There is still a CI error but it looks like its irrelevant to the current changes.

sormuras:
Restarted the job...

sbrannen:
> Rebasing on master locally works fine, and the jacoco issue was solved with the last commit until a better solution is implemented. 

Good news! 👍

marcphilipp:
@geo-m Can you please rebase on master to resolve the merge conflict?

geo-m:
Rebased on master. Context being null before reportCompletion prevented what has been discussed in this PR so i had to move it. There are no apparent effects on either test success or performance.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
```diff
@@ -120,6 +121,12 @@ default C execute(C context, DynamicTestExecutor dynamicTestExecutor) throws Exc
 	default void after(C context) throws Exception {
 	}
 
+	default void nodeSkipped(C context, TestDescriptor descriptor, SkipResult result) {
```
sbrannen: The new `nodeSkipped()` and `nodeFinished()` methods both need `@since` and `@API` declarations as well as Javadoc.

sbrannen: @junit-team/junit-lambda, we need to decide whether these new methods in `Node` should be _EXPERIMENTAL_, _STABLE_, or...

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
```diff
@@ -0,0 +1,190 @@
...
+		List<String> testWatcherMethodNames = getTestWatcherMethodNames();
+
+		LogRecordListener listener = new LogRecordListener();
+		LoggerFactory.addListener(listener);
+		ExecutionEventRecorder recorder = executeTestsForClass(TestWatcherSimpleExceptionHandlingTestCase.class);
+		LoggerFactory.removeListener(listener);
+
+		assertAll(
+			() -> assertEquals(8,
+				listener.stream().filter(l -> l.getSourceClassName().equals(TestMethodTestDescriptor.class.getName())
```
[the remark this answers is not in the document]
sbrannen: Also, you should be able to replace the `stream()` invocation with `org.junit.platform.commons.logging.LogRecordListener.stream(Class<?>)` and simplify the filter.

geo-m:
Requested changes and first draft of documentation and javadoc added for review.

There is a change i am considering. By TestExecutionResult's contract, throwables in cases of aborted or failed tests may be null. Although i dont see any case in jupiter engine where this may happen, i should probably keep it as an Optional<Throwable> even in Watcher methods.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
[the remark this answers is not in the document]
geo-m: @sbrannen thanks for the tips!

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
marcphilipp: **Team Decision:** Let's make them `MAINTAINED` like the other methods in `Node`.

marcphilipp:
Correct. 👍

sbrannen:
> @marcphilipp Rebased since the PR has been idle for a long time, and made a couple of changes.

Thanks for rebasing on `master`.

What is the status of this PR? Are you planning on making further changes?

sbrannen:
@marcphilipp, FYI: your review is still _pending_.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/Node.java
sbrannen: The new `nodeSkipped()` and `nodeFinished()` methods still need `@API` declarations (with `since = "1.4"`), and the `@since` tags need to be changed to `1.4`.

geo-m:
@sbrannen due to conscription ( = unable to set up a working environment), im not planning any changes besides the ones you might require.

geo-m:
@sbrannen due to conscription ( = unable to set up a working environment), im not planning any changes besides the ones you might require.

sbrannen:
Thanks for adding the `@API` declarations.

I'm not sure if we'll keep in the `consumers` declaration in place, since we've yet to do that elsewhere, but I am accepting your changes in any case. 👍

sbrannen:
> @sbrannen due to conscription ( = unable to set up a working environment), im not planning any changes besides the ones you might require.

OK. Thanks for letting us know.

We'll probably take the PR from here...

sbrannen:
@marcphilipp, would you please approve the changes (if you're OK with them) or dismiss your review? 😇

sbrannen:
**Update**: I have marked most discussions as _resolved_, except the following which may still need to be addressed after the PR has been merged into `master`.

https://github.com/junit-team/junit5/pull/1393#discussion_r205951408

sbrannen:
> Agreed! Please open a new issue after merging this PR so we don't forget to address the remaining items for 5.4 M1.

Sure... I'll do _my best_ to recall to do that. 😉

sbrannen:
Local merge is currently _in progress_.

Please do not make any further changes to this PR.

sbrannen:
This has been merged into master in d1cee80903803e30fdb58c968039e55b10579d0d and further refined in 6ad7622c5fe6f42549f42d8bab807dafc7d408e8 and d1cee80903803e30fdb58c968039e55b10579d0d.

sbrannen:
Thanks for the contribution, @geo-m!

sbrannen:
@geo-m, I wanted to _mention_ you in [this Tweet](https://twitter.com/sam_brannen/status/1084128633126563841), but I couldn't find you on Twitter. Do you have a Twitter account?

geo-m:
@sbrannen That's great news.I do not have a twitter account but thank you nonetheless. I look forward to contributing more!

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java
sbrannen: FYI: the following is what I was hinting at...

https://github.com/junit-team/junit5/blob/d1cee80903803e30fdb58c968039e55b10579d0d/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestWatcherTests.java#L164-L166

vyarosh:
Hey guys! 
I see that the issue has been checked "Change is documented in the User Guide and Release Notes".
But the reason of my comment is that I've failed to find it in the relevant user guide [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/)
Could you please replay me a link on any mention of this extension? I think it's quite important to be mentioned in guide. 
Thank you!

sormuras:
https://junit.org/junit5/docs/current/user-guide/#extensions-test-result-processing

vyarosh:
@sormuras sweet, thanks a lot! I wonder how I could missed it...

PR #1401, opened 2018-05-02
Add assertThrows overloads that take a ThrowingSupplier

### LINKED ISSUE #1394: Improve assertThrows failure message if the lambda returns a result instead of throwing
## Overview

When testing code whose expected behavior is to throw an exception, it is sometimes useful to have a custom failure message for the case where the code does not throw an exception but instead returns a value:

```java
try {
  double processedValue = process(bytes);
  fail("Expected an exception, but got " + processedValue);
} catch (IllegalArgumentException expected) {
}
```

One of the few disadvantages of using `assertThrows` instead of the try/catch/fail pattern above is that `assertThrows` does not provide an easy way to customize the failure message to include a result that was returned when an exception was not thrown.

One might be tempted to write something like:

```java
assertThrows(
   IllegalArgumentException.class,
   () -> {
      double processedValue = process(bytes);
      fail("Expected an exception, but got " + processedValue);
   });
```

... but in general, it's a bad idea to have more than one statement in the body of the lambda passed to `assertThrows`, since in the case where the test passes those additional assertions will never be executed:

```java
ImmutableList<Integer> xs = ImmutableList.of();
assertThrows(
    UnsupportedOperationException.class,
    () -> {
        xs.add(0);
        assertThat(xs).isEmpty(); // oops, never executed!
    });
```

What do you think about adding support to `assertThrows` for a custom failure message that includes the result of the expression that was expected to throw an exception?

## Deliverables

This can be achieved without adding an additional parameter to the `assertThrows` API, e.g.:

```java
public static <T extends Throwable> T assertThrows(
    Class<T> expectedType, ThrowingSupplier<?> supplier) { ... }
```

If the provided supplier returns a value, the implementation can include its string representation in the failure message.

Adding `ThrowingSupplier` overloads avoids adding another parameter to the API and allows existing code that is passing `ThrowingSupplier`-compatible lambdas to the existing `assertThrows` methods to get the improved failure messages for free.

For an example of what the implementation might look like, see:
https://github.com/junit-team/junit5/compare/master...cushon:assertThrows

cushon:
If the given ThrowingSupplier fails to throw the expected exception and
instead returns a value, the string representation of the value is
included in the failure message to aide debugging.

Fixes #1394

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java
@@ -18,6 +18,7 @@ import static org.junit.jupiter.api.AssertionUtils.nullSafeGet;
 import java.util.function.Supplier;
 
 import org.junit.jupiter.api.function.Executable;
+import org.junit.jupiter.api.function.ThrowingSupplier;
 import org.opentest4j.AssertionFailedError;
 
 /**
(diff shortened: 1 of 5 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java
```diff
@@ -66,8 +87,16 @@ private AssertThrows() {
 		}
 
 		String message = buildPrefix(nullSafeGet(messageOrSupplier))
-				+ String.format("Expected %s to be thrown, but nothing was thrown.", getCanonicalName(expectedType));
+				+ String.format("Expected %s to be thrown, but nothing was thrown.", getCanonicalName(expectedType))
+				+ (result != null ? String.format(" (returned %s)", result) : "");
```
sbrannen: As mentioned in #1394...

> In any case, we cannot implement it as you have it currently in your PoC since a method may actually return `null`.
> 
> In other words, we would have to maintain the current behavior for `Executable` and have different behavior for `ThrowingSupplier`.

cushon:
I updated the release notes and added some detail to the javadoc for the new overloads. Any suggestions for improvements to the text are welcome :)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -1169,6 +1169,23 @@ public static void assertAll(String heading, Stream<Executable> executables) thr
...
+	 * <em>Asserts</em> that execution of the supplied {@code executable} throws
+	 * an exception of the {@code expectedType} and returns the exception.
+	 *
+	 * <p>If no exception is thrown, or if an exception of a different type is
+	 * thrown, this method will fail.
+	 *
+	 * <p>If you do not want to perform additional checks on the exception instance,
+	 * simply ignore the return value.
+	 *
+	 * <p>If the given {@link ThrowingSupplier<?>} returns a result instead of throwing the expected exception,
```
sormuras: Checkstyle choked on the `<` character. If I'm not mistaken you may omit the `<?>`.

sormuras:
Thanks @cushon!

Issue #395, opened 2016-07-12
Will JUnit 5 support anything like @Theories from JUnit 4?


childnode:
hi,

yes, I asked my aunt google for quickly but not digged too deep in any discussion: so please regret this question ;)

since [Theories](https://github.com/junit-team/junit4/wiki/Theories) had some nice ideas and Popper has absorbed it but then "accidently" died:
❓ anything new, cool aside Parameterized tests in JUnit5?

Anything that will adress Theories?

Anything that will help to write a test class with different dataproviders for different methods?  `i.e. > run test method1 with {A,B} run test method2 with {C,D}`

Anything that will help to test combinations of different dataProvider data sets?!
`i.e. > for two dataprovider {A,B} and {C,D} run each test method with {AC,AD,BC,BD}`

/me is just bored from rewriting workarounds and pasting snippets for test class templates

~Marcel
# documentation

marcphilipp:
Parameterized tests are on our agenda for 5.0.0. We'll consider providing an option of how to combine multiple data providers. However, it's a bit early to say what they will look like exactly. On the one hand we would really want to enable using `Streams` to use large numbers of input parameters. On the other hand, this might hinder combining them to use all possible combinations.

nicolaiparlog:
I like JUnit's [core principle](https://github.com/junit-team/junit5/wiki/Core-Principles) to "prefer extension points over features". Could this be an case for that?

If there were a good way to write extensions that create dynamic tests (e.g. with #371), then extensions could try different approaches to how parameterized tests could be created with maximum ease (likely for different scenarios).

This does not exclude JUnit5 from shipping with such a feature but it would be great if it could itself only use the extension point (much like `@Disabled`) and have to "privileged access" to internals.

smoyer64:
This is yet another use case where the AfterDiscoveryCallback proposed in #354 could be used to "expand" the number of tests run (similar to JUnitParams and UpREST).

childnode:
hi there,

thanks for the input. @marcphilipp I'm somewhat confused about the development cycle ... you have released a "milestone" but do not have a plan which "milestones" are coming; but: a forecasted release date of late 2016; no proposals for old features or how to migrate to Junit5 yet? 
While I'm not saying it's wrong to +not+ make a "upfront, shitty big-pictured-waterfalled-detail-plan" :)

@smoyer64 I'll have a look at this, thx

~Marcel

marcphilipp:
[...]

PR #1422, opened 2018-05-16
JUnit 5 Theories

nrllewellyn:
## Overview

(Re-opening as a separate pull request, since first PR was corrupted by a botched Git rebase.)

Adds theories, in a form similar to JUnit 4's Theories. This will close ticket #395.

Substantial changes from JUnit 4:

- Methods allowed as sources of data points
- Streams and Iterators as accepted data point types, but only when produced by methods (since both are stateful, they have to come from a method so they can be read more than once)
- Suppliers system for simple data points (similar in form to parameterized tests, but allows mixing of in-line data points with class-level data points)
  - Specified via the `ArgumentsSuppliedBy` meta-annotation. This ties the annotation to a supplier class
  - Supplier class implements `TheoryArgumentSupplier` (or extends `AbstractTheoryArgumentSupplier` for common use cases). This class reads the annotation and produces the corresponding data points
  - Examples:
```
@Theory
public void myTheory(@IntValues(0, 42, 99) int someIntValue)

@Theory
public void myTheory(@ByteValues(0, 42) byte somByteValue, SomeClass valueFromClassLevelDataPoint)
```
- Qualifiers directly on theory parameters. Example:
```
@DataPoints(qualifiers = "nonZero")
private static final List<Double> GREATER_THAN_ZERO = Arrays.asList(3, 5, 7);

@DataPoint
private static final Double ZERO = 0;

@DataPoint(qualifiers = "nonZero")
private static final double NEGATIVE_NUMBER =  -4;

@Theory
public void testDivisionOperator(double dividend, @Qualifiers("nonZero") double divisor)
```

A few (very rough) examples can be found in:
junit-jupiter-theories/src/test/java/org/junit/jupiter/theories/TheoriesTestCase.java

---

---

### Definition of Done


marcphilipp:
@nrllewellyn Thanks for the PR! Before doing a detailed review, I wanted to let you know that we discussed whether or not we want to maintain a Theories equivalent in the JUnit 5 codebase in the JUnit 5 team. We've come to the conclusion that property-based testing may be better off in a custom test engine instead of in Jupiter. For example, @jlink has written [JQwik](http://jqwik.net/) which already supports a lot more features than JUnit 4's theories, e.g. shrinking, not reporting every invocation separately (as `TestTemplates` do) etc. If you feel passionate enough to maintain this extension as a stand-alone project, I'd be happy to give you feedback on the implementation. If you do, please add it to the [wiki page](https://github.com/junit-team/junit5/wiki/Third-party-Extensions). However, please understand that we don't want to include it into JUnit core at this point.

nrllewellyn:
@marcphilipp  Thanks for the detailed info. Perhaps my team uses Theories differently than they were originally intended--we always used them as parameterized tests with "try all combinations of parameters" functionality, whereas JQwik seems to focus more on random data.

That said, it's obviously your team's call on what you let into your library. :-)

Thanks for the offer of the code review. Let me read up a bit more on JQwik and some of the other available extensions first--I don't want to overlap too much with what's already available. I'll get back to you if I want to go ahead with maintaining it as a separate library.

marcphilipp:
The fact that @sormuras also recently has written a sample [`CartesianProductProvider`](https://github.com/junit-team/junit5-samples/blob/master/junit5-jupiter-extensions/src/main/java/com/example/cartesian/CartesianProductProvider.java) extension is another indication that there seems to be a need for something like that. I think this could be addressed by a `ArgumentsProvider` implementation for `@ParameterizedTests`, e.g. one that collects data points from fields methods.

sormuras:
Sounds like a plan.

marcphilipp:
I've created #1427. Let's continue the discussion there.

mxmlnglt:
> The fact that @sormuras also recently has written a sample [`CartesianProductProvider`](https://github.com/junit-team/junit5-samples/blob/master/junit5-jupiter-extensions/src/main/java/com/example/cartesian/CartesianProductProvider.java) extension...

404... the class now resides at [CartesianProductProvider.java](https://github.com/junit-team/junit-examples/blob/main/junit-jupiter-extensions/src/main/java/com/example/cartesian/CartesianProductProvider.java) 😉

Issue #1178, opened 2017-11-22
DynamicTest should be able to provide tags and source information to TestDescriptor


jhorstmann:
## Overview

Currently the source for dynamic tests is a always a `MethodSource` pointing to the `@TestFactory` annotated method. In case the dynamic tests are based on external files, like in a cucumber implementation, it should be possible to specify the source.

### Example

```gherkin
Given a file "helloword.feature" containing
  """
  Feature: Hello World Feature

    @hello @world
    Scenario: Hello World Scenario
        Given the user name is 'World'
         When the user clicks the hello button
         Then the output is 'Hello World'
  """
  When dynamic tests are created for this feature
  Then the test description should have a file source of "helloworld.feature" line 4
   And the test description should have the tags ["hello", "world"]
```
## Related Issues

- #434
- #1293 

## Deliverables

- [ ] `DynamicTest` / `DynamicContainer` should have additional factory methods allowing to specify `TestSource` and `Set<TestTag>`
- [ ] The above information should be used when creating the `DynamicTestTestDescriptor` and `DynamicContainerTestDescriptor`

(The actual creation of the dynamic tests for a cucumber scenario is out of scope)

sbrannen:
Interesting idea.

However, there are some problems with the proposal.

First and foremost, both `TestTag` and `TestSource` reside in the `junit-platform-engine` artifact and are therefore not visible from the JUnit Jupiter programming model.

With regard to tags, it's less of an issue since tags could be supplied as `Set<String>` instead of `Set<TestTag>`; whereas, providing support for something like `TestSource` would require a new API that can be converted into `TestSource`.

Another thing to keep in mind is that tags are processed during the overall _discovery_ phase; whereas, dynamic tests and containers are registered during the _execution_ phase. So additional filtering by tags would have to be performed for dynamically registered tests and containers as well.

@junit-team/junit-lambda, thoughts?

sbrannen:
Related to:


- #434
- #1293

robeden:
I would also like to be able to have DynamicTests deal with tags. Perhaps the two portions of this request should be broken apart so they could be dealt with separately?

> Another thing to keep in mind is that tags are processed during the overall discovery phase; whereas, dynamic tests and containers are registered during the execution phase. So additional filtering by tags would have to be performed for dynamically registered tests and containers as well.

I don't like it as well as the framework dealing with it, but I suppose the TestFactory could do the filtering if it had access to information about what's included/excluded for the run.

[...]

PR #1428, opened 2018-05-22
Add TestSource support to dynamic containers and tests

sormuras:
## Overview

Addresses #1178

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.3.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.3.0-M1.adoc
@@ -65,6 +65,9 @@ on GitHub.
   _alias_ for `Arguments.of()`. `arguments()` is intended to be used via `import static`.
 * New `get<Class>(index)` Kotlin extension method to make `ArgumentsAccessor` friendlier
   to use from Kotlin.
+* New support for supplying a custom test source `URI` when creating a dynamic container
+  or test. See factory methods `dynamicContainer(String, URI, ...)` in `DynamicContainer`
+  and `dynamicTest(String, URI, Executable)` in `DynamicTest` for details.
 
 
 [[release-notes-5.3.0-M1-junit-vintage]]
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
@@ -71,13 +71,13 @@ public class DynamicContainer extends DynamicNode {
 
 	/**
 	 * Factory for creating a new {@code DynamicContainer} for the supplied display
-	 * name, the test source uri, and stream of dynamic nodes.
+	 * name, the test source {@link URI}, and stream of dynamic nodes.
 	 *
 	 * <p>The stream of dynamic nodes must not contain {@code null} elements.
 	 *
 	 * @param displayName the display name for the dynamic container; never
 	 * {@code null} or blank
-	 * @param testSourceUri the test source uri for the dynamic test; can be {@code null}
+	 * @param testSourceUri the test source URI for the dynamic test; can be {@code null}
 	 * @param dynamicNodes stream of dynamic nodes to execute;
 	 * never {@code null}
 	 * @since 5.3
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
@@ -12,6 +12,7 @@ package org.junit.jupiter.api;
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
+import java.net.URI;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
@@ -49,7 +50,7 @@ public class DynamicContainer extends DynamicNode {
 	 * @see #dynamicContainer(String, Stream)
 	 */
 	public static DynamicContainer dynamicContainer(String displayName, Iterable<? extends DynamicNode> dynamicNodes) {
-		return new DynamicContainer(displayName, StreamSupport.stream(dynamicNodes.spliterator(), false));
+		return dynamicContainer(displayName, null, StreamSupport.stream(dynamicNodes.spliterator(), false));
 	}
 
 	/**
@@ -65,13 +66,32 @@ public class DynamicContainer extends DynamicNode {
 	 * @see #dynamicContainer(String, Iterable)
 	 */
 	public static DynamicContainer dynamicContainer(String displayName, Stream<? extends DynamicNode> dynamicNodes) {
-		return new DynamicContainer(displayName, dynamicNodes);
+		return dynamicContainer(displayName, null, dynamicNodes);
+	}
+
+	/**
+	 * Factory for creating a new {@code DynamicContainer} for the supplied display
+	 * name, the test source uri, and stream of dynamic nodes.
+	 *
+	 * <p>The stream of dynamic nodes must not contain {@code null} elements.
+	 *
+	 * @param displayName the display name for the dynamic container; never
+	 * {@code null} or blank
+	 * @param testSourceUri the test source uri for the dynamic test; can be {@code null}
+	 * @param dynamicNodes stream of dynamic nodes to execute;
+	 * never {@code null}
+	 * @since 5.3
+	 * @see #dynamicContainer(String, Iterable)
+	 */
+	public static DynamicContainer dynamicContainer(String displayName, URI testSourceUri,
+			Stream<? extends DynamicNode> dynamicNodes) {
+		return new DynamicContainer(displayName, testSourceUri, dynamicNodes);
 	}
 
 	private final Stream<? extends DynamicNode> children;
 
-	private DynamicContainer(String displayName, Stream<? extends DynamicNode> children) {
-		super(displayName);
+	private DynamicContainer(String displayName, URI testSourceUri, Stream<? extends DynamicNode> children) {
+		super(displayName, testSourceUri);
 		Preconditions.notNull(children, "children must not be null");
 		this.children = children;
 	}
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
@@ -42,16 +42,21 @@ public abstract class DynamicNode {
 
 	/**
 	 * Get the display name of this {@code DynamicNode}.
+	 *
+	 * @return the display name
 	 */
 	public String getDisplayName() {
 		return this.displayName;
 	}
 
 	/**
-	 * Get the optional test source of this {@code DynamicNode}.
+	 * Get the optional test source {@link URI} of this {@code DynamicNode}.
+	 *
+	 * @return an {@code Optional} containing the test source {@link URI};
+	 * never {@code null} but potentially empty
 	 * @since 5.3
 	 */
-	public Optional<URI> getTestSourceURI() {
+	public Optional<URI> getTestSourceUri() {
 		return Optional.ofNullable(testSourceUri);
 	}
 
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
@@ -12,6 +12,9 @@ package org.junit.jupiter.api;
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
+import java.net.URI;
+import java.util.Optional;
+
 import org.apiguardian.api.API;
 import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.commons.util.ToStringBuilder;
@@ -29,8 +32,12 @@ public abstract class DynamicNode {
 
 	private final String displayName;
 
-	DynamicNode(String displayName) {
+	/** Custom test source {@link URI} instance associated with this node; potentially {@code null}. */
+	private final URI testSourceUri;
+
+	DynamicNode(String displayName, URI testSourceUri) {
 		this.displayName = Preconditions.notBlank(displayName, "displayName must not be null or blank");
+		this.testSourceUri = testSourceUri;
 	}
 
 	/**
@@ -40,9 +47,20 @@ public abstract class DynamicNode {
 		return this.displayName;
 	}
 
+	/**
+	 * Get the optional test source of this {@code DynamicNode}.
+	 * @since 5.3
+	 */
+	public Optional<URI> getTestSourceURI() {
+		return Optional.ofNullable(testSourceUri);
+	}
+
 	@Override
 	public String toString() {
-		return new ToStringBuilder(this).append("displayName", displayName).toString();
+		return new ToStringBuilder(this) //
+				.append("displayName", displayName) //
+				.append("testSourceUri", testSourceUri) //
+				.toString();
 	}
 
 }
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
@@ -65,11 +65,11 @@ public class DynamicTest extends DynamicNode {
 
 	/**
 	 * Factory for creating a new {@code DynamicTest} for the supplied display
-	 * name, the test source uri, and executable code block.
+	 * name, the test source {@link URI}, and executable code block.
 	 *
 	 * @param displayName the display name for the dynamic test; never
 	 * {@code null} or blank
-	 * @param testSourceUri the test source uri for the dynamic test; can be {@code null}
+	 * @param testSourceUri the test source URI for the dynamic test; can be {@code null}
 	 * @param executable the executable code block for the dynamic test;
 	 * never {@code null}
 	 * @since 5.3
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
@@ -14,6 +14,7 @@ import static java.util.Spliterator.ORDERED;
 import static java.util.Spliterators.spliteratorUnknownSize;
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
+import java.net.URI;
 import java.util.Iterator;
 import java.util.function.Function;
 import java.util.stream.Stream;
@@ -59,7 +60,23 @@ public class DynamicTest extends DynamicNode {
 	 * @see #stream(Iterator, Function, ThrowingConsumer)
 	 */
 	public static DynamicTest dynamicTest(String displayName, Executable executable) {
-		return new DynamicTest(displayName, executable);
+		return new DynamicTest(displayName, null, executable);
+	}
+
+	/**
+	 * Factory for creating a new {@code DynamicTest} for the supplied display
+	 * name, the test source uri, and executable code block.
+	 *
+	 * @param displayName the display name for the dynamic test; never
+	 * {@code null} or blank
+	 * @param testSourceUri the test source uri for the dynamic test; can be {@code null}
+	 * @param executable the executable code block for the dynamic test;
+	 * never {@code null}
+	 * @since 5.3
+	 * @see #stream(Iterator, Function, ThrowingConsumer)
+	 */
+	public static DynamicTest dynamicTest(String displayName, URI testSourceUri, Executable executable) {
+		return new DynamicTest(displayName, testSourceUri, executable);
 	}
 
 	/**
@@ -101,8 +118,8 @@ public class DynamicTest extends DynamicNode {
 
 	private final Executable executable;
 
-	private DynamicTest(String displayName, Executable executable) {
-		super(displayName);
+	private DynamicTest(String displayName, URI testSourceUri, Executable executable) {
+		super(displayName, testSourceUri);
 		this.executable = Preconditions.notNull(executable, "executable must not be null");
 	}
 
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
@@ -81,7 +81,7 @@ public class TestFactoryTestDescriptor extends TestMethodTestDescriptor implemen
 			Object instance = extensionContext.getRequiredTestInstance();
 			Object testFactoryMethodResult = executableInvoker.invoke(getTestMethod(), instance, extensionContext,
 				context.getExtensionRegistry());
-			TestSource source = getSource().orElseThrow(
+			TestSource defaultTestSource = getSource().orElseThrow(
 				() -> new JUnitException("Illegal state: TestSource must be present"));
 			try (Stream<DynamicNode> dynamicNodeStream = toDynamicNodeStream(testFactoryMethodResult)) {
 				int index = 1;
@@ -89,7 +89,7 @@ public class TestFactoryTestDescriptor extends TestMethodTestDescriptor implemen
 				while (iterator.hasNext()) {
 					DynamicNode dynamicNode = iterator.next();
 					Optional<JupiterTestDescriptor> descriptor = createDynamicDescriptor(this, dynamicNode, index++,
-						source, getDynamicDescendantFilter());
+						defaultTestSource, getDynamicDescendantFilter());
 					descriptor.ifPresent(dynamicTestExecutor::execute);
 				}
 			}
@@ -117,10 +117,11 @@ public class TestFactoryTestDescriptor extends TestMethodTestDescriptor implemen
 	}
 
 	static Optional<JupiterTestDescriptor> createDynamicDescriptor(JupiterTestDescriptor parent, DynamicNode node,
-			int index, TestSource testSource, DynamicDescendantFilter dynamicDescendantFilter) {
+			int index, TestSource defaultTestSource, DynamicDescendantFilter dynamicDescendantFilter) {
 		UniqueId uniqueId;
 		Supplier<JupiterTestDescriptor> descriptorCreator;
-		TestSource source = computeOptionalTestSource(node).orElse(testSource);
+		Optional<TestSource> optionalTestSource = node.getTestSourceUri().map(UriSource::from);
+		TestSource source = optionalTestSource.orElse(defaultTestSource);
 		if (node instanceof DynamicTest) {
 			DynamicTest test = (DynamicTest) node;
 			uniqueId = parent.getUniqueId().append(DYNAMIC_TEST_SEGMENT_TYPE, "#" + index);
@@ -138,10 +139,6 @@ public class TestFactoryTestDescriptor extends TestMethodTestDescriptor implemen
 			return Optional.of(descriptor);
 		}
 		return Optional.empty();
-	}
-
-	static Optional<TestSource> computeOptionalTestSource(DynamicNode node) {
-		return node.getTestSourceURI().map(UriSource::from);
 	}
 
 }
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
@@ -30,6 +30,7 @@ import org.junit.platform.commons.util.CollectionUtils;
 import org.junit.platform.commons.util.PreconditionViolationException;
 import org.junit.platform.engine.TestSource;
 import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.descriptor.UriSource;
 
 /**
  * {@link org.junit.platform.engine.TestDescriptor TestDescriptor} for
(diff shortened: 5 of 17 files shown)


sormuras:
Including `junit-platform-engine` as an optional dependency via...

```groovy
plugins {
	id 'nebula.optional-base' version '3.0.3'
}

description = 'JUnit Jupiter API'

dependencies {
	api("org.opentest4j:opentest4j:${ota4jVersion}")
	api(project(':junit-platform-commons'))
	compile(project(':junit-platform-engine'), optional)
	compileOnly('org.jetbrains.kotlin:kotlin-stdlib')
}
```

...is not necessary. `compileOnly` deps are not listed in the generated `pom.xml`.

sbrannen:
> `compileOnly` deps are not listed in the generated `pom.xml`.

That's why we don't want to use `compileOnly`, since we _do_ want `junit-platform-engine` listed in the generated `pom.xml` as an `optional` dependency.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
```diff
@@ -40,6 +45,29 @@ public String getDisplayName() {
...
+		return Optional.ofNullable(testSource);
+	}
+
+	DynamicNode setTestSource(TestSource testSource) {
+		this.testSource = testSource;
+		return this;
+	}
+
+	/**
+	 * Create new instance of this dynamic container or test with a custom test source.
```
sbrannen: Why would we want to create a new instance?

Why not just return `this` to support method chaining?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
```diff
@@ -40,6 +45,29 @@ public String getDisplayName() {
...
+	 *
+	 * @param testSource custom test source instance to associate with this node
+	 * @return new node instance
+	 * @see #getTestSource()
+	 */
+	public abstract DynamicNode withTestSource(TestSource testSource);
```
sbrannen: How about just implementing this instead of making it `abstract`?

Or... do you foresee introducing additional methods in a pseudo-builder pattern style in the future?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
```diff
@@ -84,4 +85,10 @@ private DynamicContainer(String displayName, Stream<? extends DynamicNode> child
 		return children;
 	}
 
+	@Override
+	public DynamicContainer withTestSource(TestSource testSource) {
+		DynamicContainer clone = new DynamicContainer(getDisplayName(), getChildren());
```
sbrannen: methinks we should avoid _cloning_ whenever possible.

Similarly, if we have copy-constructor, why not pass in the test source as well? In other words, do we need to make the test source field mutable?

junit-jupiter-api/junit-jupiter-api.gradle
```diff
@@ -2,8 +2,9 @@ description = 'JUnit Jupiter API'
 
 dependencies {
 	api("org.opentest4j:opentest4j:${ota4jVersion}")
-	api(project(":junit-platform-commons"))
-	compileOnly("org.jetbrains.kotlin:kotlin-stdlib")
+	api(project(':junit-platform-commons'))
+	compileOnly(project(':junit-platform-engine'))
```
sbrannen: Let's ensure this shows up as an `optional` dependency in the generated Maven POM (see other comment in PR).

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
```diff
@@ -113,4 +114,10 @@ public Executable getExecutable() {
 		return this.executable;
 	}
 
+	@Override
+	public DynamicTest withTestSource(TestSource testSource) {
+		DynamicTest clone = new DynamicTest(getDisplayName(), getExecutable());
+		clone.setTestSource(testSource);
```
sbrannen: ditto, WRT cloning, etc.

sormuras:
Team decision: introduce static factory methods taking 3 parameters instead of the `withTestSource()` approach:

### DynamicTest
`dynamicTest(String displayName, TestSource testSource, Executable executable) `

### DynamicContainer
`dynamicContainer(String displayName, TestSource testSource, Iterable<DynamicNode> dynamicNodes)`

sormuras:
> ... listed in the generated pom.xml as an optional dependency.

It's listed now.

sbrannen:
Regarding `optional`, thanks!

`compile(project(':junit-platform-engine'), optional)`

I didn't realize Gradle supported that `optional` flag there.

Is that a new Gradle feature?

sormuras:
>  Is that a new Gradle feature?

No, `optional` is introduced by https://github.com/nebula-plugins/gradle-extra-configurations-plugin

sbrannen:
Ahhh.... OK. Thanks for the link!

I somehow overlooked the corresponding addition in the `plugins` section.

sormuras:
@sbrannen Rebased and refactored the PR. Please review again. Thanks!

sormuras:
@sbrannen Could you please do a quick review again?

documentation/src/docs/asciidoc/release-notes/release-notes-5.3.0-M1.adoc
```diff
@@ -51,7 +52,10 @@ on GitHub.
   _alias_ for `Arguments.of()`. `arguments()` is intended to be used via `import static`.
 * New `get<Class>(index)` Kotlin extension method to make `ArgumentsAccessor` friendlier
   to use from Kotlin.
-
+* New support for supplying a custom test source when creating a dynamic container or
+  test. See factory methods `dynamicContainer(String, TestSource, ...)` in
+  `DynamicContainer` and `dynamicTest.(String, TestSource, Executable)` in `DynamicTest`
```
sbrannen: Please remove the extra dot (`.`).

junit-jupiter-api/junit-jupiter-api.gradle
```diff
@@ -1,9 +1,14 @@
...
+}
+
 description = 'JUnit Jupiter API'
 
 dependencies {
 	api("org.opentest4j:opentest4j:${ota4jVersion}")
-	api(project(":junit-platform-commons"))
-	compileOnly("org.jetbrains.kotlin:kotlin-stdlib")
+	api(project(':junit-platform-commons'))
+	compile(project(':junit-platform-engine'), optional)
```
sbrannen: Since the "optional" is easy to overlook, it might be helpful to add a comment pointing out that `optional` is handled by the nebula plug-in.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
```diff
@@ -65,13 +66,31 @@ public static DynamicContainer dynamicContainer(String displayName, Iterable<? e
 	 * @see #dynamicContainer(String, Iterable)
 	 */
 	public static DynamicContainer dynamicContainer(String displayName, Stream<? extends DynamicNode> dynamicNodes) {
-		return new DynamicContainer(displayName, dynamicNodes);
+		return dynamicContainer(displayName, null, dynamicNodes);
+	}
+
+	/**
+	 * Factory for creating a new {@code DynamicContainer} for the supplied display
+	 * name, the test source and stream of dynamic nodes.
```
sbrannen: "the test source and stream of dynamic nodes." --> "test source, and stream of dynamic nodes."

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
```diff
@@ -83,5 +102,4 @@ private DynamicContainer(String displayName, Stream<? extends DynamicNode> child
 	public Stream<? extends DynamicNode> getChildren() {
 		return children;
 	}
-
```
sbrannen: IIRC, it's our policy (albeit perhaps unspoken) to include a blank line before the closing brace.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
```diff
@@ -29,8 +32,12 @@
 
 	private final String displayName;
 
-	DynamicNode(String displayName) {
+	/** Custom test source instance associated with this node; potentially {@code null}. */
+	private final TestSource testSource;
```
sbrannen: Alarms going off....

Unless I'm missing something, the direct dependency on `TestSource` in a field makes **all** dynamic containers and dynamic tests dependent on the `junit-platform-engine` module.

Consequently, the "optional" dependency on `junit-platform-engine` is meaningless.

In other words, with the hard dependency on `TestSource` it will be impossible to use dynamic tests without having the `junit-platform-engine` in the classpath.

So, @junit-team/junit-lambda, we need to rethink this.

sbrannen: This brings us back to my original comment on the matter: https://github.com/junit-team/junit5/issues/1178#issuecomment-346371161

> providing support for something like `TestSource` would require a new API that can be converted into `TestSource`.

sbrannen: Never mind: as @sormuras pointed out, the `TestSource` is indeed a run-time dependency but not necessarily a compile-time dependency. So that should allow it to be "optional".

But let's add some integration tests to make sure all is well. 😉

sormuras:
Team said: refactor to not depend directly on `junit-jupiter-engine` types. Instead provide a `java.nio.path.Path` based API.

jhorstmann:
Implementing this without dependency on the actual implementation sounds good. However, I think that `URI` might be the better abstraction instead of `Path`:

 - It's normally not needed to access the contents of a `TestSource`
 - It could directly map to the existing `UriSource` class
 - It would allow specifying a classpath resource as a source using the `classpath:` or `jar:` schemes, or any other scheme that is not directly resolvable using `FileSystem`

It would be nice to support an optional position (line, column) inside the source, but for that I think the existing `TestSource` hierarchy should be refactored a little to extract a `PositionAware` interface.

sormuras:
> It would be nice to support an optional position (line, column) inside the source, but for that I think the existing TestSource hierarchy should be refactored a little to extract a PositionAware interface.

Mh, we could use the `fragment` part of a `URI` to encode file positions.

sbrannen:
Instead of `fragment`, I'd recommend the use of query parameters.

Here's a hacked proof of concept:

```java
public static void main(String[] args) throws Exception {
	File currentDir = new File(".").getCanonicalFile();
	URI uri = new URI("file://" + currentDir.getAbsolutePath() + "/script.js?line=2&column=3");
	File file = new File(uri.getPath());
	System.err.println("URI:   " + uri);
	System.err.println("File:  " + file);
	System.err.println("Query: " + Arrays.stream(uri.getQuery().split("&")).map(s -> {
		try {
			return URLDecoder.decode(s, "UTF-8");
		}
		catch (UnsupportedEncodingException e) {
			e.printStackTrace();
			return "";
		}
	}).filter(s -> s.trim().length() > 0).collect(toList()));
}
```

The output on my machine looks something like this:

```
URI:   file:///Users/xxx/test/script.js?line=2&column=3
File:  Users/xxx/test/script.js
Query: [line=2, column=3]
```

We'd still have to actually parse the key=value pairs, but it's a start (or rather a _proof of concept_). 😉

sbrannen:
> Perhaps, we should start with a simple `URI` without `query` and/or `fragment` handling for 5.3 M1. If this works out, we'll enhance the `URI` handling in a later milestone release.

That's also an option. 👍

sbrannen:
> Aye. Played with `query` as well but the amount of extra logic is not small.

Well, if we were to support `query`, I'd recommend we simply implement a "best effort" algorithm that only supports `line` and `column` parameters, and if there are any exceptions... just log them and swallow them.

I don't think it's too much code to implement it like that.

My $0.02 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
```diff
@@ -65,13 +66,32 @@ public static DynamicContainer dynamicContainer(String displayName, Iterable<? e
 	 * @see #dynamicContainer(String, Iterable)
 	 */
 	public static DynamicContainer dynamicContainer(String displayName, Stream<? extends DynamicNode> dynamicNodes) {
-		return new DynamicContainer(displayName, dynamicNodes);
+		return dynamicContainer(displayName, null, dynamicNodes);
+	}
+
+	/**
+	 * Factory for creating a new {@code DynamicContainer} for the supplied display
+	 * name, the test source uri, and stream of dynamic nodes.
```
sbrannen: uri -> URI

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
```diff
@@ -65,13 +66,32 @@ public static DynamicContainer dynamicContainer(String displayName, Iterable<? e
...
+	 *
+	 * <p>The stream of dynamic nodes must not contain {@code null} elements.
+	 *
+	 * @param displayName the display name for the dynamic container; never
+	 * {@code null} or blank
+	 * @param testSourceUri the test source uri for the dynamic test; can be {@code null}
```
sbrannen: uri -> URI

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
```diff
@@ -40,9 +47,20 @@ public String getDisplayName() {
 		return this.displayName;
 	}
 
+	/**
+	 * Get the optional test source of this {@code DynamicNode}.
+	 * @since 5.3
+	 */
+	public Optional<URI> getTestSourceURI() {
```
sbrannen: URI -> Uri

sbrannen: missing `@return` tag

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
```diff
@@ -59,7 +60,23 @@
 	 * @see #stream(Iterator, Function, ThrowingConsumer)
 	 */
 	public static DynamicTest dynamicTest(String displayName, Executable executable) {
-		return new DynamicTest(displayName, executable);
+		return new DynamicTest(displayName, null, executable);
+	}
+
+	/**
+	 * Factory for creating a new {@code DynamicTest} for the supplied display
+	 * name, the test source uri, and executable code block.
```
sbrannen: uri -> URI

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
```diff
@@ -59,7 +60,23 @@
...
+	 *
+	 * @param displayName the display name for the dynamic test; never
+	 * {@code null} or blank
+	 * @param testSourceUri the test source uri for the dynamic test; can be {@code null}
```
sbrannen: uri -> URI

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
```diff
@@ -131,11 +140,8 @@ protected void invokeTestMethod(JupiterEngineExecutionContext context, DynamicTe
 		return Optional.empty();
 	}
 
-	private JUnitException invalidReturnTypeException(Throwable cause) {
-		String message = String.format(
-			"@TestFactory method [%s] must return a Stream, Collection, Iterable, or Iterator of %s.",
-			getTestMethod().toGenericString(), DynamicNode.class.getName());
-		return new JUnitException(message, cause);
+	static Optional<TestSource> computeOptionalTestSource(DynamicNode node) {
```
sbrannen: Please rename to `getCustomTestSource()` or similar.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/DefaultUriSource.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.net.URI;
+import java.util.Objects;
+
+import org.apiguardian.api.API;
+import org.junit.platform.commons.util.Preconditions;
+import org.junit.platform.commons.util.ToStringBuilder;
+
+/**
+ * Default uri-based test source implementation.
```
sbrannen: `Default implementation of {@link UriSource}.`

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/DefaultUriSource.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+@API(status = STABLE, since = "1.3")
+class DefaultUriSource implements UriSource {
+
+	private static final long serialVersionUID = 1L;
+
+	private final URI uri;
+
+	DefaultUriSource(URI uri) {
+		this.uri = Preconditions.notNull(uri, "uri must not be null");
```
sbrannen: uri must not be null -> URI must not be null

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
```diff
@@ -52,6 +54,50 @@ public static FilePosition from(int line, int column) {
 		return new FilePosition(line, column);
 	}
 
+	/**
+	 * Create an optional {@code FilePosition} parsing the supplied
```
sbrannen: parsing -> by parsing

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
```diff
@@ -52,6 +54,50 @@ public static FilePosition from(int line, int column) {
...
+	 * {@code query} string.
+	 *
+	 * <p>Examples of valid {@code query} strings:
+	 * <ul>
+	 *     <li>{@code "line=23"}</li>
+	 *     <li>{@code "line=23&column=42"}</li>
+	 * </ul>
+	 *
+	 * @param query the query string; may be {@code null}
+	 * @since 1.3
```
sbrannen: missing `@return` tag

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
```diff
@@ -52,6 +54,50 @@ public static FilePosition from(int line, int column) {
...
+							line = value;
+						}
+						if (key.equals("column")) {
+							column = value;
+						}
+					}
+				}
+			}
+			catch (IllegalArgumentException e) {
+				LoggerFactory.getLogger(FilePosition.class).debug(e, () -> "parsing query failed: " + query);
```
sbrannen: "parsing query failed: " --> "Failed to parse 'line' and/or 'column' from query string: "

sbrannen: Please store the logger in a static field.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
```diff
@@ -33,4 +39,42 @@
...
 
+	/**
+	 * Create a new {@code UriSource} using the supplied {@code uri}.
+	 * <p>
+	 * This implementation tries resolve the {@code uri} to local file
+	 * system path-based source first. If that fails for any reason, an
+	 * instance of a simple default uri source class storing the supplied
+	 * {@code uri} <em>as-is</em> is returned.
+	 *
+	 * @param uri the uri instance; must not be {@code null}
```
sbrannen: uri instance -> URI instance

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
```diff
@@ -33,4 +39,42 @@
...
+	 * @return a uri source instance
+	 * @since 1.3
+	 * @see org.junit.platform.engine.support.descriptor.FileSource
+	 * @see org.junit.platform.engine.support.descriptor.DirectorySource
+	 */
+	static UriSource from(final URI uri) {
+		Preconditions.notNull(uri, "uri must not be null");
```
sbrannen: uri must ... -> URI must ...

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
```diff
@@ -33,4 +39,42 @@
 	 */
 	URI getUri();
 
+	/**
+	 * Create a new {@code UriSource} using the supplied {@code uri}.
```
sbrannen: uri -> URI

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
```diff
@@ -33,4 +39,42 @@
...
+	 * <p>
+	 * This implementation tries resolve the {@code uri} to local file
+	 * system path-based source first. If that fails for any reason, an
+	 * instance of a simple default uri source class storing the supplied
```
sbrannen: an instance of a simple default uri source class --> an instance of the default {@code UriSource} implementation

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
```diff
@@ -33,4 +39,42 @@
...
+			}
+			Path path = Paths.get(pathBasedUri);
+			if (Files.isRegularFile(path)) {
+				return FileSource.from(path.toFile(), FilePosition.fromQuery(query).orElse(null));
+			}
+			if (Files.isDirectory(path)) {
+				return DirectorySource.from(path.toFile());
+			}
+		}
+		catch (IllegalArgumentException e) {
```
sbrannen: Is catching only `IllegalArgumentException` sufficient here?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
```diff
@@ -33,4 +39,42 @@
...
+			LoggerFactory.getLogger(UriSource.class).debug(e, () -> "uri not path-based: " + uri);
```
sbrannen: Please store the logger in a static field.

sbrannen: uri not path based --> The supplied URI [%s] is not path-based. Falling back to default UriSource implementation.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
```diff
@@ -33,4 +39,42 @@
...
+	 * @see org.junit.platform.engine.support.descriptor.FileSource
+	 * @see org.junit.platform.engine.support.descriptor.DirectorySource
+	 */
+	static UriSource from(final URI uri) {
+		Preconditions.notNull(uri, "uri must not be null");
+		try {
+			URI pathBasedUri = uri;
+			String query = pathBasedUri.getQuery();
+			if (StringUtils.isNotBlank(query)) {
+				String s = pathBasedUri.toString();
```
sbrannen: `s` -> something more descriptive please. 😉

platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/FilePositionTests.java
```diff
@@ -43,13 +55,42 @@ void filePositionFromLine() throws Exception {
...
+		}
+
+		assertEquals(-1, expectedLine);
+		assertEquals(-1, expectedColumn);
+	}
+
+	@SuppressWarnings("unused")
+	static Stream<Arguments> filePositionFromQuery() {
+		return Stream.of( //
+			Arguments.of(null, -1, -1), //
```
sbrannen: `Arguments.of` --> `arguments` (via static import 😉 )

sbrannen:
I requested some minor changes, but otherwise I think this PR looks good!

The only issue I see is regarding what exception to catch (see comment in review).

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
sormuras: ✔️

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicContainer.java
sormuras: ✔️

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
sormuras: ✔️x2

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
sormuras: ✔️

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicTest.java
sormuras: ✔️

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
sormuras: Replaced the method by an inline assignment.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/DefaultUriSource.java
sormuras: ✔️

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/DefaultUriSource.java
sormuras: ✔️

sormuras: There's wild mix of lower- and upper-case vs type and parameter name usages of `Preconditions::<T> T notNull(T object, String message)`... could need polishing pass.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
sormuras: ✔️

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
sormuras: ✔️ (adding two more to the other `FilePosition::from` factory methods)

sormuras: Note: all concrete `TestSource` implementation factory methods lack `@return` tags.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
sormuras: ✔️

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
sormuras: ✔️

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
sormuras: ✔️

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
sormuras: ✔️

platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/FilePositionTests.java
sormuras: Sure. 🙂

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
sormuras: Using `RuntimeException` now to catch `java.nio.file.FileSystemNotFoundException` as well. This and `IllegalArgumentException` are thrown by `Paths::get`.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/DefaultUriSource.java
sbrannen: sure.... I imagine I'll take another look in an actual browser once it's all in master.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestFactoryTestDescriptor.java
sbrannen: 👍

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
sbrannen: Aha. Then I guess we should likely address that in a separate commit outside the scope of this PR.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
sbrannen: OK. Sounds better.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/UriSource.java
sbrannen: Oops.  Didn't realize I was commenting on a static factory method in an interface.

sbrannen:
@jhorstmann,

> It would allow specifying a classpath resource as a source using the `classpath:` or `jar:` schemes, or any other scheme that is not directly resolvable using `FileSystem`

See #1467 for the proposed `classpath:` support.

> It would be nice to support an optional position (line, column) inside the source, but for that I think the existing `TestSource` hierarchy should be refactored a little to extract a `PositionAware` interface.

Can you please expound on your rationale for introducing a `PositionAware` interface?

PR #1430, opened 2018-05-25
Fix determineFullyQualifiedClassName() when ZipPath.relativize() returns paths like ../foo

jernst:
## Overview

The JDK's `ZipPath.relativize()` returns paths like `../foo` when attempting to when attempting to look for test classes in packages contained in JAR files. So `determineFullyQualifiedClassName()` constructs "fully qualified class names" that look like `com.example.bar...foo.Someclass` instead of `com.example.foo.Someclass`. Reimplemented the method without using `relativize()`, and simplified.

---

---

### Definition of Done

- [unchanged] Method [preconditions](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [unchanged] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/meta/API.html)
- [N/A] Change is documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [?] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java
@@ -12,7 +12,6 @@ package org.junit.platform.commons.util;
 
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
-import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 import static org.junit.platform.commons.util.BlacklistedExceptions.rethrowIfBlacklisted;
 import static org.junit.platform.commons.util.ClassFileVisitor.CLASS_FILE_SUFFIX;
@@ -30,7 +29,6 @@ import java.util.Optional;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
-import java.util.stream.Stream;
 
 import org.junit.platform.commons.logging.Logger;
 import org.junit.platform.commons.logging.LoggerFactory;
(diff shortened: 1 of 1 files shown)


marcphilipp:
@jernst Thanks for the contribution! When does this happen? We have a test that scans a JAR file. Can you please provide an additional test that fails before your changes and succeeds afterwards?

marcphilipp:
Thanks for the explanation! I still would like to understand when `ZipPath.relativize()` returns paths like `../foo`. This should be pretty easy to reproduce. All we need is the JAR file it happened with, right?

jernst:
Hold your horses. There seems to be a case where this does not work. Pending investigation.

Question: just what exactly is the "contract" of `determineFullyQualifiedClassName()`? There are no comments whatsoever, nor unit tests :-) and I'm not certain any more my guess is correct.

marcphilipp:
There are unit tests for `ClasspathScanner`, e.g. `ClasspathScannerTests.findAllClassesInPackageWithinJarFile()`. `determineFullyQualifiedClassName()` is a private helper method that is supposed to return the FQCN given the base package's `baseDir`, it's name (`basePackageName`), and the class file.

marcphilipp:
Just to be clear: We're not going to merge this PR without a test case that fails without your changes. 😉

marcphilipp:
@jernst Any updates on this one?

jernst:
@marcphilipp no, sorry. Development has taken a different direction and so I have neither time to spend on this any more, nor do I need it at the moment. Perhaps time to close in some fashion?

marcphilipp:
@jernst Do you recall which JDK version you were using?

PR #1443, opened 2018-06-04
Introduce TestInstanceFactory extension API

### LINKED ISSUE #672: Introduce extension API for creating the test instance

kdubb:
## Overview

`TestInstanceFactory` is a new extension API that allows custom creation of test instance objects.

* It must be applied at the class level.
* If multiple `TestInstanceFactory` extensions are found to be registered on a test class an error is thrown.
* Supports `@Nested` classes, ensuring that only one factory is registered on each level.

Issue: #672 

---

---

### Definition of Done


kdubb:
@sbrannen This extension solves issues for test instances that need to be vended in a special manner; specifically this greatly simplifies testing using CDI for [https://github.com/weld/weld-junit](Weld Junit).  Would love to see this or version of this implemented very soon so I thought I'd take a stab at it. Let me know your thoughts and suggested changes, if any.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -201,11 +203,73 @@ private Object instantiateAndPostProcessTestInstance(JupiterEngineExecutionConte
...
+
+		if (factories.size() > 1) {
+			String factoryNames = factories.stream().map(factory -> factory.getClass().getSimpleName()).collect(
+				Collectors.joining(","));
+
+			String errorMessage = String.format(
+				"Too many TestInstanceFactory extensions [%s] registered on test class: %s", factoryNames,
+				testClass.getSimpleName());
+
+			throw new JUnitException(errorMessage);
```
sormuras: At least throw an instance of `ExtensionContextException`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -201,11 +203,73 @@ private Object instantiateAndPostProcessTestInstance(JupiterEngineExecutionConte
...
+	}
+
+	protected Object invokeTestInstanceFactory(Object outerInstance, ExtensionRegistry registry,
+			ExtensionRegistry parentRegistry, ExtensionContext extensionContext) {
+
+		TestInstanceFactory factory = resolveTestInstanceFactory(registry, parentRegistry);
+		if (factory == null) {
+			return invokeTestInstanceConstructor(outerInstance, registry, extensionContext);
+		}
+		else {
```
sormuras: `else` not needed here.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -201,11 +203,73 @@ private Object instantiateAndPostProcessTestInstance(JupiterEngineExecutionConte
...
+			try {
+				if (outerInstance != null) {
+					return factory.instantiateNestedTestClass(this.testClass, outerInstance, extensionContext);
+				}
+				else {
+					return factory.instantiateTestClass(this.testClass, extensionContext);
+				}
+			}
+			catch (RuntimeException exception) {
```
sormuras: Both `catch` are too generic. And please use `BlacklistedExceptions.rethrowIfBlacklisted(cause);`

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TestInstanceFactoryTests.java
```diff
@@ -0,0 +1,156 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestInstanceFactory;
+import org.junit.jupiter.engine.AbstractJupiterTestEngineTests;
+import org.junit.platform.engine.test.event.ExecutionEventRecorder;
+import org.junit.platform.launcher.LauncherDiscoveryRequest;
+
+/**
+ * Integration tests that verify support for {@link TestInstanceFactory}.
+ *
+ * @since 5.0
```
sormuras: `5.3`

kdubb:
@sormuras  Your changes have been pushed

documentation/src/docs/asciidoc/user-guide/extensions.adoc
```diff
@@ -241,6 +241,21 @@ Examples:
 - `org.example.MyCondition`: deactivates the condition whose FQCN is exactly
   `org.example.MyCondition`.
 
+[[extensions-test-instance-factories]]
+=== Test Instance Factories
+`{TestInstanceFactory}` defines the API for `Extensions` that wish to _provide new_
```
sbrannen: A corresponding link must also be added to `/documentation/src/docs/asciidoc/link-attributes.adoc`.

documentation/src/docs/asciidoc/user-guide/extensions.adoc
```diff
@@ -241,6 +241,21 @@ Examples:
...
+test instances.
+
+Common use cases include acquiring the test instance from a dependency injection
+framework or invoking a static factory method to create the test instance.
+
+[WARNING]
+====
+Registering multiple extensions that implement `{TestInstanceFactory}` will result in an
+exception being thrown for all tests in the test class. It is the users responsibility to
```
sbrannen: ... or  test class hierarchy

sbrannen: users --> user's

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstanceCreationException.java
```diff
@@ -0,0 +1,37 @@
...
+import org.junit.platform.commons.JUnitException;
+
+/**
+ * Thrown if an error is encountered with the invocation of
+ * any {@link TestInstanceFactory}.
+ *
+ * @since 5.3
+ */
+@API(status = EXPERIMENTAL, since = "5.3")
+public class TestInstanceCreationException extends JUnitException {
```
sbrannen: I think `TestInstantiationException` would be a more suitable name

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstanceFactory.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * {@code TestInstanceFactory} defines the API for {@link Extension
+ * Extensions} that wish to <em>create</em> test instances.
+ *
+ * <p>Common use cases include creating test instances with special construction
+ * requirements or acquiring the test from a dependency injection framework.
+ *
+ * <p>Extensions that implement {@code TestInstanceFactory} must be
+ * registered at the class level.
+ *
```
sbrannen: The class-level JavaDoc for `TestInstanceFactory` should also mention that only one such extension may be registered per test class or test class hierarchy, analogous to the documentation in the User Guide.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstanceFactory.java
```diff
@@ -0,0 +1,76 @@
...
+	 * called when the {@code testClass} is a {@link Nested} test class.
+	 *
+	 * <p><strong>Note</strong>: the {@code ExtensionContext} supplied to a
+	 * {@code TestInstanceFactory} will always return an empty
+	 * {@link java.util.Optional} value from {@link ExtensionContext#getTestInstance()
+	 * getTestInstance()}. A {@code TestInstanceFactory} should therefore
+	 * only attempt to create the required test instance.
+	 *
+	 * @param testClass the test class to instantiate or otherwise obtain
+	 * @param outerInstance instance of outer test class (if any)
```
sbrannen: If the method is solely dedicated to instantiating a nested test class (i.e., a non-static nested class), it then does not make sense for the `outerInstance` to ever be `null`. Thus the documentation must state `;never {@code null}`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstanceFactory.java
```diff
@@ -0,0 +1,76 @@
...
+	 * @param context the current extension context; never {@code null}
+	 * @return The required test instance; never {@code null}
+	 * @throws TestInstanceCreationException when an error occurs with the invocation of a factory
+	 */
+	Object instantiateNestedTestClass(Class<?> testClass, Object outerInstance, ExtensionContext context)
```
sbrannen: Another option would be to have a single such method in this new API and make the `outerInstance` an `Optional`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -201,11 +205,70 @@ private Object instantiateAndPostProcessTestInstance(JupiterEngineExecutionConte
...
+			ExtensionRegistry parentRegistry) {
+
+		List<TestInstanceFactory> factories = registry.getExtensions(TestInstanceFactory.class);
+		if (factories.isEmpty()) {
+			return null;
+		}
+
+		if (parentRegistry != null) {
+			List<TestInstanceFactory> parentFactories = parentRegistry.getExtensions(TestInstanceFactory.class);
+			factories.removeAll(parentFactories);
```
sbrannen: If we are going to ignore factories registered in parent contexts -- if and only if there is a locally registered factory --  we then need to explicitly document that behavior.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -201,11 +205,70 @@ private Object instantiateAndPostProcessTestInstance(JupiterEngineExecutionConte
...
+			if (outerInstance != null) {
+				return factory.instantiateNestedTestClass(this.testClass, outerInstance, extensionContext);
+			}
+			else {
+				return factory.instantiateTestClass(this.testClass, extensionContext);
+			}
+		}
+		catch (Throwable exception) {
+			BlacklistedExceptions.rethrowIfBlacklisted(exception);
+			throw exception;
```
sbrannen: This combination does not make sense.

What are you trying to achieve?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -201,11 +205,70 @@ private Object instantiateAndPostProcessTestInstance(JupiterEngineExecutionConte
...
+			ExtensionRegistry parentRegistry, ExtensionContext extensionContext) {
+
+		TestInstanceFactory factory = resolveTestInstanceFactory(registry, parentRegistry);
+		if (factory == null) {
+			return invokeTestInstanceConstructor(outerInstance, registry, extensionContext);
+		}
+
+		try {
+			if (outerInstance != null) {
+				return factory.instantiateNestedTestClass(this.testClass, outerInstance, extensionContext);
```
sbrannen: This approach makes it impossible for such a constructor to accept arguments provided by a `ParameterResolver` -- for example, `TestInfo`, `TestReporter`, mocks from Mockito, beans from Spring, etc.

We need to decide if we want this new extension API to suffer from this restriction.

documentation/src/docs/asciidoc/user-guide/extensions.adoc
kdubb: 👍

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
kdubb: I have clarified the behavior in both the javadoc and user guide.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
kdubb: I misunderstood what @sormuras had suggested and exactly wash `rethrowIfBlacklisted` was to be used for. I have just removed this handling entirely to let the exception propagate as normal.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
kdubb: I do not believe this is a restriction that impacts the majority of use cases for factories. I believe factories are specifically for when customized construction is required by external libraries; in which case arguments from JUnit 5 sources are probably not going to be required. Also, the current extension API seems to cover these situations well and using a factory doesn't negate post-processing or any other extension.

kdubb:
@sbrannen I believe the changes you've requested have been pushed. Let me know if anything is still outstanding.

documentation/src/docs/asciidoc/link-attributes.adoc
```diff
@@ -65,6 +65,7 @@ endif::[]
 :RepetitionInfo:                    {javadoc-root}/org/junit/jupiter/api/RepetitionInfo.html[RepetitionInfo]
 :TestExecutionExceptionHandler:     {javadoc-root}/org/junit/jupiter/api/extension/TestExecutionExceptionHandler.html[TestExecutionExceptionHandler]
 :TestInfo:                          {javadoc-root}/org/junit/jupiter/api/TestInfo.html[TestInfo]
+:TestInstanceFactory:               {javadoc-root}/org/junit/jupiter/api/extension/TestInstanceFactory.html[TestInstanceFactoryh]
```
sbrannen: TestInstanceFactoryh --> TestInstanceFactory

sbrannen: TestInstanceFactoryh --> TestInstanceFactory

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
sbrannen: ok. thanks

documentation/src/docs/asciidoc/link-attributes.adoc
sbrannen: Remove trailing "h" in link name.

TestInstanceFactoryh --> TestInstanceFactory

sbrannen:
My comments:

> This approach makes it impossible for such a constructor to accept arguments provided by a `ParameterResolver` -- for example, `TestInfo`, `TestReporter`, mocks from Mockito, beans from Spring, etc.
> 
> We need to decide if we want this new extension API to suffer from this restriction.

Response from @kdubb:

> I do not believe this is a restriction that impacts the majority of use cases for factories. I believe factories are specifically for when customized construction is required by external libraries; in which case arguments from JUnit 5 sources are probably not going to be required. Also, the current extension API seems to cover these situations well and using a factory doesn't negate post-processing or any other extension.

@junit-team/junit-lambda, the team needs to reach a consensus on this before continuing with this PR.

sbrannen:
@kdubb,

> I believe the changes you've requested have been pushed. 

Thanks

> Let me know if anything is still outstanding.

I noticed an extra "h", but I think that's all for now.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstanceFactory.java
```diff
@@ -0,0 +1,66 @@
...
+	 * only attempt to create the required test instance.
+	 *
+	 * @param testClass the test class to instantiate or otherwise obtain
+	 * @param outerInstance the outer instance when requesting a {@link Nested} test class;
+	 * will be {@link Optional#empty} when {@code testClass} is not nested.
+	 * @param context the current extension context; never {@code null}
+	 * @return the required test instance; never {@code null}
+	 * @throws TestInstantiationException when an error occurs with the invocation of a factory
+	 */
+	Object instantiateTestClass(Class<?> testClass, Optional<Object> outerInstance, ExtensionContext context)
```
marcphilipp: Since the team already foresees that we might eventually need to add a callback into the existing parameter resolution mechanism, I think it would be better to use a parameter object, e.g. `TestInstanceFactoryContext` that groups `testClass` and `outerInstance`. This would allow us to later add a method to the context etc. (see `ParameterResolver`/`ParameterContext` for an example of this approach).

sbrannen: 👍

kdubb:
@marcphilipp @sbrannen Not sure if you guys say but I committed the changes @marcphilipp requested.  Is there anything else that needs to happen? Would have love to have this in M1 if possible.

kdubb:
@marcphilipp Test for PER_CLASS was added and rebased onto latest master

sbrannen:
Thanks for adding the additional tests, @kdubb.

sbrannen:
This has been merged into `master` in 5b4c642be9b1d09f9b9cebad7dd55fa40aae78bc and polished in subsequent commits.

Issue #1447, opened 2018-06-05
Ensure discovered test descriptor graphs are acyclic


sormuras:
## Overview

At the moment, the `DefaultLauncher` only ensures that an engine root descriptor is non-null. If an engine implementation returns a cyclic test descriptor graph, that graph is passed without any checking to internal filtering and then back to the engine for execution. This may lead to stack overflow issues when traversing the graph without checking for cycles.

For example here:

https://github.com/junit-team/junit5/blob/19a4cde09f1e36f4e270c537f8cf9a474e1746a8/junit-platform-surefire-provider/src/main/java/org/junit/platform/surefire/provider/RunListenerAdapter.java#L116

## Deliverables

- [ ] Detect cycles in test descriptor graphs

PR #1451, opened 2018-06-05
Ensure discovered descriptor graphs are acyclic

sormuras:
## Overview

Prior this commit only a null-check was performed by the `DefaultLauncher` against the root test descriptor returned by an engine. Other properties where not checked, especially if the graph of that root descriptor contained cycles it could lead to stack overflow errors when traversing
the graph for execution.

This commit addresses this issue by explicitly checking the returned graph of the root test descriptor for cycles. If a cycle is detected a `PreconditionViolationException` is thrown.

Fixes: #1447

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
@@ -44,6 +44,7 @@ class DefaultLauncher implements Launcher {
 	private static final Logger logger = LoggerFactory.getLogger(DefaultLauncher.class);
 
 	private final TestExecutionListenerRegistry listenerRegistry = new TestExecutionListenerRegistry();
+	private final EngineDiscoveryResultValidator discoveryResultValidator = new EngineDiscoveryResultValidator();
 	private final Iterable<TestEngine> testEngines;
 
 	/**
@@ -128,10 +129,7 @@ class DefaultLauncher implements Launcher {
 		UniqueId uniqueEngineId = UniqueId.forEngine(testEngine.getId());
 		try {
 			TestDescriptor engineRoot = testEngine.discover(discoveryRequest, uniqueEngineId);
-			Preconditions.notNull(engineRoot,
-				() -> String.format(
-					"The discover() method for TestEngine with ID '%s' must return a non-null root TestDescriptor.",
-					testEngine.getId()));
+			discoveryResultValidator.validate(testEngine, engineRoot);
 			return Optional.of(engineRoot);
 		}
 		catch (Throwable throwable) {
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidator.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidator.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.launcher.core;
+
+import java.util.ArrayDeque;
+import java.util.HashSet;
+import java.util.Queue;
+import java.util.Set;
+
+import org.junit.platform.commons.util.Preconditions;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.TestEngine;
+import org.junit.platform.engine.UniqueId;
+
+/**
+ * Perform common validation checks on the result from the `discover()` method.
+ *
+ * @since 1.3
+ */
+class EngineDiscoveryResultValidator {
+
+	/**
+	 *  Perform common validation checks.
+	 *
+	 * @throws org.junit.platform.commons.util.PreconditionViolationException if any check fails
+	 */
+	void validate(TestEngine testEngine, TestDescriptor root) {
+		Preconditions.notNull(root,
+			() -> String.format(
+				"The discover() method for TestEngine with ID '%s' must return a non-null root TestDescriptor.",
+				testEngine.getId()));
+		Preconditions.condition(isAcyclic(root),
+			() -> String.format("The discover() method for TestEngine with ID '%s' returned a cyclic graph.",
+				testEngine.getId()));
+	}
+
+	/**
+	 * @return {@code true} if the tree does <em>not</em> contain a cycle; else {@code false}.
+	 */
+	boolean isAcyclic(TestDescriptor root) {
+		Set<UniqueId> visited = new HashSet<>();
+		visited.add(root.getUniqueId());
+		Queue<TestDescriptor> queue = new ArrayDeque<>();
+		queue.add(root);
+		while (!queue.isEmpty()) {
+			for (TestDescriptor child : queue.remove().getChildren()) {
+				if (!visited.add(child.getUniqueId())) {
+					return false; // id already known: cycle detected!
+				}
+				if (child.isContainer()) {
+					queue.add(child);
+				}
+			}
+		}
+		return true;
+	}
+
+}
--- a/platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/AbstractTestDescriptorTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/AbstractTestDescriptorTests.java
@@ -34,7 +34,7 @@ import org.junit.platform.engine.UniqueId;
  */
 class AbstractTestDescriptorTests {
 
-	EngineDescriptor engineDescriptor;
+	private EngineDescriptor engineDescriptor;
 
 	@BeforeEach
 	void initTree() {
--- a/platform-tests/src/test/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidatorTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidatorTests.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.launcher.core;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.UniqueId;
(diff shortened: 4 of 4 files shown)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -69,6 +70,27 @@
 		return testEngines;
 	}
 
+	/**
+	 * @return {@code true} if the graph does _not_ contain a cycle; else {@code false}.
+	 */
+	static boolean isAcyclic(TestDescriptor root) {
+		Set<UniqueId> visited = new HashSet<>();
+		visited.add(root.getUniqueId());
+		Stack<TestDescriptor> stack = new Stack<>();
```
jbduncan: Minor nit: Could be `Deque<TestDescriptor> stack = new ArrayDeque<>` to increase performance slightly. :)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -69,6 +70,27 @@
 		return testEngines;
 	}
 
+	/**
+	 * @return {@code true} if the graph does _not_ contain a cycle; else {@code false}.
```
jbduncan: I wonder if the `_not_` should be replaced with `<em>not</em>`?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -69,6 +70,27 @@
...
+				if (!visited.add(child.getUniqueId())) {
+					return false; // id already known: cycle detected!
+				}
+				if (child.isContainer()) {
+					stack.add(child);
+				}
+			}
+		}
+		return true;
+	}
```
jbduncan: @sormuras I'm not clear on what algorithm `isAcyclic()` uses to check for cycles. Can you place a comment towards the top of the body so that it's clear, so something like the following?

```java
static boolean isAcyclic() {
    // <name-of-algorithm-here>
    . . .
}
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
sormuras: Any link to a proof? Are we talking 10%, 1% or 0.001% here?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
sormuras: Good point.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
sormuras: Hm, never invented a name for some lines of code before. _stack-based-traversing-with-bread-crumb-checks_?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
jbduncan: > Any link to a proof?

Yep! In the Java 8 API docs for [`ArrayDeque`](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayDeque.html), the class javadocs say,

> This class is likely to be faster than Stack when used as a stack, and faster than LinkedList when used as a queue.

> Are we talking 10%, 1% or 0.001% here?

I don't know, actually! And I don't actually know how to use JMH or another benchmarking tool to show _how_ much... My bad. :P

sormuras: hehe, okay. Taking the _likely_ as solid hint. (-:

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
jbduncan: Haha!

I was hoping that your algorithm had an official name like _Kahn's algorithm_ for topological sorting or _Tarjan's algorithm_ for finding strongly-connected components, as I understand from these [StackOverflow](https://stackoverflow.com/a/261595/2252930) [answers](https://stackoverflow.com/a/261621/2252930) that those algorithms can be used to find cycles in graphs.

If alternatively your algorithm is home-grown and it was inspired by an existing graph traversal algorithms like _depth-first search_ or _breadth-first search_, then it would help clarify things (for me, at least) if that was noted down somewhere, along with what change(s) were made to the traversal algorithm to allow cycles to be detected.

But maybe I'm asking for _way_ too much here, in which case I'd be happy to slowly read through the code in my own time later today to figure out what it's doing. :)

sbrannen: Well, we don't actually have a _directed graph_. Rather, we have a _tree_.

So we should probably change the title of the issue and PR to avoid confusion. 😉

sbrannen: And that's a breadth-first traversal of the tree of test descriptors.

sormuras: > Well, we don't actually have a directed graph. Rather, we have a tree.

A _tree_ is still a _graph_.

> So we should probably change the title of the issue and PR to avoid confusion.

 No where "directed" is mentioned. ;)

sbrannen: Sure, but I'd still refer to it as a _tree_ in this context. 😇

jbduncan: > And that's a breadth-first traversal of the tree of test descriptors.

@sbrannen Cool, thank you for the clarification/confirmation! 😀👍

jbduncan: Ah cool, rereading the code with the knowledge that it uses breadth-first search, I understand what it's doing now: it's using a `stack` and continuous calls to `stack.pop().getChildren()` to iterate over the entire tree in breath-first order, and it keeps track of `visited` nodes to ensure there are no cycles. Got it! 👍

I'd personally love it if a minor comment were placed at the top of the body like `// based on breath-first search`, but I won't insist. 🙂

sbrannen: Well, if @sormuras describes it as "breath-first [sic]", I would find that totally amusing...

Take a deep _breath_ first... and then dive deep! 🤣

jbduncan: Haha! *facepalm*

This is what I get for typing things with my phone. :P

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -69,6 +71,27 @@
...
+		visited.add(root.getUniqueId());
+		Deque<TestDescriptor> stack = new ArrayDeque<>();
+		stack.push(root);
+		while (!stack.isEmpty()) {
+			for (TestDescriptor child : stack.pop().getChildren()) {
+				if (!visited.add(child.getUniqueId())) {
+					return false; // id already known: cycle detected!
+				}
+				if (child.isContainer()) {
+					stack.add(child);
```
marcphilipp: Using `pop()` and `add()` will reverse the iteration order. Shouldn't hurt, but might make debugging harder.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -132,6 +155,9 @@ private Root discoverRoot(LauncherDiscoveryRequest discoveryRequest, String phas
 				() -> String.format(
 					"The discover() method for TestEngine with ID '%s' must return a non-null root TestDescriptor.",
 					testEngine.getId()));
+			Preconditions.condition(isAcyclic(engineRoot),
+				() -> String.format("The discover() method for TestEngine with ID '%s' returned a cyclic graph.",
+					testEngine.getId()));
```
marcphilipp: I think we should extract these checks into a separate `EngineDiscoveryResultValidator` class or sth. similar.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
jbduncan: @marcphillip Your saying that has prompted me to think more, and I think you're almost right; AFAICT, it will actually cause the tree to be traversed in _depth-first order_ rather than reversing the iteration order. As you said, it wouldn't hurt (except for debugging), but I think swapping the stack out for a `Queue<> queue = new ArrayDeque<>()` should fix things.

marcphilipp: You‘re right, it‘s indeed depth-first. However since we‘re visiting all descriptors anyway it doesn’t really matter. 😉

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
gaganis: Since you were discussing Stack vs Array, let me add my one cent. Stack is actually a subclass of Vector which is synchronized and in essence is part of the _'legacy'_ classes  preexisting the java 2 introduced collections framework. In [Vector's javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/Vector.html) we can read:

> As of the Java 2 platform v1.2, this class was retrofitted to implement the List interface, making it a member of the Java Collections Framework. Unlike the new collection implementations, Vector is synchronized. If a thread-safe implementation is not needed, it is recommended to use ArrayList in place of Vector.

Also in [Stack's javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/Stack.html) we can see another hint that Stack is legacy, reading:
> A more complete and consistent set of LIFO stack operations is provided by the Deque interface and its implementations, which should be used in preference to this class. 

On the other hand when we want a `stack` it is natural to use `Stack`... why would someone want to use a `Deque`..? It makes no sense... :flushed:

jbduncan: @gaganis I know, it's very weird and the complete opposite of intuitive. This is one of the many drawbacks of Java's policy on maintaining backwards compatibility as much as possible.

I don't know why they originally implemented Stack by extending Vector all those years ago, or why they made it a concrete class instead of an interface which would have allowed alternative stack implementations. However, as I think you've noticed, due to those historical decisions and how Stack consequently uses `synchronized` internally, Stack does perform not as well as ArrayDeque as a stack in single-thread situations.

As for why we're only provided with a `Deque` interface and not also a super-interface that provides just `push()` and `pop()`, I've no idea as to why it was done that way. Seems a bit silly to me, personally.

...Not sure what else to say. :)

PR #1461, opened 2018-06-08
Add support for parallel test execution

marcphilipp:
This PR adds support for parallel test execution (#60). Please refer to the added user guide sections for details.

The majority of this work was done by @leonard84 and me over the course of the last few months. I intend to squash the commits before merging this PR. For reviewing, I've decided to keep them in place so it's easier to see why a change was made.

If you want to see it in action, just execute `SlowTests` in an IDE (tested in IntelliJ IDEA).

---

---

### Definition of Done


documentation/src/docs/asciidoc/user-guide/running-tests.adoc
```diff
@@ -755,3 +755,28 @@ useful.
 | +(micro \| integration) & (foo \| baz)+
 | all _micro_ or _integration_ tests for *foo* or *baz*
 |===
+
+[[running-tests-capturing-output]]
+=== Capturing Standard Output/Error
+
+Since version 1.3, the JUnit Platform provides opt-in support for capturing output
+printed to `System.out` and `System.err`. To enable it, simply set the
+`junit.platform.launcher.capture.stdout` and/or `junit.platform.launcher.capture.stderr`
```
sbrannen: I think the prefix should be `junit.platform.launcher.output.capture` or `junit.platform.output.capture`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DynamicNodeTestDescriptor.java
```diff
@@ -30,6 +32,11 @@
 		this.index = index;
 	}
 
+	@Override
+	public ExecutionMode getExecutionMode() {
+		return getParent().map(parent -> ((JupiterTestDescriptor) parent).getExecutionMode()).orElse(CONCURRENT);
```
sbrannen: Will Jupiter now default to concurrent execution?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -103,6 +110,46 @@
...
+	protected ExecutionMode getExecutionMode(AnnotatedElement element) {
+		// @formatter:off
+		return findAnnotation(element, Execution.class)
+				.map(Execution::value)
+				.map(JupiterTestDescriptor::toExecutionMode)
+				.orElseGet(() -> getParent()
+						.filter(parent -> parent instanceof Node)
+						.map(parent -> ((Node<?>) parent).getExecutionMode())
+						.orElse(ExecutionMode.CONCURRENT));
+		// @formatter:on
```
sbrannen: Will Jupiter now default to concurrent execution?

junit-platform-engine/src/main/java/org/junit/platform/engine/ConfigurationParameters.java
```diff
@@ -83,6 +85,44 @@
...
+	 * If no such system property exists, an attempt will be made to look up
+	 * the value in the {@linkplain #CONFIG_FILE_NAME JUnit Platform properties
+	 * file}.
+	 *
+	 * <p>In case the transformer throws an exception, it will be wrapped in a
+	 * {@link JUnitException} with a helpful message.
+	 *
+	 * @param key the key to look up; never {@code null} or blank
+	 * @param transformer the transformer to apply in case a value is found;
+	 * never {@code null} or blank
```
sbrannen: remove "or blank" for the transformer

junit-platform-engine/src/main/java/org/junit/platform/engine/ConfigurationParameters.java
```diff
@@ -83,6 +85,44 @@
...
+	 * @since 1.3
+	 */
+	@API(status = STABLE, since = "1.3")
+	default <T> Optional<T> get(String key, Function<String, T> transformer) {
+		return get(key).map(input -> {
+			try {
+				return transformer.apply(input);
+			}
+			catch (Exception ex) {
+				String message = String.format("Failed to convert configuration parameter with key '%s' for input: %s",
```
sbrannen: --> Failed to transform configuration parameter with key '%s' and initial value '%s'

junit-platform-engine/src/main/java/org/junit/platform/engine/support/config/PrefixedConfigurationParameters.java
```diff
@@ -0,0 +1,65 @@
...
+
+	private final ConfigurationParameters delegate;
+	private final String prefix;
+
+	/**
+	 * Create a new view of the supplied {@link ConfigurationParameters} that
+	 * applies the supplied prefix to all queries.
+	 */
+	public PrefixedConfigurationParameters(ConfigurationParameters delegate, String prefix) {
+		this.delegate = delegate;
```
sbrannen: Needs precondition checks

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/UseResource.java
```diff
@@ -0,0 +1,64 @@
...
+ * @see Resources
+ * @see ResourceAccessMode
+ * @see UseResources
+ * @since 5.3
+ */
+@API(status = EXPERIMENTAL, since = "5.3")
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Repeatable(UseResources.class)
+public @interface UseResource {
```
sbrannen: Wouldn't `@ResourceLock` be a better suited name?

documentation/src/docs/asciidoc/user-guide/running-tests.adoc
marcphilipp: Will do!

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
marcphilipp: Only if the `junit.jupiter.execution.parallel.enabled` config param is set tot `true`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/UseResource.java
marcphilipp: Possibly... let's discuss it in the next team call, shall we?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/config/PrefixedConfigurationParameters.java
marcphilipp: ✅

junit-platform-engine/src/main/java/org/junit/platform/engine/ConfigurationParameters.java
marcphilipp: ✅

junit-platform-engine/src/main/java/org/junit/platform/engine/ConfigurationParameters.java
marcphilipp: ✅

marcphilipp:
I just pushed two additional commits that address most of the feedback. Only remaining issue is the name of `@UseResource`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/UseResource.java
leonard84: Isn't the fact that it is currently implemented via a lock an implementation detail? The idea of `@UseResource` was so that the tests could declare what resources they use and what guarantees they expect/observe. If we change the implementation to use something else than a Lock then `@ResourceLock` wouldn't make sense.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
sbrannen: OK. In that code it looks like it's defaulting to concurrent, so I'll have to review in greater depth to find where the flag is honored.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/UseResource.java
sbrannen: I think this is one of those scenarios where an implementation detail leaks to higher levels of the overall abstraction simply because it's part of the high-level use case.

To me, `@UseResource` is too generic in that it does not immediately provide a mental link to the concurrency model in question. After all, we're talking about executing things in parallel. So there is going to have to be some sort of _locking_ in order to avoid contention -- right?

On the other hand, I understand your desire not to leak implementation details, but I'm still not fond of `@UseResource`. 😉

sbrannen: In terms of consistency, IIRC, the only places we use "verbs" for annotation names are for `@ExtendWith` and `@RegisterExtension`.

Otherwise, we try to avoid using verbs for annotation names.

sbrannen: > Possibly... let's discuss it in the next team call, shall we?

If I'm there... gladly.

sbrannen:
I approved the changes requested thus far and agree that we still need to decide on the name of `@UseResource`.

If I have time, I'll review the internals in greater detail.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/UseResource.java
sbrannen: If people don't like `@ResourceLock`, I think that `@SharedResource` would at least be better than `@UseResource`.

Rationale: the developer is not instructing the framework to _use_ a resource for the test. Rather, the developer is instructing the framework that the test itself uses a shared resource, to which the framework should then synchronize access.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
marcphilipp: Just have a look at `JupiterTestEngine`. 😉

sbrannen: Alrighty. Will do...

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/UseResource.java
gaganis: I have to agree with @sbrannen that `@UseResource` is not a good name for this. `@ResourceLock` would make more sense to me.

Adding the Lock concept does help me to understand the underlying mental(concurrency) model. 

When seeing for the first time `@UseResource` for some reason I imagined that this annotation would be added to some `resource` ie some critical piece of code eg a method, that would be shared and the framework would orchestrate tests around it. _[In retrospect, feeling a bit stupid now for thinking that.]_

After reviewing the code I think the important point of this mechanism is the exclusivity of execution of tests, ie these tests characterized as _readers_ can execute at the same time while tests characterized as _writers_ can only run by themselves.

The concept of a lock suggests the above quite well for me.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,154 @@
...
+		try {
+			// Try to use constructor available in Java >= 9
+			Constructor<ForkJoinPool> constructor = ForkJoinPool.class.getDeclaredConstructor(Integer.TYPE,
+				ForkJoinWorkerThreadFactory.class, UncaughtExceptionHandler.class, Boolean.TYPE, Integer.TYPE,
+				Integer.TYPE, Integer.TYPE, Predicate.class, Long.TYPE, TimeUnit.class);
+			return constructor.newInstance(configuration.getParallelism(), defaultForkJoinWorkerThreadFactory, null,
+				false, configuration.getCorePoolSize(), configuration.getMaxPoolSize(),
+				configuration.getMinimumRunnable(), null, configuration.getKeepAlive(), TimeUnit.SECONDS);
+		}
+		catch (Exception e) {
```
gaganis: Wouldn't catching the all inclusive Exception here also mask actually crashes in the try part of this code, eg NPEs?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java
```diff
@@ -94,4 +109,12 @@ public void reportingEntryPublished(TestIdentifier testIdentifier, ReportEntry e
 
 	}
 
+	interface EagerTestExecutionListener extends TestExecutionListener {
```
leonard84: What are eager EagerTestExecutionListener? Shouldn't we add ordering instead?

platform-tests/src/test/java/org/junit/platform/console/tasks/XmlReportAssertions.java
```diff
@@ -26,24 +28,27 @@
  */
 class XmlReportAssertions {
 
-	private static Validator schemaValidator;
+	private static AtomicReference<Schema> schema = new AtomicReference<>();
```
leonard84: Do you really need an `AtomicReference` why not just use `getSchema` as initializer?

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/CompositeLockTests.java
```diff
@@ -0,0 +1,95 @@
...
+		ReentrantLock lock2 = new ReentrantLock();
+
+		new CompositeLock(asList(lock1, lock2)).acquire();
+
+		assertTrue(lock1.isLocked());
+		assertTrue(lock2.isLocked());
+	}
+
+	@Test
+	void releasesAllLocksInReverseOrder() throws Exception {
```
leonard84: A test for acquire inOrder is missing.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/LockManagerTests.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+class LockManagerTests {
+
+	private LockManager lockManager = new LockManager();
+
+	@Test
+	void returnsNopLockWithoutExclusiveResources() {
+		Collection<ExclusiveResource> resources = emptySet();
+
+		List<Lock> locks = getLocks(lockManager, resources);
```
leonard84: This does not assert that a `NopLock` was returned, same goes for the other tests.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/LockManagerTests.java
```diff
@@ -0,0 +1,106 @@
...
+		List<Lock> locks1 = getLocks(lockManager, resources);
+		List<Lock> locks2 = getLocks(lockManager, resources);
+
+		assertThat(locks1).hasSize(1);
+		assertThat(locks2).hasSize(1);
+		assertThat(locks1.get(0)).isSameAs(locks2.get(0));
+	}
+
+	@Test
+	void returnsSingleLockForExclusiveResourcWithBothLockModes() {
```
leonard84: `returnsSingleLockForExclusiveResourcWithBothLockModes` + `UsingStrongestLockMode`

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
marcphilipp: All `ParallelExecutionConfiguration` methods return primitive types thus I don't see where a NPE should come from. Am I missing something?

junit-platform-commons/src/main/java/org/junit/platform/commons/logging/LogRecordListener.java
```diff
@@ -30,7 +30,7 @@
 @API(status = INTERNAL, since = "1.1")
 public class LogRecordListener {
 
-	private final List<LogRecord> logRecords = new ArrayList<>();
+	private final List<LogRecord> logRecords = new CopyOnWriteArrayList<>();
```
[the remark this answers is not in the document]
marcphilipp: Changed in 650fa52f2e12c11619dfae698732d8edbdb54e26.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,154 @@
...
+				concurrentTasksInReverseOrder.addFirst(exclusiveTask);
+			}
+			else {
+				nonConcurrentTasks.add(exclusiveTask);
+			}
+		}
+		for (ExclusiveTask task : nonConcurrentTasks) {
+			task.compute();
+		}
+		for (ExclusiveTask forkedTask : concurrentTasksInReverseOrder) {
```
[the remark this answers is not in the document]
marcphilipp: I extracted three methods in 4bf6044e5bc70755be086a16c5f2abcdca5c859d instead. Please let me know if that's easier to understand now.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java
marcphilipp: For now this is a one-off internal scenario. I found this easier to understand and less error-prone than adding ordering support.

platform-tests/src/test/java/org/junit/platform/console/tasks/XmlReportAssertions.java
marcphilipp: It's surprisingly hard to implement a thread-safe singleton without locking. If you have a better solution, I'm open to change it. 😉

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/LockManagerTests.java
marcphilipp: Improved in 4b40b64719148d7b92224b7da53d5d5f3d11510a.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/LockManagerTests.java
marcphilipp: Renamed to `returnsWriteLockForExclusiveResourceWithBothLockModes` in 4b40b64719148d7b92224b7da53d5d5f3d11510a.

platform-tests/src/test/java/org/junit/platform/console/tasks/XmlReportAssertions.java
jbduncan: Did someone say _thread-safe singleton_? :)

Would Effective Java's recommended solution - a single-instance enum - work in this case? E.g.

```java
enum ClassName extends AppropriateInterface {
    INSTANCE;

   // needed methods go here
}
```

marcphilipp: I had tried that at some point but it felt clunky. Revisiting it now, it's not actually that bad so I changed it in 135a07d7ce28de1605bb69fbd2e24e9cdf8fc689.

@jbduncan Thanks! 🙂

jbduncan: @marcphilipp You're very welcome! :smiley:

marcphilipp:
**Team Decision:** Use `@ResourceLock` for M1 and ask the community for feedback.

sormuras:
Wohoo!

sbrannen:
👏

PR #1470, opened 2018-06-18
Rename file to match class name

piefel:
## Overview

There’s a bug in the source: The file 
`junit-platform-console/src/main/java/org/junit/platform/console/options/URIConverter.java`
contains the class `org.junit.platform.console.options.UriConverter` –

---

---

### Definition of Done


marcphilipp:
Good catch, thanks! 👍

jbduncan:
I remember submitting a PR for a similar fix with a similarly-named file a few months ago, so indeed, good catch @piefel! :+1:

piefel:
It seems everbody on the team is using MacOS these days. (Or even Windows.)

marcphilipp:
Even though it’s (at least in this case) not technically a bug because the class was not public, it would be useful to have an automated check in place. Any ideas? Does Checkstyle support that?

sbrannen:
> Even though it’s (at least in this case) not technically a bug because the class was not public, 

Right: definitely not a "bug" but potentially misleading nonetheless.

> it would be useful to have an automated check in place. 

I agree it would be useful to have some sort of automated check. Though I guess it would only make sense if the Java source file contains a single, non-public type.

> Any ideas? Does Checkstyle support that?

Dunno... 🤷🏻‍♂️

piefel:
I don’t know about your ‘bug’ definition, but JUnit 5 refused to be compiled on a Linux machine using an Oracle JDK, and that looks pretty much like a bug to me.

sbrannen:
JUnit 5 has CI builds on Linux and Windows (and macOS as well I believe).

- https://junit.ci.cloudbees.com/job/JUnit5/job/master/
- https://travis-ci.org/junit-team/junit5
- https://ci.appveyor.com/project/marcphilipp/junit5/branch/master

The last time I checked it's not a requirement of a Java source file to have a name that matches a non-public type declared in that file as long as the file does not declare any top-level public type.

Thus it would appear that it is a _bug_ with the compiler you used.

Did you compile from the command line with the Gradle build?

Or did you compile in an IDE, and if so which one?

sbrannen:
> Thus it would appear that it is a _bug_ with the compiler you used.

OK. I retract that. 😉 

It turns out that compilers are not required to enforce that a type name match the file name. Rather, it is an "option" in certain circumstances.

Specifically, Section  ["7.6. Top Level Type Declarations"](https://docs.oracle.com/javase/specs/jls/se8/html/jls-7.html#jls-7.6) of the JLS states the following.

> If and only if packages are stored in a file system (§7.2), the host system may choose to enforce the restriction that it is a compile-time error if a type is not found in a file under a name composed of the type name plus an extension (such as .java or .jav) if either of the following is true:
> 
> - The type is referred to by code in other compilation units of the package in which the type is declared.
> - The type is declared public (and therefore is potentially accessible from code in other packages).

The compilers used by all CI builds, core committers, and contributors thus far have never enforced the optional first condition mentioned above.

Thus, we simply never ran into this situation.

Given that, @piefel, we would be very interested in knowing exactly how you encountered this situation (i.e., Gradle build, IDE, JDK version, OS version, etc.).

sbrannen:
FYI: I added the 🐛 label.

piefel:
OK, there you got me. I copied over parts of the repo in an attempt to use the TreePrinter as an execution listener, and compiled my project using Maven and Java 8. I didn’t use the full JUnit build. Using the command line directly works and does _not_ enforce the file name. I wonder wether the maven-compiler-plugin tightens the configuration.

BTW, the build does not work anymore with Java 9, or at least 9.0.4: There is a warning about a missing `serialVersionUID` in `UriSource` (which is an interface, BTW). Well, Java 9 is outdated anyway, but `settings.gradle` asks for Java 9 or higher.

sbrannen:
> I wonder wether the maven-compiler-plugin tightens the configuration.

Perhaps...

sbrannen:
> BTW, the build does not work anymore with Java 9, or at least 9.0.4: There is a warning about a missing `serialVersionUID` in `UriSource` (which is an interface, BTW). Well, Java 9 is outdated anyway, but `settings.gradle` asks for Java 9 or higher.

Thanks for mentioning that.

@sormuras actually just ran into that issue as well and added a pseudo-fix in commit ae40f6e123ebd4cb3a2ad7020a427b4e9bd120c5 for use with JitPack.

We think it must be a bug in the compiler, but we have not yet investigated that further.

In any case, I think I'll update `settings.gradle` to enforce JDK 10 for all builds now as a result.

sbrannen:
Regarding the compiler warning for `UriSource`, it would appear that it is due to the fact that we recently introduced a `static` factory method in `UriSource`, and that would appear to be related to the following JDK bug:

https://bugs.openjdk.java.net/browse/JDK-8191637

That bug was fixed in JDK 10.0.1 and 11 EA.

sbrannen:
Further discussion on that "`Serializable` for interfaces" bug: https://stackoverflow.com/questions/47390432/functional-interface-with-default-behavior-with-serializable#comment83187695_47390432

sbrannen:
FYI: the Gradle build now requires JDK 10 or higher: 545e4d9f7ddb1722a90542eb506b3991b3a9ae14

sbrannen:
FWIW, the `README` had already documented the JDK 10+ requirement: https://github.com/junit-team/junit5#building-from-source

sormuras:
> FWIW, the README had already documented the JDK 10+ requirement: https://github.com/junit-team/junit5#building-from-source

Aye. GA 5.2 was already built with 10.0.1 -- only left the option to build with 9 for our Eclipse users. :-)

Issue #1524, opened 2018-08-05
Inconsistent output from TreePrinter in ConsoleLauncher with parallel test execution


littleclay:
When running all the tests in a few simple test classes with the 1.3.0-M1 ConsoleLauncher standalone jar with the parallel test execution, I am seeing inconsistent output. Sometimes there will be a NullPointerException in the TreePrinter, sometimes it will execute but some of the executed tests will not show up in the tree output (though they are in the final summary), sometimes it will display the ConsoleLauncher help output, and sometimes there is a mix of all of those behaviors. One example is:

```
(aaron) $ java -jar junit-platform-console-standalone-1.3.0-M1.jar --cp tests.jar -p com.foo --config junit.jupiter.execution.parallel.enabled=true --details tree

java.lang.NullPointerException
	at org.junit.platform.console.tasks.TreePrinter.print(TreePrinter.java:54)
	at org.junit.platform.console.tasks.TreePrinter.print(TreePrinter.java:65)
	at org.junit.platform.console.tasks.TreePrinter.print(TreePrinter.java:65)
	at org.junit.platform.console.tasks.TreePrinter.print(TreePrinter.java:65)
	at org.junit.platform.console.tasks.TreePrinter.print(TreePrinter.java:49)
	at org.junit.platform.console.tasks.TreePrintingListener.testPlanExecutionFinished(TreePrintingListener.java:56)
	at org.junit.platform.launcher.core.TestExecutionListenerRegistry$CompositeTestExecutionListener.lambda$testPlanExecutionFinished$7(TestExecutionListenerRegistry.java:102)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at org.junit.platform.launcher.core.TestExecutionListenerRegistry.notifyTestExecutionListeners(TestExecutionListenerRegistry.java:59)
	at org.junit.platform.launcher.core.TestExecutionListenerRegistry.access$100(TestExecutionListenerRegistry.java:28)
	at org.junit.platform.launcher.core.TestExecutionListenerRegistry$CompositeTestExecutionListener.testPlanExecutionFinished(TestExecutionListenerRegistry.java:102)
	at org.junit.platform.launcher.core.DefaultLauncher.lambda$execute$5(DefaultLauncher.java:155)
	at org.junit.platform.launcher.core.DefaultLauncher.withInterceptedStreams(DefaultLauncher.java:166)
[...]

PR #1528, opened 2018-08-07
Fixing concurrent access issue in TreePrintingListener

littleclay:
The addNode operation in the TreePrintingListener was non-atomic and
subject, therefore, to a race condition. This change makes the methods
that access the map synchronized to remove that race condition. Since
the access of the map is locked on the "this" object by the synchronized
keyword, the use of the ConcurrentHashMap is no longer needed and is
therefore an extra set of locking that is unnecessary.

Issue: #1524

## Overview

---

---

### Definition of Done

### DIFF
--- a/junit-platform-console/src/main/java/org/junit/platform/console/tasks/TreePrintingListener.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/tasks/TreePrintingListener.java
@@ -27,7 +27,7 @@ import org.junit.platform.launcher.TestPlan;
  */
 class TreePrintingListener implements TestExecutionListener {
 
-	// concurrent access to the map is guarded by the "this" object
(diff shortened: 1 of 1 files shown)


sbrannen:
Slated for 5.3 RC1 and assigned to @sormuras for review.

marcphilipp:
Alternate proposal: #1529

littleclay:
@marcphilipp Thanks for the catch! I was looking at the `nodesByUniqueId.map().orElse()` since that is a two part operation (checking if the parent exists and getting the root if not). But thinking about it now, there will always be a parent since the test plan is added as the root at the testPlanStarted method, which isn't multi threaded, and parents will always be added before their children so they will always exist. So simply locking the children list of the nodes themselves you have to hold it for a shorter period of time.

I'm going to close this pull request since locking the TreeNode add child method itself is a better solution.

Issue #1500, opened 2018-07-19
Document how to use @RegisterExtension for static extensions from Kotlin


stefano-m:
## Overview

- JUnit 5 version: **`5.2.0`**
- Kotlin version: **`1.2.30`**

Hello, I am trying to use the [Programmatic Extension Registration via Static Fields](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic-static-fields) to register an extension in tests written in the Kotlin programming language.  However, I cannot get the `beforeAll` method to be called.  Note that using the `@BeforeAll` annotation works fine.

Note that to have "static" methods in Kotlin, one must use the `@JvmStatic` annotation inside an `object`.

### Example 1: Working `@BeforeAll` annotation:

```kotlin
class TestBeforeAllAnnotation {
    companion object {

        var setMeBeforeAllTests :String? = null

        @JvmStatic
        @BeforeAll
        fun annotatedBeforeAll() {
            setMeBeforeAllTests = "before"
        }
    }

    @Test
    fun testAnnotatedBeforeAll() {
        assertEquals("before", setMeBeforeAllTests) // OK
    }
}
```

### Example 2: Non-working `BeforeAllCallback` extension registration

```kotlin
class BeforeAllCallBackExtension : BeforeAllCallback {

    val atConstructionTime = "done"

    var fromBeforAllCall: String? = null

    override fun beforeAll(context: ExtensionContext?) {
        println("BEFORE ALL CALLED")
        fromBeforAllCall = "before"
    }

}

class TestProgrammaticExtension {
    companion object {
        @JvmStatic
        @RegisterExtension
        val extension = BeforeAllCallBackExtension()
    }

    @Test
    fun filedAtConstructionTimeWorks() {
        assertEquals("done", extension.atConstructionTime) // OK
    }

    @Test
    fun fieldSetInCallBackFails() {
        assertEquals("before", extension.fromBeforAllCall) // FAILS
    }
}
```

sbrannen:
Thanks for raising the issue.

We'll do our best to look into this in time for 5.3.

marcphilipp:
Using `@JvmField` instead of `@JvmStatic` works, otherwise Kotlin generates a private field which we filter out.

@stefano-m Would you be interested in documenting this in the User Guide?

marcphilipp:
I changed the issue title to improve documentation.

stefano-m:
[...]

PR #1534, opened 2018-08-11
Document how to register extensions via static fields in Kotlin

stefano-m:
## Overview

JUnit 5 extensions can be registered programmatically.  When one wants to
register extensions that support lifecycle methods, they should be registered
as static fields.

In the Kotlin programming language this is normally achieved using the
`@JvmStatic` annotation.  However that will not work because a private field
will be generated, and private fields are filtered out by JUnit 5.  The correct
approach is to use `@JvmField` instead.

This commit documents the correct approach and provides a Kotlin example
analogous to the existing Java one.

Reference: #1500

---

---

### Definition of Done


documentation/src/test/java/example/registration/WebServerDemo.kt
```diff
@@ -0,0 +1,70 @@
...
+    // tag::user_guide[]
+    companion object {
+        @JvmField
+        @RegisterExtension
+        val server = WebServerExtension.builder()
+                .enableSecurity(false)
+                .build()
+    }
+    // end::user_guide[]
+    // @formatter:on
```
marcphilipp: I don't think we need the formatter comments here, do we?

documentation/src/test/java/example/registration/WebServerDemo.kt
```diff
@@ -0,0 +1,70 @@
...
+    // tag::user_guide[]
+
+    @Test
+    fun getProductList() {
+        // end::user_guide[]
+        // tag::user_guide[]
```
marcphilipp: The above two lines can be removed, right?

documentation/src/test/java/example/registration/WebServerDemo.kt
```diff
@@ -0,0 +1,70 @@
...
+        // Use WebClient to connect to web server using serverUrl and verify response
+        val webClient = WebClient()
+        val serverUrl = server.serverUrl
+        assertEquals(200, webClient.get("$serverUrl/producs").responseStatus)
+    }
+}
+// end::user_guide[]
+
+class WebServerExtension : BeforeAllCallback {
```
marcphilipp: Can we reuse the Java equivalents of these classes?

sbrannen: 👍

documentation/src/test/java/example/registration/WebServerDemo.kt
sbrannen: Yes. Good catch!

I just fixed that in the "origin": 35732161b6fc20d5825c1c049345f0e0e2eacd9e

documentation/src/test/java/example/registration/WebServerDemo.kt
sbrannen: Yes, we do: for the builder pattern. Otherwise, that ends up being a single line.

It's of course a matter of taste, but this PR just copies the style I chose for the Java example. So the examples are consistent in that regard. 😉

documentation/src/test/java/example/registration/WebServerDemo.kt
stefano-m: Hi,

regarding this and other comments. I merely ported the java extension to kotlin. I thought I could reuse the Java classes, but am not sure how I can achieve that without name clashes (i.e. the `WebServerDemo` class has the same name in both the Java and Kotlin examples).

I am not familiar with how JUnit 5 itself is developed, I noticed that the examples are themselves tested, but don't know whether this Kotlin example is tested too.

Happy to make the changes needed with some guidance.

documentation/src/test/java/example/registration/WebServerDemo.kt
stefano-m: I'll change that here too.

sbrannen: Thanks!

documentation/src/test/java/example/registration/WebServerDemo.kt
sbrannen: > I am not familiar with how JUnit 5 itself is developed, I noticed that the examples are themselves tested, but don't know whether this Kotlin example is tested too.

Yes, the Kotlin demos are also tested in the build. Note the inclusion of `*.Demo` classes below.

https://github.com/junit-team/junit5/blob/35732161b6fc20d5825c1c049345f0e0e2eacd9e/documentation/documentation.gradle#L17-L30

stefano-m: E.g. if I merely remove the `WebServerExtension`, `WebClient` and `WebResponse` classes the `:documentation:compileTestKotlin` task fails with "Unresolved reference" errors.

sbrannen: To avoid the naming conflict, please rename your Kotlin demo class to `WebServerKotlinDemo`.

sbrannen: Otherwise, you _should_ be able to reuse the `WebServerExtension`, `WebClient`, and `WebResponse` Java types. They are all top-level classes in the `example.registration` package.

sbrannen: > Happy to make the changes needed with some guidance.

Is that sufficient guidance? 😉

sbrannen: > if I merely remove the `WebServerExtension`, `WebClient` and `WebResponse` classes the `:documentation:compileTestKotlin` task fails with "Unresolved reference" errors.

I'm not a Kotlin expert, but perhaps you need to move those classes to their own files and make them `public` types.

If that doens't help, perhaps there is an issue regarding the order in which Java and Kotlin test classes are compiled.

stefano-m:
I've just realized that I put `WebServerDemo.kt` in the wrong directory (the `java` one rather than the `koltin` one). Fixing that now. Apologies for that.

documentation/src/test/java/example/registration/WebServerDemo.kt
sbrannen: Actually, now that I think about it, I think we do in fact compile Kotlin classes before Java classes.

So that would explain why the Java types are not visible during the Kotlin test class compilation task execution.

Maybe we need to remove those types from the Java demo, keep them in the Kotlin demo, and then reference the Kotlin types from the Java demo?

If that proves to be too much work, we can also just live with the duplication. After all, they are only dummy examples. 😉

sbrannen:
> I've just realized that I put `WebServerDemo.kt` in the wrong directory (the `java` one rather than the `koltin` one). Fixing that now. Apologies for that.

Oh. Oops!

So, everything is sorted out now with regard to removing the duplicated types?

stefano-m:
> So, everything is sorted out now with regard to removing the duplicated types?

Still not compiling. Maybe a classpath issue?

documentation/src/test/java/example/registration/WebServerDemo.kt
stefano-m: I came to the same conclusion!

I tried to add `compileTestJava` as a dependency to `compileTestKotlin` and gradle failed saying that I had introduced a circular dependency, so I realized that I should keep the classes in Kotlin and remove the Java ones.

I was a bit surprised about this dependency though, I would have expected the inverse.

The file name should not matter as that is a "compilation unit" in Kotlin, but the class should definitely be renamed.

I am going to give a shot at removing the classes from the java file and keeping them in the kotlin one.

stefano-m:
I have solved the compilation issues by moving the common classes from Java to Kotlin, but the kotlin test is not picked up by the `:documentation:consoleLauncherTest`, which is a shame because we cannot prove that it works.

Note that if I move the tests inside the `AssertionsKotlinDemo` **class**, then they run. Simply moving the `WebServerDemoForKotlin` class (I renamed it to avoid clashes with the Java one) in the `AssertionsKotlinDemo.kt` **file** does not work.

stefano-m:
Got it! 

`documentation.gradle` says that only `*.Tests` and `*.Demo` are considered. Renaming the class `KotlinWebServerDemo` fixes it.

stefano-m:
And I just realized I missed the above comment mentioning the importance of the class name.

Anyway, run `./gradlew clean build` passes locally, so I'm going to push the changes shortly.

marcphilipp:
For the record: Classes in `src/main/kotlin` can reference classes in `src/main/java` and vice versa. The only compilation error I got was when I tried to reference the package-private `WebServerExtension` in `src/main/java` from the public `@RegisterExtension` field in Kotlin.

documentation/src/test/java/example/registration/WebServerDemo.kt
marcphilipp: We do for Java, but for Kotlin we currently don't have a formatter that would join those lines. Thus, we should be able to remove them.

documentation/src/test/kotlin/example/registration/WebServerDemo.kt
```diff
@@ -0,0 +1,69 @@
...
+    // end::user_guide[]
+    // @formatter:on
+    // tag::user_guide[]
+
+    @Test
+    fun getProductList() {
+        // Use WebClient to connect to web server using serverUrl and verify response
+        val webClient = WebClient()
+        val serverUrl = server.serverUrl
+        assertEquals(200, webClient.get("$serverUrl/producs").responseStatus)
```
marcphilipp: There's a `t` missing.

documentation/src/test/kotlin/example/registration/WebServerDemo.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.BeforeAllCallback
+import org.junit.jupiter.api.extension.ExtensionContext
+import org.junit.jupiter.api.extension.RegisterExtension
+
+// tag::user_guide[]
+class KotlinWebServerDemo {
+    // end::user_guide[]
+    // @formatter:off
+    // tag::user_guide[]
```
marcphilipp: Please remove the above three lines.

documentation/src/test/kotlin/example/registration/WebServerDemo.kt
```diff
@@ -0,0 +1,69 @@
...
+    companion object {
+        @JvmField
+        @RegisterExtension
+        val server = WebServerExtension.builder()
+                .enableSecurity(false)
+                .build()
+    }
+    // end::user_guide[]
+    // @formatter:on
+    // tag::user_guide[]
```
marcphilipp: Ditto.

documentation/src/test/kotlin/example/registration/WebServerDemo.kt
stefano-m: Well spotted!
Done.

documentation/src/test/kotlin/example/registration/WebServerDemo.kt
stefano-m: Done.

documentation/src/test/kotlin/example/registration/WebServerDemo.kt
stefano-m: Done.

PR #1545, opened 2018-08-14
[Event] Add JUnit 5 event at JDD 2018

szpak:
"Spock vs JUnit 5 - Clash of the Titans" by Marcin Zajączkowski
https://jdd.org.pl/lecture.html#id=47734


marcphilipp:
Thanks, Marcin!

PR #1556, opened 2018-08-20
[WIP] Add ArchUnit-based tests to `platform-tooling-support-tests`

sormuras:
## Overview

This PR adds ArchUnit-based tests to the `platform-tooling-support-tests` project.

_At the moment it only enforces a cycle check that Degraph already applies. Caveat: both tools don't find the "bad edges" that `jdeps` is able to detect._

@codecholeric Am I using Arch-Unit as intended here? What could be improved?

---

---

### DIFF
--- a/platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ArchUnitTests.java
+++ b/platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ArchUnitTests.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package platform.tooling.support.tests;
+
+import static com.tngtech.archunit.library.dependencies.SlicesRuleDefinition.slices;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
+
+import com.tngtech.archunit.core.domain.JavaClasses;
+import com.tngtech.archunit.core.importer.ClassFileImporter;
+
+import org.junit.jupiter.api.Test;
+import platform.tooling.support.Helper;
+
+class ArchUnitTests {
+
+	@Test
+	void acyclicImportPackages() {
+		var packageNames = List.of("org.junit.platform", "org.junit.jupiter", "org.junit.vintage");
+		var classes = new ClassFileImporter().importPackages(packageNames);
+		// about 431 classes found in classpath of this project
+		acyclic(classes);
+	}
+
+	@Test
+	void acyclicImportJars() {
+		var jarFiles = Helper.loadJarFiles();
+		var classes = new ClassFileImporter().importJars(jarFiles);
+		// about 928 classes found in all jars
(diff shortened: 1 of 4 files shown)


codecholeric:
Normally I would say with JUnit 5 you would write your test as

```
@AnalyzeClasses(packges = {"org.junit.platform", "org.junit.jupiter", "org.junit.vintage"})
class ArchUnitTests {
    @ArchTest
    static final ArchRule no_cycles = slices().matching("org.junit.(*)..").should().beFreeOfCycles();

    @ArchTest
    static void sanityCheck(JavaClasses classes) {
        assertTrue(classes.iterator().hasNext()); // not empty...
    } 
}
```

But since you need the second test for Jar files and you really only have one test on the same classes (so caching is no advantage), I would leave the test how it is.

If you should add more rules, you could still ponder about creating different tests for special checks (like the Jars) and put the other tests in a suite like this, where classes are automatically cached (if you only have 900 classes, I guess it doesn't take that long anyway. We have > 50000, and then it does make a noticeable difference :wink:)

sormuras:
Thanks for reviweing.

> @AnalyzeClasses(packages = {"org.junit.platform", "o...

Performs class-path scanning, right? Or is there an option to provide (hard-coded) paths to jars?

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ArchUnitTests.java
```diff
@@ -0,0 +1,47 @@
...
+		// about 928 classes found in all jars
+		assertTrue(classes.size() > 800, "expected more than 800 classes, got: " + classes.size());
+
+		assertAcyclic(classes);
+
+		// LoggerFactory.java:80 -> sets field LoggerFactory$DelegatingLogger.julLogger
+		// GeneralCodingRules.NO_CLASSES_SHOULD_USE_JAVA_UTIL_LOGGING.check(classes);
+
+		// LoggerFactory.java:155 -> new Throwable()
+		// GeneralCodingRules.NO_CLASSES_SHOULD_THROW_GENERIC_EXCEPTIONS.check(classes);
```
sormuras: @codecholeric How do I filter out a single class from a JavaClass instance? `org.junit.platform.commons.logging.LoggerFactory` should be allowed to break both generic rules.

codecholeric: You can always filter the classes by

```
JavaClasses subset = classes.that(predicate);
```

In your case it could be

```
import static com.tngtech.archunit.base.DescribedPredicate.not;
import static com.tngtech.archunit.lang.conditions.ArchPredicates.are;
import static com.tngtech.archunit.core.domain.JavaClass.Predicates.equivalentTo;

classes.that(are(not(equivalentTo(LoggerFactory.class)))
```

codecholeric:
There is an option `@AnalyzeClasses(locations = MyLocationProvider.class)` where you can freely implement `MyLocationProvider implements LocationProvider`. 
You could hard-code your Jars in there, or resolve a SystemProperty, etc.

sormuras:
I'm happy now with latest revision: https://github.com/junit-team/junit5/pull/1556/files#diff-2b9d518bc08ce207484660198b801365

Using `@AnalyzeClasses(locations = ArchUnitTests.AllJars.class)`, 5x `@ArchTest`. D'accord @codecholeric?

image

codecholeric:
Looks good to me :+1:

PR #1574, opened 2018-08-31
Refine the user guide and example

oohira:
I'm not sure which is appropriate English, `larger number` or `bigger number` ;-p

---


jbduncan:
@oohira Hmm. Even though I'm a native speaker, I'm actually not sure which one is more grammatically correct (if any one of them can be considered more "correct" than the other)!

To my mind, "larger number" reads better than "bigger number".

Do we have any grammarians here who can clarify things? :)

sbrannen:
I think "larger value" is perfectly fine in this scenario.

In any case, you can see that this question has been asked numerous times on [the Internet](https://www.google.com/search?q=larger+number+or+bigger+number&ie=utf-8&oe=utf-8). 😉

sbrannen:

Thanks

jbduncan:
Cool, thanks for sharing your research regarding "larger value" with us @sbrannen. :)

oohira:
@jbduncan @sbrannen Thanks for your comments!

Issue #162, opened 2016-02-21
Introduce SPI for generating display names


nicolaiparlog:
In many cases the strings passed to `@DisplayName` are just a prettified version of the class/method name, e.g.:

``` java
@DisplayName("List")
class ListTest {
    @Nested
    @DisplayName("when empty")
    class WhenEmpty {
        @Test
        @DisplayName("can add")
        void canAdd() {
            ...
        }
    }
}
```

It might be a nice-to-have if this could happen automatically, e.g.:

``` java
@DisplayName(prettify = true)
class ListTest {
    @Nested
    class WhenEmpty {
        @Test
        void canAdd() {
            ...
        }
        @Test
        @DisplayName("can't get")
        void cantGet() {
            ...
        }
    }
}
```

### Related Issues

- #145
- #153
- #432 
- #1036 
- #1577 
- #1589 

## Deliverables

- [ ] Introduce an extension for generating display names.
- [ ] Provide an out-of-the-box implementation of the extension that can be enabled via a JUnit configuration property.

schauder:
how about making 'prettify' the default action when no explicit name is given?

sbrannen:
I like the proposal from @schauder, and I have changed the title of this issue accordingly.

@junit-team/junit-lambda, what does the rest of the team think about making this change?

jlink:
I think we should discuss all related issues (#148 and #153) together and make consistent changes: name, display name and unique ID have intermingled concerns which we should separate first.

Von meinem iPad gesendet

bechte:
I like the idea and also support going forward into this direction. 

We should discuss in detail what prettify exactly means as we somehow set a standard for naming test methods and I know there are different styles in the community:

https://dzone.com/articles/7-popular-unit-test-naming

With this change we might make a statement which style should be used as we support it the best.

We simply should keep that in mind as we move on. Still, I support this change.


--

Stefan Bechtold

eCard: http://ecard.bechte.de

http://www.bechte.de, http://twitter.com/bechte

jlink:
I'd rather not have prettification as default because there'll always be some corner cases (eg. acronyms and abbreviations) where any automatic prettification scheme will fail. 

Von meinem iPad gesendet

signed:
I agree with @bechte about the different testing styles and the corner cases mentioned by @jlink. I'd prefer not having this complexity and ‘one size fits all’ naming convention in junit core.

How about providing an ExtensionPoint to provide the display name?
[...]

PR #1588, opened 2018-09-16
Introduce Display Name Generator SPI

sormuras:
## Overview

Addresses #162 

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -88,9 +99,22 @@
...
-			Function<E, String> defaultDisplayNameGenerator) {
+	protected static DisplayNameGenerator getDisplayNameGenerator(Class<?> testClass) {
+		Preconditions.notNull(testClass, "Test class must not be null");
+		Optional<ExtendWith> optionalExtendWith = findAnnotation(testClass, ExtendWith.class);
+		if (optionalExtendWith.isPresent()) {
+			for (Class<?> candidate : optionalExtendWith.get().value()) {
+				if (!DisplayNameGenerator.class.isAssignableFrom(candidate)) {
+					continue;
+				}
+				return (DisplayNameGenerator) ReflectionUtils.newInstance(candidate);
```
sormuras: This is _quick and dirty_ hack to get a handle to a user implementation of the `DisplayNameGenerator` interface.

Having support for such a declarative way is debatable ... and if it is needed/wanted, it shouldn't use the `ExtendWith` annotation.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DefaultDisplayNameGenerator.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Default implementation of the {@link DisplayNameGenerator} API.
+ *
+ * @since 5.4
+ */
+class DefaultDisplayNameGenerator implements DisplayNameGenerator {
+
+	@Override
+	public String generateDisplayNameForNestedClass(Class<?> nestedClass) {
+		if (nestedClass.getSimpleName().contains("_")) {
```
sormuras: Needs a better oracle for enabling the "make a sentence" path. Same goes for the analog `if` below in `generateDisplayNameForMethod`.

documentation/src/test/java/example/RepeatedTestsDemo.java
```diff
@@ -50,13 +50,13 @@ void repeatedTestWithRepetitionInfo(RepetitionInfo repetitionInfo) {
 	@RepeatedTest(value = 1, name = "{displayName} {currentRepetition}/{totalRepetitions}")
 	@DisplayName("Repeat!")
 	void customDisplayName(TestInfo testInfo) {
-		assertEquals(testInfo.getDisplayName(), "Repeat! 1/1");
+		assertEquals("Repeat! 1/1", testInfo.getDisplayName());
```
sbrannen: Please go ahead and push this to `master` since it is not related to this PR and should be adopted ASAP anyway.

documentation/src/test/java/example/RepeatedTestsDemo.java
```diff
@@ -50,13 +50,13 @@ void repeatedTestWithRepetitionInfo(RepetitionInfo repetitionInfo) {
...
 	}
 
 	@RepeatedTest(value = 1, name = RepeatedTest.LONG_DISPLAY_NAME)
 	@DisplayName("Details...")
 	void customDisplayNameWithLongPattern(TestInfo testInfo) {
-		assertEquals(testInfo.getDisplayName(), "Details... :: repetition 1 of 1");
+		assertEquals("Details... :: repetition 1 of 1", testInfo.getDisplayName());
```
sbrannen: ditto

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/extension/ExtensionComposabilityTests.java
```diff
@@ -53,18 +52,20 @@ void ensureJupiterExtensionApisAreComposable() {
 
 		List<String> expectedMethodSignatures = expectedMethods.stream()
 				.map(this::methodSignature)
+				.filter(name -> !name.contains("jacoco"))
```
sbrannen: Are those jacoco methods by any chance "synthetic"?

If so, please simply filter out synthetic methods instead of checking if the method name contains "jacoco".

In any case, please go ahead and push this to `master` since it is not related to this PR and should be adopted ASAP anyway.

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/extension/ExtensionComposabilityTests.java
```diff
@@ -53,18 +52,20 @@ void ensureJupiterExtensionApisAreComposable() {
...
 				.sorted()
 				.collect(toList());
 
 		List<String> expectedMethodNames = expectedMethods.stream()
 				.map(Method::getName)
+				.filter(name -> !name.contains("jacoco"))
```
sbrannen: ditto

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/DisplayNameGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.platform.commons.util.ClassUtils;
+
+/**
+ * {@code DisplayNameGenerator} defines the {@link Extension} API for
+ * generating display names programmatically.
+ *
+ * @since 5.4
+ */
+@API(status = EXPERIMENTAL, since = "5.4")
+public interface DisplayNameGenerator extends Extension {
```
sbrannen: This should not be an `Extension`. 

Although the deliverables for the issue state that it should be an "extension", the idea was not that it would actually be a JUnit Jupiter "extension" in the sense that it would receive an `ExtensionContext` or be registered like other extensions.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/DisplayNameGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import java.lang.reflect.Method;
+
+import org.apiguardian.api.API;
+import org.junit.platform.commons.util.ClassUtils;
+
+/**
+ * {@code DisplayNameGenerator} defines the {@link Extension} API for
```
sbrannen: I'm thinking we should refer to this as an "SPI" instead of an "Extension API" in order to avoid confusion.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/DisplayNameGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+@API(status = EXPERIMENTAL, since = "5.4")
+public interface DisplayNameGenerator extends Extension {
+
+	default String generateDisplayNameForClass(Class<?> testClass) {
+		String name = testClass.getName();
+		int index = name.lastIndexOf('.');
+		return name.substring(index + 1);
+	}
+
+	default String generateDisplayNameForNestedClass(Class<?> nestedClass) {
```
sbrannen: I have an issue with calling this "Nested".

I realize it's intended to only address `@Nested` test classes, but that is misleading at best, since a static nested test class (which is perfectly valid) is also a "NestedClass".

I'm thinking we should perhaps differentiate between static and non-static nested classes.

I have some other ideas about "static nested test classes" that I will likely address in a new issue.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DefaultDisplayNameGenerator.java
```diff
@@ -0,0 +1,53 @@
...
+import java.lang.reflect.Method;
+
+import org.junit.jupiter.api.extension.DisplayNameGenerator;
+
+/**
+ * Default implementation of the {@link DisplayNameGenerator} API.
+ *
+ * @since 5.4
+ */
+class DefaultDisplayNameGenerator implements DisplayNameGenerator {
```
sbrannen: This cannot be the `DefaultDisplayNameGenerator` since that would be a breaking change.

Instead, the default should in fact be called `DefaultDisplayNameGenerator`, but it must not override any of the `default` methods defined in the SPI.

This implementation should go in a separate class (perhaps only as a _demo_ for the time being).

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
sbrannen: I agree: let's get rid of the hack and do something better suited to the task at hand.

I'll add some notes to the original issue rather than discuss that here in this PR.

documentation/src/test/java/example/RepeatedTestsDemo.java
sormuras: Done via https://github.com/junit-team/junit5/commit/d024097eb5e6b5ffd28edfc783ecf7a2512dd27e

documentation/src/test/java/example/RepeatedTestsDemo.java
sormuras: Done via https://github.com/junit-team/junit5/commit/d024097eb5e6b5ffd28edfc783ecf7a2512dd27e

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
```diff
@@ -77,10 +79,6 @@ public final Method getTestMethod() {
 
 	@Override
 	public String getLegacyReportingName() {
-		return generateDefaultDisplayName(testMethod);
-	}
-
-	private static String generateDefaultDisplayName(Method testMethod) {
 		return String.format("%s(%s)", testMethod.getName(),
```
sbrannen: Shouldn't this simply delegate to the `DEFAULT` style in order to adhere to the DRY principle?

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/extension/ExtensionComposabilityTests.java
```diff
@@ -64,7 +64,7 @@ void ensureJupiterExtensionApisAreComposable() {
 
 		// 3) Dynamically implement all Extension APIs
 		Object dynamicKitchenSinkExtension = Proxy.newProxyInstance(getClass().getClassLoader(),
-			extensionApis.toArray(new Class<?>[extensionApis.size()]), (proxy, method, args) -> null);
+			extensionApis.toArray(new Class<?>[0]), (proxy, method, args) -> null);
```
sbrannen: You could push this to `master` as a separate polishing commit.

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/extension/ExtensionComposabilityTests.java
```diff
@@ -95,12 +95,9 @@ void ensureJupiterExtensionApisAreComposable() {
...
 
-		// @formatter:off
-		assertAll(
-				() -> assertThat(actualMethodSignatures).isEqualTo(expectedMethodSignatures),
-				() -> assertThat(actualMethodNames).isEqualTo(expectedMethodNames)
-		);
-		// @formatter:on
+		assertAll("Checking method signatures and names",
+			() -> assertLinesMatch(expectedMethodSignatures, actualMethodSignatures),
+			() -> assertLinesMatch(expectedMethodNames, actualMethodNames));
```
sbrannen: Why did you change the assertions here?

sbrannen:
@sormuras, what was the resolution for the jacoco vs. synthetic discussion?

https://github.com/junit-team/junit5/pull/1588#discussion_r218036391

sormuras:
> what was the resolution for the jacoco vs. synthetic discussion?

Mh, my comment is buried somewhere or was lost before being posted. Anyway, the underlying issue were the `default` methods introduced in the `DisplayNameGenerator` interface. They triggered JaCoCo to introduce the new _real_ method. This new method was not a bridge, and not synthetic.

As I removed the `default` methods from the inteface, the JaCoCo method vanished as well.

documentation/src/test/java/example/DisplayNameGeneratorDemo.java
```diff
@@ -0,0 +1,68 @@
...
+	static class UpperCaseMethodNames implements DisplayNameGenerator {
+
+		@Override
+		public String generateDisplayNameForClass(Class<?> testClass) {
+			return Style.DEFAULT.generateDisplayNameForClass(testClass);
+		}
+
+		@Override
+		public String generateDisplayNameForNestedClass(Class<?> nestedClass) {
+			return Style.KELVIN.generateDisplayNameForNestedClass(nestedClass);
```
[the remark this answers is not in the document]
sormuras: True. Will remove it from the demo.

sbrannen:
> As I removed the default methods from the inteface, the JaCoCo method vanished as well.

Aha. Interesting!

But that makes sense now. 

Thanks for the explanation.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,152 @@
...
+ * @see DisplayName
+ * @see DisplayNameGenerator
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@API(status = EXPERIMENTAL, since = "5.4")
+public @interface DisplayNameGeneration {
+
+	Class<? extends DisplayNameGenerator> generator() default Style.class;
```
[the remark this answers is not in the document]
sormuras: Ah, I thought that pointing to an interface was possible here. Will do.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,152 @@
...
+			@Override
+			public String generateDisplayNameForMethod(Method testMethod) {
+				Preconditions.notNull(testMethod, "Test method must not be null");
+				return String.format("%s(%s)", testMethod.getName(),
+					ClassUtils.nullSafeToString(Class::getSimpleName, testMethod.getParameterTypes()));
+			}
+		},
+
+		UNDERSCORE {
+			private String replaceUnderscore(String name) {
```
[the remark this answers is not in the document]
sormuras: Moved.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,152 @@
...
+				return replaceUnderscore(DEFAULT.generateDisplayNameForNestedClass(nestedClass));
+			}
+
+			@Override
+			public String generateDisplayNameForMethod(Method testMethod) {
+				return replaceUnderscore(DEFAULT.generateDisplayNameForMethod(testMethod));
+			}
+		},
+
+		// TODO CAMEL
```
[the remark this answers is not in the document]
sormuras: I though you wanted to implement this. Rename it as you do. ;-)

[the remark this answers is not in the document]
sormuras: I can do it as well, using your implementation hints from #162

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,152 @@
...
+		//
+		//			@Override public String generateDisplayNameForMethod(Method testMethod) {
+		//				return null;
+		//			}
+		//		},
+
+		/**
+		 * TODO Javadoc
+		 */
+		KELVIN {
```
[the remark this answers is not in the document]
sormuras: Based on the feedback of https://twitter.com/sormuras/status/1041213134239936512 it will be an often desired/wanted choice.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
sormuras: Good catch!

sormuras: Erm, no! The line belongs to `getLegacyReportingName`, which should not be changed. Or should it? 🤔

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/extension/ExtensionComposabilityTests.java
sormuras: Will revert all changes made to `ExtensionComposabilityTests.java` in this PR.

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/extension/ExtensionComposabilityTests.java
sormuras: `assertLinesMatch` provides a better reporting layout. But as above, I will revert all changes made to `ExtensionComposabilityTests.java` in this PR.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
sormuras: This method implements `org.junit.platform.engine.TestDescriptor#getLegacyReportingName` -- delegating to `DEFAULT` seems odd here.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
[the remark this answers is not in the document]
sbrannen: I don't know that that is a valid conclusion.

The tweet only displays a single level of nesting using `@Nested` and does not show the results of multiple levels of nesting which I think may be a bit awkward.

Furthermore, the "likes" might simply be for the automatic conversion of underscores to spaces (i.e., the lacking need to use `@DisplayName`).

One cannot infer from the "likes" that people actually want the use of ellipses or concatenation of display names with the display name of an enclosing class.

With the proposed "underscore" name generator, one can achieve practically the same level of readability without the need for `@DisplayName`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
sbrannen: Why? That's what it used to do.... delegate to the default.

_You_ changed it so that the code is now duplicated. 😇

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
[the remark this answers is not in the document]
sbrannen: I'm fine with implementing it. I'd likely enjoy it. 😉 

But it's also fine with me if you implement it. Just let me know your decision.

In any case, based on my understanding of "reviews" of existing solutions (in that Stack Overflow thread and elsewhere)... using multiple regular expressions would likely be considerably slower than a brute force iterative approach.

Perhaps we start with a RegEx approach to get something working and then switch to a custom, iterative algorithm if necessary. 🤔

sbrannen:
**Open Issues**: (not covered in a review)

- Should `DisplayNameGeneration` be `@Inherited` (in inheritance hierarchies)?
- Should `DisplayNameGeneration` be _inherited_ (in `@Nested` hierarchies)?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
sbrannen: Right. We could handle it like that.

But at the same time, you make a valid point: it's just a coincidence that they are the same.

Thus, if we delegate to the _default_, we should at least make sure we have a "regression test" in place for the legacy one to ensure that it doesn't accidentally diverge in the future (if we change the _default_).

sbrannen: Actually, we should have a regression test in place in any case. 😉

sbrannen: Regarding violation of the DRY principle, I'm leaning toward your original idea, that it's safer to duplicate for these two use cases.

I'll leave it up to you... 😇

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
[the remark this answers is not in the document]
sormuras: You can't generate a complete and indicative sentence with the underscore style. Only fragments.

The point here is concat fragments and have a one-sentence-story telling what the test is testing. Without the need to hop through a hierarchy. It really adds value when a test named:

- _A year is not supported if it is 0._

...fails compared to one that is named

- _if it is 0_ (scroll up in the report) _A year is not supported_

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
sormuras: ...let's keep the code duplication.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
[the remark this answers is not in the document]
sbrannen: Sure. I'm not saying there is no added value in such a feature.

I'm only saying that I'm not convinced that we need to rush that into core at the moment.

Maybe we should let it get some traction (and suggestions for improvement if necessary) somewhere like Pioneer.

Or if we do include it out of the box in Jupiter, I imagine we should make it _experimental_ in its initial form.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
sbrannen: SGTM!

Care to introduce the regression test to go along with it? 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,168 @@
...
+		/**
+		 * TODO Javadoc
+		 */
+		CAMEL_CASE {
+			private final Pattern camelCasePattern = Pattern.compile(String.format("%s|%s|%s",
+				"(?<=[A-Z])(?=[A-Z][a-z])", "(?<=[^A-Z])(?=[A-Z])", "(?<=[A-Za-z])(?=[^A-Za-z])"));
+
+			@Override
+			public String generateDisplayNameForClass(Class<?> testClass) {
+				return splitCamelCase(DEFAULT.generateDisplayNameForClass(testClass));
```
sbrannen: Please delegate to `UNDERSCORE` instead of `DEFAULT` so that underscores are also replaced in camel-cased method names.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
sormuras: _und Neo so:_ 😵

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
sbrannen: Or do you think we should handle underscores differently for the "camel-case" style?

Maybe converting `"_"` to `" - "`?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,168 @@
...
+
+			private String splitCamelCase(String name) {
+				return camelCasePattern.matcher(name).replaceAll(" ");
+			}
+		},
+
+		/**
+		 * TODO Javadoc
+		 */
+		SENTENCES {
```
sbrannen: Naming is difficult here.

I think it would help to prototype various examples with various levels of nesting to see what we are actually generating, and perhaps then a better name will "jump out at us".

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,168 @@
...
+			@Override
+			public String generateDisplayNameForClass(Class<?> testClass) {
+				return DEFAULT.generateDisplayNameForClass(testClass);
+			}
+
+			@Override
+			public String generateDisplayNameForNestedClass(Class<?> nestedClass) {
+				return nestedClass.getSimpleName().replace('_', ' ') + "...";
```
sbrannen: I don't understand why an ellipsis would be desired at this level but not for the enclosing, top-level class.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,168 @@
...
+				return name.replace('_', ' ');
+			}
+		},
+
+		/**
+		 * TODO Javadoc
+		 */
+		CAMEL_CASE {
+			private final Pattern camelCasePattern = Pattern.compile(String.format("%s|%s|%s",
+				"(?<=[A-Z])(?=[A-Z][a-z])", "(?<=[^A-Z])(?=[A-Z])", "(?<=[A-Za-z])(?=[^A-Za-z])"));
```
sbrannen: I'm assuming that's from that SO thread, in which case I seem to recall that it doesn't work for something like "camelCase" but that it does work for "CamelCase".

So we'll definitely have add a **lot** of tests here. 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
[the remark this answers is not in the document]
sbrannen: 👍

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
sormuras: Yes, that latter would be what I'd expect. Implementing it now.

- Source `CamelCase_ToUnderScore` 
- Target `Camel Case - To Under Score`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
sormuras: Good idea. Need a tons of templates for testing anyway, Morpheus says.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
sormuras: The name of the test class is used as the caption, the title, the specification. You could prepend it to each sentence like:
- "Feature A Spec: B is given [...] when doing X [...] then [...]."
- "Feature A Spec: C is given [...] when doing Y [...] then [...]."
- _"Feature A Spec: {P} Q {R}"_

So, you usually don't want to repeat the top-most enclosing box over and over again. The really intersting part is the start of each sentence, which start at level 1 (when 0 is the top-level).

sormuras: Start watching here https://vimeo.com/289852238#t=1370s for minute or two. Or five. :)

https://vimeo.com/289852238#t=1370s

sormuras:
> I think it would help to prototype various examples with various levels of nesting to see what we are actually generating, and perhaps then a better name will "jump out at us".

From "fragments" using `@DisplayName` for each fragment, thus duplicating the same information as encoded in the associated method name:

fragments

to "indicative sentences", merging contexts and auto-prettify display names based on the method names:

sentences

sormuras:
For the sake of "getting this PR ready soon", I'll move the `DisplayNameGeneration.Style.SENTENCES` into another PR or into [extension sample](https://github.com/junit-team/junit5-samples/tree/master/junit5-jupiter-extensions)  or to Pioneer.

sormuras:
Same goes for the `CamelCase`-style: https://gist.github.com/sormuras/be5ecf68f3695ccc638e75ee196027c7

sormuras:
> Should @DisplayNameGeneration be @Inherited (in inheritance hierarchies)?

:+1: ... it is already supported by using `AnnotationUtils#findAnnotation(...)` in `JupiterTestDescriptor#getDisplayNameGenerator`, methinks.

> Should @DisplayNameGeneration be inherited (in @Nested hierarchies)?

Would be a nice-to-have feature.

> What package(s) do we want the new annotation, interface, and implementations to reside in?

Close to `@DisplayName` ... or introduce an `org.junit.jupiter.api.spi` package to collect new SPIs and keep the `org.junit.jupiter.api` surface clean(er)?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -88,9 +97,33 @@
...
 	}
 
-	protected static <E extends AnnotatedElement> String determineDisplayName(E element,
-			Function<E, String> defaultDisplayNameGenerator) {
+	protected static DisplayNameGenerator getDisplayNameGenerator(Class<?> testClass) {
+		Preconditions.notNull(testClass, "Test class must not be null");
+		// first locate direct or meta-direct present annotations
+		DisplayNameGeneration generation = findAnnotation(testClass, DisplayNameGeneration.class).orElse(null);
+		// next, crawl enclosing classes
+		if (generation == null) {
```
sormuras: Class-based inheritance or enclosed-based relation first?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+ * @since 5.4
+ * @see DisplayName
+ * @see DisplayNameGeneration
+ */
+@API(status = EXPERIMENTAL, since = "5.4")
+public interface DisplayNameGenerator {
+
+	/**
+	 * Generate a display name for the given top-level test class.
+	 *
```
sbrannen: top-level or `static` nested class

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+	 * TODO Javadoc
+	 */
+	String generateDisplayNameForClass(Class<?> testClass);
+
+	/**
+	 * Generate a display name for the given nested test class.
```
sbrannen: nested --> `@Nested` as a link to the Javadoc for `org.junit.jupiter.api.Nested`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+	 *
+	 * TODO Javadoc
+	 * TODO Find better name, split into two different methods?
```
sbrannen: leave it "as is" in combination with the requested Javadoc changes

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+	/**
+	 * Generate a display name for the given method.
+	 *
+	 * TODO Javadoc
+	 *
+	 * @implNote The class instance passed as {@code testClass} may differ from
+	 * the returned class by {@code testMethod.getDeclaringClass()}: e.g., when
+	 * a test method is inherited from a super class.
+	 *
+	 * @param testClass current test class this method "belongs" to
```
sbrannen: Please reword. "belongs" is a bit unclear. 😉

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
sbrannen: Class-based first since that will also pick up annotations on interfaces as well as local declarations which must always override annotations on an enclosing class.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -88,9 +97,33 @@
...
+			Class<?> current = testClass.getEnclosingClass();
```
sbrannen: `current` -> `enclosingClass`

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -88,9 +97,33 @@
...
+			while (current != null) {
+				DisplayNameGeneration enclosedGeneration = current.getAnnotation(DisplayNameGeneration.class);
```
sbrannen: `enclosedGeneration` -> `enclosingGeneration`

sbrannen: We cannot use `getAnnotation()` since it will not pick up meta-annotations.

Please use `AnnotationUtils.findAnnotation()` instead.

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/DisplayNameGenerationTests.java
```diff
@@ -0,0 +1,234 @@
...
+		void test(TestInfo testInfo) {
+			testInfo.getDisplayName();
+		}
+
+		@Test
+		void testUsingCamelCaseStyle() {
+		}
+
+		@Test
+		void testUsingCamelCase_and_also_UnderScores() {
```
sbrannen: Please add a test for camel-case, underscores, and method parameters.

sormuras:
All comments addressed, starting with the documentation.

sormuras:
Removed the WIP prefix from the issue title. @sbrannen shall I merge to `master` and you polish the documentation there?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,126 @@
...
+	 */
+	Class<? extends DisplayNameGenerator> generator() default DisplayNameGenerator.class;
+
+	/**
+	 * The pre-defined style to use.
+	 *
+	 * @return the style to use, can be overridden by a custom display
+	 * name {@link #generator()} implementation
+	 */
+	Style value() default Style.DEFAULT;
```
marcphilipp: Do we need the two separate properties or could we define `DEFAULT` and `UNDERSCORE` as `Class<? extends DisplayNameGenerator>` constants in `DisplayNameGenerator`?

sbrannen: That's also an option, but then it might make sense to get rid of the enum and rework the current _delegation_ for reused features.

Actually, I haven't put too much thought into it, but would that change require that we get rid of the enum altogether?

sbrannen: Rationale: I'm concerned about instantiating the enum type via reflection, b/c I assume that's not permitted.

sormuras: Moving types as Marc suggested for reviewing it in a dedicated commit - may revert it, if we don't like it.

sormuras: See diff in https://github.com/junit-team/junit5/pull/1588/commits/6e7f2a3325c31cbb85882f7c097dadd14a9e4341 -- only drawback is that only the default generator is held as an internal singleton instance. All other pre-defined generators (that is only `ReplaceUnderscores` for now) need to be instantiated at runtime.

sbrannen: We could always introduce a static cache to overcome that drawback (map of instances keyed by FQCN).

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,47 @@
...
+@API(status = EXPERIMENTAL, since = "5.4")
+public @interface DisplayNameGeneration {
+
+	/**
+	 * Custom display name generator.
+	 *
+	 * @return custom display name generator implementation, defaults
+	 *         to {@link DisplayNameGenerator.StandardGenerator}
+	 */
+	Class<? extends DisplayNameGenerator> value() default DisplayNameGenerator.StandardGenerator.class;
```
sbrannen: Since there is now a single attribute in this annotation, there should be no `default` for `value`.

sormuras: Sure. Left over from before the refactoring.

sormuras: Force-pushed required changes.

sormuras:
@marcphilipp anything else? If not, I'll continue with documentation.

andreybleme:
Hi @sbrannen and @sormuras, what's avoiding this change to be merged at `master`? Just asking because it blocks #1595, which I'll start working.

sormuras:
A final review by @marcphilipp tonight or ... soon. ;-)

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
```diff
@@ -0,0 +1,122 @@
...
+			// handling validation exceptions during the TestEngine discovery phase.
+			if (StringUtils.isBlank(displayName)) {
+				logger.warn(() -> String.format(
+					"Configuration error: @DisplayName on [%s] must be declared with a non-empty value.", element));
+			}
+			else {
+				return displayName;
+			}
+		}
+		// else let a 'DisplayNameGenerator' generator generate a display name
```
marcphilipp: :x: remove "generator"

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/DisplayNameGenerationTests.java
```diff
@@ -0,0 +1,245 @@
...
+		void test_with_underscores() {
+		}
+
+		@DisplayName("@DisplayName prevails")
+		@Test
+		void testDisplayNamePrevails() {
+		}
+	}
+
+	@DisplayNameGeneration(DisplayNameGenerator.StandardGenerator.class)
```
marcphilipp: :o: `DisplayNameGenerator.Standard.class` class so it's more consistent with `ReplaceUnderscores`?

sormuras: ✔️

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
sormuras: ✔️

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
```diff
@@ -0,0 +1,122 @@
...
+		// else: create an instance of the supplied generator implementation class and return it
+		return ReflectionUtils.newInstance(displayNameGeneratorClass);
+	}
+
+	/**
+	 * Find the first {@code DisplayNameGeneration} annotation that is either
+	 * <em>directly present</em>, <em>meta-present</em>, <em>indirectly present</em>
+	 * on the supplied {@code testClass} or on an enclosing class.
+	 */
+	private static DisplayNameGeneration getDisplayNameGenerationOrNull(Class<?> testClass) {
```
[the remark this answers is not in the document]
sormuras: Little `null` is sad. So sad. Will revert it to `private static Optional<DisplayNameGeneration> getDisplayNameGeneration`, as it was before.

✔️

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+	/**
+	 * Standard display name generator.
+	 *
+	 * <p>The implementation matches the published behaviour when Jupiter 5.0.0
+	 * was released.
+	 */
+	class StandardGenerator implements DisplayNameGenerator {
+		@Override
+		public String generateDisplayNameForClass(Class<?> testClass) {
+			Preconditions.notNull(testClass, "Test class must not be null");
```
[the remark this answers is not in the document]
sormuras: ✔️

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+	String generateDisplayNameForMethod(Class<?> testClass, Method testMethod);
+
+	/**
+	 * Compile a string representation from all simple parameter type names.
+	 *
+	 * @param method the method providing parameter types for the result
+	 * @return a string representation of all parameter types of the
+	 *         supplied method or {@code "()"} if the method has no parameters
+	 */
+	static String parameterTypesAsString(Method method) {
```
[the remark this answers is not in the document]
sormuras: ✔️

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGeneration.java
```diff
@@ -0,0 +1,46 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@API(status = EXPERIMENTAL, since = "5.4")
+public @interface DisplayNameGeneration {
+
+	/**
+	 * Custom display name generator.
+	 *
+	 * @return custom display name generator implementation
```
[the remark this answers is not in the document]
sormuras: ✔️

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -86,6 +87,33 @@ characters, and even emojis -- that will be displayed by test runners and test r
 include::{testDir}/example/DisplayNameDemo.java[tags=user_guide]
 ----
 
+==== Display Name Generators
+
+JUnit Jupiter supports since version 5.4.0-SNAPSHOT custom display name generators.
```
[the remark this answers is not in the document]
sormuras: ✔️

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+@API(status = EXPERIMENTAL, since = "5.4")
+public interface DisplayNameGenerator {
+
+	/**
+	 * Generate a display name for the given top-level or {@code static} nested test class.
+	 */
+	String generateDisplayNameForClass(Class<?> testClass);
+
+	/**
+	 * Generate a display name for the given {@link Nested} inner test class.
```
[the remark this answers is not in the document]
sormuras: ✔️

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -86,6 +87,32 @@ characters, and even emojis -- that will be displayed by test runners and test r
 include::{testDir}/example/DisplayNameDemo.java[tags=user_guide]
 ----
 
+==== Display Name Generators
+
+JUnit Jupiter supports custom display name generators via the `@DisplayNameGeneration` annotation.
+Values provided with a `@DisplayName` annotations always take precedence over display names
```
marcphilipp: :x: Remove the "s" from "annotations"

PR #1610, opened 2018-09-29
Publish documentation snapshot only if necessary

sormuras:
## Overview

Prior to this commit the current documentation sources were always compiled, built, prepared, and published even if nothing was edited.

Now, the file `published-checksum.txt` containing a checksum for each documentation source asset is used to determine if a build and publish is necessary. It will show up here: https://github.com/junit-team/junit5/tree/gh-pages/docs/snapshot

---


src/publishDocumentationSnapshotOnlyIfNecessary.sh
```diff
@@ -0,0 +1,31 @@
...
+  # no changes detected: we're done
+  #
+  echo "Already published documentation with same source checksum."
+else
+  #
+  # update checksum file and trigger new documentation build and upload
+  #
+  echo "Creating and publishing documentation..."
+  cp --force "${current}" "${published}"
+  echo "./gradlew --scan gitPublishPush"
```
sormuras: Dooh. Why is "poor man's dry-run mode" still active?!

sormuras:
Including all `src/main/java/**.java` files to detect potential changes in Javadoc comments.

src/publishDocumentationSnapshotOnlyIfNecessary.sh
```diff
@@ -26,6 +29,6 @@ else
   # update checksum file and trigger new documentation build and upload
   #
   echo "Creating and publishing documentation..."
-  cp --force "${current}" "${published}"
-  echo "./gradlew --scan gitPublishPush"
+  cp --force "${current}" "${published}"ca
```
marcphilipp: `ca`?

marcphilipp: Why so forcefully?

sormuras: `ca` ... _ceine ahnung_ ;-)

`--force` to overwrite existing one.

sormuras:
The script-based guard prevents any Gradle task to start at all. See how many CPU cycles are "wasted" for nothing: https://scans.gradle.com/s/sdegvcf4c4sra/performance/execution

Issue #1139, opened 2017-11-03
Introduce extension API for accessing arguments passed to tests


rumpfc:
## Issue

Extension methods all have `ExtensionContext` which provides us many useful information like `displayname`, `method` or `tags`. This helps  in writing custom test reports.

What I'm missing is a way to get argument instances of a `@ParameterizedTest`. When not running tests extended by a `ParameterResolver` using `ExtensionContext.Store` you can only make a workaround to get a clue of what arguments were used for the test:

- Filter displayname (objects need to override `toString()` for important informations)
- get values from annotations

If parameters are simple strings, integers etc. there is no problem when parsing displayname. The problem starts when one of the parameters is an object containing lots of informations. A new instance would have to be created when analyzing the test, but this doesn't garanty an object with exact the same content like in the test and might also lead to memory issues.

Most simple way is to reimplement Parameterized Test Template, extend from each class, catch resolved parameters and store them directly in ExtensionContext.Store, but I'm not a fan of writing Wrapper classes for such small extensions in existing codes.

## Possible Suggestions

- Provide a method `Arguments[] ExtensionContext.getArguments()` or `List<Arguments> ExtensionContext.getArguments()` which returns a list of all Arguments used for the test (empty list if no arguments were used).
- Store arguments in `ExtensionContext.Store` with a predefined namespace in your `ParameterizedTestParameterResolver` class (then there is no need to do it in custom `ArgumentsProvider` classes)
- Provide an Extension interface which is called between `BeforeTestExecutionCallback` and `@Test`. As parameters it should have `ExtensionContext` and `Arguments[]`.

## Related Issues

- #944
- #1668
- #1884

marcphilipp:
This might be useful for more than just `@ParameterizedTest` methods. Right now the arguments for a test method get resolved immediately before calling it, i.e. after `BeforeTestExecutionCallback` extensions are called. When would you need the arguments? Would it suffice to get access to them after the test has been executed?

rumpfc:
Test analysis are mostly done after the test execution, so I think it's ok to just have them in one of the AfterXXX interfaces. Getting arguments before test execution would also be nice, but instead of storing them in an attribute I don't know what someone should do with them before the actual test.

[...]

PR #1614, opened 2018-10-03
[WIP] Introduce extension API for accessing arguments passed to tests

paul-brooks:
## Overview

Addresses : #1139

Adding a BeforeParameterizedTestExecutionCallback to be called prior to test method execution with a copy of the actual test arguments.

Points for discussion:

- Naming of the callback.
- Currently, the arguments are resolved twice via the DefaultParameterResolver.  It would be better to do this once (and pass a copy to the callback) but this would require a change to the ExecutableInvoker API.  I'm happy to do this but thought I'd wait from feedback first.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
@@ -23,12 +23,14 @@ import org.apiguardian.api.API;
 import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.AfterTestExecutionCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.BeforeParameterizedTestExecutionCallback;
 import org.junit.jupiter.api.extension.BeforeTestExecutionCallback;
 import org.junit.jupiter.api.extension.Extension;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.TestExecutionExceptionHandler;
 import org.junit.jupiter.engine.execution.AfterEachMethodAdapter;
 import org.junit.jupiter.engine.execution.BeforeEachMethodAdapter;
+import org.junit.jupiter.engine.execution.DefaultParameterResolver;
 import org.junit.jupiter.engine.execution.ExecutableInvoker;
 import org.junit.jupiter.engine.execution.JupiterEngineExecutionContext;
 import org.junit.jupiter.engine.extension.ExtensionRegistry;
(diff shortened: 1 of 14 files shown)


sbrannen:
Please see my notes in https://github.com/junit-team/junit5/issues/1139#issuecomment-422790354 as well, especially the last paragraph.

paul-brooks:
Thanks for the feedback Sam.

I figured as much about  the naming.  I'll just revert to ArgumentsProcessor.

> Please see my notes in #1139 (comment) as well, especially the last paragraph.

I did miss that bit about the flag at the end of that comment, however wouldn't that mean that the `ExecutableInvoker` would need to know about and also invoke any `ArgumentsProcessor`'s that are present (and also handle exceptions that they may throw)? 

Isn't this what the `TestMethodTestDescriptor` already does for the existing callbacks?

It seems that we'd be adding to the responsibilities of the `ExecutableInvoker` with functionality that already exists in the `TestMethodTestDescriptor`.

Ultimately my thinking was that for test methods (which are only called from `TestMethodTestDescriptor` hierarchy?) we could resolve the parameters there and then pass them to a new method on `ExecutableInvoker`, thus only resolving them once and keeping the `ExecutableInvoker` clean.

Cheers

Paul

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestMethodTestDescriptor.java
```diff
@@ -160,6 +163,22 @@ private void invokeBeforeTestExecutionCallbacks(JupiterEngineExecutionContext co
 		}
 	}
 
+	private void resolveParametersAndInvokeTestMethod(JupiterEngineExecutionContext context,
+			DynamicTestExecutor dynamicTestExecutor, ThrowableCollector throwableCollector) {
+		ExtensionRegistry registry = context.getExtensionRegistry();
+		Object[] arguments = parameterResolver.resolveParameters(getTestMethod(), Optional.empty(),
```
sbrannen: The test instance cannot be "empty". You'll have to retrieve that and pass it on like in `invokeTestMethod()`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/DefaultParameterResolver.java
```diff
@@ -0,0 +1,173 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+import org.junit.jupiter.engine.extension.ExtensionRegistry;
+import org.junit.platform.commons.logging.Logger;
+import org.junit.platform.commons.logging.LoggerFactory;
+import org.junit.platform.commons.util.Preconditions;
+
+public class DefaultParameterResolver {
```
sbrannen: Since this not an implementation of `ParameterResolver`, we cannot call it `DefaultParameterResolver`.

Instead, we should name it something like `ParametersResolver`. The "s" is admittedly easy to overlook. So maybe we'll come up with something better later.

paul-brooks: Not sure if you're agreeing to extract this as a class now?  

Anyway, I've pushed some more changes with what I think you're saying.

marcphilipp:
Blocked by https://github.com/junit-team/junit5/issues/1139#issuecomment-513185506.

lastthyme (approved):

PR #1634, opened 2018-10-13
add global resource lock

Danny02:
Why:
Some libraries provide test modules which work by modifying private
global variables inside the library (i.e. Project Reactor virtual time
step verifier). When such library is used as a basis for the whole
app, nearly all test are affected by this manipulation. Parallel
test execution becomes inpracticable, because all methods would need
to be annotated with a READ ResourceLock and the few test methods
which modify the global state one of type READ_WRITE.

## Overview

The changes are only a quick&dirty implementation to get the discussion started. 

I would really like to implement the feature, if you can give me some guidance about integrating it "correctly" into the framework.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/GlobalResourceLock.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/GlobalResourceLock.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api.parallel;
+
+import org.apiguardian.api.API;
+
+import java.lang.annotation.*;
+import java.util.UUID;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+@API(status = EXPERIMENTAL, since = "5.3")
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+public @interface GlobalResourceLock {
+}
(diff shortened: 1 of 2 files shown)


marcphilipp:
I don't think a single global lock is sufficient for all such use cases. Instead, we could introduce a configuration parameter for "default read locks" (a default read-write lock doesn't make sense now, does it?), e.g. a comma-separated list as in the following example, and use the normal `@ResourceLock` annotation to specify that `READ_WRITE` access is required.

```
junit.jupiter.execution.locks = foo,bar
```

```java
@ResourceLock(value = "foo", mode = READ_WRITE)
@Test
void aTest() { ... }
```

@Danny02 Would that work for you?

@junit-team/junit-lambda WDYT?

@leonard84 This one might be interesting for you, too.

Danny02:
Hi @marcphilipp,

I thought about that too. The thing is that there is no need to have multiple "default read locks", besides documentation.

As in your example:
The behavior of a "foo" or "bar" READ_WRITE lock is the same, namely blocking all other test executions.

Other thoughts I had:
I think introducing this behavior only to the jupiter engine and not the platform is problematic (as I did in my example code). I would want the lock (global or default) to work for all tests not only junit ones. I for example also the jqwik test engine.

leonard84:
I don't know if this is such a good idea, are really 100% of your tests using this shared state? Adding a lock on a class also implies `SameThreadExecution` for all its children, and due to lock coarsening we pull up all locks to the class level.

Danny02:
To give you a better understanding. I'm writing a Spring-Boot webflux application, which uses reactive streams [Project Reactor](https://projectreactor.io/) for everything (from database queries to HTTP requests/responses). 
Same of my test methods do not touch this library at all, but most ( > 90%) tests call some code which uses the reactive streams.

At the moment I have tagged the methods using the virtual time and have configured two surefire executions. The problem is that I also have to configure my IDE to execute the tagged tests separately.

Another idea would be to respect annotations on package objects. I could then easily add my own READ lock on the toplevel package?

Danny02:
I'm locking at the code atm and have two questions.
1. @leonard84 Why do we use lock coarsening? Wouldn't that mean if the class has some READ lock and a method another READ_WRITE lock, that suddenly the whole class uses this READ_WRITE lock?
2. I saw that the locks are managed in the platform, but the annotations are collected in the test engine. Does this mean that third party test engines have to collect the annotations themself?

PR #1635, opened 2018-10-14
Introduce 'empty or null' source annotations for parameterized tests

jmaniquet:
## Overview

I would like to propose two new annotations :

- `@EmptyStringSource`: provides a parameterized test with the following values : the `null` String, the empty String, and the blank String
- `@EmptyListSource`: provides a parameterized test with the following values : the `null` List and the empty List

### Reasoning

- Given a method having a String input, I often have preconditions throwing an error on the null/empty/blank values
- I have then to cover all those use cases in unit tests
- Same for the null/empty list
- I can't pass `null` to `@ValueSource` or any other annotation

I ended up developing those annotations to avoid redundant code. And following that, I ended up copy/pasting them in several projects.

So I thought maybe it could be baseline in `junit-jupiter-params`.

What do you think?

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptyAndNullLists.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptyAndNullLists.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.params.provider;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+
+/**
+ * {@code EmptyListSource} is an {@link ArgumentsSource} which provides the
+ * following values :
+ *
+ * <ul>
+ * 	<li>The null list</li>
+ * 	<li>The empty list</li>
+ * </ul>
+ *
+ * <p>The supplied values will be provided as arguments to the annotated
+ * {@code @ParameterizedTest} method.
+ *
+ * <p>It is often useful to have the same test operate on both of these values.
+ *
+ * @since 5.0
+ * @see org.junit.jupiter.params.provider.ArgumentsSource
+ * @see org.junit.jupiter.params.ParameterizedTest
+ */
+@Target({ ElementType.ANNOTATION_TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@API(status = EXPERIMENTAL, since = "5.4")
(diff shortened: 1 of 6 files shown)


leonard84:
I think the names are confusing, `EmptyList` implies for me just that a list without values, `null` is not an empty `List`. The same applies to the String variant.

sbrannen:
> So I thought maybe it could be baseline in `junit-jupiter-params`.
> 
> What do you think?

I agree with @leonard84 that the proposed annotation names are misleading.

But in any case, I think these are rather corner cases and that inclusion of support for them in JUnit Jupiter might be overkill.

For example, if you're testing a `String` input and want to verify the expected behavior for `null` and _empty_ values, you typically write two separate tests if the expected behavior is different; whereas, if the expected behavior is the same for both of those values it's usually quite simple to simply write the test twice (or delegate to a private helper method, or use dynamic tests).

However, having said that, I suppose one _might_ want to test the "same behavior" scenario as a parameterized test.

Regarding blank strings, the term "blank" is completely subjective and cannot be assumed to be `n` spaces. A _blank_ string may contain various types of whitespace characters in any combination imaginable. So if we do implement a "null or empty" source for strings, I do not think we should include support for "blank" strings as well.

sbrannen:
Regarding the "null or empty" list source, why would we want to limit such support to lists? Wouldn't it make more sense to support arrays, maps, and all common collection types?

But then... that might be opening up a can of worms.

sbrannen:
@junit-team/junit-lambda, Tentatively slated for 5.4 M1 solely for the purpose of _team discussion_.

sbrannen:
FWIW, if you want to implement support for null or empty lists, you really don't even need a custom `ArgumentsProvider`.

You can do something as simple as the following:

```java
public class MethodSources {
	public static Stream<List<?>> emptyAndNullLists() {
		return Stream.of(null, Collections.emptyList());
	}
}
```

```java
@Target(ElementType.METHOD)
@MethodSource("org.example.MethodSources#emptyAndNullLists()")
public @interface EmptyAndNullLists {
}
```

```java
class Tests {

	@ParameterizedTest
	@EmptyAndNullLists
	void emptyAndNullLists(List<?> list) {
		assertTrue(list == null || list.isEmpty());
	}
}
```

😉

jmaniquet:
> it's usually quite simple to simply write the test twice (or delegate to a private helper method, or use dynamic tests).

When not using junit5, i use the 'two test with delegate private method' technique.
But when using junit5 and parameterized support, the addtional method starts to feel like more of an annoyance than anything else, so i end up with this to avoid it.

As for the proposed names, i agree that they were confusing; so NullAndEmptyStrings / NullAndEmptyCollections (instead of list) seem fine to me.

As for the blank support, i believe it have some uses, like an end user accidentally putting a space in a form field. So i see value in treating it like the empty string.
I agree it could be better designed, to accomodate various meanings for 'blank' .
So what about an annotation attribute, something like :
`String` [] blankValues() default " ";

I also thought about array and map support, but i wanted to have feedback first on that idea.

sbrannen:
If you're following this issue, please see https://github.com/junit-team/junit5/issues/1637#issuecomment-430552277 for my latest proposal for this topic.

sbrannen:
@jmaniquet,

> I also thought about array and map support, but i wanted to have feedback first on that idea.

For future reference, the best way to gather feedback on ideas is to open an issue for discussion first and then submit a PR based on the consensus reached in the issue.  That way we avoid writing code and tests that later get superseded. 😉

sbrannen:
@jmaniquet, I would appreciate your feedback on the latest proposals in #1637.

TIA

sbrannen:
Closing this PR in favor of #1637.

PR #1648, opened 2018-10-24
Wait for dynamic tests before finishing test factory

### LINKED ISSUE #1643: hwen using parallel execution, Callback methods/extensions should be called after all dynamic tests of a @TestFactory method are finished
Given a test case like the following:

```java
class ConcurrentDynamicTestCase {
	@AfterEach
	void afterEach() {
		...
	}
	@TestFactory
	DynamicTest testFactory() {
		return dynamicTest("slow", () -> {
			Thread.sleep(500);
		});
	}
}
```

## Expected behavior

`afterEach` is called _after_ all dynamic tests from `testFactory()` are finished.

## Current behavior

`afterEach` is called before.

marcphilipp:
When executing tests in parallel, lifecycle methods and callbacks called
after a `@TestFactory` method are now always executed after the dynamic
tests returned by the method.

Fixes #1643.

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
@@ -16,11 +16,14 @@ import static org.junit.platform.engine.TestExecutionResult.failed;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
 import org.junit.platform.commons.JUnitException;
+import org.junit.platform.commons.util.ExceptionUtils;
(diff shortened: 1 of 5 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTestTask.java
```diff
@@ -155,4 +139,35 @@ private void reportCompletion() {
...
+		}
+
+		@Override
+		public void awaitFinished() throws InterruptedException {
+			for (Future<?> future : futures) {
+				try {
+					future.get();
+				}
+				catch (ExecutionException e) {
+					ExceptionUtils.throwAsUncheckedException(e);
```
marcphilipp:
```suggestion
					ExceptionUtils.throwAsUncheckedException(e.getCause());
```

Issue #1611, opened 2018-10-01
Introduce better support for accessing fields


britter:
## Feature Request

_Add APIs for accessing fields of test classes and instances._

When authoring an extension one often needs to access fields of a test class or instance fields of a test instance. Usually one is interested in fields of a certain type possibly annotated with a certain method. There is currently no API to do this in JUnit so one has to fall back to raw reflection code.

I see two ways to implement this:

1. Add more methods to `ReflectionSupport` - this will work but I don't think we should put more and more functionality into these helper classes

```java
/**
 * Get the values of the static fields annotated with the given annotation from the given class.
 */
List<Object> getAnnotatedFields(Class<?>, Class<? extends Annotation>)

/**
 * Get the values of the static fields annotated with the given annotation of the given
 * type from the given class.
 */
<T> List<T> ReflectionSupport.getAnnotatedFieldsOfType(Class<?>, Class<? extends Annotation>, Class<T> type)

/**
 * Get the values of the instance fields annotated with the given annotation from the given object.
 */
List<Object> ReflectionSupport.getAnnotatedFields(Object, Class<? extends Annotation>)

/**
 * Get the values of the instance fields annotated with the given annotation of the given
 * type from the given object.
 */
<T> List<T> ReflectionSupport.getAnnotatedFieldsOfType(Object, Class<? extends Annotation>, Class<T> type)

```

2. Follow a more object-oriented approach and create specialized objects that help with this:

```java
/**
 * A wrapper around the test class being executed.
 */
public interface TestClass {

    List<Object> getAnnotatedFieldValues(Class<? extends Annotation>)

    <T>  List<T> getAnnotatedFieldValues(Class<? extends Annotation>, Class<T>)
}

/**
 * A wrapper around the test instance being executed.
 */
public interface TestInstance {

    List<Object> getAnnotatedFieldValues(Class<? extends Annotation>)

    <T>  List<T> getAnnotatedFieldValues(Class<? extends Annotation>, Class<T>)
} 
```

The specialized objects could be accessed from the `ExtensionContext`. Methods like `getRequiredTestClass` et al. could be deprecated.

## Deliverables

- [x] Introduce API for easier access of fields inside of test classes and test instances.
- [x] Document in User Guide.
   - ❌ deemed unnecessary since the new features are documented in Javadoc.
- [x] Document in Release Notes.

britter:
[...]

PR #1650, opened 2018-10-26
Introduce ExtensionContext.getTestInstances() API

### LINKED ISSUE #1618: API for accessing outer test instances
## Overview

- (x) **Feature request.** Provide an API to access outer test instance

Extensions that manage test instance fields often need access to the fields declared in outer test instances. Examples for this are the [MockitoExtension](https://github.com/mockito/mockito/blob/release/2.x/subprojects/junit-jupiter/src/main/java/org/mockito/junit/jupiter/MockitoExtension.java) and the [Testcontainers extension](https://github.com/testcontainers/testcontainers-java/pull/887). The `ExtensionContext` currently does not provide access to the outer test instance:

```java
        Optional<ExtensionContext> current = context.getParent();

        while(current.isPresent()) {
            ExtensionContext ctx = current.get();
            Optional<Object> testInstance = ctx.getTestInstance()
            if (testInstance.isPresent()) { // will never be true
                // do something with testInstance
            }
            current = ctx.getParent();
        }
```

So currently extension authors have to fall back to putting test instances into the extension store and retrieve it from there. This should be more straight forward.

## Deliverables

- [ ] API for accessing outer test instances

marcphilipp:
Resolves #1618.

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -10,6 +10,7 @@
 
 package org.junit.jupiter.api.extension;
 
+import static org.apiguardian.api.API.Status.MAINTAINED;
 import static org.apiguardian.api.API.Status.STABLE;
 
 import java.lang.reflect.AnnotatedElement;
@@ -160,7 +161,9 @@ public interface ExtensionContext {
 	 * @return an {@code Optional} containing the test instance; never
 	 * {@code null} but potentially empty
 	 * @see #getRequiredTestInstance()
+	 * @see #getTestInstances()
 	 */
+	@API(status = MAINTAINED, since = "5.4")
 	Optional<Object> getTestInstance();
 
 	/**
@@ -173,10 +176,48 @@ public interface ExtensionContext {
 	 * @return the test instance; never {@code null}
 	 * @throws PreconditionViolationException if the test instance is not present
 	 * in this {@code ExtensionContext}
+	 *
+	 * @see #getRequiredTestInstances()
 	 */
+	@API(status = MAINTAINED, since = "5.4")
 	default Object getRequiredTestInstance() {
 		return Preconditions.notNull(getTestInstance().orElse(null),
 			"Illegal state: required test instance is not present in the current ExtensionContext");
+	}
+
+	/**
+	 * Get the test instances associated with the current test or container,
+	 * if available.
+	 *
+	 * <p>While top-level tests only have a single test instance, nested tests
+	 * have one instance per containing test class.
+	 *
+	 * @return an {@code Optional} containing the test instances; never
+	 * {@code null} but potentially empty
+	 * @see #getRequiredTestInstances()
+	 *
+	 * @since 5.4
+	 */
+	@API(status = MAINTAINED, since = "5.4")
+	Optional<TestInstances> getTestInstances();
+
+	/**
+	 * Get the <em>required</em> test instances associated with the current test
+	 * or container.
+	 *
+	 * <p>Use this method as an alternative to {@link #getTestInstances()} for use
+	 * cases in which the test instances are required to be present.
+	 *
+	 * @return the test instances; never {@code null}
+	 * @throws PreconditionViolationException if the test instances are not present
+	 * in this {@code ExtensionContext}
+	 *
+	 * @since 5.4
+	 */
+	@API(status = MAINTAINED, since = "5.4")
+	default TestInstances getRequiredTestInstances() {
+		return Preconditions.notNull(getTestInstances().orElse(null),
+			"Illegal state: required test instances are not present in the current ExtensionContext");
 	}
 
 	/**
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -10,7 +10,7 @@
 
 package org.junit.jupiter.api.extension;
 
-import static org.apiguardian.api.API.Status.MAINTAINED;
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.STABLE;
 
 import java.lang.reflect.AnnotatedElement;
@@ -163,7 +163,6 @@ public interface ExtensionContext {
 	 * @see #getRequiredTestInstance()
 	 * @see #getTestInstances()
 	 */
-	@API(status = MAINTAINED, since = "5.4")
 	Optional<Object> getTestInstance();
 
 	/**
@@ -179,7 +178,6 @@ public interface ExtensionContext {
 	 *
 	 * @see #getRequiredTestInstances()
 	 */
-	@API(status = MAINTAINED, since = "5.4")
 	default Object getRequiredTestInstance() {
 		return Preconditions.notNull(getTestInstance().orElse(null),
 			"Illegal state: required test instance is not present in the current ExtensionContext");
@@ -190,7 +188,7 @@ public interface ExtensionContext {
 	 * if available.
 	 *
 	 * <p>While top-level tests only have a single test instance, nested tests
-	 * have one instance per containing test class.
+	 * have one additional instance for each enclosing test class.
 	 *
 	 * @return an {@code Optional} containing the test instances; never
 	 * {@code null} but potentially empty
@@ -198,7 +196,7 @@ public interface ExtensionContext {
 	 *
 	 * @since 5.4
 	 */
-	@API(status = MAINTAINED, since = "5.4")
+	@API(status = EXPERIMENTAL, since = "5.4")
 	Optional<TestInstances> getTestInstances();
 
 	/**
@@ -214,7 +212,7 @@ public interface ExtensionContext {
 	 *
 	 * @since 5.4
 	 */
-	@API(status = MAINTAINED, since = "5.4")
+	@API(status = EXPERIMENTAL, since = "5.4")
 	default TestInstances getRequiredTestInstances() {
 		return Preconditions.notNull(getTestInstances().orElse(null),
 			"Illegal state: required test instances are not present in the current ExtensionContext");
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api.extension;
+
+import static org.apiguardian.api.API.Status.MAINTAINED;
+
+import java.util.List;
+import java.util.Optional;
+
+import org.apiguardian.api.API;
+
+/**
+ * {@code ParameterContext} encapsulates the <em>test instances</em> of a test.
+ *
+ * <p>While top-level tests only have a single test instance, nested tests
+ * have one instance per containing test class.
+ *
+ * @since 5.4
+ * @see ExtensionContext#getTestInstances()
+ * @see ExtensionContext#getRequiredTestInstances()
+ */
+@API(status = MAINTAINED, since = "5.4")
+public interface TestInstances {
+
+	/**
+	 * Get the innermost test instance.
+	 *
+	 * <p>The innermost instance is the one closest to the test method.
+	 */
+	Object getInnermost();
+
+	/**
+	 * Get the enclosing test instances, excluding the innermost test instance,
+	 * ordered from outermost to innermost.
+	 */
+	List<Object> getEnclosing();
+
+	/**
+	 * Get all test instances, ordered from outermost to innermost.
+	 */
+	List<Object> getAll();
+
+	/**
+	 * Find the test instance of the supplied required type.
+	 */
+	<T> Optional<T> find(Class<T> requiredType);
+
+}
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
@@ -37,7 +37,7 @@ public interface TestInstances {
 	 *
 	 * @return the innermost test instance; never {@code null}
 	 */
-	Object getInnermost();
+	Object getInnermostInstance();
 
 	/**
 	 * Get the enclosing test instances, excluding the innermost test instance,
@@ -46,7 +46,7 @@ public interface TestInstances {
 	 * @return the enclosing test instances; never {@code null} or containing
 	 * {@code null}, but potentially empty
 	 */
-	List<Object> getEnclosing();
+	List<Object> getEnclosingInstances();
 
 	/**
 	 * Get all test instances, ordered from outermost to innermost.
@@ -54,7 +54,7 @@ public interface TestInstances {
 	 * @return all test instances; never {@code null}, containing {@code null},
 	 * or empty
 	 */
-	List<Object> getAll();
+	List<Object> getAllInstances();
 
 	/**
 	 * Find the first test instance that is an instance of the supplied required
@@ -64,6 +64,6 @@ public interface TestInstances {
 	 * @return the first test instance of the required type; never {@code null}
 	 * but potentially empty
 	 */
-	<T> Optional<T> find(Class<T> requiredType);
+	<T> Optional<T> findInstance(Class<T> requiredType);
 
 }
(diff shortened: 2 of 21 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -160,7 +161,9 @@
 	 * @return an {@code Optional} containing the test instance; never
 	 * {@code null} but potentially empty
 	 * @see #getRequiredTestInstance()
+	 * @see #getTestInstances()
 	 */
+	@API(status = MAINTAINED, since = "5.4")
```
britter: I think this has been released prior to 5.4

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -173,12 +176,50 @@
 	 * @return the test instance; never {@code null}
 	 * @throws PreconditionViolationException if the test instance is not present
 	 * in this {@code ExtensionContext}
+	 *
+	 * @see #getRequiredTestInstances()
 	 */
+	@API(status = MAINTAINED, since = "5.4")
```
britter: I think this has been released prior to 5.4

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+@API(status = MAINTAINED, since = "5.4")
+public interface TestInstances {
+
+	/**
+	 * Get the innermost test instance.
+	 *
+	 * <p>The innermost instance is the one closest to the test method.
+	 */
+	Object getInnermost();
```
britter: We could also add `getOutermost()`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * Get the enclosing test instances, excluding the innermost test instance.
```
britter: In which order? Probably the same order as `getAll()`...

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	List<Object> getEnclosing();
+
+	/**
+	 * Get all test instances, ordered from outermost to innermost.
```
britter: How about introducing an ordering enum like `HierarchyTraversalMode`?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
```diff
@@ -0,0 +1,55 @@
...
+	 */
+	List<Object> getAll();
+
+	/**
+	 * Find the test instance of the supplied required type.
+	 */
+	Optional<Object> find(Class<?> requiredType);
```
britter: Is this for finding test classes implementing a certain interface or extending a certain base class?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -164,7 +166,7 @@ public JupiterEngineExecutionContext prepare(JupiterEngineExecutionContext conte
 
 		// @formatter:off
 		return context.extend()
-				.withTestInstanceProvider(testInstanceProvider(context, registry, extensionContext))
+				.withTestInstancesProvider(testInstanceProvider(context, registry, extensionContext))
```
britter:
```suggestion
				.withTestInstancesProvider(testInstancesProvider(context, registry, extensionContext))
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
```diff
@@ -246,40 +248,44 @@ private TestInstanceFactory resolveTestInstanceFactory(ExtensionRegistry registr
 		return null;
 	}
 
-	private TestInstanceProvider testInstanceProvider(JupiterEngineExecutionContext parentExecutionContext,
+	private TestInstancesProvider testInstanceProvider(JupiterEngineExecutionContext parentExecutionContext,
```
britter:
```suggestion
	private TestInstancesProvider testInstancesProvider(JupiterEngineExecutionContext parentExecutionContext,
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/DefaultTestInstances.java
```diff
@@ -0,0 +1,70 @@
...
+
+	private final List<Object> instances;
+
+	private DefaultTestInstances(List<Object> instances) {
+		this.instances = instances;
+	}
+
+	@Override
+	public Object getInnermost() {
+		return instances.get(instances.size() - 1);
```
britter: If you define `instances` as `LinkedList` you could use `LinkedList.getLast()`

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/DefaultTestInstances.java
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	@Override
+	public List<Object> getEnclosing() {
+		return instances.size() <= 1 ? Collections.emptyList() : instances.subList(0, instances.size() - 1);
+	}
+
+	@Override
+	public List<Object> getAll() {
+		return instances;
```
britter: Better return an unmodifiable list here?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -681,6 +682,8 @@ else if (methodPart.endsWith(")")) {
 	 * @return an {@code Optional} containing the outermost instance; never {@code null}
 	 * but potentially empty
 	 */
+	@API(status = DEPRECATED, since = "5.4")
+	@Deprecated
```
britter: Add pointer to the new API?

sbrannen: This is an internal API. So if we no longer use it, we can delete the "dead code".

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
sbrannen: indeed 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
sbrannen: indeed 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
sbrannen: According to the status quo in this proposal, it's an "instance of" check, so.... _both_. 😉

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/DefaultTestInstances.java
marcphilipp: It's already unmodifiable because the constructor is private and only ever called with unmodifiable lists.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
marcphilipp: Well, someone could mix an older version of an engine with a newer version of the Platform, e.g. when upgrading their build tool/IDE but not their engine dependency. Thus, I think we might want to keep it around for a while. Thoughts?

marcphilipp:
> However I wonder how this places together with the API suggested in #1611. For example I would expect that these methods here return instances of type `TestInstance`. So I wonder whether it is a good idea to fix this for 5.4 or whether we should better wait for #1611?

I think we should go ahead and add this for 5.4 since we'll likely follow the direction of your latest proposal on #1611 and introduce wrappers for `ExtensionContext` anyway.

marcphilipp:
The decision we need to discuss is what the API in `ExtensionContext` should look like. We already have `getTestInstance(): Optional<Object>`. So the simplest solution would be to add `getEnclosingInstances(): Optional<List<Object>>`. Since that looked a little odd to me, I introduced a `TestInstances` object instead so it was `getTestInstances(): Optional<TestInstances>` but that's totally up for debate.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
marcphilipp: I'm not sure if that would be useful. 🤔

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
jbduncan: If it is decided that this method should be kept, then I wonder if it could and should be changed to be something like the following, so that the test instance is already casted to the correct type when it's returned (if I've understood correctly what this method is meant to do):

```java
Optional<T> find(Class<T> requiredType);
```

@marcphillip Do you have any thoughts?

marcphilipp: @jbduncan Thanks for chiming in! That makes perfect sense and I changed it rightaway.

jbduncan: @marcphillip Brilliant! Thank you for responding so quickly, and I'm glad that I understood enough to raise a worthwhile suggestion. :)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
marcphilipp: Done in 16d4e70651.

marcphilipp:
I've rebased and squashed the branch and added the missing documentation.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
sormuras: It was marked **STABLE** in 5.0 and can't be reduced to **MAINTAINED** in 5.4.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
sormuras: Ditto. Should stay `STABLE`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
sbrannen: I agree with @sormuras: changing from `STABLE` to anything else within the 5.x timeline is a violation of the API rules.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -173,12 +176,50 @@
...
 		return Preconditions.notNull(getTestInstance().orElse(null),
 			"Illegal state: required test instance is not present in the current ExtensionContext");
 	}
 
+	/**
+	 * Get the test instances associated with the current test or container,
+	 * if available.
+	 *
+	 * <p>While top-level tests only have a single test instance, nested tests
+	 * have one instance per containing test class.
```
sbrannen: Let's refer to them a "enclosing classes" in line with `java.lang.Class.getEnclosingClass()`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -173,12 +176,50 @@
...
+	 *
+	 * @return an {@code Optional} containing the test instances; never
+	 * {@code null} but potentially empty
+	 * @see #getRequiredTestInstances()
+	 *
+	 * @since 5.4
+	 */
+	@API(status = MAINTAINED, since = "5.4")
```
sbrannen: Do we really want to go with `MAINTAINED` at this point? 

Perhaps `EXPERIMENTAL` would be better suited for this feature set.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.apiguardian.api.API.Status.MAINTAINED;
+
+import java.util.List;
+import java.util.Optional;
+
+import org.apiguardian.api.API;
+
+/**
+ * {@code ParameterContext} encapsulates the <em>test instances</em> of a test.
```
sbrannen: --> `TestInstances`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * <p>While top-level tests only have a single test instance, nested tests
+ * have one instance per containing test class.
```
sbrannen: containing --> enclosing

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @since 5.4
+ * @see ExtensionContext#getTestInstances()
+ * @see ExtensionContext#getRequiredTestInstances()
+ */
+@API(status = MAINTAINED, since = "5.4")
+public interface TestInstances {
```
sbrannen: Please add `@return` Javadoc tags for all methods declared in `TestInstances`, including immutability contracts if applicable.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/DefaultTestInstances.java
```diff
@@ -0,0 +1,70 @@
...
+
+	public static DefaultTestInstances of(TestInstances testInstances, Object instance) {
+		List<Object> allInstances = new ArrayList<>(testInstances.getAll());
+		allInstances.add(instance);
+		return new DefaultTestInstances(Collections.unmodifiableList(allInstances));
+	}
+
+	private final List<Object> instances;
+
+	private DefaultTestInstances(List<Object> instances) {
```
sbrannen: We should probably make use of `Preconditions` here just to be on the safe side.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/DefaultTestInstances.java
```diff
@@ -0,0 +1,70 @@
...
+		return instances.size() <= 1 ? Collections.emptyList() : instances.subList(0, instances.size() - 1);
+	}
+
+	@Override
+	public List<Object> getAll() {
+		return instances;
+	}
+
+	@Override
+	public <T> Optional<T> find(Class<T> requiredType) {
```
sbrannen: Missing `Preconditions.notNull(...)`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestInstances.java
sbrannen: Please add `@param` Javadoc tags for all methods declared in `TestInstances`, including nullability contracts if applicable.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -782,6 +782,8 @@ else if (methodPart.endsWith(")")) {
 	 * @return an {@code Optional} containing the outermost instance; never {@code null}
 	 * but potentially empty
 	 */
+	@API(status = DEPRECATED, since = "5.4")
```
sbrannen: 5.4 --> 1.4

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
marcphilipp: Agreed.

marcphilipp:
@sbrannen Thanks for the thorough review! I've addressed all of your feedback, please take another look and approve. 😉

sbrannen:
> @sbrannen Thanks for the thorough review! I've addressed all of your feedback, please take another look and approve. 😉

Thanks for making the changes. I _approved_ the PR. 👍

sbrannen:
As an afterthought...

I admittedly find it a bit odd having methods named something as simple/generic as `getAll()` and `getEnclosing()` in this new API.

I suppose it might be OK within the context of the API but still rather generic (and potentially misleading) in terms of UX.

For example, when I see `getEnclosing()` I assume it returns a single thing, not a list of things. This singularity is also misleading when juxtaposed with `getInnermost()` (which does return a single thing).

What about `getEnclosingInstances()` (and possibly `getAllInstances()`)?

sormuras:
I favor the more explicit and context-agnostig longer versions that Sam proposed.

PR #1653, opened 2018-10-28
Introduce Try type to preserve failure cause

### LINKED ISSUE #602: Provide additional context in error messages
## Overview

As requested I am opening a new issue on this. See https://github.com/junit-team/junit5/issues/586#issuecomment-266522369 for background.

I'm new with JUnit, and my early attempts have been frustrated by messages such as: 
`PreconditionViolationException: Could not load class with name: DirectoryTest`

Well, WHY couldn't you load the class??

I'm left trying to guess what the problem could be.

It seems to me that a lot could be done to make error messages actually helpful. 

## Related Issues

- #1666

## Deliverables

- [ ] Where possible, provide additional context in error messages or log exceptions (e.g., in `ReflectionUtils`).

marcphilipp:
Exceptions reported due to failed reflective operations such as loading
a class, reading a field's value, or looking up a method by name now
include the original exception as their cause to make it easier to debug
underlying issues.

Such operations in `ReflectionUtils` and `ReflectionSupport` no longer
return an instance of `Optional` but one of a new `Try` type which
encapsulates either a successful invocation with a (nullable) resulting
value or a failure with an exception as its cause. Depending on the
caller, the cause can then be logged or rethrown as appropriate.

Resolves #602.

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/function/Try.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/function/Try.java
@@ -104,27 +104,27 @@ public abstract class Try<V> {
 	/**
 	 * If this {@code Try} is a success, apply the supplied transformer to its
 	 * value and return a new successful or failed {@code Try} depending on the
-	 * action's outcome; if it's a failure, do nothing.
+	 * transformer's outcome; if this {@code Try} is a failure, do nothing.
 	 *
-	 * @param action the action to try; must not be {@code null}
+	 * @param transformer the transformer to try; must not be {@code null}
 	 * @return a succeeded or failed {@code Try}; never {@code null}
 	 */
-	public abstract <U> Try<U> andThenTry(Transformer<V, U> action);
+	public abstract <U> Try<U> andThenTry(Transformer<V, U> transformer);
 
 	/**
-	 * If this {@code Try} is a success, apply the supplied action to its value
-	 * and return a new successful or failed {@code Try} depending on the
-	 * action's outcome; if it's a failure, do nothing.
+	 * If this {@code Try} is a success, apply the supplied function to its
+	 * value and return the resulting {@code Try}; if this {@code Try} is a
+	 * failure, do nothing.
 	 *
-	 * @param action the action to apply; must not be {@code null}
+	 * @param function the function to apply; must not be {@code null}
 	 * @return a succeeded or failed {@code Try}; never {@code null}
 	 */
-	public abstract <U> Try<U> andThen(Function<V, Try<U>> action);
+	public abstract <U> Try<U> andThen(Function<V, Try<U>> function);
 
 	/**
 	 * If this {@code Try} is a failure, call the supplied action and return a
 	 * new successful or failed {@code Try} depending on the action's outcome;
-	 * if it's a success, do nothing.
+	 * if this {@code Try} is a success, do nothing.
 	 *
 	 * @param action the action to try; must not be {@code null}
 	 * @return a succeeded or failed {@code Try}; never {@code null}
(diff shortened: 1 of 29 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -497,18 +538,34 @@ static boolean isWideningConversion(Class<?> sourceType, Class<?> targetType) {
...
+	 * <p>If an exception occurs while reading the field or if the value of the
+	 * field is {@code null}, a failed {@link Try} is returned that contains the
+	 * corresponding exception.
+	 *
+	 * @param field the field to read; never {@code null}
+	 * @param instance the instance from which the value is to be read; may
+	 * be {@code null} for a static field
+	 * @see #readFieldValue(Field)
+	 * @see #readFieldValue(Class, String, Object)
+	 * @since 1.3
```
sbrannen:
```suggestion
	 * @since 1.4
```

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -481,9 +502,29 @@ static boolean isWideningConversion(Class<?> sourceType, Class<?> targetType) {
...
+	 * Try to read the value of a potentially inaccessible static field.
+	 *
+	 * <p>If an exception occurs while reading the field or if the value of the
+	 * field is {@code null}, an failed {@link Try} is returned that contains
+	 * the corresponding exception.
+	 *
+	 * @param field the field to read; never {@code null}
+	 * @see #readFieldValue(Field, Object)
+	 * @see #readFieldValue(Class, String, Object)
+	 * @since 1.3
```
sbrannen:
```suggestion
	 * @since 1.4
```

smoyer64:
This is probably the most specialization of the Either pattern I've ever used - I created one I called Possible but it's not nearly so fleshed out as this one.  I consider this pattern essential for functional programming and I'm going to be very tempted to include ``junit-platform-commons`` as a compile-scoped dependency.  It's a shame they didn't include this in Java to wean people away from using ``Exceptions``s as flow-control.

platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
```diff
@@ -368,99 +408,102 @@ void invokePrivateStaticMethod() throws Exception {
 	}
 
 	@Test
-	void loadClassPreconditions() {
```
sbrannen: I might have overlooked something, but it appears that you have removed tests for deprecated utility methods. If that's the case, I'd recommend leaving such tests in place until the deprecated code has been officially removed.

junit-platform-commons/src/main/java/org/junit/platform/commons/function/Try.java
```diff
@@ -0,0 +1,360 @@
...
+	/**
+	 * If this {@code Try} is a success, get the contained value; if it's a
+	 * failure, call the supplied {@link Function} with the contained exception
+	 * and throw the resulting {@link Exception}.
+	 *
+	 * @param exceptionTransformer the transformer to be called with the
+	 * contained exception, if available; must not be {@code null}
+	 * @return the contained value, if available
+	 */
+	public abstract <E extends Exception> V getOrThrow(Function<? super Exception, E> exceptionTransformer) throws E;
```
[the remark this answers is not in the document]
marcphilipp: Because creating a custom exception instance should not throw an exception.

platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
marcphilipp: Unless I have overlooked something, I've kept around a sufficient amount of tests for each deprecated method to make sure they delegate to their `try...` variant correctly but have deleted any surplus tests for special cases already tested for the `try...` variant. Makes sense?

marcphilipp:
Thanks for the thorough review, @sbrannen! I have followed most of your suggestions. The only left to discuss is `Exception` vs. `Throwable`.

marcphilipp:
@smoyer64 I could publish the type as a micro library in case you're interested.

smoyer64:
@marcphilipp I'd definitely be interested in that ... I've been looking for a very small functional library that has some of the additions available in Scala, Kotlin, etc.  I originally started with Java 7 using [FunctionalJava](https://www.functionaljava.org/docs.html) but so much of that was superceded in Java 8 that I couldn't justify it simply for some of the support  structures.  If the proposed micro-library includes just a few of these (I mentioned Either above) it would be great.  And I'll certainly sign on as a developer.

Comparing ``Try`` with [Possible](https://github.com/selesy/hibp-client/blob/b93a24a94aca9ddfdfee92521e8693db133c4e48/hibp-common/src/main/java/com/selesy/hibp/Possible.java), I think they're pretty compatible.  The main (functional) difference is the static methods that create the object.  I actually like your use of ``Callable`` ... it's something I hadn't thought of but works better in some of my use cases.  I'd like to suggest that it would be worth adding the ``of`` methods ([here](https://github.com/selesy/hibp-client/blob/b93a24a94aca9ddfdfee92521e8693db133c4e48/hibp-common/src/main/java/com/selesy/hibp/Possible.java#L37) and [here](https://github.com/selesy/hibp-client/blob/b93a24a94aca9ddfdfee92521e8693db133c4e48/hibp-common/src/main/java/com/selesy/hibp/Possible.java#L48)) that I originally created as they allow the creation of a object when you've already executed the method that's potentially throwing the ``Exception`` (as when some exceptions are ignored).

jbduncan:
> @smoyer64 I could publish the type as a micro library in case you're interested.

I'd be interested in this as well, actually!

I've been interested in Vavr's `Try` (and some other bits and bobs, but I can't remember what now) for a long time, but I couldn't justify to myself using a functional collections library at the same time (or even some other functional bits which I'm not familiar with yet). So 👍 to a new micro-library.

sbrannen:
> So 👍 to a new micro-library.

Sounds like @marcphilipp might soon earn the title "micro brewer". 😛

smoyer64:
@sbrannen - you mean like this:

micro-brewer

sbrannen:
See also: #1666

junit-platform-commons/src/main/java/org/junit/platform/commons/function/Try.java
```diff
@@ -0,0 +1,374 @@
...
+ *
+ * <p>Moreover, it makes it particularly convenient to attach follow-up actions
+ * should the {@code Try} have been successful (cf. {@link #andThen} and
+ * {@link #andThenTry}) or fallback actions should it not have been (cf.
+ * {@link #orElse} and {@link #orElseTry}).
+ *
+ * @since 1.4
+ */
+@API(status = MAINTAINED, since = "1.4")
+public abstract class Try<V> {
```
leonard84: Is this inspired by http://www.vavr.io/vavr-docs/#_try ?

platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
sbrannen: Yep, if that's the case... that makes sense.

marcphilipp:
Merged manually to master in 19eb6d62...0e0f2b95.

PR #1677, opened 2018-11-12
AnnotatedElement parameter support in declaratively registered extensions' constructors

inponomarev:
## Overview

Configuring extensions programmatically in `@RegisterExtension`-annotated static fields is fine. But sometimes we need static configuration for declaratively registered extensions, i. e. we want to initialize private final fields in `@ExtendWith`-defined extension classes.

The problem is that declaratively registered extensions are instantiated using default constructors. The proposed change adds support for constructors that accept `AnnotatedElement` as the only parameter. In this case we can extract all the information needed from other annotations on the element.

E. g. we could declare something like (see documentation/src/test/java/example/registration/ParameterizedExtensionDemo.java):

```java
@ExtendWith(ParameterizedExtension.class)
public @interface WithParameterizedExtension {
    String value();
}

public class ParameterizedExtension implements ParameterResolver {
    private final String parameter;
    public ParameterizedExtension(AnnotatedElement annotatedElement) {
        parameter = annotatedElement.getAnnotation(WithParameterizedExtension.class).value();
    }
....
```
And use this extension in the following manner:

```java
@WithParameterizedExtension("foo")
public class ParameterizedExtensionDemo {

    @Test
    void parameterIsPassedToExtension(String param) {
        assertEquals("foo", param);
    }
}
```

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/extensions.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/extensions.adoc
@@ -74,6 +74,15 @@ in which they are declared in the source code. For example, the execution of tes
 both `MyFirstTests` and `MySecondTests` will be extended by the `FooExtension` and
 `BarExtension`, **in exactly that order**.
 
+===== Configuring declaratively registered extension using annotations
+
+During registration process, extension is instantiated via its default constructor.
+
+However, if extension has a constructor with `AnnotatedElement` parameter, it will be called
+and the respective `@ExtendWith`-annotated element will be passed to it.
+
+This allows to extract other annotations on the element and configure the extension.
+
 [[extensions-registration-programmatic]]
 ==== Programmatic Extension Registration
 
--- a/documentation/src/test/java/example/registration/ParameterizedExtension.java
+++ b/documentation/src/test/java/example/registration/ParameterizedExtension.java
@@ -1,27 +1,39 @@
+/*
+ * Copyright 2015-2018 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * http://www.eclipse.org/legal/epl-v20.html
+ */
+
 package example.registration;
+
+import java.lang.reflect.AnnotatedElement;
 
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.api.extension.ParameterResolutionException;
 import org.junit.jupiter.api.extension.ParameterResolver;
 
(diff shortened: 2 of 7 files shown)


inponomarev:
> Extension implementations have access to Configuration Parameters

That's great, and it would be nice if I was able to declare an extension **constructor** with `ExtensionContext` as a parameter, because I need to populate some immutable extension fields. 

To write such a constructor is what I tried to do in the first place. And then I found out that JUnit runs only the default constructor. `ExtensionContext` as a constructor parameter would certainly be a better choice, but I didn't find the valid way to build the correct `ExtensionContext` there. Maybe this is what needs to be changed here. 

> Annotations provide a very limited way of valid return types .. primitives, String, Class, enums, annotations, and arrays 

...and that's what perfectly fits the vast majority of the use cases. When I write an extension for my library/framework, I only demand the user to provide a path to resources for my extension to operate with. This what annotation return types are good for. It is the task of extension, not user, to deserialize / instantiate everything.

marcphilipp:
@inponomarev `Extension` classes should not have state, i.e. `private` `final` fields. Instead, you can read the configuration from your custom annotations and store it in the `ExtensionContext` on demand, e.g. using `Store.getOrComputeIfAbsent()`.

inponomarev:
Thanks @marcphilipp, I got your point about `Store.getOrComputeIfAbsent()` and agree with the point that `Extension`s should not have state.  Still, **final** fields of immutable objects are not state (i. e. something that can be morphed) :-) 

What about `ExtensionContext` as a possible `Extension`'s constructor parameter? Still disagree?

marcphilipp:
I would still consider that state because an extension may be registered globally without access to an `AnnotatedElement`. In general, extensions should not make assumptions about where they are registered so they can support all possible registration mechanisms (global, declarative, or programmatic).

Thus, in your example, I'd search the annotated elements upward (using `ExtensionContext.getElement()` and `ExtensionContext.getParent()`) from within the extension and use the closest configuration annotation I can find. Does that make sense?

inponomarev:
@marcphilipp 👍 👍  thank you for your suggestion! I was able to achieve what I desired using recursive search with `getElement`/`getParent` and  `Store.getOrComputeIfAbsent()`, exactly as you told.

One final remark. I was misled by the following part of User Guide:

> when an extension is registered via @RegisterExtension, it can be configured programmatically — for example, in order _**to pass arguments to the extension’s constructor**_, a static factory method, or a builder API.

Maybe this part should be rewritten to avoid mentioning the configuration of extensions this way.

marcphilipp:
I think it's ok for extension authors to decide that they'll only support `@RegisterExtension`. Alternatively, they could decide to support constructor parameters but also provide a no-args constructor for declarative registration and lookup the configuration from annotations. You explicitly asked for configuration via annotations, though. So, please take what I said above with a grain of salt. 😉

PR #1718, opened 2019-01-02
Introduce Launcher.execute(TestPlan)

### LINKED ISSUE #1695: Add Launcher execute() method that accepts the TestPlan returned by discover()
## Overview

IDEs and build tools use the `org.junit.platform.launcher.Launcher` interface to discover and run tests. However, the interface does not provide a means to run rests with a previous discovered test plan. This results, at least for IntelliJ, in that test engines will be asked twice to come up with the same test plan. See: https://youtrack.jetbrains.com/issue/IDEA-203173 for details.

Since discovery often involves classpath scanning this can be a real performance hit in large projects.

## Proposal

Add interface method `void execute(TestPlan)` to the interface to allow reuse of already discovered test plan.

### Alternative

If the suggestion above is difficult to realize. Cache a discovered test plan that results from an identical `LaunchDiscoveryRequest` object.

## Deliverables

- [ ] Introduce a new `void execute(TestPlan)` method in the `Launcher` API.
  - [ ] Implement appropriate tests.
- [ ] Add checks to `TestPlan` that throw an exception if the _user_ attempts to modify the state of the `TestPlan` after discovery has completed.
  - [ ] For the above, we will need some form of _state_ within the `TestPlan` to track if discovery has completed.
  - [ ] Implement appropriate tests.
- [ ] Document in User Guide.
- [ ] Document in Release Notes.

marcphilipp:
This commit adds a new overload for `Launcher.execute()` that accepts a
`TestPlan` directly thus avoiding running test discovery twice when
`discover()` is called prior to `execute()`. The `JUnitPlatform` runner
is changed to make use of this new feature.

Resolves #1695.

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
@@ -64,20 +64,14 @@ public interface Launcher {
 	 * {@link LauncherDiscoveryRequest} by querying all registered engines and
 	 * collecting their results.
 	 *
-	 * @apiNote This method should only be called to generate a preview of the
-	 * test tree when executing tests is not desired. First calling this method
-	 * to get access to the {@link TestPlan} and then {@link #execute} causes
-	 * test discovery to be executed twice which may result in a significant
-	 * performance degradation. Instead, {@link #execute} should be called
-	 * directly and a {@link TestExecutionListener} that overrides
-	 * {@link TestExecutionListener#testPlanExecutionStarted(TestPlan)} should
-	 * be registered to get access to the test tree, for example to render it in
-	 * an IDE.
+	 * @apiNote This method may be called to generate a preview of the test
+	 * tree. The resulting {@link TestPlan} is unmodifiable and may be passed to
+	 * {@link #execute(TestPlan, TestExecutionListener...)} for execution.
 	 *
 	 * @param launcherDiscoveryRequest the launcher discovery request; never
 	 * {@code null}
-	 * @return a {@code TestPlan} that contains all resolved {@linkplain
-	 * TestIdentifier identifiers} from all registered engines
+	 * @return an unmodifiable {@code TestPlan} that contains all resolved
+	 * {@linkplain TestIdentifier identifiers} from all registered engines
 	 */
 	TestPlan discover(LauncherDiscoveryRequest launcherDiscoveryRequest);
 
@@ -91,14 +85,33 @@ public interface Launcher {
 	 * <p>Supplied test execution listeners are registered in addition to already
 	 * registered listeners but only for the supplied launcher discovery request.
 	 *
-	 * @apiNote In order to get access to the test tree, for example to render
-	 * it in an IDE, register a {@link TestExecutionListener} that overrides
-	 * {@link TestExecutionListener#testPlanExecutionStarted(TestPlan)} instead
-	 * of calling {@link #discover} first.
+	 * @apiNote Calling this method will cause test discovery to be executed for
+	 * all registered engines. If the same {@link LauncherDiscoveryRequest} was
+	 * previously passed to {@link #discover(LauncherDiscoveryRequest)}, you
+	 * should instead call {@link #execute(TestPlan, TestExecutionListener...)}
+	 * and pass the already acquired {@link TestPlan} to avoid the potential
+	 * performance degradation (e.g., classpath scanning) of running test
+	 * discovery twice.
 	 *
 	 * @param launcherDiscoveryRequest the launcher discovery request; never {@code null}
 	 * @param listeners additional test execution listeners; never {@code null}
 	 */
 	void execute(LauncherDiscoveryRequest launcherDiscoveryRequest, TestExecutionListener... listeners);
+
+	/**
+	 * Execute the supplied {@link TestPlan} and notify
+	 * {@linkplain #registerTestExecutionListeners registered listeners} about
+	 * the progress and results of the execution.
+	 *
+	 * <p>Supplied test execution listeners are registered in addition to
+	 * already registered listeners but only for the execution of the supplied
+	 * test plan.
+	 *
+	 * @param testPlan the test plan to execute; never {@code null}
+	 * @param listeners additional test execution listeners; never {@code null}
+	 * @since 1.4
+	 */
+	@API(status = STABLE, since = "1.4")
+	void execute(TestPlan testPlan, TestExecutionListener... listeners);
 
 }
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestPlan.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestPlan.java
@@ -52,7 +52,7 @@ import org.junit.platform.engine.TestDescriptor.Visitor;
  * @see TestExecutionListener
  */
 @API(status = STABLE, since = "1.0")
-public final class TestPlan {
+public class TestPlan {
 
 	private final Set<TestIdentifier> roots = Collections.synchronizedSet(new LinkedHashSet<>(4));
 
@@ -82,7 +82,8 @@ public final class TestPlan {
 		return testPlan;
 	}
 
-	private TestPlan(boolean containsTests) {
+	@API(status = INTERNAL, since = "1.4")
+	protected TestPlan(boolean containsTests) {
 		this.containsTests = containsTests;
 	}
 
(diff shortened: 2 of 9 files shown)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -117,23 +117,32 @@ public void registerTestExecutionListeners(TestExecutionListener... listeners) {
...
 		Preconditions.notNull(listeners, "TestExecutionListener array must not be null");
 		Preconditions.containsNoNullElements(listeners, "individual listeners must not be null");
-		execute(discoverRoot(discoveryRequest, "execution"), discoveryRequest.getConfigurationParameters(), listeners);
+		execute(UnmodifiableTestPlan.from(discoverRoot(discoveryRequest, "execution")), listeners);
+	}
+
+	@Override
+	public void execute(TestPlan testPlan, TestExecutionListener... listeners) {
+		Preconditions.notNull(testPlan, "TestPlan must not be null");
+		Preconditions.condition(testPlan instanceof UnmodifiableTestPlan, "TestPlan was not returned by this Launcher");
```
mmerdes: Did you consider to check the identity of the test plan somehow (e.g. by some fingerprint)?
I guess a modified UnmodifiableTestPlan could somehow be engineered, maybe by serialization tricks...

documentation/src/docs/asciidoc/release-notes/release-notes-5.4.0-M2.adoc
```diff
@@ -24,11 +24,14 @@ repository on GitHub.
 
 ==== Deprecations and Breaking Changes
 
-* ❓
+* `TestPlan.add(TestIdentifier)` was always considered internal but was missing
+  a corresponding `@API` annotation.
```
sbrannen: `TestPlan.add(TestIdentifier)` has always been considered _internal_ and is now appropriately annotated with `@API(status = INTERNAL)`.

documentation/src/docs/asciidoc/release-notes/release-notes-5.4.0-M2.adoc
```diff
@@ -24,11 +24,14 @@ repository on GitHub.
...
 
 ==== New Features and Improvements
 
-* ❓
+* `Launcher` now provides an `execute(TestPlan, TestExecutionListener...)`
```
sbrannen: The `Launcher` API now ...

documentation/src/docs/asciidoc/release-notes/release-notes-5.4.0-M2.adoc
```diff
@@ -24,11 +24,14 @@ repository on GitHub.
...
+  method that allows to execute a previously discovered `TestPlan`.
```
sbrannen: allows to --> allows one to ...

junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
```diff
@@ -91,14 +85,32 @@
...
-	 * @apiNote In order to get access to the test tree, for example to render
-	 * it in an IDE, register a {@link TestExecutionListener} that overrides
-	 * {@link TestExecutionListener#testPlanExecutionStarted(TestPlan)} instead
-	 * of calling {@link #discover} first.
+	 * @apiNote Calling this method will cause test discovery to be executed for
+	 * all registered engines. If the same {@link LauncherDiscoveryRequest} was
+	 * previously passed to {@link #discover(LauncherDiscoveryRequest)}, you
+	 * should instead call {@link #execute(TestPlan, TestExecutionListener...)}
+	 * and pass the already acquired {@link TestPlan} to avoid the potential
+	 * performance degradation (e.g. classpatch scanning) of running test
```
sbrannen:
```suggestion
	 * performance degradation (e.g., classpath scanning) of running test
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/Launcher.java
```diff
@@ -91,14 +85,32 @@
...
 	 */
 	void execute(LauncherDiscoveryRequest launcherDiscoveryRequest, TestExecutionListener... listeners);
 
+	/**
+	 * Execute the supplied {@link TestPlan} and notify
+	 * {@linkplain #registerTestExecutionListeners registered listeners} about
+	 * the progress and results of the execution.
+	 *
+	 * <p>Supplied test execution listeners are registered in addition to already
+	 * registered listeners but only for the supplied launcher discovery request.
```
sbrannen: Please reword this sentence, since this method does not accept a "request".

marcphilipp:
@sbrannen PTAL!

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestPlan.java
```diff
@@ -52,7 +52,7 @@
  * @see TestExecutionListener
  */
 @API(status = STABLE, since = "1.0")
-public final class TestPlan {
+public class TestPlan {
```
jbduncan: @marcphilipp Have you considered making `UnmodifiableTest` it's own separate class that just delegates to a `TestPlan` rather than extending it, so that `TestPlan` can be kept final?

sbrannen: How would that help? `UnmodifiableTestPlan` must be an "instance of" `TestPlan` in order for JUnit to return it from `org.junit.platform.launcher.Launcher.discover(LauncherDiscoveryRequest)`.

sbrannen:
> @sbrannen PTAL!

will do.....

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestPlan.java
```diff
@@ -82,7 +82,7 @@ public static TestPlan from(Collection<TestDescriptor> engineDescriptors) {
 		return testPlan;
 	}
 
-	private TestPlan(boolean containsTests) {
+	protected TestPlan(boolean containsTests) {
```
sbrannen: Although it would technically not prevent subclassing, it's perhaps a good idea to document this (now protected) constructor as `@API(status = INTERNAL ...)`.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -175,11 +184,12 @@ private Root discoverRoot(LauncherDiscoveryRequest discoveryRequest, String phas
...
 
-	private void execute(Root root, ConfigurationParameters configurationParameters,
-			TestExecutionListener... listeners) {
+	private void execute(UnmodifiableTestPlan unmodifiableTestPlan, TestExecutionListener[] listeners) {
+		Root root = unmodifiableTestPlan.getRoot();
+		ConfigurationParameters configurationParameters = root.getConfigurationParameters();
 		TestExecutionListenerRegistry listenerRegistry = buildListenerRegistryForExecution(listeners);
 		withInterceptedStreams(configurationParameters, listenerRegistry, testExecutionListener -> {
-			TestPlan testPlan = TestPlan.from(root.getEngineDescriptors());
+			TestPlan testPlan = unmodifiableTestPlan.getDelegate();
```
sbrannen: Is it necessary to work with the "modifiable" `TestPlan` from this point on, or would it suffice to work with the `UnmodifiableTestPlan` during the execution phase?

Now that I think about, wouldn't it be reasonable to disallow modification of the `TestPlan` by a registered `TestExecutionListener`?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestPlan.java
jbduncan: Good question @sbrannen! I admit I didn't have this PR checked out on my computer when I wrote my suggestion, so I missed that `Launcher.discover(LauncherDiscoveryRequest)` needed to return a plain `TestPlan`. Apologies! :bowing_man: 

In that case, I don't see anyway of allowing `TestPlan` to be kept final, apart from giving `TestPlan` a constructor that accepts a parameter which determines if the `TestPlan` is modifiable or not, but it's not clear to me that that's any cleaner than the current solution.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
marcphilipp: Makes sense. Changed in my latest commit.

Issue #1767, opened 2019-02-09
Feature Request: Add @Depend(String[] methods) annotation to give 'methods' higher priority than the annotated method


duckladydinh:
Hi everyone,

I am creating a pull request for this feature. It should work similar to TestNG (@Test dependsOn) but it will not disable the test if some test fails. It just gives methods a priority like the @Order.

It would be great if someone can give me some advice how to improve it :). Thank you.

:)

sbrannen:
Tentatively slated for 5.5 M1 for the purpose of _team discussion_.

sbrannen:
Hi @duckladydinh,

Thanks for opening the issue and submitting the PR!

> I am creating a pull request for this feature. It should work similar to TestNG (`@Test(dependsOn = ...)`) but it will not disable the test if some test fails. It just gives methods a priority like `@Order`.

Although I think there is merit in mere "dependent ordering" as in the PR, I think it would be worth investigating the possibility for "not executing" methods if a method that they depend on fails.

For example, if `C` depends on `A` and `A` fails, then `C` should be _aborted_. That's basically what TestNG does, and I think it might be possible to implement this elegantly in JUnit Jupiter by implementing the `TestWatcher` API introduced in JUnit Jupiter 5.4. Of course, I have not tried to implement it myself, so I'm just thinking out loud here.

duckladydinh:
Hi @sbrannen , thank you for the suggestion. I think I have something done already based on your suggestion. My intention is to keep the DependsOnAnnotation as it is, and add a separate TestWatcher. My initial test succeeds but I think I needs a bit more testing to ensure it is stable :).

thespags:
Hi is there any update on when this PR might be merged in. There hasn't been any follow up since end of last year. Thank you.

spags-twilio:
Hi. I wanted to check in if there is any update or concerns? This is one of the features available in TestNG that I miss when working in JUnit5. It helps isolate problems in tests when you have good dependency management.

ahoehma:
Any news here?

sbrannen:
@ahoehma, this issue still has the `waiting-for-interest` label.

jzhang-SD:
This feature would be very useful

thespags:
I am still interested.

ericparton:
I'm not sure what happened here, but this would be a useful feature to have. It's something that TestNG can do, and the lack of support makes migrating from TestNG to JUnit difficult.

armlesshobo:
[...]

Issue #111, opened 2016-01-13
Specifying the order of tests in test reports


luontola:
## Motivation

With testing frameworks which consider test names as documentation/specification, the order of test results matters. IDEs and CIs should report all tests in the order they were written (not necessarily same as their execution order), especially if they are used for generating documentation.

For example see [Specsy's Fibonacci example](http://specsy.org/documentation.html#naming-tests) which has two tests ("The first two Fibonacci numbers are 0 and 1" and "Each remaining number is the sum of the previous two") whose names make sense only when read in the order they were written. This is also required by Cucumber and similar tools (assuming Cucumber features, scenarios and steps are represented as nested tests).
## Current state

AFAIK, currently JUnit 5 does not maintain the order of tests, because methods like `TestPlan.getChildren` return a `Set`. I hope that the test reporting order will be made a first-class citizen.
## Ideas for a solution

JUnit 4 maintained the test order by returning a `List` from `Description.getChildren`, but relying on a mutable class has its problems, and I'm not sure whether maintaining test order was a design goal or just an implementation detail in JUnit 4.

[Jumi](http://jumi.fi/) uses the [fi.jumi.api.drivers.TestId](https://github.com/orfjackal/jumi/blob/v0.5.437/jumi-api/src/main/java/fi/jumi/api/drivers/TestId.java) class to specify both test order and test identity using a "path", which is a list of integers.

A solution which I think would work for JUnit 5 is to combine its string-based `org.junit.gen5.launcher.TestId` with information about the order of tests. If each part of the unique ID (see #106) would have both `String name` and `int order`, and the unique ID would be a list of those pairs, then they would be easy to sort predictably: order firstly by `part.order`, secondly by `part.name`, thirdly by the next part. The `order` may be the same for all parts at the same level, in which case the `name` will determine the order (for example packages should be sorted like this). It would need to be discussed whether the order should affect the ID's equality; I have arguments against and for it.

This issue relates to #48 (scenario tests), but not directly to #13 (runtime order). My proposed solution relates to #106 and #68 (unique ID).

marcphilipp:
Since JDK 7 the order of methods as returned by the JDK's reflection API is not deterministic anymore. Therefore, to guarantee/specify a certain order additional work has to be done.

[...]

Issue #48, opened 2015-12-03
Introduce first-class support for scenario tests


sbrannen:
The JUnit Lambda prototype supported two options for the lifecycle of a test instance that could be configured via the `@TestInstance` annotation:

1. `PER_METHOD`: the default; semantically identical to JUnit 4
1. `PER_CLASS`: a.k.a., _scenario testing_ mode; analogous to TestNG

Support for `@TestInstance` has, however, been removed from `master` and will be replaced with equivalent functionality configured via dedicated annotations that make the runtime behavior of the test more apparent.

The current proposal is to introduce the following:

- `@ScenarioTest`: class-level annotation used to denote that a test class contains steps that make up a single scenario test.
- `@Step`: method-level annotation used to denote that a test method is a single _step_ within the scenario test.

The `@Step` annotation will need to provide an attribute that can be used to declare the next step within the scenario test. Steps will then be ordered according to the resulting dependency graph and executed in exactly that order. 

For example, a scenario test could look similar to the following:

```java
@ScenarioTest
class WebSecurityScenarioTest {

    @Step(next = "login")
    void visitPageRequiringAuthorizationWhileNotLoggedIn() {
        // attempt to visit page which requires that a user is logged in
        // assert user is redirected to login page
    }

    @Step(next = "visitSecondPageRequiringAuthorizationWhileLoggedIn")
    void login() {
        // submit login form with valid credentials
        // assert user is redirected back to previous page requiring authorization
    }

    @Step(next = "logout")
    void visitSecondPageRequiringAuthorizationWhileLoggedIn() {
        // visit another page which requires that a user is logged in
        // assert user can access page
    }

    @Step(next = END)
    void logout() {
        // visit logout URL
        // assert user has been logged out
    }

}
```

mmerdes:
@sbrannen do you prefer referencing the next step by name?  i assume these references would be refactoring-safe, so no real problem here. what about typos? i guess just using numbered steps will be awkward if you introduce steps in the middle...

jlink:
Is it in scope for Alpha1 (end of next week)?

bechte:
Thanks for writing down the concept and providing a detailed example. I think it is great and I also like the next attribute pointing to the methodName. Two reasons:

* IDE refactoring of the method name should also replace the next attribute string
* IDE vendors will have a nice and easy way to support auto-completion for steps

[...]

Issue #18, opened 2015-11-26
Introduce support for declaring dependencies between test methods


nemecec:
See https://github.com/junit-team/junit.contrib/tree/master/assumes
I hope the new JUnit add this in the core. Or, at least makes it easy to built it as an extension.

marcphilipp:
Technically this issue was reported first, but #1767 has more discussion so I'm closing it as _duplicate_.

PR #1768, opened 2019-02-09
Introduce @DependsOn annotation for method ordering

duckladydinh:
### Introduction

Hi,

I want to implement the feature as I suggested in #1767: A mechanism for scenario testing. I really want to make a contribution to open source project. If there is something not good, it would be great if you can give me some advice :). 

Thank you.

---

###### You can add the following annotations to your test class.

**_@TestMethodOrder(DependsOnAnnotation.class)_** - Most important: It sorts the test methods according to the dependent order, but this alone does not disable child method if any ancestor method fails.

**_@ExtendWith(DependsOnTestWatcher.class)_** - Second most important: It will disable child method if any ancestor method fails.

###### You can add the following annotation to your test method.

**_@DependsOn("ancestorMethodName")_** - Most important: It will provide the relative relationship that you want to realize.

---

---

### Definition of Done


sbrannen:
Tentatively slated for 5.5 M1 for the purpose of _team discussion_.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Depend.java
```diff
@@ -0,0 +1,48 @@
...
+ * <em>test methods</em>.
+ *
+ * @since 5.5.0-SNAPSHOT
+ * @see MethodOrderer.DependAnnotation
+ */
+@Target({ ElementType.FIELD, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@API(status = EXPERIMENTAL, since = "5.4")
+public @interface Depend {
```
sbrannen: Please rename to `DependsOn`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Depend.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * {@code @Depend} is an annotation that is used to configure the
+ * {@linkplain #methods order} in which the annotated element (i.e., field or
+ * method) should be evaluated or executed after given methods.
+ *
+ * <p>When used with the
+ * {@link MethodOrderer.DependAnnotation} {@link MethodOrderer}, the category applies to
+ * <em>test methods</em>.
+ *
+ * @since 5.5.0-SNAPSHOT
```
sbrannen:
```suggestion
 * @since 5.5
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Depend.java
```diff
@@ -0,0 +1,48 @@
...
+ * @see MethodOrderer.DependAnnotation
+ */
+@Target({ ElementType.FIELD, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@API(status = EXPERIMENTAL, since = "5.4")
```
sbrannen:
```suggestion
@API(status = EXPERIMENTAL, since = "5.5")
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Depend.java
sbrannen:
```suggestion
public @interface DependsOn {
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Depend.java
```diff
@@ -0,0 +1,48 @@
...
+public @interface Depend {
+
+	/**
+	 * The methods that must be executed before the annotated method.
+	 *
+	 * <p>Elements are ordered based on priority where methods whose names are
+	 * provided have higher priority than the annotated method.
+	 */
+	String[] methods();
```
sbrannen:
```suggestion
	String[] value();
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -172,6 +175,76 @@ private static int getOrder(MethodDescriptor descriptor) {
 		}
 	}
 
+	/**
+	 * {@code MethodOrderer} that sorts methods based on the {@link Depend @Depend}
+	 * annotation.
+	 */
+	class DependAnnotation implements MethodOrderer {
```
sbrannen:
```suggestion
	class DependsOnAnnotation implements MethodOrderer {
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -172,6 +175,76 @@ private static int getOrder(MethodDescriptor descriptor) {
...
+		private static final Logger logger = LoggerFactory.getLogger(Random.class);
+
+		private static Map<String, Integer> dependencySize;
+		private static Map<String, String[]> digraph;
```
sbrannen: A `MethodOrderer` must not have `static` state. Please redesign this.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -172,6 +175,76 @@ private static int getOrder(MethodDescriptor descriptor) {
...
+
+		@Override
+		public void orderMethods(MethodOrdererContext context) {
+			digraph = context.getMethodDescriptors().stream().filter(
+				descriptor -> descriptor.findAnnotation(Depend.class).isPresent()).collect(
+					Collectors.toMap(descriptor -> descriptor.getMethod().getName(),
+						descriptor -> descriptor.findAnnotation(Depend.class).map(Depend::methods).get()));
```
sbrannen: Please redesign this so that it does not look up annotations twice.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
sbrannen: What is the meaning of `digraph` in this context?

Please either rename it or document its purpose.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -172,6 +175,76 @@ private static int getOrder(MethodDescriptor descriptor) {
...
+					if (!dependencySize.containsKey(name)) {
+						int sz = DFS(name);
+						if (sz == -1) {
+							throw new IllegalArgumentException("Cycle Detected!!!");
+						}
+					}
+				}
+			}
+			catch (IllegalArgumentException exception) {
+				logger.error(exception,
```
sbrannen: I would argue that the exception should not be swallowed here.

If the user has a configuration error in the test class, that should fail the test class.

Please rework this to throw a `JUnitException` instead.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -172,6 +175,76 @@ private static int getOrder(MethodDescriptor descriptor) {
...
+			}
+
+			context.getMethodDescriptors().sort(Comparator.comparing(DependAnnotation::getDependencySize));
+		}
+
+		private static int getDependencySize(MethodDescriptor descriptor) {
+			return dependencySize.getOrDefault(descriptor.getMethod().getName(), 0);
+		}
+
+		private static int DFS(String name) {
```
sbrannen: Assuming you mean depth-first search by "DFS", please rename this to `depthFirstSearch()` and document what it does in any case.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
sbrannen: If you cannot figure out a good way to eliminate the double lookup, please at least change the above to the following.

```java
			// @formatter:off
			digraph = context.getMethodDescriptors().stream()
					.filter(descriptor -> descriptor.isAnnotated(Depend.class))
					.collect(toMap(
							descriptor -> descriptor.getMethod().getName(),
							descriptor -> descriptor.findAnnotation(Depend.class).map(Depend::methods).get()));
			// @formatter:on

```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -172,6 +175,76 @@ private static int getOrder(MethodDescriptor descriptor) {
...
+			int total = 1;
+
+			if (ancestors != null) {
+				for (String ancestor : ancestors) {
+					int sz = DFS(ancestor);
+
+					// cycle detected
+					if (sz == -1) {
+						return -1;
+					}
```
sbrannen: Now that I've looked more closely at the code, it appears to me that it would be better to simply throw an exception with an informative message at the first detection of a cycle, like this:

```java
if (sz == -1) {
	throw new IllegalArgumentException(
		String.format("Cycle detected between %s and %s", name, ancestor));
}
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
sbrannen: I take that back.

A `MethodOrderer` is applied during the "discovery phase", thus any such exception will be swallowed anyway and only logged at `DEBUG` level by the Jupiter test engine.

Thus, it's best to log it as an error here as you have done.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
duckladydinh: I have remove all states in the class and provide these as parameters to depthFirstSearch. I have also added an explanation for what it is used for in comment.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
duckladydinh: Done renaming and commenting it usage.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
duckladydinh: I have edited as suggested.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
duckladydinh: I have remove all these class variable except for Logger and provide them as parameters to depthFirstSearch.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
duckladydinh: Done renaming.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
duckladydinh: I have changed as above using the isAnnotated() method. 
I can also redesign using for loop, if and add each element one by one to avoid double lookup but I am not really sure if it is much better so I just use the isAnnotated method for now. If double lookup should be avoided at all cost, please tell me and I will rewrite it that way.

sbrannen:
Thanks for making the changes. 

It's much easier to follow now.

documentation/src/test/java/example/ExampleDependsOnAnnotationTestCase.java
```diff
@@ -0,0 +1,50 @@
...
+
+@TestMethodOrder(DependsOnAnnotation.class)
+public class ExampleDependsOnAnnotationTestCase {
+	@Test
+	void d() {
+		System.out.println("Test 1");
+	}
+
+	@Test
+	@DependsOn(value = "d")
```
sbrannen: The reason we switched from `methods` to `value` is so that developers do not have to type `value =`.

So, please change all occurrences of `@DependsOn(value = "X")` to `@DependsOn("X")`. 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -172,6 +175,74 @@ private static int getOrder(MethodDescriptor descriptor) {
...
+				descriptor -> descriptor.isAnnotated(DependsOn.class)).collect(
+					toMap(descriptor -> descriptor.getMethod().getName(),
+						descriptor -> descriptor.findAnnotation(DependsOn.class).map(DependsOn::value).get()));
+
+			// Give each an @Order's value equivalent to its number of previous dependencies
+			Map<String, Integer> dependencySize = new HashMap<>();
+
+			try {
+				// loop through all vertexes (methods) in graph
+				digraph.keySet().forEach(name -> {
```
sbrannen: You can replace this with the following one-liner.

```java
digraph.keySet().forEach(name -> depthFirstSearch(name, digraph, dependencySize));
```

duckladydinh: Done.

documentation/src/test/java/example/ExampleDependsOnAnnotationTestCase.java
duckladydinh: Done and the new test is also automated.

asolntsev:
Please, DO NOT introduce this feature in JUnit! Tests SHOULD BE INDEPENDENT!

I always praised JUnit for not allowing to set specific tests order. Please don't kill my believes....

ludenus:
I find dependsOn feature extremely useful for integration tests in testNG. I would love to have it in JUnit.

pavelpp:
@ludenus if your tests are dependant, then essentially it's a single long test with multiple steps. Just combine step calls under one test method and you'll get the same result.

smoyer64:
Those of you commenting on the merits and/or detriments of ordering tests should also see #18, #48 and #111.  I don't think that anyone has ever said tests should be ordered by default - and in practice I don't need ordering for unit tests.  When it comes to integration/system/acceptance tests, I find ordering invaluable for certain scenarios - e.g. for testing REST interfaces, it's convenient to test GET on a non-existent resource, test POSTing a new resource and then test GET again to verify resources that do exist are properly returned.  There are many times that Selenium tests also benefit from ordered tests.

As noted above, it's true that you can put the entire test sequence in a single test method and achieve the same result but the goal, in my opinion, is to add clarity and reuse (yes, a large test method can simply be a series of calls to "parts" of the test that can be reordered).  The one thing you can't accomplish easily with a single test method is to have the test name tell you exactly why the test failed.  If you create one method for the entire scenario, you're going to have to figure out what failed via the assertion's output.

asolntsev:
@smoyer64 yes, you will have to figure out what failed via the assertion's output, and it's absolutely not a problem. You will need to figure it out anyway.

pavelpp:
@smoyer64 
>  The one thing you can't accomplish easily with a single test method is to have the test name tell you exactly why the test failed. 
Is it really JUnit's responsibility to provide a nice and readable report? There are great frameworks out there, which do that for you.

jbduncan:
> Is it really JUnit's responsibility to provide a nice and readable report? There are great frameworks out there, which do that for you.

@pavelpp Oh, really? Would you mind sharing an example or two? :)

(It's my belief that JUnit test reports should at least be readable. I can't speak for the JUnit 5 team, though. :wink:)

pavelpp:
> @pavelpp Oh, really? Would you mind sharing an example or two? :)
> 
@jbduncan take Allure for example. I'm not using it myself, but I've seen nice examples, where each method marked with a @Step annotation is nicely logged in the report. https://docs.qameta.io/allure/#_junit_5

> (It's my belief that JUnit test reports should at least be readable. I can't speak for the JUnit 5 team, though. 😉)

I think JUnit's reports are "readable enough". If seeing test failure reason is not easy for step methods within one macro-test, then maybe JUnit 5 should solve this reporting problem first, instead of introducing a whole new concept of test dependencies, wdyt?

duckladydinh:
> Please, DO NOT introduce this feature in JUnit! Tests SHOULD BE INDEPENDENT!
> 
> I always praised JUnit for not allowing to set specific tests order. Please don't kill my believes....
Hi, I am just curious, it's a feature and not set up by default, so if you don't need it you can simply not use. What can possibly go wrong with your belief?

> @ludenus if your tests are dependant, then essentially it's a single long test with multiple steps. Just combine step calls under one test method and you'll get the same result.
How about reusability? I mean we do not need this for unit test, but let's say for scenario tests, where multiple scenarios depend on similar methods, how would we implement it?

pavelpp:
> How about reusability? I mean we do not need this for unit test, but let's say for scenario tests, where multiple scenarios depend on similar methods, how would we implement it?

I don't see how this is a problem. Java and JUnit provide all means to make these methods reusable.

smoyer64:
@pavelpp 

>Is it really JUnit's responsibility to provide a nice and readable report? There are great frameworks out there, which do that for you.

@sormuras spent a lot of time on the tree printer and the whole team really thought through both the report hierarchy (and added ``@DisplayName``) specifically to make the default behavior concise, usable and beautiful out-of-the-box.  The ability to extend the reporting capabilities or to replace them completely is awesome IMHO.

I'd like to use the fruits of those efforts to best effect and, in Eclipse or VS Code, I like the test tree with just test names and pass/fail status.  If those methods are small, I know what's broken without looking at the specific assert that didn't pass.

duckladydinh:
> > How about reusability? I mean we do not need this for unit test, but let's say for scenario tests, where multiple scenarios depend on similar methods, how would we implement it?
> 
> I don't see how this is a problem. Java and JUnit provide all means to make these methods reusable.

How? Can you give an example? Let's say I have the following tests:

1. Create a company configuration.
2. Change name.
3. Change address.
4. Change email.
5. ...

All the "Change something" depends on the first test. 

How are you going to implement it in Java and JUnit without dependent ordering without creating the the company every time?

pavelpp:
> @pavelpp
> 
> > Is it really JUnit's responsibility to provide a nice and readable report? There are great frameworks out there, which do that for you.
> 
> @sormuras spent a lot of time on the tree printer and the whole team really thought through both the report hierarchy (and added `@DisplayName`) specifically to make the default behavior concise, usable and beautiful out-of-the-box. The ability to extend the reporting capabilities or to replace them completely is awesome IMHO.
> 

So let me get this straight - by introducing tests dependencies you hope to avoid the need to look at the stack trace? That does not strike me as a convincing argument. I mean no offence, but proper logging and methods naming can solve all your problems. Besides - you will never-ever get a detailed enough report if something fails deeper in one of the steps (i.e. while calling some service) and you will still need to look at the stack trace.

Perhaps bdd-style frameworks would work better in this case, as they do exactly what you want, but they don't position steps as tests, though a failing step will show exactly at what stage scenario failed it still requires you to drill down into test report to analyse why that happened.

IMHO people are worried this will be introduced in JUnit 5 because we've seen far too many examples of this feature misuse.

duckladydinh:
> > How about reusability? I mean we do not need this for unit test, but let's say for scenario tests, where multiple scenarios depend on similar methods, how would we implement it?
> 
> I don't see how this is a problem. Java and JUnit provide all means to make these methods reusable.

Let's say I have the following tests:

1. Create a company configuration.
2. Change name.
3. Change address.
4. Change email.
5. ...

All the "Change something" depends on the first test. 

@pavelpp How are you going to implement it in Java and JUnit without dependent ordering without creating a new company every time?  

If you write all in 1 test, the first assert will just break the test, which means you may have to run the test 100 times if there is 100 "Change something" that fail. You expect to look at the stack trace 100 times instead of 1? 

If you write them in independent tests and repeat a small method call "Create company", you have to create a new company resource too many times, which is too expensive and may seriously just break your database (clean up database before every single run is expensive).

> but proper logging and methods naming can solve all your problems.

That helps a bit in code reuse but how about resource reuse?

All of that just fail in this very simple case if you just give it a try.

asolntsev:
I don't see any problems here. Nothing really fails.

```java
@Test
public void myLongIntegrationTest() {
createCompany(); // it could be private method, "step", page object, factory - whatever
changeName();
changeAddress();
changeEmail();
}
```

duckladydinh:
> I don't see any problems here. Nothing really fails.
> 
> ```java
> @Test
> public void myLongIntegrationTest() {
> createCompany(); // it could be private method, "step", page object, factory - whatever
> changeName();
> changeAddress();
> changeEmail();
> }
> ```

Hi @asolntsev , there is a big problem with this solution. Let's say all the 'change something' have errors. If you just put them together like that, you will need to repeat the test at least 3 times just to find which method will fail since the whole thing will fail if the first method fails. Whilst for dependent ordering, you can see all the failed methods in 1 run and all their logs and stack traces in one run, why would you want to do it 3 times instead of 1?

In this simple case, there are only 3 methods for the company configuration, so you see it as no problem. But in actual work that I do every day, there are between 20-40 methods for each configurations. If I put all of them together like that, I have to continuously and repeatedly log in and out of the server (server needs restarting after each fix). Nothing will fail, but it is just a terrible idea. With dependent tests, I can just log in the server once, view all logs, fix all issues at the same time. Without it, I have to do 20-40 times for the same problem. The benefit is obviously clear in this case.

So @asolntsev , do you agree? Unless there is to run the whole big test from beginning to end even if one method fails in the middle, otherwise, I cannot see how it can be done without dependent tests.

pavelpp:
@duckladydinh 
>you will need to repeat the test at least 3 times just to find which method will fail
What you are saying does not make any sense to me. Why repeat test 3 times? What is the purpose of this? What I hear is that it's just a poor test design, not a problem of the test framework.

asolntsev:
@duckladydinh Sorry, I don't understand how dependant tests solve this problem. No matter if `createCompany` is a test or helper method, if it fails, all following steps will also fail. I do not understand how you can find all errors in all tests in just one run if the first step/test/method is failing.

pavelpp:
@duckladydinh I think your problem here is that you are re-using a test to create some state. Never-ever do that. State preparation should be fast and reliable. Either use service endpoints or direct write to db to create testable state.

duckladydinh:
Hi @pavelpp

> What you are saying does not make any sense to me. Why repeat test 3 times? What is the purpose of this?

The reason to repeat 3 times is that because I assume all the changeSomething tests fail. First, the changeName fails and the changeAddress and Email will never execute, so we cannot know if there is anything wrong with it. After you fix the changeName, the changeAddress fails and changeEmail will never execute, so again, we cannot know if there is anything wrong with it. Then after you fix the changeAddress, the changeEmail fails, which is the 3rd time. That makes at least 3 times in total assuming that the fix is too trivial to verify.

>  What I hear is that it's just a poor test design, not a problem of the test framework. And this is exactly the reason why I'm against this feature, as it encourages such bad test design.

This is a very simple test scenario. Can you please give it a try? I am really curious if there is something better for this. I do not invent a new test design. Instead I just follow a design that has been agreed and improved by countless people in my company for 18 years. I do not think it is the best in the world, but can you provide any alternative here for just this very simple scenario?

> I think your problem here is that you are re-using a test to create some state. Never-ever do that.

This is called scenario tests and all of them are essential tests belonging to one configuration. They are related even by their own nature, what is wrong with sharing some state between them, which is basically against their own logic here? This is integration test, not unit test where you should absolutely not have states. More importantly, this feature has been implemented in TestNG for years and I am not sure but JUnit 4 should have it also. They are added since there are practical use cases that require them. 

>  State preparation should be fast and reliable. Either use api or direct write to db to create testable state.

Should be... , it is great if it is but in practice, performance is not something that will be fast just because we think it should. We have several teams and experts for building the system infrastructure, and if there is something that can be done they should have already done that. And one of my job as a support is to write tests and fix bugs, I cannot just go to the other teams and tell them to change their workflow or technologies so that I can do my job (that is if a better way really exists). Most importantly, everything has been solved pretty well with dependent test ordering. Why should we not use it?

>  Sorry, I don't understand how dependant tests solve this problem. 

Secondly @asolntsev it solves my problem if you use dependent test ordering as you can see in the picture. All 3 methods will fail and we can know what is wrong with all of them in the first run.

image

> No matter if createCompany is a test or helper method, if it fails, all following steps will also fail.

Of course, if createCompany fails, it is the same as one big test in the above solution, but it means that dependent tests are always at least equal or better than independent tests in this kind of scenario. Is it not good enough as a reason? But in practice, between 1 method 'createCompany' and 40 methods 'changeSomething', which one do you think is more likely to fail?

> I do not understand how you can find all errors in all tests in just one run if the first step/test/method is failing.

Okay, agreed. Dependent tests create multiple layers of tests. You cannot do anything nice across layers, but you can do it in one layer like in my example. If createCompany fails, it is over. If changeSomething fails, you clearly see the benefit.

sormuras:
Sometimes, I long for a reporting or logging feature like marking "milestones" within a single test that spans serveral actions. Something like this:

```
@Test void scenario4711(TestReporter reporter) {
  Db.prepareScenario(4711);
  try {
    reporter.milestoneReached("Database prepared");
   // changeAddress();
   reporter.milestoneReached("Address changed");
   // changeEmail();
   reporter.milestoneReached("Email changed");
   // changeName();
   reporter.milestoneReached("Name changed");
  } finally {
    Db.rollback();
  }
}
```

jbduncan:
> Sometimes, I long for a reporting or logging feature like marking "milestones" within a single test that spans serveral actions...

Now that would be nice to have. :)

duckladydinh:
> As an aside, test ordering and concurrent test execution seem to be conflicting ideas 

actually, I think it would be nice if we have a feature to run tests which have the same priority (not intersecting with each other) concurrently.

talios:
On 17 Mar 2019, at 3:46, duckladydinh wrote:

> More importantly, this feature has been implemented in TestNG for years and I am not sure but JUnit 4 should have it also. They are added since there are practical use cases that require them.

*delurking*

One thing in regards to the TestNG usage of dependencies - TestNG *also* introduced the concept of skipped tests, which - in this case, when a test method/test class fails its dependency the test is skipped, and reported as such - not just as a mere failure - it's more of a specialisation saying "hey, we didn't even ATTEMPT to the run this".

It's useful, but not critical.


---
"The ease with which a change can be implemented has no relevance at all to whether it is the right change for the (Java) Platform for all time." &mdash; Mark Reinhold.

Mark Derricutt
http://www.theoryinpractice.net
http://www.chaliceofblood.net
http://plus.google.com/+MarkDerricutt
http://twitter.com/talios
http://facebook.com/mderricutt

duckladydinh:
> It's useful, but not critical.

Hi @talios , as long as you agree it is useful, that's enough.

You know, let's take the @Disable for example, it is implemented, but why? Is it crucial? Isn't it faster to just comment out the @Test (Well, maybe there is another use for the @Disable but there are also many other uses for dependent tests also)? It is just similar. Again, it is just an optional feature that gives us more flexibility, not limiting us.

And whether it is critical or not, I think it really depends on your work: Your work says it is critical, then it is critical. Personally, I sometimes find it impossible to not use dependent tests. Like @asolntsev and @pavelpp , they both complained that dependent testing is bad, but none could find a solution to the given very simple scenario, let alone more complex ones in real world. @asolntsev 's solution has not just one disadvantage as I listed above but more. But okay, probably my standard is just too high so it may not be critical.

> One thing in regards to the TestNG usage of dependencies - TestNG *also* introduced the concept of skipped tests, which - in this case, when a test method/test class fails its dependency the test is skipped, and reported as such - not just as a mere failure - it's more of a specialisation saying "hey, we didn't even ATTEMPT to the run this".

Well, what is wrong with skipping then if the dependent test already failed (let's say server receives but not replies)? Would you like to run them all? Let's say each has a timeout of 30 seconds and there are 100 of them, so we must  wait for an hour even if we know that they will fail?

**Anyway, this is just my personal view :). If anyone finds something wrong, please speak your mind. I want to learn more.**

asolntsev:
> Of course, if createCompany fails, it is the same as one big test in the above solution, but it means that dependent tests are always at least equal or better than independent tests in this kind of scenario. Is it not good enough as a reason? But in practice, between 1 method 'createCompany' and 40 methods 'changeSomething', which one do you think is more likely to fail?

@duckladydinh  (Sorry for a long delay)
I absolutely don't understand your argumentation. How do you conclude that _dependent tests are always at least equal_? How this logic works? No, I don't think that dependent tests are at least equal. 

> All 3 methods will fail and we can know what is wrong with all of them in the first run.

I still don't see how dependant tests give you a knowledge what is wrong with all of them. I see on your picture that method `changeAddress` failed, but we don't know if `changeEmail` and `changeName` are ok or not. You still need to run tests again after fixing `changeAddress`.

duckladydinh:
> I absolutely don't understand your argumentation
Okay @asolntsev , I will try to make it slower so that you can understand it clearer. If you have questions again, please feel free to ask me.

> No, I don't think that dependent tests are at least equal.
I think that the best way to prove this is that you should give me a situation where it is worse first. Then I will immediately explain to you why it is still at least equal in any example you may give.

 (I really want to ask why you think dependent test is bad :) and how your logic works :) and how you prove that your belief is not completely wrong :) or do you have any research paper called "Dependent Tests considered harmful" or something like that :) , but let's forget it for now :) )

> I see on your picture that method changeAddress failed, but we don't know if changeEmail and changeName are ok or not.
We **do know** because as you can see they **are red** (not yellow, not green, but clearly red, I am not sure if the picture color changes in you PC, but the picture I posted shows 3 failed methods that are red), which means they **fail**. So since changeEmail and changeName **are red**, we know that they are not okay. Do you get this point? I confirm again, we know that they are not okay, because they are red (they fail).

> I still don't see how dependant tests give you a knowledge what is wrong with all of them.
Probably, you never try it out on your PC because of your belief so I can understand why you cannot see the information. You have to try it out to see it. But here is the explanation.

You have to click on the failed test (on the method with **red color**, if you click on the parent method, it may not show) and scroll down and you will see the exception for every failed test. The knowledge we get here is the exceptions. Am I clear at this point? 

If you do not understand why there is exception when you double click on the red method, then I suggest that you should try it out. It will be like magic if you really never see it but it is the way it works. Okay? Just implement some empty function and put "throw new RuntimeException();" in their body. And you will see them in the log.

I repeat, exception has stack trace, which shows us what is wrong with  the test and since we can see all exceptions here, we know what is wrong with all of them.

---
Please do not hesitate to ask me if you still do not get it, but it would be better if you could please give a counter example for what you do not agree in my argument. I will do my best to be clearer so that you can understand it better. Okay?

asolntsev:
@duckladydinh Thank you for clarifications, it seems that I get your point now.
But this is logically wrong.

Look:
> since changeEmail and changeName **are red**, we know that they are not okay

it's not true. `changeEmail` and `changeName` may be okay, but they are occasionally red because `changeAddress` did something wrong. 

And imagine the opposite case: `changeAddress` is red, but `changeEmail` and `changeName` are green. Does it mean that they are okay? Not. They may be broken, but occasionally green because `changeAddress` did something wrong. That's the essence of dependant tests.

P.S. I see you are trolling me with stacktraces and colors. Cool, I like trolling. :)
But I recommend to make some background screening next time before trolling somebody. ;)

duckladydinh:
> And imagine the opposite case: changeAddress is red, but changeEmail and changeName are green. Does it mean that they are okay? 

Well, @asolntsev , because you do not use dependent testing, so you may not know the rule: _Methods in the same priority must not intersect with each other_. If changeAddress can affect the other 2 (intentionally like in Japan, the name and  address are related), then they should be given different priorities in that case.

In my dummy scenario, I assume that they are independent methods with the same precondition (the company has been created) that can freely run at the same time, and only in that situation can you use dependent testing.  Hence I would say **YES** since the logic is perfectly fine.

> That's the essence of dependant tests.

That's not :V. It is just because you are breaking the basic **RULE**.

> But I recommend to make some background screening next time before trolling somebody. ;)

Sorry, too sleepy to make it :'(.

pavelpp:
@duckladydinh Let me clarify this - in the scenario above you run createCompany() once, then methods changeAddress, changeEmail, changeName follow, but none of them depend on each other, only on createCompany. Is my understanding correct? In that case we are talking about a classic example of a test setup method like @BeforeClass. Correct me if I'm wrong.

duckladydinh:
Hi @pavelpp , In terms of whether it works or not, you are absolutely right. @BeforeClass will work. However, are you really okay with that solution?

1. Did you ask yourself: **Does the solution scale?** In this dummy example, where we have 2 layers, it is fine, but have you ever thought of any scenarios with 4, 5, 6 layers? What will you do in these cases?

3. The test logic is messed up since **environment setup should not be the test itself**. This is just **my personal taste**, but when you say that @BeforeClass it is a "setup method", I believe that you should distinct itself from the normal test logic. I usually use it for server connection and no test there since it may give **wrong assumption** to new people that creating a company is required to set up the environment. Well, but it is just my personal taste. If you have 1 test, it works fine. If you have 100 tests having the same setup, definitely no.
 
3. **No matter how you think about it, you are using dependent testing without knowing**. It does not matter if @BeforeClass has been already introduced or not. By using it as a **test**, you have already forced an ordering relationship on your tests (the createCompany test always has to run first, and the other after). It does not matter if the name of the annotation is @BeforeClass or @DependsOn, the tests are no longer independent of each other. In other words, I don't know why you are okay with it this way despite your belief. Do you hate the name @DependsOn, maybe?
 
4. **Still not good enough**. If the @BeforeClass fails, the remaining tests still have to run instead of being skipped, and the reason why it is bad is as mentioned above.

> Well, what is wrong with skipping then if the dependent test already failed (let's say server receives but not replies)? Would you like to run them all? Let's say each has a timeout of 30 seconds and there are 100 of them, so we must wait for an hour even if we know that they will fail?

Well, anyway, that solution does work in this dummy case, even if it is unscalable, messing up code logic (this may be only my personal view), against your own belief of independence in tests, lacking of features like in a real dependent testing. Unfortunately, it should be a no-no in real complex business scenarios.

asolntsev:
> know the rule: _Methods in the same priority must not intersect with each other_. If changeAddress can affect the other 2 (intentionally like in Japan, the name and address are related), then they should be given different priorities in that case.

@duckladydinh Wow, now we are talking!
You say "methods.. must not intersect". It exactly means "tests must be independent" :)

What means "they should be given different priorities"? What happens then?

Overall, this principle of "must not intersect with each other" is too shaky. What is definition of "intersect"? I believe it cannot be formalised. It's too easy to make an error when deciding if some methods "intersect" or not. It's a clear way to create unstable tests. 

> environment setup should not be the test itself

Wow. Yes. I absolutely agree. But you propose exactly that: your first test `createCompany` **does setup preconditions for the following tests**.

duckladydinh:
> You say "methods.. must not intersect". It exactly means "tests must be independent" :)

Hi @asolntsev , it does not. When trying to tweak a sentence, please include the clear context and then the meaning will be clear: Methods **in the same priority** must not intersect.

> tests must be independent

The scenario tests model the real business logic, which cannot be changed by programmers. You can only model them in your test not change them. But basically, according to your comment, you should see now that the so called "dependent testing" is actually more independent than your 'independent' testing by putting everything in one block. In the end, it seems clear to me that the only thing you cannot accept is the name.

> Overall, this principle of "must not intersect with each other" is too shaky. What is definition of "intersect"? I believe it cannot be formalised. 

By "intersect", let's define it as not "independent". You should at least understand it, right, since it is your belief :) ? And if you yourself fail to comprehend your belief, then I am speechless. And it is not a standardized principle. It is a common sense. If they depends on each other, then they must be put in different layers.

> It's too easy to make an error when deciding if some methods "intersect" or not. It's a clear way to create unstable tests.

Actually you know, if you have a method that can change the state of something that does not have any relationship with it, then the problem lies in your code itself, not the test default. If such terribly designed methods exist, even if you try to put everything in one block like you did in the dummy case, it is still unstable since the program you are testing is itself unstable. I suggest that you should maybe find some suitable design pattern, try functional programming or just hire better developers, which will lessen the problems.

More important, it is exactly what @pavelpp says

> What I hear is that it's just a poor test design

> your first test createCompany does setup preconditions for the following tests.

By environment setup, what I means is for example, creating a connection to the platform, or just the database. There is a technical relationship between such setup and tests since it is required for execution but what is the business relationship? None. That's the difference so please stop putting them together. I am sure that most company founders do not even know to use database.

---
And @asolntsev I have a suggestion. Instead of trying to say why dependent testing, a scalable solution that can solve real problems, is bad. Why don't you at least show the advantage of putting everything in one block over dependent testing?

asolntsev:
@duckladydinh sorry, I am lost. You like dependant test, while saying that they should not _intersect_, meaning that they should not depend on each other. Or they should be assigned different _priorities_. Sorry, it's not a common sense. It's too complex system with terms like "intersect", "layers", "priority" which are too hard to define precisely.

> Why don't you at least show the advantage of putting everything in one block over dependent testing?

Wait, I never said I like to put all steps to one long test method. I don't practice it. What I said is that you do effectively the same when you create smaller tests that depend on each other. _Couple of dependant tests is as bad as one long test_.

What I really value is small, fast, independent tests. They prepare preconditions using API, inserting data to database etc. - by any means except UI. _When I test UI, I am not allowed to trust UI_. And yes, probably you need to hire better developers to create such an API. ;)

duckladydinh:
> Wait, I never said I like to put all steps to one long test method. I don't practice it. 

you did 'show' it in writing. You solution to the example scenario is exactly like that. Please at least have an idea of what you have written.

> What I really value is small, fast, independent tests.

you get all of that with dependent tests. It is first faster, since you don't need to run configuration twice. Smaller since it reuses code and dependent testing is actually more independent than your practice as shown in your example solution. Just try it and see. Please at least try it first.

---

...and anyway, @asolntsev , I can explain more but unless your job involves more complicated tests, just forget about it :) Powerful tools are not good if you do not need it :)

asolntsev:
@duckladydinh You are making some conclusions about my job without any knowledge about it. I don't want to continue. My job does involve complicated tests, but we can make them fast and stable. We can do it effectively. 

You decide, if you can write simple and stable solution, or need "powerful" tools.

duckladydinh:
> @duckladydinh You are making some conclusions about my job without any knowledge about it. I don't want to continue. My job does involve complicated tests, but we can make them fast and stable. We can do it effectively.

Sorry @asolntsev , by no way I mean your job is easy or something. I mean if you do not have the use case for it, let's not continue. It is pointless if you say you are lost after not even reading carefully and not even knowing what you have written yourself.

> You decide, if you can write simple and stable solution, or need "powerful" tools.

At least you know that it is "If".Anyway, if 'powerful' tools can help me simplify the simple things even more and stabilize the stable things even more, why not use it? Unless you can try it out, you do not know anything about it. It does not make sense to say the way you do it is better.

dharmeshsvg:
@duckladydinh : Can you please try to run below test using @DependsOn annotation? I think code is failing w.r.t. below scenario:
	@Test
	void independentTest() {
		assertEquals(0, 0);
	}
	@Test
	void gamma() {
		assertEquals(1, 1);
	}
	@Test
	@DependsOn("gamma")
	void delta() {
		assertEquals(2, 2);
	}
	@Test
	@DependsOn("delta")
	void beta() {
		assertEquals(3, 3);
	}
	@Test
	@DependsOn("beta")
	void alpha() {
		assertEquals(4, 4);
	}

As per expected behavior All the test should become pass but 2-3 test become skip in this case.

duckladydinh:
Hi @dharmeshsvg , thank you for testing it out. I believe I got the your example right on my machine. Just to make sure, did you include the @TestMethodOrder(DependsOnAnnotation.class)? That is the most important one. If you just use @ExtendWith(DependsOnTestWatcher.class), then there will be 3 tests skipped.

Screenshot from 2019-04-21 17-04-27
Screenshot from 2019-04-21 17-07-35

dharmeshsvg:
> Hi @dharmeshsvg , thank you for testing it out. I believe I got the your example right on my machine. Just to make sure, did you include the @TestMethodOrder(DependsOnAnnotation.class)? That is the most important one. If you just use @ExtendWith(DependsOnTestWatcher.class), then there will be 3 tests skipped.
> 
> Update: Hi @dharmeshsvg, I just commit your test also as ExampleDependsOnAnnotationClassicTest since I think it makes sense to have one with only the MethodSorter.
> 
> Screenshot from 2019-04-21 17-04-27
> Screenshot from 2019-04-21 17-07-35

@duckladydinh : You are right . It's because of "@TestMethodOrder(DependsOnAnnotation.class)". I have used this annotation but some how its not working in my local that's why i am facing this issue.

duckladydinh:
Hi @dharmeshsvg , so does the problem still persist in you local now :smiley: ? If it does, just let me know, I will see what I can fix.

And again, thank you for testing it out :smile: .

rmarinsky:
ганьба!

artsok:
@rmarinsky It's alright bro :)

dharmeshsvg:
> Hi @dharmeshsvg , so does the problem still persist in you local now  ? If it does, just let me know, I will see what I can fix.
> 
> And again, thank you for being my very first user  .

@duckladydinh : Yes it's still persist in my local. I have find out exact step to reproduce issue.
Please add following  annotation into "ExampleDependsOnAnnotationClassicTest" test:
@ExtendWith(DependsOnTestWatcher.class)
@TestMethodOrder(MethodOrderer.DependsOnAnnotation.class)

If  we use only "@TestMethodOrder" annotation then its working fine. but if we used both annotation then 3 test becomes skipped.

Can you please check it from your side?

duckladydinh:
@dharmeshsvg Interesting, so you only extract the files to test it. Probably some feature is not supported. For now I can only show that it is working fine with the latest version (5.5) as in the picture. I will try your package first and let you know later.

Screenshot from 2019-04-24 23-03-33

duckladydinh:
Okay, @dharmeshsvg , the problem is very simple. You extract it the wrong way :). Instead of creating a new interface TestMethodOrder and MethodOrderer, what you need to do is to use what is already included in the JUNIT framework. Otherwise, interfaces will be nothing if you do not register them with the platform.

In detail:

1. Please remove your TestMethodOrder and MethodOrderer (keep the DependsOnAnnotation class) interface.
2. Make the DependsOnAnnotation class a separated class.
3. Fix the import in the tests and run again.

Expected: Everything is perfect.

It's great to know the test is perfect :smiley: .
[DependsOnTest.zip](https://github.com/junit-team/junit5/files/3114239/DependsOnTest.zip)

desruisseaux:
My comment will probably not add much to the debate, but I would just like to add my support for dependent tests too because they save a lot of time when fixing regression bugs. We have lot of situations where if test of feature A fails, then test of feature B is likely to fail too if B uses A. Making sure that A is tested before B makes faster to identify the root cause of a regression. It is not an obligation, but just a way to save a lot of time when a single regression can cause hundreds of tests to fail and we need to find _the_ test on which to focus.

I know that the typical counter-argument is to create mocks, but I have at least one case where creating the mock was more than one week of full-time development effort and put us in the absurd situation where debugging the mock was almost as much effort as debugging the real code.

asolntsev:
@desruisseaux Thank you for your opinion. At the beginning I was even agree with you: I also want to run tests in a defined order when I want to reproduce some concrete failed build (which fails only in given order of tests). But then I use test suite. 

In your case, "creating the mock requires one week of full-time development effort" sound crazy. It's not ok. Creating mocks should be much faster and easier. I would help you with a pleasure to understand what went wrong with your mocks.

desruisseaux:
In my use case the tests failures do not depend on execution order: the exact same set of tests fail regardless order. But without dependency information, the root cause of failures can be lost in hundred of error messages and it can take a significant amount of time to find it. With tests in dependency order, the first test failure is likely to be the root cause of all other test failures.

Actually I don't really need ordering, so this feature could be considered orthogonal with the `@Order` annotation. What I need is a mechanism allowing JUnit to said _"Test B fails, but the feature tested by B depends on a feature tested by A which failed too. You way may want to look at test A first."_

About the mock that required a long time to write, it is simply a mock creating some Java data objects from queries a little bit like what Hibernate does. Even if the mock needs to answer to only a very limited set of queries, creating a mini-Hibernate is not easy when the Java data objects are complex. It takes time not only to write the code with hard-coded values, but also to extract some data from the relational database to put in the mini-Hibernate. The relationships between tables make this extraction bug-prone and a lot of debugging time was spent because we made mistakes when hard-coding some data in the mini-Hibernate.

Another example: suppose that I created my own matrix library (I know that there is many libraries around, but let suppose that for whatever reason I needed to write my own). I create tests for matrix multiplications, matrix inversions, _etc._ Then later in the same module I have another class resolving systems of linear equations, which make use of matrix operations. Should I create a `MatrixMock` with hard-coded values of the kind "when you see this sequence of 100 numbers in an `invert` operation, returns that hard-coded sequence of 100 numbers"? I could, but some calculations require so many matrix operations that this approach quickly become impracticable. A more practical approach is to use the real matrix library, but keep a note somewhere which said _"if the tests for matrix multiplications and inversions failed, then it is not even worth to execute this linear system solver test; it will certainly fails too."_ This is this kind of information that I would like to encode in JUnit tests.

marcphilipp:
@desruisseaux Thanks for providing a detailed description of your use case! I think it's an interesting one. Would you mind opening a separate issue for it?

jbduncan:
@desruisseaux I'd be very interested too in hearing more about your use case in a separate issue, as I've found once or twice in the past that having tests run in the order you're describing would have been quite helpful. :)

desruisseaux:
No problem for creating a separated issue, but I though it was the purpose of this `DependsOn` annotation proposal (or that it could be addressed by this proposal, even if the use case is a bit different).

pavelpp:
@desruisseaux Correct me if I'm wrong, but basically you are saying there's no point in running certain tests if some preconditions aren't met (certain other tests fail), is that right? If I had to solve this, I would rather create dependent test jobs on CI with smoke tests and downstream jobs, which are only triggered if those smoke tests pass. This will save you time as well, because longer dependant tests won't run at all. Of course your case might be more complex and then there's a question about running tests locally.

desruisseaux:
Hello @pavelpp. Indeed there is no point in running certain tests when certain other tests fail, but my goal is not to save test execution time. It is rather to save developer time in analysing why tests failed. Taking the above example with linear system solver, let suppose that a developer makes a change in a code not obviously related to matrices, but that change breaks matrix inversion code (yes, the developer should had an unit test for the code he changed, but no so many projects have a 100% test coverage. Maybe he though that he changed trivial code). He run `mvn test` on the command line, then get 50 test failures. Not only in `MatrixTest.testInvert()`, but also in `LinearSystemTest.testSolver()` and every other methods testing a feature that needs, directly or indirectly, matrix inversion. Where to start investigation? JUnit reports test failures in random order and there is no obvious way to identity the root cause. The developer may pick up a random test possibly far from matrix inversion code, starts a debugging session and spend time stepping through unaffected code before to reach the point where he sees the bug that his change is causing in `Matrix.invert()` method.

So my goal is not to said _"There is no point to run `LinearSystemTest.testSolver()` if `MatrixTest.testInvert()` failed"_, but rather to said _"`LinearSystemTest.testSolver()` failed, but that test depends on a feature which is tested by `MatrixTest.testInvert()` and that test failed too. Maybe the test failure is not `LinearSystem.solve()` fault; you should fix `MatrixTest.testInvert()` first and see if that correction resolves indirectly `LinearSystemTest.testSolver()` failure."_ (note: my previous comment tries to argument why the usual recommendation — to use mocks — is not always practical).

My scenario is not heavy tests executed by CI, but rather tests executed quickly enough for making reasonable to run `mvn test` before every commit. But with thousands of tests, even if test execution is fast, debugging can be painful if a single change causes 50 tests to fail with no help for identifying which test is closest to root cause.

pavelpp:
@desruisseaux OK, got it. In reality - how often do you have lots of unit tests depending on a single feature? I suspect not that many.

desruisseaux:
Not exactly depending on a single features, but depending on many features. Feature A depends on features B, C and D, with feature B depending on features C, E and F, _etc_. The number of branches in a dependency tree can grow fast, and this is the high number of possibilities that make difficult to identify the root cause (i.e. looking at a test failure which is the head of the tree, the root cause is a node somewhere in the tree).

This kind of situation is actually quite common at least in numerical computations (e.g. a calculation depends on linear system solver, which itself depends on matrix operations, which may itself depends on complex numbers, etc.). A regression e.g. in the complex number functions can have consequences that propagate in hundreds of other test cases if that function is used (directly or indirectly) often.

pavelpp:
@desruisseaux and that is why in my opinion mocking should be used, because if somebody has such a complex dependency matrix it will be a waste of time to debug the cascading failures. It's better to invest time once into proper mocking framework. Just my opinion of course.

desruisseaux:
But as explained above, mocking can take a lot of time too (how to create a mock for matrix inversions? With hard-coded values of all sequences of numbers that the matrix may see during the execution of any test in the whole library?) and result in absurd situations where we spend weeks creating and debugging the mock. That time would be better spent in debugging the actual code instead.

asolntsev:
> But as explained above, mocking can take a lot of time too (how to create a mock for matrix inversions?) 

Still, the true answer is that you should create mocks. The problem with mocks that you described earlier sounds crazy. Mocks should be simple. Mocks should not read any data from database. Mocks should answer with static data snippets (or close to that). Building mocks should be fast and easy.

If I can help somehow to improve your mock, let me know. :)
Just a couple of links on the topic:

- [Integrated Tests Are A Scam](https://blog.thecodewhisperer.com/permalink/integrated-tests-are-a-scam)
- [Antistatic](https://www.youtube.com/watch?v=dFQSOlOOoXE&list=PLfazdZ9SzB9eDJIugtfH7KeVLLAP1pDLh)  -- here you will find how to create a mock for matrix inversions

desruisseaux:
Some problems are easy to "mockify", but not all kind of problems. Part of the mock that took a long time to create is [there](https://github.com/opengeospatial/geoapi/blob/master/geoapi-conformance/src/main/java/org/opengis/test/referencing/PseudoEpsgFactory.java). It is not many lines, but was long to debug because some values are tedious to extract from the database and any mistake, after the values goes through potentially complex map projection formulas, results in a wrong output values difficult to relate to the wrong input values. In the week it took me to create that mock (together with another similar mock), almost all bugs were in those mocks and practically none of them in the real code.

Creating matrix inversion mock is not helpful. It may be helpful for a single test doing one or two inversions, not with test of a calculation doing thousands of matrix inversions. The same is true for any function which is potentially invoked indirectly thousands of time from any other feature that we want to test. It is not practical to mockify everything.

thespags:
This would be a great feature to add.

Issue #1777, opened 2019-02-15
Exception thrown from TestExecutionListener fails the test plan


sbrannen:
## Overview

Currently, if a `TestExecutionListener` throws an exception, that exception will propagate and cause the entire test plan to fail.

## Deliverables

- [ ] Ensure that exceptions thrown from methods in a `TestExecutionListener` are swallowed and logged at `ERROR` level.
- [ ] Update Javadoc for `TestExecutionListener` to describe the expectation that such listeners never throw exceptions and what happens if a listener throws an exception.

gopinath-langote:
I can pick this up. If no one is picking

sbrannen:
@gopinath-langote, feel free to submit a PR!

The desired behavior is analogous to the support for `TestWatcher` callback invocations.

If you have problems finding where/how to test the changes, just let us know.

sbrannen:
> I can pick this up. If no one is picking

If you decide to start working on it, please just say so here so that we can mark it as "in progress".

gopinath-langote:
@sbrannen 
I can work on this.


I need more information on this though.  Could you please redirect me to exact class (with full package ).

Is this for all the methods in the class or all?
How to test the behavior?


Regards,
Gopinath

sbrannen:
The easiest way to find where something like this is used is to search for uses within the code base.

For example, I just searched for where `org.junit.platform.launcher.TestExecutionListener.executionFinished(TestIdentifier, TestExecutionResult)` is invoked, and the results show me:

- `org.junit.platform.launcher.core.ExecutionListenerAdapter`
- `org.junit.platform.launcher.core.TestExecutionListenerRegistry`

So you'll need to analyze invocations of all methods in `TestExecutionListener` and determine the best place to add try-catch blocks.

sbrannen:
>     2. Is this for all the methods in the class or all?

It's for the entire API, so all methods in the `TestExecutionListener` interface.

sbrannen:
Regarding where to test it, you'll need to find an appropriate place within the `platform-tests` module.

And you'll need to create an "exception throwing" `TestExecutionListener` (stub or mock), register it, execute a simple test plan, and verify that the exceptions were all swallowed and logged. See uses of `org.junit.jupiter.engine.TrackLogRecords` for achieving the latter.

gopinath-langote:
Hello @sbrannen ,
Thanks for the information. I will start working on this.

sbrannen:
[...]

PR #1789, opened 2019-02-20
Safeguard against exceptions thrown in TestExecutionListeners

gopinath-langote:
Issue: #1777

## Overview

Prior to this PR, Test plan used to fail upon exception occurs while executing `TestExecutionListener` or `EagerTestExecutionListener`. 
To improve the platform behavior, we decided to not fail the test plan vs log an exception occurred in `TestExecutionListener` or `EagerTestExecutionListener` (WARN level) and continue test plan.

---

---

### Definition of Done


gopinath-langote:
@sbrannen 
I started working on #1777 

A couple of questions:
1. Do we want it as ERROR log or warn log?
2. What about `EagerTestExecutionListener` which extends `TestExecutionListener`, do wan to prevent exceptions there also?

sbrannen:
>     1. Do we want it as ERROR log or warn log?

Let's keep it at WARN for now.

>     2. What about `EagerTestExecutionListener` which extends `TestExecutionListener`, do wan to prevent exceptions there also?

Yes, I think that's also a good idea, even if that's currently only an internal interface used for capturing SYSOUT and SYSERR.

Or, @marcphilipp, do you see the latter as unnecessary?

gopinath-langote:
@sbrannen 
Please review the PR 
+ Let me know if we decided to handle `EagerTestExecutionListener` failures too.


Regards,
Gopi

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java
```diff
@@ -48,11 +52,11 @@
...
-		Arrays.stream(listeners)
-				.filter(EagerTestExecutionListener.class::isInstance)
-				.map(EagerTestExecutionListener.class::cast)
-				.forEach(this.eagerTestExecutionListeners::add);
-		// @formatter:on
+        Arrays.stream(listeners)
+                .filter(EagerTestExecutionListener.class::isInstance)
+                .map(EagerTestExecutionListener.class::cast)
+                .forEach(this.eagerTestExecutionListeners::add);
+        // @formatter:on
```
sbrannen: Please undo the indentation change here.

platform-tests/src/test/java/org/junit/platform/launcher/core/ThrowableTestExecutionListener.java
```diff
@@ -0,0 +1,54 @@
...
+
+package org.junit.platform.launcher.core;
+
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.engine.reporting.ReportEntry;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+import org.junit.platform.launcher.TestPlan;
+
+public class ThrowableTestExecutionListener implements TestExecutionListener {
```
sbrannen:
```suggestion
public class ThrowingTestExecutionListener implements TestExecutionListener {
```

sbrannen:
> Please review the PR

I don't have time to review it in detail at the moment, but I imagine that somebody from the team will review it in the coming week or so.

> Let me know if we decided to handle `EagerTestExecutionListener` failures too.

Yes, please go ahead and do that. Better safe than sorry, and we can always undo that later if we deem it necessary.

Thanks

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java
```diff
@@ -50,7 +54,7 @@ void registerListeners(TestExecutionListener... listeners) {
 		// @formatter:off
 		Arrays.stream(listeners)
 				.filter(EagerTestExecutionListener.class::isInstance)
-				.map(EagerTestExecutionListener.class::cast)
+ 				.map(EagerTestExecutionListener.class::cast)
```
marcphilipp:
```suggestion
				.map(EagerTestExecutionListener.class::cast)
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java
```diff
@@ -63,48 +67,117 @@ private void notifyEagerTestExecutionListeners(Consumer<EagerTestExecutionListen
...
 	}
 
-	TestExecutionListener getCompositeTestExecutionListener() {
+	CompositeTestExecutionListener getCompositeTestExecutionListener() {
 		return new CompositeTestExecutionListener();
 	}
 
-	private class CompositeTestExecutionListener implements TestExecutionListener {
+	class CompositeTestExecutionListener implements TestExecutionListener {
+		public List<TestExecutionListener> getAll() {
```
marcphilipp: Where is this used?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java
```diff
@@ -63,48 +67,117 @@ private void notifyEagerTestExecutionListeners(Consumer<EagerTestExecutionListen
...
 		@Override
 		public void dynamicTestRegistered(TestIdentifier testIdentifier) {
-			notifyTestExecutionListeners(listener -> listener.dynamicTestRegistered(testIdentifier));
+			notifyTestExecutionListeners(listener -> {
+				try {
+					listener.dynamicTestRegistered(testIdentifier);
+				}
+				catch (Throwable throwable) {
+					rethrowIfBlacklistedAndLogTest(throwable, listener, testIdentifier, "dynamicTestRegistered");
+				}
```
marcphilipp: Please move the exception handling into `notifyTestExecutionListeners()` instead of duplicating it.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java
gopinath-langote: in `DefaultLauncher`

`DefaultLauncher`, as the registry to get the `CompositeTestExecutionListener` which is the collection of all listeners and handled the failure too.

Instead of `DefaultLauncher` handing exception one more time, I made this `CompositeTestExecutionListener` visible to `DefaultLuancher` in the same package.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java
gopinath-langote: I thought of doing the same, but we will not get the exact listener method name on logging it in `notifyTestExecutionListeners()`.

sbrannen: Why don't you just pass the method name to `notifyTestExecutionListeners()` as an additional parameter?

gopinath-langote: Yes, we can, 
There are two variables here,
1. method name 
2. the log for `plan` level listener method needs `TestPlan`  & test level listener method need `TestIdentifier`.


Them, we need two variants of `notifyTestExecutionListeners()` one taking `TestPlan` as parameter & second taking `TestIdentifier`.

how does that sounds to you @sbrannen 




Thanks,
Gopi

sbrannen: I think that would be reasonable and still less code duplication. Give it a shot.

gopinath-langote: Sure. Let's give it a shot and see 👍

gopinath-langote: @sbrannen  @marcphilipp 

I was thinking to extract `TestExecutionListenerNotifier` class out to extract the notification logic at one place from `TestExecutionListenerRegistry`, `ExecutionListenerAdapter` or (may be from `DefaultLauncher`.

Currently, this logic (handling with try/catch & logging is at multiple class.


The resulting class will look like, 

```
public class TestExecutionListenerNotifier {
	private static final Logger logger = LoggerFactory.getLogger(TestExecutionListenerNotifier.class);

	public static void notify(TestExecutionListener listener, TestIdentifier testIdentifier, String listenerMethodName, Consumer<TestExecutionListener> consumer) {
		try {
			consumer.accept(listener);
		}
		catch (Throwable throwable) {
			rethrowIfBlacklistedAndLogTest(throwable, listener, testIdentifier, listenerMethodName);
		}

	}
	private static void rethrowIfBlacklistedAndLogTest(Throwable throwable, TestExecutionListener listener,
												TestIdentifier testIdentifier, String methodName) {
		BlacklistedExceptions.rethrowIfBlacklisted(throwable);
		logger.warn(throwable,
				() -> String.format("Failed to invoke ExecutionListener [%s] for method [%s] with test display name [%s]",
						listener.getClass().getName(), methodName, testIdentifier.getDisplayName()));
	}

}```

Here, I am thinking to use static methods.

Any thought on this?

**Comment by gopinath-langote** [comment]
@sbrannen  @marcphilipp Any updates on this PR?


Regards,
Gopinath

File: junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java (hunk 4 above)
Comment (marcphilipp) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: I checked out your branch and I could find no usage of this method in `DefaultLauncher` or anywhere else. I think we can make it private again and revert the changes in `DefaultLauncher` or am I missing sth.?

Comment (gopinath-langote) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: Hello @marcphilipp 
The usages of `TestExecutionListener` interface in `DefaultLauncher` are mentioned below.
These are already handled as of `CompositeTestExecutionListener`. But if you don't explicitly use the `Composite` type, it will be hard to reason about it that the exceptions are handled in some internal class.



1. https://github.com/junit-team/junit5/blob/f114fd009778bb4a33cda97a42bdf68d71348221/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java#L192


2. https://github.com/junit-team/junit5/blob/f114fd009778bb4a33cda97a42bdf68d71348221/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java#L200


3. https://github.com/junit-team/junit5/blob/f114fd009778bb4a33cda97a42bdf68d71348221/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java#L207


Let me know if I need to clarify more.


Regards,
Gopinath

Comment (marcphilipp) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: Thanks, but where is `getAll()` called?

Comment (gopinath-langote) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: Nowhere.

Comment (gopinath-langote) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: Are you suggesting to remove DefaultLauncher Changes, and not handle exception there.

As they are implicitely handled by `CompositeTestExecutionListener`?

Comment (marcphilipp) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: I may be missing sth. but I don't see any behavioral changes in `DefaultLauncher`. `CompositeTestExecutionListener` was already used there before. And if that now handles exceptions we don't need to change `DefaultLauncher` at all, do we?

Comment (gopinath-langote) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: @marcphilipp You are right, `DefaultLauncher` uses `CompositeTestExecutionListener` instead of Interface `TestExecutionListener` which implicitly means all the exceptions are handled by default.

I was thinking from readability perspective that, If we make the interface method signature to throw an exception then we should explicitly handle the exception in `DefaultLuancher`. Maybe I am thinking more ahead.


We can definitely stop here & keep `DefaultLuancher` as it is.

Let's take one call on this.


Also have a look at below comment too.

Comment (marcphilipp) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: We're not going to change the interface so let's revert the changes to `DefaultLauncher` and remove this unused method.

File: junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java (hunk 5 above)
Comment (marcphilipp) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: We don't need to handle exceptions in `ExecutionListenerAdapter` because it's an internal class that's only ever called with a `CompositeTestExecutionListener` which now handles exceptions.

File: junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java (hunk 4 above)
Comment (gopinath-langote) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: Sure. I will do that

File: junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java (hunk 5 above)
Comment (gopinath-langote) [reply junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/TestExecutionListenerRegistry.java]: Sure. I will go through it one more time.

thanks for feedback I will get back with commits :)

**Comment by gopinath-langote** [comment]
@marcphilipp All changes done from my side as discussed please review whenever you get time :)

**Comment by gopinath-langote** [comment]
Updated with upstream master

**Comment by gopinath-langote** [comment]
@marcphilipp 
Thanks for review & polishing commits those are really helpful & better.

Just for my knowledge, what it the implication of adding the logger to `log4j2-test.xml` file?

Regards,
Gopinath

**Comment by sbrannen** [comment]
> Just for my knowledge, what it the implication of adding the logger to `log4j2-test.xml` file?

He disabled the warning log level for those new log entries in the build.

**Comment by gopinath-langote** [comment]
@sbrannen Thanks for the clarification. Good to know that, next time onwards it will be helpful for me to the changes.

File: documentation/src/docs/asciidoc/release-notes/release-notes-5.5.0-M2.adoc [hunk 6]
```diff
@@ -25,6 +25,8 @@ on GitHub.
 
 * `AnnotationSupport.findRepeatableAnnotations()` now finds repeatable annotations used as
   meta-annotations on other repeatable annotations.
+* Exception thrown by `TestExecutionListeners` no longer cause test execution to abort.
```
marcphilipp: Note to self: There's an `s` missing here.

platform-tests/src/test/java/org/junit/platform/launcher/core/TestExecutionListenerRegistryTest.java
```diff
@@ -0,0 +1,251 @@
...
+import org.junit.platform.engine.TestExecutionResult;
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.reporting.ReportEntry;
+import org.junit.platform.engine.support.descriptor.DemoMethodTestDescriptor;
+import org.junit.platform.launcher.TestExecutionListener;
+import org.junit.platform.launcher.TestIdentifier;
+import org.junit.platform.launcher.TestPlan;
+
+@TrackLogRecords
+class TestExecutionListenerRegistryTest {
```
marcphilipp: Note to self: This should be `TestExecutionListenerRegistryTests`

documentation/src/docs/asciidoc/release-notes/release-notes-5.5.0-M2.adoc
gopinath-langote: @marcphilipp Just one doubt. Should I update with `Note to self:` comments? Or these comments are for you only?

marcphilipp: @gopinath-langote Sorry for taking so long to reply. Those were meant for myself to fix before merging this which I'll do as soon as CI passes.

marcphilipp:
Thanks for your patience and perseverance, @gopinath-langote! 👍

Issue #981, opened 2017-07-24
Introduce additional tests for support of generics


sbrannen:
## Overview

Issues #333, #976, and #978 addressed the topics of generics and overridden methods in various ways; however, as conjectured in #978, we may still have lingering bugs in this regard.

## Deliverables

For each of the following, introduce additional tests across multiple-level class hierarchies using generic interfaces, generic abstract classes, etc.

In other words, try to _break_ the search algorithms.

- [ ] `ReflectionUtils.findMethod(*)`
- [ ] `ReflectionUtils.findMethods(*)`

sbrannen:
This issue is *up for grabs* in case anyone in the community wants to take a stab at breaking our reflection utilities. 😉 

PRs with failing tests are extremely welcome!

sbrannen:
You're using the *alpha* version of JUnit 5.

We just released 5.0 M6, and 5.0 RC1 work is being performed in `master`.

sbrannen:
If you perform a fresh `git clone https://github.com/junit-team/junit5.git`, everything should build just fine.

cldssty:
I'd like to take a stab if that's okay!

sbrannen:
@cldssty, sure.... go for it!

jadeleeuw:
I found a situation that breaks the search algorithm, given that my expectation of its behaviour is correct, so forgive my ignorance if this works as intended.

Given the two interfaces below:
```
interface A {
    default void foo(Double number) {
        System.out.println("A");
    }
}

interface B<T extends Number> {
    default void foo(T number) {
        System.out.println("B");
    }
}
```
And a class:
```
class Bar implements B<Integer>, A {}
```

`findMethod(Bar.class, "foo", Double.class)` will return the `foo` method of interface B. This is because that method is the first method found by the search algorithm that satisfies the signature, even though a more "specific"/non-generic  method satisfying that signature is also available.

Turning the interfaces around (e.g. `class Bar implements A, B<Integer> {}`) does result in the `foo` from interface A being returned by `findMethod`.

I can imagine that a user in this situation might expect the most "specific"/non-generic method to be returned instead of the first method satisfying the signature, especially because invoking the `foo` method with a double will of course execute the method of interface A.

In essence: I am unsure whether the current behaviour of the search algorithm is working as you intended, or if this can be considered a bug. I can push a PR with a test case demonstrating the situation if needed.

sbrannen:
[...]

PR #1790, opened 2019-02-21
Enforce findMethod to prioritize specific over generic methods

jadeleeuw:
## Overview

As discussed in #981, `ReflectionUtils.findMethod` currently returns the first method that satisfies the signature requirements rather than the method with the most specific (non-generic) matching signature.

This can cause methods with matching generic signatures to be chosen over a matching non-generic signature. This test enforces that a matching non-generic signature is chosen over a matching generic signature.

This PR will fail the build until `ReflectionUtils.findMethod` prioritizes a method with non-generic parameters that satisfy the required signature over a method with generic parameters that satisfies the required signature. 

I think that the issue in the search algorithm lies [here](https://github.com/junit-team/junit5/blob/311bb71a793c07d7e3000a5d32204e0cf598f590/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java#L1237-L1242), where a method with a matching signature is immediately returned, making it ordering sensitive.
Perhaps instead of immediately returning the matching method, all matching methods should be saved and the most specific/non-generic method should be determined and returned?

I'd be happy to try and investigate the possibilities of such a solution.

---

---

### Definition of Done

### DIFF
--- a/platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
@@ -887,6 +887,17 @@ class ReflectionUtilsTests {
 	}
 
 	@Test
+	void findMethodChooseSpecificOverGenericMethod() throws NoSuchMethodException {
+		String methodName = "foo";
+		Class<?> parameterType = Double.class;
+		Class<?> testClass = InterfaceWithSpecificAndGenericDefaultMethodImpl.class;
+
+		Method specificMethod = testClass.getMethod(methodName, parameterType);
+		Optional<Method> optional = findMethod(testClass, methodName, parameterType);
+		assertThat(optional).contains(specificMethod);
+	}
+
+	@Test
 	void findMethodsPreconditions() {
 		// @formatter:off
 		assertThrows(PreconditionViolationException.class, () -> findMethods(null, null));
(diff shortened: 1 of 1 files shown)


sbrannen:
Thanks for creating the PR.

> I'd be happy to try and investigate the possibilities of such a solution.

Before implementing any code to address this, let's let team decide what course of action it wishes to take.

sbrannen:
Tentatively slated for 5.5 M1 for _team discussion_

platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
```diff
@@ -886,6 +886,18 @@ private void assertFindMethodByParameterNames(String methodName, Class<?> parame
...
+	@Test
+	void findMethodChooseSpecificOverGenericMethod() throws NoSuchMethodException {
+		String methodName = "foo";
+		Class<?> parameterType = Double.class;
+		Class<?> testClass = InterfaceWithSpecificAndGenericDefaultMethodImpl.class;
+
+		Method specificMethod = testClass.getMethod(methodName, parameterType);
+		Optional<Method> optional = findMethod(testClass, methodName, parameterType);
+
+		assertThat(optional).isEqualTo(specificMethod);
```
sbrannen: That will always fail since you're comparing an `Optional` to a `Method`.

I think you need to use `assertThat(optional).contains(method)`.

jadeleeuw: Woops! You are right. Oversight from my part. Will immediately address this.

Issue #1795, opened 2019-02-27
Test-drive JUnit 5 build on Azure Pipelines


sormuras:
## Overview

Test-drive our build on Azure Pipelines [1].

[1] https://azure.microsoft.com/en-us/services/devops/pipelines/

## Deliverables

- [x] Add `azure-pipelines.yml` with minimal configuration.
- [x] Does it work on Linux?
- [x] Does it work on Mac?
- [x] Does it work on Windows?
- [x] How to configure which JDK is used?
- [x] Is OpenJDK-EA supported?
- [x] On failing builds, are logs easy to access and scan?
- [x] Are Gradle Build Scans still published?
- [x] What about speed?
- [x] SNAPSHOT deployments: documentation (user-guide, javadoc) and artifacts

brunoborges:
Confirming a minimal setup for Linux works as expected.

`azure-pipeline.yml`

 - https://github.com/brunoborges/junit5/commit/4ee8d238a8647314d2925f78deaa56f6fdec6a4b

Here's a build test:

 - https://dev.azure.com/brunoborges-github/junit5/_build/results?buildId=25&view=logs

And here's the Gradle scan:

 - https://scans.gradle.com/s/6723pxshyiv2y

sormuras:
Awesome, @brunoborges -- thanks for spiking an example setup. PR? ;)

sormuras:
Superseded by #1810

PR #1796, opened 2019-02-27
Make Azure Pipeline build work on Windows/Mac/Linux

brunoborges:
## Overview

Enhances the build pipeline on Azure DevOps to use three agents (Windows, Linux, and Mac). Related to #1795

---

---

### DIFF
--- a/azure-gradle-step.yml
+++ b/azure-gradle-step.yml
@@ -1,13 +1,13 @@
 steps:
   - task: Gradle@2
-  inputs:
-    workingDirectory: ''
-    gradleWrapperFile: 'gradlew'
-    gradleOptions: '-Xmx3072m'
-    options: '--scan --warning-mode=all -Dplatform.tooling.support.tests.enabled=true'
-    javaHomeOption: 'JDKVersion'
-    jdkVersionOption: '1.11'
-    jdkArchitectureOption: 'x64'
-    publishJUnitResults: true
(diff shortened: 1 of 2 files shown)


sormuras:
Thanks for PR! All 3 platform build worked out of the box.

- Does Pipelines support a leading `.` at `azure-pipelines.yml`: `.azure-pipelines.yml`?
- Is it possible to move the `azure-gradle-step.yml` into the `src/azure` directory and reference it from there?
- Unit test should only be collected from one (say Linux) platform. Over [9.000 tests](https://dev.azure.com/brunoborges-github/junit5/_build/results?buildId=28&view=ms.vss-test-web.build-test-results-tab) are ... too much. :-)

sormuras:
> * Is it possible to move the `azure-gradle-step.yml` into the `src/azure` directory and reference it from there?

Pushed changes directly to your PR branch. Let's see how it works out...

sormuras:
> * Does Pipelines support a leading `.` at `azure-pipelines.yml`: `.azure-pipelines.yml`?

Does not work out.

brunoborges:
It supports hidden file. Any location actually. I just need to update the pipeline settings on Azure DevOps.

sormuras:
Understood and checked it works on a different repository and pipeline: https://github.com/sormuras/mainrunner/commit/2c156553752c3ed91a3823ab556e791ce4701da3

PR #1798, opened 2019-03-03
Introduce configurable discovery algorithm

### LINKED ISSUE #1739: Provide reusable base implementation of test discovery
All class-based test engines should handle the following selectors during test discovery:

- `ClasspathResourceSelector`
- `ClasspathRootSelector`
- `ModuleSelector`
- `PackageSelector`
- `UniqueIdSelector`
- `ClassSelector`
- `MethodSelector`

The code required to handle the first four of these is straightforward but repetitive and [easy to forget](https://stackoverflow.com/questions/52780940/kotlintestengine-discover-doesnt-return-any-tests). Moreover, resolving `UniqueIdSelectors` is non-trivial but the implementations of the Jupiter and Vintage engine provide examples of how to handle them correctly. Both use a similar yet slightly different approach. If we implemented the base algorithm once in an extensible way, it could be reused by other test engines. Doing so would reduce the burden on test engine authors and increase consistency across test engine implementations.

## Deliverables

- [ ] Extract an extendable base algorithm for resolving selectors from existing implementations
- [ ] Provide extensible configuration mechanism
- [ ] Provide sensible defaults for class-based test engines
- [ ] Use base algorithm in Jupiter and Vintage
- [ ] Reach out to authors of third-party test engines to get feedback and help them adopt the new base algorithm

marcphilipp:
The `EngineDiscoveryRequestResolver` class implements a test discovery
in a way that can be reused by different test engines. The Jupiter and
Vintage test engines are changed to use it.

Main entry point is `EngineDiscoveryRequestResolver.builder()` which
allows to add resolvers (via a new `SelectorResolvers` interface) and
visitors (reusing `TestDescriptor.Visitor`) that will be called to
resolve `DiscoverySelectors` in a given `EngineDiscoveryRequest`.

In order to support class based test engines, the predefined
`ClassContainerSelectorResolver` may be added via the builder as well.
It resolves `ClasspathRootSelectors`, `ModuleSelectors`, and
`PackageSelectors` into `ClassSelectors` by scanning for classes in the
corresponding class container.

The fundamental principle behind `SelectorResolver` is that both the
designated parent of a `TestDescriptor` and its designated children can
be specified using `DiscoverySelectors`. For example, in the case of the
Jupiter engine, when resolving a nested test class, there's a resolver
that specifies the parent using the `ClassSelector` of the enclosing
class and the children as the set of `MethodSelectors` of the testable
methods it contains joined with the set of `ClassSelectors` of nested
classes it contains.

Overall, the algorithm works as follows:

1. Enqueue all selectors in the supplied request to be resolved.
2. While there are selectors to be resolved, get the next one.
   Otherwise, the resolution is finished.
   1. Iterate over all registered `SelectorResolvers` in the order they
      were registered in and find the first one that returns a non-empty
      resolution.
   1. If such a resolution exists, enqueue the contained selectors.
   1. For each exact match in the resolution, expand its children and
      enqueue them as well.
   1. Iterate over all registered visitors and let the engine descriptor
      accept them.

Descriptors with the same unique ID are only resolved once, even if
specified using multiple selectors.

Resolves #1739.

## Overview

---

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/ClassSelectorResolver.java
```diff
@@ -0,0 +1,133 @@
...
+			Class<?> testClass = it.getTestClass();
+			// @formatter:off
+            return Resolution.match(Match.exact(it, () -> {
+                Stream<MethodSelector> methods = findMethods(testClass, isTestOrTestFactoryOrTestTemplateMethod).stream()
+                        .map(method -> selectMethod(testClass, method));
+				Stream<ClassSelector> nestedClasses = findNestedClasses(testClass, isNestedTestClass).stream()
+						.map(DiscoverySelectors::selectClass);
+				return Stream.concat(methods, nestedClasses).collect(toCollection((Supplier<Set<DiscoverySelector>>) LinkedHashSet::new));
+            }));
+            // @formatter:on
```
jbduncan: @marcphilipp It looks to me that the code within this `@formatter:(off|on)` block has some space/tab indenting issues. :)

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/EngineDiscoveryRequestResolution.java
```diff
@@ -0,0 +1,262 @@
...
+
+	private final Logger logger;
+	private final EngineDiscoveryRequest request;
+	private final Context defaultContext;
+	private final List<SelectorResolver> resolvers;
+	private final List<TestDescriptor.Visitor> visitors;
+	private final TestDescriptor engineDescriptor;
+	private final Map<DiscoverySelector, Resolution> resolvedSelectors = new LinkedHashMap<>();
+	private final Map<UniqueId, Match> resolvedUniqueIds = new LinkedHashMap<>();
+	private final Queue<DiscoverySelector> remainingSelectors = new LinkedList<>();
```
jbduncan: Have you considered using an `ArrayDeque` instead of a `LinkedList` here?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/ClassSelectorResolver.java
sormuras: /me mumbles _Turning formatters off in 2019..._ 😉

jbduncan: @sormuras Indeed. :wink: 

But on a more serious note, it makes me wonder if Eclipse has a setting that would allow method chains to be formatted with one method per line, without the JUnit 5 team having to manually use `@formatter(off|on)`... 🤔

(Am I right to think that using google-java-format, which IIUC would solve this problem more easily, was considered by the JUnit 5 team but decided against?)

sormuras: > ... using google-java-format ... was considered by the JUnit 5 team but decided against?

True. But I'm only allowed to bring up that topic once a year. Mentioned it last week already -- so when is 2020?! 😈

marcphilipp: @jbduncan I don't see any space/tab indenting issues. What am I missing?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/EngineDiscoveryRequestResolution.java
marcphilipp: I haven't TBH. Do you think it would make a difference in this scenario performance-wise?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/ClassSelectorResolver.java
sormuras: image

sormuras: Could be an GitHub display only issue. Didn't check the whitespace chars in my IDE...

jbduncan: Indeed it could be just a GitHub issue; I don't have an IDE with this PR checked out to confirm. :sweat_smile:

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/EngineDiscoveryRequestResolution.java
jbduncan: I'm not sure, actually. I personally prefer `ArrayDeque` because its Javadoc says that it's "faster than `LinkedList` when used as a queue", so when in doubt, that's the implementation I go for.

But having said this, I don't know how `ArrayDeque` using an array for storage would impact its memory consumption on a long-running instance of the JUnit Platform (as in, I don't know (1) if the queue will need to live for a "long" time, and (2) whether it's possible for the queue to get very long and then empty, causing a memory leak if the array isn't resized back to a small size).

So I'd say that using a `LinkedList` here is perfectly acceptable until we have evidence that it's a performance bottleneck or a memory hog. :)

marcphilipp: I've now opted for `ArrayDeque` for the reason you mentioned. The queue is short-lived and will only live for the duration of test discovery so we shouldn't have to worry about it being a memory leak.

jbduncan: Sounds good to me @marcphilipp! Cheers. 👍

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/ClassSelectorResolver.java
sbrannen: Nope: not a GitHub. They were spaces instead of tabs. Fixed here: f114fd009778bb4a33cda97a42bdf68d71348221 😉

jbduncan: Haha, thank you for confirming and cleaning it up, @sbrannen. :smile:

Issue #1810, opened 2019-03-12
Polish Azure Pipelines build


sormuras:
After-math for #1807 and closes #1795.

## Deliverables

- [x] Provide credentials to publish SNAPSHOT assets (artifacts and documentation)
- [x] Update https://github.com/junit-team/junit5/blob/master/README.md#continuous-integration-builds table with shields pointing to Azure Pipelines
- [x] Configure "Publish Test Results Job" to include only propery results: https://github.com/junit-team/junit5/pull/1807#discussion_r264530632
- [x] Configure "Checkout task" to only checkout the current branch
- [x] The two last jobs (publish snapshot assets) should only start if "Linux (JDK 11)" was successfully executed
- [x] Remove AppVeyor configuration
- [x] Uninstall/disconnect AppVeyor service
- [x] Remove Travis CI configuration
- [x] Uninstall/disconnect Travis CI service
- [x] ~Configure dependency cache on Azure Pipelines~ will be done in https://github.com/junit-team/junit5/issues/1826

sormuras:
Added the "overall pipeline status badge" via

It is possible to show off all (or some) jobs in a table:

- Linux (OpenJDK 11) 🤜 [![Build Status](https://dev.azure.com/junit-team/junit5/_apis/build/status/junit-team.junit5?branchName=master&jobName=Linux%20(OpenJDK%2011))](https://dev.azure.com/junit-team/junit5/_build/latest?definitionId=1&branchName=master)
- Linux (OpenJDK 12) 🤜 [![Build Status](https://dev.azure.com/junit-team/junit5/_apis/build/status/junit-team.junit5?branchName=master&jobName=Linux%20(OpenJDK%2012))](https://dev.azure.com/junit-team/junit5/_build/latest?definitionId=1&branchName=master)
- Linux (OpenJDK 13) 🤜 [![Build Status](https://dev.azure.com/junit-team/junit5/_apis/build/status/junit-team.junit5?branchName=master&jobName=Linux%20(OpenJDK%2013))](https://dev.azure.com/junit-team/junit5/_build/latest?definitionId=1&branchName=master)

sormuras:
> Click **Advanced settings** in the **Get Sources** task to see some of the above options.

I can't find the "Advanced settings..." described here https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/pipeline-options-for-git?view=azure-devops to activate

ethomson:
[...]

PR #1818, opened 2019-03-16
ci: only publish on success

ethomson:
Only run the publish steps when the build was successful, per https://github.com/junit-team/junit5/issues/1810#issuecomment-473516522.

Note that since the publish dependencies _only_ depend on the `linux_11` build, it's possible to publish even if the macOS (for instance) build fails.  I'm happy to tweak this if requested.

---

---

### Definition of Done

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -93,7 +93,7 @@ jobs:
       env:
         ORG_GRADLE_PROJECT_nexusUsername: $(nexusUsername)
         ORG_GRADLE_PROJECT_nexusPassword: $(nexusPassword)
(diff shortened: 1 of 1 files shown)


sormuras:
> Note that since the publish dependencies only depend on the linux_11 build, it's possible to publish even if the macOS (for instance) build fails. I'm happy to tweak this if requested.

I'm fine with that "as-is" for now, as `linux_11` is our reference build and performs all required tests.

Would you mind rewording the [commit message](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#commit-messages) to read sth. like the following?
```
Only run publish steps when build was successful

Issue: #1810
```
https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#commit-messages

ethomson:
> Would you mind rewording the commit message to read sth. like the following?

Fixed!

sormuras:
Thanks and merged in a minute...

PR #1839, opened 2019-03-29
Make emptyValue configurable in @CsvSource and @CsvFileSource

### LINKED ISSUE #1788: Make emptyValue configurable in @CsvSource and @CsvFileSource
I did not find a way to set the emptyValue of CsvParserSettings in case of having "null" and not "" for empty values.

For example, the values could be 
`1;null;20.7;6.75;9; 6.25`
And it is not neccessary to have
`1;;20.7;6.75;9; 6.25`

I could provide a pull request with the proposal that at the end could be so configured:

`@CsvFileSource(numLinesToSkip = 1, delimiter = ';', emptyValue='null', resources =[....]`

By default is
`@CsvFileSource(numLinesToSkip = 1, delimiter = ';', emptyValue='', resources =[....]`
`@CsvFileSource(numLinesToSkip = 1, delimiter = ';', resources =[....]`

Ciruman:
## Overview

When using `@CsvSource` or `@CsvFileSource` in a parameterized test,
if the CSV parser does not read any character from the input, and the
input is within quotes, and emptyValue is configured, an emptyValue is
returned instead of default `""`.

---

---

### Definition of Done

Closes #1788

---

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
@@ -76,7 +76,10 @@ public @interface CsvFileSource {
 	 * <p>Typically used to skip header lines.
 	 *
 	 * <p>Defaults to {@code 0}.
+	 *
+	 * @since 5.1
 	 */
+	@API(status = EXPERIMENTAL, since = "5.1")
 	int numLinesToSkip() default 0;
 
 	/**
@@ -85,7 +88,10 @@ public @interface CsvFileSource {
 	 * <p>This value replaces quoted empty strings read from the input.
 	 *
 	 * <p>Defaults to {@code ""}.
+	 *
+	 * @since 5.5
 	 */
+	@API(status = EXPERIMENTAL, since = "5.5")
 	String emptyValue() default "";
 
 }
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
@@ -80,11 +80,11 @@ public @interface CsvFileSource {
 	int numLinesToSkip() default 0;
 
 	/**
-	 * The empty value definition to use when reading the CSV files.
+	 * The empty value to use when reading the CSV files.
 	 *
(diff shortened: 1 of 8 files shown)


junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
```diff
@@ -79,4 +79,11 @@
 	 */
 	int numLinesToSkip() default 0;
 
+	/**
+	 * The empty value definition to use when reading the CSV files.
```
sbrannen: Please add a second paragraph explaining when the empty value is used.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvSource.java
```diff
@@ -55,4 +55,11 @@
 	 */
 	char delimiter() default ',';
 
+	/**
+	 * The empty value definition to use when reading the {@linkplain #value lines}.
```
sbrannen: Please add a second paragraph explaining when the empty value is used.

Ciruman:
> Please add an entry to the Release Notes and document the new feature in the corresponding section of the User Guide.

I don´t find where the other arguments are documented( I expected here: writing-test.md), now I am wondering if I document the other arguments(delimiter, numLinesToSkip, etc.) also or just update the @CsvFileSource note is enough, like this:

```
NOTE: In contrast to the syntax used in `@CsvSource`, `@CsvFileSource` uses a double
quote `"` as the quote character. See the `"United States of America"` value in the
example above. An empty, quoted value `""` results in an empty `String` unless emptyValue
is set; whereas, an entirely _empty_ value is interpreted as a `null` reference. An
`ArgumentConversionException` is raised if the target type of a `null` reference is a
primitive type.
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
Ciruman: Done

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvSource.java
Ciruman: Done

Ciruman:
> Looking good! I polished the documentation a bit further.

@marcphilipp Thank you!

marcphilipp:
Thanks, @Ciruman! 👍

PR #1873, opened 2019-05-10
[Cleanup] Simplify use of if(Optional.isPresent()) using functional style

ielatif:
Simple cleanup.

---


sormuras:
Thanks, @ielatif! Merged with a modified the commit message.

PR #1883, opened 2019-05-17
Add null value attribute to CsvSource and CsvFileSource

### LINKED ISSUE #1856: Add nullValue attribute to @CsvSource and @CsvFileSource
## Description

The idea is when parsing the values, null will be also returned if the value is equals to the string.

I think this parameter name could be nullValue.

Currently:
1;;20.7;6.75;9; 6.25 -> Second Value(not present) is null

In this issue, I would like to address the possibility to set an alternative to "not present/null" values

1;null;20.7;6.75;9; 6.25 here, if nullValue is set to "null", second value after parsing would be null
1;vacio;20.7;6.75;9; 6.25 here, if nullValue is set to "vacio", second value after parsing would be null
1;nill;20.7;6.75;9; 6.25 here, if nullValue is set to "nill", second value after parsing would be null

Then a test would look like this:
```java
    @Test
    void emptyValueIsAnEmptyWithCustomNullValueString() {
        Stream<Object[]> arguments = provideArguments(',', "","empty", "vacio , , empty , ''");

        assertThat(arguments).containsExactly(new String[] { "vacio", null, null, "" });
    }
    private Stream<Object[]> provideArguments(char delimiter, String emptyValue, String nullValue, String... value) {
        CsvSource annotation = mock(CsvSource.class);
        when(annotation.value()).thenReturn(value);
        when(annotation.delimiter()).thenReturn(delimiter);
        when(annotation.emptyValue()).thenReturn(emptyValue);
        when(annotation.nullValue()).thenReturn(nullValue);

        CsvArgumentsProvider provider = new CsvArgumentsProvider();
        provider.accept(annotation);
        return provider.provideArguments(null).map(Arguments::get);
    }
```

It could be used like this:

```java
@CsvSource(value = {"null,null,null,null", "null,2,null,null", "1,null,null,null", "20,110,true,110", "20,110,false,20"}, nullValue = "null")
```
Parsed to:
```java
null null null null 
null 2 null null
1 null null null
20 110 true 110
20 110 false 20
```

with the same meaning to:

```java
@CsvSource({",,,", ",2,,", "1,,,", "20,110,true,110", "20,110,false,20"})
```
Parsed to:
```java
null null null null 
null 2 null null
1 null null null
20 110 true 110
20 110 false 20
```

In case of having a custom nullValue configured. I think, in case of finding no value, should be returned null too.
```java
@CsvSource(value = {",,,", "null,2,null,null", "1,null,null,null", "20,110,true,110", "20,110,false,20"}, nullValue = "null")
```
Parsed to:
```java
null null null null 
null 2 null null
1 null null null
20 110 true 110
20 110 false 20
```
## Related issue
https://github.com/junit-team/junit5/issues/1788

Farbauti89:
## Overview
This PR implements the proposed enhancement in https://github.com/junit-team/junit5/issues/1856 .

It will introduce a `nullSymbols` attribute in the `CsvSource` and `CsvFileSource` Annotation. 
The `nullSymbols` will be converted into `null` if it is found in the CSV.

Resolves #1856 

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
@@ -35,7 +35,6 @@ on GitHub.
 * Exceptions thrown by test engines during discovery and execution are now reported to
   `TestExecutionListeners`.
 
-
 [[release-notes-5.6.0-M1-junit-jupiter]]
 === JUnit Jupiter
 
@@ -59,6 +58,7 @@ on GitHub.
 * New `TypeBasedParameterResolver<T>` abstract base class that serves as a generic adapter
   for the `ParameterResolver` API and simplifies the implementation of a custom resolver
   that supports parameters of a specific type.
+* New `nullSymbols` attribute in `@CsvFileSource` and `@CsvSource`.
 
 
 [[release-notes-5.6.0-M1-junit-vintage]]
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1149,17 +1149,20 @@ cannot be set simultaneously.
 `@CsvSource` uses a single quote `'` as its quote character. See the `'lemon, lime'` value
 in the example above and in the table below. An empty, quoted value `''` results in an
 empty `String` unless the `emptyValue` attribute is set; whereas, an entirely _empty_
-value is interpreted as a `null` reference. An `ArgumentConversionException` is raised if
-the target type of a `null` reference is a primitive type.
+value is interpreted as a `null` reference.
+By specifying one or more `nullSymbols` any value can be interpreted as a `null` reference.
+An `ArgumentConversionException` is raised if the target type of a `null` reference
+is a primitive type.
 
 [cols="50,50"]
 |===
-| Example Input                            | Resulting Argument List
+| Example Input                                                             | Resulting Argument List
 
-| `@CsvSource({ "apple, banana" })`        | `"apple"`, `"banana"`
-| `@CsvSource({ "apple, 'lemon, lime'" })` | `"apple"`, `"lemon, lime"`
-| `@CsvSource({ "apple, ''" })`            | `"apple"`, `""`
-| `@CsvSource({ "apple, " })`              | `"apple"`, `null`
+| `@CsvSource({ "apple, banana" })`                                         | `"apple"`, `"banana"`
+| `@CsvSource({ "apple, 'lemon, lime'" })`                                  | `"apple"`, `"lemon, lime"`
+| `@CsvSource({ "apple, ''" })`                                             | `"apple"`, `""`
+| `@CsvSource({ "apple, " })`                                               | `"apple"`, `null`
+| `@CsvSource(value = { "apple, banana, NaN" }, nullSymbols = {"NaN"})`     | `"apple"`, `"banana"`, `null`
 |===
 
 [[writing-tests-parameterized-tests-sources-CsvFileSource]]
@@ -1192,8 +1195,9 @@ NOTE: In contrast to the syntax used in `@CsvSource`, `@CsvFileSource` uses a do
 quote `"` as the quote character. See the `"United States of America"` value in the
 example above. An empty, quoted value `""` results in an empty `String` unless the
 `emptyValue` attribute is set; whereas, an entirely _empty_ value is interpreted as a
-`null` reference. An `ArgumentConversionException` is raised if the target type of a
-`null` reference is a primitive type.
+`null` reference. By specifying one or more `nullSymbols` any value can be interpreted
+as a `null` reference. An `ArgumentConversionException` is raised if the target type
+of a `null` reference is a primitive type.
 
 [[writing-tests-parameterized-tests-sources-ArgumentsSource]]
 ===== @ArgumentsSource
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
@@ -15,6 +15,9 @@ import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Stream;
 
+import com.univocity.parsers.common.DefaultConversionProcessor;
+import com.univocity.parsers.common.processor.ObjectRowListProcessor;
+import com.univocity.parsers.conversions.Conversions;
 import com.univocity.parsers.csv.CsvParser;
 
 import org.junit.jupiter.api.extension.ExtensionContext;
@@ -41,25 +44,38 @@ class CsvArgumentsProvider implements ArgumentsProvider, AnnotationConsumer<CsvS
 
 	@Override
 	public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+
 		AtomicLong index = new AtomicLong(0);
+		DefaultConversionProcessor processor = getConversionProcessor();
 
 		// @formatter:off
-		return Arrays.stream(this.annotation.value())
-				.map(line -> {
-					String[] parsedLine = null;
-					try {
-						parsedLine = this.csvParser.parseLine(line + LINE_SEPARATOR);
-					}
-					catch (Throwable throwable) {
-						handleCsvException(throwable, this.annotation);
-					}
-					Preconditions.notNull(parsedLine,
-						() -> "Line at index " + index.get() + " contains invalid CSV: \"" + line + "\"");
-					return parsedLine;
-				})
-				.peek(values -> index.incrementAndGet())
-				.map(Arguments::of);
-		// @formatter:on
+        return Arrays.stream(this.annotation.value())
+                .map(line -> {
+                    String[] parsedLine = null;
+                    try {
+                        parsedLine = this.csvParser.parseLine(line + LINE_SEPARATOR);
+                        if (parsedLine != null) {
+                            parsedLine = Arrays.copyOf(processor.applyConversions(parsedLine, null), parsedLine.length, String[].class);
+                        }
+                    } catch (Throwable throwable) {
+                        handleCsvException(throwable, this.annotation);
+                    }
+                    Preconditions.notNull(parsedLine,
+                                          () -> "Line at index " + index.get() + " contains invalid CSV: \"" + line + "\"");
+                    return parsedLine;
+                })
+                .peek(values -> index.incrementAndGet())
+                .map(Arguments::of);
+        // @formatter:on
+	}
+
+	private DefaultConversionProcessor getConversionProcessor() {
+		ObjectRowListProcessor processor = new ObjectRowListProcessor();
+		if (this.annotation.nullSymbols().length > 0) {
+			processor.convertAll(Conversions.toNull(this.annotation.nullSymbols()));
+		}
+
+		return processor;
 	}
 
 	static void handleCsvException(Throwable throwable, Annotation annotation) {
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileArgumentsProvider.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileArgumentsProvider.java
@@ -23,6 +23,9 @@ import java.util.Spliterator;
 import java.util.function.BiFunction;
 import java.util.stream.Stream;
 
+import com.univocity.parsers.common.DefaultConversionProcessor;
+import com.univocity.parsers.common.processor.ObjectRowListProcessor;
+import com.univocity.parsers.conversions.Conversions;
 import com.univocity.parsers.csv.CsvParser;
 
 import org.junit.jupiter.api.extension.ExtensionContext;
@@ -116,11 +119,14 @@ class CsvFileArgumentsProvider implements ArgumentsProvider, AnnotationConsumer<
 
 		private final CsvFileSource annotation;
 
+		private final DefaultConversionProcessor conversionProcessor;
+
 		private Object[] nextCsvRecord;
 
 		CsvParserIterator(CsvParser csvParser, CsvFileSource annotation) {
 			this.csvParser = csvParser;
 			this.annotation = annotation;
+			this.conversionProcessor = getConversionProcessor();
 			advance();
 		}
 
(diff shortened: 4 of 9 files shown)


junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
```diff
@@ -94,4 +94,6 @@
 	@API(status = EXPERIMENTAL, since = "5.5")
 	String emptyValue() default "";
 
+	String[] nullSymbol() default {};
```
sormuras: - `@API(status = EXPERIMENTAL, since = "5.5")` is missing.
- As the return type is an array of `String`, the method should be called `nullSymbols`, ending with a `s` ... 🤔

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvSource.java
```diff
@@ -67,4 +67,5 @@
 	@API(status = EXPERIMENTAL, since = "5.5")
 	String emptyValue() default "";
 
+	String[] nullSymbol() default {};
```
sormuras: Same as above: missing `@API` and `s`

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -47,26 +50,40 @@ public void accept(CsvSource annotation) {
 		settings.getFormat().setQuoteEscape('\'');
 		settings.setEmptyValue(this.annotation.emptyValue());
 		settings.setAutoConfigurationEnabled(false);
+
 		CsvParser csvParser = new CsvParser(settings);
 		AtomicLong index = new AtomicLong(0);
 
+		final DefaultConversionProcessor processor = getConversionProcessor();
```
sormuras: - Remove `final` modifier.
- Why are did use a `DefaultConversionProcessor` as the type of the variable instead of `ConversionProcessor`?

Farbauti89:
Hello Maintainers,

I am a first time commiter in this project and not sure about the "Definition of Done".
Do you tick the checkboxes when you check the PR or is this something I should do?

Also this is a PreView PR because I wanted to make sure the change is cool for you.
If it's okay I will wait with the documentation until then.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
Farbauti89: Thanks for the reply!
I wasn't really sure about the `@API` Annotation. I guess this is the following check from the "Definition of Done"?

>  Method preconditions are checked and documented in the method's Javadoc

About the plural of `nullSymbol`: You are totally right!

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvSource.java
Farbauti89: Thanks again - I also fixed this here.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
Farbauti89: The `final` modifier is gone.

I can't use the `ConversionProcessor` because the `applyConversions` Method is not part of this Interface. Therefore the `DefaultConversionProcessor` is the most common class that I could have used.

I also thought that `ConversionProcessor` should define `applyConversions`. However this is mainly because of the name of this interface. The documentation of the class is pretty specific about what its intend is:

> A basic interface for classes that associate {@link Conversion} implementations with fields of a given input/output.

sormuras:
Please run 'gradlew spotlessApply' to fix source code format violations automatically.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
sormuras: Understood.

Farbauti89:
@sormuras i fixed the formatting issue.
Sorry for the inconvenience :(

I also have a questions about the "Definition of Done".
Do you as a maintainer tick the checkboxes or is this something I should do when I think the PR is ready?

sormuras:
Thanks for the PR! 👍 

> Do you as a maintainer tick the checkboxes or is this something I should do when I think the PR is ready?

They are an indicator how for a PR is. Check them when you think a task is done. We'll go over them and double-check anyway.

Ciruman:
I like how you solved the problem with the processor and looks good to me.  Thank you for the PR!

Farbauti89:
> Looks good overall, left some comments in-line.

-@sormuras 

Hello @sormuras , I think I'am done with this PR.
Github tells me that you still request one change. I think it's the comment above.
If there is anything else to fix let me know :)

Ciruman:
As an improvement, in case of having nullSymbols = {}, there is no need to create a processor and there is no need to process the pasedLine in both CSV Argument Providers.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileArgumentsProvider.java
```diff
@@ -124,11 +127,14 @@ private CsvParser createCsvParser(InputStream inputStream) {
...
 		private final CsvFileSource annotation;
 
+		private final DefaultConversionProcessor conversionProcessor;
+
 		private Object[] nextCsvRecord;
 
 		CsvParserIterator(CsvParser csvParser, CsvFileSource annotation) {
 			this.csvParser = csvParser;
 			this.annotation = annotation;
+			this.conversionProcessor = getConversionProcessor();
```
Ciruman: If this.annotation.nullSymbols() is equals to default value, the conversionProcessor could be skiped.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -47,26 +50,40 @@ public void accept(CsvSource annotation) {
 		settings.getFormat().setQuoteEscape('\'');
 		settings.setEmptyValue(this.annotation.emptyValue());
 		settings.setAutoConfigurationEnabled(false);
+
 		CsvParser csvParser = new CsvParser(settings);
 		AtomicLong index = new AtomicLong(0);
 
+		DefaultConversionProcessor processor = getConversionProcessor();
```
Ciruman: If this.annotation.nullSymbols() is equals to default value, the conversionProcessor could be skiped.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileArgumentsProvider.java
Farbauti89: Hi @Ciruman,
thanks for the suggestion! 
Before implementing it I would like to explain why I wrote the code the way I did :)

I tried to implement my solution wit Null-Objects in mind. If the `nullSymbols` is an empty Array I still instantiate a `DefaultConversionProcessor` but because I don't add a `Conversion` to it it won't have any behaviour.
This way I don't have to introduce any more Null-Checks. I think it's a trade-off we have to find.
**Better readability** (imho) **vs Performance** (without making any Benchmarks)

Another way could be to use an `Optional<DefaultConversionProcessor>`.
This would result in a code like this when the conversion needs to be done:

 ```
AtomicReference<String[]> wrapper = new AtomicReference<>(parsedLine);
Object[] processedLine = optionalProcessor.map(defaultConversionProcessor -> defaultConversionProcessor.applyConversions(wrapper.get(), null)).orElse(wrapper.get());
parsedLine = Arrays.copyOf(processedLine, parsedLine.length, String[].class);
```

Which is imho even harder to read.

What are your thoughts?
(cc @sormuras)

File: junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java (hunk 5 above)
Comment (Farbauti89) [reply junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java]: Hi @Ciruman,

have a look at [my other comment](https://github.com/junit-team/junit5/pull/1883#discussion_r286417700) for my explanation.

**Comment by sbrannen** [comment]
@Ciruman, this PR now contains a lot of noise.

Can you please rebase your work on `master`, squash into a single commit, and force push the changes?

**Comment by Ciruman** [comment]
@sbrannen I did not create the PR. The PR was created by @Farbauti89 . @Farbauti89, can you update the PR?

**Comment by sbrannen** [comment]
> @sbrannen, I did not create the PR.

Indeed. Sorry about mixing up the user names.

**Comment by Farbauti89** [comment]
Hi @sbrannen,

> this PR now contains a lot of noise.
> 
> Can you please rebase your work on `master`, squash into a single commit, and force push the changes?

I did, as you commanded - at least I hope so :)

File: documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc [hunk 6]
```diff
@@ -28,6 +28,8 @@ on GitHub.
 
 * New `printFailuresTo(PrintWriter, int)` method in `TestExecutionSummary` that allows one
   to specify the maximum number of lines to print for exception stack traces.
+* New `nullSymbols` attribute in `@CsvFileSource` and `@CsvSource`.
```
sbrannen: This belongs in the JUnit Jupiter section, not the JUnit Platform section.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1125,17 +1125,18 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=CsvSource_example]
 `@CsvSource` uses a single quote `'` as its quote character. See the `'lemon, lime'` value
 in the example above and in the table below. An empty, quoted value `''` results in an
 empty `String` unless the `emptyValue` attribute is set; whereas, an entirely _empty_
-value is interpreted as a `null` reference. An `ArgumentConversionException` is raised if
-the target type of a `null` reference is a primitive type.
+value is interpreted as a `null` reference. By specifying one or more `nullSymbols` any value can be interpreted as a `null` reference.
```
sbrannen: Please wrap at 90 characters in Asciidoc files.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -47,26 +50,40 @@ public void accept(CsvSource annotation) {
...
-						handleCsvException(throwable, this.annotation);
-					}
-					Preconditions.notNull(parsedLine,
-						() -> "Line at index " + index.get() + " contains invalid CSV: \"" + line + "\"");
-					return parsedLine;
-				})
-				.peek(values -> index.incrementAndGet())
-				.map(Arguments::of);
-		// @formatter:on
+        return Arrays.stream(this.annotation.value())
```
sbrannen: Please use tabs instead of spaces for indentation.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
```diff
@@ -94,4 +94,14 @@
 	@API(status = EXPERIMENTAL, since = "5.5")
 	String emptyValue() default "";
 
+	/**
+	 * A List of Strings that should be interpreted as {@code null}.
+	 *
+	 * <p>Defaults to {@code {}}.
+	 *
+	 * @since 5.5
```
sbrannen: 5.5 has already been released, so please update to 5.6 accordingly.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
```diff
@@ -94,4 +94,14 @@
 	@API(status = EXPERIMENTAL, since = "5.5")
 	String emptyValue() default "";
 
+	/**
+	 * A List of Strings that should be interpreted as {@code null}.
+	 *
+	 * <p>Defaults to {@code {}}.
```
sbrannen: Is that syntactically correct for Javadoc?

sbrannen:
Tentatively slated for 5.6 M1

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
Farbauti89: I'm not sure what you exactly mean?
The Paragraph-Tag should be fine. It is used everywhere else even in other Projects.

If you mean the `{}` part I also think it's okay. At least the preview in my IDE and the generated HTML looks good.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
Farbauti89: The version is adjusted.
Also done in the `CsvSource`

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
Farbauti89: Sorry for mixing the indentations. I switched them back to tabs.

documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
Farbauti89: I moved the documentation to the correct section.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
sbrannen: Yes, I was referring to the `{}` part.

Eclipse displays that as `{}}` in generated Javadoc.

But... if the HTML generated by the build renders it as `{}`, then that's fine.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileSource.java
sbrannen: Thanks

sbrannen:

In general, I think the PR looks pretty good, but I have not performed an in-depth technical review of the implementation yet.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvFileArgumentsProvider.java
sormuras: I follow your line of reasoning and am fine with the better readability your approach provides. If a real performance issue arises from this implementation issue, we may tackle (proof and improve) it later.

Farbauti89:
@sormuras I just rebased the PR.

sormuras:
Thanks. Reviewing today.

sormuras:
Merged. Thanks again, @Farbauti89 🎉

sbrannen:
FYI: the implementation of this feature has been revised in 10065175af4660f12853c6067503f5af292e8980 and f239cb4ba9256f7f5a5d4b50caee5447dade3781.

sbrannen:
... and even further revised in 3efdced662ef1b4b741139c58874d5dac9607d50 and 179d82f9823ea3246cf89b6cecb2b5b5c163cd40. 😉

Farbauti89:
@sbrannen the implementation you did in https://github.com/junit-team/junit5/commit/10065175af4660f12853c6067503f5af292e8980 is far simpler. Now that I see it I am really wondering why I implemented this behaviour with univocity 🤦‍♂  

The other changes also improve the readability.

Thanks for notifying me about this. I really think I learned something :)

sbrannen:
> @sbrannen the implementation you did in [1006517](https://github.com/junit-team/junit5/commit/10065175af4660f12853c6067503f5af292e8980) is far simpler. Now that I see it I am really wondering why I implemented this behaviour with univocity 🤦‍♂

No worries. We, as software engineers, often do our best to avoid reinventing the wheel (which is a good thing). So if a library already has a feature we need, it's tempting to just use it.

In this case, none of us realized there was a simpler (and more efficient) solution while reviewing the PR in the browser. I only noticed it while reviewing the changes in my IDE after the PR had been merged into `master`.

> Thanks for notifying me about this. I really think I learned something :)

You're welcome. And... I also learned a lot more about the internals of Univocity in the process. So we all learned something. 😉

PR #1885, opened 2019-05-17
Introduce declarative timeouts

### LINKED ISSUE #80: Introduce global timeouts for tests
## Feature Request

Currently, in JUnit 4 it is not possible to have global timeouts for the tests. The `Timeout` rule does not guarantee that any `@After` method will be called. See

https://github.com/junit-team/junit/issues/1231#issuecomment-167437481

and the recently updated

https://github.com/junit-team/junit/wiki/Timeout-for-tests

It would be good to have a global timeout in JUnit Jupiter. Even better, it would be great if it would be possible to set different global timeouts for `@BeforeEach` `@Test`, and `@AfterEach`, which could be overridden (if needed) directly in the test files via annotations.

## Related Issues

- #20 
- #1246

marcphilipp:
This PR introduces declarative timeouts for testable and lifecycle methods. The API surface consists of a new `@Timeout` implementation that can be applied to `@Test`, `@TestTemplate`, `@TestFactory`, and lifecycle methods. In addition, it can be applied to a test class which has the same meaning as declaring it on all of its testable, but not its lifecycle methods.

Moreover, default timeouts can be defined using configuration parameters for all of the above methods, all lifecycle methods, all testable methods, or each type individually.

The timeout is implemented in a preemptive manner, but without having to execute the test in a different thread. Instead, a single-thread `ExecutorService` is used to schedule interrupting the original test thread. This way, `ThreadLocal` will continue to work which is important for integration tests that use frameworks like Spring and its `@Transactional` support.

Resolves #80.

---

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutDurationParser.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+class TimeoutDurationParser {
+
+	private static final Pattern PATTERN = Pattern.compile("([1-9]\\d*)((?:[nμm]?s)|m|h|d)?",
```
sormuras: Should the pattern support an optional white-spaces between the number and the unit string?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutInvocation.java
```diff
@@ -0,0 +1,86 @@
...
+	public T proceed() throws Throwable {
+		InterruptTask interruptTask = new InterruptTask(Thread.currentThread());
+		ScheduledFuture<?> future = executor.schedule(interruptTask, timeout.getValue(), timeout.getUnit());
+		Throwable failure = null;
+		T result = null;
+		try {
+			result = delegate.proceed();
+		}
+		catch (Throwable t) {
+			failure = t;
```
jbduncan: I wonder what should happen here if the throwable is blacklisted. IMO in that situation it should be thrown as-is so that if, for example, it's an OOME, then the JVM can crash relatively gracefully.

@marcphilipp Thoughts?

marcphilipp: Agreed, we shouldn't wrap blacklisted throwables like OOMEs. 👍

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutDurationParser.java
marcphilipp: I don't feel strongly about this. @junit-team/junit-lambda Thoughts?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+	private Optional<TimeoutDuration> parseTimeoutDuration(String propertyName) {
+		return cache.computeIfAbsent(propertyName, key -> extensionContext.getConfigurationParameter(key).map(value -> {
+			try {
+				return parser.parse(value);
+			}
+			catch (Exception e) {
+				logger.warn(e,
+					() -> String.format("Ignored invalid timeout '%s' set via the '%s' configuration parameter.", value,
+						key));
+				return null;
```
sbrannen: That violates the contract for `ConcurrentHashMap`:

> Like Hashtable but unlike HashMap, this class does not allow null to be used as a key or value.

Or am I missing something?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutDurationParser.java
sbrannen: I think it should support whitespace or perhaps just a single space.

After all, the `toString()` implementation inserts a space. 😉

marcphilipp: Ok, let's allow a single space.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutConfiguration.java
marcphilipp: Actually, I did have a test for this and

> If the mapping function returns {@code null}, no mapping is recorded.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutExtensionTests.java
```diff
@@ -105,25 +105,29 @@ void appliesTimeoutOnAnnotatedTestFactoryMethods() {
...
-			Execution execution = findExecution(results.all(), displayName);
-			assertThat(execution.getDuration()) //
-					.isGreaterThanOrEqualTo(Duration.ofMillis(10)) //
-					.isLessThan(Duration.ofSeconds(1));
-			assertThat(execution.getTerminationInfo().getExecutionResult().getThrowable().orElseThrow()) //
-					.isInstanceOf(TimeoutException.class) //
-					.hasMessageEndingWith("timed out after 10000000 nanoseconds");
-		});
+	Stream<DynamicTest> appliesTimeoutOnTestableMethodsInAnnotatedClasses() {
+		return Stream.of(TimeoutAnnotatedClassTestCase.class, InheritedTimeoutAnnotatedClassTestCase.class).map(
```
jbduncan: @marcphilipp I wonder if it would be feasible to use `@ParameterizedTest` here? Or would it cause a package or dependency cycle, or is it tricky to parameterize on instances of type `Class`? :)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutDurationTests.java
```diff
@@ -0,0 +1,40 @@
...
+		var oneSecond = new TimeoutDuration(1, SECONDS);
+
+		assertThat(oneSecond) //
+				.isEqualTo(oneSecond) //
+				.isEqualTo(new TimeoutDuration(1, SECONDS)) //
+				.hasSameHashCodeAs(new TimeoutDuration(1, SECONDS)) //
+				.isNotEqualTo("foo") //
+				.isNotEqualTo(new TimeoutDuration(2, SECONDS)) //
+				.isNotEqualTo(new TimeoutDuration(1, MINUTES));
+	}
```
jbduncan: @marcphilipp Have you considered using [EqualsVerifier](https://jqno.nl/equalsverifier/manual/) here? :)

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutConfiguration.java
sbrannen: Ahhh... good to know. Thanks!

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
```diff
@@ -0,0 +1,199 @@
...
+
+	private void interceptLifecycleMethod(Invocation<Void> invocation,
+			ReflectiveInvocationContext<Method> invocationContext, ExtensionContext extensionContext,
+			TimeoutProvider defaultTimeoutProvider) throws Throwable {
+		TimeoutDuration timeoutConfiguration = readTimeoutFromAnnotation(
+			Optional.of(invocationContext.getExecutable())).orElse(null);
+		intercept(invocation, invocationContext, extensionContext, timeoutConfiguration, defaultTimeoutProvider);
+	}
+
+	private Optional<TimeoutDuration> readTimeoutFromAnnotation(Optional<AnnotatedElement> executable) {
```
sbrannen: executable --> element

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	private TimeoutDuration getDefaultTimeout(ExtensionContext extensionContext,
+			TimeoutProvider defaultTimeoutProvider) {
+		return defaultTimeoutProvider.apply(getGlobalTimeoutConfiguration(extensionContext)).orElse(null);
+	}
+
+	private TimeoutConfiguration getGlobalTimeoutConfiguration(ExtensionContext extensionContext) {
+		ExtensionContext root = extensionContext.getRoot();
+		return root.getStore(NAMESPACE).getOrComputeIfAbsent("global_timeout_config",
```
sbrannen: Please make "global_timeout_config" a constant.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutExtensionTests.java
sbrannen: > I wonder if it would be feasible to use `@ParameterizedTest` here? Or would it cause a package or dependency cycle, 

That results in a project cycle in Eclipse, since we have not configured Eclipse's (relatively) new support for separating test sources/dependencies.

> or is it tricky to parameterize on instances of type `Class`? :)

That's actually easy: `@ValueSource(classes = { ... })`. In fact, I just used that feature today: https://github.com/spring-projects/spring-framework/commit/50cc23ca55293036e950260b8a90f4b4a3472fff#diff-442e1385f74c82823d0c9520ec395febR70

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	@Override
+	public void interceptAfterAllMethod(Invocation<Void> invocation,
+			ReflectiveInvocationContext<Method> invocationContext, ExtensionContext extensionContext) throws Throwable {
+		interceptLifecycleMethod(invocation, invocationContext, extensionContext,
+			TimeoutConfiguration::getDefaultAfterAllMethodTimeout);
+	}
+
+	private void readAndStoreTimeout(ExtensionContext context) {
```
sbrannen: After some mental exercise (while reviewing in the browser), I understand that this is correctly supporting lookups at the class level and method level and storing the result in the correct level in the extension context hierarchy, but... I think we should add some "developer documentation" here to make that more transparent for the community and our future selves. 😉

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutExtensionTests.java
jbduncan: Ah cool, thanks for enlightening me regarding both Eclipse and classes in annotations, @sbrannen! :+1:

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
marcphilipp: I've move the method closer to where it's being used and renamed it. I hope that will do. 🙂

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutDurationTests.java
jbduncan: _ping_ :)

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
sbrannen: ok. thanks.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/Constants.java
```diff
@@ -187,6 +205,205 @@
 	public static final String PARALLEL_CONFIG_CUSTOM_CLASS_PROPERTY_NAME = PARALLEL_CONFIG_PREFIX
 			+ CONFIG_CUSTOM_CLASS_PROPERTY_NAME;
 
+	/**
+	 * Property name used to set the default timeout for all testable and
```
sbrannen: indentation for method-level Javadoc is all wonky in this class now.

mixture of spaces and tabs? 🤔

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/Constants.java
```diff
@@ -187,6 +205,205 @@
...
+	 * lifecycle methods.
+	 *
+	 * <p>The value of this property will be used unless overridden by a more
+	 * specific property or a {@link org.junit.jupiter.api.Timeout @Timeout}
+	 * annotation present on the method or an enclosing test class (for testable
+	 * methods).
+	 *
+	 * <p>Please refer to the <a href="#supported-values-timeouts">class
```
sbrannen: Does that work for linking to the `<h3>` element?

I thought one had to create an _anchor_ explicitly.

But... if that works like that, then cool.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutDurationParser.java
```diff
@@ -0,0 +1,66 @@
...
+class TimeoutDurationParser {
+
+	private static final Pattern PATTERN = Pattern.compile("([1-9]\\d*) ?((?:[nμm]?s)|m|h|d)?",
+		CASE_INSENSITIVE | UNICODE_CASE);
+	private static final Map<String, TimeUnit> UNITS_BY_ABBREVIATION;
+
+	static {
+		Map<String, TimeUnit> unitsByAbbreviation = new HashMap<>();
+		unitsByAbbreviation.put("ns", NANOSECONDS);
+		unitsByAbbreviation.put("μs", MICROSECONDS);
```
sbrannen: Now that I've thought about it, do we really expect a timeout with the granularity of microseconds or nanoseconds to be meaningful in a test suite, especially if tests are executed in parallel?

I can't really imagine anything more fine-grained than milliseconds being robust in a test suite.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/Constants.java
marcphilipp: I tried it and it worked.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/Constants.java
sbrannen: Or... is that a new bug in the GitHub UI?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutDurationParser.java
marcphilipp: Probably not, I just included it for consistency with the annotation. Or do you think we should define our own enum?

At least `TimeUnit` does not have a `FOREVER` (unlike `ChronoUnit`)... 😏

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/Constants.java
marcphilipp: Doesn't look wonky to me. Could you please post a screenshot?

sbrannen: See above: likely just a bug in the display in the browser due to a change in GitHub (or my browser).

I checked the _source_, and it's fine.

Sorry for the false alarm.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutDurationParser.java
sbrannen: > Probably not, I just included it for consistency with the annotation. 

Understood.

> Or do you think we should define our own enum?

Good question. I'm not sure that "days" makes sense either. "hours" is already pushing it. 

So maybe it would provide a better user experience if users could only select time units that are reasonable for the use case at hand. For example, we could limit it to the following and internally map to `TimeUnit` enum constants for those.

- milliseconds
- seconds
- minutes
- hours

WDYT?

> At least `TimeUnit` does not have a `FOREVER` (unlike `ChronoUnit`)... 😏

Indeed.

sormuras: Let's stick with `TimeUnit` "as-is". It is used with a `long` through-out the Java SE API as well: https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/util/concurrent/class-use/TimeUnit.html

marcphilipp: I agree with @sormuras on this one.

@sbrannen Ok to leave "as-is" or would you want to change the parser?

sbrannen: I think it's OK to stick with `TimeUnit` as the enum.

As for changing the parser, I have mixed feelings. TBH, I don't think anyone will ever have a test timeout in days, but as for whether or not it will ever make sense to support microseconds or nanoseconds... I cannot predict the future.

So maybe it's OK to just support all `TimeUnit` constants as in the status quo.

One option would be to provide a _tip_ in the Javadoc and User Guide pointing out that one should pick reasonable time units, but then again... maybe that's too self explanatory. 😇 

Long story, short: I'll leave it up to you.

Issue #1875, opened 2019-05-11
Remove parentheses when generating display name with ReplaceUnderscores


juliette-derancourt:
I don't know if this has already been discussed, but I noticed something when I discovered the display name generation (which is a cool feature btw!).

When using `ReplaceUnderscores` as a [DisplayNameGenerator](https://github.com/junit-team/junit5/blob/c4ccb87c6afd33e5a09414dd8bb26728d890a684/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java), we can witness the following behavior:

```java
@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
class SomeTest {

    @Test
    void this_is_a_test() {
        // ...
    }

    @ParameterizedTest
    @CsvSource({"a, b"})
    void this_is_a_test_with_arguments(String a, String b) {
        // ...
    }

}
```

IDE output:

<img width="243" alt="tests" src="https://user-images.githubusercontent.com/28359762/57570951-df369480-7407-11e9-9934-0baecd21371d.PNG">

In the first case, I would suggest removing the extra parentheses in the output, as it brings nothing but noise, IMO.

In the second one I'm not sure. Personally I would remove them too, but then we would lose the information about the parameters (I don't know if we really need it though?). Maybe at least put a space between the end of the name and the opening bracket? (Sorry I'm a bit picky ahah)

Tell me what you think! I'd be happy to do it if you agree :)

sbrannen:
Tentatively slated for 5.5 M2 for _team discussion_

sbrannen:
> In the first case, I would suggest removing the extra parentheses in the output, as it brings nothing but noise, IMO.

In general, the idea behind it was consistency and simplicity (in terms of the implementation).

But... I suppose the empty parentheses are a bit superfluous for that use case.

> In the second one I'm not sure. Personally I would remove them too, but then we would lose the information about the parameters (I don't know if we really need it though?). Maybe at least put a space between the end of the name and the opening bracket? (Sorry I'm a bit picky ahah)

Hmmm.... I'd say we should definitely display the parameters if a display name is overloaded.

Unfortunately, a `DisplayNameGenerator` doesn't really have a way to determine if a display name is overloaded (i.e., used multiple times) within a given test class. Note: I am not talking about whether or not a test method is overloaded, because they could have different custom display names.

Thus, I am leaning toward continuing to display the parameter list.

Let's see what the rest of the team thinks.

[...]

PR #1886, opened 2019-05-19
Remove parentheses when generating display name with ReplaceUnderscores

juliette-derancourt:
## Overview

Implements the improvement suggested in #1875.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.5.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.5.0-M2.adoc
@@ -65,6 +65,8 @@ on GitHub.
   details)
 * New Kotlin friendly `assertDoesNotThrow` assertion added as top-level functions in the
   `org.junit.jupiter.api` package
+* Remove extra parentheses when generating a display name with ReplaceUnderscores for a
+  test method without parameters.
 
 
 [[release-notes-5.5.0-M2-junit-vintage]]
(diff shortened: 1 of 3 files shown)


sormuras:
Looks good so far -- only a `gradlew spotlessApply` run is missing to get the CI builds pass.

juliette-derancourt:
> Looks good so far -- only a gradlew spotlessApply run is missing to get the CI builds pass.

Thanks! Should be better now :)

juliette-derancourt:
I noticed something strange while doing this: the test `underscoreGenerator()` in [DisplayNameGenerationTests](https://github.com/junit-team/junit5/blob/b3ee93ff9e4ffe81a84bad04af5d9decb12d8161/junit-jupiter-engine/src/test/java/org/junit/jupiter/api/DisplayNameGenerationTests.java#L55) still passes if I leave the parentheses at the end of a display name.

It seems to be because of the implementation of [assertLinesMatch](https://github.com/junit-team/junit5/blob/b3ee93ff9e4ffe81a84bad04af5d9decb12d8161/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java#L1581), which is used here. This assertion internally relies on the `boolean matches(String regex)` method of `String.java`, which takes a regex and therefore ignore the `()` at the end.

Don't know if this matters or not?

sormuras:
> Should be better now :)

Great. Looks good. Will merge it soon.

> [...] `assertLinesMatch` [...] which takes a regex and therefore ignore the `()` at the end.

Yes, that's the expected behaviour of this "staged-matching assertion utility". It's up to the test author to pass a stricter regex as an expected line to prevent "false-positives".

PR #1910, opened 2019-06-01
Add config param for default execution mode of top-level classes

### LINKED ISSUE #1682: Add parallel execution mode that runs only test classes in parallel
I'm in the process of upgrading our parallel test setup to JUnit5. My JUnit4 surefire configuration looks like this:

```
<configuration>
    <parallel>classes</parallel>
    <threadCount>1</threadCount>
    <perCoreThreadCount>true</perCoreThreadCount>
</configuration>
```

With this configuration, test classes run in parallel but individual test methods within a class run within the same thread (see [Surefire Docs](
https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html)).

I tried to emulate this behavior with `@Execution(ExecutionMode.SAME_THREAD)` on my test classes. But as discussed on Gitter, this actually causes the test and all its method to run in the same thread as the test classes parent node which is the test engine.

@marcphilipp suggested two possible workarounds:
- you could put the annotation on each test method but that's probably overkill
- you could put all your test methods into `@Nested` classes and use `@Execution(ExecutionMode.SAME_THREAD)` on the nested class

Both workarounds are quite invasive and not very intuitive for authors of new tests.

It would be great if we had an additional execution mode that runs only test classes in parallel or if `@Execution(ExecutionMode.SAME_THREAD)` on test classes could be adapted to do that.

marcphilipp:
The new `junit.jupiter.execution.parallel.mode.classes.default`
configuration parameter can be used to run top-level classes in parallel
but their methods sequentially or vice versa.

Resolves #1682.

---

### Definition of Done


documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1715,6 +1715,33 @@ declarative <<writing-tests-parallel-execution-synchronization, synchronization>
...
+.Configuration parameters to execute top-level classes in sequentially but their methods in parallel
+----
+junit.jupiter.execution.parallel.enabled = true
+junit.jupiter.execution.parallel.mode.default = concurrent
+junit.jupiter.execution.parallel.mode.classes.default = same_thread
+----
+
+Not configuring a value for `junit.jupiter.execution.parallel.mode.classes.default` is
+equivalent to using the same value as for `junit.jupiter.execution.parallel.mode.default`.
+
```
sormuras: Add a paragraph showing both _default_ keys set to `concurrent`:

```
----
junit.jupiter.execution.parallel.enabled = true
junit.jupiter.execution.parallel.mode.default = concurrent
junit.jupiter.execution.parallel.mode.classes.default = concurrent
----
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
```diff
@@ -0,0 +1,439 @@
...
+import org.junit.platform.engine.support.hierarchical.ExclusiveResource;
+import org.junit.platform.engine.support.hierarchical.ThrowableCollector;
+
+/**
+ * {@link TestDescriptor} for tests based on Java classes.
+ *
+ * @since 5.5
+ */
+@API(status = INTERNAL, since = "5.5")
+public abstract class ClassBasedTestDescriptor extends JupiterTestDescriptor {
```
sormuras: This class was just renamed from `ClassTestDescriptor` by inserting the `Based` literal, right?

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
marcphilipp: That's the same as the following so I didn't think it made sense to include it. WDYT?

```
----
junit.jupiter.execution.parallel.enabled = true
junit.jupiter.execution.parallel.mode.default = concurrent
----
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
marcphilipp: That and `instantiateTestClass` is now abstract instead of only being overridden in `NestedClassTestDescriptor`.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sormuras: Yeah, might feel like an unnecessary repetition, but it doesn't hurt to be explicit here.

_Not configuring a value for [...] is equivalent to using the same value as [...]._
_It unrolls to this configuration "2x concurrent", which means classes and methods are executed in any order._

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
sormuras: :+1:

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sormuras: Perhaps we also should _paint out_ a potential execution order example like this?

- same, same: `T{a1, a2, a3, b1, b2, b3, c1, c2}`
- same, concurrent: `TA{a1, a3, a2}, TB{b3, b2, b1}, TC{c2, c1}`
- concurrent, same: `TA{a1, a2, a3}, TB{b1, b2, b3}, TC{c1, c2}`
- concurrent, concurrent: `TX{a1, b2, c2, b3}, TY{b1, a3}, TZ{c1, a2}`

marcphilipp: A series of gantt-like charts would be a good illustration. 🤔

sormuras: ### same, same
```
T -> a1 - a2 - b1 - b2
```

### same, concurrent
```
T1 -> a1 - a2
T2 -> b1 - b2
```

### concurrent, same
```
T1 -> a1 - b1
T2 -> a2 - b2
```

### concurrent, concurrent
```
T1 -> a1
T2 -> a2
T3 -> b1
T4 -> b2
```

Correct?

marcphilipp: Depends on what `a1` and so on mean. 😉

marcphilipp: I'll go ahead, merge this PR and will create a follow-up issue to improve the documentation.

PR #1918, opened 2019-06-05
Cleaning

BielinskiLukasz:
## Overview

Hello. I did some small cleaning in your repositories (simple and small changes) in Java code. Maybe you can use one or more. Have a nice day!
List of changes:
- add final keywords
- remove unnecessary keywords and semicolons
- simplify return statements
- create negate methods
- replace Optional = null by Optional.empty()
- simplify:
    - list initialization
    - create String from bytes read from file as Files.readString(xmlFile)
    - replace !Optional.isPresent() by Optional.isEmpty()
    - replace String.valueOf(c) by c in String creation // char c

---

---

### Definition of Done

### DIFF
--- a/documentation/src/test/java/example/testinterface/TestLifecycleLogger.java
+++ b/documentation/src/test/java/example/testinterface/TestLifecycleLogger.java
@@ -25,7 +25,7 @@ import org.junit.jupiter.api.TestInstance.Lifecycle;
 @TestInstance(Lifecycle.PER_CLASS)
 interface TestLifecycleLogger {
 
-	static final Logger logger = Logger.getLogger(TestLifecycleLogger.class.getName());
+	Logger logger = Logger.getLogger(TestLifecycleLogger.class.getName());
 
 	@BeforeAll
 	default void beforeAllTests() {
--- a/documentation/src/test/java/example/testinterface/TestLifecycleLogger.java
+++ b/documentation/src/test/java/example/testinterface/TestLifecycleLogger.java
@@ -25,7 +25,7 @@ import org.junit.jupiter.api.TestInstance.Lifecycle;
 @TestInstance(Lifecycle.PER_CLASS)
 interface TestLifecycleLogger {
 
-	Logger logger = Logger.getLogger(TestLifecycleLogger.class.getName());
+	static final Logger logger = Logger.getLogger(TestLifecycleLogger.class.getName());
 
 	@BeforeAll
 	default void beforeAllTests() {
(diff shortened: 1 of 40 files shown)


documentation/src/test/java/example/testinterface/TestLifecycleLogger.java
```diff
@@ -25,7 +25,7 @@
 @TestInstance(Lifecycle.PER_CLASS)
 interface TestLifecycleLogger {
 
-	static final Logger logger = Logger.getLogger(TestLifecycleLogger.class.getName());
+	Logger logger = Logger.getLogger(TestLifecycleLogger.class.getName());
```
toniedzwiedz: I wouldn't say these keywords were superfluous. There's no reason to make a logger per instance and I don't see why anyone would want to reassign the logger.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TestInfoParameterResolver.java
```diff
@@ -85,7 +85,7 @@ public String toString() {
 		}
 
 		private static Object nullSafeGet(Optional<?> optional) {
-			return optional != null ? optional.orElse(null) : null;
+			return optional.isPresent() ? optional.orElse(null) : null;
```
toniedzwiedz: This change breaks null-safety. `null.isPresent()` will crash and burn.

BielinskiLukasz: Thanks for the review, you are right, I didn't think about that situation

documentation/src/test/java/example/testinterface/TestLifecycleLogger.java
BielinskiLukasz: Of course, you are right :) And also I have. Fields in an interface are by default public, static and final.

BielinskiLukasz:
I made changes also in unit test classes so it caused problems with checks. I will try to clean this tomorrow (revert changes in unit tests).

BielinskiLukasz:
I must revert many of own changes. Maybe I will create a new pull request without changes in unit tests (and NullpointerException found by @toniedzwiedz). If that's better idea please close that request or write a comment. Thanks a lot! :)

sbrannen:
> I must revert many of own changes. Maybe I will create a new pull request without changes in unit tests (and NullpointerException found by @toniedzwiedz). If that's better idea please close that request or write a comment. Thanks a lot! :)

OK. I'll close this PR now.

Please submit new PRs for the individual categories of changes you propose and take note of the tips I provide below.

That will make it easier for the team to digest.

We typically don't care about a field or a local variable in a test class being `final`. So we will likely choose not to accept such proposed changes.

The same holds for stylistic decisions like the presence of a trailing semicolon for enum constants. We've done that consistently and do not plan to change that.

PR #1922, opened 2019-06-07
add ClassFileMethodOrderer

### LINKED ISSUE #1919: Provide a MethodOrderer that uses the order in the source file
The `MethodOrderer` implementation `OrderAnnotation` uses an `@Order` annotation to allow sorting the execution order of tests (which is generally a smell, but sometimes necessary). But that means another annotation for every test method. It would be nicer to simply use the order of the methods in the source. As this information is not provided by the Java reflection API, we'd have to parse the class file. Maybe we don't need a full parser, but a simple heuristic may be enough.

Some years ago I already wrote a simple JUnit 4 runner for this [here](https://github.com/t1/test-tools/blob/master/src/main/java/com/github/t1/testtools/OrderedJUnitRunner.java)

## Deliverables

- [ ] `ClassFileMethodOrderer` implementing `MethodOrderer`

t1:
Resolves #1919.

## Overview

Add a `ClassFileMethodOrderer` that orders test methods according to the order in the class file (which is the same as the order in the source file)

---

---

### Definition of Done

- [x] Change is documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/)
- [ ] Change is documented in the [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)


sbrannen:
FYI: you have some build errors on the CI server.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -242,4 +249,75 @@ public void orderMethods(MethodOrdererContext context) {
...
+		private static final Map<Class<?>, String> CACHE = new HashMap<>();
+
+		@Override
+		public void orderMethods(MethodOrdererContext context) {
+			context.getMethodDescriptors().sort(COMPARATOR);
+		}
+
+		private static final Comparator<MethodDescriptor> COMPARATOR = Comparator //
+				.comparing(ClassFileMethodOrderer::getMethodIndex) //
+				.thenComparing(descriptor -> parameterList(descriptor.getMethod()));
```
marcphilipp: What about inherited methods?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -242,4 +249,75 @@ public void orderMethods(MethodOrdererContext context) {
...
+		}
+
+		private static String toByte(int character) {
+			return Character.toString((char) character);
+		}
+
+		private static byte[] read(Class<?> type) {
+			// the simple name doesn't include any containing classes
+			String typeName = type.getName().substring(type.getPackage().getName().length() + 1);
+			try (InputStream inputStream = type.getResourceAsStream(typeName + ".class")) {
```
marcphilipp: This won't work for nested classes, will it?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -242,4 +249,75 @@ public void orderMethods(MethodOrdererContext context) {
...
+		}
+
+		private static final Comparator<MethodDescriptor> COMPARATOR = Comparator //
+				.comparing(ClassFileMethodOrderer::getMethodIndex) //
+				.thenComparing(descriptor -> parameterList(descriptor.getMethod()));
+
+		private static int getMethodIndex(MethodDescriptor methodDescriptor) {
+			Method method = methodDescriptor.getMethod();
+			String pattern = toStringWithLengthPrefix(method);
+			String classFile = CACHE.computeIfAbsent(method.getDeclaringClass(), cls -> new String(read(cls), UTF_8));
```
forax: keeping the declaring class in a cache means that it will never be unloaded !

sbrannen: Good point!

I guess we'd better store it in the `LruCache` that will be introduced in conjunction with #2012.

forax: I think it's better to use a java.lang.ClassValue which is basically a class cache.

sbrannen: Interesting! I wasn't aware of `ClassValue`, so thanks for pointing that out.

LifeIsStrange:
Any update?

marcphilipp:
Closing with rationale of https://github.com/junit-team/junit5/issues/1919#issuecomment-825563384.

Issue #1900, opened 2019-05-25
New annotation "@Requirement"


Bukama:
_On the last JUG I met @sormuras and he motivated me to participate for the first time by presenting a feature-request. So here it is._

**Motivation:**
When we write tests we do this mostly because of two reasons:

1. Avoid technical errors (like NPE)
2. Ensure our code fullfills the defined requirements / avoid functional errors

But as of today (_as far as I know_) the only way to "link" a test to a requirement is by using it's (displayed) name, for example putting the id of the requirement (_I'll call this "req-id" in the following lines_) in front of the name - like `req123_testSomethingInThisMethod`. I think you all agree that this is not a good way to show that this method is a test to ensure the requirement with the id `req123`.

**Suggestion:**
Therefore I would like to suggest do add a new annotation `@Requirement`.

This annotation should be used to show that a particular test is written to ensure the specific requirement. The annotation takes the req-id as a string parameter, e.g. (`@Requirement("REQ-123")`). In the test result this req-id is then published as an attribut of the test case. This allows tools which parse the result to show if all tests annotated with a given req-id have passed and therefore the requirement is fullfilled.

I don't see the need of making this annotation repeatable on a method. In my opinion each test method should only verify one aspect. I also don't see the need of using this annotation on class level as I in my opinion a test calls - especially for small methods - may contain technical and functional tests.

Maybe this can also be taken into the considerations about a standard test result format (https://github.com/ota4j-team/opentest4j/issues/9)

**Lookout:**

To be honest only printing an additional attribut containing the id of a requirment is only one part of a useful funcitonality to check if all requirements are fullfilled. As without a list of definied requirements the report only containts information about tests which are annotation with a requirment but the report (tool) doesn't know if there were tests to all requirements. While I'm quite sure the suggested annotation should be a feature of JUnit I'm not sure if the "linking functionality" is seen as one by the JUnit-Team. I'll make a short description of what I mean, then maybe it's more clear:

To compare a list of requirements to the list of test results a comparator needs an input of this list. So I see the need of some module which takes this list (e.g. a file, service call, etc.) as an input and compares it with the test results and then creates a report which shows the reader which requirements don't have test at all, or which tests of a given requirement have passed/failed.

But maybe this is just a second step and - as mentioned - maybe not even in focus of JUnit.

marcphilipp:
FWIW Spock has an `@Issue` annotation.

@Bukama Have you considered using `@Tag()` instead?

Bukama:
[...]

PR #1940, opened 2019-06-27
Introduce @Requirement annotation

Bukama:
## Overview

I'm opening this pull request for issue #1900 to ask for some feedback about my attempt to implement the new annotation.

I orient myself at the "@Displayname" annotation and did the following steps

1. Defined the new annotation
2. Extended the extensionContext, testDescriptor and tesInfo interfaces for it
3. Added the new attribute to the AbstractTestDescriptor and modifided extending classes
4. Extended the XSD and the XMLWriter to write the new attribute

On the one hand side I think that should do it, on the other side I think something is missing. What's definitly missing is an updated documentation module. Havn't done that yet as I want to ask for feedback while updating the documentation in parallel.

_Nevertheless thanks to @sormuras helping me fixing IntelliJ and giving a short introduction into gradle and the junit modules_

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Requirement.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Requirement.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2015-2019 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import java.lang.annotation.*;
+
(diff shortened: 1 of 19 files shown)


marcphilipp:
@Bukama Wow, that looks like it was a lot of work! We generally recommend to check with the team before starting coding.

I don't think we should dive into the details of the changes. A few high-level remarks from my side: I think if we introduce a new high-level concept for `TestIdentifier` it should be more generic so it can be used for different purposes but not as generic that it becomes useless. A list of custom key-value pairs comes to mind. For Jupiter we'd then also need an extensible mechanism to attach those to a `TestDescriptor`. Alternatively, we could attach all readable annotations. For reporting these, including "requirement" we're going to need a new reporting format (as mentioned in #1900). I'm afraid we can't just add elements to the XSD because it's not controlled by us and would probably break downstream reporting tools.

marcphilipp:
**Team Decision:** Since we don't want to implement this feature in this way, let's continue the discussion in #1900.

Bukama:
@marcphilipp Thanks for your response - then I did missunderstood your comment in #1900 But it's alright :)

Issue #1958, opened 2019-07-15
Add support for String delimiters in @CsvSource and @CsvFileSource


nrundle:
Configuring the delimiter in `@CsvSource` when using Groovy is extremely limited due to the lack of support of character literals.

The only way to configure the delimiter in the annotation that I could find is to cast it to char which requires the class the use of `@CompileStatic`.

The [univocity parser](https://github.com/uniVocity/univocity-parsers/blob/v2.8.2/src/main/java/com/univocity/parsers/csv/CsvFormat.java) that is used can take either a `char` or `String` delimiter.

In order to offer better support for writing tests in Groovy, please update this annotation so that a string delimiter may be used.

## Deliverables

- [ ] Add support for String delimiters in `@CsvSource` and `@CsvFileSource`

marcphilipp:
You mean like this?

```groovy
@CompileStatic
class GroovyParamsTests {

    private static final char COMMA = (char) ','

    @CsvSource(value = ["1, 2", "3, 4"], delimiter = COMMA)
    @ParameterizedTest
    void test(int i, int j) {

    }
}
```

marcphilipp:
Changing it from `char` to `String` would be a breaking change. On the other hand,  `@CsvSource` is still experimental although widely used.

Tentatively slated for 5.6 M1 for _team discussion_.

nrundle:
Yes, I experimented with using `@CompileStatic`, but unfortunately then I lose access to other Groovy features that I was using.

So far, I like being able to use the @CsvSource and see the inputs and expected outputs right above the test.

sormuras:
Perhaps something Groovy 3.x will support w/o resorting to `String`?

Because only to support `","` -- a single character sequence -- I'd hesitate to change the API of `@CsvSource`.

nrundle:
Perhaps rather than changing the API, a new one could be created.  Since the univocity parser allows string delimiters with more than one character, I thought it might be a nice enhancement which not only supports groovy's lack of a character literal, but also allows for a wider set of delimiters to be used.

sbrannen:
> Perhaps rather than changing the API, a new one could be created. Since the univocity parser allows string delimiters with more than one character, I thought it might be a nice enhancement which not only supports groovy's lack of a character literal, but also allows for a wider set of delimiters to be used.

I suppose we could introduce something like `delimiterString`. It's not the most elegant and would require a more complicated lookup algorithm (that rejects configuration of both attributes), but it's possible.

[...]

PR #1972, opened 2019-08-03
Add support for String delimiters in @CsvSource and @CsvFileSource

juliette-derancourt:
## Overview

Solves #1958 by adding a `delimiterString` attribute to the `@CsvSource` annotation.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
@@ -48,7 +48,7 @@ on GitHub.
 ==== New Features and Improvements
 
 * Documented the support of commented lines in CSV file with `@CsvFileSource`.
-* Add support for String delimiters in `@CsvSource`.
+* Add support for String delimiters in `@CsvSource` and `@CsvFileSource`.
 
 [[release-notes-5.6.0-M1-junit-vintage]]
 === JUnit Vintage
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
@@ -48,7 +48,7 @@ on GitHub.
 ==== New Features and Improvements
 
 * Documented the support of commented lines in CSV file with `@CsvFileSource`.
-
+* Add support for String delimiters in `@CsvSource`.
 
 [[release-notes-5.6.0-M1-junit-vintage]]
 === JUnit Vintage
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1160,6 +1160,10 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=CsvFileSource_example
 include::{testResourcesDir}/two-column.csv[]
 ----
 
+As with `@CsvSource`, the default delimiter is `,` but you can use another character by
+setting the `delimiter` attribute. Alternatively, the `delimiterString` attribute allows
+to use a `String` delimiter. However, both attributes cannot be set simultaneously.
+
 NOTE: In contrast to the syntax used in `@CsvSource`, `@CsvFileSource` uses a double
 quote `"` as the quote character. See the `"United States of America"` value in the
 example above. An empty, quoted value `""` results in an empty `String` unless the
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1122,6 +1122,10 @@ include::{testDir}/example/ExternalMethodSourceDemo.java[tags=external_MethodSou
 include::{testDir}/example/ParameterizedTestDemo.java[tags=CsvSource_example]
 ----
 
+The default delimiter is `,` but you can use another character by setting the `delimiter`
+attribute. The `delimiterString` attributes also allows you to use a String as a delimiter.
+However, both attributes cannot be set simultaneously.
+
 `@CsvSource` uses a single quote `'` as its quote character. See the `'lemon, lime'` value
 in the example above and in the table below. An empty, quoted value `''` results in an
 empty `String` unless the `emptyValue` attribute is set; whereas, an entirely _empty_
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1123,7 +1123,7 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=CsvSource_example]
 ----
 
 The default delimiter is `,` but you can use another character by setting the `delimiter`
-attribute. The `delimiterString` attributes also allows you to use a String as a delimiter.
+attribute. Alternatively, the `delimiterString` attribute allows to use a `String` delimiter.
 However, both attributes cannot be set simultaneously.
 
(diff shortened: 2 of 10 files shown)


juliette-derancourt:
I'm not quite satisfied by the Javadoc, but I don't really know how to put it better...
Also, I don't know if there's a way to enforce the fact that you can't set both attributes at the same time directly on the annotation?

sormuras:
> I'm not quite satisfied by the Javadoc, but I don't really know how to put it better...

That's Sam's realm. 

> Also, I don't know if there's a way to enforce the fact that you can't set both attributes at the same time directly on the annotation?

There isn't. Only option is to fail at runtime.

juliette-derancourt:
> Didn't look into the changes, but the CI complains about a missing 'spotlessApply' run.

My bad, fixed it!

> There isn't. Only option is to fail at runtime.

Too bad...
Overall, I feel that the issue could be handled more elegantly, but I guess it's the best that can be done without changing the existing API

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvSource.java
```diff
@@ -50,10 +50,21 @@
...
 	 */
-	char delimiter() default ',';
+	char delimiter() default '\0';
+
+	/**
+	 * The column delimiter to use when reading the {@linkplain #value lines}.
+	 * This can't be set simultaneously with {@link #delimiter delimiter}.
+	 * If neither attributes are set, the delimiter will be ','.
+	 *
+	 * <p>Defaults to {@code ""}.
```
marcphilipp: I think we should remove the defaults from the Javadoc because they might confuse users.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1122,6 +1122,10 @@ include::{testDir}/example/ExternalMethodSourceDemo.java[tags=external_MethodSou
 include::{testDir}/example/ParameterizedTestDemo.java[tags=CsvSource_example]
 ----
 
+The default delimiter is `,` but you can use another character by setting the `delimiter`
+attribute. The `delimiterString` attributes also allows you to use a String as a delimiter.
```
marcphilipp:
```suggestion
attribute. Alternatively, the `delimiterString` attribute allows to use a `String` delimiter.
```

documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
```diff
@@ -48,7 +48,7 @@ on GitHub.
 ==== New Features and Improvements
 
 * Documented the support of commented lines in CSV file with `@CsvFileSource`.
-
+* Add support for String delimiters in `@CsvSource`.
```
marcphilipp: I'm wondering whether we should also add `delimiterString` to `@CsvFileSource`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvSource.java
juliette-derancourt: Indeed, I was thinking the same thing

documentation/src/docs/asciidoc/release-notes/release-notes-5.6.0-M1.adoc
juliette-derancourt: I was wondering about that too, I don't know if that would be useful... maybe @nrundle have some insight on this?

juliette-derancourt: It could also be done just for the sake of consistency

nrundle: I’d do it in order to be consistent.

Personally, I like being able to see the test values right by the test so I haven’t used file source.

marcphilipp: @junit-team/junit-lambda Any objections?

sormuras: No objections.

marcphilipp: @juliette-derancourt I'd say let's add it to `@CsvFileSource` for the sake of consistency then.

sbrannen: +1

junit-jupiter-params/src/test/java/org/junit/jupiter/params/provider/CsvFileArgumentsProviderTests.java
```diff
@@ -164,52 +230,33 @@ void throwsExceptionForInvalidCharset() {
...
 
-	private Stream<Object[]> provideArguments(InputStream inputStream, String lineSeparator, char delimiter,
-			String emptyValue) {
-		String expectedResource = "foo/bar";
-		CsvFileSource annotation = annotation("ISO-8859-1", lineSeparator, delimiter, emptyValue, expectedResource);
+	private Stream<Object[]> provideArguments(InputStream inputStream, CsvFileSource annotation) {
+		String expectedResource = annotation.resources()[0];
 
 		CsvFileArgumentsProvider provider = new CsvFileArgumentsProvider((testClass, resource) -> {
 			assertThat(resource).isEqualTo(expectedResource);
```
juliette-derancourt: @marcphilipp I don't really understand the purpose of this assertion?

juliette-derancourt:
I added the same attribute on `@CsvFileSource`.
I took the liberty to refactor the tests a bit, hope you don't mind ^^
Tell me if you agree with the introduction of [CsvFileSourceMock](https://github.com/junit-team/junit5/blob/14c2b2f65d5da20fe88cd6ca6d86618d0faeeb16/junit-jupiter-params/src/test/java/org/junit/jupiter/params/provider/CsvFileSourceMock.java) (could be named CsvFileSourceBuilder?), as I felt that building the annotations via polymorphic methods started to be a bit noisy.

I'd also like to remove the duplication between CsvArgumentsProvider and CsvFileArgumentsProvider, but I don't know which approach to choose:
- Regroup the delimiter choice logic in a something like a `Delimiter` class with a static method
- Make CsvArgumentsProvider and CsvFileArgumentsProvider extend the same abstract class, something like CsvAnnotationConsumer<?> (would like this better, but `@CsvSource` and `@CsvFileSource` would also have to have a common parent)

What do you think?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsParser.java
```diff
@@ -0,0 +1,84 @@
...
+package org.junit.jupiter.params.provider;
+
+import java.lang.annotation.Annotation;
+
+import com.univocity.parsers.csv.CsvParser;
+import com.univocity.parsers.csv.CsvParserSettings;
+
+import org.junit.platform.commons.PreconditionViolationException;
+
+public class CsvArgumentsParser {
```
marcphilipp:
```suggestion
class CsvArgumentsParser {
```

sbrannen:
I've revised this a bit in dff526f3f0dcebb101f40446f5d5df207160f7a4.

I really like the `CsvFileSourceMock`. It's much better with that in place.

Thanks again for the PR!

sbrannen:
@junit-team/junit-lambda Let's try to remember to check the `@since` and `@API` declarations. They were missing in the merged PR, even though the "Public API has Javadoc and @API annotations" TODO was checked off.

marcphilipp:
Point taken. Mea culpa.

Issue #1963, opened 2019-07-24
Make org.junit.platform.console.tasks.*Listener classes public


kriegfrj:
I have a project where it would be nice to be able to plug in arbitrary test listeners to get the output that you want.

It would be nice to be able to use the listeners in the console package to optionally produce nicely-formatted test output. Unfortunately, I can't do this easily because the classes are package private, so I have to use reflection if I want to use them.

kriegfrj:
Having played with it a little bit, I think it would also be nice if all the listener-related code were put in a different package to the executor code. I'm looking at an OSGi-type usage of it, and having this separation of packages would make it easier to manage the package dependencies.

sbrannen:
Tentatively slated for 5.6 M1 for _team discussion_

sormuras:
Iirc, we introduced the "reporting" module to contain such public listeners.

marcphilipp:
**Team Decision:** Let's polish the listeners to keep the public API small and move them to the reporting module.

marcphilipp:
@kriegfrj Would you like to submit a PR?

kriegfrj:
I'll see if I can have a look next week some time. Can't promise anything. I have another couple of suggestions from my experience in trying to use the LegacyXmlGeneratingListener too which might tie in (eg, perhaps `org.unit.platform.launcher.listeners.LegacyReportingUtils` also belongs in the reporting module and in a different package.

kriegfrj:
In terms of clarifying Hebrew scope odd this feature, @sormuras suggested not including the color stuff in the reporting api. My vote (for what little it may be worth), as someone who might want to use this functionality, would be to include it. Anyone who wants to use these listeners may also want the output with color codes. I got one would like to. It might be useful to make it customizable somehow, but in wouldn't like the color functionality to remain part of the private api as that would mean having to reinvent it in my own code.

sormuras:
My suggestion is to keep all console-only related code, like the "color stuff", reside in the console module. We should only move (read: re-implement) the target-agnostic data structures and their related listeners to the reporting module.

marcphilipp:
@kriegfrj We're having second thoughts about making the listeners with all their configuration options public. Would a class that contains a set of factory methods, e.g. `createTreePrintingListener(boolean enableAnsiColors): TestExecutionListener`, work in your use case?

kriegfrj:
[...]

PR #1977, opened 2019-08-10
Make console printing listeners public (and other reporting enhancements)

kriegfrj:
## Overview

This is an implementation of #1963, with a couple of other changes thrown in.

I created a new package in junit-platform-reporting called `org.junit.platform.reporting.console` and moved all the three listener class implementations into that package. I also had to remove a couple of the ancillary classes along with them that were used for configuring the listeners (`Theme`, `Color`), or for implementing some of their core functionality (`TreePrinter`). This part was fairly straightforward, and hopefully uncontroversial. The listener classes and those used for configuration were made public; the implementation classes were left package private.

In order to make the classes more user friendly, I added some overloaded constructors with default options and added some documentation. The documentation is based on my reading of the code - hopefully I didn't get it too wrong (in particular, if someone could check my comments about the relative merits & functionality of the `TreePrintingListener` vs `VerboseTreePrintingListener`, that would be appreciated).

I marked all of the newly publicized classes with the `@API` annotation, marking them as experimental as of 1.6.

Internally, the console listeners all had a flag `disableAnsiColors`. I have taken the opportunity to flip the polarity on this and called it `useAnsiColors`, as this is easy to do while the API is not public. The console executor still processes the command line the same way and takes care of the conversion.

Another changes was to move `LegacyReportingUtils` from `junit-platform-launcher` to `junit-platform-reporting`. Reasoning:

* It seems to belong more properly in that module rather than in the launcher module - it is a reporting tool, not a launching tool.
* The only place in the JUnit 5 source tree where it was actually used is in the reporting module (`LegacyXmlReportGeneratingListener`).
* The fact that this class was split and in the wrong package/module caused me a few headaches when I was trying to use it in an OSGi context.

Finally, I added a feature to `LegacyXmlReportGeneratingListener` so that you can override the behaviour as to when an xml file is generated. The original implementation would generate one file per root in the test hierarchy. The new feature allows you to override this so that you can (eg) produce one xml file per test class, or per other container level. This is a feature that I would have found useful for the Bnd JUnit Platform tester implementation. This feature is isolated to its own commit and if necessary could be hived off to its own PR.

Thoughts/comments/suggestions welcome!

### DIFF
--- a/junit-platform-console/src/main/java/org/junit/platform/console/options/AvailableOptions.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/options/AvailableOptions.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.junit.platform.engine.discovery.ClassNameFilter;
+import org.junit.platform.reporting.console.Theme;
 
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
--- a/junit-platform-console/src/main/java/org/junit/platform/console/options/CommandLineOptions.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/options/CommandLineOptions.java
@@ -27,6 +27,7 @@ import java.util.Optional;
 import java.util.stream.Stream;
 
 import org.apiguardian.api.API;
+import org.junit.platform.reporting.console.Theme;
 
 /**
  * @since 1.0
--- a/junit-platform-console/src/main/java/org/junit/platform/console/tasks/ConsoleTestExecutor.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/tasks/ConsoleTestExecutor.java
@@ -26,13 +26,16 @@ import org.junit.platform.commons.JUnitException;
 import org.junit.platform.commons.util.ClassLoaderUtils;
 import org.junit.platform.console.options.CommandLineOptions;
 import org.junit.platform.console.options.Details;
-import org.junit.platform.console.options.Theme;
 import org.junit.platform.launcher.Launcher;
 import org.junit.platform.launcher.LauncherDiscoveryRequest;
 import org.junit.platform.launcher.TestExecutionListener;
 import org.junit.platform.launcher.core.LauncherFactory;
 import org.junit.platform.launcher.listeners.SummaryGeneratingListener;
 import org.junit.platform.launcher.listeners.TestExecutionSummary;
+import org.junit.platform.reporting.console.FlatPrintingListener;
+import org.junit.platform.reporting.console.Theme;
+import org.junit.platform.reporting.console.TreePrintingListener;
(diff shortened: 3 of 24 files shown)


kriegfrj:
Looks like the Azure tests have failed. Failures seem to be related to the fact that I have moved those listeners into a different module. I'm not quite sure how to go about debugging them though because it doesn't fail when I run the Gradle tests on my machine.

Also forgot to add this:
---

sormuras:
@kriegfrj - first thanks for the PR.

A general note: please keep the focus on a single task/issue within a single PR. This PR is hard to review as it touches at least three different areas/goals.

Next, I'd like to move (read polish and re-implement) only the data structures and their listeners to the reporting module. Printing and applying console colors should stay in the console module, imho. As I wasn't able to attend the team meeting last Friday, I'll take this opinion to the next meeting.

sormuras:
> Another changes was to move `LegacyReportingUtils` from `junit-platform-launcher` to `junit-platform-reporting`. Reasoning:
>
>It seems to belong more properly in that module rather than in the launcher module - it is a reporting tool, not a launching tool.
> The only place in the JUnit 5 source tree where it was actually used is in the reporting module (`LegacyXmlReportGeneratingListener`).
> The fact that this class was split and in the wrong package/module caused me a few headaches when I was trying to use it in an OSGi context.

A PR with only this change - after we and the CI instances reviewed it - would be a no-brainer to merge.

sormuras:
> Looks like the Azure tests have failed. Failures seem to be related to the fact that I have moved those listeners into a different module. I'm not quite sure how to go about debugging them though because it doesn't fail when I run the Gradle tests on my machine.

Set system property `platform.tooling.support.tests.enabled` to `true` when running a local "full" build using Gradle. This will enable all long-running integration tests that are executed on the CI instances.

sormuras:
Regarding `LegacyReportingUtils` from `junit-platform-launcher` to `junit-platform-reporting` ... it's not that easy. Due to:

https://github.com/junit-team/junit5/blob/5e41ebe612fd67e905f96f1dd3184a071b65be17/junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/LegacyReportingUtils.java#L27

...we have declare it `@Deprecated` first and remove it "later".

kriegfrj:
> A PR with only this change [*Move `LegacyReportingUtils`*] - after we and the CI instances reviewed it - would be a no-brainer to merge.

Split this one into #1979.

kriegfrj:
> Finally, I added a feature to `LegacyXmlReportGeneratingListener` so that you can override the behaviour as to when an xml file is generated. The original implementation would generate one file per root in the test hierarchy. The new feature allows you to override this so that you can (eg) produce one xml file per test class, or per other container level. This is a feature that I would have found useful for the Bnd JUnit Platform tester implementation. This feature is isolated to its own commit and if necessary could be hived off to its own PR.

This has been hived off into #1980 for discussion.

PR #1979, opened 2019-08-12
Move LegacyReportingUtils to junit-platform-reporting

### LINKED ISSUE #1978: Move LegacyReportingUtils to junit-platform-reporting
For historical reasons, `org.junit.platform.launcher.listeners.LegacyReportingUtils` is really in the wrong package and in the wrong module. Its purpose is reporting related, not launching related.

Proposal:

1. Copy it to `org.junit.platform.reporting.legacy` package in `junit-platform-reporting` module.
2. Deprecate the version in `junit-platform-launcher` with appropriate commentary, for future removal.

kriegfrj:
Moved LegacyReportingUtils from junit-platform-launcher to junit-platform-reporting as being more appropriate. Old implementation in the launcher package left in place for compatibility with 3rd party code, but deprecated for future removal. Resolves #1978.

Left the old unit test in-place to guard against regressions. Future dev should be done in the new copy.

---

### DIFF
--- a/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/LegacyReportingUtils.java
+++ b/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/LegacyReportingUtils.java
@@ -8,12 +8,11 @@
  * https://www.eclipse.org/legal/epl-v20.html
  */
 
-package org.junit.platform.launcher.listeners;
+package org.junit.platform.reporting.legacy;
 
-import static org.apiguardian.api.API.Status.DEPRECATED;
+import static org.apiguardian.api.API.Status.MAINTAINED;
 
 import org.apiguardian.api.API;
-import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.engine.support.descriptor.ClassSource;
 import org.junit.platform.launcher.TestIdentifier;
 import org.junit.platform.launcher.TestPlan;
@@ -22,12 +21,12 @@ import org.junit.platform.launcher.TestPlan;
  * Utility methods for dealing with legacy reporting infrastructure, such as
  * reporting systems built on the Ant-based XML reporting format for JUnit 4.
  *
+ * This class was formerly from {@code junit-platform-launcher}
+ * in {@link org.junit.platform.launcher.listeners} package.
+ *
  * @since 1.0.3
- * @deprecated Use {@link org.junit.platform.reporting.legacy.LegacyReportingUtils}
- * instead.
  */
-@Deprecated
-@API(status = DEPRECATED, since = "1.6")
+@API(status = MAINTAINED, since = "1.6")
 public class LegacyReportingUtils {
 
 	private LegacyReportingUtils() {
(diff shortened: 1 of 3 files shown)


junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/LegacyReportingUtils.java
```diff
@@ -0,0 +1,85 @@
...
+ * Utility methods for dealing with legacy reporting infrastructure, such as
+ * reporting systems built on the Ant-based XML reporting format for JUnit 4.
+ *
+ * This class was formerly from {@code junit-platform-launcher}
+ * in {@link org.junit.platform.launcher.listeners} package.
+ *
+ * @since 1.0.3
+ */
+@API(status = MAINTAINED, since = "1.6")
+public class LegacyReportingUtils {
```
ricardomedeirosdacostajunior:
``` suggestion
public final class LegacyReportingUtils {
```

kriegfrj:
> Please remove the tests for the old implementation and delegate to the old implementation from the new location.

Done!

kriegfrj:
> Unless I'm missing sth. the class is still duplicated and does not delegate.

Sorry, I misread your previous comment - I thought you meant to delegate the test to the old implementation from the new (test) location (which is what I did).

Before I go ahead and implement this please let me explain, as my original decision to duplicate rather than delegate was a calculated one. My reasons were as follows:

My original intent behind this issue was to decouple the launching functionality and the launcher jar from the reporting functionality/jar. Delegating from the deprecated implementation to the new location will re-introduce a dependency for the launcher jar on the reporting jar where none currently exists. This is even more apparent with the advent of #2095 - delegating would introduce an `Import-Package` clause in the launcher jar for the `org.junit.platform.reporting.legacy` package and a "uses" clause for `org.junit.platform.launcher.listeners` package on the same reporting package. This means you wouldn't be able to use the launcher jar in OSGi unless the reporting jar is also present. We could make this import optional but that's extra work and maintenance.

The main argument for delegating is to make future maintenance easier while keeping the two implementations in sync. However, the version in the launcher jar is deprecated, there is an easy replacement which is maintained, and this is all well-documented. So I don't think there is a great need to worry too much about future maintenance of the implementation in the launcher jar, nor even care if the two implementations diverge. This is also why I originally had two sets of tests (so that the tests for the implementation in new location could evolve with changes in the implementation if necessary, while the tests in the old location could guard against regressions but not necessarily need to be updated for more functionality).

When I weighed the pros and cons I felt that in the circumstances the decoupling was more important than the maintainability issue, hence I went for duplication rather than delegation.

If after having processed the above you still think it preferable to delegate rather than duplicate, I'll make the change.

marcphilipp:
> Delegating from the deprecated implementation to the new location will re-introduce a dependency for the launcher jar on the reporting jar where none currently exists.

Which is why I proposed to "delegate to the old implementation from the new location".

From experience things that are deprecated have the nasty habit of sticking around longer than one wishes. Thus, I still think we should delegate rather than duplicate.

kriegfrj:
> Which is why I proposed to "delegate to the old implementation from the new location".
>
Wow, I've really scored poorly on my reading comprehension in this PR. :smile: Forgive me, I read it back-to-front.

That makes a bit more sense.
 
> From experience things that are deprecated have the nasty habit of sticking around longer than one wishes. Thus, I still think we should delegate rather than duplicate.

Ok, I have resubmitted the PR with this change.

I totally concur about deprecations hanging around longer than they should. However, my concern would be that delegating rather than duplicating is likely to exacerbate that issue - if anyone makes changes, they are likely to make them to the deprecated code base which will keep it in sync with the preferred code base. If they are allowed to diverge, then it is more likely that people will be forced to migrate. Anyway, that's your call.

Issue #1980, opened 2019-08-12
LegacyXmlReportGeneratingListener to have configurable roots


kriegfrj:
`LegacyXmlReportGeneratingListener` generates one xml report file per root `TestIdentifier` instance in the in the `TestPlan` being executed. The file name is of the form `Test-nnnn.xml` where `nnn` is the name of the root node.

I recently encountered a situation where I would have liked to be able to change both the number of files generated, and the algorithm used to derive their filenames. Unfortunately, `LegacyXmlReportGeneratingListener` has been written in such a way to make this impossible without extensive use of reflection, as the necessary methods that need to be overridden are all private:

https://github.com/junit-team/junit5/blob/7dc2df5cede39eafe17f0cc026f7dfdcc96215ee/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/LegacyXmlReportGeneratingListener.java#L102-L127

`executionFinished()` (and `executionSkipped()`) both call `writeXmlReportInCaseOfRoot()` to do the (optional) writing of an xml file. It would be ideal to simply tap into it here somewhere, but being private makes it impossible without reflection.

## Proposal

1. Combine the functionality of `isRoot()` and the code that extracts the root name in `writeXmlReportInCaseOfRoot()`. This could be done by having a single method which returns an `Optional<String>` - empty indicates a non-root (no file to be written), a string indicates the name to use in the test file. Say we call it `getRootFileName(TestIdentifier)`.
2. Make it possible to override `getRootFileName(TestIdentifier)`, either by:
   1. making it `protected` method and thereby allowing a subclass to override it, or else
   2. by making it a field of type `Function<Optional<String>, TestIdentifier>` and initializing it to the current implementation in the constructor by default.

adityas71094:
@kriegfrj, Even I have a use case where I need to generate a test report for every test class. I later combine all these files into one report file. But right now I am not able to combine as all files have the same name.


It will be great if we are able to call a setFileName(String fileName) or a setFile(File reportFile) on LegacyXmlReportGeneratingListener.

mseele:
@adityas71094 please see https://github.com/junit-team/junit5/issues/1989

kriegfrj:
@adityas71094 , thanks for your comment.

Please see PR #1981. You should be able to use it to accomplish what you want by passing in an implementation for the `rootFileName` parameter that returns the name of the file that you want for each class. Eg, something like this:

```java
[...]

PR #1981, opened 2019-08-12
Make LegacyXmlReportGeneratingListener configurable

kriegfrj:
Added ability to change the behavior of when XML files are written. Implements #1980.

This is a sample implementation (as mentioned in #1980) using a `Function<>` member rather than a protected method.

Happy to change this to use a protected method if this is considered a superior solution.


sormuras:
The source issue #1980 was already closed.

PR #1993, opened 2019-08-28
Auto-detect enum type from method signature

### LINKED ISSUE #1973: Smarter @EnumSource implementation
I've just started using `@EnumSource` in a parameterized test. Very handy.

However, after having used it with a few methods, it occurred to me that it could be a little handier and a little DRYer.

Consider:

```java
@ParameterizedTest
@EnumSource(MyEnum.class)
public void myTestMethod(MyEnum param) {
}
```
We have to specify the enum type twice. This seems to be needless repetition and a possible source of errors.

My proposal is in two parts:

1. The "value" parameter of the `@EnumSource` annotation should be inferred by the engine based on the type of the parameter.
2. As a further enhancement, if there is no other configuration needed for the `@EnumSource` annotation, then it would be nice if it were possible to omit it altogether.

### Proposal part 1

Ability to write tests as follows:

```java
@ParameterizedTest
// Equivalent to @EnumSource(MyEnum.class)
@EnumSource 
public void myTestMethod(MyEnum param) {
}
```

```java
@ParameterizedTest
// Equivalent to @EnumSource(value = MyEnum.class, names = {"VAL1", "VAL2"}))
@EnumSource(names = { "VAL1", "VAL2" }) 
public void myTestMethod(MyEnum param) {
}
```

### Propsal part 2

Ability to write tests as follows:

```java
@ParameterizedTest
// Equivalent to @EnumSource(MyEnum.class)
public void myTestMethod(MyEnum param) {
}
```

marcphilipp:
The `value` parameter of the `EnumSource` annotation is now optional.
When omitted, the declared type of the first method parameter is used.
The test will fail if it does not reference an enum type.

Resolves #1973.

---

### Definition of Done

### DIFF
--- a/documentation/src/test/java/example/ParameterizedTestDemo.java
+++ b/documentation/src/test/java/example/ParameterizedTestDemo.java
@@ -24,10 +24,11 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.time.LocalDate;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
 import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.List;
-import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
@@ -109,36 +110,41 @@ class ParameterizedTestDemo {
 
 	// tag::EnumSource_example[]
 	@ParameterizedTest
-	@EnumSource(TimeUnit.class)
-	void testWithEnumSource(TimeUnit timeUnit) {
-		assertNotNull(timeUnit);
+	@EnumSource(ChronoUnit.class)
+	void testWithEnumSource(TemporalUnit unit) {
+		assertNotNull(unit);
 	}
 	// end::EnumSource_example[]
 
+	// tag::EnumSource_example_autodetection[]
+	@ParameterizedTest
+	@EnumSource
+	void testWithEnumSourceWithAutoDetection(ChronoUnit unit) {
+		assertNotNull(unit);
+	}
+	// end::EnumSource_example_autodetection[]
+
 	// tag::EnumSource_include_example[]
 	@ParameterizedTest
-	@EnumSource(value = TimeUnit.class, names = { "DAYS", "HOURS" })
-	void testWithEnumSourceInclude(TimeUnit timeUnit) {
-		assertTrue(EnumSet.of(TimeUnit.DAYS, TimeUnit.HOURS).contains(timeUnit));
+	@EnumSource(names = { "DAYS", "HOURS" })
+	void testWithEnumSourceInclude(ChronoUnit unit) {
+		assertTrue(EnumSet.of(ChronoUnit.DAYS, ChronoUnit.HOURS).contains(unit));
 	}
 	// end::EnumSource_include_example[]
 
 	// tag::EnumSource_exclude_example[]
 	@ParameterizedTest
-	@EnumSource(value = TimeUnit.class, mode = EXCLUDE, names = { "DAYS", "HOURS" })
-	void testWithEnumSourceExclude(TimeUnit timeUnit) {
-		assertFalse(EnumSet.of(TimeUnit.DAYS, TimeUnit.HOURS).contains(timeUnit));
-		assertTrue(timeUnit.name().length() > 5);
+	@EnumSource(mode = EXCLUDE, names = { "ERAS", "FOREVER" })
+	void testWithEnumSourceExclude(ChronoUnit unit) {
+		assertFalse(EnumSet.of(ChronoUnit.ERAS, ChronoUnit.FOREVER).contains(unit));
 	}
 	// end::EnumSource_exclude_example[]
 
 	// tag::EnumSource_regex_example[]
 	@ParameterizedTest
-	@EnumSource(value = TimeUnit.class, mode = MATCH_ALL, names = "^(M|N).+SECONDS$")
-	void testWithEnumSourceRegex(TimeUnit timeUnit) {
-		String name = timeUnit.name();
-		assertTrue(name.startsWith("M") || name.startsWith("N"));
-		assertTrue(name.endsWith("SECONDS"));
+	@EnumSource(mode = MATCH_ALL, names = "^.*DAYS$")
+	void testWithEnumSourceRegex(ChronoUnit unit) {
+		assertTrue(unit.name().endsWith("DAYS"));
 	}
 	// end::EnumSource_regex_example[]
 
@@ -261,7 +267,7 @@ class ParameterizedTestDemo {
 	// tag::implicit_conversion_example[]
 	@ParameterizedTest
 	@ValueSource(strings = "SECONDS")
-	void testWithImplicitArgumentConversion(TimeUnit argument) {
+	void testWithImplicitArgumentConversion(ChronoUnit argument) {
 		assertNotNull(argument.name());
 	}
 	// end::implicit_conversion_example[]
@@ -297,11 +303,11 @@ class ParameterizedTestDemo {
 	// @formatter:off
 	// tag::explicit_conversion_example[]
 	@ParameterizedTest
-	@EnumSource(TimeUnit.class)
+	@EnumSource(ChronoUnit.class)
 	void testWithExplicitArgumentConversion(
 			@ConvertWith(ToStringArgumentConverter.class) String argument) {
 
-		assertNotNull(TimeUnit.valueOf(argument));
+		assertNotNull(ChronoUnit.valueOf(argument));
 	}
 
 	// end::explicit_conversion_example[]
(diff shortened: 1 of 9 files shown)


documentation/src/test/java/example/ParameterizedTestDemo.java
```diff
@@ -109,36 +110,41 @@ void nullEmptyAndBlankStrings(String text) {
 
 	// tag::EnumSource_example[]
 	@ParameterizedTest
-	@EnumSource(TimeUnit.class)
-	void testWithEnumSource(TimeUnit timeUnit) {
-		assertNotNull(timeUnit);
+	@EnumSource(ChronoUnit.class)
```
sbrannen: Although `ChronoUnit implements TemporalUnit`, that's a rather special case, and I'm not sure it benefits the average user to demonstrate that.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/NullEnum.java
```diff
@@ -0,0 +1,20 @@
...
+
+package org.junit.jupiter.params.provider;
+
+/**
+ * Dummy enum class to be used as default value of annotation attributes.
+ *
+ * @since 5.6
+ * @see EnumSource#value()
+ */
+enum NullEnum {
```
sbrannen: I see you decided to make this package private.

Hope that doesn't cause any issues.

In the very least, it would make the Javadoc for `NullEnum` unavailable if package-private types are not included in the generated Javadoc.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumSource.java
```diff
@@ -56,7 +58,7 @@
 	 * @see #names
```
sbrannen: Probably good to document here what the semantics of the `default` value are.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumSource.java
```diff
@@ -158,6 +161,10 @@ private static void validatePatterns(EnumSource enumSource, Set<String> names) {
 			}
 		}
 
+		private interface Validator {
```
sbrannen: Much better than `BiConsumer<EnumSource, Set<String>>`. Thanks.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumArgumentsProvider.java
```diff
@@ -26,28 +27,47 @@
...
-		return constants.stream().map(Arguments::of);
+	private <E extends Enum<E>> Set<? extends E> getEnumConstants(ExtensionContext context) {
+		Class<E> enumClass = determineEnumClass(context);
+		return EnumSet.allOf(enumClass);
+	}
+
+	@SuppressWarnings({ "unchecked", "rawtypes" })
+	private <E extends Enum<E>> Class<E> determineEnumClass(ExtensionContext context) {
+		Class enumClass = enumSource.value();
+		if (enumClass.equals(NullEnum.class)) {
```
sbrannen: `==` should suffice here -- right?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumArgumentsProvider.java
```diff
@@ -26,28 +27,47 @@
...
+			Method method = context.getRequiredTestMethod();
+			Class<?>[] parameterTypes = method.getParameterTypes();
+			Preconditions.condition(parameterTypes.length > 0,
+				() -> "At least one parameter is required on test method: " + method.toGenericString());
```
sbrannen: At least one parameter is required on test method ... --> Test method X must declare at least one parameter.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumArgumentsProvider.java
```diff
@@ -26,28 +27,47 @@
...
+			Preconditions.condition(Enum.class.isAssignableFrom(parameterTypes[0]),
+				() -> "First parameter must reference an Enum type (alternatively, use the annotation's 'value' parameter to specify the type explicitly): "
```
sbrannen: 'value' parameter -> 'value' attribute

sbrannen:
Oops: I "reviewed" in the form of multiple comments. Sorry about that.

In any case, I think this looks pretty good.

documentation/src/test/java/example/ParameterizedTestDemo.java
marcphilipp: My intent was to demonstrate under which circumstances one needs to specify the enum class explicitly. Thus the switch from `TimeUnit` to `ChronoUnit`. I'm open to suggestions, of course. 🙂

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/NullEnum.java
marcphilipp: > Hope that doesn't cause any issues.

What issues would that cause?

> In the very least, it would make the Javadoc for NullEnum unavailable if package-private types are not included in the generated Javadoc.

You'd make it unavailable if package-private types _are_ included, right?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumArgumentsProvider.java
marcphilipp: It should unless there are different `ClassLoader` involved. Thus, I usually tend to use `equals()` to be safe.

documentation/src/test/java/example/ParameterizedTestDemo.java
marcphilipp: Shall I add a clarifying sentence?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumSource.java
marcphilipp: Done in 47aa8af2cd282a53790de0bb05edd7efaec14f2c.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumArgumentsProvider.java
marcphilipp: Done in 331468f2e51292417150f2027ac8c09d13fbc6df.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumArgumentsProvider.java
marcphilipp: Done in 331468f2e51292417150f2027ac8c09d13fbc6df.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EnumArgumentsProvider.java
sbrannen: > It should unless there are different `ClassLoader` involved.

Well, `java.lang.Class<T>` does not override `equals()`. So having a different `ClassLoader` would not make a difference in the outcome.

I suppose it's just a matter of taste.

documentation/src/test/java/example/ParameterizedTestDemo.java
sbrannen: Yes, please do.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/NullEnum.java
sbrannen: > > Hope that doesn't cause any issues.
> 
> What issues would that cause?

I just think it's somewhat unusual to have a non-public type as a default value for an annotation attribute. Other than potentially not publishing the Javadoc for it, I'm not immediately aware of any concrete runtime issues, but perhaps there's some use case we're not thinking of in which a user may wish to refer to the type in compiled user code.

> > In the very least, it would make the Javadoc for NullEnum unavailable if package-private types are not included in the generated Javadoc.
> 
> You'd make it unavailable if package-private types _are_ included, right?

If we leave it "as is" (package private) and do not publish package private members in the generated Javadoc, then the Javadoc would not be published for `NullEnum`.

sbrannen: For example, our build configures `memberLevel = JavadocMemberLevel.PROTECTED` which should correlate to:

> Shows only protected and public classes and members.

marcphilipp: I don't feel strongly about not making it `public`. Thus, I'll do it for the sake of consistency.

marcphilipp: Done in 5f243c2d0d1c823f4bac006d73cf90332dbcd038.

documentation/src/test/java/example/ParameterizedTestDemo.java
marcphilipp: Done in cfae38b44235ec077444d36cf944c7e26b6961e5.

PR #2012, opened 2019-09-15
Cache methods to improve performance of TestSource creation

### LINKED ISSUE #1965: Test discovery is very slow with scalatest
Our tests are written in scalatest and run by adding the annotation `@RunWith(classOf[JUnitRunner])` and using the maven surefire plugin. In one test class we have 14.000 tests. These tests take roughly 4 seconds to execute but 11 minutes to discover.

The issue seems to be connected with `VintageTestDescriptor`. It tries to find the method of a test using `VintageTestDescriptor.findMethod` which in turn enumerates all methods to find the one of the test.

I see two possible solutions:

* Do not call into `ReflectionUtils.findMethods` _for every test_. This means enumerating ~14.000 methods of that class (plus suplerclass methods) 14.000 times. If we could reuse the computed `List<Method>` of a class instead of recomputing it, that would solve the problem.

* Do not enumerate all methods to find one. This would mean changing `List<Method> findMethods(Class<?> clazz, Predicate<Method> predicate, HierarchyTraversalMode traversalMode)` and `List<Method> findAllMethodsInHierarchy(Class<?> clazz, HierarchyTraversalMode traversalMode)` to work all the way with `Stream`s instead of `Lists`s so that we do not need to continue enumerating methods once we have found the method, by leveraging the `Stream.findFirst` method. This would help in our particular case. Most of the time is actually spent in `ReflectionUtils.isMethodShadowedByLocalMethods`, which is only called for superclass methods. All of our tests are defined in the concrete class, so we would never need to enumerate that far.

I will attach a profile of `mvn test` during the test discovery phase for a better analysis.

marcphilipp:
Prior to this commit, we scanned the hierarchy of a test class for
methods once for every test method. This lead to quadratic runtime
behavior and was problematic for large test classes. Now, the new
`TestSourceProvider` caches the methods of a class so that we only need
to filter that list once for every test method.

## Before

Screenshot 2019-09-14 at 20 53 07

## After

Screenshot 2019-09-15 at 11 13 18

Resolves #1965.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
@@ -28,7 +28,7 @@ public class LruCache<K, V> extends LinkedHashMap<K, V> {
 
 	private static final long serialVersionUID = 1L;
 
-	private int maxSize;
+	private final int maxSize;
 
 	public LruCache(int maxSize) {
 		super(maxSize + 1, 0.75f, true);
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2015-2019 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.util;
+
+import static org.apiguardian.api.API.Status.INTERNAL;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import org.apiguardian.api.API;
+
+/**
+ * A simple LRU cache with a maximum size.
+ *
+ * @param <K> the type of keys maintained by this cache
+ * @param <V> the type of values maintained by this cache
+ */
+@API(status = INTERNAL, since = "1.6")
+public class LruCache<K, V> extends LinkedHashMap<K, V> {
+
+	private static final long serialVersionUID = 1L;
+
+	private int maxSize;
+
+	public LruCache(int maxSize) {
+		super(maxSize + 1, 0.75f, true);
+		this.maxSize = maxSize;
+	}
+
+	@Override
+	protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
+		return size() > maxSize;
+	}
+
+}
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
@@ -20,6 +20,8 @@ import org.apiguardian.api.API;
 /**
  * A simple LRU cache with a maximum size.
  *
+ * <p>This class is not thread-safe.
+ *
  * @param <K> the type of keys maintained by this cache
  * @param <V> the type of values maintained by this cache
  */
@@ -30,8 +32,18 @@ public class LruCache<K, V> extends LinkedHashMap<K, V> {
 
 	private final int maxSize;
 
+	/**
+	 * Create a new LRU cache that maintains at most the supplied number of
+	 * entries.
+	 *
+	 * <p>For optimal use of the internal data structures, you should pick a
+	 * number that's one below a power of two since this is based on a
+	 * {@link java.util.HashMap} and the eldest entry will be evicted after
+	 * adding the entry that increases the {@linkplain #size() size} to be above
+	 * {@code maxSize}.
+	 */
 	public LruCache(int maxSize) {
-		super(maxSize + 1, 0.75f, true);
+		super(maxSize + 1, 1, true);
 		this.maxSize = maxSize;
 	}
 
--- a/junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
+++ b/junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
@@ -17,12 +17,13 @@ import static org.junit.platform.commons.util.FunctionUtils.where;
 import static org.junit.platform.commons.util.ReflectionUtils.findMethods;
 
 import java.lang.reflect.Method;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.apiguardian.api.API;
 import org.junit.platform.commons.support.ModifierSupport;
+import org.junit.platform.commons.util.LruCache;
 import org.junit.platform.engine.TestSource;
 import org.junit.platform.engine.support.descriptor.ClassSource;
 import org.junit.platform.engine.support.descriptor.MethodSource;
@@ -34,7 +35,7 @@ import org.junit.runner.Description;
 @API(status = INTERNAL, since = "5.6")
 public class TestSourceProvider {
 
-	private final Map<Class<?>, List<Method>> methodsCache = new ConcurrentHashMap<>();
+	private final Map<Class<?>, List<Method>> methodsCache = Collections.synchronizedMap(new LruCache<>(32));
 
 	public TestSource findTestSource(Description description) {
 		Class<?> testClass = description.getTestClass();
--- a/junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
+++ b/junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
@@ -42,7 +42,7 @@ public class TestSourceProvider {
 		if (testClass != null) {
 			String methodName = description.getMethodName();
 			if (methodName != null) {
-				Method method = findMethod(testClass, methodName);
+				Method method = findMethod(testClass, sanitizeMethodName(methodName));
 				if (method != null) {
 					return MethodSource.from(testClass, method);
 				}
@@ -52,11 +52,15 @@ public class TestSourceProvider {
 		return null;
 	}
 
-	private Method findMethod(Class<?> testClass, String methodName) {
+	private String sanitizeMethodName(String methodName) {
 		if (methodName.contains("[") && methodName.endsWith("]")) {
 			// special case for parameterized tests
-			return findMethod(testClass, methodName.substring(0, methodName.indexOf("[")));
+			return methodName.substring(0, methodName.indexOf("["));
 		}
+		return methodName;
+	}
+
+	private Method findMethod(Class<?> testClass, String methodName) {
 		List<Method> methods = methodsCache.computeIfAbsent(testClass, clazz -> findMethods(clazz, m -> true)).stream() //
 				.filter(where(Method::getName, isEqual(methodName))) //
 				.collect(toList());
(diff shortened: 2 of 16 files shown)


junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.platform.engine.support.descriptor.MethodSource;
+import org.junit.runner.Description;
+
+/**
+ * @since 5.6
+ */
+@API(status = INTERNAL, since = "5.6")
+public class TestSourceProvider {
+
+	private final Map<Class<?>, List<Method>> methodsCache = new ConcurrentHashMap<>();
```
sbrannen: Please correct me if I'm wrong, but it appears that there is a single instance of `TestSourceProvider` per test plan. Thus, the `methodsCache` will grown unbounded, which could potentially cause memory issues for large test suites.

I'd therefore recommend that the `methodsCache` implement a _least recently used_ (LRU) eviction policy in order to limit the size of the cache over time. This can be achieved by tracking a "max size" and enforcing that in an overridden variant of `LinkedHashMap.removeEldestEntry(Entry<K, V>)`. The custom `LinkedHashMap` can then be wrapped in `Collections.unmodifiableMap()`.

It's hard to guess what the "max size" should be, but we could experiment with values like 8, 16, 32, etc. to see how it goes. It could also be made configurable, but that's likely overkill.

Thoughts?

jbduncan: (Alternatively if using a `LinkedHashMap` is infeasible because concurrency is needed, then the JUnit 5 team could try shadowing [Caffeine](https://github.com/ben-manes/caffeine) and using its cache class, at the cost of increasing the number of dependencies and increasing JUnit 5 artifact sizes.)

sbrannen: Oops...... did I type `unmodifiableMap()`? 😱 

I naturally meant `Collections.synchronizedMap()`.

jbduncan: @sbrannen Ahh, now that sounds like a pretty good first idea to try. :)

sbrannen: Yes, that works quite well, as I did [here](https://github.com/spring-projects/spring-framework/blob/cf5cb9484a3fd36a9418a4c7890f2bf82ffa369c/spring-test/src/main/java/org/springframework/test/context/cache/DefaultContextCache.java#L320).

ben-manes: [ConcurrentLinkedHashMap](https://github.com/ben-manes/concurrentlinkedhashmap) is commonly copied into projects since it is small, fast, and minimal. It might be a good choice if the synchronization becomes a bottleneck and you don’t want a larger, feature rich dependency.

sbrannen: Thanks for the pointer, @ben-manes.

marcphilipp: I followed the `LinkedHashMap` approach for now.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @param <K> the type of keys maintained by this cache
+ * @param <V> the type of values maintained by this cache
+ */
+@API(status = INTERNAL, since = "1.6")
+public class LruCache<K, V> extends LinkedHashMap<K, V> {
+
+	private static final long serialVersionUID = 1L;
+
+	private int maxSize;
```
sbrannen:
```suggestion
	private final int maxSize;
```

junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
```diff
@@ -0,0 +1,43 @@
...
+
+	public LruCache(int maxSize) {
+		super(maxSize + 1, 0.75f, true);
```
sbrannen: Why `maxSize + 1`?

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
```diff
@@ -0,0 +1,76 @@
...
+			}
+			return ClassSource.from(testClass);
+		}
+		return null;
+	}
+
+	private Method findMethod(Class<?> testClass, String methodName) {
+		if (methodName.contains("[") && methodName.endsWith("]")) {
+			// special case for parameterized tests
+			return findMethod(testClass, methodName.substring(0, methodName.indexOf("[")));
```
sbrannen: Why perform recursion here?

Why not just `methodName = methodName.substring(0, methodName.indexOf("["));`?

junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/RunListenerAdapter.java
```diff
@@ -36,11 +37,13 @@
 
 	private final TestRun testRun;
 	private final EngineExecutionListener listener;
+	private TestSourceProvider testSourceProvider;
```
sbrannen:
```suggestion
	private final TestSourceProvider testSourceProvider;
```

junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/RunnerExecutor.java
```diff
@@ -27,15 +28,17 @@
 public class RunnerExecutor {
 
 	private final EngineExecutionListener engineExecutionListener;
+	private TestSourceProvider testSourceProvider;
```
sbrannen:
```suggestion
	private final TestSourceProvider testSourceProvider;
```

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
marcphilipp: Probably because I don't like reassigning method parameters. 🤔

junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
marcphilipp: Because if the map already contains `maxSize` entries, it will add another one which will cause `HashMap` to resize and only then will it evict the eldest entry.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
sbrannen: Hmmmm... hmmmmmmmmmm....

sbrannen: If you really dislike reassigning method parameters, you could introduce a local variable such as `effectiveMethodName` to avoid the recursion, but in the end it's not a big deal for this use case.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
sbrannen: OK. Thanks for clarifying.

However, if we really want to ensure that the map is not unnecessarily resized, I don't think adding 1 will suffice due to the `loadFactor`. A [little research](https://stackoverflow.com/a/6400874/388980) reveals that `maxSize * 4/3` is a more suitable `initialCapacity` to avoid resizing.

WDYT?

marcphilipp: I think the crucial sentence from `HashMap`'s Javadoc is this one:

> If the initial capacity is _greater_ than the maximum number of entries divided by the load factor, no rehash operations will ever occur.

Thus, if load factor is `.75`, it needs to be `(maxSize * 4/3f) + 1`.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
marcphilipp: I extracted the first part into a separate method.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
marcphilipp: So, I think I found a good solution. Simply use a `loadFactor` of 1 and an `initialCapacity` of `maxSize + 1`. The new test shows that works reliably and is better than the solution above because it doesn't overestimate the maximum number of entries.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
```diff
@@ -30,8 +32,18 @@
...
+	/**
+	 * Create a new LRU cache that maintains at most the supplied number of
+	 * entries.
+	 *
+	 * <p>For optimal use of the internal data structures, you should pick a
+	 * number that's one below a power of two since this is based on a
+	 * {@link java.util.HashMap} and the eldest entry will be evicted after
+	 * adding the entry that increases the {@linkplain #size() size} to be above
+	 * {@code maxSize}.
+	 */
```
jbduncan: @marcphilipp I can tell that there is a link between powers of two and the fact that this cache extends `HashMap`, but it's not obvious to me what that link is, so I wondered if you could expand the javadoc to clarify things?

marcphilipp: Well, I could add a `@see HashMap`. Would that help?

jbduncan: @marcphilipp Thanks for the suggestion! However, when I search for "two", "2" or "power" in the [javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/HashMap.html) for `HashMap`, I don't get any useful information that clarifies the link between powers of two and hash maps.

I'm sure I'm missing something, so would you mind adding an `@see` pointing to the area in the javadoc that describes the relationship?

marcphilipp: You're right, it's not really documented, probably on purpose... 😉 

Basically, it uses an internal array of buckets with a length that's always a power of two. For every object, it will compute the hash an put it into one of the buckets. When it needs to resize, it doubles the number of buckets and rehashes all entries.

jbduncan: Ahh, okay, got it! :+1: 

It's unclear to me then if adding `@see HashMap` would help. What do you think about adding your description of `HashMap`'s internal resizing policy to the javadoc comment? :)

junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
sbrannen: OK. Let's stick with that approach. Assuming the hash codes returned for `java.lang.Class` result in an even distribution across the buckets, we shouldn't have collisions that result in slower performance.

I think we have now successfully over-engineered this feature. 🤓

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/TestSourceProvider.java
sbrannen: That looks much better. Thanks.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/LruCache.java
marcphilipp: I don't want to describe `HashMap`'s internals in our Javadoc in too much detail. It's an internal API so the comment is mainly there as a reminder. 😉

PR #2026, opened 2019-09-24
Add timeout config parameter

### LINKED ISSUE #1959: @Timeout should support debugging
We tried to use the new `@Timeout` annotation introduced in https://github.com/junit-team/junit5/pull/1885 to enforce a global timeout. I put it on the base class that all our JUnit Jupiter integration tests extend from. 

It works, but unfortunately without any kind of "DisableOnDebug" feature we cannot realistically use it. When debugging a test and stepping through with a debugger it will fail early with the timeout.

If adding a global "DisableOnDebug" is not being considered, I think it would be a small fix to `org.junit.jupiter.engine.extension.TimeoutExtension` itself.

Just need to add something like this (copied from JUnit 4's `DisableOnDebug`) to `TimeoutExtension` and not apply the timeout if `isDebugging()` is `true`.

```java
private static boolean isDebugging() {
    List<String> inputArguments = ManagementFactory.getRuntimeMXBean().getInputArguments();
    Iterator<String> iterator = inputArguments.iterator();
    String argument;
    do {
        if (!iterator.hasNext()) {
            return false;
        }

        argument = iterator.next();
        if ("-Xdebug".equals(argument)) {
            return true;
        }
    } while (!argument.startsWith("-agentlib:jdwp"));

    return true;
}
```

chrischild:
## Overview

Added a new configuration parameter (junit.jupiter.execution.timeout.mode) to determine if timeouts should be applied to annotated tests.

Couple of notes/questions:
1. Since the default was to have it be enabled but still allow someone to comment the parameter out I wondered whether there was any reason to see if the value was set to "enabled" or not. To me there's no reason so I only checked to see if the value was "disabled" or "disabled_on_debug". Thoughts?

2. Testing: I created one new test to make sure that setting the value to "disabled" truly did disable the timeouts. If there's a way to test the "disabled_on_debug" let me know.

Resolves #1959.

---

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
```diff
@@ -50,6 +55,7 @@ public void beforeEach(ExtensionContext context) {
 	}
 
 	private void readAndStoreTimeoutSoChildrenInheritIt(ExtensionContext context) {
+
```
sormuras: Remove extra empty line.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
```diff
@@ -10,6 +10,9 @@
 
 package org.junit.jupiter.engine.extension;
 
+import static org.junit.jupiter.engine.config.JupiterConfiguration.TIMEOUT_MODE_PROPERTY_NAME;
+
+import java.lang.management.ManagementFactory;
```
sormuras: Please add `requires java.management;` to `junit-jupiter-engine`'s [module-info.java](https://github.com/junit-team/junit5/blob/master/junit-jupiter-engine/src/module/org.junit.jupiter.engine/module-info.java#L12) file and update the expected list of requirements in this integration test: [junit-jupiter-engine.expected.txt](https://github.com/junit-team/junit5/blob/master/platform-tooling-support-tests/projects/jar-describe-module/junit-jupiter-engine.expected.txt)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutExtensionTests.java
```diff
@@ -16,13 +16,7 @@
...
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.DynamicTest.dynamicTest;
-import static org.junit.jupiter.engine.Constants.DEFAULT_AFTER_ALL_METHOD_TIMEOUT_PROPERTY_NAME;
-import static org.junit.jupiter.engine.Constants.DEFAULT_AFTER_EACH_METHOD_TIMEOUT_PROPERTY_NAME;
-import static org.junit.jupiter.engine.Constants.DEFAULT_BEFORE_ALL_METHOD_TIMEOUT_PROPERTY_NAME;
-import static org.junit.jupiter.engine.Constants.DEFAULT_BEFORE_EACH_METHOD_TIMEOUT_PROPERTY_NAME;
-import static org.junit.jupiter.engine.Constants.DEFAULT_TEST_FACTORY_METHOD_TIMEOUT_PROPERTY_NAME;
-import static org.junit.jupiter.engine.Constants.DEFAULT_TEST_METHOD_TIMEOUT_PROPERTY_NAME;
-import static org.junit.jupiter.engine.Constants.DEFAULT_TEST_TEMPLATE_METHOD_TIMEOUT_PROPERTY_NAME;
+import static org.junit.jupiter.engine.Constants.*;
```
sormuras: Restore the explicit import statements.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutExtensionTests.java
```diff
@@ -34,17 +28,7 @@
...
-import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.DisplayName;
-import org.junit.jupiter.api.DynamicTest;
-import org.junit.jupiter.api.Nested;
-import org.junit.jupiter.api.RepeatedTest;
-import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestFactory;
-import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.*;
```
sormuras: Restore the explicit import statements.

chrischild: Thanks I fixed the other ones. I think I ran an oraginze imports and it replaced them with the explicit import. Probably need to check a setting in my IDE.

chrischild:
I'm obviously hitting some errors but when I run gradlew test locally I'm seeing everything pass. What should I be running to see these errors?

sormuras:
https://github.com/junit-team/junit5/blob/da135a48f74cd0966979df96cd3d91aaf02dfdc9/.github/workflows/main.yml#L24

Without the '--scan' option, if you don't want to.

chrischild:
Closing for now. Thought I was closer to done but I looks like that was incorrect.

junit-jupiter-engine/src/module/org.junit.jupiter.engine/module-info.java
```diff
@@ -14,6 +14,7 @@
 	requires transitive org.junit.jupiter.api;
 	requires transitive org.junit.platform.engine;
 	requires transitive org.opentest4j;
+	requires transitive java.management;
```
sormuras: > Please add `requires java.management;` to ...

As the `java.management` module is only used internally and module `org.junit.jupiter.engine` doesn't "leak" a management type via it's API, there's no need for the `transitive` modifier. Please remove it here and in the integration test.

junit-jupiter-engine/src/module/org.junit.jupiter.engine/module-info.java
```diff
@@ -14,6 +14,7 @@
 	requires transitive org.junit.jupiter.api;
 	requires transitive org.junit.platform.engine;
 	requires transitive org.opentest4j;
+	requires java.management;
```
marcphilipp: Let's make this `static` so it's optional.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
```diff
@@ -165,6 +170,26 @@ private ScheduledExecutorService getExecutor(ExtensionContext extensionContext)
...
 	}
 
+	private boolean isDebuggingOrDisabled(ExtensionContext extensionContext) {
+		if (extensionContext.getConfigurationParameter(TIMEOUT_MODE_PROPERTY_NAME).isPresent()) {
+			if (extensionContext.getConfigurationParameter(TIMEOUT_MODE_PROPERTY_NAME).get().equals(DISABLED)) {
+				return true;
+			}
+			else if (extensionContext.getConfigurationParameter(TIMEOUT_MODE_PROPERTY_NAME).get().equals(
+				DISABLED_ON_DEBUG)) {
+				for (final String argument : ManagementFactory.getRuntimeMXBean().getInputArguments()) {
```
marcphilipp: We had issues using `java.lang.management` when this was introduced in junit4:
https://github.com/junit-team/junit4/pull/1187

I think we should wrap `ManagementFactory.getRuntimeMXBean().getInputArguments()` in a try-catch and catch `NoClassDefFoundError` (or whatever is thrown when the class is not present) to make sure this doesn't blow up on Android and other platforms. We should add a test with a custom `ClassLoader` and an integration test that excludes the module from the module path.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
```diff
@@ -165,6 +170,26 @@ private ScheduledExecutorService getExecutor(ExtensionContext extensionContext)
...
+					if ("-Xdebug".equals(argument)) {
```
sbrannen: From the [Java SE 8 documentation](https://docs.oracle.com/javase/8/docs/technotes/tools/windows/java.html#BABHDABI) for `java` command-line options:

> *-Xdebug*
> 
>     Does nothing. Provided for backward compatibility.

So there's no need to check for `-Xdebug` in JUnit Jupiter.

Checking for the `-agentlib:jdwp` should suffice.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
chrischild: I'll see what I can do for creating a new test.

chrischild: So I've been giving this a crack over the last couple days and I'm not seeing much progress. Modules and loading them is a little outside my breadth of knowledge. I do want to try some more but do you have some thoughts on where I should start?

sormuras: Here's the source of the integration test that checked the behavior for when module `java.scripting` is missing:

https://github.com/junit-team/junit5/blob/releases/5.4.x/platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/MultiReleaseJarTests.java#L71

Compare the `module-info.java` files in the underlying Maven-based projects at: https://github.com/junit-team/junit5/tree/releases/5.4.x/platform-tooling-support-tests/projects/multi-release-jar -- the one in `default` contains a `requires java.scripting;` directive, the one in `no-scripting` does not.

If you any help, please let me know.

chrischild: That was a good starting point. I can see how the code works up to a certain point. I see Maven's being run and calls all the tests inside each JupiterIntegrationTests class. I don't see how javaScriptingModuleIsAvailable throws any errors but it looks like it actually is.

sormuras: > I don't see how javaScriptingModuleIsAvailable throws any errors but it looks like it actually is.

As the module declared in project `no-scripting` doesn't read `java.scripting` that very module is **not** added to the set of visible modules at run time. As our (deprecated) code only declared a `static` (read: optional) read edge to `java.scripting` we had to ensure that we only use that module, if it is available at run time. That check, which needs to go in this PR, could look similar to this outline:

https://github.com/junit-team/junit5/blob/75907415611e1a8a9bc3e9476231a0d77bc7a23a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/ScriptExecutionCondition.java#L131-L160

sormuras:
> `ManagementFactory.getRuntimeMXBean().getInputArguments()`
...
> Checking for the `-agentlib:jdwp` should suffice.
...

Parsing the input arguments feels a brittle to me. Isn't there another and safer way to determinewhether the VM is running in debug mode at run time? /cc @raphw @forax

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
chrischild: Ahh there we go. OK I think I should be able to finish this up tomorrow now. Thanks for the quick responses and they were very helpful as well.

chrischild: I am stuck. I didn't want to pollute this PR so I have created another branch with the integration test [here](https://github.com/chrischild/junit5/tree/issues/integration-test). Specifically [Integration Test](https://github.com/junit-team/junit5/commit/17e982bc6841daef3d76d8cc9fbaef567e190e23) JupiterIntegrationTests and [Customer classloader](https://github.com/junit-team/junit5/commit/7f844fba1b7c9e560952aeaa0f31b8d56dc84c1d) MangementCondition.

chrischild: Sorry I don't think this is something I can figure out on my own. I went so far as to checkout the 5.4.3 code and ran the integration tests. I also changed the class that the ScriptExecutionCondition was trying to load to java.lang.management.ManagementFactory and the class stops throwing the exception that is expected when the java.management module is not in the module-info.java file. It continues on and fails later on when it tries to create an instance of the Evaluator which it should since the java.scripting module isn't loaded. 

Just to be sure I also left the class to be loaded as the ManagementFactory and added the java.scripting package to the no-scripting module-info to see if it would successfully build and it did, so I'm not quite sure where to go from there.

chrischild: Also the CI test is failing sure to a file locking error it looks like.

sormuras: > Also the CI test is failing due to a file locking error it looks like.

I'm fixing this issue at the moment and will push it to `master` soon.

> [...] so I'm not quite sure where to go from there.

I'll have a look at it next week or so. Perhaps, it's the _wrong_ way to test here anyway. As Android doesn't use the module system (does it?) yet, it'll be enough to guard the production code and test the same way as it is done in JUnit 4.

raphw:
> > `ManagementFactory.getRuntimeMXBean().getInputArguments()`
> > ...
> > Checking for the `-agentlib:jdwp` should suffice.
> > ...
> 
> Parsing the input arguments feels a brittle to me. Isn't there another and safer way to determine whether a VM is running in debug mode at run time? /cc @raphw @forax

Not really, you can check if the VM is listening to port 5005 but this does not help if the debugging port has changed. As far as I know, there is no good mechanism.

sbrannen:
> Not really, you can check if the VM is listening to port 5005 but this does not help if the debugging port has changed. As far as I know, there is no good mechanism.

Thanks for the feedback, @raphw.

So checking only for `-agentlib:jdwp` seems to still be the best option.

junit-platform-commons/src/module/org.junit.platform.commons/module-info.java
```diff
@@ -9,7 +9,8 @@
  */
 
 module org.junit.platform.commons {
-	requires java.logging; // TODO Is "requires transitive java.logging" needed here?
```
sormuras: Removed the superseded `TODO` on-the-fly -- the module API of `org.junit.platform.commons` does not use types of `java.logging`.

junit-platform-commons/src/module/org.junit.platform.commons/module-info.java
```diff
@@ -9,7 +9,8 @@
...
+	requires java.logging;
```
sormuras: I left the `static` modifier out on purpose -- because we need the module when running on the module path.

Otherwise, users have to add it via `--add-module java.management` or via their module descriptor(s).

sormuras:
> Since the default was to have it be enabled but still allow someone to comment the parameter out I wondered whether there was any reason to see if the value was set to "enabled" or not. To me there's no reason so I only checked to see if the value was "disabled" or "disabled_on_debug". Thoughts?

With https://github.com/junit-team/junit5/pull/2026/commits/ce1892cb02b3917d061c414f0b49f08cb6017a52 I enforced `enabled` to be the actual value for when a mode value is passed.

> Testing: I created one new test to make sure that setting the value to "disabled" truly did disable the timeouts. If there's a way to test the "disabled_on_debug" let me know.

That `"disabled_on_debug"` test case and the one that simulates an environment without JMX is still missing.

chrischild:
Ah I like it. Ok I'll take a peek and try and get those final missing pieces finished soon.

chrischild:
Huh I'm stumped on the JMX environment test. I tried my tests that I had created earlier and there was never a failure. I then commented out the java.management module change made by @sormuras in the platform module-info class
```
        requires java.logging;
//	requires java.management; // needed by RuntimeUtils to determine input arguments
	requires transitive org.apiguardian.api;
```
Also removed it from junit-platform-commons.expected to make sure a build should pass. I cleaned and then rebuilt and no errors occurred during the build.

I rechecked out this branch fresh and again removed the java.management line above and re-tested a clean build a third time and again there was no errors thrown. I don't see how this is possible unless it's being loaded somewhere else, maybe?

Anyways I'll check in the new unit test for the ```disabled_on_debug``` value.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutExtensionTests.java
```diff
@@ -88,6 +89,33 @@ void doesNotApplyTimeoutOnAnnotatedTestMethodsUsingDisabledTimeoutMode() {
...
+		// Should we test if we're debugging? This test will fail if we are debugging.
+		if (RuntimeUtils.isDebug()) {
+			assertThat(execution.getTerminationInfo().getExecutionResult().getThrowable()) //
+					.isEmpty();
+		}
+		else {
+			assertThat(execution.getTerminationInfo().getExecutionResult().getThrowable().orElseThrow()) //
+					.isInstanceOf(TimeoutException.class) //
+					.hasMessage("testMethod() timed out after 10 milliseconds");
+		}
```
chrischild: Comment isn't quite clear. Right now we need to check if we're debugging as the output of the test differs if so. If this is sufficient I'll remove the comment as part of another commit.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
sormuras: Resolved via https://github.com/junit-team/junit5/commit/ce1892cb02b3917d061c414f0b49f08cb6017a52

sormuras:
> Huh I'm stumped on the JMX environment test. [...]

I'll add it later this week to this branch ... or after merging it. Or #2054 will provide such an integration test as a side effect.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TimeoutExtensionTests.java
sormuras: We'll polish the test a bit after merging it to `master` -- so the comment helps as-is.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/Constants.java
```diff
@@ -425,6 +425,25 @@
...
+	 * <h3>Examples</h3>
+	 *
+	 * <ul>
+	 * <li>{@code enabled}: enables timeouts.
+	 * <li>{@code disabled}: disables timeouts.
+	 * <li>{@code disabled_on_debug}: disables timeouts while debugging.
+	 * </ul>
+	 *
+	 * @since 5.6
+	 */
```
marcphilipp: The Javadoc should document the default value.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TimeoutExtension.java
```diff
@@ -38,6 +42,9 @@
 	private static final ExtensionContext.Namespace NAMESPACE = ExtensionContext.Namespace.create(Timeout.class);
 	private static final String TESTABLE_METHOD_TIMEOUT_KEY = "testable_method_timeout_from_annotation";
 	private static final String GLOBAL_TIMEOUT_CONFIG_KEY = "global_timeout_config";
+	public static final String ENABLED_MODE_VALUE = "enabled";
+	public static final String DISABLED_MODE_VALUE = "disabled";
+	public static final String DISABLED_ON_DEBUG_MODE_VALUE = "disabled_on_debug";
```
marcphilipp: These should be `private`.

sormuras:
Thanks @chrischild

chrischild:
My pleasure. I wish I could've figured out the environment test though. I went back again and tried to see what else I might have missed but couldn't find a solution. I'll be keeping an eye on the Android integration test so I can see how it gets solved.

Issue #2030, opened 2019-09-26
Vintage Parameterized Tests are more than two orders of magnitude slower


marschall:
JUnit 5.5.2

## Steps to reproduce

Run the following empty test. With JUnit 4 it takes about 2 seconds to run, with the JUnit 5 vintage engine it takes about 250 seconds to run.

https://github.com/marschall/junit5-Segment-hashCode/blob/master/src/test/java/com/github/marschall/junit5/segmenthashcode/ReproducerTest.java

## Context

We have a JUnit 4 Parameterized test with about 70'000 total tests. 3 Tests with about 20'000 parameters. The test runs massively slower with the JUnit 5 vintage engine compared to the the JUnit 4 engine.

Profiling highlights the following hot spots:

- `TestRun#getInProgressTestDescriptorsWithSyntheticStartEvents` has a bad complexity (probably O(N²)) because every time it iterates over all started descriptors, it gets slower the more tests you have
- `UniqueId#hashCode` is not memoized 
- `UniqueId#toString` is not memoized

See the following flight recording

<img width="883" alt="TestResult" src="https://user-images.githubusercontent.com/471021/65680904-36a46c80-e058-11e9-81df-41084a245562.png">

marcphilipp:
Thanks for reporting this issue! 👍 

I had also seen that being slow in a recent profile I did in https://github.com/junit-team/junit5/pull/2012#issue-317622439.

I will improve it for 5.6.

marcphilipp:
@marschall I see you have already started, have you?

marschall:
Yes, the changes take the runtime of the posted empty test from ~250 seconds to ~4 seconds.

PR #2031, opened 2019-09-26
Improve performance of Vintage engine

marschall:
## Overview

Improves performance Parameterized vintage engine with the following
changes:

- add map to TestRun to track in progress TestDescriptors
- memoize UniqueId#hashCode
- memoize UniqueId#toString
- avoid intermediate char[] allocation when formatting UniqueId

Issue: #2030

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
@@ -82,6 +82,10 @@ public class UniqueId implements Cloneable, Serializable {
 
 	private final UniqueIdFormat uniqueIdFormat;
 	private final List<Segment> segments;
+	// lazily computed
+	private transient int hashCode;
+	// lazily computed
+	private transient String toString;
 
 	private UniqueId(UniqueIdFormat uniqueIdFormat, Segment segment) {
 		this.uniqueIdFormat = uniqueIdFormat;
@@ -221,7 +225,23 @@ public class UniqueId implements Cloneable, Serializable {
 
 	@Override
 	public int hashCode() {
-		return this.segments.hashCode();
+		int value = this.hashCode;
+		if (value == 0) {
+			value = this.segments.hashCode();
+			if (value == 0) {
+				// handle the edge case of the computed hashCode being 0
+				value = 1;
+			}
+			// this is a benign race like String#hash
+			// we potentially read and write values from multiple threads
+			// without a happens-before relationship
+			// however the JMM guarantees us that we only ever see values
+			// that were valid at one point, either 0 or the hash code
+			// so we might end up not seeing a value that a different thread
+			// has computed or multiple threads writing the same value
+			this.hashCode = value;
+		}
+		return value;
 	}
 
 	/**
@@ -230,7 +250,19 @@ public class UniqueId implements Cloneable, Serializable {
 	 */
 	@Override
 	public String toString() {
-		return this.uniqueIdFormat.format(this);
+		String s = this.toString;
+		if (s == null) {
+			s = this.uniqueIdFormat.format(this);
+			// this is a benign race like String#hash
+			// we potentially read and write values from multiple threads
+			// without a happens-before relationship
+			// however the JMM guarantees us that we only ever see values
+			// that were valid at one point, either null or the toString value
+			// so we might end up not seeing a value that a different thread
+			// has computed or multiple threads writing the same value
+			this.toString = s;
+		}
+		return s;
 	}
 
 	/**
--- a/junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/TestRun.java
+++ b/junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/TestRun.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -49,7 +50,8 @@ class TestRun {
 	private final Map<Description, List<VintageTestDescriptor>> descriptionToDescriptors;
 	private final Map<TestDescriptor, List<TestExecutionResult>> executionResults = new LinkedHashMap<>();
 	private final Set<TestDescriptor> skippedDescriptors = new LinkedHashSet<>();
-	private final Map<TestDescriptor, EventType> startedDescriptors = new LinkedHashMap<>();
+	private final Set<TestDescriptor> startedDescriptors = new HashSet<>();
+	private final Map<TestDescriptor, EventType> inProgressDescriptors = new LinkedHashMap<>();
 	private final Set<TestDescriptor> finishedDescriptors = new LinkedHashSet<>();
 
 	TestRun(RunnerTestDescriptor runnerTestDescriptor) {
@@ -74,10 +76,9 @@ class TestRun {
 	}
 
 	Collection<TestDescriptor> getInProgressTestDescriptorsWithSyntheticStartEvents() {
-		List<TestDescriptor> result = startedDescriptors.entrySet().stream() //
+		List<TestDescriptor> result = inProgressDescriptors.entrySet().stream() //
 				.filter(entry -> entry.getValue().equals(EventType.SYNTHETIC)) //
 				.map(Entry::getKey) //
-				.filter(descriptor -> !isFinished(descriptor)) //
 				.collect(toCollection(ArrayList::new));
 		Collections.reverse(result);
 		return result;
@@ -127,14 +128,16 @@ class TestRun {
 	}
 
 	void markStarted(TestDescriptor testDescriptor, EventType eventType) {
-		startedDescriptors.put(testDescriptor, eventType);
+		inProgressDescriptors.put(testDescriptor, eventType);
+		startedDescriptors.add(testDescriptor);
 	}
 
 	boolean isNotStarted(TestDescriptor testDescriptor) {
-		return !startedDescriptors.containsKey(testDescriptor);
+		return !startedDescriptors.contains(testDescriptor);
 	}
 
 	void markFinished(TestDescriptor testDescriptor) {
+		inProgressDescriptors.remove(testDescriptor);
 		finishedDescriptors.add(testDescriptor);
 	}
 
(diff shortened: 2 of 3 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
```diff
@@ -82,10 +83,14 @@ public static UniqueId root(String segmentType, String value) {
 
 	private final UniqueIdFormat uniqueIdFormat;
 	private final List<Segment> segments;
+	private transient int hashCode;
+	private transient String toString;
```
sormuras: Transient in order to keep former serialized objects de-serializable?

marschall: Yes and setting them in `readObject`.

sormuras: Wonder if we have to go this route. As the expected life-cycle of a serialized `UniqueId` object is short. Like within an IDE-based testing session. @junit-team/junit-lambda what do think?

marschall:
I'm at a loss why `org.junit.jupiter.engine.extension.TimeoutExtensionTests > appliesDefaultTimeoutsFromConfigurationParameters()` fails, it runs locally and for the other setups.

I don't know why the macOS tests fail, I just get a ❌ and no log.

As this chance includes no API changes should I simply tick the checkbox in DoD or remove the DoD item entirely?

sormuras:
I restartet the CI run, which usually help. But it didn't help the situation this time at all:

> git checkout --progress --force c4b143c088f7c3c867a34ab0074e841de99cfcc7
> ##[error]fatal: reference is not a tree: c4b143c088f7c3c867a34ab0074e841de99cfcc7
> ##[warning]Git checkout failed on shallow repository, this might because of git fetch with depth '10' doesn't include the checkout commit 'c4b143c088f7c3c867a34ab0074e841de99cfcc7'.

_sigh_

junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
```diff
@@ -82,10 +84,14 @@ public static UniqueId root(String segmentType, String value) {
...
 
 	private UniqueId(UniqueIdFormat uniqueIdFormat, Segment segment) {
 		this.uniqueIdFormat = uniqueIdFormat;
 		this.segments = singletonList(segment);
+		this.hashCode = this.segments.hashCode();
+		this.toString = this.uniqueIdFormat.format(this);
```
marcphilipp: I'm not sure if we should do this eagerly. Doing so might slow down other use cases. WDYT?

junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/TestRun.java
```diff
@@ -74,10 +75,9 @@ RunnerTestDescriptor getRunnerTestDescriptor() {
 	}
 
 	Collection<TestDescriptor> getInProgressTestDescriptorsWithSyntheticStartEvents() {
-		List<TestDescriptor> result = startedDescriptors.entrySet().stream() //
+		List<TestDescriptor> result = inProgressDescriptors.entrySet().stream() //
```
marcphilipp: ❓ Can we change `startedDescriptors` to a `Set<TestDescriptor>` now?

junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
sormuras: This is `String::format`, ain't it?

If yes, this will be blazing fast ... 🔜 https://openjdk.java.net/jeps/348

marcphilipp: Well, maybe, but it will still be slower than doing nothing... 😉

junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/TestRun.java
```diff
@@ -50,6 +50,7 @@
 	private final Map<TestDescriptor, List<TestExecutionResult>> executionResults = new LinkedHashMap<>();
 	private final Set<TestDescriptor> skippedDescriptors = new LinkedHashSet<>();
 	private final Map<TestDescriptor, EventType> startedDescriptors = new LinkedHashMap<>();
+	private final Map<TestDescriptor, EventType> inProgressDescriptors = new HashMap<>();
```
marcphilipp: IIRC the order is important so this should be a `LinkedHashMap`.

junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
sormuras: True.

sbrannen: > I'm not sure if we should do this eagerly. Doing so might slow down other use cases. WDYT?

marcphilipp:
@marschall I've removed the fetch-depth from the GitHub Actions workflows. Please rebase on `master`.

junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
marschall: @sbrannen here's a screenshot from a [flight recording](https://github.com/marschall/junit5-Segment-hashCode/raw/master/src/main/jfr/junit-5.5.2.jfr) from this test https://github.com/marschall/junit5-Segment-hashCode/blob/master/src/test/java/com/github/marschall/junit5/segmenthashcode/ReproducerTest.java

Screenshot

marschall: @sormuras I don't see how this helps. JEP 348 talks about `String#format`, we're calling `UniqueIdFormat#format` which does not end up calling `String#format`. JEP 348 explicitly is only for JDK code. JEP 348 gave up on intrinsifing `Objects#hashCode`.

marschall:
Now setting up JDK 12 is failing

```
Exception calling "ExtractToDirectory" with "2" argument(s): "End of Central Directory record could not be found."
At line:1 char:111
+ ... catch { } ; [System.IO.Compression.ZipFile]::ExtractToDirectory('D:\a ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : InvalidDataException
 
##[error]The process 'C:\windows\System32\WindowsPowerShell\v1.0\powershell.exe' failed with exit code 1
##[error]Node run failed with exit code 1
```
```
/usr/bin/tar xzC /Users/runner/runners/2.158.0/work/_temp/temp_1171158186 -f /Users/runner/runners/2.158.0/work/_temp/c24f05d8-02a2-4fac-929f-6bb145d91e4f
tar: Unrecognized archive format
tar: Error exit delayed from previous errors.
##[error]The process '/usr/bin/tar' failed with exit code 1
##[error]Node run failed with exit code 1
```
```
/usr/bin/docker exec  d33d2f444f9c9eaa18ca7de883c14e3a99c22a9f0b461d0b4f0314bf5c287193 sh -c "cat /etc/*release | grep ^ID"
Running JavaScript Action with default external tool: node12
/bin/tar xzC /__w/_temp/temp_984953448 -f /__w/_temp/6abd6cf7-d10f-4a3b-967e-2da2937b5200

gzip: stdin: not in gzip format
/bin/tar: Child returned status 1
/bin/tar: Error is not recoverable: exiting now
##[error]The process '/bin/tar' failed with exit code 2
##[error]Node run failed with exit code 1
```

junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
sormuras: > [...] which does not end up calling `String#format`.

Wasn't sure about that and didn't verified my guess two days ago. Thanks for the heads-up.

> JEP 348 gave up on intrinsifing `Objects#hashCode`.

Hm, that and the information from above point towards computing and storing those values already in the constructor. Iff those values are always used at least once in the lifetime of an instance, methinks.

sormuras:
> Now setting up JDK 12 is failing

Beta is a beta... is a beta. _sigh again_ Same on Mac OS.

sormuras:
https://github.com/actions/setup-java/issues/23

junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueId.java
sbrannen: > @sbrannen here's a screenshot from a flight recording

OK. You convinced me.

marschall: @marcphilipp I believe both of them ultimately always end up being called

`#hashCode`

The hash based collections in:

- org.junit.platform.engine.support.discovery.EngineDiscoveryRequestResolution
- org.junit.platform.engine.support.descriptor.AbstractTestDescriptor#addChild
- org.junit.vintage.engine.execution.TestRun

`#toString`

- org.junit.platform.launcher.TestIdentifier#from(TestDescriptor)

I have to admit doing it eagerly is a bit of a cop out because there are many options and trade-offs for computing it lazily if it has to be thread safe (I assume it has to be).
- do we want to use `AtomicReference`?
- do we want to use `volatile` and an `AtomicIntegerFieldUpdater`?
- do we want to use `volatile` and a lock, if so
  - do we want a `synchronized` or a `java.util.concurrent.locks.Lock`?
  - how many locks do we want? a single static one? one per object? one per object and field?
- do we want a benign race like `String#hash`?

marcphilipp: > Iff those values are always used at least once in the lifetime of an instance, methinks.

Not necessarily since some `TestDescriptors` might get pruned from the test plan. Thus, I think we should go for lazy initialization instead. We could reuse the `MemoizingSupplier` that's already in the codebase:
https://github.com/junit-team/junit5/blob/5e41ebe612fd67e905f96f1dd3184a071b65be17/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/ExtensionValuesStore.java#L164-L192.

marcphilipp: On the other hand that would require a lot of additional objects... 🤔

marcphilipp: I think a benign race would be sufficient because if it ends up being used `hashCode()` and `toString()` will be called from a single thread before the potentially multi-threaded execution begins.

sbrannen: > do we want a benign race like String#hash?

marcphilipp:
Thanks a lot, @marschall! 👍

sbrannen:
> Thanks a lot, @marschall! 👍

+1

PR #2032, opened 2019-09-26
Implement safer variant to ensure CDATA end markers result in valid XML

chschu:
## Overview

This PR changes the implementation of `XmlReportWriter. writeCDataSafely(...)` to be even safer by always passing valid CDATA content to `javax.xml.stream.XMLStreamWriter.writeCData(...)`.

The current implementation passes a String to `javax.xml.stream.XMLStreamWriter.writeCData(...)` that might not be valid CDATA content, i.e. it might contain `]]>`. This causes issues if the implementation of that method validates its parameter value.

The PR does not change the produced XML, which is covered by `XmlReportWriterTests.writesValidXmlEvenIfExceptionMessageContainsCData()`.

---

---

### Definition of Done

- [X] Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc

### DIFF
--- a/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java
+++ b/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.TreeSet;
+import java.util.regex.Pattern;
 
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
@@ -51,8 +52,7 @@ import org.junit.platform.launcher.listeners.LegacyReportingUtils;
  */
 class XmlReportWriter {
 
-	private static final String CDATA_START = "<![CDATA[";
-	private static final String CDATA_END = "]]>";
+	private static final Pattern CDATA_SPLIT_PATTERN = Pattern.compile("(?<=]])(?=>)");
 
 	private final XmlReportData reportData;
 
@@ -296,7 +296,9 @@ class XmlReportWriter {
 	}
 
 	private void writeCDataSafely(XMLStreamWriter writer, String data) throws XMLStreamException {
-		writer.writeCData(data.replace(CDATA_END, "]]" + CDATA_END + CDATA_START + ">"));
+		for (String safeDataPart : CDATA_SPLIT_PATTERN.split(data)) {
+			writer.writeCData(safeDataPart);
+		}
 	}
 
 	private void newLine(XMLStreamWriter xmlWriter) throws XMLStreamException {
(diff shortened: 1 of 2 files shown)


marcphilipp:
@chschu Thanks! Could you please add a test case that demonstrates the issue and ensures we don't introduce it again?

chschu:
@marcphilipp I didn't want to expose more of the `XmlReportWriter` implementation just for the test. The arguments of `XMLStreamWriter.writeCData(...)` cannot be captured directly in a test, because the `XMLOutputFactory` is obtained using a static call.

The test I added checks that the data passed to the underlying `Writer` never contains the problematic substring `]]><![CDATA[`. That shouldn't ever happen if CDATA blocks are written separately.

The CI failure on Windows looks like a file locking issue that has nothing to do with my changes.

junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java
```diff
@@ -51,8 +52,7 @@
  */
 class XmlReportWriter {
 
-	private static final String CDATA_START = "<![CDATA[";
-	private static final String CDATA_END = "]]>";
+	private static final Pattern CDATA_SPLIT_PATTERN = Pattern.compile("(?<=]])(?=>)");
```
marcphilipp: ❓ Why is this pattern so complicated? Is the positive lookbehind and the positive lookahead necessary?

platform-tests/src/test/java/org/junit/platform/reporting/legacy/xml/XmlReportWriterTests.java
```diff
@@ -219,10 +220,44 @@ void writesValidXmlEvenIfExceptionMessageContainsCData() throws Exception {
...
+			}
+
+			@Override
+			public void flush() {
+			}
+
+			@Override
+			public void close() {
+			}
+		};
```
sbrannen: Please replace that block with the following.

```java
		Writer assertingWriter = new StringWriter() {

			@Override
			public void write(char[] cbuf, int off, int len) {
				assertThat(new String(cbuf, off, len)).doesNotContain("]]><![CDATA[");
			}
		};
```

junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java
sbrannen: And if the the positive lookbehind and  positive lookahead are necessary, please add comments in the code to explain exactly what the pattern matches and why.

marcphilipp: @chschu Could you please take a look? We'd like to include this in 5.6 M1.

chschu: Sorry for the delay.

The zero-width assertions are used to simplify the loop in `writeCDataSafely(...)`. Any characters captured by the pattern itself would have to be added to the results of `String.split(...)`.

I added a comment that should clarify this.

marcphilipp: Thanks! 👍

marcphilipp:
Thanks, @chschu! 👍

Issue #1914, opened 2019-06-04
Support test methods with Kotlin suspend modifier


IRus:
## Goals

Support running suspend test in JUnit Jupiter:

```kotlin
class Kit {
    @Test 
    suspend fun foo() {
        delay(1000) // suspend call
        assertEquals(1, 1)
    } 
}
```

Currently, such test can be written this way:

```kotlin
class Kit {
    @Test 
    fun foo() = runBlocking {
        delay(1000) // suspend call
        assertEquals(1, 1)
        assertThrows { /* ... */ } 
        Unit // test should return void, but `assertThrows` returns `Throwable`, so `foo` returns `Throwable` too 
    } 
}
```

Also, will be nice to provide `CoroutineScope` through params, or as receiver in extension:

```kotlin
class Kit {
    suspend fun foo(scope: CoroutinesScope) {  /* ... */  } // (1)
    suspend fun CoroutinesScope.foo() {  /* ... */  } // (2)
}
```
1 and 2 actually the same on bytecode level. `suspend` is optional.

And finally, support for `runBlockingTest`:

```kotlin
class Kit {
    suspend fun foo(scope: TestCoroutinesScope) {  /* ... */  }
    suspend fun TestCoroutinesScope.foo() {  /* ... */  } 
}
```

## What can be done currently

`ParameterResolver` can be used to provide stubs for `Continuation`, `CoroutineScope` and `TestCoroutineScope`. These stub arguments can be replaced with real arguments in invocation.

## Problems

Current extensions points not enough to implement this feature as extensions, since:

1. **Discovery**. Jupiter discovers tests that returns `void`, but `suspend fun` returns `Object`;
2. **Invocation**. `InvocationInterceptor` in 5.5-M1(SNAPSHOT) don't providing mechanism to override actual invocation, only to decoration of existing invocation. Conversion of `method` to `kotlinFunction`, and then executing using [`callSuspend`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect.full/call-suspend.html) is necessary to execute `suspend fun`.

Also, my [slides](https://ruslan.ibragimov.by/pages/cv/speaker/kotlin-night-kiev-2019-junit.pdf) about this topic.

sormuras:
Related to #1851

marcphilipp:
@IRus This is not resolved by #2042, is it?

IRus:
@marcphilipp Yes, not resolved, this is different issues

[...]

PR #1853, opened 2019-04-09
Issue: #1851 Support assertThrows() with suspending functions in Kotlin

DerkSchooltink:
## Overview

I have modified the assertThrows() functions to accept suspending functions as body. This means now all functions using the suspend modifier for a function can be used by this API as well. Conventional methods without suspend modifier can be passed as well, not resulting in different behaviour.

This should now be possible:

```
suspend fun test(): Nothing = throw Exception()

assertThrows<Exception> {
    test()
}
```

Few topics open for discussion:

- Util.kt added to provide suspending functions to assert failures, not sure if this is the most optimal place. Could be added to relevant tests as well as private functions, I am not totally sure.

Resolves #1851.

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement).

---

### Definition of Done

- [X] There are no TODOs left in the code
- [ ] Method [preconditions](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [X] [Coding conventions](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [X] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling
- [X] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [ ] Change is documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [X] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #2042, opened 2019-10-05
Issue #1851: Inline assertThrows lambda arguments

### LINKED ISSUE #1851: Support assertThrows() with suspending functions in Kotlin
Often when I write a unit test that has to assert that a certain exception is thrown I use the `assertThrows()` method. But I find myself regularly having to test suspending functions, so I have to wrap my function with a `runBlocking()`. This is a lot of repeating code if your entire codebase is heavily based on coroutines. 

### Proposed fix

I couldn't possibly have made this issue without also thinking about a solution.

Effectively, this is what I came up with:

```kotlin
private inline fun <reified T : Throwable> assertThrows(
        noinline executable: suspend () -> Unit
    ) = assertThrows(T::class.java) {
        runBlocking {
            executable()
        }
    }
```

This way you can pass suspending functions as parameter `executable`, which will be invoked in a `runBlocking()` to make sure the suspending function joins before continuing.

### tl;dr

I think it would be a good addition to JUnit Jupiter to have functions that allow easy testing of suspending functions.

ashdavies:
## Overview
**Resolves #1851 #1914**

Since provided executables for `assertThrows` invocations are marked as `noinline`, it is not possible to execute them with the same context as the call site. Thus it is not possible to call `assertThrows` on a suspended method from a `CoroutineContext`.

By including overloaded inline methods accepting the `Result<*>` parameter, it is possible to defer a `Throwable` being thrown until it can be evaluated in the java `Executable` instance, despite having already been executed in the calling context.

This approach resolves the issue of being unable to call `assertThrows` be suspended functions, without having to include an additional dependency on Coroutines, and inherits the calling Coroutine hierarchy.

Solution provided alongside #1853 which resolves the same issue through use of `runBlocking`.

---

---

### Definition of Done


marcphilipp:
Looks like it doesn't compile... 🤔 
https://scans.gradle.com/s/b6i2vqkyxxqim/console-log#L39

JLLeitschuh:
Since this method is `inline` and so no code actually calls these methods, this technically isn't an API breaking change, correct?

JLLeitschuh:
I think that it would be appropriate to include an example in the documentation on how to use this with a coroutine so that people looking to test coroutines with this new method know what to do.

ashdavies:
@JLLeitschuh is that really necessary? The idea here is to resolve not being able to inline context, and doesn't necessarily need to involve Coroutines at all. I intend to write an article about it anyhow, but so far I see this as just improving the flexibility of the existing `assertThrows` methods, rather than catering for a specific use case.

However, should it be felt as important, I can add documentation, but I'm not familiar with the documentation structure for this repository. Any hints as to where I should be looking to update/introduce documentation?

JLLeitschuh:
I just saw your documentation changes:
I was more looking for an example around how to use these new methods with coroutine results.

ashdavies:
@JLLeitschuh I was looking at how to do this, I opted for the stubbed context classes as a coroutine example would require importing all the necessary libraries just for the documentation which seems both unnecessary and difficult 😆

marcphilipp:
@IRus @DerkSchooltink Could you please confirm that this solves your coroutine use case?

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -95,8 +95,21 @@ fun assertAll(heading: String?, vararg executables: () -> Unit) =
...
+ * Example usage:
+ * ```kotlin
+ * val exception = assertThrows<IllegalArgumentException>(runCatching {
+ *     throw IllegalArgumentException("Talk to a duck")
+ * })
+ * assertEquals("Talk to a duck", exception.message)
+ * ```
+ * @see Assertions.assertThrows
+ */
+inline fun <reified T : Throwable> assertThrows(result: Result<*>): T =
```
IRus: Not sure if benefits of this method enough, to be part of API.

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -121,8 +134,21 @@ inline fun <reified T : Throwable> assertThrows(message: String, noinline execut
...
+ * Example usage:
+ * ```kotlin
+ * val exception = assertThrows<IllegalArgumentException>({ "Should throw an Exception" }, runCatching {
+ *     throw IllegalArgumentException("Talk to a duck")
+ * })
+ * assertEquals("Talk to a duck", exception.message)
+ * ```
+ * @see Assertions.assertThrows
+ */
+inline fun <reified T : Throwable> assertThrows(noinline message: () -> String, result: Result<*>): T =
```
IRus: Same question as for line 111

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
ashdavies: Inline functions cannot be called from a non-public API, it would need to be collapsed into the calling method.
Is that what you're asking?

IRus: I'm asking if it makes sense to expose this method as part of API. If they not needed (I think so), they should be collapsed.

ashdavies: Yeah it's not completely necessary, and probably doesn't provide a large amount of benefit

ashdavies:
Sorry for the delay, I'm having some issues with Java 13 on my development machine

marcphilipp:
@ashdavies Why Java 13?

ashdavies:
@marcphilipp I actually I have no idea, I had confused the projects Java 11 dependency with another project I was working on 🤦‍♂ 

I still have some issues setting up this project locally with IntelliJ but that's probably a local configuration issue (doesn't find Java classpath).

ashdavies:
The goal here is to adjust the existing API so that `noinline` can be removed from `Executable`, so the intention is to ensure existing tests pass and provide further affirmation that inlined contexts, such as a `CoroutineContext` can be used with `assertThrows`.

I've not included the actual `CoroutineContext` in the test to avoid the dependency, but I've created a `ScopedContext` class to simulate this behaviour.

marcphilipp:
Please let me know when this PR is ready for another review.

ashdavies:
@marcphilipp yep, ready for review, as mentioned, the idea is that not much additional testing is necessary since it's only allowing the lambda to be called inline with `assertThrows` thus so long as the existing assertions pass it should be doing its job. I've added another test to verify the scope behaviour.

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/ContextScope.kt
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+package org.junit.jupiter.api
+
+interface ContextScope {
```
DerkSchooltink: Related to my previous comment, but this interface doesn't seem to be useful. See previous comment for explanation.

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
```diff
@@ -60,6 +60,16 @@ class KotlinAssertionsTests {
         assertThrows<AssertionError>({ "should fail" }) { fail(null as Throwable?) }
     }
 
+    @Test
```
DerkSchooltink: A general statement for testing this implementation:

There are no scenario's that cover an actual suspending function. Even though it works with suspending functions, I've verified this myself as well, I'm still worried about future adjustments to this code that might change that behaviour. You would never find out because there are no unit tests that will fail. 

I must admit I do not know how to effectively test this either, too. But I think _something_ should be done to avoid future regression issues.

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -121,8 +125,14 @@ inline fun <reified T : Throwable> assertThrows(message: String, noinline execut
  * ```
  * @see Assertions.assertThrows
  */
-inline fun <reified T : Throwable> assertThrows(noinline message: () -> String, noinline executable: () -> Unit): T =
-    Assertions.assertThrows(T::class.java, Executable(executable), Supplier(message))
+inline fun <reified T : Throwable> assertThrows(noinline message: () -> String, executable: () -> Unit): T {
+    val result: Result<*> = runCatching(executable)
```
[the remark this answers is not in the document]
ashdavies: I know but personally I prefer using explicit types as it's not always clear when reviewing outside of the IDE, but I hadn't noticed from the surrounding code style that this was used, consistency wins 👍

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
ashdavies: Agreed, the tests I've included here only attempt to "simulate" a `CoroutineContext`, short of actually including Coroutines I don't know of a good solution to this problem. Though since it's would only be for test this could be easier to justify, thoughts @marcphilipp ?

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/ScopedContext.kt
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+package org.junit.jupiter.api
+
+fun <T> contextScope(block: ContextScope.() -> T): T = object : ContextScope {
```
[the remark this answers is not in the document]
ashdavies: The idea was that `apply` can only be called from `ContextScope` so that when calling it from within `assertThrows` it verifies that the scope has been inlined, thus behaving in the same was as `CoroutineScope.launch { }`

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -121,8 +125,14 @@ inline fun <reified T : Throwable> assertThrows(message: String, noinline execut
...
+    val executable = Executable {
```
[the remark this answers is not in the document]
ashdavies: Honestly, I just find the difference in line length jarring

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -126,12 +125,9 @@ inline fun <reified T : Throwable> assertThrows(message: String, executable: ()
...
  */
 inline fun <reified T : Throwable> assertThrows(noinline message: () -> String, executable: () -> Unit): T {
-    val result: Result<*> = runCatching(executable)
-    val executable = Executable {
-        result.getOrThrow()
-    }
+    val result = runCatching(executable)
 
-    return Assertions.assertThrows(T::class.java, executable, Supplier(message))
+    return Assertions.assertThrows(T::class.java, Executable { executable }, Supplier(message))
```
DerkSchooltink: `executable` should be `result`, doesn't compile

ashdavies: 🤦‍♂

local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
DerkSchooltink: I think you committed this by accident :P

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
IRus: `kotlinx.coroutines` can be added in test classpath, so then you'll not need this ContextScope

DerkSchooltink: > `kotlinx.coroutines` can be added in test classpath, so then you'll not need this ContextScope

Good idea! Using [kotlinx-coroutines-test](https://github.com/Kotlin/kotlinx.coroutines/tree/master/kotlinx-coroutines-test) should work just fine. Hopefully this is allowed from a dependency pollution standpoint.

marcphilipp: As long as it's on the test classpath only, that's fine!

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
```diff
@@ -60,6 +61,14 @@ class KotlinAssertionsTests {
...
+    fun `expected context exception testing`() = runBlocking<Unit> {
+        assertThrows<AssertionError>("Should fail async") { failAsync("Should fail async") }
```
DerkSchooltink: Maybe you can use an anonymous function here to avoid having to create the private function below, which clutters up the tests:

```
        assertThrows<AssertionError>("Should fail async") {
            suspend { fail("Should fail async") }()
        }
```

IRus: Agree, creating coroutine in place looks good

DerkSchooltink:
@marcphilipp can you review the current implementation? It would be cool to merge this PR to get closer to the 5.7 milestone :)

junit-jupiter-engine/junit-jupiter-engine.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
 	testImplementation(project(":junit-platform-launcher"))
 	testImplementation(project(":junit-platform-runner"))
 	testImplementation(project(":junit-platform-testkit"))
+	testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:${Versions.coroutines}")
```
marcphilipp: The version should be declared in `dependencies.gradle.kts` instead.

ashdavies:
So..... can this be merged? 😅

DerkSchooltink:
> So..... can this be merged? 😅

Merge it!

IRus:
:shipit:

ashdavies:
> > So..... can this be merged? 😅
> 
> Merge it!

I don't have write permission 😅

DerkSchooltink:
> > > So..... can this be merged? 😅
> > 
> > 
> > Merge it!
> 
> I don't have write permission 😅

@marcphilipp it is up to you then I guess :P

dependencies/dependencies.gradle.kts
```diff
@@ -25,5 +25,6 @@ dependencies {
 		api("com.tngtech.archunit:archunit-junit5-api:${Versions.archunit}")
 		api("com.tngtech.archunit:archunit-junit5-engine:${Versions.archunit}")
 		api("org.slf4j:slf4j-jdk14:${Versions.slf4j}")
+		api("org.jetbrains.kotlinx:kotlinx-coroutines-core:${Versions.coroutines}")
```
JLLeitschuh: Should this be `compileOnly`?

IRus: This file used as BOM: https://github.com/junit-team/junit5/blob/master/junit-jupiter-engine/junit-jupiter-engine.gradle.kts#L11

Actual dependency added here: https://github.com/junit-team/junit5/pull/2042/files#diff-923b2e78200b0be136b71b3fec482d83R22

See previous comment: https://github.com/junit-team/junit5/pull/2042#discussion_r385882647

JLLeitschuh: 🔥 Thanks for checking!

marcphilipp:
I'll add an entry to the release notes and will then merge it.

marcphilipp:
Thanks everyone!

Please let me know if the release note entry is worded correctly:
77930f85397faf19543a29e47a629cfe2d2cc237

ashdavies:
LGTM, thanks for merging 👍

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -95,8 +95,11 @@ fun assertAll(heading: String?, vararg executables: () -> Unit) =
  * ```
  * @see Assertions.assertThrows
  */
-inline fun <reified T : Throwable> assertThrows(noinline executable: () -> Unit): T =
-    Assertions.assertThrows(T::class.java, Executable(executable))
+inline fun <reified T : Throwable> assertThrows(executable: () -> Unit): T {
+    val result = runCatching(executable)
+
+    return Assertions.assertThrows(T::class.java) { result.getOrThrow() }
+}
```
JLLeitschuh: Sorry, just had this realization. `runCatching` was introduced in Kotlin 1.3.

Does this have the implication that this method will now only be supported in Kotlin 1.3 and higher?

IRus: This shouldn't be a problem, since coroutines become stable in [1.3](https://blog.jetbrains.com/kotlin/2018/10/kotlin-1-3/), and this method designed for coroutines users

marcphilipp: Hmm, good question. We have this integration test in place that's [compiled against Kotlin 1.1](https://github.com/junit-team/junit5/blob/b01802956c05822d59ec73cbf377529a3abfb5f8/platform-tooling-support-tests/projects/gradle-kotlin-extensions/build.gradle.kts#L27-L28). Shouldn't that be broken? 🤔 
https://github.com/junit-team/junit5/blob/3f7fed61f2edcee2a5551dbde75382b1dbe21267/platform-tooling-support-tests/projects/gradle-kotlin-extensions/src/test/kotlin/com/example/project/ExtensionFunctionsTests.kt

marcphilipp: Or does that work because we use the 1.3.61 version of stdlib?
https://github.com/junit-team/junit5/blob/b01802956c05822d59ec73cbf377529a3abfb5f8/platform-tooling-support-tests/projects/gradle-kotlin-extensions/build.gradle.kts#L20

ashdavies: We could probably adjust the implementation to use the same behaviour as `runCatching` without it, though it's a bit more verbose that I had hoped.

```kotlin
inline fun <reified T : Throwable> assertThrows(executable: () -> Unit): T {
  val throwable: Throwable? = try {
    executable()
  } catch (caught: Throwable) {
    caught
  } as? Throwable

  return Assertions.assertThrows(T::class.java) {
    if (throwable != null) {
      throw throwable
    }
  }
}
```

marcphilipp: @ashdavies Could you try to make the implementation test fail?

marcphilipp: Resolves in `master` in c5d658c5aa8e5550410e9691d90e3216fe9a55e4.

ashdavies: Thanks for addressing, hadn't yet had the chance to check

marcphilipp: No worries!

PR #2043, opened 2019-10-06
Log containers and tests which are excluded by post discovery filters

### LINKED ISSUE #1514: Mention containers and tests that were ignored because of their tags
Tagging and filtering works great, but developers could be irritated when their test is not executed (because of a tag) and they don't know why (maybe being new to the code base).

Jupiter could print a message that tests were ignored because their tag was not activated. If these information are easy to gather without performance impact, it could even print some numbers, something like:

```
12 classes and 132 methods are tagged with "integration" and not executed
```

antimpatel:
## Overview
Mention containers and tests that were ignored because of their tags or any other reason.

Current Behavior:
During the execution of post-discovery filters, tests which are excluded does not print any message/reason, which leaves developers to be unaware of the reason for exclusion. Also, No reason was passed from TagFilter class.

Solution:
Stores the Exclusion reason against the set of TestDescriptors and print this information after the discovery phase. Passed the inclusion or exclusion reason to FilterResult for TagFilter.

Closes #1514

---

---

### Definition of Done

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -87,7 +88,20 @@ public final class TagFilter {
 	 * @see TestTag#isValid(String)
 	 */
 	public static PostDiscoveryFilter includeTags(List<String> tagExpressions) throws PreconditionViolationException {
-		return includeMatching(tagExpressions, Stream::anyMatch);
+		Supplier<String> inclusionReason = () -> inclusionReasonExpressionSatisfy(tagExpressions);
+		Supplier<String> exclusionReason = () -> exclusionReasonExpressionNotSatisfy(tagExpressions);
+
+		return includeMatching(tagExpressions, Stream::anyMatch, inclusionReason, exclusionReason);
+	}
+
+	private static String inclusionReasonExpressionSatisfy(List<String> tagExpressions) {
+		return String.format("Test included Because it satisfy expression(s): [%s]", formatToString(tagExpressions));
+	}
+
+	private static String exclusionReasonExpressionNotSatisfy(List<String> tagExpressions) {
+		return String.format("Test excluded Because it does not satisfy expression(s): [%s]",
+			formatToString(tagExpressions));
+
 	}
 
 	/**
@@ -124,24 +138,28 @@ public final class TagFilter {
 	 * @see TestTag#isValid(String)
 	 */
 	public static PostDiscoveryFilter excludeTags(List<String> tagExpressions) throws PreconditionViolationException {
-		return includeMatching(tagExpressions, Stream::noneMatch);
+		Supplier<String> inclusionReason = () -> inclusionReasonExpressionNotSatisfy(tagExpressions);
+		Supplier<String> exclusionReason = () -> exclusionReasonExpressionSatisfy(tagExpressions);
+
+		return includeMatching(tagExpressions, Stream::noneMatch, inclusionReason, exclusionReason);
 	}
 
-	private static String inclusionReasonTagContains(Set<TestTag> tags) {
-		return String.format("Test included Because it contains tags: [%s]", formatToString(tags));
+	private static String inclusionReasonExpressionNotSatisfy(List<String> tagExpressions) {
+		return String.format("Test included Because it does not satisfy expression(s): [%s]",
+			formatToString(tagExpressions));
 	}
 
-	private static String exclusionReasonTagContains(Set<TestTag> tags) {
-		return String.format("Test excluded Because it contains tags: [%s]", formatToString(tags));
-
+	private static String exclusionReasonExpressionSatisfy(List<String> tagExpressions) {
+		return String.format("Test excluded Because it satisfy expression(s): [%s]", formatToString(tagExpressions));
 	}
 
-	private static String formatToString(Set<TestTag> testTags) {
-		return testTags.stream().map(TestTag::getName).collect(Collectors.joining(","));
+	private static String formatToString(List<String> tagExpressions) {
+		return tagExpressions.stream().map(String::trim).collect(Collectors.joining(","));
 	}
 
 	private static PostDiscoveryFilter includeMatching(List<String> tagExpressions,
-			BiPredicate<Stream<TagExpression>, Predicate<TagExpression>> matcher) {
+			BiPredicate<Stream<TagExpression>, Predicate<TagExpression>> matcher, Supplier<String> inclusionReason,
+			Supplier<String> exclusionReason) {
 
 		Preconditions.notEmpty(tagExpressions, "list of tag expressions must not be null or empty");
 		List<TagExpression> parsedTagExpressions = parseAll(tagExpressions);
@@ -149,8 +167,7 @@ public final class TagFilter {
 			Set<TestTag> tags = descriptor.getTags();
 			boolean included = matcher.test(parsedTagExpressions.stream(), expression -> expression.evaluate(tags));
 
-			return FilterResult.includedIf(included, () -> inclusionReasonTagContains(tags),
-				() -> exclusionReasonTagContains(tags));
+			return FilterResult.includedIf(included, inclusionReason, exclusionReason);
 		};
 	}
 
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
@@ -90,11 +90,11 @@ public final class TagFilter {
 	}
 
 	private static String inclusionReasonExpressionSatisfy(List<String> tagExpressions) {
-		return String.format("Test included Because it satisfy expression(s): [%s]", formatToString(tagExpressions));
+		return String.format("included because tags match expression(s): [%s]", formatToString(tagExpressions));
 	}
 
 	private static String exclusionReasonExpressionNotSatisfy(List<String> tagExpressions) {
-		return String.format("Test excluded Because it does not satisfy expression(s): [%s]",
+		return String.format("excluded because tags do not match tag expression(s): [%s]",
 			formatToString(tagExpressions));
 
 	}
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
@@ -18,6 +18,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.apiguardian.api.API;
@@ -126,6 +127,19 @@ public final class TagFilter {
 		return includeMatching(tagExpressions, Stream::noneMatch);
 	}
 
+	private static String inclusionReasonTagContains(Set<TestTag> tags) {
+		return String.format("Test included Because it contains tags: [%s]", formatToString(tags));
+	}
+
+	private static String exclusionReasonTagContains(Set<TestTag> tags) {
+		return String.format("Test excluded Because it contains tags: [%s]", formatToString(tags));
+
+	}
+
+	private static String formatToString(Set<TestTag> testTags) {
+		return testTags.stream().map(TestTag::getName).collect(Collectors.joining(","));
+	}
+
 	private static PostDiscoveryFilter includeMatching(List<String> tagExpressions,
 			BiPredicate<Stream<TagExpression>, Predicate<TagExpression>> matcher) {
 
@@ -133,8 +147,10 @@ public final class TagFilter {
 		List<TagExpression> parsedTagExpressions = parseAll(tagExpressions);
 		return descriptor -> {
 			Set<TestTag> tags = descriptor.getTags();
-			return FilterResult.includedIf(
-				matcher.test(parsedTagExpressions.stream(), expression -> expression.evaluate(tags)));
+			boolean included = matcher.test(parsedTagExpressions.stream(), expression -> expression.evaluate(tags));
+
+			return FilterResult.includedIf(included, () -> inclusionReasonTagContains(tags),
+				() -> exclusionReasonTagContains(tags));
 		};
 	}
 
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
@@ -16,11 +16,8 @@ import static org.junit.platform.commons.util.CollectionUtils.toUnmodifiableList
 
 import java.util.List;
 import java.util.Set;
-import java.util.function.BiPredicate;
-import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.apiguardian.api.API;
 import org.junit.platform.commons.PreconditionViolationException;
@@ -88,10 +85,8 @@ public final class TagFilter {
 	 * @see TestTag#isValid(String)
 	 */
 	public static PostDiscoveryFilter includeTags(List<String> tagExpressions) throws PreconditionViolationException {
-		Supplier<String> inclusionReason = () -> inclusionReasonExpressionSatisfy(tagExpressions);
-		Supplier<String> exclusionReason = () -> exclusionReasonExpressionNotSatisfy(tagExpressions);
 
-		return includeMatching(tagExpressions, Stream::anyMatch, inclusionReason, exclusionReason);
+		return includeMatching(tagExpressions);
 	}
 
 	private static String inclusionReasonExpressionSatisfy(List<String> tagExpressions) {
@@ -138,10 +133,8 @@ public final class TagFilter {
 	 * @see TestTag#isValid(String)
 	 */
 	public static PostDiscoveryFilter excludeTags(List<String> tagExpressions) throws PreconditionViolationException {
-		Supplier<String> inclusionReason = () -> inclusionReasonExpressionNotSatisfy(tagExpressions);
-		Supplier<String> exclusionReason = () -> exclusionReasonExpressionSatisfy(tagExpressions);
 
-		return includeMatching(tagExpressions, Stream::noneMatch, inclusionReason, exclusionReason);
+		return excludeMatching(tagExpressions);
 	}
 
 	private static String inclusionReasonExpressionNotSatisfy(List<String> tagExpressions) {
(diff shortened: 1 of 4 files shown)


ptlanu22:
@marcphilipp can you take a look, ?

antimpatel:
@marcphilipp can you take a look?

marcphilipp:
@antimpatel Sorry for the delay. We're finishing things for 5.6 M1 right now and will get back to this PR once that's released.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -250,4 +252,14 @@ private void execute(TestDescriptor engineDescriptor, EngineExecutionListener li
...
 	}
 
+	/**
+	 * prints the reason for test cases that are excluded in post discovery filter phase.
+	 * @param root of the all engines.
+	 */
+	private void printFilteredReason(Root root) {
+		root.getExclusionReasonToDescriptorMap().forEach((key, value) -> {
+			String displayNames = value.stream().map(TestDescriptor::getDisplayName).collect(Collectors.joining(", "));
+			logger.debug(() -> String.format("Method names [%s] excluded, reason [%s]", displayNames, key));
```
juliette-derancourt: Why this `debug` level? Maybe `info` would make more sense?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
```diff
@@ -33,6 +36,12 @@
 	private final Map<TestEngine, TestDescriptor> testEngineDescriptors = new LinkedHashMap<>(4);
 	private final ConfigurationParameters configurationParameters;
 
+	/**
+	 * Maps to store list of {@link TestDescriptor} which are excluded.
+	 * key: Reason of exclusion.
+	 */
+	private final Map<String, List<TestDescriptor>> exclusionReasonToDescriptorMap = new LinkedHashMap<>();
```
juliette-derancourt: I think that if you rename the map to something like `excludedTestDescriptorsByReason`, you could get rid of the comment. Also, does it really need to be a `LinkedHashMap`?

platform-tests/src/test/java/org/junit/platform/launcher/TagFilterTests.java
```diff
@@ -150,19 +150,49 @@ private void includeSingleTag(PostDiscoveryFilter filter) {
...
 		assertTrue(filter.apply(classWithTag1).excluded());
 		assertTrue(filter.apply(classWithTag1AndSurroundingWhitespace).excluded());
 		assertTrue(filter.apply(classWithBothTags).excluded());
 
+		assertThat(filter.apply(classWithTag1).getReason().get()).contains(
+			"Test excluded Because it contains tags: [tag1]");
+		assertThat(filter.apply(classWithTag1AndSurroundingWhitespace).getReason().get()).contains(
+			"Test excluded Because it contains tags: [tag1]");
+		assertThat(filter.apply(classWithBothTags).getReason().get()).contains(
+			"Test excluded Because it contains tags: [tag1,tag2]");
```
juliette-derancourt: To make that a little bit more readable, you could maybe regroup both exclusion/reason assertions inside a method like `assertExcluded(FilterResult filterResult, String reason)` and call it here like `assertExcluded(filter.apply(...), "reason")`... (and have the `assertIncluded` counterpart for the inclusions)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
```diff
@@ -126,15 +127,30 @@ public static PostDiscoveryFilter excludeTags(List<String> tagExpressions) throw
...
 		Preconditions.notEmpty(tagExpressions, "list of tag expressions must not be null or empty");
 		List<TagExpression> parsedTagExpressions = parseAll(tagExpressions);
 		return descriptor -> {
 			Set<TestTag> tags = descriptor.getTags();
-			return FilterResult.includedIf(
-				matcher.test(parsedTagExpressions.stream(), expression -> expression.evaluate(tags)));
+			boolean included = matcher.test(parsedTagExpressions.stream(), expression -> expression.evaluate(tags));
+
+			return FilterResult.includedIf(included, () -> inclusionReasonTagContains(tags),
+				() -> exclusionReasonTagContains(tags));
```
juliette-derancourt: The main issue I see here is that the provided reason will not always be true, if I'm not mistaken.

Say that we have a class with `tag2` and we only want to include tests with `tag1`. When the filter will be applied to this class, the `included` boolean here will be `false`. It will result in the following exclusion reason: `"Test excluded because it contains tags: [tag2]"`. But shouldn't it actually be `"Test excluded because it does *not* contains tags: [tag1]"`?

The issue is that here, you don't have the original intention to *include* or *exclude* the provided tags, since it's hidden inside the `matcher`...

juliette-derancourt: On another note, do we actually need an inclusion reason?

ptlanu22: @juliette-derancourt I thought of this but was not clear if we require different exclusion reason because of their tags. Although  ``` "Test excluded because it does *not* contains tags: [tag1]"?``` this seems more accurate. I added inclusion reason to maintain symmetry, isn't it relevant to add inclusion reason  as well,

juliette-derancourt: > I thought of this but was not clear if we require different exclusion reason because of their tags. Although "Test excluded because it does *not* contains tags: [tag1]"? this seems more accurate.

I'm not saying we should use this wording or even do this, I'm just saying that the reason you're printing now is false in the case of a "tag inclusion" :)
We actually need to decide if we want to log something if we're in that case (i.e. if some tests are not executed because they don't have the "included tag" on them) or if we're just sticking to the "excluded tag" case. (Don't know if this makes sense outside of my head ahah)

> I added inclusion reason to maintain symmetry, isn't it relevant to add inclusion reason as well?

Well you're not using the inclusion reason in your current implementation, only the exclusion one (see in [here](https://github.com/junit-team/junit5/pull/2043/files#diff-1b699f4d83a93825dbcb40dc72f85278R97)). But the symmetry argument makes sense I guess

ptlanu22: Here What we can do is, we can check whether this method is called from includeTags or excludeTags method. accordingly, we need to populate the message, 

Say that we have a class with tag2 and we only want to include tests with tag1, meaning caller function is **includeTags()**, When the filter will be applied to this class, the included boolean here will be false. Hence we need to give exclusion reason for method, which can be 
like '**Test excluded because the method does not contain any of the tags: [tag1]**'

I believe this makes much more sense since one has to know the methods and reason for which tests have not executed. if we agree then I can push the changes along with other changes.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
```diff
@@ -82,7 +95,13 @@ void prune() {
 	}
 
 	private boolean isExcluded(TestDescriptor descriptor, Filter<TestDescriptor> postDiscoveryFilter) {
-		return descriptor.getChildren().isEmpty() && postDiscoveryFilter.apply(descriptor).excluded();
+		FilterResult filterResult = postDiscoveryFilter.apply(descriptor);
+		boolean excluded = descriptor.getChildren().isEmpty() && filterResult.excluded();
+		if (excluded) {
+			exclusionReasonToDescriptorMap.computeIfAbsent(filterResult.getReason().orElse("Unknown"),
+				list -> new LinkedList<>()).add(descriptor);
+		}
```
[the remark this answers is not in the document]
ptlanu22: Agree, Logic should not hide inside excluded function.
Since Filter result is computed inside `isExcluded` method, what I can do it I can compute `FilterResult`  inside the `applyPostDiscoveryFilters` and pass it to isExcluded in place of passing postDiscoveryFilter. like` isExcluded(TestDescriptor descriptor, FilterResult filterResult) `.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -250,4 +252,14 @@ private void execute(TestDescriptor engineDescriptor, EngineExecutionListener li
 		}
 	}
 
+	/**
+	 * prints the reason for test cases that are excluded in post discovery filter phase.
+	 * @param root of the all engines.
+	 */
+	private void printFilteredReason(Root root) {
```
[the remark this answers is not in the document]
ptlanu22: Correct, the second option makes sense. will change that

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
ptlanu22: Will Rename It, It doesn't matter which map implementation we use, ordered is always better. :)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
juliette-derancourt: @ptlanu22 Sorry to keep you waiting, yes let’s go with the "does not contain" reason if the test is not explicitely included.

I guess you could remove the matcher parameter and make two methods to separate the include/exclude cases ;)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
juliette-derancourt: @ptlanu22 Let’s put two logs here:
- one info log with the number of tests that were not executed, along with the reason
- one debug log with the list of the affected tests (in order to keep the first log concise)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
sbrannen: We are actually handling tag expressions at this point. Thus, we want to state something along the lines of:


```
The following containers and tests were excluded since their tags do not match the tag expression '<expression>': <containers and tests>
```

juliette-derancourt: Well here we are giving the exclusion reason for one test/container, the actual reduce + logging happens in the launcher 👇
Therefore, the wording of the reason does not really matter, right?

juliette-derancourt: Well here we are giving the exclusion reason for one test/container, the actual reduce + logging happens in the launcher 👇
Therefore, the wording of the reason does not really matter, right?

sbrannen: Sorry for getting this out of context.

Yes, the aggregation occurs in the launcher.

In any case, the wording does matter. A test does not "contain" tags. Rather, a test or container can be "assigned" tags, and the test or container is then included or excluded based on whether the assigned tags match the supplied tag expression.

Hope that's clearer.

juliette-derancourt: Yes of course, I’m not saying it _doesn’t_ matter (I know I just said that 😂), but I meant that it will not be the actual thing that will be in the log, it’s just a key for aggregating the reasons :)
But yeah you’re right, contain is not the right word!

antimpatel: @juliette-derancourt / @sbrannen , Here I believe we have to compromise on inclusion/exclusion reasons. so instead of tags, I changed the message to expression, that is still not perfect, but better than a message with Tags.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
```diff
@@ -81,12 +98,22 @@ void prune() {
...
 	}
 
 	private void acceptInAllTestEngines(TestDescriptor.Visitor visitor) {
 		this.testEngineDescriptors.values().forEach(descriptor -> descriptor.accept(visitor));
 	}
 
+	/**
+	 * prints the reason for test cases that are excluded in post discovery filter phase.
+	 */
+	private void printFilteredReason() {
```
juliette-derancourt: Could you add an info log here saying that `x container and y tests were excluded because <reason>`?
The `TestDescriptor` has `isContainer()` and `isTest()` methods you can use.
Also, the reason would have to be changed to something like `tags (do not) match the tag expression <tagExpression>` so that it makes a valid sentence.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
```diff
@@ -81,12 +98,22 @@ void prune() {
...
+		excludedTestDescriptorsByReason.forEach((key, value) -> {
+			String displayNames = value.stream().map(TestDescriptor::getDisplayName).collect(Collectors.joining(", "));
+			logger.debug(() -> String.format("Method names [%s] excluded, reason [%s]", displayNames, key));
```
juliette-derancourt:
```suggestion
			logger.debug(() -> String.format("The following containers and tests were excluded because %s: %s", key, displayNames));
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
```diff
@@ -123,18 +138,36 @@ public static PostDiscoveryFilter excludeTags(String... tagExpressions) throws P
...
+		return String.format("Test included Because it does not satisfy expression(s): [%s]",
+			formatToString(tagExpressions));
+	}
+
+	private static String exclusionReasonExpressionSatisfy(List<String> tagExpressions) {
+		return String.format("Test excluded Because it satisfy expression(s): [%s]", formatToString(tagExpressions));
+	}
+
+	private static String formatToString(List<String> tagExpressions) {
+		return tagExpressions.stream().map(String::trim).collect(Collectors.joining(","));
```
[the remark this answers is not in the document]
antimpatel: I would like to highlight here that its expression that is gonna printed,  
yes sorting should be there.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
```diff
@@ -123,18 +138,36 @@ public static PostDiscoveryFilter excludeTags(String... tagExpressions) throws P
...
 	}
 
 	private static PostDiscoveryFilter includeMatching(List<String> tagExpressions,
```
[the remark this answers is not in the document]
antimpatel: There is already a public caller method for this, common code is already abstracted out. since postDiscoveryFilter only care about the reason for inclusion/exclusion, isn't it good to just pass reasons? that's why I didn't create a new method, to keep matching logic at one place. :)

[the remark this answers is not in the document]
juliette-derancourt: Here the three parameters you're passing (matcher, inclusion and exclusion reasons) will always be the same trio of values, depending on wether the tags are excluded or included, so you may as well make it explicit. Plus, it will be more readable imo. Personally, I think that sometimes readability wins over abstraction ;P

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
```diff
@@ -86,7 +85,18 @@ public static PostDiscoveryFilter includeTags(String... tagExpressions) throws P
 	 * @see TestTag#isValid(String)
 	 */
 	public static PostDiscoveryFilter includeTags(List<String> tagExpressions) throws PreconditionViolationException {
-		return includeMatching(tagExpressions, Stream::anyMatch);
+
+		return includeMatching(tagExpressions);
+	}
```
juliette-derancourt: Ok now I see what you meant in https://github.com/junit-team/junit5/pull/2043#discussion_r357924111... and I think I agree 😅
My bad, I guess I didn't look at it properly...
I mean I still don't really like the way it looked before, but you were right this is no better. So I don't know, if you want to put it back the way it was or if you have another idea, go on 😉 (sorry again)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
```diff
@@ -81,12 +98,24 @@ void prune() {
...
+	private boolean isExcluded(TestDescriptor descriptor, FilterResult filterResult) {
+
+		return descriptor.getChildren().isEmpty() && filterResult.excluded();
 	}
 
 	private void acceptInAllTestEngines(TestDescriptor.Visitor visitor) {
 		this.testEngineDescriptors.values().forEach(descriptor -> descriptor.accept(visitor));
 	}
 
+	private void printReasonforFilteredTest(Map<String, List<TestDescriptor>> excludedTestDescriptorsByReason) {
```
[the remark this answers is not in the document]
juliette-derancourt: Or maybe `logTestDescriptorExclusionReasons`?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
```diff
@@ -81,12 +98,24 @@ void prune() {
...
+		excludedTestDescriptorsByReason.forEach((key, value) -> {
+			String displayNames = value.stream().map(TestDescriptor::getDisplayName).collect(Collectors.joining(", "));
+			long containersCount = value.stream().filter(TestDescriptor::isContainer).count();
+			long methodCount = value.size() - containersCount;
+			logger.info(() -> String.format(" %d containers and %d tests were excluded because: %s", containersCount,
```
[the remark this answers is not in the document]
juliette-derancourt: This would be completed by the exclusion reason suggested in https://github.com/junit-team/junit5/pull/2043#discussion_r362565251

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
ptlanu22: I believe the current code is more readable than the previous one, let's go with this. :P

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/Root.java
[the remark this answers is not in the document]
ptlanu22: yup, missed this. will correct.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TagFilter.java
```diff
@@ -86,7 +85,18 @@ public static PostDiscoveryFilter includeTags(String... tagExpressions) throws P
...
+
+	private static String inclusionReasonExpressionSatisfy(List<String> tagExpressions) {
+		return String.format("Test included Because it satisfy expression(s): [%s]", formatToString(tagExpressions));
+	}
+
+	private static String exclusionReasonExpressionNotSatisfy(List<String> tagExpressions) {
+		return String.format("Test excluded Because it does not satisfy expression(s): [%s]",
```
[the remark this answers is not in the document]
ptlanu22: keeping it like excluded because tags do not match tag expression(s) : makes sense in case of multiple expressions.

ptlanu22:
@juliette-derancourt are we good here?

juliette-derancourt:
Hey @ptlanu22!
Sorry again for the delay on this one ;)
I merged it with ae9b05434fa1d7dfa5b9ff2096189b30ef6e1d6c and edited the wording a bit in 35175b1298b0bfa17fb87d19decf649d7e5ccdc0.
Thanks again for the PR! 😄

PR #2044, opened 2019-10-06
Improve handling of duplicate Descriptions

### LINKED ISSUE #2021: JUnit5-Vintage produces duplicate TestIdentifiers when running JUnit3 suites with duplicate test names
## Steps to reproduce

Reproducer: https://github.com/junit-team/junit5-samples/pull/113

I admit that JUnit3 is kind of old, however https://github.com/apache/jmeter contains quite a few `suite()`-driven tests. It would be nice if Vintage executor could account for `TestSuite` names as well when producing `TestIdentifier`s

## Context

 - Used versions (Jupiter/Vintage/Platform): 5.5.1
 - Build Tool/IDE: Gradle 5.5.1

marcphilipp:
Some runners don't memoize Description instances and even pass different
instances when invoking testStarted and testFinished. This makes it
particularly difficult when there are multiple equal Description
instances in a Runner.

Prior to this commit, such cases were deemed unresolvable which lead to
multiple dynamic TestDescriptors with the same unique ID being reported
which is a violation of the contract of TestExecutionListener.

Now, we assume the testStarted/testSkipped event for a Description with
multiple equal Descriptions is referring to the first Description that
hasn't yet been started or skipped. The corresponding TestDescriptor is
then reported as started/skipped. When testFinished is called for such
a Description, we check whether the same thread has previously reported
an equal Description as started. If so, we report the corresponding
TestDescriptor as finished.

Fixes #2021.

---

---

### Definition of Done

### DIFF
--- a/junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineExecutionTests.java
+++ b/junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineExecutionTests.java
@@ -42,14 +42,15 @@ import org.junit.platform.engine.UniqueId;
 import org.junit.platform.engine.reporting.ReportEntry;
 import org.junit.platform.launcher.LauncherDiscoveryRequest;
 import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
+import org.junit.platform.testkit.engine.EngineExecutionResults;
 import org.junit.platform.testkit.engine.EngineTestKit;
-import org.junit.platform.testkit.engine.Events;
 import org.junit.runner.Description;
 import org.junit.runner.RunWith;
 import org.junit.runner.Runner;
 import org.junit.runner.notification.RunNotifier;
 import org.junit.runners.Suite;
 import org.junit.runners.Suite.SuiteClasses;
+import org.junit.vintage.engine.samples.junit3.JUnit3ParallelSuiteWithSubsuites;
 import org.junit.vintage.engine.samples.junit3.PlainJUnit3TestCaseWithSingleTestWhichFails;
 import org.junit.vintage.engine.samples.junit4.CompletelyDynamicTestCase;
 import org.junit.vintage.engine.samples.junit4.EmptyIgnoredTestCase;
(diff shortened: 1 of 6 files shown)


marcphilipp:
@vlsi I would appreciate it if you could take a look. 🙂

junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineExecutionTests.java
```diff
@@ -683,12 +685,56 @@ void executesCompletelyDynamicTestCaseDiscoveredByUniqueId() {
...
+	void executesJUnit3ParallelSuiteWithSubsuites() {
+		var suiteClass = JUnit3ParallelSuiteWithSubsuites.class;
+		var results = execute(suiteClass);
+		results.containerEvents() //
+				.assertStatistics(stats -> stats.started(4).dynamicallyRegistered(0).finished(4).succeeded(4)) //
+				.assertEventsMatchExactly( //
+					event(engine(), started()), //
+					event(container(suiteClass), started()), //
+					event(container("Case"), started()), //
+					event(container("Case")), //
```
vlsi: Can this be Case1/Case2?
Are the events distinguishable?

vlsi:
Ah, I don't completely follow the logic here, however, I wonder if the resulting test events are distinguishable. In other words, are Case1/Case2 distinguishable?

PS. Can "beta" version of JUnit be used as Gradle's `includeBuild`?
If that is possible, I can try that with the current JMeter.
However, I have rewritten the tests that used ActiveSuite and duplicate test names.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineExecutionTests.java
marcphilipp: Since they run concurrently the order isn't reliable.

marcphilipp:
> PS. Can "beta" version of JUnit be used as Gradle's `includeBuild`?

I've never tried that but I don't see why it shouldn't work.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/VintageTestEngineExecutionTests.java
vlsi: Does it make sense to add a second test where regular `TestSuite` is used instead of `ActiveSuite`?
It should reproduce the same issue, however, the order would be reliable.

PR #2060, opened 2019-10-12
Introduce discovery listeners and abort discovery on failures by default

### LINKED ISSUE #2052: Improve error reporting for exceptions thrown by SelectorResolvers in EngineDiscoveryRequestResolution
Exceptions thrown by custom `SelectorResolvers` for test discovery are not propagated back to the `TestEngine` when using the `EngineDiscoveryRequestResolver`, therefore also not the `Launcher` and consequently also not the user. Currently if discovery in a `SelectorResolver` fails, JUnit only reports that no tests were found.

I am building a custom `TestEngine`, because I have some domain-specific needs, that Jupiter does not fulfill. To reduce the workload for coding the discovery phase of the engine I started using the `EngineDiscoveryRequestResolver` suggested in #1739 and introduced with #1798.
If Tests are malformed, i. e. my `SelectorResolver` checks whether the Tests have been specified correctly, discovery in my engine should outright fail by throwing an exception. Suggested in #1909 and introduced with #2006 exceptions thrown by the `TestEngine` are reported to the `TestExecutionListeners` and potentially provide the user with feedback why the discovery has failed. Unfortunately that is not the case for the `EngineDiscoveryRequestResolver` respectively the `EngineDiscoveryRequestResolution`. As far as I can tell the class just [catches all `Throwables`](https://github.com/junit-team/junit5/blob/d6c023e10349109d7e57d642b4191e2ba5e156fd/junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/EngineDiscoveryRequestResolution.java#L95) and simply logs them internally. The user never gets to see the reason why the Test Discovery failed or rather why no tests could be found.

Might it be possible to let the `Exceptions` caught in the `EngineDiscoveryRequestResolver` respectively the `EngineDiscoveryRequestResolution` to not be caught in that class, but to let them travel up the hierarchy to the `TestEngine` to be then caught by the `Launcher` and as a result to be propagated back to the user?
Or is there any other way to communicate the reason for failed test discovery back to the user?

## Deliverables

- [ ] Propagate exceptions thrown by `SelectorResolvers` to the `TestEngine`

marcphilipp:
## Overview

This PR adds the ability to add `LauncherDiscoveryListeners` to `LauncherDiscoveryRequests`. Such listeners are notified when engines resolve or fail to resolve a `DiscoverySelector` and can decide how to handle this. The default listener (up for debate!), currently only logs messages. The new `abortOnFailure()` listener, however, throws exceptions thereby effectively aborting test discovery.

Resolves #2052.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
@@ -10,6 +10,7 @@
 
 package org.junit.platform.engine;
 
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.STABLE;
 
 import java.util.List;
@@ -44,28 +45,42 @@ public interface EngineDiscoveryRequest {
 	 * filtered by a particular type.
 	 *
 	 * @param selectorType the type of {@link DiscoverySelector} to filter by
-	 * @return all selectors of this request that are instances of {@code selectorType}
+	 * @return all selectors of this request that are instances of
+	 * {@code selectorType}; never {@code null} but potentially empty
 	 */
 	<T extends DiscoverySelector> List<T> getSelectorsByType(Class<T> selectorType);
 
 	/**
-	 * Get the {@link DiscoveryFilter DiscoveryFilters} for this request, filtered
-	 * by a particular type.
+	 * Get the {@link DiscoveryFilter DiscoveryFilters} for this request,
+	 * filtered by a particular type.
 	 *
-	 * <p>The returned filters are to be combined using AND semantics, i.e. all of
-	 * them have to include a resource for it to end up in the test plan.
+	 * <p>The returned filters are to be combined using AND semantics, i.e. all
+	 * of them have to include a resource for it to end up in the test plan.
 	 *
 	 * @param filterType the type of {@link DiscoveryFilter} to filter by
-	 * @return all filters of this request that are instances of {@code filterType}
+	 * @return all filters of this request that are instances of
+	 * {@code filterType}; never {@code null} but potentially empty
 	 */
 	<T extends DiscoveryFilter<?>> List<T> getFiltersByType(Class<T> filterType);
 
 	/**
 	 * Get the {@link ConfigurationParameters} for this request.
+	 *
+	 * @return the configuration parameters; never {@code null}
 	 */
 	ConfigurationParameters getConfigurationParameters();
 
-	@API(status = API.Status.EXPERIMENTAL, since = "1.6")
+	/**
+	 * Get the {@link EngineDiscoveryListener} for this request.
+	 *
+	 * <p>The default implementation returns a no-op listener that ignores all
+	 * calls so that engines that call this methods can be used with an earlier
+	 * version of the JUnit Platform that did not yet include it.
+	 *
+	 * @return the discovery listener; never {@code null}
+	 * @since 1.6
+	 */
+	@API(status = EXPERIMENTAL, since = "1.6")
 	default EngineDiscoveryListener getDiscoveryListener() {
 		return EngineDiscoveryListener.NOOP;
 	}
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
@@ -65,4 +65,6 @@ public interface EngineDiscoveryRequest {
 	 */
 	ConfigurationParameters getConfigurationParameters();
 
+	EngineDiscoveryListener getDiscoveryListener();
+
 }
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
@@ -65,6 +65,9 @@ public interface EngineDiscoveryRequest {
 	 */
 	ConfigurationParameters getConfigurationParameters();
 
-	EngineDiscoveryListener getDiscoveryListener();
+	@API(status = API.Status.EXPERIMENTAL, since = "1.6")
+	default EngineDiscoveryListener getDiscoveryListener() {
+		return EngineDiscoveryListener.NOOP;
+	}
 
 }
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
@@ -75,7 +75,7 @@ public interface EngineDiscoveryRequest {
 	 *
 	 * <p>The default implementation returns a no-op listener that ignores all
 	 * calls so that engines that call this methods can be used with an earlier
-	 * version of the JUnit Platform that did not yet include it.
+	 * version of the JUnit Platform that did not yet include this API.
 	 *
 	 * @return the discovery listener; never {@code null}
 	 * @since 1.6
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/SelectorResolutionResult.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/SelectorResolutionResult.java
@@ -12,8 +12,10 @@ package org.junit.platform.engine;
 
 import java.util.Optional;
 
+import org.apiguardian.api.API;
 import org.junit.platform.commons.util.ToStringBuilder;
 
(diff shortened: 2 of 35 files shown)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/LauncherDiscoveryListener.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.junit.platform.engine.EngineDiscoveryListener;
+import org.junit.platform.engine.UniqueId;
+
+public interface LauncherDiscoveryListener extends EngineDiscoveryListener {
+
+	default void engineDiscoveryStarted(UniqueId engineId) {
+	}
+
+	default void engineDiscoveryFinished(UniqueId engineId, Optional<Throwable> failure) {
```
sormuras: No `Optional` in parameter types.

Either allow `null` as an indicator for no failure or provide two finished callbacks:
- `default void engineDiscoveryFinished(UniqueId engineId) {}`
- `default void engineDiscoveryFinishedWithFailure(UniqueId engineId, Throwable failure) {}`

Or wrap it in a `EngineDiscoveryFinishedResult` holder that has an `Optional<Throwable> failure()` accessor.

marcphilipp: > No `Optional` in parameter types.

We don't always follow that "rule" here. Nevertheless, you have a point and I think it would be more consistent to use a result object.

junit-platform-engine/src/main/java/org/junit/platform/engine/SelectorResolutionResult.java
```diff
@@ -0,0 +1,63 @@
...
+	public SelectorResolutionResult(Status status, Throwable throwable) {
+		this.status = status;
+		this.throwable = throwable;
+	}
+
+	public Status getStatus() {
+		return status;
+	}
+
+	public Throwable getThrowable() {
```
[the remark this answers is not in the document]
marcphilipp: Good catch! 👍

marcphilipp:
> Is there then no real means for the end-user, who is going to write the tests, to change the `EngineDiscoveryListener`?

I built it this way so it's backwards compatible. I think it would be surprising for build tools and IDEs if the Launcher started throwing exceptions all of a sudden. However, I do see your point. We could determine the default `EngineDiscoveryListener` from a config parameter that you could, for example, set via a system property. 🤔 

> And is there no means for the developer, who is writing a `TestEngine` to swap out the `EngineDiscoveryListener`?

I don't think it's the responsibility of the engine to decide how to deal with unresolvable selectors since this should be handled consistently for all currently present engines. If a `TestEngine` runs into an absolutely fatal error, it can always throw an exception.

lbernwald:
> > Is there then no real means for the end-user, who is going to write the tests, to change the `EngineDiscoveryListener`?
> 
> I built it this way so it's backwards compatible. I think it would be surprising for build tools and IDEs if the Launcher started throwing exceptions all of a sudden. However, I do see your point. We could determine the default `EngineDiscoveryListener` from a config parameter that you could, for example, set via a system property. 🤔
> 
Yes, I agree. I think setting `EngineDiscoveryListener` from a config parameter would be a viable option. I don't know if I understood you correctly, but wouldn't throwing an exception in an `EngineDiscoveryListener` just result in an exception thrown on the `Engine` level and the `Launcher` responsible for handling said exception. The `Launcher` wouldn't necessarily throw an exception then. The `DefaultLauncher` catches all Throwables during discovery anyway, right?

> > And is there no means for the developer, who is writing a `TestEngine` to swap out the `EngineDiscoveryListener`?
> 
> I don't think it's the responsibility of the engine to decide how to deal with unresolvable selectors since this should be handled consistently for all currently present engines. If a `TestEngine` runs into an absolutely fatal error, it can always throw an exception.

Yes, you're right. I had kind of a thought error there. Because the `TestEngine` is handling the discovery, it can already decide what to do if a test is malformed and potentially throw an exception just like you said. The handling of said exceptions should then definitely be handled by the `Launcher`.

marcphilipp:
> I don't know if I understood you correctly, but wouldn't throwing an exception in an `EngineDiscoveryListener` just result in an exception thrown on the `Engine` level and the `Launcher` responsible for handling said exception. The `Launcher` wouldn't necessarily throw an exception then. The `DefaultLauncher` catches all Throwables during discovery anyway, right?

Yes, the `DefaultLauncher` catches `Throwables` thrown by an engine during discovery. With the change in this PR that includes exceptions thrown by any `EngineDiscoveryListener`. The exception will then be passed to all `LauncherDiscoveryListeners` wrapped in an `EngineDiscoveryResult`. The `AbortOnFailureLauncherDiscoveryListener` rethrows the passed the exception which is propagated by the `Launcher` to its caller.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/discovery/LauncherDiscoveryListeners.java
```diff
@@ -30,12 +38,50 @@ public static LauncherDiscoveryListener logging() {
...
+					.orElseThrow(() -> {
+						String allowedValues = Arrays.stream(LauncherDiscoveryListenerType.values()) //
+								.map(type -> type.parameterValue) //
+								.collect(joining("', '", "'", "'"));
+						return new JUnitException("Invalid value of configuration parameter '"
+								+ DEFAULT_DISCOVERY_LISTENER_CONFIGURATION_PROPERTY_NAME + "': " //
+								+ defaultListenerType + " (allowed are " + allowedValues + ")");
+					})) //
+				.orElseGet(LauncherDiscoveryListeners::abortOnFailure);
+	}
```
mmerdes: A rather deeply nested indentation - Did you consider a temp variable or helper method to reduce indentation?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherDiscoveryRequestBuilder.java
```diff
@@ -204,15 +205,23 @@ else if (filter instanceof DiscoveryFilter<?>) {
...
 		return new DefaultDiscoveryRequest(this.selectors, this.engineFilters, this.discoveryFilters,
-			this.postDiscoveryFilters, launcherConfigurationParameters, getLauncherDiscoveryListener());
+			this.postDiscoveryFilters, launcherConfigurationParameters, discoveryListener);
 	}
 
-	private LauncherDiscoveryListener getLauncherDiscoveryListener() {
-		if (this.discoveryListeners.isEmpty()) {
-			return LauncherDiscoveryListeners.logging();
+	private LauncherDiscoveryListener getLauncherDiscoveryListener(ConfigurationParameters configurationParameters) {
+		LauncherDiscoveryListener defaultDiscoveryListener = LauncherDiscoveryListeners.fromConfigurationParameter(
```
mmerdes: Not sure about the naming - would `configuredDiscoveryListener ` or `configuredDefaultDiscoveryListener ` convey the intent more clearly?

junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
```diff
@@ -44,28 +45,42 @@
...
 	 */
 	ConfigurationParameters getConfigurationParameters();
 
-	@API(status = API.Status.EXPERIMENTAL, since = "1.6")
+	/**
+	 * Get the {@link EngineDiscoveryListener} for this request.
+	 *
+	 * <p>The default implementation returns a no-op listener that ignores all
+	 * calls so that engines that call this methods can be used with an earlier
+	 * version of the JUnit Platform that did not yet include it.
```
mmerdes: 'this API' instead of 'it'?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/discovery/LauncherDiscoveryListeners.java
marcphilipp: I didn't look quite as bad in an (my) IDE but I extracted a method nevertheless.

junit-platform-engine/src/main/java/org/junit/platform/engine/EngineDiscoveryRequest.java
marcphilipp: Done.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherDiscoveryRequestBuilder.java
marcphilipp: TBH I think it's clear enough. After all it mimics the new configuration parameter.

PR #2068, opened 2019-10-16
Publish correct Gradle Module Metadata with Gradle 6

jjohannes:
## Overview

With Gradle 6, `maven-publish` automatically published Gradle Module Metadata files (`.module`) in addition to POM metadata files. With this, more information, like additional dependency version constraints, can be provided to Gradle builds consuming JUnit 5 modules. This PR does the following to publish correct and richer Gradle Module Metadata with Gradle 6:
- Upgrade the build to Gradle 6 RC1
- Set the JvmTargetVersion for all modules to `8`
- Add a _platform dependency_ from each module to the JUnit BOM. This allows Gradle to **always** align the versions if multiple modules are in a dependency graph.
- Changes the Javadoc and sources publishing configuration to use the new built-in support of Gradle 

In addition this PR fixes a deprecation warnings and an issue with inner classes not correctly declared in a Kotlin DSL script.

An example of how the alignment works can be found here: https://github.com/jjohannes/what-if-gradle-metadata/tree/master/usecases/junit5-align-versions

---

---

### Definition of Done

- **n/a** Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- **n/a**  Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- **n/a**  Change is documented in the [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)


platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/GradleModuleFileTests.java
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * @since 1.6
+ */
+class GradleModuleFileTests {
+
+	@Test
+	void jupiterAggregatorGradleModuleMetadataVariants() throws Exception {
+		var expected = List.of(">> HEAD >>", //
+			"{", //
```
sormuras:
```
"""
Text Blocks, where art thou?
"""
```

sormuras:
Seems like this integration test project needs some Gradle 6 ❤️
https://github.com/junit-team/junit5/tree/master/platform-tooling-support-tests/projects/gradle-kotlin-extensions

marcphilipp:
I have a fix for the failing integration test locally. I'll push it as soon as @jjohannes gives me access to the branch. 😉

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
```diff
@@ -71,6 +81,11 @@ if (project in mavenizedProjects) {
 
 	apply(plugin = "publishing-conventions")
 
+	java {
+		publishJavadoc()
+		publishSources()
+	}
```
marcphilipp: 🎉

jjohannes:
Thanks @marcphilipp for fixing the test issues. I gave you access to my fork.

Regarding signing of GMM: It is not working for `SNAPSHOT` versions and that will not be fixed. The problem is that we cannot fix it unless we re-design the way signing is working in Gradle today. Since a `module` file refers each artifact/file directly by its name, we need to know the concrete file name. However, for snapshots, that can only be decided at publishing time when negotiating it with the repository as it depends on the previously published snapshots. Signing however happens earlier in a separate Gradle task. So if we modify GMM during publishing, which we have to do for snapshots, we would have to re-run signing (of that file) afterwards somehow. That would be a larger change.

For non-snapshot versions this is not an issue and signing is working fine.

It seems to be common practice **not** to sign snapshots. JUnit 5 is also not doing it as far as I can tell. Therefore we think this is an acceptable limitation for now. Do you think it will be an issue?

marcphilipp:
> Do you think it will be an issue?

I think some people (like me) will have signing configured using the default project properties locally and run into this issue. But they can work around as I just did in the latest commit. Is there an issue in gradle/gradle to track this?

marcphilipp:
Thanks Jendrik, much appreciated! 👍

PR #2070, opened 2019-10-18
Run Gradle with --parallel in GitHub Actions

marcphilipp:
Trying to speed-up GitHub Actions builds.


sbrannen:
Why don't we just add `org.gradle.parallel=true` to `gradle.properties`?

I was going to do that (or at least suggest it) anyway.

Issue #1793, opened 2019-02-24
Support repeatable annotations for built-in conditions


rocketraman:
Junit should support multiple condition annotations on classes and tests. One use case for this is a test that should not run in two situations: the first is that it should never run in the CI environment (CI=true), and the second is if a dev sets another environment variable (SKIP_X=true). One approach is to use the `@DisabledIfEnvironmentVariable` twice.

The only solution currently is to do `@DisabledIfEnvironmentVariable(named = "SKIP_X", matches = "true")`, and then set `SKIP_X` in the CI environment. However, with this approach we lose some self-documentation -- the explicit `CI=true` check documents the fact that this test doesn't run currently in the CI environment, but may in the future, and is an easy way to find all such cases. It is also far less maintainable, as each such case requires modifying the CI configuration settings.

## Deliverables

- [x] Support multiple condition annotations on classes and tests

sbrannen:
Tentatively slated for 5.5 M1 for _team discussion_

sbrannen:
Proof of concept: _in progress_

sbrannen:
The proof of concept can be seen here:

https://github.com/junit-team/junit5/compare/issues/1793-multiple-condition-annotations

marcphilipp:
**Team Decision:** use new repeatable annotation lookup for execution condition annotations

to be determined later: use new repeatable annotation lookup for registering multiple `@ArgumentSource` annotations of the same type for parameterized tests.

sbrannen:
**Update**: the latest work on this can be seen in the following branch for anyone interested.

https://github.com/junit-team/junit5/commits/issues/1793-repeatable-conditions

sbrannen:
@junit-team/junit-lambda, I have added the _team discussion_ label in order to discuss the semantics we want for the proposed feature.

Some topics to consider:

- Repeated annotations like `@EnabledIfEnvironmentVariable` and `@EnabledIfSystemProperty` could be supported with `AND` or `OR` semantics but not both simultaneously (at least not with the current approach). Developers could naturally desire one or the other: the "best" semantics are therefore debatable.
- We currently support `OR` semantics for annotations like `@EnabledOnOs` since we accept multiple OSes via an array. Thus, we may choose to simply align with that existing feature and apply `OR` semantics to repeated annotations for environment variables and system properties; however, doing so would break from the traditional semantics of individually processed conditions.
[...]

PR #2094, opened 2019-11-08
Support repeatable annotations for built-in conditions

sbrannen:
Prior to this commit, built-in conditions could not be declared
multiple times for a given annotated element (i.e., class or method).
This prevents users from specifying multiple system properties or
environment variables for which an element should be enabled or
disabled.

This commit introduces repeatable annotation support for
@EnabledIfEnvironmentVariable, @DisabledIfEnvironmentVariable,
@EnabledIfSystemProperty, and @DisabledIfSystemProperty.

Closes: #1793

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/EnabledIfEnvironmentVariableCondition.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/EnabledIfEnvironmentVariableCondition.java
@@ -13,13 +13,9 @@ package org.junit.jupiter.api.condition;
 import static java.lang.String.format;
 import static org.junit.jupiter.api.extension.ConditionEvaluationResult.disabled;
 import static org.junit.jupiter.api.extension.ConditionEvaluationResult.enabled;
-import static org.junit.platform.commons.util.AnnotationUtils.findAnnotation;
-
-import java.util.Optional;
 
 import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
-import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.platform.commons.util.Preconditions;
 
 /**
@@ -28,21 +24,24 @@ import org.junit.platform.commons.util.Preconditions;
  * @since 5.1
  * @see EnabledIfEnvironmentVariable
  */
-class EnabledIfEnvironmentVariableCondition implements ExecutionCondition {
+class EnabledIfEnvironmentVariableCondition
+		extends AbstractRepeatableAnnotationCondition<EnabledIfEnvironmentVariable> {
 
-	private static final ConditionEvaluationResult ENABLED_BY_DEFAULT = enabled(
-		"@EnabledIfEnvironmentVariable is not present");
+	private static final ConditionEvaluationResult ENABLED = ConditionEvaluationResult.enabled(
+		"No @EnabledIfEnvironmentVariable conditions resulting in 'disabled' execution encountered");
+
+	EnabledIfEnvironmentVariableCondition() {
+		super(EnabledIfEnvironmentVariable.class);
+	}
 
 	@Override
-	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
-		Optional<EnabledIfEnvironmentVariable> optional = findAnnotation(context.getElement(),
-			EnabledIfEnvironmentVariable.class);
+	protected ConditionEvaluationResult getNoDisabledConditionsEncounteredResult() {
+		return ENABLED;
+	}
 
-		if (!optional.isPresent()) {
-			return ENABLED_BY_DEFAULT;
-		}
+	@Override
+	protected ConditionEvaluationResult evaluate(EnabledIfEnvironmentVariable annotation) {
 
-		EnabledIfEnvironmentVariable annotation = optional.get();
 		String name = annotation.named().trim();
 		String regex = annotation.matches();
 		Preconditions.notBlank(name, () -> "The 'named' attribute must not be blank in " + annotation);
--- a/junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyConditionTests.java
+++ b/junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyConditionTests.java
@@ -40,13 +40,13 @@ class DisabledIfSystemPropertyConditionTests extends AbstractExecutionConditionT
 	}
 
 	@BeforeAll
-	static void setSystemProperty() {
-		DisabledIfSystemPropertyIntegrationTests.setSystemProperty();
+	static void setSystemProperties() {
+		DisabledIfSystemPropertyIntegrationTests.setSystemProperties();
 	}
 
 	@AfterAll
-	static void clearSystemProperty() {
-		DisabledIfSystemPropertyIntegrationTests.clearSystemProperty();
+	static void clearSystemProperties() {
+		DisabledIfSystemPropertyIntegrationTests.clearSystemProperties();
 	}
 
 	/**
@@ -56,7 +56,7 @@ class DisabledIfSystemPropertyConditionTests extends AbstractExecutionConditionT
 	void enabledBecauseAnnotationIsNotPresent() {
 		evaluateCondition();
 		assertEnabled();
-		assertReasonContains("@DisabledIfSystemProperty is not present");
+		assertReasonContains("No @DisabledIfSystemProperty conditions resulting in 'disabled' execution encountered");
 	}
 
 	/**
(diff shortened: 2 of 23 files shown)


sbrannen:
@junit-team/junit-lambda, this PR is ready for review.

I'll update the User Guide and Release Notes once people have had a chance to review the implementation and tests.

sbrannen:
> I'll update the User Guide and Release Notes ...

That's done now as well.

So, if there aren't any objections, I'll merge this into `master`.

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyIntegrationTests.java
```diff
@@ -26,53 +31,66 @@
...
-	static void clearSystemProperty() {
-		System.clearProperty(KEY);
+	static void clearSystemProperties() {
+		System.clearProperty(KEY1);
+		System.clearProperty(KEY2);
 	}
 
 	@Test
 	@Disabled("Only used in a unit test via reflection")
 	void enabledBecauseAnnotationIsNotPresent() {
```
juliette-derancourt: Should be `disabledBecauseAnnotationIsNotPresent` right?

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyConditionTests.java
```diff
@@ -40,13 +40,13 @@ protected ExecutionCondition getExecutionCondition() {
 	}
 
 	@BeforeAll
-	static void setSystemProperty() {
-		DisabledIfSystemPropertyIntegrationTests.setSystemProperty();
```
juliette-derancourt: Is it a good idea to call another `@BeforeAll` method like that? (I genuinely don't know, but I find it a bit weird)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/EnabledIfEnvironmentVariableCondition.java
```diff
@@ -28,21 +24,24 @@
  * @since 5.1
  * @see EnabledIfEnvironmentVariable
  */
-class EnabledIfEnvironmentVariableCondition implements ExecutionCondition {
+class EnabledIfEnvironmentVariableCondition
+		extends AbstractRepeatableAnnotationCondition<EnabledIfEnvironmentVariable> {
+
+	private static final ConditionEvaluationResult ENABLED = ConditionEvaluationResult.enabled(
+		"No 'disabled' @EnabledIfEnvironmentVariable conditions encountered");
```
juliette-derancourt: I see what you meant here, but the `disabled @Enabled...` may be a bit confusing, no? Have no better idea though...

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/EnabledIfSystemPropertyIntegrationTests.java
```diff
@@ -25,59 +30,84 @@
  */
 class EnabledIfSystemPropertyIntegrationTests {
 
-	private static final String KEY = "EnabledIfSystemPropertyTests.key";
+	private static final String KEY1 = "EnabledIfSystemPropertyTests.key1";
+	private static final String KEY2 = "EnabledIfSystemPropertyTests.key2";
 	private static final String ENIGMA = "EnabledIfSystemPropertyTests.enigma";
```
juliette-derancourt: You removed the prefix everywhere but here, don't know if it's intentional?

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/EnabledIfSystemPropertyIntegrationTests.java
```diff
@@ -25,59 +30,84 @@
...
 	private static final String BOGUS = "EnabledIfSystemPropertyTests.bogus";
 
 	@BeforeAll
 	static void setSystemProperty() {
```
juliette-derancourt: Maybe `setSystemProperties` like in the other one?

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyConditionTests.java
juliette-derancourt: Maybe there could be an `AbstractSystemPropertyTests` or something like that to handle these shared set-ups? The one in `EnabledIfSystemPropertyIntegrationTests` is also the same

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyIntegrationTests.java
sbrannen: The name is actually correct, but... I fully understand how that's confusing. 😉 

It should be "enabled" because `@DisabledIfSystemProperty` is not present on the test method. The corresponding unit test `org.junit.jupiter.api.condition.DisabledIfSystemPropertyConditionTests.enabledBecauseAnnotationIsNotPresent()` verifies that. Note that the unit test ignores the presence of `@Disabled`.

Does that make sense now?

sbrannen: Having said that, the `@Disabled` declaration and use of `fail()` are unnecessary for that test method, so I'll remove them. 👍

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyConditionTests.java
sbrannen: > Is it a good idea to call another @BeforeAll method like that? (I genuinely don't know, but I find it a bit weird)

It certainly is a bit strange. I admit.

But I think it's OK for this particular scenario, since the exact same set of system properties are required to make the integration tests and unit tests pass.

sbrannen: > The one in EnabledIfSystemPropertyIntegrationTests is also the same.

It's similar but not exactly the same. The two classes uses different prefixes for the keys (i.e., related to the scope of each test class). I did that when these tests were originally introduced in order to avoid possible conflicts (e.g., during potential parallel execution of the test classes). Perhaps that's over-engineering... but it doesn't bother me having it stay like that.

If the rest of the team feels differently about that, we can change it though.

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/EnabledIfSystemPropertyIntegrationTests.java
sbrannen: Good catch. That was not intentional. I'll fix it.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/EnabledIfEnvironmentVariableCondition.java
sbrannen: Indeed. That can certainly be confusing.

The thing is... every `ExecutionCondition` either returns `enabled()` or `disabled()`, and we only care if `disabled()` is the result. 

I guess we could change it to something like `No @EnabledIfEnvironmentVariable conditions resulting in 'disabled' execution encountered`.

Is that better?

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/EnabledIfSystemPropertyIntegrationTests.java
sbrannen: Good catch. I'll change that, too.

sbrannen:
Thanks for the review, @juliette-derancourt.

I made changes, rebased on master, and force pushed.

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyIntegrationTests.java
sbrannen: I think my comment got lost, but... I removed the `@Disabled` annotation and use of `fail()` in that method since they didn't make sense.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/EnabledIfEnvironmentVariableCondition.java
juliette-derancourt: Yes, I find it more understandable indeed!

junit-jupiter-engine/src/test/java/org/junit/jupiter/api/condition/DisabledIfSystemPropertyIntegrationTests.java
juliette-derancourt: Ah, I see.
This whole enabled/disabled thing everywhere is pretty disorienting ahah 😅

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/EnabledIfEnvironmentVariableCondition.java
sbrannen: Done

PR #2095, opened 2019-11-08
Provide JARs with OSGi metadata

### LINKED ISSUE #195: Provide JARs with OSGi metadata
The JARs that are produced during a build should contain OSGi metadata. This would allow to consume them as OSGi bundles right away.

rotty3000:
fixes  #195

Signed-off-by: Raymond Augé <raymond.auge@liferay.com>

## Overview

Provide JARs with OSGi metadata

---

---

### Definition of Done

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -10,4 +10,5 @@ repositories {
 dependencies {
 	implementation(kotlin("gradle-plugin"))
 	implementation("de.marcphilipp.gradle:nexus-publish-plugin:0.4.0")
+	implementation("biz.aQute.bnd:biz.aQute.bnd.gradle:4.3.1")
 }
--- a/buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
@@ -1,3 +1,7 @@
+import aQute.bnd.gradle.BundleTaskConvention
+import aQute.bnd.gradle.FileSetRepositoryConvention
+import aQute.bnd.gradle.Resolve
+
 plugins {
 	`java-library`
 	eclipse
@@ -136,6 +140,105 @@ if (project in mavenizedProjects) {
 					description.set(provider { "Module \"${project.name}\" of JUnit 5." })
 				}
 			}
+		}
+	}
+
+	// This task enances `jar` and `shadowJar` tasks with the bnd
+	// `BundleTaskConvention` convention which allows for generating OSGi
+	// metadata into the jar
+	tasks.withType<Jar>().matching {
+		task: Jar -> task.name == "jar" || task.name == "shadowJar"
+	}.configureEach {
+		val btc = BundleTaskConvention(this)
+
+		// These are bnd instructions necessary for generating OSGi metadata.
+		// We've generalized these so that they are widely applicable limiting
+		// module configurations to special cases.
+		btc.setBnd("""
+			# These are the general rules for package imports.
+			Import-Package: \
+				!org.apiguardian.api,\
+				org.junit.platform.commons.logging;status=INTERNAL,\
+				kotlin.*;resolution:="optional",\
+				*
+
+			# This tells bnd not to complain if a module doesn't actually import
+			# the kotlin packages, but enough modules do to make it a default.
+			-fixupmessages.kotlin.import: "Unused Import-Package instructions: \\[kotlin.*\\]";is:=ignore
+
+			# This tells bnd to ignore classes it files in `META-INF/versions/`
+			# because bnd doesn't yet support multi-release jars.
+			-fixupmessages.wrong.dir: "Classes found in the wrong directory: \\{META-INF/versions/...";is:=ignore
+
+			# Don't scan for Class.forName package imports.
+			# See https://bnd.bndtools.org/instructions/noclassforname.html
+			-noclassforname: true
+
+			# Don't add all the extra headers bnd normally adds.
+			# See https://bnd.bndtools.org/instructions/noextraheaders.html
+			-noextraheaders: true
+
+			# Don't add the Private-Package header.
+			# See https://bnd.bndtools.org/instructions/removeheaders.html
+			-removeheaders: Private-Package
+
+			# Add the custom buildSrc/src/main/kotlin/APIGuardianAnnotations.kt
+			# plugin to bnd
+			-plugin.apiguardian.annotations: ${APIGuardianAnnotations::class.qualifiedName}
+
+			# Instruct the APIGuardianAnnotations how to operate.
+			# See https://bnd.bndtools.org/instructions/export-apiguardian.html
+			-export-apiguardian: *;version=${project.version}
+		""")
+
+		// Add the convention to the jar task
+		convention.plugins.put("bundle", btc)
+
+		doLast {
+			// Do the actual work putting OSGi stuff in the jar.
+			btc.buildBundle()
+		}
+
+		finalizedBy("verifyOSGi")
+	}
+
+	// Bnd's Resolve task uses a properties file for it's configuration. This
+	// task writes out the properties necessary for it to verify the OSGi
+	// metadata.
+	tasks.register<WriteProperties>("verifyOSGiProperties") {
+		setOutputFile("${buildDir}/verifyOSGiProperties.bndrun")
+		property("-standalone", "true")
+		property("-runee", "JavaSE-${Versions.jvmTarget}")
+		property("-runrequires", "osgi.identity;filter:='(osgi.identity=${project.name})'")
+		property("-runsystempackages", "jdk.internal.misc,sun.misc")
+	}
+
+	// Bnd's Resolve task is what verifies that a jar can be used in OSGi and
+	// that it's metadata is valid. If the metadata is invalid this task will
+	// fail.
+	tasks.register<Resolve>("verifyOSGi") {
+		dependsOn("verifyOSGiProperties")
+		setBndrun("${buildDir}/verifyOSGiProperties.bndrun")
+		setReportOptional(false)
+		withConvention(FileSetRepositoryConvention::class) {
+
+			// By default bnd will use jars found in:
+			// 1. project.sourceSets.main.runtimeClasspath
+			// 2. project.configurations.archives.artifacts.files
+			// to validate the metadata.
+			// This adds jars defined in `testRuntimeClasses` also so that bnd
+			// can use them to validate the metadata without causing those to
+			// end up in the dependencies of those projects.
+			bundles(sourceSets["test"].runtimeClasspath)
+		}
+	}
+
+	// The ${project.description}, for some odd reason, is only available
+	// afterEvaluate.
+	afterEvaluate {
+		tasks.withType<Jar>().configureEach {
+			convention.findPlugin(BundleTaskConvention::class.java)
+				?.bnd("Bundle-Name: ${project.description}")
 		}
 	}
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,3 +17,8 @@ org.gradle.jvmargs=-Xmx1g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryEr
 org.gradle.caching=true
 org.gradle.parallel=true
 
+# Use this flag so that when compiling against project dependencies
+# the jar is created and used instead of using src/main/classes.
+#
+# See https://docs.gradle.org/current/userguide/java_library_plugin.html#sub:java_library_known_issues_windows_performance
+systemProp.org.gradle.java.compile-classpath-packaging=true
--- a/junit-platform-console-standalone/junit-platform-console-standalone.gradle.kts
+++ b/junit-platform-console-standalone/junit-platform-console-standalone.gradle.kts
@@ -1,3 +1,5 @@
+import aQute.bnd.gradle.BundleTaskConvention;
+
 plugins {
 	`java-library-conventions`
 	id("com.github.johnrengelman.shadow")
@@ -41,6 +43,19 @@ tasks {
 			into("META-INF")
 		}
 
+		withConvention(BundleTaskConvention::class) {
+			bnd("""
+				# Customize the imports because this is an aggregate jar
+				Import-Package: \
+					!org.apiguardian.api,\
+					kotlin.*;resolution:="optional",\
+					*
+				# Disable the APIGuardian plugin since everything was already
+				# processed, again because this is an aggregate jar
+				-export-apiguardian:
+			""")
+		}
+
 		mergeServiceFiles()
 		manifest.apply {
 			inheritFrom(jar.get().manifest)
@@ -58,5 +73,13 @@ tasks {
 					"Multi-Release" to true
 			))
 		}
+	}
+
+	// This jar contains some Java 9 code
+	// (org.junit.platform.console.ConsoleLauncherToolProvider which implements
+	// java.util.spi.ToolProvider which is @since 9).
+	// So in order to resolve this, it can only run on Java 9
+	verifyOSGiProperties {
+		property("-runee", "JavaSE-9")
 	}
 }
--- a/buildSrc/src/main/kotlin/APIGuardianAnnotations.kt
+++ b/buildSrc/src/main/kotlin/APIGuardianAnnotations.kt
@@ -1,0 +1,91 @@
+import aQute.bnd.header.Attrs
+import aQute.bnd.header.OSGiHeader
+import aQute.bnd.header.Parameters
+import aQute.bnd.osgi.Analyzer
+import aQute.bnd.osgi.Clazz
+import aQute.bnd.osgi.Descriptors.TypeRef
+import aQute.bnd.osgi.Instruction
+import aQute.bnd.osgi.Instructions
+import aQute.bnd.service.AnalyzerPlugin
+
+/*
+This is a plugin for bnd which helps analyze the usages of
+org.apiguardian.api.API found in the project bytecode. You can read more
+about this here: https://bnd.bndtools.org/instructions/export-apiguardian.html
+
+Once the next version of bnd releases (likely 5.0.0) this plugin will be
+included in bnd and this class can be removed.
+
+Please ping @rotty3000 to cleanup when that happens.
+*/
+open class APIGuardianAnnotations : AnalyzerPlugin {
+
+	companion object {
+		const val API_ANNOTATION: String = "org/apiguardian/api/API"
+		const val INTERNAL_STATUS: String	= "INTERNAL"
+		const val STATUS_PROPERTY: String	= "status"
+		const val EXPORT_APIGUARDIAN: String = "-export-apiguardian"
+		const val MANDATORY_DIRECTIVE: String = "mandatory:"
+		const val NO_IMPORT_DIRECTIVE: String = "-noimport:"
+	}
+
+	internal enum class Status {
+		INTERNAL,
+		DEPRECATED,
+		EXPERIMENTAL,
+		MAINTAINED,
+		STABLE
+	}
+
+	@Throws(Exception::class)
+	override fun analyzeJar(analyzer:Analyzer):Boolean {
+		// Opt-in is required.
+		val header = OSGiHeader.parseHeader(analyzer.getProperty(EXPORT_APIGUARDIAN))
+		if (header.isEmpty()) return false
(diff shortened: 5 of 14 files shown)


sbrannen:
Tentatively slated for 5.6 M2 for _team discussion_

sbrannen:
I'd really like to see some integration tests in place that actually verify that the generated OSGi metadata is consumable.

@junit-team/junit-committers, thoughts?

rotty3000:
Great feedback. I wanted to make the first pass small concise. I'm happy to make any adjustments and I don't see any issues making the adjustments you want. Finally I can surely add a verification of the metadata. :)

rotty3000:
> Where is the Export-Package statement?
fixed
> API guardian must not be imported
fixed
> Kotlin must always be an optional import.
wonderful
> Why is org.opentest4j the only import with version information?
good question! working on it.

rotty3000:
made `Bundle-Name` === `${project.description}`

sbrannen:
Did you forget to push the changes? Or... are you still working on the updates locally before pushing? (but.... absolutely no rush)

rotty3000:
I've pushed all the above changes now. A major addition is only exporting packages marked with `@API` from all jars.

However there is an issue pending that I'm not sure why but for some reason bnd is not identifying the import versions for project dependencies. This is strange since I do see that bnd has the correct classpath and the jars are there. For example `junit-platform-engine` does not properly import `org.junit.platform.commons.*` with versions. I need to consult with @bjhargrave to figure out what might be the cause.

Also pending is verification of the metadata.

rotty3000:
@sbrannen is it known that it's not possible to build any single module that has project dependencies? For instance when I try `./gradlew :junit-jupiter-engine:jar` it fails with
```
Execution failed for task ':junit-platform-commons:compileModule'.
> Extension with name 'sourceSets' does not exist. Currently registered extension names: [ext, eclipse, idea, defaultArtifacts, spotless]
```
it's not a huge deal, just makes debugging the build pretty hard if I can only execute the entire thing.

sbrannen:
> `./gradlew :junit-jupiter-engine:jar`

I just executed that, and it worked for me.

Maybe you did a `clean` first?

sbrannen:
Hmmm.... `./gradlew clean :junit-jupiter-engine:jar` works for me as well.

You executed that against `master` -- right?

sbrannen:
> However there is an issue pending that I'm not sure why but for some reason bnd is not identifying the import versions for project dependencies. This is strange since I do see that bnd has the correct classpath and the jars are there. For example `junit-platform-engine` does not properly import `org.junit.platform.commons.*` with versions. I need to consult with @bjhargrave to figure out what might be the cause.
> 
> Also pending is verification of the metadata.

OK. Please keep us posted on both accounts.

build.gradle.kts
```diff
@@ -140,6 +142,31 @@ subprojects {
...
+							!org.apiguardian.api,\
+							kotlin.*;resolution:="optional",\
+							*
+						-fixupmessages.kotlin.import: "Unused Import-Package instructions: \\[kotlin.*\\]";is:=ignore
+						-removeheaders: Private-Package
+						-noextraheaders: true
+
+						# The following exports only packages having classes annotated @API
+						# First collect all classes annotated with @API
+						possibleAPI: ${'$'}{classes;HIERARCHY_INDIRECTLY_ANNOTATED;org.apiguardian.api.API}
```
sbrannen: Cool. I didn't know you could do things like that with `bnd`. 👍 

But... that will automatically export `org.junit.jupiter.engine.config` (which should not be exported because all classes annotated with `@API` in that package are actually `@API(status = INTERNAL, ...)`) -- right?

So it looks like we will still have to fine tune things a bit.

Or can you do even more magic with `bnd` to check the `status` attribute of `@API` declarations?

rotty3000:
Let's see what we can come up with. ☺️ It might take me a bit. If you like
I can mark the PR as WIP. Or maybe you have a label...

settings.gradle.kts
```diff
@@ -1,4 +1,12 @@
 pluginManagement {
+	resolutionStrategy {
+		eachPlugin {
+			if (requested.id.id == "biz.aQute.bnd.builder") {
+				useModule("biz.aQute.bnd:biz.aQute.bnd.gradle:4.3.0")
+			}
+		}
+	}
+
```
marcphilipp: Please add the version to the `plugins` block instead and delete `resolutionStrategy` again.

settings.gradle.kts
```diff
@@ -10,6 +18,12 @@ pluginManagement {
 		id("org.asciidoctor.convert") version "1.5.8.1"
 		id("me.champeau.gradle.jmh") version "0.4.8"
 		id("io.spring.nohttp") version "0.0.3.RELEASE"
+		id("biz.aQute.bnd.builder")
+	}
+
+	repositories {
+		gradlePluginPortal()
+		mavenCentral()
```
marcphilipp: Why do we need `mavenCentral()`?

build.gradle.kts
```diff
@@ -140,6 +142,31 @@ subprojects {
 		}
 
 		afterEvaluate {
+			if (project != project(":junit-jupiter")) {
+				tasks["jar"].withConvention(aQute.bnd.gradle.BundleTaskConvention::class) {
+					setBnd("""
```
marcphilipp: Doesn't `bnd = """` work?

build.gradle.kts
```diff
@@ -140,6 +142,31 @@ subprojects {
 		}
 
 		afterEvaluate {
+			if (project != project(":junit-jupiter")) {
+				tasks["jar"].withConvention(aQute.bnd.gradle.BundleTaskConvention::class) {
```
marcphilipp: Will this work for subprojects that have disabled the `jar` task and use `shadowJar` instead?

build.gradle.kts
```diff
@@ -140,6 +142,31 @@ subprojects {
 		}
 
 		afterEvaluate {
+			if (project != project(":junit-jupiter")) {
```
marcphilipp: Why is this in an `afterEvaluate` block?

marcphilipp: Please add this to `java-library-conventions.gradle.kts` instead of to the root build script.

settings.gradle.kts
rotty3000: Due to an issue with the gradle plugin portal reported here https://github.com/gradle/plugin-portal-requests/issues/9 version 4.3.0 is not accessible and so we have to use the approach shown until the issue can be resolved. Sorry about that.

settings.gradle.kts
rotty3000: Also due to the issue with the gradle plugin portal...

build.gradle.kts
rotty3000:
```
Script compilation error:

  Line 147:                                     bnd = """
                 ^ Val cannot be reassigned
```

build.gradle.kts
rotty3000: That's a good question. I only briefly looked at `shadowJar` but I wasn't clear it's real purpose. Now that you mention it I'll have to dig deeper into it. If you could fill me in with a small description of what the intention is there it'll help me make the best choice on how to handle it.

build.gradle.kts
rotty3000: Because that was the only place I could get  `${project.description}` to resolve correctly. Outside of this I was getting null. Any suggestions to work around that would be appreciated.

rotty3000: Great, I wasn't aware of it's existence.

rotty3000: I managed to move the bnd configuration over to `java-library-conventions.gradle.kts`

rotty3000: this was also solved.

build.gradle.kts
marcphilipp: We use the shadow plugin to build a fat jar with relocated dependencies, e.g. in `junit-jupiter-params`. You could probably use the (not very pretty) approach from here to catch both tasks: https://github.com/junit-team/junit5/blob/master/buildSrc/src/main/kotlin/java-library-conventions.gradle.kts#L118

settings.gradle.kts
marcphilipp: I see. Thanks for the explanation!

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
```diff
@@ -180,6 +183,29 @@ tasks.jar {
 				"Implementation-Vendor" to "junit.org"
 		)
 	}
+
+	withConvention(BundleTaskConvention::class) {
+		setBnd("""
+			Bundle-Name: ${project.description}
```
marcphilipp: Isn't this evaluated before the descriptions are set?

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
```diff
@@ -180,6 +183,29 @@ tasks.jar {
...
+				*
+			-fixupmessages.kotlin.import: "Unused Import-Package instructions: \\[kotlin.*\\]";is:=ignore
+			-removeheaders: Private-Package
+			-noextraheaders: true
+
+			# The following exports only packages having classes annotated @API
+			# First collect all classes annotated with @API
+			possibleAPI: ${'$'}{classes;HIERARCHY_INDIRECTLY_ANNOTATED;org.apiguardian.api.API}
+			# Strip off the classNames to obtain a list of package names
+			possiblePackages: ${'$'}{replace;${'$'}{possibleAPI};\.[^.]*${'$'};}
```
marcphilipp: I'm curious if we could we parse the module descriptors instead... 🤔

sbrannen: That would be great it we could!

Better to not duplicate this information.

sormuras: We could parse their source code (similar to [this](https://github.com/sormuras/bach/blob/master/src/bach/Bach.java#L697-L728)) or use the already compiled descriptors -- they are compiled at this stage, aren't they?

kriegfrj:
I like where this project is going, thanks guys. A couple of comments:

* Given the way that JUnit 5 is versioned, perhaps we could consider making use of OSGi contracts instead of straight-out import/export package mechanism? I don't have a lot of first-hand experience with OSGi contracts, but from what little knowledge I do have, they might be a fit. Perhaps @bjhargrave can comment on this.

* I would also like to see us leverage the OSGi capability mechanism in some way to define dependency relationships between test bundles and test engine bundles. That way, the resolver can help with making sure that the correct engines are included in the set of run bundles. For example, on the test engine bundle side, something like this:

```
Provide-Capability: org.junit.platform.engine;
    org.junit.platform.engine.id='junit-jupiter;version="[5.5.1]"' # By convention, make this == TestEngine.getId()
    version=5.5.1
```

...then on the test bundle side, you can declare (eg) the following:

```
Require-Capability: org.junit.platform.engine;
   org.junit.platform.engine.id='junit-jupiter;version="[5.5,6)",junit-vintage;version=4.12,org.myorg.my-test-engine;version=1.0'
```

Now the resolver will be able to ensure that the test bundles are paired with the correct versions of the correct test engines. The syntax might need a bit of tweaking (I'm not an expert in this area), but hopefully the idea makes sense.

build.gradle.kts
rotty3000: k, this is solved.

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
rotty3000: I split this up so that only that one pieces is during `afterEvaluate`

rotty3000:
Ok, only pending are:
- the `@API(status = INTERNAL, ...)`
- the verification

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
rotty3000: The module descriptor will always trail your actual code. You should let tooling give you a result _as close to perfect as a algorithm can bring it_ and then let you take it the last mile. This will eliminate guess work and automatically find details humans aren't good at identifying.

In short, let's leave that part for a future update, shall we? :)

rotty3000: This is not to say that I will not solve the @API exports issue... but rather then parsing of the module descriptor suggestion. ;)

rotty3000:
@kriegfrj let's not shock our goodly junit developers with too much too early :D 

However, I do like your suggestion. Maybe you could open a separate issue specific to capabilities and requirements?

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
rotty3000: ... if that's ok with you all!

kriegfrj:
> @kriegfrj let's not shock our goodly junit developers with too much too early :D
>

:smile: sorry!

> However, I do like your suggestion. Maybe you could open a separate issue specific to capabilities and requirements?

Sure - good idea to keep it separate. Where should I raise the issue - here, on bnd github, or both?

rotty3000:
I'd suggest on this project since that's where you intended to make the change, IIUC.

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
marcphilipp: Fair enough, we could instead add an integration test that verifies the two things are aligned.

rotty3000:
Ok, so, the `@API(status = INTERNAL)` is solved. That also allowed me to do a big cleanup. the changes are much more concise again. You'll note that I added an _inline_ bnd plugin benefiting from the kotling buildSrc that was already there. But do note that bnd is adding support for APIGuardian in the future release (but I didn't want to wait so I added the plugin here).

Only verification remains.

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
rotty3000: resolving this since we have a solution now

rotty3000:
Ok, so before we move on, I wanted to point out that all this did find some oddities:

- `org.junit.platform.commons.logging` is not exported because all it's types are marked as `@API(status = INTERLAN)` however it's imported by almost all other modules.

So, what we're thinking is that _really_ `@API` should always be exported, but that the package exports should also always be marked with their `status` attribute. In the case of INTERNAL, it's really an API for junit itself, like a `exports ... to ...` in JPMS. We can do that also in OSGi. So I'm going to change the heuristic.

sbrannen:
> Ok, so, the `@API(status = INTERNAL)` is solved. 

Great

> But do note that bnd is adding support for APIGuardian in the future release

Oh wow. That's a pleasant surprise.

> Only verification remains.

👍

rotty3000:
Ok, so this PR is finally feature complete, including verification.

gradle.properties
```diff
@@ -17,3 +17,8 @@ org.gradle.jvmargs=-Xmx1g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryEr
 org.gradle.caching=true
 org.gradle.parallel=true
 
+# Use this flag so that when compiling against project dependencies
+# the jar is created and used instead of using src/main/classes.
+#
+# See https://docs.gradle.org/current/userguide/java_library_plugin.html#sub:java_library_known_issues_windows_performance
+systemProp.org.gradle.java.compile-classpath-packaging=true
```
marcphilipp: Why is this required?

rotty3000:
@marcphilipp I will add some comments around all the tasks to explain their use. Standby! :)

gradle.properties
rotty3000: Ok, let me see. 

It begins with this very good question by @sbrannen:
> Why is org.opentest4j the only import with version information?

This was also a surprise to me because bnd gradle plugin was normally always able to determine the version to import a package because it would scan the classpath (essentially `project.sourceSet.main.compile`) and observe the package versions it found within the __jars__ that are there.

Remember, the bnd gradle plugin only adds metadata to the __jar__.

However, this was not working on this project! Why not?

I observed that this project uses the `java-library` gradle plugin which is an observably uncommon scenario as most gradle projects use the `java` plugin.

The distinctive difference here is that the `java` gradle plugin having dependencies of type `project` (e.g. `compile project(":foo")`) will result the project __jars__ being in `project.sourceSet.main.compile`. As I indicated earlier, this allowed bnd to observe the OSGi metadata for those dependent projects.

Now, the `java-library` plugin does _not_ by default cause a project dependency's jar to be built, AND when looking at the ``project.sourceSet.main.compile` will only contain the project output directory. Bnd does not emit metadata into the output directory and therefore there is no metadata to observer.

Setting the flag above to `true` tells the `java-library` plugin to revert to building the jar and using this in `project.sourceSet.main.compile`.

rotty3000: Also, as I indicated in the comment, you can read more about this here https://docs.gradle.org/current/userguide/java_library_plugin.html#sub:java_library_known_issues_windows_performance

rotty3000: Or better https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_classes_usage

rotty3000:
@marcphilipp I added lots of comments, but please if you have more questions I can take those too.

buildSrc/build.gradle.kts
```diff
@@ -10,4 +10,5 @@ repositories {
 dependencies {
 	implementation(kotlin("gradle-plugin"))
 	implementation("de.marcphilipp.gradle:nexus-publish-plugin:0.4.0")
+	implementation("biz.aQute.bnd:biz.aQute.bnd.gradle:4.3.0")
```
kriegfrj: Should we bump to 4.3.1 now that the maintenance release is publicly available?

junit-platform-console-standalone/junit-platform-console-standalone.gradle.kts
```diff
@@ -59,4 +74,10 @@ tasks {
 			))
 		}
 	}
+
+	// This jar contains some Java 9 code (I think from shadowed jansi)
+	// So in order to resolve this, it can only run on Java 9
+	verifyOSGiProperties {
+		property("-runee", "JavaSE-9")
+	}
```
marcphilipp: This feels wrong to me. Do you have a failure message?

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
```diff
@@ -139,6 +143,105 @@ if (project in mavenizedProjects) {
...
+				org.junit.platform.commons.logging;status=INTERNAL,\
+				kotlin.*;resolution:="optional",\
+				*
+
+			# This tells bnd not to complain if a module doesn't actually import
+			# the kotlin packages, but enough modules do to make it a default.
+			-fixupmessages.kotlin.import: "Unused Import-Package instructions: \\[kotlin.*\\]";is:=ignore
+
+			# This tells bnd to ignore classes it files in `META-INF/versions/`
+			# because bnd doesn't yet support multi-version jars.
```
marcphilipp:
```suggestion
			# because bnd doesn't yet support multi-release jars.
```

rotty3000: fixing

junit-platform-console-standalone/junit-platform-console-standalone.gradle.kts
rotty3000: I will check to see exactly what classes are compiled with Java 9 bytecode (bnd uses the class' magic number to make this determination).

rotty3000: The class in question is `org.junit.platform.console.ConsoleLauncherToolProvider` which implements `java.util.spi.ToolProvider` which is `@since 9` :)

rotty3000:
As requested I added a few checks to the manifest test.

rotty3000:
failures here do not seem to be related. :(

rotty3000:
See the same failures on this other sanity check PR https://github.com/junit-team/junit5/pull/2101

sormuras:
`Manifest attribute Bundle-Version expected to be 5.6.0-SNAPSHOT, but is: 5.6.0.SNAPSHOT`

There's `.` written to the manifest instead of a `-`. Who generates the string value that's used for `Bundle-Version`?

rotty3000:
Ah, I didn't see this error. Dot is required by OSGi version syntax. I will parse the version into the correct syntax for this check. Standby!

PR #2106, opened 2019-11-20
[ReflectionUtils] Use class' classloader to load method params

### LINKED ISSUE #2104: `ReflectionUtils.loadRequiredParameterType()` should use `ClassLoader` for the supplied `Class`
In this code here:

https://github.com/junit-team/junit5/blob/b4fef4a8a01c0bfeb8f008a331ec5dc5a9df0869/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java#L1208-L1215

... the `tryToLoad(String)` method uses the default classloader to attempt to load the parameter class. It would probably make more sense to use the `tryToLoad(String, ClassLoader)` variant, and use `clazz.getClassLoader()` instead, because the test class' methods should resolve from the perspective of the test class itself.

Related to #1987.

## Steps to reproduce

I intend to submit a PR to demonstrate the issue (hopefully along with a fix).

## Context

 - Test discovery clients that use `selectMethod(Class<?>...)` variants for methods with parameters.

## Deliverables

* `selectMethod(Class<?>, String, String)` should work across classloaders.

kriegfrj:
Fixes #2104, along with corresponding test case.

## Overview

`ReflectionUtils.loadRequiredParameterType()` will now use class' classloader to resolve parameter type names into `Class` objects.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
@@ -12,9 +12,15 @@ package org.junit.platform.commons.util;
 
 import static org.apiguardian.api.API.Status.INTERNAL;
 
+import java.io.File;
+import java.net.MalformedURLException;
 import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Paths;
 import java.security.CodeSource;
 import java.util.Optional;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import org.apiguardian.api.API;
 
@@ -48,6 +54,43 @@ public final class ClassLoaderUtils {
 			/* otherwise ignore */
 		}
 		return ClassLoader.getSystemClassLoader();
+	}
+
+	static URL toURL(String filePath) {
+		try {
+			return Paths.get(filePath).toUri().toURL();
+		}
+		catch (MalformedURLException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	/**
+	 * Returns a classloader that has visibility to the current classpath, but won't delegate to the parent for classes that
+	 * match the supplied filter. If you attempt to load a class using this loader that matches the filter, then
+	 * it will be loaded directly from the classpath and a new {@link Class} object created.
+	 *
+	 * <p>This is sometimes useful for testing if you have a class in your JVM that's already visible to the current
+	 * class loader, but for testing purposes you need to load it from the context of another class loader. This allows
+	 * you to emulate environments that have multiple classloaders (eg, OSGi).
+	 *
+	 * <p>For an example, see {@link ReflectionUtilsTest#findMethodByParameterTypesFromForeignClassLoader}.
+	 *
+	 * @param filter predicate to match on the type names. If the predicate returns true, then the classloader won't
+	 * delegate to the parent classloader but will load the class directly from the classpath.
+	 * @return The excluding class loader.
+	 */
+	public static URLClassLoader excludingClassLoader(Predicate<String> filter) {
+		return new URLClassLoader(Stream.of(System.getProperty("java.class.path").split(File.pathSeparator)).map(
+			ClassLoaderUtils::toURL).toArray(URL[]::new), new ClassLoader() {
+				@Override
+				protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+					if (filter.test(name)) {
+						throw new ClassNotFoundException("Skipping " + name + " for testing");
+					}
+					return super.loadClass(name, resolve);
+				}
+			});
 	}
 
 	/**
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
@@ -1207,7 +1207,7 @@ public final class ReflectionUtils {
 
 	private static Class<?> loadRequiredParameterType(Class<?> clazz, String methodName, String typeName) {
 		// @formatter:off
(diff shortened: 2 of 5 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
```diff
@@ -50,6 +56,43 @@ public static ClassLoader getDefaultClassLoader() {
 		return ClassLoader.getSystemClassLoader();
 	}
 
+	static URL toURL(String filePath) {
+		try {
+			return Paths.get(filePath).toUri().toURL();
```
sormuras: Replace `Paths.get` with `Path.of` factory - remove the then superseded `import java.nio.file.Paths;` statement as well.

marcphilipp: `Path.of` was introduced in Java 11 so we cannot use it here.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
```diff
@@ -50,6 +56,43 @@ public static ClassLoader getDefaultClassLoader() {
...
+				@Override
+				protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+					if (filter.test(name)) {
+						throw new ClassNotFoundException("Skipping " + name + " for testing");
+					}
+					return super.loadClass(name, resolve);
+				}
+			});
+	}
+
```
marcphilipp: Why did you make these two new methods part of the production code even though they're only used in tests?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
sormuras: My fault.

_We all play on 11, right?_

sormuras: Java 11 [java.nio.file.Paths](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Paths.html)'s Javadcoc contains the followin note:

> **API Note:**
It is recommended to obtain a `Path` via the `Path.of` methods instead of via the `get` methods defined in this class as this class may be deprecated in a future release.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
kriegfrj: The Javadoc for `ClassLoaderUtils` said it was for internal use only.

I'm happy to take guidance on a better place to put it - I wasn't sure where to put test utils. That said, an excluding class loader is also a useful thing to have around when you're doing testing that involves class loaders, so longer term maybe it's not such a silly idea to have it in the production code?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -1207,7 +1207,7 @@ public static boolean isMethodPresent(Class<?> clazz, Predicate<Method> predicat
 
 	private static Class<?> loadRequiredParameterType(Class<?> clazz, String methodName, String typeName) {
 		// @formatter:off
-		return tryToLoadClass(typeName)
+		return tryToLoadClass(typeName, clazz.getClassLoader() == null ? ClassLoaderUtils.getDefaultClassLoader() : clazz.getClassLoader())
```
marcphilipp: I'm not convinced this is actually more correct than before. What if the parameter class is from a different module etc. and should be loaded with a different class loader? If users need to use a special class loader, they could use the findMethod(Class, String, Class...) overload, couldn't they?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
sbrannen: We already have several custom class loaders used in tests.

- org.junit.jupiter.api.AssertThrowsAssertionsTests.EnigmaClassLoader
- org.junit.jupiter.engine.extension.TestInstanceFactoryTests.ProxyClassLoader
- org.junit.jupiter.engine.support.OpenTest4JAndJUnit4AwareThrowableCollectorTests.TestClassLoader

So it might be good to have some central implementations for various testing scenarios, but I don't think that necessarily warrants being in "production" code.

We could also make use of Gradle's "test fixtures" support.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
kriegfrj: The methods of a class - including their parameters - form part of the class' definition.
The class is defined and resolved by the classloader that loaded it. The resolution process involves loading all of the classes referenced by this class' definition (including the method parameter types).

If you use another classloader to load the param types, it *may* work (eg, if you use a parent classloader), but it may not. On the other hand, if you use the same classloader that the class itself (and its method signatures) was defined with, it's guaranteed to work.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClassLoaderUtils.java
kriegfrj: That is fair. Let me know where you'd like the testing utilities to go and I'll look at moving it there.

kriegfrj:
For posterity: @sbrannen used this same solution in da6a0cae9beff18227c2dd21a12cd490c7bcfd92.

Issue #1987, opened 2019-08-20
Support explicit `ClassLoader` in `ClassSelector` and `MethodSelector`


kriegfrj:
This is a specific sub-issue of #806.

I've been adapting JUnit Platform to run in an OSGi environment as part of the Bndtools project (see bndtools/bnd#3145). This project consists of a `BundleEngine` which discovers test bundles in the running OSGi framework, and then delegates to other TestEngines it finds in the environment for each such bundle that it finds.

It became apparent during this implementation that the discovery selectors that do class loading (eg, selectClass(String) and selectMethod(String, String)) ideally needed variants that allowed you to explicitly specify the classloader, rather than always using the default classloader. I had to write a fair bit of code to work around the lack of such a feature when I implemented `BundleEngine`.

The call to `ReflectionUtils.tryToLoadClass(String)`:

https://github.com/junit-team/junit5/blob/cfe9a9928a6c040e8fd6edf96ed70e7eaac94a80/junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/ClassSelector.java#L74-L76

...could then simply be replaced with a call to the `ReflectionUtils.tryToLoadClass(String, ClassLoader)` variant (and possibly update the precondition error message to indicate which classloader it used to try and load the class).

For backwards compatibility, if the classloader is not explicitly specified, simply use the default classloader (current behaviour).

sbrannen:
This is also closely related to #201.

marcphilipp:
@kriegfrj The only currently supported mechanism is to set the calling thread's context classloader which is used in `ReflectionUtils`.

kriegfrj:
Hi @marcphilipp ,
Actually, that's not the only supported mechanism. Another way (which I used in BundleEngine) is to use the `selectClass(Class)` variant instead of `selectClass(String)`, and load the class yourself using the appropriate ClassLoader before handing off to the engine. My proposal here would be essentially the same but moved into `ClassSelector` and `selectClass()` as overloads.

marcphilipp:
Is that enough? I don't think we have any tests that ensure we only use the attached class loader for loading the methods etc. for that test class.

kriegfrj:
> Is that enough? I don't think we have any tests that ensure we only use the attached class loader for loading the methods etc. for that test class.

[...]

PR #2107, opened 2019-11-20
Add `selectClass()`/`selectMethod()` variants in `DiscoverySelectors` that take a `ClassLoader`

kriegfrj:
- Implements #1987

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.0-M1.adoc
@@ -32,6 +32,9 @@ repository on GitHub.
   which return a `Stream`.
 * New `tryToLoadClass(...)` variant in `ReflectionSupport` that accepts an explicit
   `ClassLoader`, allowing classes to be resolved with custom `ClassLoader` arrangements.
+* New overloaded constructors for `ClassSelector` and `MethodSelector` that take an
+  explicit `ClassLoader` as a parameter, allowing selectors to select classes in custom
+  `ClassLoader` arrangements like in OSGi.
 * `ReflectionSupport.findMethod(Class<?>, String, String)` now uses the `ClassLoader` of
   the supplied `Class` to load parameter types instead of using the _default_
   `ClassLoader`. This allows parameter types to be resolved with custom `ClassLoader`
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/ClassSelector.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/ClassSelector.java
@@ -16,6 +16,7 @@ import java.util.Objects;
 
 import org.apiguardian.api.API;
 import org.junit.platform.commons.PreconditionViolationException;
+import org.junit.platform.commons.function.Try;
 import org.junit.platform.commons.util.ReflectionUtils;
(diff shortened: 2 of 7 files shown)


juliette-derancourt:
Tentatively slated for 5.6 M2 solely for the purpose of *team discussion*

marcphilipp:
@kriegfrj Why is this better than `selectClass(classLoader.loadClass(className))`?

kriegfrj:
A fair question...

Perhaps, for similar reasons that selectClass(className) is better than selectClass(ClassLoaderUtils.getDefaultClassLoader().loadClass(className)), maybe?

Aside from being slightly more compact, it also:

1. Loads the class lazily (selector might get filtered before even needing to load the class),
2. Handles the.possible ClassNotFoundException and wraps it in a PreconditionViolationException.

But I'm happy if you guys decide otherwise.

sbrannen:
Hi @kriegfrj,

It looks like you're almost done with this PR (rebased on `main` and updated). 👍

So, it appears that only an entry in the release notes is missing.

Do you think you'll have time to finish this PR this week?

If not, someone in the team can take merge the PR and add an entry in the release notes.

Please just let us know about your availability for completing this issue.

Thanks!

sbrannen (changes requested):
Whenever we introduce an overloaded constructor or method, we do our best to append _optional_ arguments to the formal parameter list instead of prepending them.

This aligns with conventional best practices for API design.

In light of that, please modify all constructors/methods in this PR to accept the `ClassLoader` last in the formal parameter list (unless there is a compelling reason not to do so).

Thanks

kriegfrj:
> Whenever we introduce an overloaded constructor or method, we do our best to append _optional_ arguments to the formal parameter list instead of prepending them.
> 
> This aligns with conventional best practices for API design.
> 
> In light of that, please modify all constructors/methods in this PR to accept the `ClassLoader` last in the formal parameter list (unless there is a compelling reason not to do so).
>

I think the reason I put them first (it's been a while!) was that, from a domain perspective, all the other argument lists go from least specific selector segment to most specific (eg, (class, method) or (class, method, parameters) ). In keeping with that style seemed to make the most sense to put the class loader first.

I started putting it last, and realised that it causes a source-level compatibility as some of the tests started to fail to compile due to the ambiguous overloads when supplied with a `null` parameter (this may also be part of the reason I put them first). Having looked at it though, I think in practice that this will be ok because the preconditions all insist that the `String` arguments are non-null, so there should be no existing code out in the wild that could run afoul of this newly-introduced source ambiguity. So I think in practice only the test cases that are testing the preconditions (attempting to pass in a `null`) are affected.

I've also updated the release notes.

I also realised that, since I submitted the original PR, we also have the `NestedClassSelector` and `NestedMethodSelector` which might need similar treatment... I'm not sure if I'll have time for that too.

sbrannen:
Thanks for making the requested changes and updating the release notes. 👍 

> I also realised that, since I submitted the original PR, we also have the `NestedClassSelector` and `NestedMethodSelector` which might need similar treatment... I'm not sure if I'll have time for that too.

Good catch.

I opened #3298 to track that.

sbrannen (approved):

I'll make any additional changes after merging.

sbrannen:
This has been merged into `main` in c12634cbc65b9da528681cf3abc4b79814a3656e and revised in 142476c62c7c0e00e8b84a890a5a1319826b7cdf.

Thanks

Issue #1596, opened 2018-09-23
Create IndicativeSentences Display Name Generator


sormuras:
## Overview

As a follow-up from #1588 use the `DisplayNameGenerator` SPI to turn names of test classes and test methods using camel-case and/or underscores in their notation into human-readable display names that form complete sentences.

## Example

From "fragments" using `@DisplayName` for each fragment, thus duplicating the same information as encoded in the associated method name...

https://github.com/junit-team/junit5/blob/91279b2cb70636ea4c24bb4a1b5a4e976adab405/documentation/src/test/java/example/TestingAStackDemo.java#L27-L97

...resulting in this picture...

![fragments](https://user-images.githubusercontent.com/2319838/45813051-dc006400-bcd1-11e8-9543-ac5279c2c6c0.png)

...to "indicative sentences", merging contexts and auto-prettify display names based on the method names:

![sentences](https://user-images.githubusercontent.com/2319838/45813065-e1f64500-bcd1-11e8-9b10-717b6d1ff82e.png)

## Related Issues

- #1595 

## Deliverables

- [ ] Create and implement `IndicativeSentencesGenerator` in `org.junit.jupiter.api.DisplayNameGenerator`
- [ ] Integrate this new pre-defined generator into `org.junit.jupiter.engine.descriptor.DisplayNameUtils`
- [ ] Extend `org.junit.jupiter.api.DisplayNameGenerationTests` to verify the implementation

olcbean:
I would like to take a stab at this, if nobody is working on it. Once I am ready: should I open the PR against Pioneer or better here and after the review you will move it (if needed)?

Just one more quick nit question to make sure that we are on the same page: in the attached snippet and png there is a transition that I am not sure it is feasible to make (`with new Stack()` -> `using its noarg constructor.`)
```
@Test 
@DisplayName("is instantiated with new Stack()") 
```
```
A stack is instantiated using its noarg constructor.
```
I would go ahead and assume that the name should be `A stack is instantiated with new Stack()` instead;) 
And the same thinking for `A stack when new ...` instead of `A new stack ...`.

juliette-derancourt:
Is this issue still relevant? It seems that the related PR was closed because you suggested to move it to JUnit Pioneer, but I see no such commit there... Is there something to do?

JoosJuliet:
Can I grab this one? I can't find any assigned person. Could I do this?

RoryBlevins:
[...]

PR #1615, opened 2018-10-03
Add CamelCase Display Name Generator

zoltanersek:
Issue: #1595

## Overview

Please describe your changes here and list any open questions you might have.

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement).

---

### Definition of Done

- [x] There are no TODOs left in the code
- [ ] Method [preconditions](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [ ] [Coding conventions](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [ ] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling
- [ ] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](http://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/meta/API.html)
- [ ] Change is documented in the [User Guide](http://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](http://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [x] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #2124, opened 2019-12-10
Issue/1596 indicative sentences

Kazhuru:
## Overview

Hi!, as followed up in the issue #1596, this is the pull request of the implementation of a IndicativeSentencesGenerator supporting the RemplaceUnderscores class and with DisplayName's.

I have a open question: https://github.com/junit-team/junit5/issues/1596#issuecomment-562296145 Here I proposed that might be better to use a variable string to separate the result of the nodes of the test tree, instead of a fixed comma, but I am not sure how to pass that variable.

---

---
### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
@@ -146,8 +146,8 @@ public interface DisplayNameGenerator {
 	 * {@code DisplayNameGenerator} that supports {@code ReplaceUnderscores} to generate complete sentences.
 	 *
 	 * <p>This generator extends the functionality of {@link ReplaceUnderscores} by
-	 * replacing all underscores ({@code '_'}) found in class and method names
-	 * with spaces ({@code ' '}).
+	 * generating a human-readable display names that form complete sentences divided each
+	 * class, nested class and test by a ({@code ','}).
 	 */
 	class IndicativeSentencesGenerator extends ReplaceUnderscores {
 
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
@@ -13,6 +13,7 @@ package org.junit.jupiter.api;
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
 import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
 
 import org.apiguardian.api.API;
 import org.junit.platform.commons.util.ClassUtils;
@@ -138,6 +139,111 @@ public interface DisplayNameGenerator {
 
 		private static boolean hasParameters(Method method) {
 			return method.getParameterCount() > 0;
+		}
+	}
+
+	/**
+	 * {@code DisplayNameGenerator} that supports {@code ReplaceUnderscores} to generate complete sentences.
+	 *
+	 * <p>This generator extends the functionality of {@link ReplaceUnderscores} by
+	 * replacing all underscores ({@code '_'}) found in class and method names
+	 * with spaces ({@code ' '}).
+	 */
+	class IndicativeSentencesGenerator extends ReplaceUnderscores {
+
+		@Override
+		public String generateDisplayNameForClass(Class<?> testClass) {
+			String indicativeName = removeRootCharacter(super.generateDisplayNameForClass(testClass));
+			return indicativeName;
+		}
+
+		@Override
+		public String generateDisplayNameForNestedClass(Class<?> nestedClass) {
+			String indicativeName = classReplaceToIndicativeSentence(nestedClass);
+			return indicativeName;
+		}
+
+		@Override
+		public String generateDisplayNameForMethod(Class<?> testClass, Method testMethod) {
+			String methodName = underscoreSnakeCaseName(super.generateDisplayNameForMethod(testClass,testMethod));
+			String indicativeName = classReplaceToIndicativeSentence(testClass);
+
+			return indicativeName + ", " + methodName ;
+		}
+
+		private String classReplaceToIndicativeSentence(Class<?> testClass) {
+			String classIndicativeSentence = null;
+			Class<?> classWithEnclosingParent = testClass.getEnclosingClass();
+			DisplayName classWithDisplayName = testClass.getAnnotation(DisplayName.class);
+			DisplayNameGeneration classWithAnnotation = testClass.getAnnotation(DisplayNameGeneration.class);
+
+				if(classWithEnclosingParent == null ) {
+					if(classWithDisplayName != null)
+						classIndicativeSentence = classWithDisplayName.value();
+					else
+						classIndicativeSentence = generateDisplayNameForClass(testClass);
+				}
+				else {
+					if(classWithAnnotation != null) {
+						if (classWithAnnotation.value() == IndicativeSentencesGenerator.class) {
+							if(classWithDisplayName != null)
+								classIndicativeSentence = classWithDisplayName.value();
+							else
+								classIndicativeSentence = generateDisplayNameForClass(testClass);
+						}
+					}
+					else {
+						if(classWithDisplayName != null)
+							classIndicativeSentence = classWithDisplayName.value();
+						else
+							classIndicativeSentence = super.generateDisplayNameForNestedClass(testClass);
+
+						classIndicativeSentence = classReplaceToIndicativeSentence(classWithEnclosingParent) +
+										", " +  classIndicativeSentence;
+					}
+				}
+
+			return underscoreSnakeCaseName(classIndicativeSentence);
+		}
+
+		private String removeRootCharacter(String testName)
+		{
+			int indexOfChar = testName.indexOf('$');
+
+			if(testName.indexOf('$')>=0)
+				testName = testName.substring(indexOfChar+1);
+
+			return testName;
+		}
+
+		private String underscoreSnakeCaseName(String testName)
+		{
+			if(testName.length() <= 1)
+				return testName;
+			else
+			{
+				String [] splitTestWords = testName.split(" ");
+
+				for (int i = 1; i < splitTestWords.length; i++) {
+					if (checkTwoCapsString(splitTestWords[i]) == false)
+						splitTestWords[i] = Character.toLowerCase(splitTestWords[i].charAt(0)) + splitTestWords[i].substring(1);
+				}
+
+				return String.join(" ",splitTestWords);
+			}
+		}
+
+		private boolean checkTwoCapsString(String inputWord) {
+			char fragmentChar;
+			int lowerCaseCount = 0;
+
+			for(int i=0;i < inputWord.length();i++) {
+				fragmentChar = inputWord.charAt(i);
+				if (Character.isUpperCase(fragmentChar))
+					lowerCaseCount++;
+			}
+
+			return (lowerCaseCount >=2);
 		}
 	}
 
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
@@ -20,6 +20,7 @@ import java.util.function.Supplier;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.DisplayNameGeneration;
 import org.junit.jupiter.api.DisplayNameGenerator;
+import org.junit.jupiter.api.DisplayNameGenerator.IndicativeSentencesGenerator;
 import org.junit.jupiter.api.DisplayNameGenerator.ReplaceUnderscores;
 import org.junit.jupiter.api.DisplayNameGenerator.Standard;
 import org.junit.jupiter.engine.config.JupiterConfiguration;
@@ -50,6 +51,11 @@ final class DisplayNameUtils {
 	 * Pre-defined display name generator instance replacing underscores.
 	 */
 	private static final DisplayNameGenerator replaceUnderscoresGenerator = new ReplaceUnderscores();
+
+	/**
+	 * Pre-defined display name generator instance with indicative sentences.
+	 */
+	private static final DisplayNameGenerator indicativeSentencesGenerator = new IndicativeSentencesGenerator();
 
 	static String determineDisplayName(AnnotatedElement element, Supplier<String> displayNameSupplier) {
 		Preconditions.notNull(element, "Annotated element must not be null");
@@ -98,6 +104,9 @@ final class DisplayNameUtils {
 					}
 					if (displayNameGeneratorClass == ReplaceUnderscores.class) {
 						return replaceUnderscoresGenerator;
+					}
+					if (displayNameGeneratorClass == IndicativeSentencesGenerator.class) {
+						return indicativeSentencesGenerator;
 					}
 					return ReflectionUtils.newInstance(displayNameGeneratorClass);
 				}) //
--- a/junit-jupiter-engine/src/test/java/org/junit/jupiter/api/DisplayNameGenerationTests.java
+++ b/junit-jupiter-engine/src/test/java/org/junit/jupiter/api/DisplayNameGenerationTests.java
@@ -68,6 +68,22 @@ class DisplayNameGenerationTests extends AbstractJupiterTestEngineTests {
 	}
 
 	@Test
+	void indicativeSentencesGenerator() {
+		var expectedDisplayNames = List.of( //
+						"CONTAINER: IndicativeStyleTestCase", //
+						"TEST: @DisplayName prevails", //
+						"TEST: IndicativeStyleTestCase, test", //
+						"TEST: IndicativeStyleTestCase, test (TestInfo)", //
+						"TEST: IndicativeStyleTestCase, test with underscores", //
+						"TEST: IndicativeStyleTestCase, testUsingCamelCase and also UnderScores", //
+						"TEST: IndicativeStyleTestCase, testUsingCamelCase and also UnderScores keepingParameterTypeNamesIntact (TestInfo)", //
+						"TEST: IndicativeStyleTestCase, testUsingCamelCaseStyle" //
+		);
+		check(IndicativeStyleTestCase.class, expectedDisplayNames);
+		//check(IndicativeSentencesInheritedFromSuperClassTestCase.class, expectedDisplayNames);
+	}
+
+	@Test
 	void noNameGenerator() {
 		check(NoNameStyleTestCase.class, List.of( //
 			"CONTAINER: nn", //
@@ -94,6 +110,19 @@ class DisplayNameGenerationTests extends AbstractJupiterTestEngineTests {
 			"TEST: the stack is no longer empty", //
 			"TEST: throws an EmptyStackException when peeked", //
 			"TEST: throws an EmptyStackException when popped" //
+		));
+	}
+
+	@Test
+	void checkDisplayNameGeneratedForIndicativeGeneratorTestCase() {
+		check(IndicativeGeneratorTestCase.class, List.of( //
+			"CONTAINER: A stack", //
+			"CONTAINER: A stack, when is new", //
+			"CONTAINER: A stack, when is new, after pushing an element to an empty stack", //
+			"TEST: A stack, is instantiated with new constructor", //
+			"TEST: A stack, when is new, after pushing an element to an empty stack, is no longer empty", //
+			"TEST: A stack, when is new, throws EmptyStackException when peeked", //
+			"TEST: A stack, when is new, throws EmptyStackException when popped" //
 		));
 	}
 
@@ -170,6 +199,10 @@ class DisplayNameGenerationTests extends AbstractJupiterTestEngineTests {
 	static class UnderscoreStyleTestCase extends AbstractTestCase {
 	}
 
+	@DisplayNameGeneration(DisplayNameGenerator.IndicativeSentencesGenerator.class)
+	static class IndicativeStyleTestCase extends AbstractTestCase {
+	}
+
 	@DisplayNameGeneration(NoNameGenerator.class)
 	static class NoNameStyleTestCase extends AbstractTestCase {
 	}
@@ -177,7 +210,6 @@ class DisplayNameGenerationTests extends AbstractJupiterTestEngineTests {
 	// No annotation here! @DisplayNameGeneration is inherited from super class
 	static class UnderscoreStyleInheritedFromSuperClassTestCase extends UnderscoreStyleTestCase {
 	}
-
 	// -------------------------------------------------------------------
 
 	@DisplayName("A stack")
@@ -238,6 +270,55 @@ class DisplayNameGenerationTests extends AbstractJupiterTestEngineTests {
 				@Test
 				void peek_returns_that_element_without_removing_it_from_the_stack() {
 					assertEquals(anElement, stack.peek());
+					assertFalse(stack.isEmpty());
+				}
+			}
+		}
+	}
+
+	// -------------------------------------------------------------------
+
+	@DisplayName("A stack")
+	@DisplayNameGeneration(DisplayNameGenerator.IndicativeSentencesGenerator.class)
+	static class IndicativeGeneratorTestCase {
+
+		Stack<Object> stack;
+
+		@Test
+		void is_Instantiated_With_New_Constructor() {
+			new Stack<>();
+		}
+
+		@Nested
+		class When_Is_New {
+
+			@BeforeEach
+			void create_With_New_Stack() {
+				stack = new Stack<>();
+			}
+
+			@Test
+			void throws_EmptyStackException_When_Peeked() {
+				assertThrows(EmptyStackException.class, () -> stack.peek());
+			}
+
+			@Test
+			void throws_EmptyStackException_When_Popped() {
+				assertThrows(EmptyStackException.class, () -> stack.pop());
+			}
+
+			@Nested
+			class After_pushing_an_element_to_an_empty_stack {
+
+				String anElement = "an element";
+
+				@BeforeEach
+				void push_An_Element() {
+					stack.push(anElement);
+				}
+
+				@Test
+				void is_no_longer_empty() {
 					assertFalse(stack.isEmpty());
 				}
 			}


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
```diff
@@ -51,6 +52,11 @@
 	 */
 	private static final DisplayNameGenerator replaceUnderscoresGenerator = new ReplaceUnderscores();
 
+	/**
+	 * Pre-defined display name generator instance with indicative sentences.
```
juliette-derancourt:
```suggestion
	 * Pre-defined display name generator instance producing indicative sentences.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,134 @@ private static boolean hasParameters(Method method) {
...
+
+		private String classReplaceToIndicativeSentence(Class<?> testClass) {
+			String classIndicativeSentence = null;
+			Class<?> classWithEnclosingParent = testClass.getEnclosingClass();
+			DisplayName classWithDisplayName = testClass.getAnnotation(DisplayName.class);
+			DisplayNameGeneration classWithAnnotation = testClass.getAnnotation(DisplayNameGeneration.class);
+
+			if (classWithEnclosingParent == null) {
+				if (classWithDisplayName != null)
+					classIndicativeSentence = classWithDisplayName.value();
```
juliette-derancourt: You should return the string here instead of storing it and have the return statement at the end of the method (same for the other occurences below). This way, you can also get rid of the initialization at line 173.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,134 @@ private static boolean hasParameters(Method method) {
...
+
+		/**
+		 * Generate a string which runs through the test name of a name parsed
+		 * by {@link ReplaceUnderscores} and removes the uppercase of the first letter.
+		 *
+		 * @param testName the Test Class from to extract the parameter types from;
+		 * never {@code blank}.
+		 * @return a string without underscore conversion capital letters.
+		 */
+		private String underScoreNameFormatting(String testName) {
```
juliette-derancourt: Do we really want to do that? I'm not sure we should go in such a level of detail...

sbrannen:
> I have a open question: [#1596 (comment)](https://github.com/junit-team/junit5/issues/1596#issuecomment-562296145) Here I proposed that might be better to use a variable string to separate the result of the nodes of the test tree, instead of a fixed comma, but I am not sure how to pass that variable.

Well, it would have to be supplied by the user.

Thus, I could foresee two options.

1. Create a new annotation to allow it to be configured per test class hierarchy.
2. Introduce a new configuration parameter to allow it to be configured globally.

I'm not sure if # 1 wouldn't be overkill. So I guess that's debatable; however, the global option would likely be necessary in any case (if we make it configurable).

Kazhuru:
Thanks for the feedback @juliette-derancourt, I'll start solving your revisions with a next commit.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,134 @@ private static boolean hasParameters(Method method) {
...
+		 * @param testName the Test Class from to extract the parameter types from;
+		 * never {@code blank}.
+		 * @return a string without the $ symbol and the root data that comes from
+		 * {@code Class.getName()}.
+		 */
+		private String removeRootCharacter(String testName) {
+			Preconditions.notBlank(testName, "Input Name parameter should not be blank");
+			int indexOfChar = testName.indexOf('$');
+
+			if (testName.indexOf('$') >= 0)
```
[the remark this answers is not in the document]
Kazhuru: You are right, the reason of the method is to recover the last element, I will modify the logic with a split.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,134 @@ private static boolean hasParameters(Method method) {
...
+	 * <p>This generator extends the functionality of {@link ReplaceUnderscores} by
+	 * generating a human-readable display names that form complete sentences divided each
+	 * class, nested class and test by a ({@code ','}).
+	 */
+	class IndicativeSentencesGenerator extends ReplaceUnderscores {
+
+		@Override
+		public String generateDisplayNameForClass(Class<?> testClass) {
+			String indicativeName = removeRootCharacter(super.generateDisplayNameForClass(testClass));
+			return indicativeName;
```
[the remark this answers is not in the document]
Kazhuru: Roger that! I didn't see that detail.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: I see, as I had the idea to concatenate with the parents recursively, so I decided that way, but yeah sure with the returns maybe a cleaner way.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: The purpose of these methods is for,
maintaining integrity when variations appear with the use of @DisplayName and if you write an underscore convention with uppercase letter at the beginning or in all except the first one, 
example,


```
@DisplayName("A stack")
class TestCase { ...

  @Nested
  class After_pushing_an_element_to_an_empty_stack {

    @Test
    void throws_EmptyStackException_When_Peeked() 
}

----> A stack
--------> A stack, After pushing an element to an empty stack
----------->A stack, After pushing an element to an empty stack, throws EmptyStackException When Peeked 
```
That it looks a kind of strange the output with uppercase letters the first letter in some words and in others not.

Anyway, if you still consider this formatting is too much, with out them the output will look like the example in that test and obviosuly the purpose of the feature would not be affected.

Kazhuru:
> I'm not sure if # 1 wouldn't be overkill. So I guess that's debatable; however, the global option would likely be necessary in any case (if we make it configurable).

Thank you for the answer @sbrannen I was somewhat lost in how the annotations were handled, but with your comment it gave me the lead to know where to look, I wil try to do it the second way.

Kazhuru:
Shame that didn't not work on the OpenJDK Cross-version :( ,  I was confident that my Graddle builds we're happy, the error that is failing is happening in:
`platform.tooling.support.tests.ModularUserGuideTests.runTestsFromUserGuideWithinModularBoundaries` 
When running the build tasks in platform-tooling-support and in documentation's asciidoctor everything was right.
Maybe the CI error generated because i did not added the annotation `@Documented` in `IndicativeSentencesSeparator.java` file.

marcphilipp:
@Kazhuru The failure on JDK 14 is _not_ caused by your changes but by me upgrading to 14.ea.27. @sormuras is currently looking into fixing it on `master`. You can then rebase when we're ready.

documentation/src/test/java/example/DisplayNameGeneratorDemo.java
```diff
@@ -40,36 +39,16 @@ void if_it_is_negative(int year) {
...
-		@Override
-		public String generateDisplayNameForNestedClass(Class<?> nestedClass) {
-			return super.generateDisplayNameForNestedClass(nestedClass) + "...";
-		}
-
-		@Override
-		public String generateDisplayNameForMethod(Class<?> testClass, Method testMethod) {
-			String name = testClass.getSimpleName() + ' ' + testMethod.getName();
-			return name.replace('_', ' ') + '.';
+		@DisplayName("if it is one of the following years.")
```
marcphilipp: How about the following?

```java
		@ParameterizedTest(name = "{0}")
		@ValueSource(ints = { 2016, 2020, 2048 })
		void if_it_is(int year) {
		}
```

sormuras:
> @Kazhuru The failure on JDK 14 is _not_ caused by your changes but by me upgrading to 14.ea.27. @sormuras is currently looking into fixing it on `master`. You can then rebase when we're ready.

Let's see how this plays out: https://github.com/junit-team/junit5/commit/70d10c6e24993fafa8b6980b948e0aaa2432c7c7

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
marcphilipp: I agree with @juliette-derancourt. Let's not do any capitalization changes.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesSeparator.java
```diff
@@ -0,0 +1,44 @@
...
+ * in the IndicativeSentencesGenerator, if this notation was not declared
+ * IndicativeSentencesGenerator will use ", " as separator by default.
+ *
+ * @see DisplayNameGenerator
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@API(status = EXPERIMENTAL, since = "5.4")
```
marcphilipp:
```suggestion
@API(status = EXPERIMENTAL, since = "5.6")
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesSeparator.java
```diff
@@ -0,0 +1,44 @@
...
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+
+/**
+ * {@code @IndicativeSentencesSeparator} is used to declare a custom separator
+ * in the IndicativeSentencesGenerator, if this notation was not declared
+ * IndicativeSentencesGenerator will use ", " as separator by default.
+ *
+ * @see DisplayNameGenerator
```
marcphilipp: `@since` is missing.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,144 @@ private static boolean hasParameters(Method method) {
...
 	}
 
+	/**
+	 * {@code DisplayNameGenerator} that supports {@code ReplaceUnderscores} to generate complete sentences.
+	 *
+	 * <p>This generator extends the functionality of {@link ReplaceUnderscores} by
+	 * generating a human-readable display names that form complete sentences divided each
+	 * class, nested class and test by a ({@code ','}).
+	 */
+	class IndicativeSentencesGenerator extends ReplaceUnderscores {
```
marcphilipp: Let's make this `@API(status = experimental, since = "5.6")` and add a corresponding `@since` tag to the Javadoc.

sormuras:
> > @Kazhuru The failure on JDK 14 is _not_ caused by your changes but by me upgrading to 14.ea.27. @sormuras is currently looking into fixing it on `master`. You can then rebase when we're ready.
> 
> Let's see how this plays out: [70d10c6](https://github.com/junit-team/junit5/commit/70d10c6e24993fafa8b6980b948e0aaa2432c7c7)

Current master is green again: https://github.com/junit-team/junit5/commits/master

Kazhuru:
Thanks for the code review @marcphilipp I'll start to check in to, also the rebase to master to get the @sormuras fix in a next commit.

documentation/src/test/java/example/DisplayNameGeneratorDemo.java
Kazhuru: Interesting, I did not test the output with ParametrizedTest I'll check how it works and add this on the test's file

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: All right, the next commit will have the changes to revert that and also in the DisplayNamesTest file.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesSeparator.java
Kazhuru: Thanks!, was not sure what version should be there, changed on the next commit. 👍

documentation/src/test/java/example/DisplayNameGeneratorDemo.java
Kazhuru: The result of your example gives this output:
Screen Shot 2019-12-17 at 12 17 20 PM

In fact to DisplayGenerator is getting the "(int)" at the last of the name this also happens in ReplaceUndescore, 
the concatenation with the function `if_it_is()` with the separator and his parent class is working,
and output seems odd, because the parameter `name = "{0}"` just print the numbers, if we change that with a string name like: `@ParameterizedTest(name = " Year {0} is a leap year.")` will output this:

Screen Shot 2019-12-17 at 12 38 33 PM

I think the problem is that the **ParametrizedTest** data printing cycle does not go through **DisplayGenerator**, maybe those results could be outside the scope of this extension proposal?

Kazhuru:
Thanks for the quick responses guys! , that commit was a merge because I created a conflict trying to rebase upstream sorry for the mess 😓, also the rebase solved the conflicts with the OpenJDK 14, but Windows CI now throws,
`JUnit Jupiter:KotlinAssertionsDemo:timeout exceeded with preemptive termination() `
 maybe something happened in the last commit to master.

sormuras:
> [...] but Windows CI now throws,
> `JUnit Jupiter:KotlinAssertionsDemo:timeout exceeded with preemptive termination()`
> [...]

Restartet tests. That one is flaky on CI systems...

Kazhuru:
Yep @sormuras that reset worked, Thanks for the help

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -160,12 +160,53 @@ include::{testDir}/example/DisplayNameGeneratorDemo.java[tags=user_guide]
...
   | '-- if it is zero() [OK]
-  '-- A year is a leap year... [OK]
-    +-- A year is a leap year if it is divisible by 4 but not by 100. [OK]
-    '-- A year is a leap year if it is one of the following years. [OK]
-      +-- Year 2016 is a leap year. [OK]
-      +-- Year 2020 is a leap year. [OK]
-      '-- Year 2048 is a leap year. [OK]
+  '-- A year is a leap year [OK]
+    +-- A year is a leap year, if it is divisible by 4 but not by 100. [OK]
+    '-- A year is a leap year, if it is one of the following years. [OK]
```
MaximumEndurance: We might add the deleted lines (166-168) after this sentence just so that we are actually adding the examples as we state above

Kazhuru: Sure! no problem re-adding the examples, I think I chopped them up thinking it would be simpler.

MaximumEndurance: You are completely right. I just meant, probably that line should also be removed with if we remove the examples. It may be confusing for the reader. It's your take on adding or removing the examples, you are right, probably every reader would know examples of leap year. The comment was to point out a discrepancy. Thanks for looking into it!

documentation/src/test/java/example/DisplayNameGeneratorDemo.java
marcphilipp: Fair enough.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesSeparator.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @since 5.6
+ * @see DisplayNameGenerator
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@API(status = EXPERIMENTAL, since = "5.6")
+public @interface IndicativeSentencesSeparator {
```
marcphilipp: Let's rename this annotation to `@IndicativeSentencesGeneration`, rename `value` to `separator` (so we can easily add additional config options in the future), and meta-annotate it with `@DisplayNameGeneration(IndicativeSentences.class)`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesSeparator.java
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * Custom separator for indicative sentences generator.
+	 *
+	 * @return custom separator for indicative sentences
+	 */
+	String value();
```
marcphilipp: Let's set `""` as the dummy default value so users can use the annotation without any parameters and we can detect whether they set the separator explicitly.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,100 @@ private static boolean hasParameters(Method method) {
...
+		}
+
+		@Override
+		public String generateDisplayNameForMethod(Class<?> testClass, Method testMethod) {
+			return classReplaceToIndicativeSentence(testClass) + getSentenceSeparator(testClass)
+					+ super.generateDisplayNameForMethod(testClass, testMethod);
+		}
+
+		private String classReplaceToIndicativeSentence(Class<?> testClass) {
+			Class<?> classWithEnclosingParent = testClass.getEnclosingClass();
```
marcphilipp:
```suggestion
			Class<?> enclosingClass = testClass.getEnclosingClass();
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,100 @@ private static boolean hasParameters(Method method) {
...
+			DisplayName classWithDisplayName = testClass.getAnnotation(DisplayName.class);
+			DisplayNameGeneration classWithAnnotation = testClass.getAnnotation(DisplayNameGeneration.class);
```
marcphilipp: Please use `AnnotationSupport.findAnnotation` in the above two lines to support meta annotations.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesSeparator.java
Kazhuru: Sure!, on it with the refactor

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesSeparator.java
Kazhuru: Okay, I hadn't thought about the case of the annotation without the parameter.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Roger, applying changes to the following commit.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,100 @@ private static boolean hasParameters(Method method) {
...
+	 * {@code DisplayNameGenerator} that supports {@code ReplaceUnderscores} to generate complete sentences.
+	 *
+	 * <p>This generator extends the functionality of {@link ReplaceUnderscores} by
+	 * generating a human-readable display names that form complete sentences divided each
+	 * class, nested class and test by a ({@code ','}).
+	 *
+	 * @since 5.6
+	 */
+	@API(status = EXPERIMENTAL, since = "5.6")
+	class IndicativeSentencesGenerator extends ReplaceUnderscores {
```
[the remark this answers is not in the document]
Kazhuru: Makes sense, I named it because I thought it generated something, I'll start a name refactoring.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
juliette-derancourt: Plus, the method returns an `Optional` so you might be able to remove some if statements below ;)

juliette-derancourt: Also, I think it would be more readable if you'd rename these variables `displayName` and `displayNameGeneration`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,100 @@ private static boolean hasParameters(Method method) {
...
+		 * Gets the separator for {@link IndicativeSentencesSeparator} when extracting the
+		 * annotation from {@code IndicativeSentencesSeparator}, if it doesn't find it,
+		 * then search for the parent classes, if no separator is found use @code{", "} by default.
+		 *
+		 * @param currentClass the Test Class the separator either custom or default
+		 * @return the indicative sentence separator
+		 * {@code Class.getName()}.
+		 */
+		private String getSentenceSeparator(Class<?> currentClass) {
+			IndicativeSentencesSeparator separator = currentClass.getAnnotation(IndicativeSentencesSeparator.class);
```
juliette-derancourt: You can also make use of `findAnnotation` here

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +141,100 @@ private static boolean hasParameters(Method method) {
...
+		 *
+		 * @param testName the Test Class from to extract the parameter types from;
+		 * never {@code blank}.
+		 * @return a string without the $ symbol and the root data that comes from
+		 * {@code Class.getName()}.
+		 */
+		private String removeRootCharacter(String testName) {
+			Preconditions.notBlank(testName, "Input Name parameter should not be blank");
+			String[] testNameSplit = testName.split("\\$");
+			int last = testNameSplit.length - 1;
```
juliette-derancourt: Now we may have a different problem... what if there's a `$` in the test name? 😁 (it's possible I think?)

juliette-derancourt:
> Thus, I could foresee two options.
> 1. Create a new annotation to allow it to be configured per test class hierarchy.
> 2. Introduce a new configuration parameter to allow it to be configured globally.

> I'm not sure if # 1 wouldn't be overkill. So I guess that's debatable; however, the global option would likely be necessary in any case (if we make it configurable).

+1 for the global configuration

I agree that # 1 is a bit overkill, but maybe it could be useful if we want to make other parts configurable? That way the "sentence" can be adapted for each test.
Not sure people will go into such trouble to make test names pretty though... 🤓

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: thanks @juliette-derancourt, done in the @marcphilipp refactoring suggestions :D

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Oh I did not thought about it, I'll check what happens

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @since 5.6
+ * @see DisplayNameGenerator
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@API(status = EXPERIMENTAL, since = "5.6")
+public @interface IndicativeSentencesGeneration {
```
juliette-derancourt: I'm wondering if we could add `@DisplayNameGeneration(DisplayNameGenerator.IndicativeSentences.class)` here, so that it could be omitted on the class? Would it make sense?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Yeah, seems that we can do stuff like this kind of naming:
Screen Shot 2019-12-19 at 3 12 32 PM

Also, yes it makes troubles when in case happens  with that method, so i decided to remove ir and use the method `class.simpleName() ` to solve this.

Kazhuru:
There are some limits to fail on a timeout test? 🤔  currently droping only on MAC's CI: `org.junit.jupiter.engine.extension.TimeoutExtensionTests > appliesTimeoutOnTestableMethodsInAnnotatedClasses()[1] FAILED`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
Kazhuru: Do you mean that the whole implementation should be on that file?

marcphilipp: That's also what I meant in https://github.com/junit-team/junit5/pull/2124#discussion_r359873187. If we annotate `IndicativeSentencesGeneration` with `@DisplayNameGeneration(DisplayNameGenerator.IndicativeSentences.class)`, users could use `@IndicativeSentencesGeneration` without having to also annotate their classes with `@DisplayNameGeneration(DisplayNameGenerator.IndicativeSentences.class)`.

documentation/src/test/java/example/DisplayNameIndicativeSentencesDemo.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.jupiter.api.DisplayNameGeneration;
+import org.junit.jupiter.api.DisplayNameGenerator;
+import org.junit.jupiter.api.IndicativeSentencesGeneration;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class DisplayNameIndicativeSentencesDemo {
+
+	@Nested
+	@DisplayNameGeneration(DisplayNameGenerator.IndicativeSentences.class)
```
marcphilipp:
```suggestion
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +143,78 @@ private static boolean hasParameters(Method method) {
...
+		private String classReplaceToIndicativeSentence(Class<?> testClass) {
+			Class<?> enclosingParent = testClass.getEnclosingClass();
+			Optional<DisplayName> displayName = AnnotationSupport.findAnnotation(testClass, DisplayName.class);
+			Optional<DisplayNameGeneration> displayNameGeneration = AnnotationSupport.findAnnotation(testClass,
+				DisplayNameGeneration.class);
+
+			if (enclosingParent == null) {
+				return displayName.map(DisplayName::value).orElseGet(() -> generateDisplayNameForClass(testClass));
+			}
+			else { //&& displayNameGeneration.getClass() == IndicativeSentences.class
```
marcphilipp: ❓ Is this sth. you meant to check?

Kazhuru: Oh that was just a miss comment that i]I got left to remove, I was just playing around with those methods values while debugging, I'll get rid of it on the https://github.com/junit-team/junit5/pull/2124#discussion_r360117501 changes.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
Kazhuru: Yeah, now I can see what you mean, I've never heard of Meta-Annotations, luckily, this page
https://dzone.com/articles/what-are-meta-annotations-in-java had a very good example that explained to me what I wasn't picking up. 😅

Kazhuru:
This last commit makes this implementation of Display Generator able to support any extension of DisplayNameGenerator instead of just extending underscores by this way:
`@IndicativeSentencesGeneration(separator = " >> ", value = DisplayNameGenerator.ReplaceUnderscores.class)`

Kazhuru:
Also FYI, taking advantage of the fact that I already have the files well learned and how the flow works, I have a branch with the implementation about the CamelCase Display we wanted on the issue #1615 here is the link of my branch just in case: https://github.com/Kazhuru/junit5/commit/548379f365ae627b19e16d2525737814baf849ed.
PD: By doing this I realized that the way I was doing this issue only supported the underscore generator.
We can add this changes also in these pull request or in another new?

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -160,14 +160,57 @@ include::{testDir}/example/DisplayNameGeneratorDemo.java[tags=user_guide]
...
       +-- Year 2020 is a leap year. [OK]
       '-- Year 2048 is a leap year. [OK]
 ```
 
+[[writing-tests-display-name-generator-extensions]]
+==== Extensions of Display Name Generators
+The class DisplayNameGenerator already supports the extensions of `ReplaceUnderscores`
+and `IndicativeSentences`.
+
+By using the annotation that extends `Standard`:
```
marcphilipp: ❓ Did you mean this?

```suggestion
By using the annotation with a class that extends `Standard`:
```

documentation/src/test/java/example/DisplayNameGeneratorDemo.java
```diff
@@ -40,7 +39,7 @@ void if_it_is_negative(int year) {
 	}
 
 	@Nested
-	@DisplayNameGeneration(IndicativeSentences.class)
+	@DisplayNameGeneration(DisplayNameGenerator.IndicativeSentences.class)
```
marcphilipp: Please import `IndicativeSentences`.

documentation/src/test/java/example/DisplayNameIndicativeSentencesDemo.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.junit.jupiter.api.DisplayNameGenerator;
+import org.junit.jupiter.api.IndicativeSentencesGeneration;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class DisplayNameIndicativeSentencesDemo {
+
+	@Nested
+	@IndicativeSentencesGeneration(separator = " >> ", value = DisplayNameGenerator.ReplaceUnderscores.class)
```
marcphilipp: Please import `ReplaceUnderscores`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
```diff
@@ -0,0 +1,61 @@
...
+	 * @return custom separator for indicative sentences
+	 */
+	String separator() default "";
+
+	/**
+	 * Custom display name generator.
+	 *
+	 * @return custom display name generator class
+	 */
+	Class<? extends DisplayNameGenerator> value();
```
marcphilipp: ❓ I think we should rename this to `generator` and give it a default value.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
```diff
@@ -0,0 +1,61 @@
...
+ * {@code @IndicativeSentencesGeneration} is used to declare a custom separator
+ * by {@code IndicativeSentences}, if this notation was not declared
+ * will be use ", " as separator by default.
+ *
+ * @since 5.6
+ * @see DisplayName
+ * @see DisplayNameGenerator
+ */
+@DisplayNameGeneration(DisplayNameGenerator.IndicativeSentences.class)
+@Target(ElementType.TYPE)
```
marcphilipp: We should add `ElementType.ANNOTATION_TYPE` here to allow using it as a meta-annotation (and add a test for that).

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +143,97 @@ private static boolean hasParameters(Method method) {
 		}
 	}
 
+	/**
+	 * {@code DisplayNameGenerator} that generate complete sentences.
```
marcphilipp:
```suggestion
	 * {@code DisplayNameGenerator} that generates complete sentences.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +143,97 @@ private static boolean hasParameters(Method method) {
...
+	 *
+	 * <p>This extends the functionality of {@link ReplaceUnderscores} by
```
marcphilipp: This is no longer true.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +143,97 @@ private static boolean hasParameters(Method method) {
...
+		 *
+		 * @param currentClass the Test Class the separator either custom or default
+		 * @return the indicative sentence separator
+		 * {@code Class.getName()}.
+		 */
+		private String getSentenceSeparator(Class<?> currentClass) {
+			Optional<IndicativeSentencesGeneration> indicativeSentencesGeneration = AnnotationSupport.findAnnotation(
+				currentClass, IndicativeSentencesGeneration.class);
+
+			if (indicativeSentencesGeneration.isPresent())
```
marcphilipp: Please add braces to all `if`/`else` statements.

Kazhuru:
Thanks for the several times of code reviews @marcphilipp and everyone, I'm going to check the changes and resolve them in a next commit. 😄

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
```diff
@@ -0,0 +1,61 @@
...
+@Inherited
+@API(status = EXPERIMENTAL, since = "5.6")
+public @interface IndicativeSentencesGeneration {
+
+	String DEFAULT_SEPARATOR = ", ";
+	/**
+	 * Pre-defined display name generator instances.
+	 */
+	DisplayNameGenerator standardGenerator = new DisplayNameGenerator.Standard();
+	DisplayNameGenerator replaceUnderscoresGenerator = new DisplayNameGenerator.ReplaceUnderscores();
```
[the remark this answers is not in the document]
Kazhuru: Sure, I put those instances in the interface because it was needed for the indicative sentences and I did not was sure in which part of the code should be, thanks for the tips

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +143,97 @@ private static boolean hasParameters(Method method) {
...
+			return IndicativeSentencesGeneration.DEFAULT_SEPARATOR;
+		}
+
+		private DisplayNameGenerator getDisplayNameGenerator(Class<?> currentClass) {
+			Optional<IndicativeSentencesGeneration> indicativeSentencesGeneration = AnnotationSupport.findAnnotation(
+				currentClass, IndicativeSentencesGeneration.class);
+
+			if (indicativeSentencesGeneration.isPresent()) {
+				if (indicativeSentencesGeneration.get().value() == ReplaceUnderscores.class)
+					return IndicativeSentencesGeneration.replaceUnderscoresGenerator;
```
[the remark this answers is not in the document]
Kazhuru: Moved as a static method in `DisplayNameGenerator`

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
Kazhuru: Yes, with this changes I'll update that and value for generator in the documentation

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
Kazhuru: Sounds good to me!, more expressive than value 😅

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Thanks for the typo correction, I'll be more careful on that

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Indeed, checking all the java-docs and documentation to make sure everything is up to date

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: On it. 👍

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
Kazhuru: Oh didn't know, I'll try to do it, thanks for the feedback!

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -80,6 +81,24 @@ static String parameterTypesAsString(Method method) {
...
+	 * @return description
+	 */
+	static DisplayNameGenerator getDisplayNameGenerator(Class<?> testClass) {
+
+		if (testClass == Standard.class) {
+			return Standard.standardGenerator;
+		}
+		if (testClass == ReplaceUnderscores.class) {
+			return ReplaceUnderscores.replaceUnderscoresGenerator;
+		}
```
marcphilipp: `IndicativeSentences` is missing here, isn't it?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -88,6 +107,11 @@ static String parameterTypesAsString(Method method) {
 	 */
 	class Standard implements DisplayNameGenerator {
 
+		/**
+		 * Pre-defined standard display name generator instance.
+		 */
+		static final DisplayNameGenerator standardGenerator = new Standard();
```
marcphilipp: Let's just call it `INSTANCE` (same below).

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: You're right since it was not used I forgot to add that, on it!

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Roger, lees repetitive when using it.

Kazhuru:
Sorry for the delayed changes, all of them on those last 2 commits, the CI is failing because my spotlessApply is outdate with rebase upstream everything will be green again.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
juliette-derancourt: Isn't this logic already implemented [here](https://github.com/Kazhuru/junit5/blob/issue/1596-indicative-sentences/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java#L102-L111)? Maybe it could be made a public method of `DisplayNameUtils`?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +168,131 @@ private static boolean hasParameters(Method method) {
...
+		private String classReplaceToIndicativeSentence(Class<?> testClass) {
+			Class<?> enclosingParent = testClass.getEnclosingClass();
+			Optional<DisplayName> displayName = AnnotationSupport.findAnnotation(testClass, DisplayName.class);
+			Optional<DisplayNameGeneration> displayNameGeneration = AnnotationSupport.findAnnotation(testClass,
+				DisplayNameGeneration.class);
+
+			if (enclosingParent == null) {
+				return displayName.map(DisplayName::value).orElseGet(() -> generateDisplayNameForClass(testClass));
+			}
+			else {
```
juliette-derancourt: I don't know if I already mentioned that, but personally I would get rid of both `else` blocks, since a value has already been returned in the `if` block. I feel that in most cases it makes the code more readable. What do you think?

@junit-team/junit-lambda Is there some kind of tacit style rule about this? 😛

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -141,4 +168,131 @@ private static boolean hasParameters(Method method) {
...
+		 * annotation from {@code IndicativeSentencesGeneration}, if it doesn't find it,
+		 * then search for the parent classes, if no separator is found use @code{", "} by default.
+		 *
+		 * @param testClass Class to get Indicative sentence annotation separator either custom or default
+		 * @return the indicative sentence separator
+		 */
+		private String getSentenceSeparator(Class<?> testClass) {
+			Optional<IndicativeSentencesGeneration> indicativeSentencesGeneration = getIndicativeSentencesGeneration(
+				testClass);
+			if (indicativeSentencesGeneration.isPresent()) {
```
juliette-derancourt: Same kind of thought here, I would negate this `if` condition and get rid of the `else` block... but I guess it's a matter of taste really ^^

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
```diff
@@ -0,0 +1,57 @@
...
+
+	String DEFAULT_SEPARATOR = ", ";
+	Class<? extends DisplayNameGenerator> DEFAULT_GENERATOR = DisplayNameGenerator.Standard.class;
+
+	/**
+	 * Custom separator for indicative sentences generator.
+	 *
+	 * @return custom separator for indicative sentences
+	 */
+	String separator() default "";
```
juliette-derancourt: Why don't we default to `", "` here?

documentation/src/test/java/example/DisplayNameIndicativeSentencesDemo.java
```diff
@@ -0,0 +1,45 @@
...
+import org.junit.jupiter.api.DisplayNameGenerator.ReplaceUnderscores;
+import org.junit.jupiter.api.IndicativeSentencesGeneration;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class DisplayNameIndicativeSentencesDemo {
+
+	@Nested
+	@IndicativeSentencesGeneration(separator = " >> ", generator = ReplaceUnderscores.class)
+	class A_cat_named_schrodinger {
```
juliette-derancourt: Maybe more like `Schrodinger_s_cat`? 😁

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -160,14 +160,60 @@ include::{testDir}/example/DisplayNameGeneratorDemo.java[tags=user_guide]
   | | +-- For example, year -1 is not supported. [OK]
   | | '-- For example, year -4 is not supported. [OK]
   | '-- if it is zero() [OK]
-  '-- A year is a leap year... [OK]
```
juliette-derancourt: Maybe this ellipsis at the end of the class name wasn't such a bad idea? 🤔

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Yeah! these class has 2 methods that I was considering to use instead of the actual new ones, it took me some a time this Friday wondering a good way to solve these needs (IMHO after Marc's feedback), 

To adjust the methods and accessibility of `DisplayNameUtils` `getDisplayNameGenerator` needed a `JupiterConfiguration` parameter because this method is more general and uses the default Display Name Generator setting, I'm not so sure how to pass that configuration or make a new one and where.

For the method `getDisplayNameGeneration` also useful because we needed to find the IndicativeSentencesGeneration Annotation and it has the class IndicativeSentences,  but it only return the first DisplayNameGenerator annotation not the annotation and the parameter information that were needed.

I thought these details were too specific to start modifying methods where other classes depended on them so I decided to do it this way. Anyway if you feel that this approach falls into the problem of duplicate code, I can try to do it using the `DisplayNameUtils` class maybe with an override to the `getDisplayNameGenerator` method.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
juliette-derancourt: It seems to be done [here](https://github.com/Kazhuru/junit5/blob/0c98ec8d3e030062020ca6550e59274fdf338309/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java#L239) anyway

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -78,6 +81,26 @@ static String parameterTypesAsString(Method method) {
...
+		if (testClass == Standard.class) {
+			return Standard.INSTANCE;
+		}
+		if (testClass == ReplaceUnderscores.class) {
+			return ReplaceUnderscores.INSTANCE;
+		}
+		if (testClass == IndicativeSentences.class) {
+			return IndicativeSentences.INSTANCE;
+		}
+		return (DisplayNameGenerator) ReflectionUtils.newInstance(testClass);
```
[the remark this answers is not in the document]
Kazhuru: Yeah I also thought about it, I did not know about that, I'll try to expand that precondition to make the method safer.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -113,9 +138,11 @@ public String generateDisplayNameForMethod(Class<?> testClass, Method testMethod
 	 */
 	class ReplaceUnderscores extends Standard {
 
+		static final DisplayNameGenerator INSTANCE = new ReplaceUnderscores();
+
 		@Override
 		public String generateDisplayNameForClass(Class<?> testClass) {
-			return replaceUnderscores(super.generateDisplayNameForClass(testClass));
+			return replaceUnderscores(testClass.getSimpleName());
```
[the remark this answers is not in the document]
Kazhuru: Nop this was a very interesting change indeed that I asked in glitter, 
when generating the report of a test with `replaceUnderscores` : 

	@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)
	static class StackTestCase {

		Stack<Object> stack;

		@Test
		void new_stack() {
			new Stack<>();
		}
        [...]
     }
I got an output like this: 
evidence

I don't know if this is a desired output with the class parents with a $, but it seems odd to read that if the goal is making a result pretty to read. 

All of this is because`generateDisplayNameForClass` in Standard uses `class.getName()` in `generateDisplayNameForClass` and then strips the packages name with a `LastIndexOf('.')`, for inherence `super` also Underscores had the same output issue and then also Indicative Sentences because it to concatenate the results of the selected Generator.

I tried to make this change on Standard but to many test has this kind of result _ContainerClass$MyClassTest_ expected, so decided to let Standard  be and make this change.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
Kazhuru: I love how you hit the nail on the head with ideas that I thought for a while 🤓  with no so transparent decisions, I left "" as default because the recommendations on this comment:
https://github.com/junit-team/junit5/pull/2124#discussion_r359873426.
But yeah as you say in that section im checking for an empty string of separator and then using the default static variable.
Anyway I can change that quick for sure if thats better.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
Kazhuru: Do you mean the 3 dots?

documentation/src/test/java/example/DisplayNameIndicativeSentencesDemo.java
Kazhuru: haha yeah, sure!

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Sure I didn't think about it, I already make this changes on mi branch and look indeed more readable, thanks this changes comes in the next commit.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
Kazhuru: Sure same as my last answer, thanks for the tips. 🙌

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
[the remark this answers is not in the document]
juliette-derancourt: I think you can use [something like this](https://stackoverflow.com/a/4584552) ;)

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
juliette-derancourt: Yes!

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
juliette-derancourt: Ahah, telepathy? xD
I think the initial remark was more about the fact that we need to set a default value, not necessarily about the value itself
@marcphilipp what do you think?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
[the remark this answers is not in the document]
juliette-derancourt: I see what you mean

@sormuras Was there a reason to reuse the `Standard` implementation in `ReplaceUnderscores`? Do you think this change would make sense?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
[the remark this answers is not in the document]
Kazhuru: Oh, currently I made this change by this annotation:
```
Preconditions.condition(testClass.getEnclosingClass() == DisplayNameGenerator.class,
			"Class must be a DisplayNameGenerator implementation");
```
But yeah, I'll change that '==' to isAssignableFrom() following your suggestion :D

mmerdes:
@marcphilipp 
Please take another look at the review.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
marcphilipp: I think I was thinking about the generator being used without the annotation. If we require the use of the annotation using the default value of the annotation attribute is fine with me.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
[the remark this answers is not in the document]
marcphilipp: @Kazhuru Are you going to make this change?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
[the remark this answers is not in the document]
sormuras: The idea was to keep the "relative path to the package" of a test class intact. Using `getSimpleName()` may result in many similar test names.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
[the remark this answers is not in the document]
Kazhuru: Yeah @marcphilipp , sorry for the long time without giving signs of life these last months :(, I got into a lot of work and personal stuff and they left me little time to check the progress of the feature, I'm going to make this change in a next commit and the other details that were commented below.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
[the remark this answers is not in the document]
Kazhuru: All right I'll keep the result of the generators with the relative path of the package, and the proper changes on the test's.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/IndicativeSentencesGeneration.java
Kazhuru: Oh I see now anyway this default value was not used because it later is replaced, roger on that, then to set the value as default here! 👍

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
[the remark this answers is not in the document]
marcphilipp: No worries, we've all been there. 🙂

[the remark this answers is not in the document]
marcphilipp: @Kazhuru Any updates?

juliette-derancourt:
Hey @Kazhuru! Hope you're doing well :)
Would you mind if I finished the last details on this PR in order to merge it?

Kazhuru:
Hi @juliette-derancourt, I'm sorry I've been so ausent for so long and the reply so late, I've been delaying the changes for personal problems with the quarantine now, go ahead and do the details you need to make the merge, thanks and I hope you're well too.

juliette-derancourt:
No worries! The situation is not the greatest indeed... =/
Good luck, hope it will get better soon! :)

juliette-derancourt:
Merged in d3e901aadf7548a3b72a2ccfa97391711b182065, thanks @Kazhuru!

PR #2175, opened 2020-02-06
Simplified Kotlin Assertions

Fleshgrinder:
## Overview

- Inlined `ExecutableStream` and `ExecutableCollection` type aliases since each is only used in two places that don‘t warrant dedicated type aliases. There is also [KT-24700](https://youtrack.jetbrains.com/issue/KT-24700) that makes private type aliases kind of dangerous.
- Removed the two map extension functions and instead inlined it at a single spot, all other functions are now calling that function and there is no need for the additional symbols anymore.
- Removed the `toList()` calls and exchanged them with `Arrays.stream()`, there is really no need to construct a list first.
- Moved the examples to Kotlin files and made them actual tests. This ensures that the code from the examples stays up-to-date and is correct since they are treated like normal tests. In fact, I fixed some samples that where using `Duration.seconds` which actually needs to be `Duration.ofSeconds`.

We could make every trivial, forwarding function `inline` but I didn‘t do it right away because I would first like to know if there are any good reasons not to. I will leave a comment on every function that I believe should be inlined. Instead of making them `inline` we could also drop them altogether since some of them are just forwarding to their corresponding Java functions. Instead of redeclaring the whole Java API in Kotlin we could just start annotating the Java code with the [JetBrains annotations](https://github.com/JetBrains/java-annotations), the result would be the same but the experience would not only be improved for Kotlin users but for Java users as well (given they use IntelliJ).

Moving the samples to actual tests means that I had to enable tests in the API module. I am not sure if this is a good thing. Let me know what you think.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/junit-jupiter-api.gradle.kts
+++ b/junit-jupiter-api/junit-jupiter-api.gradle.kts
@@ -14,4 +14,11 @@ dependencies {
 	api(project(":junit-platform-commons"))
 
 	compileOnly("org.jetbrains.kotlin:kotlin-stdlib")
+
+	testImplementation("org.jetbrains.kotlin:kotlin-stdlib")
+	testRuntimeOnly(project(":junit-jupiter-engine"))
+}
+
+tasks.test {
+	useJUnitPlatform()
 }
--- a/junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
+++ b/junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
@@ -12,13 +12,13 @@
 package org.junit.jupiter.api
 
 import java.time.Duration
+import java.util.Arrays
 import java.util.function.Supplier
 import java.util.stream.Stream
 import org.apiguardian.api.API
 import org.apiguardian.api.API.Status.EXPERIMENTAL
 import org.junit.jupiter.api.function.Executable
 import org.junit.jupiter.api.function.ThrowingSupplier
-import java.util.Arrays
 
 /**
  * @see Assertions.fail
--- a/junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
+++ b/junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
@@ -18,119 +18,85 @@ import org.apiguardian.api.API
 import org.apiguardian.api.API.Status.EXPERIMENTAL
 import org.junit.jupiter.api.function.Executable
 import org.junit.jupiter.api.function.ThrowingSupplier
+import java.util.Arrays
 
 /**
  * @see Assertions.fail
  */
 fun fail(message: String?, throwable: Throwable? = null): Nothing =
-    Assertions.fail<Nothing>(message, throwable)
+    Assertions.fail(message, throwable)
 
 /**
  * @see Assertions.fail
  */
 fun fail(message: (() -> String)?): Nothing =
-    Assertions.fail<Nothing>(message)
+    Assertions.fail(message)
 
 /**
  * @see Assertions.fail
  */
 fun fail(throwable: Throwable?): Nothing =
-    Assertions.fail<Nothing>(throwable)
-
-/**
- * [Stream] of functions to be executed.
- */
-private typealias ExecutableStream = Stream<() -> Unit>
-private fun ExecutableStream.convert() = map { Executable(it) }
+    Assertions.fail(throwable)
 
 /**
  * @see Assertions.assertAll
  */
-fun assertAll(executables: ExecutableStream) =
-    Assertions.assertAll(executables.convert())
+fun assertAll(executables: Stream<() -> Unit>) =
+    assertAll(null, executables)
 
 /**
  * @see Assertions.assertAll
  */
-fun assertAll(heading: String?, executables: ExecutableStream) =
-    Assertions.assertAll(heading, executables.convert())
-
-/**
- * [Collection] of functions to be executed.
- */
-private typealias ExecutableCollection = Collection<() -> Unit>
-private fun ExecutableCollection.convert() = map { Executable(it) }
+fun assertAll(heading: String?, executables: Stream<() -> Unit>) =
+    Assertions.assertAll(heading, executables.map { Executable(it) })
 
 /**
  * @see Assertions.assertAll
  */
-fun assertAll(executables: ExecutableCollection) =
-    Assertions.assertAll(executables.convert())
+fun assertAll(executables: Collection<() -> Unit>) =
+    assertAll(executables.stream())
 
 /**
  * @see Assertions.assertAll
  */
-fun assertAll(heading: String?, executables: ExecutableCollection) =
-    Assertions.assertAll(heading, executables.convert())
+fun assertAll(heading: String?, executables: Collection<() -> Unit>) =
+    assertAll(heading, executables.stream())
 
 /**
  * @see Assertions.assertAll
  */
 fun assertAll(vararg executables: () -> Unit) =
-    assertAll(executables.toList().stream())
+    assertAll(Arrays.stream(executables))
 
 /**
  * @see Assertions.assertAll
  */
 fun assertAll(heading: String?, vararg executables: () -> Unit) =
-    assertAll(heading, executables.toList().stream())
+    assertAll(heading, Arrays.stream(executables))
 
 /**
- * Example usage:
- * ```kotlin
- * val exception = assertThrows<IllegalArgumentException> {
- *     throw IllegalArgumentException("Talk to a duck")
- * }
- * assertEquals("Talk to a duck", exception.message)
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertThrowsSample
  * @see Assertions.assertThrows
  */
 inline fun <reified T : Throwable> assertThrows(noinline executable: () -> Unit): T =
     Assertions.assertThrows(T::class.java, Executable(executable))
 
 /**
- * Example usage:
- * ```kotlin
- * val exception = assertThrows<IllegalArgumentException>("Should throw an Exception") {
- *     throw IllegalArgumentException("Talk to a duck")
- * }
- * assertEquals("Talk to a duck", exception.message)
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertThrowsWithMessageSample
  * @see Assertions.assertThrows
  */
 inline fun <reified T : Throwable> assertThrows(message: String, noinline executable: () -> Unit): T =
     assertThrows({ message }, executable)
 
 /**
- * Example usage:
- * ```kotlin
- * val exception = assertThrows<IllegalArgumentException>({ "Should throw an Exception" }) {
- *     throw IllegalArgumentException("Talk to a duck")
- * }
- * assertEquals("Talk to a duck", exception.message)
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertThrowsWithLazyMessageSample
  * @see Assertions.assertThrows
  */
 inline fun <reified T : Throwable> assertThrows(noinline message: () -> String, noinline executable: () -> Unit): T =
     Assertions.assertThrows(T::class.java, Executable(executable), Supplier(message))
 
 /**
- * Example usage:
- * ```kotlin
- * val result = assertDoesNotThrow {
- *     // Code block that is expected to not throw an exception
- * }
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertDoesNotThrowSample
  * @see Assertions.assertDoesNotThrow
  * @param R the result type of the [executable]
  */
@@ -139,12 +105,7 @@ fun <R> assertDoesNotThrow(executable: () -> R): R =
     Assertions.assertDoesNotThrow(ThrowingSupplier(executable))
 
 /**
- * Example usage:
- * ```kotlin
- * val result = assertDoesNotThrow("Should not throw an exception") {
- *     // Code block that is expected to not throw an exception
- * }
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertDoesNotThrowWithMessageSample
  * @see Assertions.assertDoesNotThrow
  * @param R the result type of the [executable]
  */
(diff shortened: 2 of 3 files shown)


junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -18,119 +18,85 @@ import org.apiguardian.api.API
 import org.apiguardian.api.API.Status.EXPERIMENTAL
 import org.junit.jupiter.api.function.Executable
 import org.junit.jupiter.api.function.ThrowingSupplier
+import java.util.Arrays
 
 /**
  * @see Assertions.fail
  */
 fun fail(message: String?, throwable: Throwable? = null): Nothing =
```
Fleshgrinder:
```suggestion
inline fun fail(message: String?, throwable: Throwable? = null): Nothing =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -18,119 +18,85 @@ import org.apiguardian.api.API
...
-    Assertions.fail<Nothing>(message, throwable)
+    Assertions.fail(message, throwable)
 
 /**
  * @see Assertions.fail
  */
 fun fail(message: (() -> String)?): Nothing =
```
Fleshgrinder:
```suggestion
inline fun fail(noinline message: (() -> String)?): Nothing =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -18,119 +18,85 @@ import org.apiguardian.api.API
...
-    Assertions.fail<Nothing>(message)
+    Assertions.fail(message)
 
 /**
  * @see Assertions.fail
  */
 fun fail(throwable: Throwable?): Nothing =
```
Fleshgrinder:
```suggestion
inline fun fail(throwable: Throwable?): Nothing =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -18,119 +18,85 @@ import org.apiguardian.api.API
...
-private fun ExecutableCollection.convert() = map { Executable(it) }
+fun assertAll(heading: String?, executables: Stream<() -> Unit>) =
+    Assertions.assertAll(heading, executables.map { Executable(it) })
 
 /**
  * @see Assertions.assertAll
  */
-fun assertAll(executables: ExecutableCollection) =
-    Assertions.assertAll(executables.convert())
+fun assertAll(executables: Collection<() -> Unit>) =
```
Fleshgrinder:
```suggestion
inline fun assertAll(executables: Collection<() -> Unit>) =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -18,119 +18,85 @@ import org.apiguardian.api.API
...
+    assertAll(executables.stream())
 
 /**
  * @see Assertions.assertAll
  */
-fun assertAll(heading: String?, executables: ExecutableCollection) =
-    Assertions.assertAll(heading, executables.convert())
+fun assertAll(heading: String?, executables: Collection<() -> Unit>) =
```
Fleshgrinder:
```suggestion
inline fun assertAll(heading: String?, executables: Collection<() -> Unit>) =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -18,119 +18,85 @@ import org.apiguardian.api.API
...
+    assertAll(heading, executables.stream())
 
 /**
  * @see Assertions.assertAll
  */
 fun assertAll(vararg executables: () -> Unit) =
```
Fleshgrinder:
```suggestion
inline fun assertAll(vararg executables: () -> Unit) =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -18,119 +18,85 @@ import org.apiguardian.api.API
...
-    assertAll(executables.toList().stream())
+    assertAll(Arrays.stream(executables))
 
 /**
  * @see Assertions.assertAll
  */
 fun assertAll(heading: String?, vararg executables: () -> Unit) =
```
Fleshgrinder:
```suggestion
inline fun assertAll(heading: String?, vararg executables: () -> Unit) =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -139,12 +105,7 @@ fun <R> assertDoesNotThrow(executable: () -> R): R =
     Assertions.assertDoesNotThrow(ThrowingSupplier(executable))
```
Fleshgrinder: `inline`

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -153,12 +114,7 @@ fun <R> assertDoesNotThrow(message: String, executable: () -> R): R =
     assertDoesNotThrow({ message }, executable)
```
Fleshgrinder: `inline`

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -167,84 +123,54 @@ fun <R> assertDoesNotThrow(message: () -> String, executable: () -> R): R =
     Assertions.assertDoesNotThrow(ThrowingSupplier(executable), Supplier(message))
```
Fleshgrinder: `inline`

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -167,84 +123,54 @@ fun <R> assertDoesNotThrow(message: () -> String, executable: () -> R): R =
...
- *     // Code block that is being timed.
- * }
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertTimeoutSample
  * @see Assertions.assertTimeout
- * @paramR the result of the [executable].
+ * @param R the result of the [executable].
  */
 @API(status = EXPERIMENTAL, since = "5.5")
 fun <R> assertTimeout(timeout: Duration, executable: () -> R): R =
```
Fleshgrinder:
```suggestion
inline fun <R> assertTimeout(timeout: Duration, executable: () -> R): R =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -167,84 +123,54 @@ fun <R> assertDoesNotThrow(message: () -> String, executable: () -> R): R =
...
- *     // Code block that is being timed.
- * }
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertTimeoutWithMessageSample
  * @see Assertions.assertTimeout
- * @paramR the result of the [executable].
+ * @param R the result of the [executable].
  */
 @API(status = EXPERIMENTAL, since = "5.5")
 fun <R> assertTimeout(timeout: Duration, message: String, executable: () -> R): R =
```
Fleshgrinder:
```suggestion
inline fun <R> assertTimeout(timeout: Duration, message: String, executable: () -> R): R =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -167,84 +123,54 @@ fun <R> assertDoesNotThrow(message: () -> String, executable: () -> R): R =
...
- *     // Code block that is being timed.
- * }
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertTimeoutWithLazyMessageSample
  * @see Assertions.assertTimeout
- * @paramR the result of the [executable].
+ * @param R the result of the [executable].
  */
 @API(status = EXPERIMENTAL, since = "5.5")
 fun <R> assertTimeout(timeout: Duration, message: () -> String, executable: () -> R): R =
```
Fleshgrinder:
```suggestion
inline fun <R> assertTimeout(timeout: Duration, message: () -> String, executable: () -> R): R =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -167,84 +123,54 @@ fun <R> assertDoesNotThrow(message: () -> String, executable: () -> R): R =
...
- *     // Code block that is being timed.
- * }
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertTimeoutPreemptivelySample
  * @see Assertions.assertTimeoutPreemptively
- * @paramR the result of the [executable].
+ * @param R the result of the [executable].
  */
 @API(status = EXPERIMENTAL, since = "5.5")
 fun <R> assertTimeoutPreemptively(timeout: Duration, executable: () -> R): R =
```
Fleshgrinder:
```suggestion
inline fun <R> assertTimeoutPreemptively(timeout: Duration, executable: () -> R): R =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -167,84 +123,54 @@ fun <R> assertDoesNotThrow(message: () -> String, executable: () -> R): R =
...
- *     // Code block that is being timed.
- * }
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertTimeoutPreemptivelyWithMessageSample
  * @see Assertions.assertTimeoutPreemptively
- * @paramR the result of the [executable].
+ * @param R the result of the [executable].
  */
 @API(status = EXPERIMENTAL, since = "5.5")
 fun <R> assertTimeoutPreemptively(timeout: Duration, message: String, executable: () -> R): R =
```
Fleshgrinder:
```suggestion
inline fun <R> assertTimeoutPreemptively(timeout: Duration, message: String, executable: () -> R): R =
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -167,84 +123,54 @@ fun <R> assertDoesNotThrow(message: () -> String, executable: () -> R): R =
...
- *     // Code block that is being timed.
- * }
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertTimeoutPreemptivelyWithLazyMessageSample
  * @see Assertions.assertTimeoutPreemptively
- * @paramR the result of the [executable].
+ * @param R the result of the [executable].
  */
 @API(status = EXPERIMENTAL, since = "5.5")
 fun <R> assertTimeoutPreemptively(timeout: Duration, message: () -> String, executable: () -> R): R =
```
Fleshgrinder:
```suggestion
inline fun <R> assertTimeoutPreemptively(timeout: Duration, message: () -> String, executable: () -> R): R =
```

junit-jupiter-api/junit-jupiter-api.gradle.kts
```diff
@@ -14,4 +14,11 @@ dependencies {
 	api(project(":junit-platform-commons"))
 
 	compileOnly("org.jetbrains.kotlin:kotlin-stdlib")
+
+	testImplementation("org.jetbrains.kotlin:kotlin-stdlib")
+	testRuntimeOnly(project(":junit-jupiter-engine"))
```
sbrannen: The `junit-jupiter-api` project is not allowed to contain tests, since that introduces a classpath cycle in certain IDEs (e.g., Eclipse).

All tests for `junit-jupiter-api` reside in `junit-jupiter-engine`, so please undo this change and move the tests to `junit-jupiter-engine`.

Fleshgrinder: I assumed as much and that's also why I asked for feedback regarding this. That said, these are not really tests but examples and they must be accessible for the `junit-jupiter-api` module or they cannot be referenced in the documentation blocks. I see a few possible ways to go about this:

1. Live with the fact that the examples might get outdated or are outright wrong because there is no validation (status quo).
1. Make the test classpath of `junit-jupiter-engine` available at compile time to `junit-jupiter-api` (does not really resolve the cycle issue).
1. Do what I have done here which works because `compile` happens before `compileTest` and there is thus no real cycle.
1. Instead of using `prject(":junit-jupiter-engine")` for testing of the samples we use a previously released version of JUnit. This should not be a problem because the engine does not provide anything on top of `junit-jupiter-api` that is required in the samples and which would lead to issues since the samples target only `junit-jupiter-api`.

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -23,114 +24,79 @@ import org.junit.jupiter.api.function.ThrowingSupplier
...
 
 /**
- * Example usage:
- * ```kotlin
- * val exception = assertThrows<IllegalArgumentException> {
- *     throw IllegalArgumentException("Talk to a duck")
- * }
- * assertEquals("Talk to a duck", exception.message)
- * ```
+ * @sample org.junit.jupiter.api.AssertionsSamples.assertThrowsSample
```
marcphilipp: Does the Kotlin compiler validate these? It can‘t see test classes when compiling these, though, can it?

junit-jupiter-api/junit-jupiter-api.gradle.kts
marcphilipp: How about we put them into the `testFixtures` and add a test to `platform-tests` that runs them?

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
Fleshgrinder: I don't know what happens under the hood but can find out.

junit-jupiter-api/junit-jupiter-api.gradle.kts
Fleshgrinder: That sounds good. 👍

Fleshgrinder:
Active

PR #2177, opened 2020-02-08
Added programmatic test builder API

Fleshgrinder:
## Overview

> This is a draft because the PR is still missing crucial parts like tests and documentation. However, the implementation itself is mostly finished and I would like to gather feedback. 😊

Using `junit-jupiter-params` in Kotlin is awkward for many reasons and why I propose the addition of a programmatic, functional API to JUnit that will make `junit-jupiter-params` obsolete for many use-cases.

```kotlin
private class JunitJupiterParamsKt {
    /** @see dataProvider */
    @ParameterizedTest
    @MethodSource("dataProvider")
    fun isPalindrom(candidate: String) {
        assertEquals(candidate.reversed(), candidate)
    }

    companion object {
        @JvmStatic
        fun dataProvider(): Stream<Arguments> =
            Stream.of(
                arguments("mum"),
                arguments("dad")
            )
    }
}
```

Above is an extremely simple test written in Kotlin that makes use of `junit-jupiter-params` but there are many issues with it (ordered from top to bottom as they occur in the file above):

- We have to add the kdoc block with an `@see` reference to the actual method source to please IntelliJ and stop insisting on `dataProvider` being unused (this is a total IDE issue that can be solved but it is a real issue).
- We need to add two annotations to our function to enable the parameterization.
- We need to add an annotation with a string value matching the name of the data provider function, keeping it up-to-date manually.
- We need to define the signature as we expect things to arrive, however, there are no compile-time checks that actually ensure that the arguments of the data provider are actually compatible with our signature. We might have been running tests for five minutes already until we run into the issue at runtime.
- We need to create a companion object where we collect all data providers. This means that the data providers will end up far away from the spot where they are actually being used. Forcing us to scroll and jump around to find everything that makes up a test case.
  
  There is the ability to change the test instance lifecycle to class that lifts this requirement. However, many users do not know about it and I also had to learn that tests can be written in ways that make it impossible to switch.
- We need to annotate each data provider with `@JvmStatic`.
- We need to add `Stream<Arguments>` as the return type (or add `!!` at the end of the stream construction and each `arguments` call) to disable warnings about missing nullability information.
- We need to use `Stream` which is not very idiomatic in Kotlin.

Here is the very same parameterized test written with the programmatic API that I am proposing:

```kotlin
private class JunitJupiterTestBuilderKt {
    @TestFactory fun isPalindrom() =
        testOf(case("mum"), case("dad")) { 
            assertEquals(it.reversed(), it)
        }
}
```

It solves all mentioned issues, better yet, it is also much faster than the version from above because there is no runtime discovery necessary and nothing needs to be wired by JUnit since its all done by the language. We can have all these benefits in Java too:

```java
class JunitJupiterTestBuilderJ {
    @TestFactory
    TestCases isPalindrom() {
        return testOf(
            it -> assertTrue(StringUtils.isPalindrom(it)),
            caseOf("mum"),
            caseOf("dad")
        );
    }
}
```

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/function/ThrowingConsumers.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/function/ThrowingConsumers.java
@@ -7,11 +7,12 @@
  *
  * https://www.eclipse.org/legal/epl-v20.html
  */
+
 package org.junit.jupiter.api.function;
 
-import org.apiguardian.api.API;
-
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import org.apiguardian.api.API;
 
 /**
  * {@code ThrowingConsumers} contains additional {@link ThrowingConsumer}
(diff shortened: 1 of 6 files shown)


marcphilipp:
I agree with most of the issues you describe above and using `@TestFactory` is an alternative that we provided on purpose. If only you could use method references in annotations…

Most of the issues you list only apply when using `@MethodSource`. You could alternatively implement `ArgumentsProvider` and use `@ArgumentSource`.

I‘m not yet convinced we should add this DSL to core JUnit.

Fleshgrinder:
> I‘m not yet convinced we should add this DSL to core JUnit.

Why not and what would convince you?

marcphilipp:
The problem with `@TestFactory` vs. `@ParameterizedTest` is that there‘s no support for lifecycle methods or extensions. While your type safety argument is certainly valid, you‘d also lose functionality. Moreover, everything we add to the core will need to be maintained. So I‘d like to see there’s considerable interest before adding it.

Fleshgrinder:
I don't consider this an issue because I'm not proposing to drop parameterized tests. However, I do believe that this simplified DSL serves most use cases better and requires less mental power plus docs.

Sounds good.

Fleshgrinder:
I withdraw the PR, don't think it will ever make it and since the bot is going to close it anyways soon...

jbduncan:
Worst case scenario, my experience is that Kotest is generally a better testing framework for Kotlin users than JUnit 4 or 5, assuming you don't need any JUnit-specific extensions. :)

PR #2201, opened 2020-02-27
Introduce new `DisplayNameGenerator` named `Simple`

### LINKED ISSUE #1961: Default display name for methods appends empty parentheses
In JUnit Jupiter 5.5.1 using Kotlin 1.3.41:

```kotlin
@Nested
inner class `assign to group` {
    @Test
    fun `failure -- stop`() {
    }
    @Test
    fun assigned() {
    }
}
```

Results in the following display names:

```
assign to group
  failure -- stop()
  assigned()
```

While I expect it without parentheses:

```
assign to group
  failure -- stop
  assigned
```

juliette-derancourt:
This new generator extends `Standard` and removes parenthesis
at the end of method names if they have no parameters.

Resolves #1961.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
@@ -105,13 +105,38 @@ public interface DisplayNameGenerator {
 	}
 
 	/**
-	 * {@code DisplayNameGenerator} that replaces underscores with spaces.
+	 * Simple {@code DisplayNameGenerator} that removes trailing parentheses
+	 * for methods with no parameters.
 	 *
 	 * <p>This generator extends the functionality of {@link Standard} by
+	 * removing parentheses ({@code '()'}) found at the end of method names
+	 * with no parameters.
+	 */
+	class Simple extends Standard {
+
+		@Override
+		public String generateDisplayNameForMethod(Class<?> testClass, Method testMethod) {
+			String displayName = testMethod.getName();
+			if (hasParameters(testMethod)) {
+				displayName += ' ' + parameterTypesAsString(testMethod);
+			}
+			return displayName;
+		}
+
+		private static boolean hasParameters(Method method) {
+			return method.getParameterCount() > 0;
+		}
+
+	}
+
+	/**
+	 * {@code DisplayNameGenerator} that replaces underscores with spaces.
+	 *
+	 * <p>This generator extends the functionality of {@link Simple} by
 	 * replacing all underscores ({@code '_'}) found in class and method names
 	 * with spaces ({@code ' '}).
 	 */
-	class ReplaceUnderscores extends Standard {
+	class ReplaceUnderscores extends Simple {
 
 		@Override
 		public String generateDisplayNameForClass(Class<?> testClass) {
@@ -125,20 +150,13 @@ public interface DisplayNameGenerator {
 
 		@Override
 		public String generateDisplayNameForMethod(Class<?> testClass, Method testMethod) {
-			String displayName = replaceUnderscores(testMethod.getName());
-			if (hasParameters(testMethod)) {
-				displayName += ' ' + parameterTypesAsString(testMethod);
-			}
-			return displayName;
+			return replaceUnderscores(super.generateDisplayNameForMethod(testClass, testMethod));
 		}
 
 		private static String replaceUnderscores(String name) {
 			return name.replace('_', ' ');
 		}
 
-		private static boolean hasParameters(Method method) {
-			return method.getParameterCount() > 0;
-		}
 	}
 
 }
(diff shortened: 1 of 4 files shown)


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
```diff
@@ -96,6 +102,9 @@ private static DisplayNameGenerator getDisplayNameGenerator(Class<?> testClass,
 					if (displayNameGeneratorClass == Standard.class) {
 						return standardGenerator;
 					}
+					if (displayNameGeneratorClass == Simple.class) {
```
aepfli: not sure if appropriate, but if the list of displaynameGenerators is extending further, it might be easier to be read as a switch clause - but that might also be just subjective feeling.

aepfli:
lgtm ;)

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
juliette-derancourt: I was gonna say good idea, but unfortunatly we can't use `switch` with `Class` objects... =/

juliette-derancourt:
Thanks for the review @aepfli! ;)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -105,13 +105,38 @@ public String generateDisplayNameForMethod(Class<?> testClass, Method testMethod
...
+	 * removing parentheses ({@code '()'}) found at the end of method names
+	 * with no parameters.
+	 */
+	class Simple extends Standard {
+
+		@Override
+		public String generateDisplayNameForMethod(Class<?> testClass, Method testMethod) {
+			String displayName = testMethod.getName();
+			if (hasParameters(testMethod)) {
+				displayName += ' ' + parameterTypesAsString(testMethod);
```
marcphilipp: ❓ Why the extra space?

juliette-derancourt: I was actually wondering whether to put it or not.
I originally added it in `ReplaceUnderscores` to make `some_test(Object o)` become `some test (Object)` which looks better imo.
But here if `Simple` is used in Kotlin (I know it’s not dedicated to Kotlin usage but it was the purpose of the original request and I guess most people will use it for that), it will display such method as `some test(Object)` if there is no extra space, which I find less pretty (I might be nitpicking here).
What do you think?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java
sormuras: ...before introducing the `if` statements, I converted the class to its string-representation and switched over those names. That looked awkward, too.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
marcphilipp: I think that's a fair point, I hadn't thought of that. 👍

michaldo:
Hi,
do you think displaying parameters is useful? Parameters are used with @ParameterizedTest only, which has own displaying policy.

Compare output:

- child blood group could be (BloodGroup, BloodGroup, BloodGroup)
  - Father: A, Mother: B, Child: A
  - Father: A, Mother: B, Child: B
  - Father: A, Mother: B, Child: AB

vs

- child blood group could be
  - Father: A, Mother: B, Child: A
  - Father: A, Mother: B, Child: B
  - Father: A, Mother: B, Child: AB

IMHO parameter classes decrease elegance

sbrannen:
> do you think displaying parameters is useful?

Yes, we do. 

> Parameters are used with @ParameterizedTest only

That's incorrect.

Any `@Test` method can accept parameters. Please read the user guide for further details.

michaldo:
Indeed, I read documentation and there are more options to inject parameter to test method than @ParametrizedTest

However, documentation focus on injecting something like "test-context" or extra resources like unique temp directory. These method parameters are not core test part.

I still don't see an example that printing class names of parameters increase output readability. 

I rather think there are two worlds - method names and method params. Method name are almost absolutely flexible and can be freely rearranged to achieve best readable and mess-free display name. Method params have classes with concrete names. Their always "sounds" the same. And contrast between fine tuned method display name and rough parameter class name is unavoidable.

juliette-derancourt:
> However, documentation focus on injecting something like "test-context" or extra resources like unique temp directory. These method parameters are not core test part.

You can have all sorts of parameters if you are using a `ParameterResolver`. These would be "core test".

If you want more insight as to why parameters are useful, please take a look here: https://github.com/junit-team/junit5/issues/1875#issuecomment-491521853

michaldo:
Thanks @juliette-derancourt, I wondered why parameters are part of display name and now I see answer. Overloading.

Anyway, it's nice to see someone else shares my doubts about parameters role. 

I think you are focused on technology: Java has overloading, if display name omits parameters, overloaded method will not be distinguishable. 

I rather think display name generator should focus on readability: method name should "carry" meaning of test, and author of test is responsible for meaningful method names. For example, Java Bean convention determines strict rule how method must be named (get/set). For test the only requirement for method name is meaningfulness.

When I started to play with display name generator, I was disgusted that my effort to provide natural language test output is destroyed by rough class names. I started PR to remove parameters from output, but I didn't know why they were added at all.
Hopefully, I found this PR and parameters origin. I see you are comited to parameters and I will not continue my PR.

PR #2214, opened 2020-03-13
Introduce `@EnabledIf`/`@DisabledIf` annotations

### LINKED ISSUE #1911: Convenient way to use static Java methods in @Enabled*/@Disabled*
## Idea

I would like to be able to easily use the logic placed in my static Java methods to determine if a given test(s) should be executed or not.

## Rationale

In some cases it is convenient to delegate the logic in conditional test execution to custom Java code. A real life usage - determine if strong cryptography is available in JDK to [disable all related tests](https://github.com/4finance/micro-infra-spring/blob/7d77b9bda1abe6fa236c13c8ea73eb15623ad18e/micro-infra-spring-config/src/test/groovy/com/ofg/infrastructure/property/decrypt/LoadingEncryptedFromFileSystemSpec.groovy#L17-L18) and [fail just one](https://github.com/4finance/micro-infra-spring/blob/7d77b9bda1abe6fa236c13c8ea73eb15623ad18e/micro-infra-spring-config/src/test/groovy/com/ofg/infrastructure/property/decrypt/BrokenJceInstallationWarningSpec.groovy#L9-L14) to inform external contributor in meaningful way that his/her configuration is not optimal (I know it is somehow obsolete as of Java 11 and using OpenJDK all around, but at the time it was a real issue for us with Oracle JDK).

In Spock it can be easily done with:

```groovy
@IgnoreIf({ SomeUtil.isStrongEncryptionSupported() })
class BrokenJceInstallationWarningSpec extends Specification {
    ....
```

In JUnit 5 it should be achievable with a few lines of Java Script and `@EnableIf` (looking at the JUnit 5 documentation) which is somehow inconvenient. In addition `@EnableIf` has been deprecated recently.

## API proposal

As Java is more limited that Groovy it doesn't seem to be possible to get that level of API:

```java
@EnabledX(() -> MyUtil.callMyStaticMethodReturningBoolean())
// or
@EnabledX(MyUtil::callMyStaticMethodReturningBoolean)
```

Therefore, it would probably be needed to use String literals:

```java
@EnabledX(method = "com.example.MyUtil.callMyStaticMethodReturningBoolean")
// or
@EnabledX(package = "com.example", method = "MyUtil.callMyStaticMethodReturningBoolean")
```

If package is omitted JUnit could try to assume the same package as enclosing class to try to resolve the method.

I would exclude an ability to pass method arguments or JUnit context to a called method.

Please note that I don't know the internal of JUnit 5, so some of the idea could be hard/problematic to implement. Feel free to choose some other (convenient) way to implement that feature.

BTW, `@EnabledX` is also a place holder which should be changed to something more meaningful.

## Deliverables

- [ ] Convenient API to use static Java methods in `@Enabled*`/`@Disabled*`
- [ ] Documentation update

juliette-derancourt:
Resolves #1911.

---

---

### Definition of Done


junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/DisabledIf.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apiguardian.api.API;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * {@code @DisabledIf} is used to signal that the annotated test class or
+ * test method is <em>disabled</em> only if the provided condition
+ * evaluates to {@code true}.
+ *
+ * <p>When applied at the class level, all test methods within that class
+ * will be disabled on the same specified operating systems.
```
marcphilipp: I believe you might have copied that from somewhere. 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Optional;
+
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.platform.commons.JUnitException;
+import org.junit.platform.commons.util.Preconditions;
+import org.junit.platform.commons.util.ReflectionUtils;
+
+public abstract class MethodBasedCondition implements ExecutionCondition {
```
marcphilipp: I think this class can be package-private.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
```diff
@@ -0,0 +1,88 @@
...
+		Class<?> clazz = ReflectionUtils.tryToLoadClass(className).getOrThrow(
+			cause -> new JUnitException(format("Could not load class [%s]", className), cause));
+		return findMethod(clazz, methodName);
+	}
+
+	private Method findMethod(Class<?> clazz, String methodName) {
+		return ReflectionUtils.findMethod(clazz, methodName) //
+				.orElseGet(() -> ReflectionUtils.findMethod(clazz, methodName, ExtensionContext.class) //
+						.orElseThrow(() -> new JUnitException(
+							format("Could not find method [%s] in class [%s]", methodName, clazz.getName()))));
```
marcphilipp: Can we reuse this logic from `MethodArgumentsProvider` somehow?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/DisabledIf.java
juliette-derancourt: Oopsie, you got me! 😁

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
juliette-derancourt: Indeed, it was even making the build fail ^^

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
juliette-derancourt: Hmm I thought about that at first, but since the classes were two different modules I put that aside for now. But I agree.
Where would you put the logic in common? In `ReflectionUtils`?

sormuras: > Where would you put the logic in common? In `ReflectionUtils`?

Sounds like a good place. We may export that logic to external users later, if requested.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -437,6 +437,21 @@ method. Specifically, these annotations will be found if they are directly prese
 indirectly present, or meta-present on a given element.
 ====
 
+[[writing-tests-conditional-execution-custom]]
+==== Custom Conditions
+
+A container or test may be enabled or disabled based on the boolean return of a static
+method via the `{EnabledIf}` and `{DisabledIf}` annotations. The method is provided to
```
marcphilipp: These attributes need to be defined in `documentation/src/docs/asciidoc/link-attributes.adoc`.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -437,6 +437,21 @@ method. Specifically, these annotations will be found if they are directly prese
 indirectly present, or meta-present on a given element.
 ====
 
+[[writing-tests-conditional-execution-custom]]
+==== Custom Conditions
+
+A container or test may be enabled or disabled based on the boolean return of a static
```
marcphilipp: We should describe that the method does not have to be static in case `@TestInstance(PER_CLASS)` is used and add tests for this case.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -1314,6 +1315,11 @@ public static boolean isMethodPresent(Class<?> clazz, Predicate<Method> predicat
 		return Optional.empty();
 	}
 
+	public static Method getRequiredMethod(Class<?> clazz, String methodName, Class<?>... parameterTypes) {
```
marcphilipp: This method needs Javadoc, including `@since`, and an `@API` annotation.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
juliette-derancourt: Didn't noticed this file, thanks! 👍

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
juliette-derancourt: Ah yes, this one slipped under my radar 😄

juliette-derancourt: Would this make more sense to throw a `NoSuchMethodException` here instead of `JUnitException`? 🤔

marcphilipp: When we throw exceptions, we generally only use `JUnitException` or one of its subclasses to avoid confusion with exceptions thrown from tests. Thus, I'd keep it as is.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
juliette-derancourt: Should I add a code sample to illustrate this and/or the usage of an external method?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
```diff
@@ -66,7 +66,7 @@ private Object evaluateCondition(Method method, ExtensionContext context) {
 		Preconditions.condition(areParametersSupported(method),
 			() -> format("method [%s] should take either an ExtensionContext or no parameters", method.getName()));
 
-		Object testInstance = context.getTestInstance().orElse(null);
+		Object testInstance = context.getRequiredTestInstance();
```
juliette-derancourt, line 69: @marcphilipp Am I wrong to assume that the `testInstance` will always be present in this case?
If so, is the presence of `@TestInstance(Lifecycle.PER_CLASS)` even required in the case of a non-static method?

marcphilipp: I think we're missing integration test for applying the condition on the class level in which case the test instance would not be present.

juliette-derancourt:
> Unless I'm missing sth. there's no integration test that uses `@EnabledIf`/`@DisabledIf` on a test class, is there?

No, there is not. I tried to add one on a nested class inside `EnabledIf/DisabledIfIntegrationTests`, but it doesn't work well with the way `AbstractExecutionConditionTests` is designed.
Actually, there's no such test for any of the `ExecutionCondition` implementations (unless I missed them?), so should we create a new way to test this use case?

Btw, I just noticed that using `@EnabledIf`/`@DisabledIf` with a non-static method does not work at the class level, even with a per-class lifecycle (but it works great with a static method). The test instance doesn't seem to be present in the extension context in that case, although I thought it should because of the per-class lifecycle... am I wrong? 🤔

marcphilipp:
> I tried to add one on a nested class inside `EnabledIf/DisabledIfIntegrationTests`, but it doesn't work well with the way `AbstractExecutionConditionTests` is designed.

How so? This works for me:

```java
@Nested
@DisabledIf("org.junit.jupiter.api.condition.StaticConditionMethods#returnsTrue")
class ConditionallyDisabledClass {

	@Test
	void disabledBecauseConditionMethodReturnsTrue() {
		fail("Should be disabled");
	}

}
```

> Actually, there's no such test for any of the `ExecutionCondition` implementations (unless I missed them?), so should we create a new way to test this use case?

I think I'd be ok with just having integration tests for this use case. @sbrannen Since you wrote most (all?) of the other conditions, what do you think?

> Btw, I just noticed that using `@EnabledIf`/`@DisabledIf` with a non-static method does not work at the class level, even with a per-class lifecycle (but it works great with a static method). The test instance doesn't seem to be present in the extension context in that case, although I thought it should because of the per-class lifecycle... am I wrong? 🤔

I think you're right and that's ok. Conditions get evaluated before the test class is instantiated or any lifecycle methods are executed.

sbrannen:
> > Actually, there's no such test for any of the `ExecutionCondition` implementations (unless I missed them?), so should we create a new way to test this use case?
> 
> I think I'd be ok with just having integration tests for this use case. @sbrannen Since you wrote most (all?) of the other conditions, what do you think?

I think that's OK. 

We can always revisit that later if we deem it necessary (e.g., we have doubts about the correctness of the implementation or bugs are reported).

juliette-derancourt:
> How so? This works for me:
> 
> ```java
> @Nested
> @DisabledIf("org.junit.jupiter.api.condition.StaticConditionMethods#returnsTrue")
> class ConditionallyDisabledClass {
> 
> 	@Test
> 	void disabledBecauseConditionMethodReturnsTrue() {
> 		fail("Should be disabled");
> 	}
> 
> }
> ```

It does. It just felt weird to not have a matching method in `DisabledIfConditionTests`, but I forgot `DisabledIfIntegrationTests` is also run on its own during the build.
I'll just add this then 😄 

> I think you're right and that's ok. Conditions get evaluated before the test class is instantiated or any lifecycle methods are executed.

So you're saying we can't use a non-static method at the class level?
Then I'm wondering if it's worth mentioning the use of `Lifecycle.PER_CLASS` in the user guide, since it doesn't seem to be needed in order to use non-static condition methods at method level.

> We can always revisit that later if we deem it necessary (e.g., we have doubts about the correctness of the implementation or bugs are reported).

👍

PR #2217, opened 2020-03-18
Make JAR builds reproducible

mrwilson:
## Overview

This PR makes the java bytecode generated by this project [reproducible](https://reproducible-builds.org/). This means that under identical builds condition (for example Java version), repeated builds should provide the same output byte-for-byte.

It achieves this by:

* supporting [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/specs/source-date-epoch/) as a way of fixing build-time-specific metadata added to JARs
* using Gradle's [reproducible archives](https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives) functionality to fix timestamps and file ordering within JARs
* re-archiving JARs that are non-deterministic because they use `jar` to rewrite their contents after build.

## Confirming the change

```shell
#!/bin/bash -e

rm -f checksums*

# Fix build time
export SOURCE_DATE_EPOCH=$(date +%s)

./gradlew clean build -x test --parallel --rerun-tasks

find . -name '*.jar' \
    | grep '/build/libs/' \
    | grep -v 'javadoc' \
    | sort \
    | xargs sha256sum > checksums-1.txt

./gradlew clean build -x test --parallel --rerun-tasks

find . -name '*.jar' \
    | grep '/build/libs/' \
    | grep -v 'javadoc' \
    | sort \
    | xargs sha256sum > checksums-2.txt

diff checksums-{1,2}.txt
```

The diff should be empty, that is to say, both independent runs of the build should have generated byte-for-byte identical outputs.

## Omitting Javadocs

The javadocs outputs remain non-deterministic - despite the inclusion of `noTimestamp(true)` in the project's javadoc configuration, the comment header is still being generated. It looks like the javadoc mechanism in Gradle isn't honouring this flag.

I wasn't able to solve this problem, but would be necessary to achieve full reproducibility.

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement). ✅ 

---

### Definition of Done

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,6 @@
 import java.time.OffsetDateTime
+import java.time.Instant
+import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 
 plugins {
@@ -16,7 +18,24 @@ buildScan {
 	}
 }
 
-val buildTimeAndDate = OffsetDateTime.now()
+val buildTimeAndDate by extra {
+
+	// SOURCE_DATE_EPOCH is a UNIX timestamp for pinning build metadata against
+	// in order to achieve reproducible builds
+	//
+	// More details - https://reproducible-builds.org/docs/source-date-epoch/
+
+	if (System.getenv().containsKey("SOURCE_DATE_EPOCH")) {
+
+		val sourceDateEpoch = System.getenv("SOURCE_DATE_EPOCH").toLong()
+
+		Instant.ofEpochSecond(sourceDateEpoch).atOffset(ZoneOffset.UTC)
+
+	} else {
+		OffsetDateTime.now()
+	}
+}
+
 val buildDate by extra { DateTimeFormatter.ISO_LOCAL_DATE.format(buildTimeAndDate) }
 val buildTime by extra { DateTimeFormatter.ofPattern("HH:mm:ss.SSSZ").format(buildTimeAndDate) }
 val buildRevision by extra { versioning.info.commit }
@@ -103,6 +122,11 @@ subprojects {
 	else if (project in vintageProjects) {
 		group = property("vintageGroup")!!
 		version = property("vintageVersion")!!
+	}
+
+	tasks.withType<AbstractArchiveTask>().configureEach {
+		isPreserveFileTimestamps = false
+		isReproducibleFileOrder = true
 	}
 
 	pluginManager.withPlugin("java") {
--- a/buildSrc/src/main/kotlin/java-repackage-jars.gradle.kts
+++ b/buildSrc/src/main/kotlin/java-repackage-jars.gradle.kts
@@ -1,0 +1,42 @@
+import java.util.Calendar
+import java.util.GregorianCalendar
+import java.util.jar.JarEntry
+import java.util.jar.JarFile
+import java.util.jar.JarOutputStream
+import org.gradle.api.internal.file.archive.ZipCopyAction
+
+// This registers a `doLast` action to rewrite the timestamps of the project's output JAR
+afterEvaluate {
+	val jarTask = (tasks.findByName("shadowJar") ?: tasks["jar"]) as Jar
+
+	jarTask.doLast {
+
+		val newFile = createTempFile("rewrite-timestamp")
+		val originalOutput = jarTask.archiveFile.get().getAsFile()
+
+		newFile.outputStream().use { os ->
+
+			val newJarStream = JarOutputStream(os)
+			val oldJar = JarFile(originalOutput)
+
+			oldJar.entries()
+					.toList()
+					.distinctBy { it.name }
+					.sortedBy { it.name }
+					.forEach { entry ->
+						val jarEntry = JarEntry(entry.name)
+
+						// Use the same constant as the fixed timestamps in normal copy actions
+						jarEntry.time = ZipCopyAction.CONSTANT_TIME_FOR_ZIP_ENTRIES
+
+						newJarStream.putNextEntry(jarEntry)
+
+						oldJar.getInputStream(entry).copyTo(newJarStream)
+					}
+
+			newJarStream.finish()
+		}
+
+		newFile.renameTo(originalOutput)
+	}
+}
--- a/documentation/src/docs/asciidoc/user-guide/appendix.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/appendix.adoc
@@ -1,6 +1,18 @@
 [[appendix]]
 == Appendix
 
+[[reproducible-builds]]
+=== Reproducible Builds
+
+Starting with version 5.7, JUnit 5 aims for its non-javadoc JARs to be https://reproducible-builds.org/[reproducible]
+
+Under identical build conditions, such as Java version, repeated builds should provide the
+same output byte-for-byte.
+
+This means that anyone can reproduce the build conditions of the artifacts on Maven
+Central/Sonatype and produce the same output artifact locally, confirming that the
+artifacts in the repositories were actually generated from this source code.
+
 [[dependency-metadata]]
 === Dependency Metadata
 
--- a/src/checkBuildReproducibility.sh
+++ b/src/checkBuildReproducibility.sh
@@ -1,0 +1,23 @@
+#!/bin/bash -e
+
+rm -rf checksums*
+
+export SOURCE_DATE_EPOCH=$(date +%s)
+
+function calculate_checksums() {
+    OUTPUT=$1
+
+    ./gradlew --no-build-cache clean assemble --parallel
+
+    find . -name '*.jar' \
+        | grep '/build/libs/' \
+        | grep --invert-match 'javadoc' \
+        | sort \
+        | xargs sha256sum > ${OUTPUT}
+}
+
+
+calculate_checksums checksums-1.txt
+calculate_checksums checksums-2.txt
+
+diff checksums-1.txt checksums-2.txt
--- a/.github/workflows/reproducible-build.yml
+++ b/.github/workflows/reproducible-build.yml
@@ -1,0 +1,25 @@
+name: Reproducible build
+
+on:
+  push:
+    branches:
+    - master
+    - 'releases/*'
+  pull_request:
+    branches:
+    - '*'
+
+jobs:
+  check_build_reproducibility:
+    name: 'Check build reproducibility'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: 'Set up JDK 11'
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Build and compare checksums
+      shell: bash
+      run: |
+        ./src/checkBuildReproducibility.sh
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ gradle-app.setting
 *.graphml
 coverage.db*
 .metadata
+
+checksums*
--- a/junit-platform-commons/junit-platform-commons.gradle.kts
+++ b/junit-platform-commons/junit-platform-commons.gradle.kts
@@ -3,6 +3,7 @@ import java.util.spi.ToolProvider
 plugins {
 	`java-library-conventions`
 	`java-multi-release-sources`
+	`java-repackage-jars`
 }
 
 description = "JUnit Platform Commons"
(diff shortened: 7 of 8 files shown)


sbrannen:
Tentatively slated for 5.7 M1 for _team discussion_

marcphilipp:
> The JARs that it does not apply to are `junit-platform-commons` and `junit-platform-console` as these use direct calls to the `jar` command as part of their builds which introduces non-determinism through the last modified timestamps.

Since all other jars use `junit-platform-commons`, merging this would have limited value. Is there an option for the `jar` command to use a fixed timestamp or can we add another task action to reset the timestamps of the jars?

mrwilson:
As far as I'm aware, we'd need to go with option 2 - happy to add that.

I imagine something that loops over the contents of the archive and sets last modified to the value of `buildTimeAndDate`, which can be pinned to SOURCE_DATE_EPOCH.

marcphilipp:
I think it should use the same timestamp Gradle uses:
https://github.com/gradle/gradle/blob/199fd61e4ab42e4b61fda5d70158175b078d4b9f/subprojects/core/src/main/java/org/gradle/api/internal/file/archive/ZipCopyAction.java#L42-L57

mrwilson:
@marcphilipp - I believe that the most recent commit solves the issue for the two special-case sub-projects.

buildSrc/build.gradle.kts
```diff
@@ -13,3 +13,10 @@ dependencies {
 	implementation("biz.aQute.bnd:biz.aQute.bnd.gradle:5.0.0")
 	implementation("com.github.jengelman.gradle.plugins:shadow:5.2.0")
 }
+
+allprojects {
+	tasks.withType<AbstractArchiveTask>().configureEach {
+		isPreserveFileTimestamps = false
+		isReproducibleFileOrder = true
+	}
+}
```
marcphilipp: Do we need this here since we're not publishing any jars produced by `buildSrc`?

build.gradle.kts
```diff
@@ -145,6 +176,42 @@ subprojects {
...
+
+									oldJar.getInputStream(entry).copyTo(newJarStream)
+								}
+
+						newJarStream.finish()
+					}
+
+					newFile.renameTo(originalOutput)
+				}
+			}
```
marcphilipp: I'd like to move this logic to `buildSrc`. Are you feeling up for it or shall I polish it myself?

buildSrc/build.gradle.kts
mrwilson: Good point - I included it for completeness as the output jar was picked up by the testing script, will remove.

build.gradle.kts
mrwilson: Yep, I'll modify the commits appropriately. 👍

mrwilson:
Those commits have been re-written, and the test script only returns a diff for `buildSrc` as discussed earlier in the PR.

The jar-rewriting has been moved to its own file under `buildSrc` and is imported only in the two projects that need it. Feels much cleaner!

mrwilson:
This looks like it's almost ready to go - is there a sensible place to record this feature in the User Guide / Release Notes? (Last remaining unchecked box)

marcphilipp:
I'd say as a new section in the appendix. WDYT?

mrwilson:
That sounds good - I've also tried to add another GitHub action which will perform multiple `assemble` steps and compare their checksums, to ensure that future changes don't regress this property.

documentation/src/docs/asciidoc/user-guide/appendix.adoc
```diff
@@ -1,6 +1,13 @@
 [[appendix]]
 == Appendix
 
+[[reproducible-builds]]
+=== Reproducible Builds
+
+JUnit 5 aims for its non-javadoc JARs to be https://reproducible-builds.org/[reproducible] - under identical build conditions, such as Java version, repeated builds should provide the same output byte-for-byte.
+
+This means that anyone can reproduce the build conditions of the artifacts on Maven Central/Sonatype and produce the same output artifact locally, confirming that the artifacts in the repositories were actually generated from this source code.
```
marcphilipp, line 9: Please hard-wrap these lines at 90 chars (cf. https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#documentation).

documentation/src/docs/asciidoc/user-guide/appendix.adoc
```diff
@@ -1,6 +1,13 @@
 [[appendix]]
 == Appendix
 
+[[reproducible-builds]]
+=== Reproducible Builds
+
+JUnit 5 aims for its non-javadoc JARs to be https://reproducible-builds.org/[reproducible] - under identical build conditions, such as Java version, repeated builds should provide the same output byte-for-byte.
```
marcphilipp, line 7: I think we should add "Starting with version 5.7, ...".

src/checkBuildReproducibility.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -e
+
+rm -rf checksums*
+
+export SOURCE_DATE_EPOCH=$(date +%s)
+
+function calculate_checksums() {
+    OUTPUT=$1
+
+    ./gradlew clean assemble
```
marcphilipp, line 10:
```suggestion
    ./gradlew --no-build-cache clean assemble
```

.github/workflows/main.yml
```diff
@@ -94,6 +94,20 @@ jobs:
...
+    - uses: actions/checkout@v2
+    - name: 'Set up JDK 11'
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Build and compare checksums
+      shell: bash
+      run: |
+        ./src/checkBuildReproducibility.sh
+
```
marcphilipp, line 110: Let's move this to a dedicated workflow so it can run in parallel with the main workflow.

mrwilson:
Made changes as suggested.

marcphilipp:
Thanks a lot, @mrwilson! 🎉

jbduncan:
Glad to see this merged in! Great work @mrwilson. 🎉

vlsi:
Frankly speaking, I see no reason to enforce timestamps in jar files.
It could be OK to just leave them as is (==0 timestamps) and avoid repackage step.

---

It is required to capture the build environment: timestamp, and Java version/vendor.

Even though jar most jar files have Build-Date and Build-Time headers, it is not clear how to reproduce JUnit build.
What value should be passed to `SOURCE_DATE_EPOCH` ?

buildSrc/src/main/kotlin/java-repackage-jars.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+
+			oldJar.entries()
+					.toList()
+					.distinctBy { it.name }
+					.sortedBy { it.name }
+					.forEach { entry ->
+						val jarEntry = JarEntry(entry.name)
+
+						// Use the same constant as the fixed timestamps in normal copy actions
+						jarEntry.time = ZipCopyAction.CONSTANT_TIME_FOR_ZIP_ENTRIES
```
vlsi, line 30: @mrwilson , this seems to specify the same constant value for all the jars. Is `isPreserveFileTimestamps = false` enough then? What is the purpose of rewriting the jar?

mrwilson: This is just for two specific cases where the actual `jar` command was being used to manipulate the artifact after it's been created by Gradle - I tried to replicate what the `jar` tool was doing in pure Gradle, with no success.

As an immediate solution, this resets all the timestamps of the two artifacts in question - the choice of timestamp was informed by what the `isPreserveFileTimestamps = false` would do if this step wasn't necessary.

I hope that helps.

vlsi: Thanks, that clarifies things. It would probably be great if something like that was in the comment above.

hboutemy:
FYI, I just added Gradle builds support to reproducible-central
and I tested JUnit5 latest release: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/junit/junit5/README.md
sorry, there are still a few issues

sormuras:
The results linked via 🔍 show differences like:

> `│ -Created-By: 11.0.14.1 (Oracle Corporation 11.0.14.1+1)`
> `│ +Created-By: 11.0.13 (Eclipse Adoptium 11.0.13+8)`

Seems like not all "external" properties are stable - or not all expected differences are excluded from the comparison.

What is the latest result for JUnit 5.10.1?

hboutemy:
> What is the latest result for JUnit 5.10.1?

with my limited knowledge of Gradle, I can't make the build work: I need help

```
$ ./gradlew --no-daemon clean assemble publishToMavenLocal -x test -x signMavenPublication
...

* What went wrong:
Execution failed for task ':junit-bom:publishMavenPublicationToMavenLocal'.
> Failed to publish publication 'maven' to repository 'mavenLocal'
   > Invalid publication 'maven': artifact file does not exist: '/tmp/junit-bom/junit-bom/build/publications/maven/pom-default.xml.asc'
...
```

Do you know how I can fix this?

sormuras:
The command we use in [checkBuildReproducibility.sh](https://github.com/junit-team/junit5/blob/main/gradle/scripts/checkBuildReproducibility.sh) reads:

`./gradlew --no-build-cache clean assemble --parallel -Porg.gradle.java.installations.auto-download=false -Dscan.tag.Reproducibility`

Does this help?

hboutemy:
I tried to run this command: gives build success, but I don't know what this success means

Then I tried to modify source code and rebuild: same success, even if the expected result for modified code is failure because I obvioulsy should not get the same binaries are the 5.10.1 release

then I don't know what this command does, it it does not do what I expect = check local build output against content downloaded from Maven Central

I'll try to have a look at `checkBuildReproducibility.sh` to figure out...

marcphilipp:
@hboutemy Our release builds currently require signing but you can use any PGP key you want. The key id in your local GPG can be configured by passing `-Psigning.gnupg.keyName=<your-key-id>` (e.g. the actual key would be `FF6E2C001948C5F2F38B0CC385911F425EC61B51`). I'll add a build option to future releases so that signing can be disabled.

hboutemy:
thanks @marcphilipp for helping: ok, I need to drop the `-x signMavenPublication`, then I'm now able to build 5.10.1

another question: is there a way to configure the `publishToMavenLocal` output to another place than default `.m2/repository`? This is a flexibility that would ease my work on creating a rebuild script

sormuras:
> for the Build-Time and Build-Date , maybe there is a command line attribute to add for me to get the same binaries as what is published to Maven Central

Set `SOURCE_DATE_EPOCH` environment variable, it's picked by the build.

```
# Fix build time
export SOURCE_DATE_EPOCH=$(date +%s)
```

marcphilipp:
@hboutemy Could you please raise a new issue where we can discuss dropping the `Created-By` attribute?

marcphilipp:
> another question: is there a way to configure the `publishToMavenLocal` output to another place than default `.m2/repository`? This is a flexibility that would ease my work on creating a rebuild script

Yes, you can call `publishAllPublicationsToTempRepository` instead of `publishToMavenLocal` to write to `build/repo` instead.

marcphilipp:
> thanks @marcphilipp for helping: ok, I need to drop the `-x signMavenPublication`, then I'm now able to build 5.10.1

I've just added an option to disable signing that will help with future releases: https://github.com/junit-team/junit5/commit/6238d5545851436d3760732a1ef58e67f880d52d

sormuras:
> Set the SOURCE_DATE_EPOCH environment variable, ...

Perhaps we can store the value used for `SOURCE_DATE_EPOCH` inside the Manifest.mf file for easier consumption? 🤔

vlsi:
>Created-By: 17.0.8 (Azul Systems, Inc. 17.0.8+7-LTS), it's a pain to have that detailed info on the JDK because I need to install exactly the same one (hoping I can find it)

In practice, different JVM can produce different jars, so having a detailed JVM version does help.
I am not sure it should be in MANIFEST though, and it might be better to store it side by side with the generated jars (just like we do not store the timestamps in the jars rather than we store it separately).

hboutemy:
@visi #3559 created, as asked by @marcphilipp 
let's discuss there :)

and thanks @marcphilipp for your hints, I'll be able to improve my Gradle rebuilding script in Reproducible Central to check more projects in the future

PR #2236, opened 2020-04-03
Move versions to gradle.properties

### LINKED ISSUE #2234: Consider moving dependency versions to gradle.properties (and remove Versions from buildSrc)
I see you've moved the versions to dependency constraints (which I think is nice), however, it looks like there's no point in having `Versions` in `buildSrc`.

The problems with `Versions` object are:
1) Every time `Versions` is updated the whole thing is rebuilt
2) The only option to try a new version is to edit the build scripts.

What do you think if the versions are moved to `gradle.properties` instead?
For instance:

gradle.properties:
```properties
apiguardian-api.version=1.1.0
picocli.version=4.2.0
...
```

Then dependency constraints can be created as follows:

```kotlin
val String.v: String get() = rootProject.extra["$this.version"] as String

fun DependencyConstraintHandlerScope.apiv(
    notation: String,
    versionProp: String = notation.substringAfterLast(':')
) =
    "api"(notation + ":" + versionProp.v)

dependencies {
    (constraints) {
        apiv("org.apiguardian:apiguardian-api")
        apiv("info.picocli:picocli")
        ...
    }
}
...
```

It would
1) Consolidate the versions in `gradle.properties`
2) Enable testing with newer versions via `-Ppicocli.version=5.0.0`
3) Avoid recompilation when `Versions` is updated

vlsi:
This reduces Gradle script recompilation, and it enables to specify
dependency versions from the command line.

fixes #2234

## Overview

---

---

### Definition of Done

### DIFF
--- a/buildSrc/src/main/kotlin/ProjectExtensions.kt
+++ b/buildSrc/src/main/kotlin/ProjectExtensions.kt
@@ -8,8 +8,5 @@ val Project.javaModuleName: String
 val Project.versions: Versions
     get() {
         var versions: Versions? by rootProject.extra
-        if (versions == null) {
-            versions = Versions(rootProject)
-        }
-        return versions!!
+        return versions ?: Versions(rootProject).also { versions = it }
     }
--- a/buildSrc/src/main/kotlin/ProjectExtensions.kt
+++ b/buildSrc/src/main/kotlin/ProjectExtensions.kt
@@ -7,7 +7,7 @@ val Project.javaModuleName: String
 
 val Project.versions: Versions
     get() {
-        val versions: Versions by rootProject.extra {
+        val versions by rootProject.extra {
             Versions(rootProject)
         }
         return versions
--- a/buildSrc/src/main/kotlin/ProjectExtensions.kt
+++ b/buildSrc/src/main/kotlin/ProjectExtensions.kt
@@ -1,4 +1,14 @@
 import org.gradle.api.Project
+import org.gradle.kotlin.dsl.extra
 
 val Project.javaModuleName: String
     get() = "org." + this.name.replace('-', '.')
+
+val Project.versions: Versions
+    get() {
+        val extra = rootProject.extra
+        if (!extra.has("versions")) {
+            extra.set("versions", Versions(rootProject))
+        }
+        return extra.get("versions") as Versions
+    }
--- a/buildSrc/src/main/kotlin/ProjectExtensions.kt
+++ b/buildSrc/src/main/kotlin/ProjectExtensions.kt
@@ -1,14 +1,15 @@
 import org.gradle.api.Project
 import org.gradle.kotlin.dsl.extra
-import org.gradle.kotlin.dsl.invoke
+import org.gradle.kotlin.dsl.provideDelegate
 
 val Project.javaModuleName: String
     get() = "org." + this.name.replace('-', '.')
 
 val Project.versions: Versions
     get() {
-        val versions by rootProject.extra {
-            Versions(rootProject)
+        var versions: Versions? by rootProject.extra
+        if (versions == null) {
+            versions = Versions(rootProject)
         }
-        return versions
+        return versions!!
     }
--- a/buildSrc/src/main/kotlin/ProjectExtensions.kt
+++ b/buildSrc/src/main/kotlin/ProjectExtensions.kt
@@ -1,14 +1,14 @@
 import org.gradle.api.Project
 import org.gradle.kotlin.dsl.extra
+import org.gradle.kotlin.dsl.invoke
 
 val Project.javaModuleName: String
     get() = "org." + this.name.replace('-', '.')
 
 val Project.versions: Versions
     get() {
-        val extra = rootProject.extra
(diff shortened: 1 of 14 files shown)


marcphilipp:
@vlsi I hope you don't mind, I pushed a commit to the PR branch with what I had in mind. WDYT?

buildSrc/src/main/kotlin/ProjectExtensions.kt
```diff
@@ -1,14 +1,14 @@
...
 
 val Project.javaModuleName: String
     get() = "org." + this.name.replace('-', '.')
 
 val Project.versions: Versions
     get() {
-        val extra = rootProject.extra
-        if (!extra.has("versions")) {
-            extra.set("versions", Versions(rootProject))
+        val versions: Versions by rootProject.extra {
```
marcphilipp: This will create this object every time this extension function is called. My version would only create it once.

vlsi:
@marcphilipp , I see what you mean. That is a clever approach.
What I don't quite like with that is adding a dependency requires edits to `gradle.properties`, `dependencies.gradle.kts`, `Versions.kt`. 

For instance, `bnd` is used just once (in `dependencies`), so adding `bnd` to `Versions` does not probably help a lot.

However, that is up to you.

buildSrc/src/main/kotlin/ProjectExtensions.kt
```diff
@@ -1,4 +1,12 @@
...
+import org.gradle.kotlin.dsl.extra
+import org.gradle.kotlin.dsl.provideDelegate
 
 val Project.javaModuleName: String
     get() = "org." + this.name.replace('-', '.')
+
+val Project.versions: Versions
+    get() {
+        var versions: Versions? by rootProject.extra
+        return versions ?: Versions(rootProject).also { versions = it }
```
marcphilipp, line 11: Nice! I wasn't aware setting the delegate would store it in extra properties. Now that I think about it that makes perfect sense. 👍

marcphilipp:
> What I don't quite like with that is adding a dependency requires edits to `gradle.properties`, `dependencies.gradle.kts`, `Versions.kt`.

I'm a bit torn on this myself. I'll try the more dynamic approach now.

marcphilipp:
I pushed another commit that removes the explicit properties, i.e. `versions.junit4` is now `versions["junit4"]`.

@vlsi @sormuras Which version (no pun intended) do you like better?

sormuras:
This.

> `versions["junit4"]`

Feels like a lookup from a map.

marcphilipp:
> This: `versions["junit4"]` Feels like a lookup from a map.

Do you that as a pro or con?

sormuras:
Pro. As in "This version I like better".

vlsi:
> versions["junit4"].

This looks good to me, and, in practice, you might want to have `junit4` and `junit4Min` as explicit vals (since they are used often).

---

Frankly speaking, I don't like the duplication of `${...}` in `dependencies`, and it would probably make sense to infer property name from `artifactId` there via the extension (like in my first proposal).

marcphilipp:
> Frankly speaking, I don't like the duplication of `${...}` in `dependencies`, and it would probably make sense to infer property name from `artifactId` there via the extension (like in my first proposal).

I saw what you did there and to me it felt like too much magic. In the end it's a matter of taste so we might have to agree to disagree on certain aspects.

marcphilipp:
> This looks good to me, and, in practice, you might want to have `junit4` and `junit4Min` as explicit vals (since they are used often).

I think that's a good compromise. I extraced the most commonly used versions into explicit properties.

platform-tooling-support-tests/platform-tooling-support-tests.gradle.kts
```diff
@@ -69,10 +69,10 @@ tasks.test {
...
 
 		// Pass version constants (declared in Versions.kt) to tests as system properties
-		systemProperty("Versions.apiGuardian", Versions.apiGuardian)
-		systemProperty("Versions.assertJ", Versions.assertJ)
-		systemProperty("Versions.junit4", Versions.junit4)
-		systemProperty("Versions.ota4j", Versions.ota4j)
+		systemProperty("Versions.apiGuardian", versions.apiguardian)
+		systemProperty("Versions.assertJ", versions["assertj"])
+		systemProperty("Versions.junit4", versions.junit4)
+		systemProperty("Versions.ota4j", versions.opentest4j)
```
sormuras, line 75: Should all 4 lines follow the `versions["assertj"]` pattern?

marcphilipp: See rationale in https://github.com/junit-team/junit5/pull/2236#issuecomment-609015745. But I can roll that last commit back if the rest of the team feels strongly about it.

sormuras: Understood. Let's keep them as they are for the time being.

marcphilipp:
Thanks for the initiative, @vlsi! 👍

PR #2251, opened 2020-04-07
Avoid inner class cycle detection for non-matching predicate

### LINKED ISSUE #2249: Detecting cycles in inner class hierarchies throws exceptions even for classes not marked as @Nested
Upgrading to JUnit 5.6 is breaking a unit test I have.

```java
private class MySillyClosure(...) extends Closure<RetryStrategy>(...) { 
    @override
    RetryStrategy call(...) { return MyStrategy() } 
}
```

This class is defined in my test suite and the recursive check fails even though I have not marked any of these classes as `@Nested`.

I think this is the suspect code:

https://github.com/junit-team/junit5/blob/master/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/predicates/IsTestClassWithTests.java#L46

https://github.com/junit-team/junit5/blob/master/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java#L1007-L1008

Notice that the code looks for all nested classes (and asserts on finding a cycle) before attempting to filter out using the predicate for `@Nested`. Therefore any class that has this inner inheritance (in my case Groovy `Closure` and `WriteableClosure`) is causing an error even if they are not marked as `@Nested`.

For reference, here is the closure code and its inner class:

https://github.com/groovy/groovy-core/blob/master/src/main/groovy/lang/Closure.java#L847

Is there a way for me to disable this recursion check?

sbrannen:
This PR is incomplete with regard to documentation in the release notes, but it is ready for peer review.

Fixes #2249 

### Definition of Done

### DIFF
--- a/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/discovery/predicates/IsTestClassWithTestsTests.java
+++ b/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/discovery/predicates/IsTestClassWithTestsTests.java
@@ -84,6 +84,15 @@ class IsTestClassWithTestsTests {
 		assertFalse(isTestClassWithTests.test(PrivateStaticTestCase.class));
 	}
 
+	/**
+	 * @see https://github.com/junit-team/junit5/issues/2249
+	 */
+	@Test
+	void recursiveHierarchies() {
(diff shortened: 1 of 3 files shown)


sbrannen:
Closed via 11ddd8766b8e5ca661ca554a3c144673180fbc15

PR #2253, opened 2020-04-10
Add FilePosition in FileSelector and ClasspathResourceSelector

### LINKED ISSUE #2146: Support FilePosition in discovery selectors
Currently there is no good way to discover and execute a single test when using a file based test system like Cucumber. So I would like to request support for a file position in `UriSelector`, `FileSelector` and `ClasspathResourceSelector`.

**Deliverables**

Support for a file position when using:
 - URI selectors
 - File selectors
 - Classpath resource selectors

**Motivation**

Cucumber executes feature files written in Gherkin. A feature file consists of multiple scenarios organized in a hierarchical structure. A project may contain multiple feature files, typically organized in some directory structure.

For example:
```
|- src/main/java/com/example
| |- App.java
|- src/test/java/com/example
| |- StepDefinitions.java
|- src/test/resource/com/example/
| |- single.feature
| |- rule.feature
| |- outline.feature
```

Contents of `outline.feature`:
```feature
Feature: A feature with scenario outlines

  Scenario: A scenario
    Given a scenario
    When it is executed
    Then is only runs once

  Scenario Outline: A scenario outline
    Given a scenario outline
    When it is executed
    Then <example> is used

    Examples: With some text
      | example |
      | A       |
      | B       |

    Examples: With some other text
      | example |
      | C       |
      | D       |
```

`outline.feature`  contains 5 tests in a hierarchy:

```
A feature with scenario outlines
|- A scenario
|- A scenario outline
| |- With some text
| | |- Example #1
| | |- Example #2
| |- With some other text
| | |- Example #1
| | |- Example #2
```
Now suppose I would like select both tests (`Example #1`, `Example #2`) nested under the example  `With some text` on line 17.

Currently JUnit discovery selectors (`ClasspathRootSelector`, `ClasspathResourceSelector`, `PackageSelector`, `FileSelector`, `DirectorySelector`) only allow me to discover all tests in a single directory or file.

The `UniqueIdSelector` could be used to select the `With some text` container but requires reproducing the unique id which is rather complicated and quite possibly an implementation detail that should not be relied upon.

Cucumbers  current`TestEngine` implementation currently allows for a `line` query parameter to be added in when parsing using the `UriSelector`. For example:

  - `classpath:/com/example/outline.feature?line=17`
  - `file:/home/mpkorstanje/projects/cucumber/src/test/resources/com/example/outline.feature?line=17`

Since this is not formalized I suspect other test engines that consume the `UriSelector` may crash on the query parameter and so I would like something safer. 

For IDEs being able to use a `FileSelector` with a file position would also simplify the complexity of providing support for file based test engines. Naively allowing the user to select any line in a file and execute the discovered tests for that line would provide for a minimal implementation without knowing any details about the underlying implementation of the test engine.

The same reasoning applies to `ClasspathResourceSelector`.

timtebeek:
## Overview

Fixes #2146.

---

---

### Definition of Done


junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/FilePosition.java
```diff
@@ -0,0 +1,178 @@
...
+	private final int line;
+	private final Integer column;
+
+	protected FilePosition(int line) {
+		Preconditions.condition(line > 0, "line number must be greater than zero");
+		this.line = line;
+		this.column = null;
+	}
+
+	protected FilePosition(int line, int column) {
```
timtebeek, line 125: Wasn't particularly pleased about having to make these constructors `protected` for use from `org.junit.platform.engine.support.descriptor.FilePosition` constructors, but saw no way around it. Is there any way to discourage usage through `@API` annotations?

platform-tests/src/test/java/org/junit/platform/engine/discovery/FilePositionTests.java
```diff
@@ -31,12 +32,7 @@
  * @since 1.0
  */
 @DisplayName("FilePosition unit tests")
-class FilePositionTests extends AbstractTestSourceTests {
-
-	@Override
-	Stream<FilePosition> createSerializableInstances() {
-		return Stream.of(FilePosition.from(42, 99));
-	}
```
timtebeek, line 39: `AbstractTestSourceTests` is only accessible from `org.junit.platform.engine.support.descriptor`, so I though it preferable to keep that that way. Should this particular override and the functionality it covered be reproduced here? Or should I not yet have moved this tests class at this time?

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/FilePosition.java
```diff
@@ -0,0 +1,178 @@
...
+import org.junit.platform.commons.util.StringUtils;
+import org.junit.platform.commons.util.ToStringBuilder;
+
+/**
+ * Position inside a file represented by {@linkplain #getLine line} and
+ * {@linkplain #getColumn column} numbers.
+ *
+ * @since 1.0
+ */
+@API(status = STABLE, since = "1.0")
```
timtebeek, line 32: Not particularly sure about the `since` argument here, and/or whether to declare this as `STABLE` already, given that we likely want to change the constructors from `protected` to `private`.

mpkorstanje: You can also apply the annotation to specific methods to mark them as internal. The most specific annotation applies.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
```diff
@@ -10,32 +10,35 @@
...
 /**
  * Position inside a file represented by {@linkplain #getLine line} and
  * {@linkplain #getColumn column} numbers.
  *
  * @since 1.0
  */
-@API(status = STABLE, since = "1.0")
-public class FilePosition implements Serializable {
+@API(status = Status.DEPRECATED, since = "5.7")
+public class FilePosition extends org.junit.platform.engine.discovery.FilePosition {
```
mpkorstanje, line 28: IDEs won't pick up on the @API annotation. Adding `@Deprecated` might be useful here.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
```diff
@@ -10,32 +10,35 @@
...
 
 	private static final long serialVersionUID = 1L;
```
mpkorstanje, line 30: Where are file positions de-serialized too?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
```diff
@@ -77,7 +80,7 @@ public static FilePosition from(int line, int column) {
 	 * @see #from(int)
 	 * @see #from(int, int)
 	 */
-	public static Optional<FilePosition> fromQuery(String query) {
+	public static Optional<org.junit.platform.engine.discovery.FilePosition> fromQuery(String query) {
```
mpkorstanje, line 83: Returning the super class here will break existing callers.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/ClasspathResourceSelector.java
```diff
@@ -59,6 +65,14 @@ public String getClasspathResourceName() {
 		return this.classpathResourceName;
 	}
 
+	/**
+	 * Get the selected position within the classpath resource as a
+	 * {@link FilePosition}.
+	 */
+	public FilePosition getPosition() {
```
mpkorstanje, line 72: JUnit 5 seems to follow the pattern that nullable return values are always wrapped in an Optional.

https://github.com/junit-team/junit5/blob/3f7fed61f2edcee2a5551dbde75382b1dbe21267/junit-platform-console/src/main/java/org/junit/platform/console/tasks/TreeNode.java#L74-L76

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
marcphilipp: We don't use the serialization ourselves. It's meant for third-party tools that send test events over JVM boundaries. Since we didn't change any fields, I think we should be fine.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
timtebeek: This is slightly embarrassing, but when I change it to `FilePosition` it refuses to compile:
```
> Task :junit-platform-engine:compileJava FAILED
/junit5/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java:85: error: fromQuery(String) in org.junit.platform.engine.support.descriptor.FilePosition cannot hide fromQuery(String) in org.junit.platform.engine.discovery.FilePosition
        public static Optional<FilePosition> fromQuery(String query) {
                                             ^
  return type Optional<org.junit.platform.engine.support.descriptor.FilePosition> is not compatible with Optional<org.junit.platform.engine.discovery.FilePosition>
1 error
``` 
I fail to understand why this is only an issue for this static method and not the other two static `from` methods that take one or more `int` arguments. Any suggestions on what to do here?

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -135,6 +135,52 @@ public static FileSelector selectFile(File file) {
...
+	 *
+	 * @param path the path to the file to select; never {@code null} or blank
+	 * @param line the line number; must be greater than zero
+	 * @param column the column number; must be greater than zero
+	 * @see FileSelector
+	 * @see #selectFile(File, int, int)
+	 * @see #selectDirectory(String)
+	 * @see #selectDirectory(File)
+	 */
+	public static FileSelector selectFile(String path, int line, int column) {
```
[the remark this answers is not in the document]
timtebeek: Had originally not wanted to go with `selectFile(path, FilePosition.from(12, 34))` as to me it seemed slightly odd to ask the caller to invoke two methods to construct a selector. But I can perfectly live with the change as suggested, and have applied it already.

timtebeek:
Ok I've hopefully addressed all of your suggestions as intended; I hope it's ok I marked them as "resolved"; If I shouldn't have done that please let me know!

The only outstanding issue (I think) is this slight gaffe on my part where I need some direction: https://github.com/junit-team/junit5/pull/2253/files#r406852337

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FilePosition.java
timtebeek: I've briefly explored using `Optional<? extends FilePosition>` in the superclass `FilePosition`, but that leaks out into usage of that method regrettably. :\

mpkorstanje: A method can hide a method in the parent class if it follows he Liskov Substitution Principle. So a subclass must return a more specific type then the parent class. For generics this inversed.

Suppose it would be possible to do extend a class and return a more specialized generic:

```java
class Parent {
    public List<Number> funX() {
        return new ArrayList<>();
    }
}

class Child extends Parent {
    public List<Double> funX() {
        return new SpecializedDoubleList<>();
    }
}
```

Then by invoking `funX` on the parent type we could get a list that is only intended to contain `Double` values and add an `Integer` to it.

```java
Parent p = new Child();
List<Number> numbers = p.funX();
numbers.add(Integer.valueOf(1));
```
So the problem is that generic type is too specific. It specifies both the elements that can be added to the list (as a consumer) as well as those that can be taken from it (as a producer). By changing the parent to `extends Number` we can specify that the list contains an unknown type of element, but if you were to taken elements from it, it would at-least extend `Number`.

```java
class Parent {
    public List<? extends Number> funX() {
        return new ArrayList<>();
    }
}
```

mpkorstanje: > I've briefly explored using Optional<? extends FilePosition> in the superclass FilePosition, but that leaks out into usage of that method regrettably. :\

It isn't pretty but I think that is okay. 

It is possible to assign `List<Number>` to `List<? extends Number>` so once the old `FilePosition` and the `extends Number` are removed it shouldn't break anything.

timtebeek: Applied in https://github.com/junit-team/junit5/pull/2253/commits/decb66931b828437d53497fcf091ba3209f36774 as mentioned above; probably best for @marcphilipp to have another close look :)

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/ClasspathResourceSource.java
```diff
@@ -21,6 +21,7 @@
 import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.commons.util.ToStringBuilder;
 import org.junit.platform.engine.TestSource;
+import org.junit.platform.engine.discovery.FilePosition;
```
marcphilipp, line 24: This is a breaking change (same above). I'm afraid we'll have to leave the existing `getPosition()` method in place, deprecate it, and add a new `getFilePosition` method that returns an optional of the new type.

timtebeek: Good that you caught that one! I've hopefully make the correct changes in https://github.com/junit-team/junit5/pull/2253/commits/a768cf46290e40cc2cd17b2baee630667b2e3b2e ; Could you have a look if this was what you had intended?

timtebeek: Also deprecated the old methods in: https://github.com/junit-team/junit5/pull/2253/commits/dd3d8f7a41064a1daadad7b1da9f07b8d67f5170

marcphilipp: Thanks for the quick turnaround! I'm afraid there's more, though. The old factory methods in the `Source` classes need to be preserved and deprecated, too.

timtebeek: Sure thing: https://github.com/junit-team/junit5/pull/2253/commits/f80fcb92f6b60b4400548f2723e0839e0fe145ea

timtebeek: Hmm; maybe I was too quick with my change; should this apply to `ClassSource` as well?

timtebeek: Figured `ClassSource` probably needed a restore as well, so covered in https://github.com/junit-team/junit5/pull/2253/commits/147ea9d10ebc4b9e6fc9504d230b104c4bd9c70c

timtebeek:
@marcphilipp what would be needed to finish this PR? I'm guessing at least a change to the changelog; anything else?

marcphilipp:
@timtebeek I think this looks good! I'll discuss our decision wrt. to the package with the team and will get back to you afterwards.

timtebeek:
Just noticed https://github.com/junit-team/junit5/commit/19785134f95ba100346eb2d8b0c24f707535ebe4 in master; with this PR added to the [5.7 M2](https://github.com/junit-team/junit5/milestone/49) milestone, should I merge in master and add a rough outline for the changes contained in this PR?

I know this isn't a high priority new feature or fix, but would nonetheless be nice to see this finished if possible, before any other branches introduce conflicts. :)

marcphilipp:
@timtebeek Sorry for the delay. We discussed it briefly in the last team call. The current solution still contains breaking changes so we were a bit torn whether to just duplicate the `FilePosition` class to avoid the messy (not your fault!) deprecation work and the confusion caused by it. Would you mind changing this PR so the `FilePosition` class and its tests are copied and the deprecation is reverted? Sorry for the back ans forth! As for the release notes: yes, please rebase on `master` and add a bullet point.

timtebeek:
@marcphilipp No problem at all; I've reverted the deprecation changes, created a clean copy of FilePosition and updated the release notes. Could you have another look? :)

timtebeek:
@marcphilipp any update on this issue? Just trying to keep this flowing through; if there's any changes I have to make still let me know!

timtebeek:
@juliette-derancourt Thanks a lot for the detailed feedback; I've applied all your suggestions. Let me know if there's anything else! :)

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -135,6 +135,50 @@ public static FileSelector selectFile(File file) {
...
+	/**
+	 * Create a {@code FileSelector} for the supplied file path.
+	 *
+	 * <p>This method selects the file using the supplied path <em>as is</em>,
+	 * without verifying if the file exists.
+	 *
+	 * @param path the path to the file to select; never {@code null} or blank
+	 * @param position the position inside the file; may be {@code null}
+	 * @see FileSelector
+	 * @see #selectFile(File, int, int)
```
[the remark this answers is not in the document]
timtebeek: I've specifically added references to the methods that take one fewer/extra argument, or the same amount of arguments with a single argument of a different type. Hope that's what you intended as well! :)

platform-tests/src/test/java/org/junit/platform/engine/discovery/DiscoverySelectorsTests.java
```diff
@@ -153,6 +186,30 @@ void selectClasspathResources() {
 		// standard use case
 		selector = selectClasspathResource("A/B/C/spec.json");
 		assertEquals("A/B/C/spec.json", selector.getClasspathResourceName());
+
+		// with file position
+		selector = selectClasspathResource("A/B/C/spec.json", FilePosition.from(12, 34));
+		assertEquals("A/B/C/spec.json", selector.getClasspathResourceName());
+		assertEquals(FilePosition.from(12, 34), selector.getPosition().get());
```
timtebeek, line 193:
```suggestion
```

juliette-derancourt:
Thanks for the PR @timtebeek!

PR #2270, opened 2020-04-23
Execute explicitly selected classes regardless of class name filters

### LINKED ISSUE #2259: ConsoleLauncher: warn when using --select-class with a class that is not accepted by the --include-classname rules
When using the ConsoleLauncher to execute tests in a class FirstTesterMon.java, selected by the option --select-class, no tests were found. This is correct, as I had not specified any --include-classname nonstandard rule. 

However, it would have been really helpful to get a warning message in the style of:
WARNING: you have specified to scan a class FirstTesterMon; however this class will NOT be scanned due to the rules for classname scanning. Consider specifying a --include-classname rule.

Or alternatively, and perhaps more sane: if a class is specified using --select-class always scan it even if the name does not match the --include-classname rules.

juliette-derancourt:
## Overview

Add explicitly selected classes from `--select-class` and `--select-method` to the include patterns passed when constructing the `ClassNameFilter` so that these are always executed regardless of the implicit or explicit class name filter.

Resolves #2259.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
@@ -23,7 +23,9 @@ on GitHub.
 
 ==== New Features and Improvements
 
-* ❓
+* When using `ConsoleLauncher`, explicitly selected classes from `--select-class`
+  and `--select-method` are now always executed regardless of class name patterns
+  provided with `--include-classname`.
 
 
 [[release-notes-5.7.0-M2-junit-jupiter]]
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
@@ -25,7 +25,7 @@ on GitHub.
 
 * When using `ConsoleLauncher`, explicitly selected classes from `--select-class`
   and `--select-method` are now always executed regardless of class name patterns
-  provided with `--include-classname`.
+  provided with `--include-classname` or the default class name pattern.
 
 
 [[release-notes-5.7.0-M2-junit-jupiter]]
--- a/junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
@@ -27,7 +27,9 @@ import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
+import java.util.function.Function;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import org.junit.platform.commons.util.ModuleUtils;
 import org.junit.platform.commons.util.Preconditions;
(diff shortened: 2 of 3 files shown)


platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
```diff
@@ -282,6 +298,10 @@ private LauncherDiscoveryRequest convert() {
 		return creator.toDiscoveryRequest(options);
 	}
 
+	private void assertFilters(Filter<?> filter, String filteredElement) {
```
juliette-derancourt, line 301: Not sure if this is a good idea... Might need a better name 🤔

documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
```diff
@@ -23,7 +23,9 @@ on GitHub.
 
 ==== New Features and Improvements
 
-* ❓
+* When using `ConsoleLauncher`, explicitly selected classes from `--select-class`
+  and `--select-method` are now always executed regardless of class name patterns
+  provided with `--include-classname`.
```
marcphilipp, line 28:
```suggestion
  provided with `--include-classname` or the default class name pattern.
```

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
```diff
@@ -127,4 +129,14 @@ private void addFilters(LauncherDiscoveryRequestBuilder requestBuilder, CommandL
...
 
+	private ClassNameFilter includedClassNamePatterns(CommandLineOptions patterns) {
+		List<String> classNamePatterns = new ArrayList<>();
+		classNamePatterns.addAll(patterns.getIncludedClassNamePatterns());
+		classNamePatterns.addAll(patterns.getSelectedClasses());
+		classNamePatterns.addAll(patterns.getSelectedMethods().stream() //
+				.map(name -> ReflectionUtils.parseFullyQualifiedMethodName(name)[0]) //
+				.collect(Collectors.toList()));
+		return includeClassNamePatterns(classNamePatterns.toArray(new String[0]));
+	}
```
marcphilipp, line 140: We should quote the literal class names using `Pattern.quote`, e.g. like this:

```suggestion
	private ClassNameFilter includedClassNamePatterns(CommandLineOptions options) {
		Stream<Stream<String>> patternStreams = Stream.of( //
			options.getIncludedClassNamePatterns().stream(), //
			options.getSelectedClasses().stream() //
					.map(Pattern::quote), //
			options.getSelectedMethods().stream() //
					.map(name -> ReflectionUtils.parseFullyQualifiedMethodName(name)[0]) //
					.map(Pattern::quote));
		return includeClassNamePatterns(patternStreams //
				.flatMap(Function.identity()) //
				.toArray(String[]::new));
	}
```

platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
marcphilipp: `assertFilterDescriptionContains`?

Should we rather call the filter and check it includes/excludes the expected names?

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
juliette-derancourt: Haven't thought of that, thanks! 👍

platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
juliette-derancourt: Indeed!

PR #2271, opened 2020-04-24
Add Flight Recorder based execution listener

### LINKED ISSUE #2272: Add Flight Recording Test Execution Listener
## Flight Recording Test Execution Listener

Implement a `TestExecutionListener` that sends Java Flight Recorder events.

Motivation: https://twitter.com/gunnarmorling/status/1270285860626403329?s=20
> Essentially, where I currently ask users "Can you increase your log level to DEBUG and share logs from component xyz", I'd like to use JFR.

### Links and Resources

- [Flight Recorder API Programmer’s Guide](https://docs.oracle.com/en/java/javase/14/jfapi)
- [Repositories related to Flight Recorder](https://github.com/flight-recorder)
- [JEP 328: Flight Recorder](https://openjdk.java.net/jeps/328)
- [JEP 349: JFR Event Streaming](https://openjdk.java.net/jeps/349)
- [2020 DevDotNext Digital Edition - Continuous Monitoring With JFR](https://www.youtube.com/watch?v=E9K5m1HXMSc) 📹
- [Using JDK FlightRecorder and JDK Mission Control](https://blog.arkey.fr/2020/06/28/using-jdk-flight-recorder-and-jdk-mission-control/) by Brice Dutheil

## Deliverables

- [x] Spike a JFR-based `TestExectionListener` implementation PR #2271 
- [x] Get spike to build and run it on Java 8, 11, and 14
- [x] Decide which events to commit: test only, containers, dynamic test registration, ...
- [x] Decide which data to send: unique id, display names, result
- [x] Decide which metadata to supply: category, name, label

sormuras:
## Overview

Resolves #2272

---

---

### Definition of Done


sormuras:
Hi @egahlin and @mgronlun -- after successfully prototyping a test execution listener that fires JFR events, I a) tweeted about it https://twitter.com/sormuras/status/1252113877699084292, b) threw it away, and c) started this new spike that should land in the coming `1.7.0-M2` release of JUnit Platform.

Please, can you take a look at my second attempt of `? extends jdk.jfr.Event` classes and how those custom events are created, composed, and commited in [FlightRecordingListener.java]?

Thanks in advance!

[FlightRecordingListener.java]: https://github.com/junit-team/junit5/pull/2271/files#diff-f1f1cf74ae9fe9cda577a83a2733c451

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -0,0 +1,131 @@
...
+	public FlightRecordingListener() {
+		this.testPlanExecutionEvent = new AtomicReference<>();
+		this.testExecutionEventMap = new ConcurrentHashMap<>();
+	}
+
+	@Override
+	public void testPlanExecutionStarted(TestPlan plan) {
+		TestPlanExecutionEvent event = new TestPlanExecutionEvent();
+		event.containsTests = plan.containsTests();
+		event.roots = plan.getRoots().stream().map(TestIdentifier::getUniqueId).collect(Collectors.joining("\t"));
```
egahlin, line 53: If this is a path, consider using "/" as a separator, i.e. "id1 / id2 / id3". Tabs may result in different number of spaces, or may not work at all, if visualized in a GUI.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -0,0 +1,131 @@
...
+		testPlanExecutionEvent.set(event);
+		event.begin();
+	}
+
+	@Override
+	public void testPlanExecutionFinished(TestPlan plan) {
+		TestPlanExecutionEvent event = testPlanExecutionEvent.get();
+		event.end();
```
egahlin, line 61: No need to call end(). It is implicit in commit()

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -0,0 +1,131 @@
...
+
+	@Category("JUnit")
+	static abstract class TestEvent extends Event {
+		@Label("UniqueId")
+		String uniqueId;
+		@Label("Display Name")
+		String displayName;
+	}
+
+	@Description("Dynamic Test Registration")
```
egahlin, line 116: Should be label

egahlin, line 116: Missing Category and Name("org.junit.DynamicTestRegistration")

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -0,0 +1,131 @@
...
+		event.displayName = test.getDisplayName();
+		event.reason = reason;
+		event.commit();
+	}
+
+	@Override
+	public void executionStarted(TestIdentifier test) {
+		TestExecutionEvent event = new TestExecutionEvent();
+		event.begin();
+		event.uniqueId = test.getUniqueId();
```
egahlin, line 86: Set the fields before timing starts with begin()

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -0,0 +1,131 @@
...
+ */
+@API(status = EXPERIMENTAL, since = "1.7")
+public class FlightRecordingListener implements TestExecutionListener {
+
+	private final AtomicReference<TestPlanExecutionEvent> testPlanExecutionEvent;
+	private final Map<String, TestExecutionEvent> testExecutionEventMap;
+
+	public FlightRecordingListener() {
+		this.testPlanExecutionEvent = new AtomicReference<>();
+		this.testExecutionEventMap = new ConcurrentHashMap<>();
```
egahlin, line 46: Why not assign field directly, i.e. testExecutionEventMap = new ConcurrentHashMap<>()?

sormuras:
> There is more, but I made a first pass.

Thanks a lot. Applying most suggested changes and answer your question inline.

Still struggling to convince Gradle to compile this PR. The module/package `jdk.jfr` is somehow hidden from its sight. Already tried with adding `--add-modules jdk.jfr` but to no avail. @marcphilipp will take a look later.

As JUnit Platform targets Java 1.8 as well, we'll need to guard this listener implementation from being registered on a 1.8 runtime. Right?

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: I like field declarations to be separated from their assignment.

The constructor is a good place to do that work, which not much here, yet. But you never know.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -0,0 +1,131 @@
...
+ * A {@link TestExecutionListener} that generates Java Flight Recorder
+ * events.
+ *
+ * @see <a href="https://openjdk.java.net/jeps/328">JEP 328: Flight Recorder</a>
+ * @since 1.7
+ */
+@API(status = EXPERIMENTAL, since = "1.7")
+public class FlightRecordingListener implements TestExecutionListener {
+
+	private final AtomicReference<TestPlanExecutionEvent> testPlanExecutionEvent;
```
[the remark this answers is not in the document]
sormuras: I like the similarity to the map declared next to it: an atomic reference is a single-entry map

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: The format defined by `UniqueId` uses `/` as an element separator char -- so, that's already taken. See details at:

https://github.com/junit-team/junit5/blob/19785134f95ba100346eb2d8b0c24f707535ebe4/junit-platform-engine/src/main/java/org/junit/platform/engine/UniqueIdFormat.java#L42

Using tab `\t` is not a good choice either. Guess, we need to define a human-visible, URI-encodeable unique-id separator char.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: Is `@Category` not inherited?

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: Using `#` for the time being.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -0,0 +1,138 @@
...
+		SkippedTestEvent event = new SkippedTestEvent();
+		event.uniqueId = test.getUniqueId();
+		event.displayName = test.getDisplayName();
+		event.reason = reason;
+		event.commit();
+	}
+
+	@Override
+	public void executionStarted(TestIdentifier test) {
+		TestExecutionEvent event = new TestExecutionEvent();
```
marcphilipp, line 84: ❓ Does JFR support explicit nesting of events?

junit-platform-launcher/src/module/org.junit.platform.launcher/module-info.java
```diff
@@ -30,4 +31,7 @@
 
 	uses org.junit.platform.engine.TestEngine;
 	uses org.junit.platform.launcher.TestExecutionListener;
+
+	provides org.junit.platform.launcher.TestExecutionListener
+			with org.junit.platform.launcher.listeners.FlightRecordingListener;
```
marcphilipp, line 36: This should be declared in `META-INF/services` so it works on the classpath as well, shouldn't it?

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
egahlin: Yes, but it is usually good to avoid from a visualization point of view.

This image from JMC shows what happens in the JVM (thread sleep, lock contention etc) using events from JFR. A JUnit test event would be visualized on separate lane next to the JVM event so you can see for what made a test slow, i.e a call to Thread.sleep(). 

Nested events are painted on top of each other, so it can be hard to see. Preferably only the actual test (the leaf nodes) should be represented by the Test event. If you want to have information about test containers/groups, it would be better to separate them into their own events, or perhaps leave them out completely.

https://docs.oracle.com/javase/10/troubleshoot/img/jfr-eventgraph.png

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: It is inherited.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: > [...] If you want to have information about test containers/groups, it would be better to separate them into their own events, or perhaps leave them out completely.

Good idea.

Candidates to be left out:
- Test Plan
- Dynamic Test Registration
- Test Container

They don't contribute much to a better understanding what happened in a test run. Also, the recorded "Unique Id" field already contains the hierarchical information of a test and its parents.

junit-platform-launcher/src/module/org.junit.platform.launcher/module-info.java
sormuras: Just wanted to start small: [ModularUserGuideTests](https://github.com/junit-team/junit5/blob/c7dc07b655eed11f532e983f6c178b1b02bb457b/platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java#L110)

Still, I think, we should register this listener by default, i.e. programmatically. Hard-coded into the constructor of: `TestExecutionListenerRegistry`

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: Why assign a field directly? Performance?

sormuras:
Yeah. Build is green...

...yet, module `jdk.jfr` isn't found at compile time. Which is strange. And why do we have to remove `--release` from compiler arguments?!

junit-platform-launcher/src/module/org.junit.platform.launcher/module-info.java
sormuras: It should be replaced adding this listener programmatically by default.

This is just temporary hook to get this spike emitting some data.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
marcphilipp: Less code.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -112,27 +120,38 @@ public void executionFinished(TestIdentifier test, TestExecutionResult result) {
...
 
 	@Name("org.junit.SkippedTest")
+	@Label("Skipped Test")
 	static class SkippedTestEvent extends TestEvent {
 		@Label("Reason")
 		String reason;
 	}
 
 	@Name("org.junit.TestExecution")
+	@Label("Test")
```
egahlin, line 142: @Category("JUnit"

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -112,27 +120,38 @@ public void executionFinished(TestIdentifier test, TestExecutionResult result) {
...
 		String displayName;
 	}
 
 	@Name("org.junit.DynamicTestRegistration")
+	@Label("Dynamic Test Registration")
 	static class DynamicTestEvent extends TestEvent {
 	}
 
 	@Name("org.junit.SkippedTest")
+	@Label("Skipped Test")
```
egahlin, line 135: @Category("JUnit"

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -112,27 +120,38 @@ public void executionFinished(TestIdentifier test, TestExecutionResult result) {
...
-
 		@Label("Unique Id")
 		String uniqueId;
-
 		@Label("Display Name")
 		String displayName;
 	}
 
 	@Name("org.junit.DynamicTestRegistration")
+	@Label("Dynamic Test Registration")
```
egahlin, line 130: I don't think stack traces are useful for these events as they would only point back into the test framework. Adding the annotation @StackTrace(false) will turn them off.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
egahlin: I would leave them out for now. It will reduce the noise. If there is a need later on, they can be added.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -112,27 +120,38 @@ public void executionFinished(TestIdentifier test, TestExecutionResult result) {
...
+	@Label("Test")
 	static class TestExecutionEvent extends TestEvent {
 		@Label("Result")
 		String result;
 	}
+
+	@Category("JUnit")
+	@Name("org.junit.ReportEntry")
+	@Label("Report Entry")
+	static class ReportEntryEvent extends Event {
```
egahlin, line 151: How often is a report emitted, once per test? What would it typically contain? There may be other ways to structure the data, but I need more information.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: To make it explicit? As `@Category` is already inherited from the super class.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: Good point -- really need to read https://docs.oracle.com/en/java/javase/14/jfapi/create-events.html more carefully. That entire event class might be dropped in the near future -- as it is not relevant for test result recording anyway.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
egahlin: Sorry. I missed that. I'm not used to GitHub, is there a way I can see the files in their latest state without comments (on the web)?

sormuras: Thought there was a filter for that, but couldn't find it.

Try "View file..." in this drop down menu:
image

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: This is an event triggered by a user or an extension within the lifecycle of a test.

Here's an example:
https://github.com/junit-team/junit5/blob/19785134f95ba100346eb2d8b0c24f707535ebe4/documentation/src/test/java/example/TestReporterDemo.java#L25-L42

sormuras: Here's the API documentation for `TestReporter`: https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/TestReporter.html

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
egahlin: Thanks!

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
sormuras: Reduced the scope to "Test Plan" and "Test" (w/o containers) events for the time being.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -76,74 +78,80 @@ public void executionSkipped(TestIdentifier test, String reason) {
...
 
 	@Override
 	public void executionFinished(TestIdentifier test, TestExecutionResult result) {
-		TestExecutionEvent event = testExecutionEventMap.get(test.getUniqueId());
+		if (test.isContainer() && result.getStatus().equals(TestExecutionResult.Status.SUCCESSFUL)) {
+			return;
+		}
+		TestExecutionEvent event = testExecutionEventMap.get(test.getUniqueId()); // TODO Remove?
 		event.end();
+		event.reports = event.reportEntries == null ? null : event.reportEntries.toString();
```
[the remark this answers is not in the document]
sormuras: Understood.

- Is type `String` also a good choice for an ID? That's the one the JUnit Platform already provides.
- Do tools like JMC display those relations?

I'll keep this normalization advice for later, as report entries are always strongly coupled to a test execution frame/event. And 99% or more of todays' tests don't publish any report entries anyway. Thus, the field is almost always `null`.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -76,74 +78,80 @@ public void executionSkipped(TestIdentifier test, String reason) {
...
 		event.result = result.getStatus().toString();
+		event.throwable = result.getThrowable().map(Throwable::getMessage).orElse(null); // TODO Include stacktrace?
```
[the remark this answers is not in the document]
sormuras: Check. Will rename the existing field and extend the event class with field `exceptionClass`.

sormuras:
@egahlin Some of the event-related design choices are driven by the attempt to use JFR as the basis for a unified reporting format. That means, the event data collected by should suffice to generate something similar to well-known `TEST-${name}.xml` files of today.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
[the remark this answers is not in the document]
egahlin: String works as long it is unique (for a recording).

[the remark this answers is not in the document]
egahlin: JMC will display all fields and is aware of relations. They can be used when creating custom tables, i.e group by testId.

relational:
I don't think I am an intended recipient of this email.

best,
Eirrikur

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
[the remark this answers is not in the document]
sormuras: For long-running tests (like scenario/UI tests) it is better to have distinct report event fired while the event is still running.
Thus, I'll implement the suggested normalization.

sormuras:
https://twitter.com/sormuras/status/1255842464952463360 generated from this
[user-guide.zip](https://github.com/junit-team/junit5/files/4558393/user-guide.zip) JFR recording.

junit-platform-launcher/src/main/java9/org/junit/platform/launcher/listeners/FlightRecordingListener.java
```diff
@@ -0,0 +1,180 @@
...
+	@StackTrace(false)
+	static class TestPlanExecutionEvent extends Event {
+		@Label("Contains Tests")
+		boolean containsTests;
+		@Label("Engine Names")
+		String engineNames;
+	}
+
+	@MetadataDefinition
+	@Relational
```
egahlin, line 122: Set a name for the annotation, i.e @Name("org.junit.UniqueId")

sormuras: Sure.

Removed the name annotation from the metadata definition because it isn't inherited:

```java
@Target({ ElementType.TYPE, ElementType.FIELD, ElementType.METHOD})
@MetadataDefinition
public @interface Name { ...
```

sormuras:
Convinced Gradle to build module `org.junit.platform.launcher@1.7.0-SNAPSHOT`. Its modular API printed via `jar --describe-module --file junit-platform-launcher-1.7.0-SNAPSHOT.jar` reads:

```
exports org.junit.platform.launcher
exports org.junit.platform.launcher.core
exports org.junit.platform.launcher.listeners
exports org.junit.platform.launcher.listeners.discovery

requires java.base mandated
requires java.logging transitive
requires jdk.jfr
requires org.apiguardian.api transitive
requires org.junit.platform.commons transitive
requires org.junit.platform.engine transitive

uses org.junit.platform.engine.TestEngine
uses org.junit.platform.launcher.TestExecutionListener

contains org.junit.platform.launcher.tagexpression
```

buildSrc/src/main/kotlin/java-library-conventions.gradle.kts
```diff
@@ -145,7 +145,7 @@ val compileModule by tasks.registering(JavaCompile::class) {
 	source = fileTree(moduleSourceDir)
 	destinationDir = moduleOutputDir
 	sourceCompatibility = "9"
-	targetCompatibility = "9"
+	targetCompatibility = "11"
```
sormuras, line 148: Because `junit-platform-jfr` targets Java 11 -- will be reverted when we turn it into a MR-JAR with base 8.

documentation/documentation.gradle.kts
```diff
@@ -18,7 +18,7 @@ modularProjects.forEach { evaluationDependsOn(it.path) }
 
 javaLibrary {
 	mainJavaVersion = JavaVersion.VERSION_1_8
-	testJavaVersion = JavaVersion.VERSION_1_8
+	testJavaVersion = JavaVersion.VERSION_11
```
sormuras, line 21: Because junit-platform-jfr targets Java 11 -- will be reverted when we turn it into a MR-JAR with base 8.

junit-platform-jfr/junit-platform-jfr.gradle.kts
```diff
@@ -0,0 +1,23 @@
...
+	api(project(":junit-platform-launcher"))
+}
+
+javaLibrary {
+	mainJavaVersion = JavaVersion.VERSION_11
+}
+
+tasks {
+	verifyOSGi {
+		enabled=false
```
sormuras, line 21: When enabled, it yields:

```
Resolution failed. Capabilities satisfying the following requirements could not be found:
    [<<INITIAL>>]
      ? osgi.identity: (osgi.identity=junit-platform-jfr)
          ? [junit-platform-jfr version=1.7.0.SNAPSHOT]
              ? osgi.wiring.package: (&(osgi.wiring.package=jdk.jfr))
```

junit-platform-launcher/junit-platform-launcher.gradle.kts
```diff
@@ -1,5 +1,7 @@
 plugins {
 	`java-library-conventions`
+	`java-multi-release-sources`
+	`java-repackage-jars`
```
sormuras, line 4: Gonna revert this. No longer needed, methinks.

src/checkstyle/suppressions.xml
```diff
@@ -4,4 +4,6 @@
 		files="junit-platform-commons[\\/]src[\\/]main[\\/]java9[\\/]org[\\/]junit[\\/]platform[\\/]commons[\\/]util[\\/]*"/>
 	<suppress checks="JavadocPackage"
 		files="junit-platform-console[\\/]src[\\/]main[\\/]java9[\\/]org[\\/]junit[\\/]platform[\\/]console[\\/]*"/>
+	<suppress checks="JavadocPackage"
+		files="junit-platform-launcher[\\/]src[\\/]main[\\/]java9[\\/]org[\\/]junit[\\/]platform[\\/]launcher[\\/]listeners[\\/]*"/>
```
sormuras, line 8: Gonna revert this. No longer needed, methinks.

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
```diff
@@ -108,6 +108,8 @@
 		var projectDir = Path.of("../documentation");
 		command.add(Path.of(System.getProperty("java.home"), "bin", "java").toString());
 
+		command.add("-XX:StartFlightRecording:filename=" + temp.resolve("user-guide.jfr"));
+
```
sormuras, line 112: Convert this into a real integration test.

junit-platform-jfr/junit-platform-jfr.gradle.kts
marcphilipp: @rotty3000 @bjhargrave Do you know what's wrong here?

junit-platform-jfr/src/module/org.junit.platform.jfr/module-info.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Defines the JUnit Platform JAR API.
+ *
+ * @since 1.7
+ */
+module org.junit.platform.jfr {
+	requires jdk.jfr;
+	requires transitive org.apiguardian.api;
+	requires org.junit.platform.commons;
```
marcphilipp, line 19: I don't see commons in the dependencies in the buildscript nor used in the source code. Can that be removed from here?

documentation/documentation.gradle.kts
marcphilipp: Fixed via an extra configuration.

junit-platform-jfr/junit-platform-jfr.gradle.kts
bjhargrave: It is detecting a reference to the package `jdk.jfr` in the artifact `junit-platform-jfr`. And no artifacts in the resolution check which exports the package `jdk.jfr`.

I will let @rotty3000 suggest a solution as he provided the initial verifyOSGi task.

junit-platform-jfr/src/module/org.junit.platform.jfr/module-info.java
sormuras: Think so, yes.

Also the `transitive` modifiers are not needed here -- as no package is exported.

junit-platform-jfr/junit-platform-jfr.gradle.kts
marcphilipp:
```suggestion
		enabled = false
```

rotty3000: I'm taking a look today.

bjhargrave: This may be due to Bnd not having the latest EE model properties which includes the new `jdk.jfr` package. We are just fixing that now in Bnd.

sormuras: Sounds like a potential cause - as other modules defined here that also don't export any package are working w/o special configuration directives. I guess.

bjhargrave: We will soon have a Bnd 5.1.1-SNAPSHOT build available that can be used to test if our fix solves this. I will let you know when it is available.

sormuras: Great! Take your time and looking forward to test-drive that build.

I might merge this PR soon -- with verifyOSGi disabled -- so, testing will happen on a new branch.

rotty3000: I've confirmed that the upcoming bnd 5.1.1 will address the issue here.

sormuras: Via this commit? https://github.com/bndtools/bnd/commit/7e295c95a80202ca245f51939ed631775d62ee83

bjhargrave: For 5.1.1, it would technically be https://github.com/bndtools/bnd/commit/95ab05ca689bc2a442e3ebc4cf1efdf521a16abb which is the above commit cherry-picked onto the next branch which is where 5.1.1 currently lives.

marcphilipp: @bjhargrave I got a notification that 5.1.1 has been released but it's not on Maven Central, yet. Why is that?

bjhargrave: I made the build last night and released the staging repo this morning (EDT). So it should now be available on Maven Central now.

sormuras: Already indexed via https://search.maven.org/artifact/biz.aQute.bnd/biz.aQute.bnd/5.1.1/jar

marcphilipp: Thanks, I just upgraded `master` to 5.1.1. @sormuras Could you please rebase this branch to verify that the issue is now resolved?

sormuras: Rebase done and force-pushed.

Verifying that it works with `verifyOSGi` enabled locally ... in progress.

sormuras: This still does not bode well

```text
> Task :junit-platform-jfr:verifyOSGi FAILED
Resolution failed. Capabilities satisfying the following requirements could not be found:
    [<<INITIAL>>]
      ? osgi.identity: (osgi.identity=junit-platform-jfr)
          ? [junit-platform-jfr version=1.7.0.SNAPSHOT]
              ? osgi.wiring.package: (osgi.wiring.package=jdk.jfr)

Execution failed for task ':junit-platform-jfr:verifyOSGi'.
> D:\dev\github\junit-team\junit5\junit-platform-jfr\build\verifyOSGiProperties.bndrun resolution exception
```

rotty3000: So I've just tested and there is only one minor tweak required to get past this now :)

in `buildSrc/src/main/kotlin/osgi-conventions.gradle.kts`

please make this small change:
```
@@ -65,11 +65,13 @@ val osgiPropertiesFile = file("$buildDir/verifyOSGiProperties.bndrun")
 // task writes out the properties necessary for it to verify the OSGi
 // metadata.
 val osgiProperties by tasks.registering(WriteProperties::class) {
 	outputFile = osgiPropertiesFile
 	property("-standalone", true)
-	property("-runee", "JavaSE-${Versions.jvmTarget}")
+	project.extensions.getByType(JavaLibraryExtension::class.java).let { javaLibrary ->
+		property("-runee", "JavaSE-${javaLibrary.mainJavaVersion}")
+	}
 	property("-runrequires", "osgi.identity;filter:='(osgi.identity=${project.name})'")
 	property("-runsystempackages", "jdk.internal.misc,sun.misc")
 }
 
 val osgiVerification by configurations.creating {

```
then finally:
```
@@ -18,9 +18,6 @@ javaLibrary {
 
 tasks {
 	compileModule {
 		targetCompatibility = "11"
 	}
-	verifyOSGi {
-		enabled = false
-	}
 }

```
With this all tests should now pass.

rotty3000: This issue is that the Java versioned used for the "verify" task should match the _intended target_ for that code. Since each project seems to be able to adjust it's base Java level, we need to make `verifyOSGi` aware of that as well :) This small change fixes that instead of using a global preset version.

rotty3000: If you would like I can push this as a PR to master branch including the update to 5.1.1 and then you can rebase?

marcphilipp: @rotty3000 Thanks for looking into this and providing such detailed instructions! I think we can take it from here. 👍 

That makes total sense. I'm afraid the error message was not very helpful for a non-OSGi user such as me. Is there a way to make it more verbose in case we run into a similar issue in the future?

rotty3000: You're right that the error message was not informative. Who said "Software is hard."
But to be very honest, there are very few projects _I've seen_ in the wild which actually build artifacts targeting different versions of Java in a single build and also building for OSGi, so most times a global setting is more than enough. I had to dig into gradle/kotlin DSL to even figure out how to get the project specific java version :D So what you have here is bleeding edge Java Library + OSGi build spec. But the bnd team is trying very hard to make the whole thing less obtuse and esoteric. Thanks for putting up with us :)

sormuras: Applied the "small change" as suggested, @rotty3000 -- and works locally. At least the OSGi verification check.

Now that the integration tests are started, they revealed another quirk: https://bugs.openjdk.java.net/browse/JDK-8232092?focusedCommentId=14352730&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14352730 -- but that's another story.

Thanks for your support, Raymond!

marcphilipp:
@sormuras Shouldn't happen again for that particular test: https://github.com/junit-team/junit5/commit/448d714ac270a312be66e9d60eaf45cfa2f5a4f0

Issue #2281, opened 2020-04-30
Support file system resources in @CsvFileSource


homberghp:
## Problem description

Test resources are strange beasts. The compiler does not know about them nor cares, so getting the resources into the class path in typical builds needs help from the build system (ant, maven, gradle). This leads to several problems

- running a single test file (e.g. a first after project clean or fresh project) will not work, unless a more or less complete build step is executed.
- copying the test file from one location to another without any transformation is a waste of time (and ssd wear).
- Modifying the test data and rerun the test requires a rebuild (remember the deployment of the test data to the (test) class path, so modifying just the test data and re-run the test is not enough.
- Test data should be considered input to the program, not a resource. A resource is a glorified constant, test data should be able to vary independent of program code, test or otherwise.

## Original Proposal
- Modify the InputStream creating method (see my remarks in #1970) to try file system first, then class path would be a sufficient change. This will create this feature very cheaply without changing current class path semantics.
- Update @CsvFileSource documentation to reflect this change since the change proposal in #1970 adapts a private method in the CsvFileArgumentsProvider class.

sbrannen:
As I stated in https://github.com/junit-team/junit5/issues/1970#issuecomment-621727776, we likely will **not** want to change default semantics since that would be a potentially breaking change.

Instead, adding support for looking in the file system would be "in addition" to the current classpath support. It should not "supersede" that support.

That's why I suggested a robust and backward compatible solution based on schemes (a.k.a., _prefixes_) such as `classpath:` and `file:`, where `classpath:` is the inferred default.

As a side note, I understand your preference for placing resources directly in the file system. There may be valid use cases for that. That's why I asked you to open this issue; however, I don't agree with the following.

> so getting the resources into the class path in typical builds needs help from the build system

sbrannen:
Tentatively slated for 5.7 M2 for _team discussion_

sbrannen:
As @sormuras mentioned in https://github.com/junit-team/junit5/issues/1970#issuecomment-621728986, supporting `modulepath:` as a scheme would also be an option.

homberghp:
Ultimately a URI or URL would be most flexible, but will introduce quite some effort to get it working.
Fetched test data from server might be useful in integration tests.

homberghp:
By the way, I want to compliment you on the quick response. Absolutely phenomenal.

homberghp:
I have just created a pull request with an implementation (including test and javadoc updates) on how I would envision it.
[...]

Issue #1970, opened 2019-07-30
@CsvFileSource cannot locate resource when used in base class


jbytecoder:
When using a relative resource path with `@CsvFileSource` — for example, `"test.csv"` as opposed to `"org/exmaple/test.csv"` — the resource is resolved against the runtime type of test class.

This is valid approach until we use inheritance.  If it happens that we have a base class test and subclass located in different packages then, the subclass test fails with cannot find resource. It runs normally if I copy the resource under same package as subclass. Obviously this is not an issue when using full path resources.  

I think that when source file is resolved instead of using runtime class, the class that **declares** the method annotated with `@CsvFileSource` should be used.  

## Steps to reproduce

In order to reproduce we need 2 classes in different packages, let's assume base class Foo.

```java
package org.example.foo;

import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.CsvFileSource;

public abstract class Foo {
    @ParameterizedTest
    @CsvFileSource(resources = "testEquals.csv",numLinesToSkip = 1)
    public void testEquals( String a, String b ) {
        Assertions.assertEquals(a,b);
    }
}
```

And then if we extend this class with Bar

```java
package org.example.bar;

import org.example.foo.Foo;

public class Bar extends Foo {
}
```

Then the Bar test fails, while being unable to locate resource. However if we place class `Bar` in `org.example.foo` package everything works fine.

## Context

The environment is jupiter-5.4 using gradle-5.4.1 with intelij idea 2019.2

marcphilipp:
Thanks for raising the issue! I tend to agree. Tentatively slated for 5.6 M1 for _team discussion_.

sbrannen:
I agree that this can be improved, and the change would be along the lines of what we do in the Spring TestContext Framework for various annotations such as `@ContextConfiguration`, `@Sql`, etc.

marcphilipp:
**Team Decision:** While we agree that it would be desirable to interpret such paths relative to the declaring class, it would be a considerable effort to rework our annotation search algorithm. Thus, we have assigned the `waiting-for-interest` label for now.

kevin0x90:
Hi,

i would like to work on this issue.

marcphilipp:
@kevin0x90 The implementation does not seem straightforward. I'd propose you start spiking potential solution and report back so we can discuss here, ok?

kevin0x90:
[...]

PR #2282, opened 2020-04-30
implemented #2281, including unit test enhancment and csvfilesource a…

homberghp:
## Overview

Enhance the way CsvFileSource resolves its resources by searching the file system for the named resource in addition to resolving it via the class loader of the test class, which is the current implementation. The following benefits will be provided:

- In case of sub classes in different packages (#1970) this will be a no brainer. In that case specifying the test class as `src/test/resources/testEquals.csv` would solve that whole issue.
- Resource resolution will have two steps: file system first, then default to class loader resolution. In case test cases are deployed by jar file, one gets the added benefit of being able to _overwrite_ the test data by supplying a file with the same path and file name as the test data file in the deployment artifact.
- Test data classes no longer need to be copied around if the file system resolving is used. This may be a minor issue unless files are big and/or you worry about storage hardware wear.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
@@ -23,7 +23,7 @@ on GitHub.
 
 ==== New Features and Improvements
 
-* ❓
+* File system path resolution for cvs file sources in addition to class path resolution.
 
 
 [[release-notes-5.7.0-M2-junit-jupiter]]
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1218,7 +1218,7 @@ of any custom values configured via the `nullValues` attribute.
 [[writing-tests-parameterized-tests-sources-CsvFileSource]]
 ===== @CsvFileSource
 
-`@CsvFileSource` lets you use CSV files from the classpath. Each line from a CSV file
+`@CsvFileSource` lets you use CSV files from the file system and from classpath. Each line from a CSV file
 results in one invocation of the parameterized test.
 
 The default delimiter is a comma (`,`), but you can use another character by setting the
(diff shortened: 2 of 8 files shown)


homberghp:
## Studying the scheme like proposal

- Considering to implement a URI like schema. Studying the URL and URI doc, I found that 
using proper url like https://junit5.org will work and produce a reasonable resource specifying string.
However, using the `file://` only allows full path names like '  file:/home/hom/NetBeansProjects/uriresolution/target/classes/u/u.csv' which of course would be a **no go**, because that would bind the project to a specific file system organization. The class loader resolution which is still intact will deal with the class path, presumably the module path and the jar path too. 

- The order of evaluation might need some consideration because in case of having two resources on in the project with the same name will give the first one evaluated precedence. On the business side of a project, I could imagine that this leads to unsafe situations because a local file could overwrite a constant in a jar. On the other hand, having an almost automatic overwrite mechanism in the testing part of a project might be beneficial. In that case, giving file system resolution precedence over class loader might be acceptable.

homberghp:
## Studying the scheme like proposal

- Considering to implement a URI like schema. Studying the URL and URI doc, I found that 
using proper url like https://junit5.org will work and produce a reasonable resource specifying string.
However, using the `file://` only allows full path names like '  file:/home/hom/NetBeansProjects/uriresolution/target/classes/u/u.csv' which of course would be a **no go**, because that would bind the project to a specific file system organization. The class loader resolution which is still intact will deal with the class path, presumably the module path and the jar path too. 

- The order of evaluation might need some consideration because in case of having two resources on in the project with the same name will give the first one evaluated precedence. On the business side of a project, I could imagine that this leads to unsafe situations because a local file could overwrite a constant in a jar. On the other hand, having an almost automatic overwrite mechanism in the testing part of a project might be beneficial. In that case, giving file system resolution precedence over class loader might be acceptable.

sbrannen:
Declined in favor of the team decision stated in https://github.com/junit-team/junit5/issues/2281#issuecomment-625757646.

PR #2283, opened 2020-05-01
Add possibility to register post-discovery filters via SPI

### LINKED ISSUE #1997: Add a possibility to add custom PostDiscoveryFilter via SPI or configuration property
Currently we are working on solution that will allow users to select tests to run from web UI. Every test is already marked by custom `@TestId` annotation, so we only missing possibility to register a custom `PostDiscoveryFilter` via SPI or some kind of configuration property. 

Currently the only way to add `PostDiscoveryFilter` is to manually create `LauncherDiscoveryRequest` but that's seems to be hard for users (imagine, they will need to switch from maven/gradle integrations to custom plugin)

I know there is possibility to run tests via className/method name or uniqueId but this values may be changed during test refactoring but they will still related to same entity in test management system. Thats why we are using custom annotation `@TestId` to store this relation.

The solution I have in mind is to implement custom `PostDiscoveryFilter`, register it via SPI, and then enable/configure it via environment variable (system property).

baev:
## Overview

fixes #1997

As we discussed before, I introduced additional configuration property for `LauncherConfig` that allows users to control post-discovery filter auto registration and related methods. For consistency reasons, I introduced an additional method to manually add required post-discovery filters.

This part of changes available in separate commit https://github.com/junit-team/junit5/pull/2283/commits/cb24d1cf515c9ce6e41ea59cbe6f703918aa9c56

Then, additional filters are populated to `EngineDiscoveryOrchestrator`. I don't feel right about introduced changes in `Launcher` interface, It would be nice to hear your opinion on that point. The other option would be to modify the constructor of `DefaultLauncher` (or add a constructor that creates `DefaultLauncher`  from provided `EngineDiscoveryOrchestrator`. 

The same way I feel about `EngineDiscoveryOrchestrator` -- I decided to not change constructor and use fluent setter, but maybe it would be better to change existed `EngineDiscoveryOrchestrator` to accept post-discovery filters in addition to engines.

When we agree, I'll add more tests and, probably, we should add support for these changes to TestKit as well

---

---

### Definition of Done


marcphilipp:
> As we discussed before, I introduced additional configuration property for `LauncherConfig` that allows users to control post-discovery filter auto registration and related methods. For consistency reasons, I introduced an additional method to manually add required post-discovery filters.

👍 

> Then, additional filters are populated to `EngineDiscoveryOrchestrator`. I don't feel right about introduced changes in `Launcher` interface, It would be nice to hear your opinion on that point. The other option would be to modify the constructor of `DefaultLauncher` (or add a constructor that creates `DefaultLauncher` from provided `EngineDiscoveryOrchestrator`.

Let's go with modifying the `DefaultLauncher` constructor.

> The same way I feel about `EngineDiscoveryOrchestrator` -- I decided to not change constructor and use fluent setter, but maybe it would be better to change existed `EngineDiscoveryOrchestrator` to accept post-discovery filters in addition to engines.

Yes, I'd also pass them to the constructor.

> When we agree, I'll add more tests and, probably, we should add support for these changes to TestKit as well

What did you have in mind? I think it would be sufficient, if the TestKit would always pass an empty list of filters to the orchestrator. If users want to test their `PostDiscoveryFilters`, they can pass them in the request, or am I missing sth.?

baev:
> I think it would be sufficient, if the TestKit would always pass an empty list of filters to the orchestrator. If users want to test their `PostDiscoveryFilters`, they can pass them in the request, or am I missing sth.?

Yeah, you right

baev:
@marcphilipp everything is done

baev:
If you need I can squash commits into a single one

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryOrchestrator.java
```diff
@@ -82,7 +86,10 @@ public LauncherDiscoveryResult discover(LauncherDiscoveryRequest request, String
 			testEngineDescriptors.put(testEngine, rootDescriptor);
 		}
 
-		applyPostDiscoveryFilters(testEngineDescriptors, request.getPostDiscoveryFilters());
+		final List<PostDiscoveryFilter> filters = new LinkedList<>(postDiscoveryFilters);
```
marcphilipp, line 89: nit: we don't use `final` for local variables.
```suggestion
		List<PostDiscoveryFilter> filters = new LinkedList<>(postDiscoveryFilters);
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
```diff
@@ -93,7 +94,13 @@ public static Launcher create(LauncherConfig config) throws PreconditionViolatio
 		}
 		engines.addAll(config.getAdditionalTestEngines());
 
-		Launcher launcher = new DefaultLauncher(engines);
+		Set<PostDiscoveryFilter> filters = new LinkedHashSet<>();
```
marcphilipp, line 97: ❓ I'm not sure we should be using a `Set` here but then again it's consistent with `engines` above. Thoughts?

platform-tests/src/test/java/org/junit/platform/launcher/TestPostDiscoveryFilter.java
```diff
@@ -0,0 +1,23 @@
...
+package org.junit.platform.launcher;
+
+import org.junit.platform.engine.FilterResult;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.TestTag;
+
+public class TestPostDiscoveryFilter implements PostDiscoveryFilter {
+	@Override
+	public FilterResult apply(final TestDescriptor object) {
+		final boolean include = object.getTags().stream().map(TestTag::getName).anyMatch("test-post-discovery"::equals);
```
marcphilipp, line 20: Ditto.
```suggestion
		boolean include = object.getTags().stream().map(TestTag::getName).anyMatch("test-post-discovery"::equals);
```

platform-tests/src/test/java/org/junit/platform/launcher/TestPostDiscoveryFilter.java
```diff
@@ -0,0 +1,23 @@
...
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.launcher;
+
+import org.junit.platform.engine.FilterResult;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.TestTag;
+
+public class TestPostDiscoveryFilter implements PostDiscoveryFilter {
```
marcphilipp, line 17: How about `TestPostDiscoveryTagFilter`?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
baev: Agree. Totally makes no sense to use set, only for consistency reasons.

By having a list instead of set we can kinda keep the order of filters. At some point having possibility to control the orders can give users some advantages

baev: replaced with list for now

marcphilipp:
@baev Thanks, again! 👍

PR #2267, opened 2020-04-19
WIP: Class-based platform-suites from experiments/platform-suites based on 5.6.0

nicerloop:
## Overview

As I would have use of platform-level suites, especially to hold engine-specific configuration, I grabbed the code from experiments/platform-suites and ported it onto platform 1.6.0, adding recursive suite discovery and arbitrary engine configuration.

This is by no means a finished work, nor an original work, but it fulfils my immediate needs, and seems to be in line with some interests for 5.7 (see issue #744).

As this is my first attempt to contribute to this project, I am fully open to any guidance to improve this proposal to hopefully get it integrated.

I did some copy-paste of deprecated method I did not know how to replace, I did not check to coding conventions, nor did I update any documentation: this is a work in progress.

This Works for Me with cucumber-junit-platform-engine, pitest-junit5-plugin, serenity-cucumber5, from maven and eclipse, without visible problems.

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement).

---

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [ ] [Coding conventions](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [ ] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling
- [ ] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [ ] Change is documented in the [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [ ] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #2295, opened 2020-05-13
WIP: Class-based platform-suites from experiments/platform-suites over 5.7.0-M1

nicerloop:
## Overview

I ported experiments/platform-suites over 5.7.0-M1.

See #744
Replaces #2267 

I still need guidance to improve this proposal to hopefully get it integrated.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherDiscoveryResult.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherDiscoveryResult.java
@@ -14,8 +14,10 @@ import static java.util.Collections.unmodifiableMap;
 import static org.apiguardian.api.API.Status.INTERNAL;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Optional;
 
 import org.apiguardian.api.API;
 import org.junit.platform.engine.ConfigurationParameters;
@@ -33,11 +35,13 @@ public class LauncherDiscoveryResult {
 
 	private final Map<TestEngine, TestDescriptor> testEngineDescriptors;
 	private final ConfigurationParameters configurationParameters;
+	private final Optional<TestDescriptor> parentDescriptor;
 
 	LauncherDiscoveryResult(Map<TestEngine, TestDescriptor> testEngineDescriptors,
-			ConfigurationParameters configurationParameters) {
+			ConfigurationParameters configurationParameters, Optional<TestDescriptor> parentDescriptor) {
 		this.testEngineDescriptors = unmodifiableMap(new LinkedHashMap<>(testEngineDescriptors));
 		this.configurationParameters = configurationParameters;
+		this.parentDescriptor = parentDescriptor;
 	}
 
 	public TestDescriptor getEngineTestDescriptor(TestEngine testEngine) {
@@ -48,12 +52,17 @@ public class LauncherDiscoveryResult {
 		return configurationParameters;
 	}
 
+	Optional<TestDescriptor> getParentDescriptor() {
+		return parentDescriptor;
+	}
+
 	Collection<TestEngine> getTestEngines() {
 		return this.testEngineDescriptors.keySet();
 	}
 
 	Collection<TestDescriptor> getEngineTestDescriptors() {
-		return this.testEngineDescriptors.values();
+		return getParentDescriptor().map(o -> (Collection<TestDescriptor>) Collections.singleton(o)).orElseGet(
+			this.testEngineDescriptors::values);
 	}
 
 }
--- a/junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/Suite.java
+++ b/junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/Suite.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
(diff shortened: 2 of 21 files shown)


sbrannen:
Tentatively slated for 5.7 M2 for review

swaranga:
Is there support planned support for state sharing across test classes in the same suite. Something like a @BeforeSuite that initializes some expensive resources once which can then be shared across multiple test classes in the suite?

nicerloop:
Hi, for now my interest is in having multiple suites hosting different configurations of test engine (cucumber-junit-platform-engine in my case). I prototyped it by porting experiments/platform-suites over 5.6 and adding configuration parameters. I now hope to have it integrated, starting over on top of 5.7 M1. Will probably need much more than just polishing :)

swaranga:
Was a decision made suite support will be included in 5.7 M2?

marcphilipp:
> Was a decision made whether suite support will be included in 5.7 M2?

We don't think the spike implementation in this PR is production-ready. Adding another layer to the tree might break existing support in build tools and IDEs so it's not something we want to do without careful consideration.

mpkorstanje:
>  Adding another layer to the tree might break existing support in build tools and IDEs so it's not something we want to do without careful consideration.

Which tools would you take into consideration?

 * Maven
 * Gradle
 * IDEA
 * Eclipse
 * JUnit Platform Console Launcher

Any others?

marcphilipp:
I guess at least also VS Code, NetBeans, Bazel, and Ant.

mpkorstanje:
I've checked a few projects. Nothing breaking so far.

- [X] Maven
- [X] Gradle
  - The reporting on a class that is part of a suite collides with that same class executed by `junit-jupiter` in `build/gradle/reports`. Looks like the reporting is still using JUnit 4 assumptions about top level classes. Excluding `junit-jupiter` makes this go away.
- [X] IDEA 2020.2
  - Rerun fails for tests in suite
  - `@Testable` is missing on `@Suite` so no UI support.
- [ ] Eclipse
- [ ] JUnit Platform Console Launcher
- [ ] VS Code
- [ ] NetBeans
- [ ] Bazel
- [ ] Ant

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/Suite.java
```diff
@@ -0,0 +1,45 @@
...
+ * @see ExcludeTags
+ * @see IncludeEngines
+ * @see ExcludeEngines
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+@Documented
+@API(status = EXPERIMENTAL, since = "1.7")
+public @interface Suite {
```
mpkorstanje, line 44: Should also be annotated with `@Testable`

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherDiscoveryResult.java
```diff
@@ -33,11 +35,13 @@
 
 	private final Map<TestEngine, TestDescriptor> testEngineDescriptors;
 	private final ConfigurationParameters configurationParameters;
+	private final Optional<TestDescriptor> parentDescriptor;
```
mpkorstanje, line 38: Optional should not be used as a wrapper for fields. Consider pass in a null value in the constructor value and wrap it in the getter. As al alternative to passing an null value consider creating multiple constructors.

mpkorstanje (commented):
Looking at the code there are quite a few todo items that still need to be resolved.

However I think the biggest obstacle is the way tests suites are discovered. Currently the `junit-platform-launcher` sideloads the suite discovery request in `EngineDiscoveryOrchestrator.collectRequests`.

This comes with a two deficiencies:
 * `junit-platform-launcher` has a dependency on the `junit-platform-suite-api`
 * `suites` is not a `TestEngine` and `suites` can not be excluded/included from discovery like regular engines

Because `suite` is not a test engine I'm in the rather strange situation of having to exclude all other engines so as to avoid having the tests in the suite executed twice. Once by the discovery request from `suite` and once from the regular test engine. 

Now I think the JUnit Platform Launcher shouldn't have to know about the Suite API much like it doesn't know about any other test engine APIs. And looking at the implementation in I wonder if it would be possible to create a `junit-platform-suite-engine` module that either depends on the `junit-platform-launcher` or emulates its behaviour.

marcphilipp:
> Now I think the JUnit Platform Launcher shouldn't have to know about the Suite API much like it doesn't know about any other test engine APIs. And looking at the implementation in I wonder if it would be possible to create a `junit-platform-suite-engine` module that either depends on the `junit-platform-launcher` or emulates its behaviour.
> 
> This would at the very least remove the risk of introducing an extra layer in the tree. It may however require opening up the `junit-platform-launcher` or extracting some code from it.

I vaguely remember there was some reason we didn't consider making it a separate engine but I can't put my finger on it right now. I do think that would solve some of the concerns I'm having with this spike implementation.

@nicerloop @mpkorstanje Would this sth. you'd be interested in exploring? I cannot make any guarantees right now that we're going to accept it but it would be good to see how it would look.

mpkorstanje:
I'd at least be interested in exploring the suite test engine as a separate package within in the `junit-platform-launcher`.  I suspect that untangling it from the launcher will require a bit more direction. But we can consider that as needed.

nicerloop:
Hi, I would surely be interested in exploring the engine option, but as I have very little knowledge of JUnit code-base and project beyond what I discovered in this attempt, I would need some guidance to start one the right foot.

marcphilipp:
@mpkorstanje Sounds good!

Closing this PR in favor of exploring a (standalone) engine option.

mpkorstanje:
Cheers! I've implemented a prototype that seems to work for a very simple test case.

https://github.com/mpkorstanje/junit-platform-suite-engine

I'm returning further discussion to:

https://github.com/junit-team/junit5/issues/744

PR #2322, opened 2020-06-14
Replace "blacklisted" with "unrecoverable" exceptions

marcphilipp:
This commit replaces the oppressive term "blacklist" [1] with a more
descriptive, domain-specific one. Exceptions that are rethrown by the
new `UnrecoverableExceptions.rethrowIfUnrecoverable` method, are
_unrecoverable_, i.e. they should always terminate test plan execution
immediately instead of being caught and potentially wrapped in another
exception.

To support running newer engines on an older Platform, the existing
class is deprecated instead of deleted for now.

[1] https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.2

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/BlacklistedExceptions.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/BlacklistedExceptions.java
@@ -32,7 +32,9 @@ import org.apiguardian.api.API;
  * Use at your own risk!
  *
  * @since 1.0
+ * @deprecated Use {@link UnrecoverableExceptions} instead.
  */
+@Deprecated
 @API(status = DEPRECATED, since = "1.7")
 public final class BlacklistedExceptions {
 
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertAll.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertAll.java
@@ -18,8 +18,8 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.junit.jupiter.api.function.Executable;
-import org.junit.platform.commons.util.BlacklistedExceptions;
 import org.junit.platform.commons.util.Preconditions;
+import org.junit.platform.commons.util.UnrecoverableExceptions;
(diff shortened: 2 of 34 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/FatalExceptions.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package org.junit.platform.commons.util;
+
+import static org.apiguardian.api.API.Status.INTERNAL;
+
+import org.apiguardian.api.API;
+
+/**
+ * Internal utilities for working with <em>fatal</em> exceptions.
```
sbrannen, line 18: Instead of the term "fatal", I would suggest something along the lines of "unrecoverable" meaning an exception from which the application (test plan) cannot recover.

It's certainly debatable due to the common use of "fatal errors" in computing in general, but I have been told that terms (potentially) implying "death" or "violence" should be avoided if other terms are more suitable.

sbrannen: Granted, I realize that "fatal" can also mean ["causing failure"](https://www.merriam-webster.com/dictionary/fatal). That's why I say it's debatable.

marcphilipp: I agree that "unrecoverable" is better and will change it accordingly.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/BlacklistedExceptions.java
```diff
@@ -36,26 +33,26 @@
  *
  * @since 1.0
  */
-@API(status = INTERNAL, since = "1.0")
+@API(status = DEPRECATED, since = "1.7")
 public final class BlacklistedExceptions {
```
sbrannen, line 37: Please add an `@deprecated` Javadoc tag at the class level analogous to the one at the method level.

marcphilipp: Added.

PR #2323, opened 2020-06-14
Move Assertions.fail return type documentation to fail(String)

### LINKED ISSUE #2321: Move Assertions.fail example to Assertions.fail(String)
Currently the explanation for the return value of `Assertions.fail` and an example for it exists for `Assertions.fail(String, Throwable)`:
https://github.com/junit-team/junit5/blob/be2aa2497a75dedceb37bc3297276e230fb4f293/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java#L117-L128

However, the example does not fit because it does not use the `fail(String, Throwable)` overload. Would it therefore make sense to move the return value explanation and the example to `fail(String)` and link to that from the other `fail` overloads instead?

Marcono1234:
## Overview
Fixes #2321 by moving the `Assertions.fail(...)` return type documentation and example to `Assertions.fail(String)` since the example did not fit for `fail(String, Throwable)` where it was previously.

---

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
@@ -91,8 +91,8 @@ public class Assertions {
 	 * <p>Although failing <em>with</em> an explicit failure message is recommended,
 	 * this method may be useful when maintaining legacy code.
 	 *
-	 * <p>See Javadoc for {@link #fail(String, Throwable)} for an explanation of
-	 * this method's generic return type {@code V}.
+	 * <p>See Javadoc for {@link #fail(String)} for an explanation of this method's
+	 * generic return type {@code V}.
 	 */
 	public static <V> V fail() {
 		AssertionUtils.fail();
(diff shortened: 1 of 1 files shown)


sbrannen:
review _in progress_

sbrannen:

Thanks

Marcono1234:
Sorry in case the commit message issue reference was not in the expected format, [CONTRIBUTING.md](https://github.com/junit-team/junit5/blob/master/CONTRIBUTING.md#commit-messages) suggests writing "Issue: #XYZ", though this is not recognized by GitHub apparently.

Is "Closes #XYZ" the expected commit message format? And do you want me to write a pull request changing this in CONTRIBUTING.md?

sbrannen:
Regarding "Issue:" vs "Closes", we do generally like to use "Issue:" since it shows a relationship to the issue and does not automatically close it.

When we are certain that we want the issue/PR to be automatically closed with the commit, we use "Closes".

Historically we only used "Issue", but we have started using "Closes" more and more often.

So perhaps we should update the contributing guidelines to point out when to use "Issue" and when to use "Closes".

PR #2346, opened 2020-06-29
Introduce `TypedArgumentConverter` as a more specific `ArgumentConverter`

### LINKED ISSUE #2284: Provide a less generic argument converter
When implementing an explicit converter via `SimpleArgumentConverter`, the signature of the `convert` method is the following: `Object convert(Object source, Class<?> targetType)`. Some boilerplate code is then needed to check if the types of the source/target are supported.

I personally often use one converter per use case, i.e. to convert type X to type Y.
Maybe we could implement a new argument converter that would take two types (one for the source and one for the target) and would do the checks for the user. The method to implement would then have a simpler signature.

## Deliverables

- [ ] An abstract converter like `TypedArgumentConverter<S, T>` that would ensure that source is of type S and target of type T (Simple's already taken so... `SimplerArgumentConverter`? :P)

juliette-derancourt:
Resolves #2284.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
@@ -105,6 +105,8 @@ on GitHub.
   same functionality but a more descriptive name.
 * `assertTimeoutPreemptively` in `Assertions` now reports the stacktrace of the timed out
   thread in the cause of the `AssertionFailedError`.
+* New `TypedArgumentConverter` that always convert one specific type to another, therefore
+  reducing boilerplate type checks compared to `ArgumentConverter`.
 
 [[release-notes-5.7.0-M2-junit-vintage]]
 === JUnit Vintage
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
@@ -105,8 +105,8 @@ on GitHub.
   same functionality but a more descriptive name.
 * `assertTimeoutPreemptively` in `Assertions` now reports the stacktrace of the timed out
   thread in the cause of the `AssertionFailedError`.
-* New `TypedArgumentConverter` that always convert one specific type to another, therefore
-  reducing boilerplate type checks compared to `ArgumentConverter`.
+* New `TypedArgumentConverter` for converting one specific type to another, therefore
+  reducing boilerplate type checks compared to implementing `ArgumentConverter` directly.
 
 [[release-notes-5.7.0-M2-junit-vintage]]
 === JUnit Vintage
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1442,6 +1442,14 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=explicit_conversion_e
 include::{testDir}/example/ParameterizedTestDemo.java[tags=explicit_conversion_example_ToStringArgumentConverter]
 ----
 
+If the converter is only meant to convert one type to another, you can extend
+`TypedArgumentConverter` that reduce boilerplate type checks.
+
+[source,java,indent=0]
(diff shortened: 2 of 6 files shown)


juliette-derancourt:
This one has been sitting on my computer for quite some time, and I remembered why: I'm not sure how to test this other than with an example like I did? @junit-team/junit-lambda Do you think it enough?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/TypedArgumentConverter.java
```diff
@@ -0,0 +1,61 @@
...
+	@Override
+	@SuppressWarnings("unchecked")
+	public final Object convert(Object source, ParameterContext context) throws ArgumentConversionException {
+		if (!sourceType.isAssignableFrom(source.getClass())) {
+			throw new ArgumentConversionException("Can only convert objects of type " + sourceType);
+		}
+		if (!targetType.isAssignableFrom(context.getParameter().getType())) {
+			throw new ArgumentConversionException("Can only convert to type " + targetType);
+		}
+		return convert((S) source);
```
marcphilipp, line 47: The `@SuppressWarnings` can be avoided via:

```suggestion
		return convert(sourceType.cast(source));
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/TypedArgumentConverter.java
```diff
@@ -0,0 +1,61 @@
...
+ * @since 5.7
+ * @see ArgumentConverter
+ */
+@API(status = EXPERIMENTAL, since = "5.7")
+public abstract class TypedArgumentConverter<S, T> implements ArgumentConverter {
+
+	private final Class<S> sourceType;
+	private final Class<T> targetType;
+
+	protected TypedArgumentConverter(Class<S> sourceType, Class<T> targetType) {
```
marcphilipp, line 33: Please take a look at `TypeBasedParameterResolver` which determines the parameter type reflectively which would avoid the boilerplate constructor.

juliette-derancourt: Thing is, I would have to cast the found `Class` objects into `Class<S>` and `Class<T>`? Or am I missing something?

marcphilipp: I guess so... 🤔 @sbrannen WDYT?

marcphilipp:
@juliette-derancourt We discussed this in today's call and think we should go ahead as is. Could you please add an example to the user guide and an entry to the release notes?

documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
```diff
@@ -105,6 +105,8 @@ on GitHub.
   same functionality but a more descriptive name.
 * `assertTimeoutPreemptively` in `Assertions` now reports the stacktrace of the timed out
   thread in the cause of the `AssertionFailedError`.
+* New `TypedArgumentConverter` that always convert one specific type to another, therefore
+  reducing boilerplate type checks compared to `ArgumentConverter`.
```
marcphilipp, line 109:
```suggestion
* New `TypedArgumentConverter` for converting one specific type to another, therefore
  reducing boilerplate type checks compared to implementing `ArgumentConverter` directly.
```

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1442,6 +1442,14 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=explicit_conversion_e
 include::{testDir}/example/ParameterizedTestDemo.java[tags=explicit_conversion_example_ToStringArgumentConverter]
 ----
 
+If the converter is only meant to convert one type to another, you can extend
+`TypedArgumentConverter` that reduce boilerplate type checks.
```
marcphilipp, line 1446:
```suggestion
`TypedArgumentConverter` to avoid boilerplate type checks.
```

PR #2350, opened 2020-07-05
Cache Gradle wrapper and dependencies on CI builds

oehme:
## Overview

This should speed up CI builds as Gradle distributions and external dependencies no longer have to be downloaded each time. Since there are also some Maven integration tests, the local Maven repository is also cached.

---

---

### Definition of Done

### DIFF
--- a/.github/workflows/cross-version.yml
+++ b/.github/workflows/cross-version.yml
@@ -23,6 +23,22 @@ jobs:
     container: "junitteam/build:${{ matrix.jdk }}"
     steps:
       - uses: actions/checkout@v2
+      - name: Cache Gradle wrapper and dependencies
+        uses: actions/cache@v2
+        with:
+          path: |
+            /root/.gradle/caches/
+            /root/.gradle/wrapper/dists
+          key: test-${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle', '**/*.gradle.kts', '**/gradle.properties', 'gradle/**', 'buildSrc/src/main/**') }}
+          restore-keys: |
+            test-${{ runner.os }}-gradle-
+      - name: Cache local Maven repository
+        uses: actions/cache@v2
+        with:
+          path: /root/.m2/repository
+          key: test-${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            test-${{ runner.os }}-maven-
       - name: Prepare Gradle Enterprise credentials
         run: |
           mkdir -p /root/.gradle/enterprise/
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,22 @@ jobs:
     container: junitteam/build:latest
     steps:
     - uses: actions/checkout@v2
+    - name: Cache Gradle wrapper and dependencies
+      uses: actions/cache@v2
+      with:
+        path: | 
+          /root/.gradle/caches/
+          /root/.gradle/wrapper/dists
+        key: test-${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle', '**/*.gradle.kts', '**/gradle.properties', 'gradle/**', 'buildSrc/src/main/**') }}
+        restore-keys: |
+          test-${{ runner.os }}-gradle-
+    - name: Cache local Maven repository
+      uses: actions/cache@v2
+      with:
+        path: /root/.m2/repository
+        key: test-${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          test-${{ runner.os }}-maven-
     - name: Prepare Gradle Enterprise credentials
       run: |
         mkdir -p /root/.gradle/enterprise/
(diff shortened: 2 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -101,6 +165,15 @@ jobs:
     if: github.event_name == 'push' && github.repository == 'junit-team/junit5' && (startsWith(github.ref, 'refs/heads/releases/') || github.ref == 'refs/heads/main')
     steps:
     - uses: actions/checkout@v2
+    - name: Cache Gradle wrapper and dependencies
```
oehme, line 168: This step and the one below don't invoke the Maven integration tests, so no need to cache .m2 here.

.github/workflows/cross-version.yml
```diff
@@ -23,6 +23,22 @@ jobs:
     container: "junitteam/build:${{ matrix.jdk }}"
     steps:
       - uses: actions/checkout@v2
+      - name: Cache Gradle wrapper and dependencies
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches/
+            ~/.gradle/wrapper/dists
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
```
marcphilipp, line 32: Shouldn't this also include `buildSrc/**`, `gradle/**`, and `gradle.properties`?

oehme: Good point, I've expanded the list.

oehme:
Something's not working corretly here. On Linux, the cache entry is 0MB, on Windows we get no hits at all and on MacOS it works fine.

oehme:
Ah I think I know why - All builds are using the same key, but some of them aren't actually invoking any Gradle or Maven steps. So we have to give each its own ID.

oehme:
Nope, still seeing the same issue. On Windows creating the cache entry fails because of the `.lock` files. On Linux it only works correctly for the "reproducible build" workflow, but not the others. The difference seems to be that the "reproducible build" workflow doesn't use a container, but the others do.

sormuras:
_Who put what into your ☕ , Stefan? On steroids?_ 🚀  ;-)

Watching every commit since yesterday and hoping to see a significant gain of build speed on all platforms. As long as the builds stay "correct" and the configuration of the caching setup isn't too complex, I'm really forward to the (timing) results of this PR.

oehme:
Finally figured out the issue on Linux containers - `~/.gradle`  doesn't work, you have to use `/root/.gradle` instead. The Windows issue was easily fixed by stopping the daemon (releasing its file locks).

You can now compare the performance of a typical CI build on the [main branch](https://github.com/junit-team/junit5/actions/runs/157493860) to [this cached one](https://github.com/junit-team/junit5/actions/runs/159399712) on my branch. Each individual build now takes ~3min instead of ~5min.

If you did the coverage report in parallel to the other tests, the overall build time would come down further.

oehme:
I've optimized this a bit further by only including the essentials in the cache jar, which reduces upload/download times. Saves another ~10s on every build.

oehme:
I tried reducing the size of the cache archives by excluding some files, but that doesn't seem to work. I've opened an [issue](https://github.com/actions/cache/issues/364) for the cache action. If that gets resolved, we could save some more time.

sormuras:
Great work, @oehme!

>  CI  / Mac OS (pull_request) Successful in 14m

What-ever happened here. **14m** is not ... quick.

> If you did the coverage report in parallel to the other tests, the overall build time would come down further.

True. 

> I tried reducing the size of the cache archives by excluding some files, [...]

+1'ed the issue you created for actions/cache ... but what about deleting all `gradle*.zip` files manually before caching?

--

I wonder whether https://github.com/junit-team/junit5/blob/main/src/publishDocumentationSnapshotOnlyIfNecessary.sh should also be integrated directly into the GH workflow configuration file. I created it back then in a dedicated script file to be able to invoke and test it locally as well as on different CI service providers.

oehme:
Yeah I saw those outliers too, but not just on my branch. Hard to tell what went wrong there without some timestamps. Maybe some test using an outside resource that sometimes responds slowly?

We could delete the zip files, but I didn't want to complicate the caching further, it's already a lot of copy-pasta.

vlsi:
> Cache Size: ~633 MB (664085360 B)
> /bin/tar -z -xf /__w/_temp/9da0e536-5cf5-4944-8581-e15877787035/cache.tgz -P -C /__w/junit5/junit5
> Cache restored from key: test-Linux-gradle-2bf12571c9ec30ce7d57c6b569df17c660c961e659cee88c27370df7fc661671

AFAIK there's 5GiB limit: https://github.com/actions/cache#cache-limits, so it can be challenging to fit into the limit.

vlsi:
@oehme , do you think it is faster to download Gradle from the cache rather than from services.gradle.org?
I guess the amount of work should be more-or-less the same: download archive and unpack it.

I see why caching dependencies helps: it reduces the number of Nexus/Artifactory requests, however, gradle-..zip should not gain much from caching. Am I missing something?

oehme:
The main idea is that the caches are co-located, probably in the same datacenter, rather than possibly halfway across the globe.

marcphilipp:
Let's give this a try for a while and see how it evolves over time. Thanks again, @oehme! 👍

vlsi:
I wonder what would be the timings if .gradle/wrapper dir is not cached.

My measurements for pgjdbc sugget that caching wrapper does not improve
build time.

marcphilipp:
But it doesn't make it worse, either, does it?

vlsi:
It makes it worse: when the cache is changed (e.g. dependency is updated),
it increases the cache size, so there's more to be compressed and uploaded

>

marcphilipp:
Thanks for the heads-up. I'll keep an eye on it.

sormuras:
800 megs o' cache

image

PR #2365, opened 2020-07-29
Use specific name prefix in preemptive timeout threads

zabetak:
## Overview

Having a specific name prefix is useful when debugging stack traces to
quickly identify which threads are created by the junit framework and
from where exactly.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
@@ -193,12 +193,7 @@ class AssertTimeout {
 		private static final AtomicInteger threadNumber = new AtomicInteger(1);
 
 		public Thread newThread(Runnable r) {
-			Thread t = new Thread(null, r, "junit-timeout-thread-" + threadNumber.getAndIncrement(), 0);
-			if (t.isDaemon())
-				t.setDaemon(false);
-			if (t.getPriority() != Thread.NORM_PRIORITY)
-				t.setPriority(Thread.NORM_PRIORITY);
-			return t;
+			return new Thread(null, r, "junit-timeout-thread-" + threadNumber.getAndIncrement(), 0);
 		}
 	}
 
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
@@ -19,8 +19,10 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
 
@@ -127,7 +129,7 @@ class AssertTimeout {
 			Object messageOrSupplier) {
 
 		AtomicReference<Thread> threadReference = new AtomicReference<>();
-		ExecutorService executorService = Executors.newSingleThreadExecutor();
+		ExecutorService executorService = Executors.newSingleThreadExecutor(new TimeoutThreadFactory());
 
 		try {
 			Future<T> future = executorService.submit(() -> {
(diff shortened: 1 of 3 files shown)


sbrannen:
Thanks for the PR.

I think using a custom thread name prefix is a good idea, and I wish I'd thought of that when I first introduced the preemptive timeout support.

Let's see what the rest of the team thinks.

----

Tentatively slated for 5.7 M2 for _team discussion_.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
```diff
@@ -180,4 +182,30 @@ static void assertTimeoutPreemptively(Duration timeout, Executable executable, S
...
+	 * helpful for debugging purposes.
+	 *
+	 */
+	private static class TimeoutThreadFactory implements ThreadFactory {
+		private static final AtomicInteger threadNumber = new AtomicInteger(1);
+		private final ThreadGroup group;
+
+		TimeoutThreadFactory() {
+			SecurityManager s = System.getSecurityManager();
+			group = (s != null) ? s.getThreadGroup() : Thread.currentThread().getThreadGroup();
```
marcphilipp, line 198: There's no need to do this, `Thread` already does this if no group is specified explicitly.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
```diff
@@ -180,4 +182,30 @@ static void assertTimeoutPreemptively(Duration timeout, Executable executable, S
...
+		}
+
+		public Thread newThread(Runnable r) {
+			Thread t = new Thread(group, r, "junit-timeout-thread-" + threadNumber.getAndIncrement(), 0);
+			if (t.isDaemon())
+				t.setDaemon(false);
+			if (t.getPriority() != Thread.NORM_PRIORITY)
+				t.setPriority(Thread.NORM_PRIORITY);
```
marcphilipp, line 206: I don't think we need these two `if`s as we shut down the executor afterwards.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
zabetak: I think it makes sense if the factory is about to create many threads but given that at the moment we have a factory per thread it can be removed.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
zabetak: I don't think it matters if we shutdown the executor or not. Even after shutdown the thread is not guaranteed to terminate so being a daemon or not matters. 

Moreover, in a setting where there many threads competing for CPU if these ones have low/high priority it might have an impact.

zabetak:
Thanks for the feedback. I addressed the comments and pushed a few more commits. Let me know if and when I need to squash the commits.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
```diff
@@ -180,4 +182,17 @@ static void assertTimeoutPreemptively(Duration timeout, Executable executable, S
...
+	/**
+	 * The thread factory used for preemptive timeout.
+	 *
+	 * The factory creates threads with meaningful names, helpful for debugging purposes.
+	 */
+	private static class TimeoutThreadFactory implements ThreadFactory {
+		private static final AtomicInteger threadNumber = new AtomicInteger(1);
+
+		public Thread newThread(Runnable r) {
+			return new Thread(null, r, "junit-timeout-thread-" + threadNumber.getAndIncrement(), 0);
```
marcphilipp, line 194:
```suggestion
			return new Thread(r, "junit-timeout-thread-" + threadNumber.getAndIncrement());
```

marcphilipp:
> Let me know if and when I need to squash the commits.

No need, we'll squash them when we merge the PR.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
marcphilipp: > Even after shutdown the thread is not guaranteed to terminate so being a daemon or not matters.

D you mean we should make it a daemon thread to avoid it from preventing the VM to shut down?

zabetak: Basically, I am saying that we should decide if we want daemon or non daemon threads. If we don't specify it then we rely on the status of the parent thread. In this case, I could prefer non-daemon threads.

marcphilipp:
@zabetak Could you please add an entry to the release notes in `documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc`?

marcphilipp:
Thanks, @zabetak! 👍

PR #2368, opened 2020-08-01
Emit entire non-matching lines

### LINKED ISSUE #2233: Describe mismatches found by assertLinesMatch
Currently, in `AssertLinesMatch` on `main`, I see:

```java
	private final static int MAX_SNIPPET_LENGTH = 21;
	private final static int MAX_LINES_IN_FAILURE_MESSAGE = 42;
```

Please create some way to allow the user of JUnit 5 to configure these parameters, with a value of zero meaning "unlimited".  The usefulness of `assertLinesMatch` degrades when the snippets are too short to show the problem that needs to debugged, and when the relevant lines are omitted.

The numbers 21 and 42 are certainly not "one size fits all" parameters.

Recent releases of IntellJ IDEA have a really nice feature parses the output of `assertLinesMatch` fails and shows the results in a graphical diff window.  The omitted lines make that feature far less valuable than it could be as well.

I think a static knob would be fine. I see no need for these to be changed dynamically during test suite execution.

How about just a simple system property for each of those two parameters?

Thanks!

sormuras:
## Overview

Prior to this PR, a small number (`42`) of lines were always included in the error message string of `assertLinesMatch`. This arbitrary number of lines did not always contain the lines that did not match. This PR removes the constant number and always emits the two non-matching lines entirely.

The staged matching logic is still working in a "fail-fast mode". The first non-mathing line pair is reported. Subsequent lines are not processed.

Closes #2233

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
@@ -80,126 +80,135 @@ class AssertLinesMatch {
 			return;
 		}
 
-		int expectedSize = expectedLines.size();
-		int actualSize = actualLines.size();
-
-		// trivial case: when expecting more than actual lines available, something is wrong
-		if (expectedSize > actualSize) {
-			fail(expectedLines, actualLines, messageOrSupplier, "expected %d lines, but only got %d", expectedSize,
-				actualSize);
-		}
-
-		// simple case: both list are equally sized, compare them line-by-line
-		if (expectedSize == actualSize) {
-			if (IntStream.range(0, expectedSize).allMatch(i -> matches(expectedLines.get(i), actualLines.get(i)))) {
-				return;
-			}
-			// else fall-through to "with fast-forward" matching
-		}
-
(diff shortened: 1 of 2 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
```diff
@@ -173,7 +171,9 @@ void assertLinesMatchWithFastForward() {
 					}
 				}
 
-				fail("expected line #%d:`%s` doesn't match", expectedLineNumber, snippet(expectedLine));
+				int actualLineNumber = actualLines.size() - actualDeque.size() + 1; // 1-based line number
+				fail("expected line #%d doesn't match actual line #%d%n" + "\texpected: `%s`%n" + "\t  actual: `%s`",
```
marcphilipp, line 175:
```suggestion
				fail("expected line #%d doesn't match actual line #%d%n\texpected: `%s`%n\t  actual: `%s`",
```

sormuras: Ah... my fault. I wanted to start each line in an extra line. Will add `//` to force line breaks.

PR #2371, opened 2020-08-02
Truncate arguments in display name if they exceed max length

### LINKED ISSUE #2358: Excessively long argument values in parameterized display names should be truncated
JUnit provides a nice way to configure test names, however, the names might become too long unexpectedly.

Here's a case for JUnit4:

https://github.com/pgjdbc/pgjdbc/blob/aa190cce2c27ac5f5d4fa33f5d156f8bf04cbe0c/pgjdbc/src/test/java/org/postgresql/core/UTF8EncodingTest.java#L29-L33

```java
  @Parameterized.Parameters(name = "string={1}, encoding={0}")
  public static Iterable<Object[]> data() {
    final StringBuilder reallyLongString = new StringBuilder(1024 * 1024);
    for (int i = 0; i < 185000; ++i) {
      reallyLongString.append(i);
    }
```

I guess the same is relevant for JUnit5 as well.

It would be great if JUnit could detect such cases, and it could trim the sub-parts before it produces a single multi-megabyte test name.

For instance, in the case like above JUnit could produce something like `string=012345678910111213...truncated...sha1=12312312123, encoding=utf-8`

It turns out the long test names consume disk space as Gradle records all the test outputs, and it does make `remote build cache` slower, as the build cache has to keep all the long test names (so it impacts `cache space consumption`, `cache store latency` and `cache retrieval latency`).

Of course, users can truncate test names on their own, however, I believe it would be better if JUnit prevented "unexpectedly long test name" issue automatically.

In other words:

1) If use is using placeholder-like pattern, then JUnit should trim all the parameters.
2) If the user returns test name via programmatic API (I'm not sure if that exists at all, but I assume it does :) ), then JUnit could truncate the full string.
3) There should probably be a configuration option for the default limit of the parameter value.

WDYT?

Relevant Gradle Build Cache issue: https://github.com/gradle/gradle/issues/13608#issuecomment-660874776

marcphilipp:
Resolves #2358.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
@@ -109,7 +109,7 @@ on GitHub.
   reducing boilerplate type checks compared to implementing `ArgumentConverter` directly.
 * Added `ExtensionContext.getConfigurationParameter(String, Function<String, T>)`
   convenience method for reading transformed configuration parameters from extensions.
-* Arguments of in display names of parameterized test invocations are now truncated if
+* Arguments in display names of parameterized test invocations are now truncated if
   they exceed a configurable maximum length (defaults to 512 characters).
 
 [[release-notes-5.7.0-M2-junit-vintage]]
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
@@ -109,6 +109,8 @@ on GitHub.
   reducing boilerplate type checks compared to implementing `ArgumentConverter` directly.
 * Added `ExtensionContext.getConfigurationParameter(String, Function<String, T>)`
   convenience method for reading transformed configuration parameters from extensions.
+* Arguments of in display names of parameterized test invocations are now truncated if
+  they exceed a configurable maximum length (defaults to 512 characters).
 
 [[release-notes-5.7.0-M2-junit-vintage]]
 === JUnit Vintage
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
@@ -107,6 +107,8 @@ on GitHub.
   thread in the cause of the `AssertionFailedError`.
 * New `TypedArgumentConverter` for converting one specific type to another, therefore
   reducing boilerplate type checks compared to implementing `ArgumentConverter` directly.
+* Added `ExtensionContext.getConfigurationParameter(String, Function<String, T>)`
+  convenience method for reading transformed configuration parameters from extensions.
 
 [[release-notes-5.7.0-M2-junit-vintage]]
 === JUnit Vintage
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1563,8 +1563,8 @@ The following placeholders are supported within custom display names.
 | `{0}`, `{1}`, ...        | an individual argument
 |===
 
-NOTE: When including arguments in display names, their string representation is truncated
-if it exceeds the configured maximum length. The limit is configurable via the
+NOTE: When including arguments in display names, their string representations are truncated
+if they exceed the configured maximum length. The limit is configurable via the
 `junit.jupiter.params.displayname.argument.maxlength` configuration parameter and defaults
 to 512 characters.
 
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1563,6 +1563,11 @@ The following placeholders are supported within custom display names.
 | `{0}`, `{1}`, ...        | an individual argument
 |===
 
+NOTE: When including arguments in display names, their string representation is truncated
+if it exceeds the configured maximum length. The limit is configurable via the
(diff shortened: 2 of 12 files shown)


junit-jupiter-params/src/test/java/org/junit/jupiter/params/ParameterizedTestIntegrationTests.java
```diff
@@ -219,6 +220,17 @@ void executesLifecycleMethods() {
...
 	}
 
+	@Test
+	void truncatesArgumentsThatExceedMaxLength() {
+		var results = EngineTestKit.engine(new JupiterTestEngine()) //
+				.configurationParameter(ParameterizedTestExtension.ARGUMENT_MAX_LENGTH_KEY, "2") //
+				.selectors(selectMethod(TestCase.class, "testWithCsvSource", String.class.getName())) //
+				.execute();
+		results.testEvents().assertThatEvents() //
+				.haveExactly(1, event(displayName("[1] argument=f…"), started())) //
```
juliette-derancourt, line 230: So do we consider that `…` counts as a character? Personally, with a max length of _two_ I'd naturally expect `fo…`. But I may be the only one 🤷‍♀️

marcphilipp: TBH I was myself torn on this question. @junit-team/junit-lambda Thoughts?

sormuras: If I specify a max char count, I expect to get back a string of the given max length. For max = 5 and `...` it'd yield:

```
   "123" = "123"
  "1234" = "1234"
 "12345" = "12345"
"123456" = "12..."
```

marcphilipp: I'm using the Unicode ellipsis character `…` which is just one character so it would be `1234…` for `123456` in your example.

sormuras: Ah, I see. Nice trick to retain two characters from the source string. I'll update the example above accordingly.

Gist: a given max char count should not be exceeded. It's max. Not max+n. ;-)

sbrannen: > Gist: a given max char count should not be exceeded. It's max. Not max+n. ;-)

sbrannen: >Gist: a given max char count should not be exceeded. It's max. Not max+n.

I agree with that.

documentation/src/docs/asciidoc/release-notes/release-notes-5.7.0-M2.adoc
```diff
@@ -107,6 +107,10 @@ on GitHub.
   thread in the cause of the `AssertionFailedError`.
 * New `TypedArgumentConverter` for converting one specific type to another, therefore
   reducing boilerplate type checks compared to implementing `ArgumentConverter` directly.
+* Added `ExtensionContext.getConfigurationParameter(String, Function<String, T>)`
+  convenience method for reading transformed configuration parameters from extensions.
+* Arguments of in display names of parameterized test invocations are now truncated if
```
sbrannen, line 112:
```suggestion
* Arguments in display names of parameterized test invocations are now truncated if
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
```diff
@@ -91,10 +94,17 @@ private String argumentsPattern(Object[] arguments) {
...
-				result[i] = StringUtils.nullSafeToString(arguments[i]);
+				result[i] = truncateIfExceedsMaxLength(StringUtils.nullSafeToString(arguments[i]));
 			}
 		}
 		return result;
 	}
 
+	private String truncateIfExceedsMaxLength(String argument) {
+		if (argument.length() > argumentMaxLength) {
+			return argument.substring(0, argumentMaxLength - 1) + "\u2026";
```
sbrannen, line 105: Maybe add a comment that the unicode character is an ellipsis.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1563,6 +1563,11 @@ The following placeholders are supported within custom display names.
 | `{0}`, `{1}`, ...        | an individual argument
 |===
 
+NOTE: When including arguments in display names, their string representation is truncated
+if it exceeds the configured maximum length. The limit is configurable via the
```
sbrannen, line 1567:
```suggestion
NOTE: When including arguments in display names, their string representations are truncated
if they exceed the configured maximum length. The limit is configurable via the
```

junit-jupiter-params/src/test/java/org/junit/jupiter/params/ParameterizedTestIntegrationTests.java
juliette-derancourt: So be it 😄 
But we should make that clear in the documentation then.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
marcphilipp: Done in 4f4707e082ae70e3f34925f0d1f37039b5d70691.

marcphilipp:
@sbrannen PTAL

PR #2382, opened 2020-08-13
Introduce global default lock to facilitate isolated test execution

### LINKED ISSUE #2142: Allow to run some tests isolated
We have one test that changes the global state, for example
```java
@Test
void localeTest() {
  Locale.setDefault(Locale.FRENCH);
  assertEquals("Bunjour", getMessage());
}
```
and hundreds of tests that reads the state:
```java
@Test
void someTest() {
  assertEquals("Hello", getMessage());
}
```

After enabling parallel test execution our tests start to fail at random locations with messages like that:
`[ERROR]   message 0 expected:<...Annotation.java:17: [Annotation 'AnnotationAnnotation' have incorrect indentation level 2, expected level should be 0].> but was:<...Annotation.java:17: [Die Annotation 'AnnotationAnnotation' hat eine unerwartete Einrückungstiefe von 2 (erwartet: 0)].>`

`@Execution(ExecutionMode.SAME_THREAD)` for the test `localeTest` won't help us, since the all other tests are able to run concurrent with every other test and with `localeTest` too.

This can be solved with `@ResourceLock`:
```java
@Test
@Execution(ExecutionMode.SAME_THREAD)
@ResourceLock(value="global", mode=READ_WRITE)
void localeTest() {
  Locale.setDefault(Locale.FRENCH);
  assertEquals("Bunjour", getMessage());
}

@Test
@Execution(ExecutionMode.CONCURRENT)
@ResourceLock(value="global", mode=READ)
void test1() {
}

// ...

@Test
@Execution(ExecutionMode.CONCURRENT)
@ResourceLock(value="global", mode=READ)
void test100500() {
}
```

but this is error prone, since adding a new Test without this annotation will produce a flaky tests which may fail.

It will be much better to mark some tests as "incompatible with parallel execution at all":
 ```java
@Test
@Execution(ExecutionMode.ALONE) // no other this will be executed in parallel 
void localeTest() {
  Locale.setDefault(Locale.FRENCH);
  assertEquals("Bunjour", getMessage());
}
```

As a workaround, it is possible to turn such a test into Junit 4 test, which will be executed in other universe, isolated from all other.

### PR DESCRIPTION
**Pull request by marcphilipp**
This commit introduces a new global resource lock "__global__" that all
test descriptors that are children of the engine descriptor acquire by
default in READ mode. Using the `@Isolated` annotation in the Jupiter
API causes the mode to be changed to READ_WRITE.

Will eventually resolve #2142.

---

---

### Definition of Done

  - [x] Add tests for `@Isolated` on test classes and in `@Nested` tests

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api.parallel;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+
+@API(status = EXPERIMENTAL, since = "5.7")
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Inherited
+@ResourceLock("__global__")
+public @interface Isolated {
+}
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
@@ -20,6 +20,19 @@ import java.lang.annotation.Target;
 
 import org.apiguardian.api.API;
 
+/**
+ * {@code @Isolated} is used to declare that the annotated test class should be
+ * executed in isolation from other test classes.
+ *
+ * <p>When a test class is run in isolation, no other test class is executed
+ * concurrently. This can be used to enable parallel test execution for the
+ * entire test suite while running some tests in isolation (e.g. if they modify
+ * some global resource).
+ *
+ * @since 5.7
+ * @see ExecutionMode
+ * @see ResourceLock
+ */
 @API(status = EXPERIMENTAL, since = "5.7")
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
@@ -22,7 +22,7 @@ import org.apiguardian.api.API;
 
 @API(status = EXPERIMENTAL, since = "5.7")
 @Retention(RetentionPolicy.RUNTIME)
-@Target({ ElementType.TYPE, ElementType.METHOD })
+@Target(ElementType.TYPE)
 @Inherited
 @ResourceLock("__global__")
 public @interface Isolated {
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
@@ -37,6 +37,6 @@ import org.apiguardian.api.API;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
 @Inherited
-@ResourceLock("__global__")
+@ResourceLock("org.junit.platform.engine.support.hierarchical.ExclusiveResource.GLOBAL_KEY")
 public @interface Isolated {
 }
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ExclusiveResource.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ExclusiveResource.java
@@ -29,6 +29,8 @@ import org.junit.platform.commons.util.ToStringBuilder;
 @API(status = EXPERIMENTAL, since = "1.3")
 public class ExclusiveResource {
 
+	static final String GLOBAL_RESOURCE_LOCK_KEY = "__global__";
+
 	private final String key;
 	private final LockMode lockMode;
 	private int hash;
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTreeWalker.java [hunk 1]
```diff
@@ -10,39 +10,62 @@
...
+			walk(child, child, advisor);
+		});
 		return advisor;
 	}
 
-	private void walk(TestDescriptor testDescriptor, NodeExecutionAdvisor advisor) {
+	private void walk(TestDescriptor globalLockDescriptor, TestDescriptor testDescriptor,
+			NodeExecutionAdvisor advisor) {
 		Set<ExclusiveResource> exclusiveResources = getExclusiveResources(testDescriptor);
+		Set<ExclusiveResource> allResources = new HashSet<>(exclusiveResources);
```
marcphilipp, line 49: This can be moved into the `else` branch again.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/LockManagerTests.java
```diff
@@ -93,6 +97,27 @@ void returnsWriteLockForExclusiveResourceWithBothLockModes() {
 		assertThat(locks.get(1)).isInstanceOf(WriteLock.class);
 	}
 
+	@ParameterizedTest
+	@EnumSource
+	void globalLockComesFirst(LockMode globalLockMode) {
+		Collection<ExclusiveResource> resources = asList( //
+			new ExclusiveResource("___foo", READ), //
+			new ExclusiveResource("foo", READ_WRITE), //
+			new ExclusiveResource(GLOBAL_RESOURCE_LOCK_KEY, globalLockMode), new ExclusiveResource("bar", READ_WRITE));
```
marcphilipp, line 106:
```suggestion
			new ExclusiveResource(GLOBAL_RESOURCE_LOCK_KEY, globalLockMode), //
			new ExclusiveResource("bar", READ_WRITE));
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+
+@API(status = EXPERIMENTAL, since = "5.7")
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.METHOD })
```
marcphilipp, line 25:
```suggestion
@Target(ElementType.TYPE)
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
```diff
@@ -0,0 +1,29 @@
...
+@Inherited
+@ResourceLock("__global__")
```
marcphilipp, line 27: ❓ Feedback from JUnit team call: maybe call the lock "isolated" as well?

leonard84: IMHO the name of the lock is fitting, as the lock is used by _every_ `TestDescriptor`, I would find it more confusing if every such descriptor would acquire a read lock on `__isolated__`. 
Just as a side note: https://en.wikipedia.org/wiki/Global_interpreter_lock

Maybe we could change the annotation to `@GlobalIsolation` if `@Isolated` to `__global__` is too surprising.

marcphilipp: I went with `org.junit.platform.engine.support.hierarchical.ExclusiveResource.GLOBAL_KEY` now for consistency with the constants in the `Resources` class.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTreeWalker.java
```diff
@@ -25,16 +29,20 @@
...
 		return advisor;
 	}
 
-	private void walk(TestDescriptor testDescriptor, NodeExecutionAdvisor advisor) {
+	private void walk(TestDescriptor globalLockDescriptor, TestDescriptor testDescriptor,
+			NodeExecutionAdvisor advisor) {
 		Set<ExclusiveResource> exclusiveResources = getExclusiveResources(testDescriptor);
 		if (exclusiveResources.isEmpty()) {
-			testDescriptor.getChildren().forEach(child -> walk(child, advisor));
+			advisor.useResourceLock(testDescriptor, lockManager.getLockForResources(singleton(GLOBAL_READ)));
```
leonard84, line 44: You could create a constant for the `singleton` locks, no real need to create new instances, same for the read_write below

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ExclusiveResource.java
```diff
@@ -29,6 +29,19 @@
 @API(status = EXPERIMENTAL, since = "1.3")
 public class ExclusiveResource {
 
+	/**
+	 * The key for the global resource lock that all direct children of the
+	 * engine descriptor acquire in {@linkplain LockMode#READ read} mode by
+	 * default.
+	 *
```
leonard84, line 36: Maybe extend the explanation a bit with the additional effects if any such child requests READ_WRITE for this key, i.e. lock coarsening to top-level descriptor, forced same thread exection

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTreeWalker.java
marcphilipp: Good catch. I added a `getLockForResource` method that takes a single `ExclusiveResource` now.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ExclusiveResource.java
marcphilipp: Done in f7a7742ba8c4c8dab6956a2265262920e22b9434.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Isolated.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @since 5.7
+ * @see ExecutionMode
+ * @see ResourceLock
+ */
+@API(status = EXPERIMENTAL, since = "5.7")
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+@ResourceLock("org.junit.platform.engine.support.hierarchical.ExclusiveResource.GLOBAL_KEY")
```
vlsi, line 40: I wonder if the constant could be declared in the interface, so it could be referenced from `ExclusiveResource`

marcphilipp: jupiter-api doesn't have a dependency on platform-engine or vice versa.

Issue #1224, opened 2018-01-04
Test methods cannot combine RepeatedTest and ParameterizedTest annotations


SamCarlberg:
## Overview

The title pretty much explains it. Test methods annotated with `@RepeatedTest(n)` and `@ParameterizedTest` will fail _n_ times with a `ParameterResolutionException`, then run the parameterized test as normal

#### Test class

```java
class JUnitCannotCombineRepeatedAndParameterizedTests {

  @RepeatedTest(5)
  @ParameterizedTest
  @MethodSource("source")
  void test(String param) {
    System.out.println(param);
  }

  static Stream<Arguments> source() {
    return Stream.of(Arguments.of("foo"),
                     Arguments.of("bar"));
  }

}
```

#### Expected result

The parameterized test should run 5 times with the parameters generated by the `source()` method eg

```
[repetition 1 of 5]
  foo
  bar
[repetition 2 of 5]
  foo
  bar
...
```

#### Actual result

There are 5 repetitions of the test, each one failing with this exception:

```java
org.junit.jupiter.api.extension.ParameterResolutionException: No ParameterResolver registered for parameter [java.lang.String arg0] in executable [void JUnitCannotCombineRepeatedAndParameterizedTests.test(java.lang.String)].
```

After this, the test is run with the parameters generated by `source()`

This occurs on 5.0.0 and 5.1.0-M1

sbrannen:
Well, that's actually the expected behavior.

You cannot have two `@TestTemplate` implementations executing at the same time for a single method.

In addition, it was never foreseen that one would _nest_ `@TestTemplate` implementations, which is essentially what you are proposing here.

So I'm a bit torn as to whether or not we should (or could) consider making a change like this.

@junit-team/junit-lambda, thoughts?

leksak:
I'm writing code that could make use of this functionality right now

marcphilipp:
This will be addressed by #871 which would allow to use `@RepeatedContainer` and/or `@ParameterizedContainer` on an enclosing, potentially `@Nested`, test class.

honza-zidek:
There is a Stack Overflow issue (https://stackoverflow.com/q/58005980/2886891) which would be solved by fixing this problem...

danielhodder:
I'm not sure that #871 quite captures the same  request as the one being requested here. The main use I want to use this capability for is combining different extensions together. I will give an example:

[...]

Issue #871, opened 2017-05-30
Create Container Template Extension Point


nicolaiparlog:
Test templates are a great way to create extensions that generate test cases per method. If an extension wants to generate suites of tests per class, though, this is not easily possible. I propose to implement a container template extension point that functions similarly to test template:

* it defines the number of executions `n` - but in this case per container, which makes Jupiter work all its "execute tests in this class" magic not once but `n` times
* each invocation is called with a context, which most importantly contains the index of the current iteration
* additional extensions can be registered

It might make sense to consider whether other extension points should be allowed to easily access the container invocation context, maybe from the `ExtensionContext` they already receive. (Thinking about that, wouldn't it make sense to do the same for test templates, i.e. give `ExtensionContext` a method that returns the invocation context?)

Where parameterized test methods were the canonical example of how to use test templates, theories would be the canonical example for container templates. With this extension point the theories extension would use reflection to find out how many parameters were defined and then have the class be executed so many times, each time dropping a different set of parameters into all test methods:

* `a = 0; b = 0`
    * testAddition
    * test Subtraction
* `a = 0; b = 1`
    * testAddition
    * test Subtraction
* ...

marcphilipp:
Defining `@ArgumentsSource` on the test class (as proposed in #878) would be an equivalent solution, right?

nicolaiparlog:
In parts. But it would still make the method the container for multiple tests instead of the class. And it would also require all methods to have the same parameters.

sbrannen:
Spring users often ask for such capabilities -- for example, see [SPR-16302](https://jira.spring.io/browse/SPR-16302).

sbrannen:
Although SPR-16302 is a _recent_ JIRA issue for spring, the topic has been raised multiple times over the years with regard to Spring's integration with JUnit 4. More recently, Spring users have begun to ask for such a feature in conjunction with JUnit Jupiter.

sbrannen:
With regard to generic support for iteratively executing an entire container (test class), something like `@RepeatedContainer` would be very useful for frameworks such as Spring.

Thinking out loud...

Spring already has its own annotations for declaring things like active profiles, context configuration classes, etc. Thus, it would not make sense for a parameterized Spring test class to make use of annotations from Jupiter such as `@ValueSource`, etc. Rather, a framework like Spring would be more interested in knowing which _iteration_ of the parameterized container is currently being executed so that Spring could then _pick_ the correct configuration for that iteration.

Thoughts?

[...]

PR #2409, opened 2020-09-15
Introduce @CombineTestTemplates support

danielhodder:
This adds support for combining multiple TestTemplate providers
together, in a product style, so that an implementer can use the
benefits of multiple extensions together.

Issue: #1224

## Overview

This is a fix for #1224. I had some free time so I thought I would take a look at how hard it would be. This will almost certainly not work for everyone so I have made it opt-in with a new annotation, @CombineTestTemplates. This will allow those who need this feature to use it, and everything will be normal for others.

Using the Maven test sample as a quick test we get the following test output (happy to hear suggestions for better test names): 
image

I have not yet had a chance to write test cases but I'm gonna have a crack at that later in the week if this seems interesting.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1848,6 +1848,49 @@ method with full support for the same lifecycle callbacks and extensions. Please
 NOTE: <<writing-tests-repeated-tests>> and <<writing-tests-parameterized-tests>> are
 built-in specializations of test templates.
 
+Usually if there is more than one `{TestTemplateInvocationContextProvider}` that applies
+to a method, each set is concatonated togther. This means that the extensions registered
+by each will not be available to the other. The best example is this test method
+
+[source:java]
+---
+@RepeatedTest(2)
+@ParameterizedTest
+@ValueSource(string={"a", "b"})
+public void testMethod(String value) {
+}
+---
+
+In this example the test method will be run four times; twice from the `RepeatedTest`
+annotation and twice from the `ParameterizedTest` annotation. This may not be desirable
+though since it means two of the runs will not have anything that populates the `value`
+parameter.
+
+If this is a case you need to deal with; there is an annotation `{CombineTestTemplates}`
+which changes this behaviour.
+
+[source:java]
+---
+@RepeatedTest(2)
+@ParameterizedTest
+@CombineTestTemplates
+@ValueSource(string={"a", "b"})
+public void testMethod(String value) {
+}
+---
+
+Adding this annotation causes the test runner to combine the contexts from all of the
+providers. In this example the context from the first repeated test run will be combined
+with each of the one from the parameterized test. This result in executions that can be
+described like this:
+
+{[Run 1, Parameter A], [Run 1, Parameter B], [Run 2, Parameter A], [Run 2, Parameter B]}
+
+For more details see the Javadoc on the `{CombineTestTemplates}` annotation.
+
+NOTE: Not all `{TestTemplateInvocationContextProvider}` instances will be compatible with
+this annotation.
+
 [[writing-tests-dynamic-tests]]
 === Dynamic Tests
 
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/CombineTestTemplates.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/CombineTestTemplates.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+import org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider;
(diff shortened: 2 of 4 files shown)


sbrannen:
Tentatively slated for 5.8 M1 solely for the purpose of _team discussion_

SpencerMelo:
Any idea on when we are going to be able to use it?

parnmatt:
I too am quite interested in if/when this would be accepted. 
The assignee and reviewer lists are still empty; so could be a while.

mikalaikarol:
Looking forward for it as well

mikalaikarol:
sad that there was no continuation on this PR

parnmatt:
> sad that there was no continuation on this PR

Indeed, I was hoping to use this feature at some point soon.

SpencerMelo:
It's really sad, the code was there. I don't understand why it will not move forward, for folks writing large scale tests, with a lot of data, reporting, different stuff going on, this would be useful, teams that are using junit5 at it full extents often will get annoyed by the lack of this kind of thing, anyway.

marcphilipp:
Sorry, we haven't managed to discuss this PR in the team, yet. Hence, I reopened the PR.

rmodini:
@marcphilipp thanks for reopening the pr! looking forward to this

marcphilipp:
Did any of you have any concrete use cases in mind except for combining repeated and parameterized tests (when is that actually useful?). JUnit Pioneer has [`@CartesianProductTest`](https://junit-pioneer.org/docs/cartesian-product/). Does that address your use case?

marcphilipp:
Also related: #871

danielhodder:
The original use for this was for testing database migrations if I recall. We have versioned migrations (1.0, 1.2, 2.0, etc.) which are all cumulative. We also support multiple databases. So the idea was run tests after upgrading from each version to the latest, on al the supported databases.

You'd end up with tests like:

* Test the feature works on a freshly created Oracle database
* Test the feature work on a database that started on the 1.0 schema and was upgraded to the latest on MS SQL
and so on.

The main issue I see with things like the CartesianProductTest and other similar things, is they assume a simple value. In my case I was creating new databases for each test so there's a _lot_ of setup code to get that all going, and I think each execution of the template ends up registering about half a dozen specific ParameterResolvers and Before/AfterCallback hooks.

The basic goal was to have a test helper which let you focus on writing the test code, and not gluing together the test libraries by passing dynamic parameters off to the different test helpers yourself. Basically you go from:

```java
@EnumSource(DatabaseTypes.class)
@VersionDetector
public void test(DatabaseType database, String startingVersion) {
    try (Database d = helper.createDatabase(database)) {
        DataSource source = d.getDataSource()
        // Run some scripts
        scriptHelper.runScriptsFrom(startingVersion, d.getAdminDatasource());

        try (ApplicationPlatform a = platformHelper.initialize(source)) {
            // Now test
            assertTrue(a.foo());
        }
    }
}
```

To something more like

```
@ApplicationDatabaseTest
public void test(ApplicationPlatform a) {
    assertTrue(a.foo());
}
```

And the @ApplicationDatabaseTest annotation is a meta annotation which does all the setup, script running, and application initialization. Then provides the initialized platform with a ParameterResolver. It also deals with cleaning up the test databases, and the global state the ApplicationPlatform needs to work correctly.

Hope that helps give an example of where this kind feature is super helpful. It is possible to do today but you have to write the whole thing from scratch, rather than being able to assemble it out of the existing elements Junit provides.

marcphilipp:
@danielhodder Thanks for the detailed description! I wonder if we should rather prioritize implementing #871 so that you could run the test class multiple times against different database types and the contained test methods against the different versions. Would that also work in your case?

danielhodder:
Not so much for us. We don't want to have two tests that use the same database. Basically we want a brand new database set up for each test (isolation etc.). A lot of the applications need to setup things which would collide so it's easier to not.

In fairness though that's partly because that's how our test support has always worked, and so just porting the existing behavior is easier.

Another use case for use, which I thought just today is take the previous example, and then parameterize each test with a known set of Demo Data.

Basically #871 doesn't really change the overall landscape of only being able to template a test with one provider. The goal of this PR is to create a way of having smaller, re-usable, templating methods and then compose them as needed in your test; rather than having to create a custom extension for every test class.

Don't get me wrong though #871 would be awesome and that would definitely help for other use cases we have. (Bonus points if we can have a version of that at the Suite level so you can deploy servers with it.)

ewoks:
Any updates here? Anything to be resolved before it lands to next release?

danielhodder:
I've updated the copyright notices which were failing the spotless check.

marcphilipp:
We've looked at this today in the team call. We're not yet convinced this would be easy to grasp from a user's perspective. Since #871 is less controversial and more generally useful, we think we should implement that first.

danielhodder:
I don't think I'd argue that feature would be for very advanced usage only, and very much hard to grasp, if you didn't get it. I guess this exists to solve a specific problem which can't currently be handled in extensions. If there was a way of accomplishing this outside of the core engine I'd gladly write it there. As far as I can tell that's not possible though.

This would be a feature mostly of interest to large teams build big software packages which have a lot custom test extensions, which the team may not own. In our case most of these templates are managed by our 'platform' team, while the usages are in downstream app teams. If we can combine templates then teams don't have to choose between them.

I do thing #871 is a great step and might help if it's possible to use that with inner classes. That would let you have something like:

```java
class Test {
    @Parameterized
    class InnterTest { 
         @MyDatabaseTestTempate
         public void test(DataSource db) {
         }
    }
}
```

It's not as 'clean' but it accomplishes a similar end, although it does mean you have to have separate extensions for these, rather than re-using existing ones, which was my goal.

danielhodder:
Just ran into a more simple example which might help demonstrate this one. I am trying to combine the `PamaterizedTest` annotation from JUnit 5, with the `RetryingTest` annotation from JUnit Pioneer. In the current implementation if I try and use the two together I won't get parameters on any of the runs which the retrying test template generates.

#871 would definitely help since it would also allow BeforeEach and AfterEach methods to be parameterized.

danielhodder:
I spent a bit of time and wrote a basic version of this (with a lot of hacks) here: https://github.com/danielhodder/junit-test-template-combiner. If there's enough interest I'll figure out how to publish this to maven central.

PR #2416, opened 2020-09-21
Introduce junit-platform-suite-engine

mpkorstanje:
## Overview

Implements a test engine that allows declarative execution of test suites using
the `@Suite` annotation.

Internally the Suite Engine uses the JUnit Platform Launcher. The engine works
adding the test descriptors of the discovered engines and tests as children to its
own test descriptor.

```java
package org.junit.platform.suite;

import org.junit.platform.suite.api.SelectPackages;

@Suite
@SelectPackages("org.junit.suite.testcases")
class SelectPackageSuite {

}
```

Is equivalent to:

```java
import org.junit.platform.engine.discovery.DiscoverySelectors;
import org.junit.platform.launcher.Launcher;
import org.junit.platform.launcher.LauncherDiscoveryRequest;
import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
import org.junit.platform.launcher.core.LauncherFactory;

public class Main {

    public static void main(String[] args) {
        Launcher launcher = LauncherFactory.create();
        LauncherDiscoveryRequest request = LauncherDiscoveryRequestBuilder.request()
                .selectors(DiscoverySelectors.selectPackage("org.junit.suite.testcases"))
                .build();
        launcher.execute(request);
    }
}
```

Issue: #744

---

---

### Definition of Done


mpkorstanje:
This implementation is not done yet. However before continuing it would be good to give this a review to ensure that this is fundamentally the right way to go.

Currently this PR features:

 * A declarative test suite that can be executed with Junit 5;
 * The suite supports existing selectors from `junit-platform-suite-api`;
 * and adds several new selectors to ensure parity with `DiscoverySelectors`.
 * Supports transparent translation between unique ids of the engines test descriptors and the launchers test plan.

### Implementation ###

The suite engine converts an annotated class into a discovery request. This request is executed and the resulting test plan is mapped to a tree of test descriptors. In essence the tree of test descriptors are a view on the test plan. So suppose the discovery requests produces this test plan:

```
JUnit Jupiter
|- TestA
||- method1
||- method2
|- TestB
||- method1
||- method2
```

Then that test plan is mapped to a tree of test descriptors like so:

```
SuiteEngine
|- ExampleSuite
||-JUnit Jupiter
|||- TestA
||||- method1
||||- method2
|||- TestB
||||- method1
||||- method2
````

The unique identifiers are remapped by pre-pending the unique identifier of the suite. So:

```
junit-jupiter/TestA/method1() ->  junit-suite/ExampleSuite/junit-jupiter/TestA/method1()
```

When resolving tests by unique id this is reversed.
 
### Known limitations ###

 * Only top level classes annotated with `@Suite` are recognized. 
 * No support for a `@DynamicSuite`.
 * No before or after suite hooks
 * No parallel execution
 * The selectors that were added to `junit-platform-suite-api` are not supported by `@RunWith(JUnitPlatform.class)`
 * The unique id discovery selector has intentionally not been implemented as an annotation. The unique id representation does not seem suitable for this.
 * Likewise the method discovery selector has been omitted. I could not work out a usable annotation.
 
I don't think these limitations are insurmountable, however I would consider solving them out of scope for this PR.

**Comment by mpkorstanje** [comment]
So @marcphilipp what do you think? If there are general aspects you have doubts about I'd like to hear that too.

**Comment by kriegfrj** [comment]
> So @marcphilipp what do you think? If there are general aspects you have doubts about -more a problem with the concept of a suite engine, rather then this particular implementation- I'd like to hear that too.
> 
> I did create a small demo to play around with:
> 
> * https://github.com/mpkorstanje/junit-platform-suite-demo

For reference, this technique of a "meta TestEngine" (that recursively calls other engines using the launcher API) was also a feature of the `BundleEngine` that we implemented as part of Bnd for testing in OSGi. The `BundleEngine` takes care of all the classloading magic that needs to happen in an OSGi environment and passes the loaded classes off to the other engines to run the actual tests. See https://github.com/bndtools/bnd/blob/master/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/BundleEngine.java

**Comment by mpkorstanje** [comment]
Good to know that the idea has already been tested! Any practical problems you've ran into with regards to IDEs and other bits of tooling?

>  this technique of a "meta TestEngine" (that recursively calls other engines using the launcher API) was also a feature of the BundleEngine

Unless I'm misunderstanding the purpose and function of [`findEngines`](https://github.com/bndtools/bnd/blob/master/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java#L169) it would appear that the Bundle engine is neither recursive nor using the Launcher API.

The line `.filter(engine -> engine.getId() != BundleEngine.ENGINE_ID)` would prevent the engine from calling itself.  And the engines are loaded directly via the `ServiceLoader` rather then `org.junit.platform.launcher.Launcher`.

**Comment by kriegfrj** [comment]
> Good to know that the idea has already been tested! Any practical problems you've ran into with regards to IDEs and other bits of tooling?
> 

There are bits of the code in various places that softly assume that the Engine will be at the top of the hierarchy. For example, the naming convention for the segments of the UniqueId for the tests. In our hierarchy, the BundleEngine's segment uses "engine:bnd-bundle-engine" and the sub-engines have "sub-engine:junit-jupiter", etc.

> > this technique of a "meta TestEngine" (that recursively calls other engines using the launcher API) was also a feature of the BundleEngine
> 
> Unless I'm misunderstanding the purpose and function of [`findEngines`](https://github.com/bndtools/bnd/blob/master/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java#L169) it would appear that the Bundle engine is neither recursive nor using the Launcher API.
> 
> The line `.filter(engine -> engine.getId() != BundleEngine.ENGINE_ID)` would prevent the engine from calling itself. And the engines are loaded directly via the `ServiceLoader` rather then `org.junit.platform.launcher.Launcher`.

On the contrary, it seems you've understood the purpose and function well. :smile:

Sorry, I used the terms "recursive" and "launcher API" a bit too loosely. By "recursive" I meant one `TestEngine` invoking another, which as you point out is not strictly recursive - we explicitly prevent recursion as you noted. And no, we don't use the launcher API - we bypass it to invoke the engines directly.

**Comment by marcphilipp** [comment]
@mpkorstanje Sorry for the delay! This is still on my list, just haven't had the time to do a proper review, yet.

**Comment by jlink** [comment]
I haven’t checked the implementation so maybe this has already been discussed: How are the engines called from this one being prevented from running their tests? If they are not the tests in the suite will be run twice, right?

**Comment by mpkorstanje** [comment]
> How are the engines called from this one being prevented from running their tests

They are not. Currently to avoid this the suite engine and all other engines that are not part of the suite should be explicitly included in the discovery request. While perhaps unexpected I don't think we can implement a less surprising default behavior. Did you have anything in mind?

**Comment by jlink** [comment]
> Currently to avoid this the suite engine and all other engines that are not part of the suite should be explicitly included in the discovery request. While perhaps unexpected I don't think we can implement a less surprising default behavior. Did you have anything in mind?

I think the problem occurs when you run all tests (from IDE or Gradle or Maven) since then we'd see some tests being executed twice. Maybe a reasonable default behaviour would be to NOT execute a suite at all when the request includes other test engines?

**Comment by mpkorstanje** [comment]
I'm not sure if that would be reasonable.

1. It is hard to understand why a test is not picked up -- nothing happened. On the other hand when a test is executed twice it will be apparent when looking at IDEAs and JUnits tree output that both the test engine and the suite engine ran the test. This makes the solution easier to find because you know what to stop from happening. 

And because both the problem of not having the suite executed and the problem of having tests executed twice result in the same solution it would be preferable to make the problem observable.

2. I don't believe IDEA or Maven explicitly include any engine filters. They're using either the class or package selectors. If a package or class contains tests for different engines I would expect all to be executed, regardless of whether engine executes the same set of tests or a completely different set of tests.

**Comment by dosomder** [comment]
@mpkorstanje Great work, looking for to this so I can migrate to `cucumber-junit-platform-engine`

File: junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java [hunk 1]
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	void discover() {
+		Preconditions.condition(testPlan == null, "discovery can only happen once");
+
+		LauncherDiscoveryRequest request = requestBuilder.build();
+		// @formatter:off
+		LauncherConfig launcherConfig = LauncherConfig.builder()
+				.enableTestExecutionListenerAutoRegistration(false)
+				.enablePostDiscoveryFilterAutoRegistration(false)
```
mpkorstanje, line 97: Consider explicitly disabling the discovery of launcher-discovery listeners via SPI. https://github.com/junit-team/junit5/pull/2457

carlspring:
Hi,

Are there any updates on this? What still needs to be done/tested? Is there any ETA on when this could be accepted, merged and released?

Kind regards,

Martin

marcphilipp:
@carlspring Still on my list and not forgotten, just a little short on time these days. Will try to get it in for 5.8 M1, can't promise any date, though.

carlspring:
Hi @marcphilipp ,

Thanks for your quick response!

Would you mind clarifying what still needs to be done and how much of an effort it would be? Also, just off the record and quite optimistically, when do you suppose `5.8-M1` would come out (if this feature does make it in)?

Thanks again! :)

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
mpkorstanje: We can keep this enabled. While we try to hide that a suite is discovering and executing tests using the junit platform internally, we don't need to hide that launcher discovery is started.

mpkorstanje:
@carlspring 

> Would you mind clarifying what still needs to be done and how much of an effort it would be?  ... Would this add complete support for test suites, or would more work be required?

I think the first two posts of this MR should answer that. If that leaves you with any questions you can build this locally and try it out. 

If you are looking to contribute the User Guide and Release notes have not been updated yet. I was waiting for Marcs review before writing these but in the mean time I'd welcome a PR too. Please also feel free to give the implementation a review.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteLauncherDiscoveryRequestBuilder.java
```diff
@@ -0,0 +1,182 @@
...
+import org.junit.platform.suite.api.SelectClasses;
+import org.junit.platform.suite.api.SelectClasspathResource;
+import org.junit.platform.suite.api.SelectClasspathRoots;
+import org.junit.platform.suite.api.SelectDirectories;
+import org.junit.platform.suite.api.SelectFile;
+import org.junit.platform.suite.api.SelectModules;
+import org.junit.platform.suite.api.SelectPackages;
+import org.junit.platform.suite.api.SelectUris;
+
+final class SuiteLauncherDiscoveryRequestBuilder {
```
marcphilipp, line 66: Can we share this with `junit-platform-runner`?

junit-platform-suite-engine/src/test/java/org/junit/platform/suite/engine/testsuites/SuiteSuite.java
```diff
@@ -0,0 +1,20 @@
...
+package org.junit.platform.suite.engine.testsuites;
+
+import org.junit.platform.suite.api.SelectClasses;
+import org.junit.platform.suite.api.Suite;
+
+@Suite
+@SelectClasses(SelectClassesSuite.class)
+public class SuiteSuite {
+
+}
```
marcphilipp, line 20: We should use the `java-test-fixtures` Gradle plugin and move the two sample packages to the `testFixtures` source set like we do in `junit-vintage-engine`.

marcphilipp (commented):
Great work, I think this looks very promising! I'm wondering if we could get rid of mapping between identifier and descriptor and vice-versa. I think we'd need to pass in a root `UniqueId` when calling `Launcher.discover()` and make the `TestExecutionListener` somehow accessible in `ExecutionRequest`. That's a bit tricky because `TestExecutionListener` is in `junit-platform-launcher`, though. Since I wouldn't want to make this a public API in any case, we could go for a hacky solution like a downcast, though. WDYT?

junit-platform-suite/junit-platform-suite.gradle.kts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+description = "JUnit Platform Suite (Aggregator)"
+
+dependencies {
+	internal(platform(project(":dependencies")))
+
+	api(platform(project(":junit-bom")))
+	api(project(":junit-platform-suite-api"))
+	api(project(":junit-platform-suite-engine"))
```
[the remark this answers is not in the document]
mpkorstanje: When comparing to `junit-jupiter` you would expect so yes.

However the `junit-platform-suite` module has a transitive dependency on `opentest4j` through `junit-platform-suite-engine` -> `junit-platform-engine`. When using `runtimeOnly` this dependency is not available to compile the module.

Using `runtimeOnly` does work for `junit-jupiter` because `opentest4j` is also available through `junit-jupiter-api`. However `junit-platform-suite-api` does not have a dependency on `opentest4j`.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/ClassSelectorResolver.java
```diff
@@ -0,0 +1,88 @@
...
+	ClassSelectorResolver(Predicate<String> classNameFilter, SuiteConfiguration configuration) {
+		this.classNameFilter = classNameFilter;
+		this.configuration = configuration;
+	}
+
+	@Override
+	public Resolution resolve(ClassSelector selector, Context context) {
+		Class<?> testClass = selector.getJavaClass();
+		if (isSuiteClass.test(testClass)) {
+			// Nested tests are never filtered out
```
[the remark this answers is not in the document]
mpkorstanje: Fixed in f064ebb

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/TestIdentifierAsTestDescriptor.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.platform.launcher.TestIdentifier;
+import org.junit.platform.launcher.TestPlan;
+
+/**
+ * Represents a {@link TestIdentifier} from a {@link TestPlan} as a
+ * {@link TestDescriptor} in an {@link ExecutionRequest}.
+ *
+ * @see EngineExecutionListenerAdaptor
+ */
+final class TestIdentifierAsTestDescriptor extends AbstractTestDescriptor {
```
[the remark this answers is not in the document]
mpkorstanje: Fixed in f064ebb

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestEngine.java
```diff
@@ -0,0 +1,96 @@
...
+	public String getId() {
+		return SuiteEngineDescriptor.ENGINE_ID;
+	}
+
+	/**
+	 * Returns {@code org.junit.platform.suite} as the group ID.
+	 */
+	@Override
+	public Optional<String> getGroupId() {
+		return Optional.of("org.junit.platform.suite");
```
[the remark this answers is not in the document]
mpkorstanje: Fixed in f064ebb

mpkorstanje:
> I'm wondering if we could get rid of mapping between identifier and descriptor and vice-versa. I think we'd need to pass in a root UniqueId when calling Launcher.discover() and make the TestExecutionListener somehow accessible in ExecutionRequest. That's a bit tricky because TestExecutionListener is in junit-platform-launcher, though. Since I wouldn't want to make this a public API in any case, we could go for a hacky solution like a downcast, though. WDYT?

It would at the very least break `UniqueId.getEngineId` which assumes that the engine is both the first segment in the id and named `engine`. It may also break other implicit assumptions engines make about the structure of their ids.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteLauncherDiscoveryRequestBuilder.java
mpkorstanje: While the name might create the impression that it can be used by a user to build requests to discover test suites, it is only a utility that the suite engine uses to create launcher discovery requests to discover tests the tests in a suite. In essence it maps suite annotations to a launcher discovery request.

Perhaps a better name is in order.

marcphilipp: This is exactly what `junit-platform-runner` does with the annotations, isn't it?

junit-platform-suite/junit-platform-suite.gradle.kts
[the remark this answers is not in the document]
marcphilipp: I see, but `implementation` should work, right?

junit-platform-suite-engine/src/test/java/org/junit/platform/suite/engine/testsuites/SuiteSuite.java
mpkorstanje: Ah. This is pretty nice.

Resolved with 08feabb.

marcphilipp:
> It would at the very least break `UniqueId.getEngineId` which assumes that the engine is both the first segment in the id and named `engine`. It may also break other implicit assumptions engines make about the structure of their ids.

I'm not very concerned at the moment since `UniqueId.getEngineId` would then just return `junit-platform-suite` which, technically, is correct. I doubt any engine was querying it's own ID this way before.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/ClassSelectorResolver.java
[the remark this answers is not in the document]
marcphilipp: The comment in jupiter-engine should stay, the one in suite-engine (originally copied, I assume) should be removed.

marcphilipp:
I think the CI failure is unrelated. I'll fix it on `main`.

mpkorstanje:
> I doubt any engine was querying it's own ID this way before.

I happen to know someone. :smile: 

https://github.com/cucumber/cucumber-jvm/blob/main/junit-platform-engine/src/main/java/io/cucumber/junit/platform/engine/FeatureResolver.java#L203-L223


```java

    void resolveUniqueId(UniqueIdSelector uniqueIdSelector) {
        UniqueId uniqueId = uniqueIdSelector.getUniqueId();
        // Ignore any ids not from our own engine
        if (!engineDescriptor.getUniqueId().getEngineId().equals(uniqueId.getEngineId())) {
            return;
        }

        Predicate<TestDescriptor> keepTestWithSelectedId = testDescriptor -> uniqueId
                .equals(testDescriptor.getUniqueId());

        uniqueId.getSegments()
                .stream()
                .filter(FeatureOrigin::isFeatureSegment)
                .map(UniqueId.Segment::getValue)
                .map(URI::create)
                .flatMap(this::resolveUri)
                .forEach(featureDescriptor -> {
                    featureDescriptor.prune(keepTestWithSelectedId);
                    engineDescriptor.mergeFeature(featureDescriptor);
                });
    }
```

mpkorstanje:
I do see some potential in implementing a new launcher and reusing `EngineExecutionOrchestrator` and `EngineDiscoveryOrchestrator`. I'll give it a try.

marcphilipp:
> I think the CI failure is unrelated. I'll fix it on `main`.

Fixed in 03a79c13f612099a8fc328aa2bcff23e4f8dcc2c. Please rebase.

marcphilipp:
> I happen to know someone. 😄

Hehe, that's a valid use case, of course. 🙂  I don't think that should stop us, though, since it's easily fixed and there's currently no `@SelectUniqueId` anyway.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/ClassSelectorResolver.java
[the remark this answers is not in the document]
mpkorstanje: Derp.

junit-platform-suite/junit-platform-suite.gradle.kts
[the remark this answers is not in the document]
mpkorstanje: Seems to work. Though the documentation suggest that this is for dependencies which are purely internal and not meant to be exposed to consumers - which should happen. I'm guessing this only applies internally to a gradle project?

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteLauncherDiscoveryRequestBuilder.java
mpkorstanje: Ah I see. It'd be a single class/module. I'll see about extracting that after everything else.

mpkorstanje (commented):
I've knocked down some internal walls. Looks like it is possible to reuse the `EngineDiscoveryOrchestrator` and `EngineExecutionOrchestrator`.  There still is quite a bit of plumbing left. I'll mark the PR as ready for review when I get around to finishing that up.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryOrchestrator.java
```diff
@@ -115,13 +119,14 @@ private LauncherDiscoveryResult discoverSafely(LauncherDiscoveryRequest request,
...
 	private TestDescriptor discoverEngineRoot(TestEngine testEngine, LauncherDiscoveryRequest request,
-			LauncherDiscoveryListener listener) {
+			LauncherDiscoveryListener listener, UniqueId parentId) {
 		UniqueId uniqueEngineId = UniqueId.forEngine(testEngine.getId());
+		UniqueId executionId = parentId == null ? uniqueEngineId : parentId.append(uniqueEngineId.getLastSegment()); // TODO: Do better
 		try {
-			listener.engineDiscoveryStarted(uniqueEngineId);
-			TestDescriptor engineRoot = testEngine.discover(request, uniqueEngineId);
+			listener.engineDiscoveryStarted(uniqueEngineId); // TODO: uniqueEngineId ok?
+			TestDescriptor engineRoot = testEngine.discover(request, executionId);
```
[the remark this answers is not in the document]
marcphilipp: Good question. I think we should pass in `executionId` because otherwise it would be inconsistent with what the engine passes to `selectorProcessed` and it's documented to be "the unique ID of the engine descriptor". I took a quick look and the internal implementation looked like they don't make any such assumptions.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineIdValidator.java
```diff
@@ -21,14 +24,15 @@
 /**
  * @since 1.7
  */
-class EngineIdValidator {
+@API(status = INTERNAL, since = "1.7", consumers = "suite") //TODO: Why isn't this used in testkit?
```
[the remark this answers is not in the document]
marcphilipp: Good catch! I think this should be done in the `EngineDiscoveryOrchestrator` constructor instead so we can avoid making it `public`.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteLauncherDiscoveryRequestBuilder.java
mpkorstanje: On second thought. This would require solving quite a few problems that I'd consider out of scope of this PR. I'm also not convinced the effort to remove a relatively trivial amount of not-quite-duplication is worth the added complexity of adding another module and satisfying two somewhat dissimilar use-cases..

While functionally almost the same the JUnitPlatformRunner is applying some introspection to work out if it is used as a suite or not. Unfortunately this is determined by the implicit use of either `@SelectPackages` or `@SelectClasses` rather then an explicit annotation (e.g. `@Suite`).

```java
	private LauncherDiscoveryRequest createDiscoveryRequest() {
		List<DiscoverySelector> selectors = getSelectorsFromAnnotations();

		// Allows to simply add @RunWith(JUnitPlatform.class) to any test case
		boolean isSuite = !selectors.isEmpty();
		if (!isSuite) {
			selectors.add(selectClass(this.testClass));
		}

		LauncherDiscoveryRequestBuilder requestBuilder = request().selectors(selectors);
		addFiltersFromAnnotations(requestBuilder, isSuite);
		return requestBuilder.build();
	}
```	

This could be worked around by making the `SuiteLauncherDiscoveryRequestBuilder` build a `LauncherDiscoveryRequestBuilder` rather then a `DiscoveryRequestBuilder`. Using that `LauncherDiscoveryRequestBuilder` to build the `LauncherDiscoveryRequest`. Inspecting the  `LauncherDiscoveryRequest` and adding the defaults when absent. 

This would also require making `IncludeClassNameFilter` and friends public.

```java
                LauncherDiscoveryRequestBuilder requestBuilder = suiteLauncherDiscoveryRequestBuilder.build();
                LauncherDiscoveryRequest build = requestBuilder.build();
		List<ClassSelector> classSelectors = build.getSelectorsByType(ClassSelector.class);
		List<PackageSelector> packageSelectors = build.getSelectorsByType(PackageSelector.class);
		List<IncludeClassNameFilter> includeClassNameFilters = build.getFiltersByType(IncludeClassNameFilter.class);

		// Allows to simply add @RunWith(JUnitPlatform.class) to any test case
		boolean isSuite = !classSelectors.isEmpty() && !packageSelectors.isEmpty();
		if (!isSuite) {
			requestBuilder.selectors(selectClass(this.testClass));
		} else {
			if(includeClassNameFilters.isEmpty()) {
				requestBuilder.filters(includeClassNamePatterns(STANDARD_INCLUDE_PATTERN_ARRAY));
			}
		}
		return requestBuilder.build();
```		

From the other side, the `SuiteLauncherDiscoveryRequestBuilder` currently supports adding a requests from a class by looking at it's annotations and adding a request from a unique id. The latter is needed because suites may be selected by their unique id (the class) or by the unique id of a test in that suite, or both.

This could be worked around by making the `SuiteLauncherDiscoveryRequestBuilder` build a `LauncherDiscoveryRequestBuilder` rather then a `DiscoveryRequestBuilder` but would introduce another layer of indirection as the `SuiteTestDescriptor` would have store track unique ids and test classes and build the discovery request after all selectors have been resolved delaying feedback on incorrectly used annotations.

documentation/src/test/java/example/SuiteDemo.java
```diff
@@ -0,0 +1,27 @@
...
+//tag::user_guide[]
+import org.junit.platform.suite.api.SelectPackages;
+import org.junit.platform.suite.api.Suite;
+import org.junit.platform.suite.api.SuiteDisplayName;
+
+@Suite
+@SuiteDisplayName("JUnit Platform Suite Demo")
+@SelectPackages("example")
+//end::user_guide[]
+@org.junit.platform.suite.api.IncludeClassNamePatterns
```
mpkorstanje, line 22: Both the platform launcher and platform runner only include classes with specific names when no other pattern is provided. In doing so they ignore `ExampleTestCase.failingTest` which would fail the build. 

I don't think it is a good idea to repeat this feature in the suite engine. It is a "hidden" requirement in addition to the use of the annotations. However I'm also unaware of the rationale for including it in the first place. 

What was is reason to only include classes with certain names by default?

documentation/src/docs/asciidoc/user-guide/appendix.adoc
```diff
@@ -50,10 +50,17 @@ artifacts are deployed to Sonatype's {snapshot-repo}[snapshots repository] under
...
+  `junit-platform-suite`::
+    JUnit Platform Suite artifact that transitively pulls in dependencies on
+    `junit-platform-suite-api`, `junit-platfomr-suite-engine` for simplified
+    dependency management in build tools such as Gradle and Maven.
   `junit-platform-suite-api`::
     Annotations for configuring test suites on the JUnit Platform. Supported by the
-    <<running-tests-junit-platform-runner,JUnitPlatform runner>> and possibly by
-    third-party `TestEngine` implementations.
+    <<junit-platform-suite-engine,JUnit Platform Suite Engine>> and
+    <<running-tests-junit-platform-runner,JUnitPlatform runner>>.
```
mpkorstanje, line 60: I've intentionally remove the reference to other test engines. Looking at the current implementations I don't think it is feasible for other test engines to replicate this functionality without using JUnits internals.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryOrchestrator.java
```diff
@@ -45,7 +48,7 @@
  *
  * @since 1.7
  */
-@API(status = INTERNAL, since = "1.7", consumers = "testkit")
+@API(status = INTERNAL, since = "1.7", consumers = { "testkit", "suite" })
```
mpkorstanje, line 51: The usage of `consumers` is rather inconsistent. What is the preferred usage?

```
@API(status = EXPERIMENTAL, since = "1.3", consumers = "org.junit.platform.engine.support.hierarchical")
@API(status = INTERNAL, since = "5.8", consumers = "org.junit.vintage.**")
@API(status = INTERNAL, since = "1.7", consumers = { "testkit", "suite" })
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryOrchestrator.java
```diff
@@ -72,18 +75,40 @@ public EngineDiscoveryOrchestrator(Iterable<TestEngine> testEngines,
...
+	 * <p>Applies {@linkplain org.junit.platform.launcher.EngineFilter engine
+	 * filters} and {@linkplain PostDiscoveryFilter post-discovery filters} and
+	 * {@linkplain TestDescriptor#prune() prunes} the resulting test tree.
+	 *
+	 * Additionally test engines without tests are pruned from from the discovery
+	 * result and the engines unique id will be prefixed with {@code parentId}.
+	 */
+	public LauncherDiscoveryResult discover(LauncherDiscoveryRequest request, Phase phase, UniqueId parentId) {
+		Map<TestEngine, TestDescriptor> result = discover(request, phase, parentId::appendEngine);
+		return new LauncherDiscoveryResult(pruneEngines(result), request.getConfigurationParameters());
```
mpkorstanje, line 95: Pruning the test engines is necessary because their test descriptors no longer root containers. This results in them being pruned when added as children to the parent engine. However because the `LauncherDiscoveryResult` is executed as a whole events for these empty test engine are still emitted resulting exceptions when trying to look up these nodes in in the test plan.

documentation/src/docs/asciidoc/user-guide/appendix.adoc
```diff
@@ -199,6 +208,12 @@ runner ..> launcher
 runner ..> suite_api
 runner ..> junit4
 
+suite ..> suite_api
+suite ..> suite_engine
+
+suite_engine ..> launcher
+suite_engine ..> suite_api
```
mpkorstanje, line 215: This image is becoming quite crowded now. I'm not sure how to improve it.

mpkorstanje: By adding some more space (additional dots) between the connections that cross package boundaries graphviz seems to be able to sort it out again.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteLauncherDiscoveryRequestBuilder.java
mpkorstanje: Ooh. I found a solution!

junit-platform-runner/src/main/java/org/junit/platform/runner/JUnitPlatform.java
```diff
@@ -145,160 +157,23 @@ private JUnitPlatformTestTree generateTestTree(LauncherDiscoveryRequest discover
...
+		// @formatter:on
 	}
 
-	private <A extends Annotation, V> V getValueFromAnnotation(Class<A> annotationClass, Function<A, V> extractor,
-			V defaultValue) {
-
-		A annotation = this.testClass.getAnnotation(annotationClass);
-		return (annotation != null ? extractor.apply(annotation) : defaultValue);
+	private boolean isSuite() {
+		return SUITE_ANNOTATIONS.stream().anyMatch(annotation -> testClass.getAnnotation(annotation) != null);
```
mpkorstanje, line 176: What about meta-annotations?

documentation/src/docs/asciidoc/user-guide/junit-platform-suite-engine.adoc
```diff
@@ -0,0 +1,41 @@
...
+By annotating a class with `@Suite` it is marked as a test suite on the JUnit Platform.
+As seen in the following example, selector and filter annotations can then be used to
+control the contents of the suite.
+
+[source,java,indent=0]
+----
+include::{testDir}/example/SuiteDemo.java[tags=user_guide]
+----
+
+.Additional Configuration Options
```
mpkorstanje, line 38: What about meta annotations?

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -343,8 +343,8 @@ public static ModuleSelector selectModule(String moduleName) {
 	 */
 	@API(status = EXPERIMENTAL, since = "1.1")
 	public static List<ModuleSelector> selectModules(Set<String> moduleNames) {
-		Preconditions.notNull(moduleNames, "moduleNames must not be null");
-		Preconditions.containsNoNullElements(moduleNames, "individual module name must not be null");
+		Preconditions.notNull(moduleNames, "Module names must not be null");
+		Preconditions.containsNoNullElements(moduleNames, "Individual module name must not be null");
```
mpkorstanje, line 347: For consistency.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/IsSuiteClass.java
```diff
@@ -0,0 +1,31 @@
...
+
+	private static final IsPotentialTestContainer isPotentialTestContainer = new IsPotentialTestContainer();
+
+	@Override
+	public boolean test(Class<?> testClass) {
+		return isPotentialTestContainer.test(testClass) && hasSuiteAnnotation(testClass);
+	}
+
+	private boolean hasSuiteAnnotation(Class<?> testClass) {
+		return AnnotationSupport.isAnnotated(testClass, Suite.class);
```
mpkorstanje, line 28:
```java
@Suite
@SelectPackages("org.junit.suite.testcases")
```

This is starting to look kinda redundant. A suite without selectors doesn't work anyway.

```
SelectClasses
SelectClasspathResource
SelectDirectories
SelectFile
SelectFile
SelectModules.
SelectPackages
SelectUris
```

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteLauncherDiscoveryRequestBuilder.java
mpkorstanje: Done. See: `SuiteLauncherDiscoveryRequestBuilder`.

documentation/src/docs/asciidoc/user-guide/junit-platform-suite-engine.adoc
sbrannen: In general, JUnit 5 always supports using annotations as meta-annotations. So if you're talking about finding the `@Suite` annotation as a meta-annotation, yes, we should support that.

mpkorstanje:
@marcphilipp I can't think of anything to add/improve. Ready for review again I guess.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryOrchestrator.java
```diff
@@ -173,6 +202,15 @@ private void logTestDescriptorExclusionReasons(Map<String, List<TestDescriptor>>
 		});
 	}
 
+	private Map<TestEngine, TestDescriptor> pruneEngines(Map<TestEngine, TestDescriptor> result) {
+		// @formatter:off
+		return result.entrySet()
+				.stream()
+				.filter(entry -> !entry.getValue().getChildren().isEmpty())
```
marcphilipp, line 209: ❌ Although I don't know of any engine that is completely dynamic, we should call `containsTests` here:
```suggestion
				.filter(entry -> TestDescriptor.containsTests(entry.getValue())
```

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/Configuration.java
```diff
@@ -0,0 +1,53 @@
...
+ * @see org.junit.platform.runner.JUnitPlatform
+ * @see org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder#configurationParameter(String, String)
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+@Documented
+@API(status = Status.EXPERIMENTAL, since = "1.8")
+@Repeatable(Configurations.class)
+public @interface Configuration {
```
marcphilipp, line 40: ❓ You opted for brevity here?

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/package-info.java
```diff
@@ -1,5 +1,13 @@
 /**
  * Annotations for configuring a <em>test suite</em> on the JUnit Platform.
+ *
+ * <h4>JUnit 4 Suite Support</h4>
+ * <p>Test suites can be run on the JUnit Platform in a JUnit 4 environment via
+ * {@code @RunWith(JUnitPlatform.class)} with the {@code junit-platform-runner}.
+ *
+ * <h4>JUnit 5 Suite Support</h4>
+ * <p>Test suites can be run on the JUnit Platform in a JUnit 5 environment via
+ * {@code @Suite} with the {@code junit-platform-suite-engine}.
```
marcphilipp, line 10:
```suggestion
 * {@link Suite @Suite} with the {@code junit-platform-suite-engine}.
```

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/DiscoverySelectorResolver.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.support.discovery.EngineDiscoveryRequestResolver;
+
+final class DiscoverySelectorResolver {
+
+	// @formatter:off
+	private static final EngineDiscoveryRequestResolver<SuiteEngineDescriptor> resolver = EngineDiscoveryRequestResolver.<SuiteEngineDescriptor>builder()
+			.addClassContainerSelectorResolver(new IsSuiteClass())
+			.addSelectorResolver(context -> new ClassSelectorResolver(context.getClassNameFilter(), context.getEngineDescriptor()))
+			.addTestDescriptorVisitor(context -> DiscoverySelectorResolver::discoverSuite)
```
marcphilipp, line 23: ❌ We shouldn't do this recursively, just for the engine descriptor's direct children.

documentation/src/docs/asciidoc/user-guide/appendix.adoc
```diff
@@ -173,15 +183,15 @@ jupiter ..> jupiter_api
 jupiter ..> jupiter_params
 jupiter ..> jupiter_engine
 
-jupiter_api ..> opentest4j
-jupiter_api ..> commons
+jupiter_api ....> opentest4j
+jupiter_api ...> commons
```
marcphilipp, line 187: Nice, I didn't know about these! 👍

marcphilipp (changes requested):
Very thorough work! I only found a few minor things. When this PR is merged, I'd like to release 5.8.0-M1 so we can get feedback from the community.

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/Configuration.java
mpkorstanje: I never noticed. But yes, it would be better to use `ConfigurationParameter` to retain the similarity to the discovery api.

documentation/src/docs/asciidoc/user-guide/junit-platform-suite-engine.adoc
```diff
@@ -0,0 +1,42 @@
...
+In addition to _at least one_ other test engine, you need the following artifacts and
+their dependencies on the classpath. See <<dependency-metadata>> for details regarding
+group IDs, artifact IDs, and versions.
+
+===== Explicit Dependencies
+
+* `junit-platform-suite-api` in _test_ scope
+* `junit-platform-suite-engine` in _test runtime_ scope: implementation of the
+                                `TestEngine` API for the declarative Junit Platform
+                                Suites
```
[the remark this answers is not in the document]
mpkorstanje: I couldn't work out from the documentation what the preferred pattern is. 

The `junit-jupiter` aggregator goes unmentioned and the `junit-platform-runner` details the exact scopes used. Something I don't think is possible when using `junit-platform-suite`.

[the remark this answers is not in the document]
mpkorstanje: Added a note about the aggregator. I think that should be sufficient.

mpkorstanje:
> Very thorough work! I only found a few minor things. When this PR is merged, I'd like to release 5.8.0-M1 so we can get feedback from the community.

Cheers. Looking forward.

mpkorstanje:
Btw. Should I squash all commits before they're merged?

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/DiscoverySelectorResolver.java
```diff
@@ -34,6 +33,7 @@ private static void discoverSuite(TestDescriptor descriptor) {
 
 	void resolveSelectors(EngineDiscoveryRequest request, SuiteEngineDescriptor engineDescriptor) {
 		resolver.resolve(request, engineDescriptor);
+		engineDescriptor.getChildren().forEach(DiscoverySelectorResolver::discoverSuite);
```
marcphilipp: ❓ Shouldn't this be done before pruning the tree?

marcphilipp:
> Should I squash all commits before they're merged?

Don't bother, I'll do that when merging.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/DiscoverySelectorResolver.java
mpkorstanje: Yup. :face_with_head_bandage:

mpkorstanje: Fixed  a7bb955 .

marcphilipp (approved):

marcphilipp:
Thanks so much for all the hard work, @mpkorstanje! 👍 

I sent a tweet from the team account in your honor. https://twitter.com/junitteam/status/1359130553224749059

mpkorstanje:
You're welcome. This was fun to do.

dosomder:
@mpkorstanje thanks for your work. Now that this PR is merged, could you update the answer here? https://stackoverflow.com/questions/64587739/cucumberoptions-in-cucumber-junit-platform-engine/64598725#64598725

mpkorstanje:
No, not any time soon, but feel free to post another answer.

PR #2422, opened 2020-09-23
@Disabled javadoc - use better wording to be more precise about actua…

michalbcz:
## Overview

@Disabled use word "should" in "should not be executed"

Word "should" can be interpreted as something which may happen, but it's not 100% sure. Like they say in https://tools.ietf.org/html/rfc2119. In context of JUnit5 this actually means "shall not be executed"

---

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Disabled.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Disabled.java
@@ -22,7 +22,7 @@ import org.apiguardian.api.API;
 
 /**
  * {@code @Disabled} is used to signal that the annotated test class or
- * test method is currently <em>disabled</em> and should not be executed.
+ * test method is currently <em>disabled</em> and shall not be executed.
  *
  * <p>{@code @Disabled} may optionally be declared with a {@linkplain #value
  * reason} to document why the annotated test class or test method is disabled.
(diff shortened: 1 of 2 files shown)


sbrannen:
As I [mentioned on Gitter](https://gitter.im/junit-team/junit5?at=5f6b6b5b5efe9467b10f897f), the JUnit 5 documentation does not aim to adhere to the _Key words for use in RFCs to Indicate Requirement Levels_. In the context of the `@Disabled` annotation, the word "should" can safely be interpreted as "shall" if that makes it clearer for you.

Having said that, I also do not think that the JUnit 5 documentation should aim to adhere to the _Key words for use in RFCs to Indicate Requirement Levels_, since we are not authoring any form of standardized specification.

We can of course update the documentation to make the meaning more explicit, but I would tend to favor "will" over "shall" in such cases.

If we make any such changes, we should also do so [consistently](https://gitter.im/junit-team/junit5?at=5f6b6c90efefe620f2442211), not just for `@Disabled`.

sbrannen:
Tentatively slated for 5.8 M1 solely for the purpose of _team discussion_

michalbcz:
@sbrannen I know, it wasn't meant as an offense and thanks for your quick answer on Gitter. I just thought that it would be better to do PR because others can talk about it more easily (I see you already marked it for consideration, thanks for that) and I still think it's better to use clearer language (no matter of  RFC 2119 - I just use it rather as a support for my argument). 

Anyway thanks for your work on JUnit5!

sbrannen:
> @sbrannen I know, it wasn't meant as an offense and thanks for your quick answer on Gitter. I just thought that it would be better to do PR because others can talk about it more easily (I see you already marked it for consideration, thanks for that) and I still think it's better to use clearer language (no matter of RFC 2119 - I just use it rather as a support for my argument).

Sure. No offense taken. I was merely providing full context here for those who hadn't read the discussion on Gitter (and also in order not to lose/forget the comments I'd already made).

> Anyway thanks for your work on JUnit5!

🙇

marcphilipp:
Closing in light of the above discussion.

PR #2431, opened 2020-10-01
Introduce TestSourceLocator for DynamicTests

ST-DDT:
## Overview

With dynamic tests in large classes (`TestClass`) it is difficult to find out where they were actually defined.
The containing/declaring method (`testMethod()`) helps to get a rough understanding of the source of the test and assertion stacktrace shows where it went wrong, but dynamic tests are meant to test different setups/environment/instances and might be generated/inherited from a different method or class, so it would be nice if double-clicking the test would show the actual source (`"setup3"`) of the dynamic test.

## Usage-Example

- ❌ TestClass (a)
  - ❌ testMethod (b)
    - ✔️ setup1
    - ✔️ setup2
    - ❌ setup3 `<-- This is where I (probably) need to do things`
    - ✔️ setup4

````java
class TestClass {

    @TestFactory
    List<DynamicTest> testMethod() {
        return asList(

                dynamicTest("setup1", () -> { // Might be created in a factory method
                        // [...]
                }),

                dynamicTest("setup2", () -> { // or inherited from a parent class
                        // [...]
                }),

                dynamicTest("setup3", () -> { // <--- I wish to get here (wherever "here" might be, e.g. in another/super method or class)
                        // [...]
                       assert(foo, bar); // The assertion fails here
                        // [...]
                }),

                dynamicTest("setup4", () -> {
                        // [...]
                }),

        ).map(this::resetMocks);
    }

}
````

## Content

This PR adds a `TestSourceLocator` that automatically calculates a testSourceUri if none is specified.
It isn't a finished implementation yet (It does what I believe it should do, but eclipse doesn't seem to be support it yet)
but I would like to start a discussion regarding this feature request.

I also considered creating an issue for that first, but IMO actual code is more expressive.

## Next steps

1. Discussion whether the feature request gets conceptually accepted
2. Convert to an actual PR with tests
3. Wait for merge
4. Add support to IDEs (if necessary)

---

---

### Definition of Done


sormuras:
That's an interesting way to calculate a default test source location. Will take a closer look next week.

Tentatively slated for 5.8 M1 for team discussion.

ST-DDT:
Any update on this one? Is it accepted for the upcoming milestone? Am I supposed to do something to get it accepted?

marcphilipp:
@ST-DDT We've briefly discussed this PR in our team call today. We're a bit worried that creating a stacktrace for each dynamic node might have a considerable performance impact, in particular with a large number of dynamix tests. Have you already done a performance experiment? If our hunch turns out to be true, we should make this feature opt-in via a configuration parameter (and evaluate that and create the test source in the corresponding `TestDescriptor` method as the static factory method does not have access to configuration parameters.

ST-DDT:
> We're a bit worried that creating a stacktrace for each dynamic node might have a considerable performance impact, in particular with a large number of dynamix tests. Have you already done a performance experiment? 

I haven't done any performance tests yet. I'll try to get some JMH measurements in the next few days to check that and get some somewhat reliable data.
A first search yielded this result:
https://shipilev.net/blog/2014/exceptional-performance/
Which found a cost of ~20ms per stacktrace.

> If our hunch turns out to be true, we should make this feature opt-in via a configuration parameter (and evaluate that and create the test source in the corresponding `TestDescriptor` method as the static factory method does not have access to configuration parameters.

IIRC the `TestDescriptor` is created after the `@TestFactory` has already returned, so it doesn't have access to the correct location that was used to generate the dynamic node. Unless we try to extract that information from the `Executable` (lambda) with some serious black magic.
As a last way out, we could remove the automatic `TestSource` generation and provide it as an API method, that the developer can use attach it to a dynamic node, but this would mostly remove the benefits of this PR.

ST-DDT:
According to this article you can safe quite some time, if you use the stack walker API for Java 9+ (which would require the use of multi-release jars).
https://ionutbalosin.com/2018/06/an-even-faster-way-than-stackwalker-api-for-asynchronously-processing-the-stack-frames/
Which found a cost of `~1ms + 50ms per 1000 converted stacktrace elements`.
The stackwalker api would significantly reduce the number of touched stacktrace elements.

As an alternative, we could omit the test source uri creation inside the dynamic node and just store the Exception there. The creation of the exception is very cheap (~1ms if I read the statistics later statistics and 0.1ms in the previous statistics).
The stacktrace could be "calculated"/converted async/only when needed and thus reduce the impact on the test creation itself.

marcphilipp:
> IIRC the `TestDescriptor` is created after the `@TestFactory` has already returned, so it doesn't have access to the correct location that was used to generate the dynamic node. Unless we try to extract that information from the `Executable` (lambda) with some serious black magic.

You're right.

> As a last way out, we could remove the automatic `TestSource` generation and provide it as an API method, that the developer can use attach it to a dynamic node, but this would mostly remove the benefits of this PR.

We also discussed this and came to the same conclusion.

> As an alternative, we could omit the test source uri creation inside the dynamic node and just store the Exception there. The creation of the exception is very cheap (~1ms if I read this statistics correctly and 0.1ms in the previous statistics).
> The stacktrace could be "calculated"/converted async/only when needed and thus reduce the impact on the test creation itself. Note the "expensive" part is the translation of the jvm stacktrace to the Java stack trace:

I think that's a good idea. We could store a transient exception in `DynamicNode` and convert it into a `TestSource` in `TestFactoryTestDescriptor`. That way we could make it opt-in or opt-out depending on the outcome of your performance experiments.

> According to this article you can safe quite some time, if you use the stack walker API for Java 9+ (which would require the use of multi-release jars).
> [ionutbalosin.com/2018/06/an-even-faster-way-than-stackwalker-api-for-asynchronously-processing-the-stack-frames](https://ionutbalosin.com/2018/06/an-even-faster-way-than-stackwalker-api-for-asynchronously-processing-the-stack-frames/)
> Which found a cost of `~1ms + 50ms per 1000 converted stacktrace elements`.
> The stackwalker api would significantly reduce the number of touched/converted stacktrace elements.

We already have the infrastructure set up to turn junit-jupiter-engine into a MRJAR so I think that would be the way to go. We could also consider only supporting this on JDK 9 and above so we wouldn't have to maintain two implementations.

ST-DDT:
Here the result of the first benchmark

Test-Source

````java
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class TestSourceLocatorBenchmark {

	private static final URI uri = URI.create("test");

	@Benchmark
	public DynamicNode fixedTest() {
		return DynamicTest.dynamicTest("test", uri, this::locateSource);
	}

	@Benchmark
	public URI locateSource() {
		return TestSourceLocator.locateSource(); // stack depth 17
	}

	@Benchmark
	public DynamicNode locatedTest() {
		return DynamicTest.dynamicTest("test", this::locateSource); // stack depth 20
	}

}
````

Result (Direct test source uri via stacktrace)

````txt
Benchmark                                Mode  Cnt     Score    Error  Units
TestSourceLocatorBenchmark.fixedTest     avgt    5     7,366   1,124  ns/op
TestSourceLocatorBenchmark.locateSource  avgt    5  8648,422  51,346  ns/op
TestSourceLocatorBenchmark.locatedTest   avgt    5  9220,781  65,666  ns/op
````

Creating URIs with stacktrace source uri seems to be about three orders of magnitude slower.
I'm not sure where the additional 600ns between `locateSource` and `locatedTest` originate.

If I change the `TestSourceLocator` to use MR-Jars and the StackWalker API I get the following results:

Result (Direct test source uri via stack walker)

````txt
Benchmark                                Mode  Cnt     Score     Error  Units
TestSourceLocatorBenchmark.fixedTest     avgt    5     7,984    0,972  ns/op
TestSourceLocatorBenchmark.locateSource  avgt    5  2214,013   43,695  ns/op
TestSourceLocatorBenchmark.locatedTest   avgt    5  2397,145  258,689  ns/op
````

Strangely this also affects the difference between `locateSource` and `locatedTest`.

Tomorrow I'll do some benchmarks with async/lazy stacktraces.

ST-DDT:
I did some more tests:

````java
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
public class TestSourceLocatorBenchmark {

	private static ExecutorService pool = new ThreadPoolExecutor(0, Runtime.getRuntime().availableProcessors(), 5L,
		TimeUnit.SECONDS, new LinkedBlockingQueue<>(), runnable -> {
			final Thread t = new Thread(runnable);
			t.setDaemon(true);
			return t;
		});

	@Benchmark
	public Throwable newException() {
		return new RuntimeException();
	}

	@Benchmark
	@OutputTimeUnit(TimeUnit.MILLISECONDS)
	public List<Throwable> newExceptionX1_000_000() {
		final List<Throwable> r = new ArrayList<>(1_000_000);
		for (int i = 0; i < 1_000_000; i++) {
			r.add(newException());
		}
		return r;
	}

	@Benchmark
	public URI direct() {
		final Throwable e = newException();
		final StackTraceElement ste = e.getStackTrace()[0];
		final int lineNumber = ste.getLineNumber();
		return URI.create("class:/" + ste.getClassName() + (lineNumber > 0 ? "?line=" + lineNumber : ""));
	}

	@Benchmark
	@OutputTimeUnit(TimeUnit.MILLISECONDS)
	public List<URI> directX1_000_000() {
		final List<URI> r = new ArrayList<>(1_000_000);
		for (int i = 0; i < 1_000_000; i++) {
			final Throwable e = newException();
			final StackTraceElement ste = e.getStackTrace()[0];
			final int lineNumber = ste.getLineNumber();
			r.add(URI.create("class:/" + ste.getClassName() + (lineNumber > 0 ? "?line=" + lineNumber : "")));
		}
		return r;
	}

	@Benchmark
	@OutputTimeUnit(TimeUnit.MILLISECONDS)
	public List<URI> asyncX1_000_000() throws InterruptedException, ExecutionException {
		final List<Future<URI>> f = new ArrayList<>(1_000_000);
		for (int i = 0; i < 1_000_000; i++) {
			final Throwable e = newException();
			f.add(pool.submit(() -> {
				final StackTraceElement ste = e.getStackTrace()[0];
				final int lineNumber = ste.getLineNumber();
				return URI.create("class:/" + ste.getClassName() + (lineNumber > 0 ? "?line=" + lineNumber : ""));
			}));
		}
		final List<URI> r = new ArrayList<>(1_000_000);
		for (final Future<URI> future : f) {
			r.add(future.get());
		}
		return r;
	}

	@Benchmark
	@OutputTimeUnit(TimeUnit.MILLISECONDS)
	public List<URI> parallelStreamX1_000_000() {
		final List<Throwable> f = new ArrayList<>(1_000_000);
		for (int i = 0; i < 1_000_000; i++) {
			f.add(newException());
		}

		return f.parallelStream().map(e -> {
			final StackTraceElement ste = e.getStackTrace()[0];
			final int lineNumber = ste.getLineNumber();
			return URI.create("class:/" + ste.getClassName() + (lineNumber > 0 ? "?line=" + lineNumber : ""));
		}).collect(Collectors.toList());
	}

}
````

Result:

````txt
Benchmark                                            Mode  Cnt      Score     Error  Units
TestSourceLocatorBenchmark.asyncX1_000_000           avgt    5  11497,074  594,193  ms/op
TestSourceLocatorBenchmark.directX1_000_000          avgt    5   9091,947   93,444  ms/op
TestSourceLocatorBenchmark.newExceptionX1_000_000    avgt    5   1032,536  210,882  ms/op
TestSourceLocatorBenchmark.parallelStreamX1_000_000  avgt    5   5097,745   86,647  ms/op

TestSourceLocatorBenchmark.direct                    avgt    5   8850,718   33,682  ns/op
TestSourceLocatorBenchmark.newException              avgt    5    861,982   26,859  ns/op

TestSourceLocatorBenchmark.dynamicTestStaticUri      avgt    5      7,984    0,972  ns/op
TestSourceLocatorBenchmark.dynamicTestCreatedUri     avgt    5    274,474    0,963  ns/op
TestSourceLocatorBenchmark.stackWalkerUri            avgt    5   2397,145  258,689  ns/op
````

(AdoptOpenJDK 11.0.9 @ i7-7820X CPU - 8x 3.6GHz)

The `newException` by itself is two orders of magnitude slower, than just creating a dynamic test without testSourceUri.
If we create an uri ourselves, then we are also two orders of magnitude slower, but still 3 times faster than creating an exception.

If we store the exception, then we always have the fixed cost of ~800ns/node + 8000ns/node if we actually use it.
In comparison, we have a fixed cost of ~2400ns/node if we directly use the stack walker to get the uri.
(Note: The fixed cost can be eliminated in both cases, by using some kind of static variable)

IMO the difference between the 800 vs 2400ns is not worth keeping it as a stacktrace.
I'm not sure how many dynamic nodes are created and how much it "costs" to create them.
Most projects that I know have less then 10k tests in total. Even if all of them were dynamic nodes it would only take an additional 24ms to fill in those details. Which seems negligible to me.

Should we go for Java 9+ StackWalker without toggle then? I personally also wouldn't mind the 8000ns for Java 8.

marcphilipp:
Thanks for doing the benchmarks! We'll discuss your results in our next team call.

ST-DDT:
FFR: Here a preview of the multi-release code: Stacktrace (J8) / StackWalker (J9+)
https://github.com/ST-DDT/junit5/compare/locatable-dynamic-tests...ST-DDT:locatable-dynamic-tests-j9

ST-DDT:
Any update here?

ST-DDT:
> Change is documented in the User Guide and Release Notes

Should I write documentation for this feature and where would be a good location to add it?
I'm not sure about this, because this feature works mostly out of the box without user interaction.

Will the release notes added by you, when the PR is merged or am I supposed to add it?
I couldn't find anything related to that in the contribution guidelines and a few other PRs that I looked into don't have them either.

I'll mark this PR as ready to be merged, because IMO this PR contains all that is needed. The MR-Jar and the potential other features are optional, and I will add them upon request.

marcphilipp:
**Team decision:** We've discussed this issue today and would like to wait for additional interest from the community before adding considerable complexity to the codebase.

ST-DDT:
Updated to take into account the changes that have been made to the `ClassSource` in #2637

Shinigami92:
> [...] additional interest from the community [...]

How was this interest asked? Are there any results yet?

sbrannen:
> > [...] additional interest from the community [...]
> 
> How was this interest asked? 

It was asked by making that statement and adding the `status: waiting-for-interest` label.

> Are there any results yet?

Not really. There are no up-votes (👍) for this issue and no further discussion from additional community members.

Shinigami92:
> Not really. There are no up-votes (:+1:) for this issue and no further discussion from additional community members.

Okay, I think multiple issues were made here :confused: 

1. It's a PR and not an issue and most folks just going to the issues tab in GitHub and search there for stuff.
2. I even don't expect `1.` to be really successful cause the community around junit that really goes here into GitHub is also somewhat to small.
   Cause these are only the contributors that really want something but not the community that searching for stuff that could be useful.
   The second mentioned are more going to Documentation Websites and just using this stuff.
3. No proactive (except of the label `status: waiting-for-interest`) were made. IMO your team should ask for feedback at e.g. Twitter, or the Discord Community https://join.javadiscord.net/

So in my point of view, the current expectation is somewhat like writing a little text on a post-it, pinning this post-it on downside of a bank in a park and then assume that in the next half year multiple folks have signed with an agreement the statement on it.
IMO that is not realistic.

---

Please don't take me to angry or somewhat, and I also would like to transparently talk to you and your team:
I'm a personal good friend of @ST-DDT but also a Vue.js SPA Frontend Developer.
So yeah, I personally cannot give this an upvote, cause I would not need it my self, because I'm not writing Java at all anymore, and therefore my vote would not be relevant anyway.

But he talked to me to watch this PR a long time ago and I think he put much afford into it and at least he is using it to write better test for the project in his current company. So it has a use-case.

The only thing that could be added from @ST-DDT are some user documentation, to give end users of JUnit a hint that this exists, how to use it and when it is useful.

sbrannen:
Thanks for the constructive feedback, @Shinigami92.

We plan to discuss it once the entire team has time to meet again.

ST-DDT:
Another alternative, that would still help me, is only adding the `TestSourceLocator` class without adding it to the factory methods/constructors/static initializers. That way, we have the ability to set the right location/define factory methods ourselves, without being intertwined in the library.

ST-DDT:
Merry xmas

ST-DDT:
Is there anyway I can simplify this PR to get it approved(able)?
Like removing the actual implementation related parts and just having a `SourceLocationResolver` interface/field that gets accessed?

ST-DDT:
Bump

juliette-derancourt:
This PR has been converted to an issue for the time being: #2924.

ST-DDT:
FFR: Commit-id: https://github.com/junit-team/junit5/commit/6f447168561d0b36255cac1466da0818a46e8946

PR #2448, opened 2020-10-13
Use Docker images from GitHub Container Registry

marcphilipp:


jbduncan:
Is there any way we can review the source code for the new container images?

jbduncan:
Ah, the Dockerfiles are published over at https://github.com/junit-team/build-env, aren't they? :)

marcphilipp:
Yep, that's right.

jbduncan:
LGTM so far!

Issue #806, opened 2017-04-25
Revise ClassLoader lookups across the platform


sbrannen:
## Overview

Although current use of `ClassLoaderUtils.getDefaultClassLoader()` works for the most common class loading scenarios, it is not always appropriate to default to using the thread context `ClassLoader` (TCCL) when loading user types via reflection.

This issue is inspired by the issues discussed in #805 and serves as an umbrella issue for potential class loading issues across the platform.

## Related Issues

- #201 
- #805 

## Comparison to class loading in Spring

If we take inspiration from the Spring Framework, we can see that `org.springframework.util.ClassUtils.forName(String, ClassLoader)` accepts a user-supplied `ClassLoader` and falls back to the "default ClassLoader" only if the supplied `ClassLoader` is `null`. For loading services, Spring does not use Java's `ServiceLoader` mechanism but rather its own `SpringFactoriesLoader` which performs the same task as Java's `ServiceLoader` but with explicit control over which `ClassLoader` is used. Note that `SpringFactoriesLoader` delegates to `org.springframework.util.ClassUtils.forName(String, ClassLoader)` to actually load the service class.

## Initial Analysis

JUnit does not necessarily need the flexibility of `SpringFactoriesLoader`, but for certain scenarios we will likely have to supply a custom `ClassLoader` instead of relying on the "default ClassLoader" which currently is typically the thread context ClassLoader (TCCL).

Supplying `getClass().getClassLoader()` is likely a viable option. However, we should not change the implementation of `ClassLoaderUtils.getDefaultClassLoader()`. Rather, we should simply use it less often and then only as a fallback.

FWIW, we actually did have the foresight to create `org.junit.platform.commons.util.ReflectionUtils.loadClass(String, ClassLoader)`, but... we never supply it anything other than the default ClassLoader. 😉 

## Use Cases to Investigate

The following use cases should be investigated to determine how best to look up the `ClassLoader` to use.

- `TestEngine` loading in `ServiceLoaderTestEngineRegistry`
- `TestExecutionListener` loading in `ServiceLoaderTestExecutionListenerRegistry`
- Jupiter `Extension` loading  in `ExtensionRegistry#registerAutoDetectedExtensions`
- Additional classpath entries in `ConsoleTestExecutor#createCustomClassLoader`
- `ClasspathScanner` configuration/instantiation in `ReflectionUtils`
- `ReflectionUtils#loadClass(String)`

## Deliverables

[...]

PR #2457, opened 2020-10-21
Add possibility to register launcher-discovery listeners via SPI

filiphr:
`LauncherDiscoveryListeners` can now be registered via Java's `ServiceLoader`
mechanism in addition to those
passed as part of each `LauncherDiscoveryRequest` and the default one.

## Overview

The idea of this PR is to enable registering `LauncherDiscoveryListener` through an SPI. I am creating this PR in order to hear the opinion of the JUnit Team. Is something like this acceptable for you? If it is I am going to add add the documentation and finish whatever needs to be finished from your feedback.

---

---

### Definition of Done


marcphilipp:
Tentatively slated for 5.8 M1 solely for the purpose of _team discussion_.

filiphr:
Just FYI doing something like this allows to have an SPI change the current thread context classloader and actually can help with #201 and #806. I needed to use a different classloader for loading the classes that are discovered and adding my own classloader before the discovery is started is one way to achieve this.

I am sure that this is not the only use case for this, and the use case I have does not really fit this, I am sure that there are other more suitable use cases for this.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherDiscoveryRequestBuilder.java
```diff
@@ -268,14 +273,24 @@ private LauncherDiscoveryListener getLauncherDiscoveryListener(ConfigurationPara
...
+		if (discoveryListeners.contains(defaultDiscoveryListener) && serviceLoaderListeners.isEmpty()) {
 			return LauncherDiscoveryListeners.composite(discoveryListeners);
 		}
 		List<LauncherDiscoveryListener> allDiscoveryListeners = new ArrayList<>(discoveryListeners);
-		allDiscoveryListeners.add(defaultDiscoveryListener);
+		allDiscoveryListeners.addAll(serviceLoaderListeners);
+		if (!allDiscoveryListeners.contains(defaultDiscoveryListener)) {
+			allDiscoveryListeners.add(defaultDiscoveryListener);
+		}
 		return LauncherDiscoveryListeners.composite(allDiscoveryListeners);
```
marcphilipp, line 294: I don't think this optimization is necessary. Let's get rid of the first `if`.

filiphr: Sure no problem, I was not entirely sure about this part as well, it didn't feel quite right anyways. I'll make a small optimisation so that the list is created with a correct size at least (to avoid resizes)

Frontrider:
So this would let us stuff like changing the class path for the tests?

filiphr:
I want to be really clear that the goal of this PR is not to fix the issues that need different classpath for the tests. I don't want the JUnit team to reject this PR solely on the fact of allowing users to change the classloader.

However, it indeed allows implementors of this interface to change the current thread context classloader in `LauncherDiscoveryListener#LauncherDiscoveryListener`. This only works when the discovery selectors are created without the class already loaded, i.e. only with the class name. JUnit Jupiter the uses this constructor to load the class for the tests.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherDiscoveryRequestBuilder.java
```diff
@@ -268,14 +273,24 @@ private LauncherDiscoveryListener getLauncherDiscoveryListener(ConfigurationPara
 				.map(value -> LauncherDiscoveryListeners.fromConfigurationParameter(
 					DEFAULT_DISCOVERY_LISTENER_CONFIGURATION_PROPERTY_NAME, value)) //
 				.orElseGet(LauncherDiscoveryListeners::abortOnFailure);
-		if (discoveryListeners.isEmpty()) {
+
+		List<LauncherDiscoveryListener> serviceLoaderListeners = configurationParameters.getBoolean(
+			DISCOVERY_LISTENERS_AUTODETECTION_ENABLED_PROPERTY_NAME).filter(p -> p).map(
+				p -> StreamSupport.stream(
+					new ServiceLoaderLauncherDiscoveryListenerRegistry().loadListeners().spliterator(), false).collect(
+						Collectors.<LauncherDiscoveryListener> toList())).orElse(Collections.emptyList());
```
marcphilipp, line 281: Let's try to trick Spotless into formatting this in a more readable way. Sth. like this should work:
```suggestion
		List<LauncherDiscoveryListener> serviceLoaderListeners = configurationParameters.getBoolean(
			DISCOVERY_LISTENERS_AUTODETECTION_ENABLED_PROPERTY_NAME) //
				.filter(p -> p) //
				.map(p -> StreamSupport.stream(new ServiceLoaderLauncherDiscoveryListenerRegistry().loadListeners().spliterator(), false) //
					.collect(Collectors.<LauncherDiscoveryListener>toList())) //
				.orElse(Collections.emptyList());
```

marcphilipp (changes requested):
Looks good! Could you please add this to the user guide and release notes as well?

filiphr:
> Looks good! Could you please add this to the user guide and release notes as well?


Thanks. I just pushed the update to the user guide and release notes.

One thing that I just remembered. What would you like the default flag to be set to. Currently discovering `LauncherDiscoveryListener`(s) is an opt-in. Users have to explicitly enable it via `junit.platform.discovery.listener.autodetection.enabled`

marcphilipp:
**Team decision:** Let's include this feature in 5.8.

marcphilipp (approved):
@filiphr I took the liberty of reworking a few parts based on your implementation. I hope you don't mind!

filiphr:
Really appreciated @marcphilipp. Thanks for the work on it.

One small comment though, I think that you missed a `@since 1.8` on the `isLauncherDiscoveryListenerAutoRegistrationEnabled()` in the `LauncherConfig`. I saw that you added some missing `@since 1.7` and wanted to let you know.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherConfig.java
```diff
@@ -60,6 +60,17 @@
...
 	boolean isTestEngineAutoRegistrationEnabled();
 
+	/**
+	 * Determine if launcher discovery listeners should be discovered at runtime
+	 * using the {@link java.util.ServiceLoader ServiceLoader} mechanism and
+	 * automatically registered.
+	 *
+	 * @return {@code true} if launcher discovery listeners should be
+	 * automatically registered
+	 */
```
marcphilipp, line 70:
```suggestion
 	 * @since 1.8
	 */
```

marcphilipp:
I merged this manuaally in 8b71ee6 to resolve the merge conflict. I'm glad this is now in place. Thanks for the contribution and patience, @filiphr! 👍

filiphr:
That's great news @marcphilipp. Looking forward in using this and finalising the migration to JUnit 5 in the MapStruct project.

PR #2458, opened 2020-10-23
Added osgi-test EclipseCon 2020 presentation

kriegfrj:
@marcphilipp suggested that I add our talk to the presentations list. I'm not sure where the final rendered pages are or how to produce them so that I can verify the result looks correct though.

cc @bjhargrave, @rotty3000

### DIFF
--- a/data/events.json
+++ b/data/events.json
@@ -734,5 +734,17 @@
       "marcphilipp",
       "sormuras"
     ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


marcphilipp (approved):

marcphilipp:
Thanks! However, we currently only display upcoming events based on the collected data. I'll merge your PR anyway in case we'll ever change that. That will also make it easier to add future talks by one of you guys.

PR #2464, opened 2020-10-26
osgi: Use Bnd macro processing to avoid gradle afterEvaluate

bjhargrave:
We use late-binding to get the project description via Bnd's macro
support instead of having to use afterEvaluate to early-bind to the
project description in the gradle script. 

(I forgot to include this change in the build updates for Bnd 5.2.0.)

---

---

### DIFF
--- a/buildSrc/src/main/kotlin/osgi-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/osgi-conventions.gradle.kts
@@ -23,6 +23,9 @@ tasks.withType<Jar>().matching {
 			# in the default Bundle-SymbolicName value.
 			Bundle-SymbolicName: ${'$'}{task.archiveBaseName}
 
+			# Set the Bundle-Name from the project description
+			Bundle-Name: ${'$'}{project.description}
+
 			# These are the general rules for package imports.
 			Import-Package: \
 				!org.apiguardian.api,\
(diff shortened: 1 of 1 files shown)


marcphilipp (approved):
Thanks! Are there any plans for the bnd plugin to adopt Gradle's Provider/Property APIs?

bjhargrave:
> Are there any plans for the bnd plugin to adopt Gradle's Provider/Property APIs?

Bnd 5.2.0 already does support them. In particular, the `bnd` property of `BundleTaskConvention` was updated in 5.2.0 to accept `Provider<? extends CharSequence>` for Bnd instructions. See https://github.com/bndtools/bnd/pull/4318/commits/2f810df486d1c611bb2bcd1879528e56cb361d75.

marcphilipp:
Thanks for the pointer! That would have been an alternative way to get rid of the `afterEvaluate` then, right?

bjhargrave:
> That would have been an alternative way to get rid of the `afterEvaluate` then, right?

Yes, but using the Bnd macro support, as done in this PR, is the most straightforward way. The Bnd gradle plugin provides access to the `project` and `task` properties. And these are using for several other Bnd instructions.

But you can now use a Provider to supply the Bnd instructions and that Provider would be invoked at task execution time. However, I am not sure you could not use a simple multiline string, as is currently used, since I am not sure when that string would be interpolated. You would need to make sure any interpolation was deferred until task execution time and I suspect the string would end up being interpolated at task configuration time which is too early.

PR #2499, opened 2020-12-18
Adds assertInstanceOf (proof of concept)

jbee:
Signed-off-by: Jan Bernitt <jaanbernitt@gmail.com>

## Overview

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2015-2020 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api;
+
+import static org.junit.jupiter.api.AssertionUtils.*;
+
+import java.util.function.Supplier;
+
+import org.opentest4j.AssertionFailedError;
+
+class AssertInstanceOf {
+
+	private AssertInstanceOf() {
+		/* no-op */
+	}
+
+	static void assertInstanceOf(Class<?> expectedType, Object actualValue) {
+		assertInstanceOf(expectedType, actualValue, (Object) null);
+	}
+
+	static void assertInstanceOf(Class<?> expectedType, Object actualValue, String message) {
+		assertInstanceOf(expectedType, actualValue, (Object) message);
+	}
+
+	static void assertInstanceOf(Class<?> expectedType, Object actualValue, Supplier<String> messageSupplier) {
+		assertInstanceOf(expectedType, actualValue, (Object) messageSupplier);
+	}
+
+	private static void assertInstanceOf(Class<?> expectedType, Object actualValue, Object messageOrSupplier) {
+		if (!expectedType.isInstance(actualValue)) {
+			String template = actualValue instanceof Throwable ? "Unexpected exception type"
+					: "Unexpected instance type";
+			if (actualValue == null)
+				template = "Unexpected null value";
+			String message = buildPrefix(nullSafeGet(messageOrSupplier))
+					+ format(expectedType, actualValue == null ? null : actualValue.getClass(), template);
+			throw new AssertionFailedError(message);
+		}
+	}
+}
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
@@ -3448,4 +3448,15 @@ public class Assertions {
 		return AssertTimeout.assertTimeoutPreemptively(timeout, supplier, messageSupplier);
 	}
 
+	public static void assertInstanceOf(Class<?> expectedType, Object actualValue) {
+		AssertInstanceOf.assertInstanceOf(expectedType, actualValue);
+	}
+
+	static void assertInstanceOf(Class<?> expectedType, Object actualValue, String message) {
+		AssertInstanceOf.assertInstanceOf(expectedType, actualValue, message);
+	}
+
+	static void assertInstanceOf(Class<?> expectedType, Object actualValue, Supplier<String> messageSupplier) {
+		AssertInstanceOf.assertInstanceOf(expectedType, actualValue, messageSupplier);
+	}
 }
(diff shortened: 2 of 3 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -3448,4 +3448,48 @@ public static void assertTimeoutPreemptively(Duration timeout, Executable execut
...
+	 *
+	 * <p>Like the {@code instanceof} operator a {@code null} value is not
+	 * considered to be of the {@code expectedType} and does not pass the assertion.
+	 *
+	 * <p>If necessary, the failure message will be retrieved lazily from the
+	 * supplied {@code messageSupplier}.
+	 *
+	 * @since 5.8
+	 */
+	@API(status = STABLE, since = "5.8")
```
marcphilipp, line 3491: Please mark all of the new methods "experimental".
```suggestion
	@API(status = EXPERIMENTAL, since = "5.8")
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -3448,4 +3448,48 @@ public static void assertTimeoutPreemptively(Duration timeout, Executable execut
...
+	/**
+	 * <em>Assert</em> that the supplied {@code actualValue} is an instance of the
+	 * {@code expectedType}.
+	 *
+	 * <p>Like the {@code instanceof} operator a {@code null} value is not
+	 * considered to be of the {@code expectedType} and does not pass the assertion.
+	 *
+	 * @since 5.8
+	 */
+	public static void assertInstanceOf(Class<?> expectedType, Object actualValue) {
```
marcphilipp, line 3460: I wonder if the signature of this and the other methods shouldn't look like this to avoid obviously wrong parameters at compile time:

```suggestion
	public static <T> void assertInstanceOf(Class<? extends T> expectedType, T actualValue) {
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @since 5.8
+ */
+class AssertInstanceOf {
+
+	private AssertInstanceOf() {
+		/* no-op */
+	}
+
+	static void assertInstanceOf(Class<?> expectedType, Object actualValue) {
```
marcphilipp, line 32: I think these and the other methods should return the casted value.
```suggestion
	static <T> T assertInstanceOf(Class<T> expectedType, Object actualValue) {
```

marcphilipp (changes requested):

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
jbee: I don't think this is what you want. I found that in the examples I use it you don't know what actual type you really get so if you add the restriction you would need to add a cast the actual value which then fails instead of the assert getting us back to poor error messages.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
jbee: Yeah, this feels more useful. Actual value is an `Object` and after we verified it is of type `T` we can use it as that.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
marcphilipp: You're right, I had actually tried it and meant to remove this comment before posting the review. Sorry for the noise!

jbee:
As a side note: I still could not figure out how to run tests in IDEA because when running the test this triggers kotlin build which fails. I updated to latest stable IDEA 2020.3 with Kotlin plugin 1.4.21 but I keep getting compile errors like in the attached screenshot. Which error fails the build varies.
Screenshot at 2020-12-26 16-45-35

marcphilipp:
Please correct me if I'm wrong, but it looks like you're using IntelliJ to build instead of delegating to Gradle:

Screenshot 2021-01-01 at 14 11 37

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	static <T> T assertInstanceOf(Class<T> expectedType, Object actualValue, Supplier<String> messageSupplier) {
+		return assertInstanceOf(expectedType, actualValue, (Object) messageSupplier);
+	}
+
+	private static <T> T assertInstanceOf(Class<T> expectedType, Object actualValue, Object messageOrSupplier) {
+		if (!expectedType.isInstance(actualValue)) {
+			String template = actualValue instanceof Throwable ? "Unexpected exception type"
+					: "Unexpected instance type";
```
marcphilipp, line 47: Is it really worth differentiating between exception and "normal" instances here? Couldn't we always say "Unexpected type"?

jbee:
For me the grade settings look like this 
Screenshot at 2021-01-01 15-51-38

As far as I can tell it does use gradle to build.

marcphilipp:
Hmm, that looks different. Which version of IDEA is that?

jbee:
2020.3

marcphilipp:
How did you initially open the project? Did you import it as a Gradle project?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
marcphilipp:
```suggestion
			String template = "Unexpected type";
```

marcphilipp (changes requested):

marcphilipp (approved):
Could you please add an entry to the 5.8 M1 release notes?

documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M1.adoc
```diff
@@ -31,6 +31,9 @@ on GitHub.
   `test(displayName("my test"))` can be used to match against a test whose display name is
   `my test`.
 
+* New `assertInstanceOf` methods as a replacement for `assertTrue(obj instanceof X)` which
+  provide better error messages comparable to those of `assertThrows`.
+
```
marcphilipp, line 36: This should go to the "New Features and Improvements" section of "JUnit Jupiter"

marcphilipp (changes requested):

documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M1.adoc
jbee: done

marcphilipp (approved):

marcphilipp:
@jbee Is there anything missing or can we make this PR "ready for review"?

sormuras (changes requested):
>  Using the '.*' form of import should be avoided

Please replace on demand import with single class imports.

jbee:
@sormuras imports are fixed.

sormuras (approved):

PR #2501, opened 2020-12-21
Report containers as JFR events

marcphilipp:
Prior to this commit only tests were reported as JFR events. However,
it's also useful to see events for containers, e.g. test classes, in
particular when they require expensive setup that causes other JFR
events such as garbage collection.

Moreover, this commit adds an integration test using the new JfrUnit
Jupiter extension.

Screenshot 2020-12-21 at 10 21 06

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -98,11 +98,15 @@ allprojects {
 	}
 
 	repositories {
-		// mavenLocal()
 		mavenCentral()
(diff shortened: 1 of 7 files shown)


sormuras (commented):
Using 'var' in main source files might make a backport to Java 8 more difficult.

marcphilipp:
> Using 'var' in main source files might make a backport to Java 8 more difficult.

That's true but it should simple enough.

sormuras (approved):

PR #2502, opened 2020-12-21
Add JFR events for test discovery

marcphilipp:
Since test discovery can take a considerable amount of time, we now also
report JFR events on the launcher and engine level.

Screenshot 2020-12-21 at 17 56 14


marcphilipp:
I added subcategories now:
Screenshot 2020-12-21 at 18 19 02

PR #2514, opened 2021-01-04
Introduce Launcher sessions and accompanying listener to support global setup/teardown use cases

marcphilipp:
## Overview

Implementations of `TestExecutionListener` and (starting with 5.8) `LauncherDiscoveryListener` can be registered via ServiceLoader on each `Launcher` created via `LauncherFactory`. Implementing one of those listeners, in particular the former, allows to have custom setup or teardown code executed once for the entire test run across multiple test engines. Use cases include starting and stopping an external process, replacing the class loader, initializing some library etc. However, subsequent calls to `Launcher.execute` currently cause that setup / teardown code to be executed again which is not desirable, especially if it’s slow. However, for example the Maven Surefire plugin attempts to rerun failed tests if `rerunFailingTestsCount` is set or (when [SUREFIRE-1725](https://issues.apache.org/jira/browse/SUREFIRE-1725) is released) will even execute every test class in a separate request.

We could make the `Launcher` interface extend `AutoCloseable`. Calling `close()` would signal that it’s no longer used and any teardown code should be run. Since we can’t change `Launcher` in that way without breaking existing usages (which wouldn’t call `close()`), this PR introduces a new concept: a closeable `LauncherSession`. Tools that potentially call `Launcher.discover` or `Launcher.execute` multiple times should first create a session and close it when they’re done. Instances of `LauncherSessionListener` — which are registered via ServiceLoader (by default, can be disabled) or explicitly via a `LauncherConfig` — are called when the session is opened and closed. Each listener can then get access to the `Launcher` to register discovery/execution listeners via `LauncherSession.getLauncher`, e.g. to postpone setup until the first execution.

To ensure `LauncherSessionListener` gets called even by tools that do not (yet) use sessions, the `Launcher` created via `LauncherFactory.create` opens and closes a session for each call to `Launcher.discover` or `Launcher.execute`.

### Open questions

- [ ] Should we deprecate `LauncherFactory.create` and introduce `LauncherFactory.createLauncher` next to `LauncherFactory.createSession` to raise awareness of the new concept?

### Definition of Done


junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
```diff
@@ -61,42 +63,77 @@
...
-	 * registration for test engines and test execution listeners.
+	 * Factory method for creating a new {@link LauncherSession} using the
+	 * {@linkplain LauncherConfig#DEFAULT default} {@link LauncherConfig}.
+	 *
+	 * @throws PreconditionViolationException if no test engines are detected
+	 * @see #createSession(LauncherConfig)
+	 * @since 1.8
+	 */
+	@API(status = EXPERIMENTAL, since = "1.8")
+	public static LauncherSession createSession() throws PreconditionViolationException {
```
marcphilipp, line 81: Team feedback: rename to `openSession()`.

sormuras (approved):

alterhu2020:
@marcphilipp  when this feature wiil release into official version? This interface is very important hook for global setup junit environment, thanks.

marcphilipp:
We're going to release it as part of 5.8.0-M1, probably until the end of the week.

@alterhu2020 Out of curiosity, what's your use case?

alterhu2020:
@marcphilipp  We want to initalize some global setup/teardown functions(like injecting different environment setup..) when junit is launched. 

And with using `beforeAll` and `afterAll` hook it will call in every junit testInstance class, these session listeners in this feature is very helpful I think, It provides the ability to achieve that, look forward to the new version 5.8.0-M1.  
Thanks very much for your team.

paplorinc:
@marcphilipp is awesome!

PR #2530, opened 2021-01-16
Block parallel tests longer on CI

mpkorstanje:
## Overview

Various tests in `ParallelExecutionIntegrationTests` verify that different
tests were indeed executed by different threads. To ensure that an incorrect
implementation does not block the test suite indefinitely these tests contain
an intentional race condition.

This race condition assumes awaiting at most 100ms will be sufficient to schedule
another thread. This assumption is hardware dependent and does not hold on
Github CI[1]. Increasing the maximum time spend waiting by a factor 10 while in CI
is hopefully sufficient.

1. https://github.com/junit-team/junit5/pull/2416/checks?check_run_id=1712237454
```
2021-01-16T02:00:44.8166959Z > Task :platform-tests:test
2021-01-16T02:00:44.8167430Z
2021-01-16T02:00:44.8168932Z ParallelExecutionIntegrationTests > customContextClassLoader() FAILED
2021-01-16T02:00:44.8177847Z     java.lang.AssertionError:
2021-01-16T02:00:44.8178812Z     Expected size:<3> but was:<2> in:
2021-01-16T02:00:44.8179762Z     <["ForkJoinPool-19-worker-3", "ForkJoinPool-19-worker-5"]>
2021-01-16T02:00:44.8183737Z         at org.junit.platform.engine.support.hierarchical.ParallelExecutionIntegrationTests.customContextClassLoader(ParallelExecutionIntegrationTests.java:135)
2021-01-16T02:01:05.6171111Z
2021-01-16T02:01:05.6173697Z ParallelExecutionIntegrationTests > customContextClassLoader() FAILED
2021-01-16T02:01:05.6175991Z     java.lang.AssertionError:
2021-01-16T02:01:05.6177743Z     Expected size:<3> but was:<2> in:
2021-01-16T02:01:05.6178859Z     <["ForkJoinPool-1-worker-3", "ForkJoinPool-1-worker-5"]>
2021-01-16T02:01:05.6183298Z         at org.junit.platform.engine.support.hierarchical.ParallelExecutionIntegrationTests.customContextClassLoader(ParallelExecutionIntegrationTests.java:135)
2021-01-16T02:01:09.0267715Z
2021-01-16T02:01:09.0280937Z
2021-01-16T02:01:09.0282574Z ParallelExecutionIntegrationTests > customContextClassLoader() FAILED
2021-01-16T02:01:09.0283978Z 1534 tests completed, 3 failed, 7 skipped
2021-01-16T02:01:09.0284818Z     java.lang.AssertionError:
2021-01-16T02:01:09.0290826Z     Expected size:<3> but was:<2> in:
2021-01-16T02:01:09.0291728Z     <["ForkJoinPool-1-worker-3", "ForkJoinPool-1-worker-5"]>
2021-01-16T02:01:09.0295567Z         at org.junit.platform.engine.support.hierarchical.ParallelExecutionIntegrationTests.customContextClassLoader(ParallelExecutionIntegrationTests.java:135)
```
---

---

### Definition of Done

- [X] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

### DIFF
--- a/platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionIntegrationTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionIntegrationTests.java
@@ -766,7 +766,7 @@ class ParallelExecutionIntegrationTests {
 			throws InterruptedException {
 		var value = sharedResource.incrementAndGet();
 		countDownLatch.countDown();
-		countDownLatch.await(100, MILLISECONDS);
+		countDownLatch.await(estimateSimulatedTestDurationInMiliseconds(), MILLISECONDS);
 		return value;
 	}
 
@@ -774,8 +774,23 @@ class ParallelExecutionIntegrationTests {
 			throws InterruptedException {
 		var value = sharedResource.get();
 		countDownLatch.countDown();
-		countDownLatch.await(100, MILLISECONDS);
+		countDownLatch.await(estimateSimulatedTestDurationInMiliseconds(), MILLISECONDS);
 		assertEquals(value, sharedResource.get());
+	}
+
+	/**
+	 * To simulate tests running in parallel tests will modify a shared
+	 * resource, simulate work by waiting, then check if the shared resource was
+	 * not modified by any other thread.
+	 *
+	 * Depending on system performance the simulation of work needs to be longer
+	 * on slower systems to ensure tests can run in parallel.
+	 *
+	 * Currently CI is known to be slow.
+	 */
+	private static long estimateSimulatedTestDurationInMiliseconds() {
+		var runningInCi = Boolean.valueOf(System.getenv("CI"));
+		return runningInCi ? 1000 : 100;
 	}
 
 	static class ThreadReporter implements AfterTestExecutionCallback {


marcphilipp:
Thanks for the PR! I'm a bit hesitant to merge it as it will slow down tests for everyone. I think we should make the timeout dependent on whether the `CI` env var is present or sth. like that. Thoughts?

mpkorstanje:
This will not slow down tests for every one. Due to the semantics of `CountDownLatch.await` the slowdown will only be noticeable when functionality related to parallel execution is broken.

Increasing the timeout only on CI adds a complexity to the build that is harder to trace. I don't think most of us would expect tests to detect they are being run in CI and alter their behavior (e.g. Volvo). For example in this instance I was able to quickly deduce that the build failure is most likely caused by sluggish CI. Had there been a conditional timeout I would not have been so certain.

I don't think it's a good idea to add complexity to smooth over failure cases. So personally I would very much prefer dealing with the former then the latter. Though ultimately it comes down to the projects preference.

marcphilipp:
It will actually slow down this test and other tests in the test class, because the `@ResourceLock` prevents the concurrent tests from starting and thus they don't call `countDown()`.

Before (3.3s): https://ge.junit.org/s/4nvawv23oruhw/tests#:platform-tests-test-org.junit.platform.engine.support.hierarchical-parallelexecutionintegrationtests
After (17.7s): https://ge.junit.org/s/ut3gkbqyzflpw/tests#:platform-tests-test-org.junit.platform.engine.support.hierarchical-parallelexecutionintegrationtests

I understand your concern and I totally agree that we should avoid the complexity if possible. However, in this case, I don't see another option. I think using a different timeout on CI than locally (or on Windows vs. other OSes) would be ok in this case.

mpkorstanje:
Mmh. Should those tests await the latch then? 

The latch seems to be intended to ensure a test occupies a thread so each tests will run on different threads.

I'll have to take a closer look.

marcphilipp:
> The latch seems to be intended to ensure a test occupies a thread so each test will run on different threads.

In some cases, yes. In other cases the same mechanism is used to wait long enough to be fairly certain things are _not_ executed concurrently.

mpkorstanje:
I was hoping to be able to make it explicit what each test should should wait for but that doesn't look feasible. A conditional it is then.

marcphilipp (approved):

marcphilipp:
Thanks! 👍

PR #2526, opened 2021-01-13
Introduce SHORT_DISPLAY_NAME and LONG_DISPLAY_NAME in @ParameterizedTest

chia7712:
…rizedTest

## Overview

By default, the display name of ```ParameterizedTest``` is similar to ```RepeatedTest#SHORT_DISPLAY_NAME```. If we want to make display name include "method name", we have to write string formate like ``` "{displayName} - {arguments}"``` for all test cases. It seems to me ```RepeatedTest#LONG_DISPLAY_NAME``` is a good example and it would be better to introduce ```LONG_DISPLAY_NAME``` to ```ParameterizedTest```. The benefit is that junit 5 users can replace aforementioned string format by "standard" long display name.

Resolved #2525

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement).

---

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.commons/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [ ] [Coding conventions](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [ ] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling
- [ ] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [ ] Change is documented in the [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [ ] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #2532, opened 2021-01-16
Make default parameterized display name template configurable

chia7712:
Introduce `junit.jupiter.params.displayname.default` configuration parameter to configure the default display name template for all parameterized tests.

## Overview

according to the comment (https://github.com/junit-team/junit5/pull/2526#issuecomment-760898928)

The new configuration ```junit.jupiter.params.displayname.default``` enables us to configure the display name for all parameterized test. With this new configuration, we don't need to write duplicate pattern (like ```{displayName} - {arguments}```) to all test case.

---

---

### Definition of Done


junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
```diff
@@ -39,6 +39,7 @@
 
 	private static final String METHOD_CONTEXT_KEY = "context";
 	static final String ARGUMENT_MAX_LENGTH_KEY = "junit.jupiter.params.displayname.argument.maxlength";
+	static final String DISPLAY_NAME_TEMPLATE_KEY = "junit.jupiter.params.displayname.template";
```
marcphilipp, line 42: I know I said "template" in the other issue. Looking at the sources of `@ParameterizedTest` now I think "pattern" would be better. Looking at other configuration paramters, though, I think `junit.jupiter.params.displayname.default` would be more consistent.

@junit-team/junit-lambda WDYT?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
```diff
@@ -122,13 +123,17 @@ private TestTemplateInvocationContext createInvocationContext(ParameterizedTestN
...
-				"Configuration error: @ParameterizedTest on method [%s] must be declared with a non-empty name.",
-				templateMethod));
+		// the priority of pattern is shown below.
+		// 1) configuration parameter
+		// 2) ParameterizedTest name
+		String pattern = extensionContext.getConfigurationParameter(DISPLAY_NAME_TEMPLATE_KEY).map(String::trim).orElse(
+			Preconditions.notBlank(parameterizedTest.name().trim(),
+				() -> String.format(
+					"Configuration error: @ParameterizedTest on method [%s] must be declared with a non-empty name.",
+					templateMethod)));
```
marcphilipp, line 136: Please only call `trim()` once and add validation (with an appropriate message) for the config parameter along with a test case each for both scenarios.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1589,6 +1589,15 @@ if they exceed the configured maximum length. The limit is configurable via the
...
 
+If you'd like to change the default name template for all parameterized tests in your project, you can add the following
+configuration to `junit-platform.properties`
+
+[source,properties,indent=0]
+----
+junit.jupiter.params.displayname.template = {index}
+----
+
+NOTE: `junit.jupiter.params.displayname.template` can override the `name` of `ParameterizedTest` for all test cases.
```
marcphilipp, line 1600: Actually, the logic should be inverted: `ParameterizedTest.name()` should override the config parameter. Please also document the config parameter in the Javadoc of `ParameterizedTest.name()`.

marcphilipp (changes requested):

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sbrannen: That's correct: `ParameterizedTest.name()` must override the config parameter in all cases.

But.... to make that possible, we will need to change the `default` value of `ParameterizedTest.name()`.

If we leave it declared as `String name() default DEFAULT_DISPLAY_NAME;` (i.e., the status quo), it will be impossible to know if the user explicitly set the `name` attribute to the value of `DEFAULT_DISPLAY_NAME` intentionally.

Thus, we will need to change the default of the `name` attribute from `DEFAULT_DISPLAY_NAME` to something like `"{default_display_name}"`.

If `"{default_display_name}"` is returned when invoking `name()`, we will:

1. Look up the new config param, and if it has been set, use it.
2. If the new config param has not been set, we will use the value of the `DEFAULT_DISPLAY_NAME` constant for backwards compatibility.

Otherwise, the user explicitly supplied a pattern for the `name` attribute (or their code was compiled against a previous version of the API with `DEFAULT_DISPLAY_NAME` set as the default value for `name`), and we use whatever explicit pattern is set.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
sbrannen: The term "pattern" is consistent with the existing documentation for `DEFAULT_DISPLAY_NAME`:

> Default display name pattern for the current invocation of a `@ParameterizedTest` method: `"[{index}] {argumentsWithNames}"`

sbrannen: > Looking at other configuration paramters, though, I think `junit.jupiter.params.displayname.default` would be more consistent.

Yes, let's go with `junit.jupiter.params.displayname.default`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
sbrannen: As mentioned in https://github.com/junit-team/junit5/pull/2532/files#r559023221 and subsequent comments, the priority needs to be reversed for backwards compatibility and local overrides of the new global default.

sbrannen (changes requested):

chia7712:
@sbrannen @marcphilipp Thanks for all your suggestions. I have addressed your comments. Please take a look.

sbrannen (approved):

I think this PR looks pretty good as is.

The team can take over any final polishing before merging into `main`.

But let's first see what the rest of the team has to say.

chia7712:
@sbrannen @marcphilipp any feedback? I'd like to address your comment ASAP since it would be better if this patch can be included by junit 5.8 :)

sbrannen:
> @sbrannen @marcphilipp any feedback? 

I approved the PR.

There are some minor improvements to be made to the documentation, etc., but those can be addressed by a core committer once the PR has been merged into `main`.

So we are waiting on approval from @marcphilipp.

> I'd like to address your comment ASAP since it would be better if this patch can be included by junit 5.8 :)

It looks like this can potentially make it into 5.8 M1.

chia7712:
fix conflicting file

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1649,6 +1649,20 @@ A parameterized test with named arguments ✔
...
+[source,properties,indent=0]
+----
+junit.jupiter.params.displayname.default = {index}
+----
+
+the display name for a parameterized method is determined according to the following precedence rules:
+
+1. `name` of `@ParameterizedTest`, if present
+2. the value of `junit.jupiter.params.displayname.default` (from junit-platform.properties), if present
+3. by calling `DEFAULT_DISPLAY_NAME` of `@ParameterizedTest`
```
marcphilipp, line 1664:
```suggestion
3. `DEFAULT_DISPLAY_NAME` constant defined in `@ParameterizedTest`
```

marcphilipp (approved):

PR #2566, opened 2021-03-11
Use JDK 11 to run Gradle

marcphilipp:
In preparation for the toolchain upgrade to JDK 16, we now run the build with the JDK 11 LTS version again.


sormuras (approved):
lgtm

PR #2574, opened 2021-03-18
Test on JDK 16

marcphilipp:


sormuras (approved):
I confess, I'm addicted to records! ❤️

PR #2581, opened 2021-03-26
Enable Test Distribution

marcphilipp:
This PR applies the `com.gradle.enterprise.test-distribution` plugin that uses Test Distribution (TD) agents connected to ge.junit.org to execute tests remotely. TD is disabled by default and can be enabled by setting the `enableTestDistribution` project property to `true`. In order to use Test Distribution, the user running the build needs an account on ge.junit.org with the TD role and an access key. All JUnit 5 team members can use TD locally via their existing access keys used to publish build scans.

On CI, an access key for a technical user is injected via the `GRADLE_ENTERPRISE_ACCESS_KEY` environment variable. Since secrets are only injected into builds on the main repo, PR builds cannot use TD. Builds originating in the main repo will use TD for the Linux and the JDK 17 jobs. The Windows and macOS jobs currently don't use TD since there are no Windows/macOS TD agents connected to ge.junit.org and we want to keep running the tests, not just the build, on those operating systems.


sormuras (approved):
Looks good to me; especially the cleanup applied to the integration tests.

Issue #2219, opened 2020-03-19
Multiple declarations of `@TempDir` in same parameter list or class declaration should fail


marcphilipp:
If a test/lifecycle method has multiple parameters annotated with `@TempDir` or a test class has multiple fields annotated with `@TempDir`, the corresponding test class/method should fail in order to prevent user errors that stem from assuming them to be different directories.

## Deliverables

- [x] Multiple parameters of the same callable must not inject the same temp dir
- [x] Multiple `static` fields must not inject the same temp dir
- [x] `static` field temp dir cannot be injected again in instance field
- [x] Multiple instance fields must not inject the same temp dir
- [x] `static` or instance field temp dir cannot be injected again as parameter

marcphilipp:
@Marcono1234 We reconsidered: https://github.com/junit-team/junit5/issues/1967#issuecomment-857598034

PR #2598, opened 2021-04-16
Add identifier element to `TempDir` annotation

### LINKED ISSUE #1967: Multiple @TempDirs should result in multiple directories
I just got bitten by the (IMHO counter-intuitive) behavior of `@TempDir`: When using it in multiple places, each injection point gets the **same** directory injected. That is, the following does not do what one would intuitively expect:

```java
void test(@TempDir File source, @TempDir File target) {
    ...
}
```

Both `source` and `target` point to the same directory!

Of course, you can work around this:

```java
void test(@TempDir File tempDir) {
    File source = new File(tempDir, "source");
    File target = new File(tempDir, "target");
    ...
}
```

But this feels unnecessarily complex for this, IMHO, rather common use case.

If, on the other hand, injecting the same temporary directory in multiple places is _really_ what is desired, one can even imagine the following API to still support that use case:

```java
@TempDir("id") Path samePath;

void test(@TempDir("id") File sameDir) {
    ...
}
```

## Deliverables

- [ ] An update to `TemporaryDirectory` which sets up (and tears down) a new temporary directory at each injection point.

sormuras:
## Overview

Add identifier element to `TempDir` annotation.

Closes #1967

---

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -136,7 +138,8 @@ public boolean supportsParameter(ParameterContext parameterContext, ExtensionCon
 	public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
 		Class<?> parameterType = parameterContext.getParameter().getType();
 		assertSupportedType("parameter", parameterType);
-		return getPathOrFile(parameterType, extensionContext);
+		String id = parameterContext.findAnnotation(TempDir.class).map(TempDir::value).orElse("");
```
marcphilipp, line 141: ❓ Should we validate whether `TempDir.value()` returns `null`?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -146,9 +149,10 @@ private void assertSupportedType(String target, Class<?> type) {
 		}
 	}
 
-	private Object getPathOrFile(Class<?> type, ExtensionContext extensionContext) {
+	private Object getPathOrFile(Class<?> type, ExtensionContext extensionContext, String id) {
+		String key = id.isEmpty() ? KEY : KEY + '+' + id;
```
marcphilipp, line 153: ⭕ I think we can just use the id and remove `KEY` since we're using our own namespace.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -146,9 +149,10 @@ private void assertSupportedType(String target, Class<?> type) {
...
 		Path path = extensionContext.getStore(NAMESPACE) //
-				.getOrComputeIfAbsent(KEY, key -> createTempDir(), CloseablePath.class) //
+				.getOrComputeIfAbsent(key, __ -> createTempDir(), CloseablePath.class) //
```
marcphilipp, line 155: ⭕ I liked your idea of using the key as a prefix for the temp dir (Maybe as a suffix to `TEMP_DIR_PREFIX`). We'd probably have to sanitize the string a bit for that, e.g. remove all non-alphanumeric characters. WDYT?

marcphilipp (commented):
Looks promising! I guess test coverage is pending, right?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
sormuras: Values returned by an annotation element are never `null`: https://stackoverflow.com/a/41309428/1431016

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
sormuras: Will remove (or inline) the `KEY`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
sormuras: Sanitizing is brittle at best: https://stackoverflow.com/a/31564306/1431016 -- not using the key as a prefix is the safe bet.

marcphilipp: **Team decision:** Let's not use the key as a prefix for now since sanitizing might be a rabbit hole.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDir.java
```diff
@@ -73,4 +73,11 @@
 @Documented
 @API(status = EXPERIMENTAL, since = "5.4")
 public @interface TempDir {
+	/**
+	 * Returns an identifier for temp directory.
+	 *
+	 * @return an identifier for the underlying temp directory, where an empty String represents the old identifier
```
Marcono1234, line 79: Could you please more in depth describe what the identifier is for?
Additionally it is most likely not clear to the users what "the old identifier" means.

Marcono1234 (commented):
You are probably not finished with this, but would it be possible to adjust the Javadoc to make it more clear that the same temp directory is reused unless different IDs are used?

Additionally the following sentences of the Javadoc would probably need to be rephrased as well to refer to multiple director<ins>ies</ins>:
```
<p>The temporary directory is only created
```
```
<p>The scope of the temporary directory
```
```
[...] each test
will use its own temporary directory.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDir.java
marcphilipp: @Marcono1234 Thanks for the review! I've improved the Javadoc in one of the latest commits. PTAL!

sormuras (commented):
LGTM, @marcphilipp (Note: I can not approve my own PR... 😉)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryTests.java
```diff
@@ -884,4 +891,96 @@ void nested() {
...
 	}
+
+	static class MultipleTempDirsTestCase {
+
+		@TempDir("a")
+		static Path A;
+		static Path B;
+
+		@BeforeAll
+		static void beforeAll(@TempDir("a") Path dir1, @TempDir("b") Path dir2) {
```
Marcono1234, line 902: The test uses the same ID for multiple `@TempDir` paths, e.g. field `A` and parameter `dir`.
I thought that this will not be allowed in the future according to https://github.com/junit-team/junit5/issues/1967#issuecomment-850821121?

You are even mixing `static` with non-`static` scopes here, e.g. `static` `A` with parameter `dir4` of `beforeEach`.

marcphilipp: You're right in pointing out that I will have to change that when tackling the other issue.

Marcono1234: Thanks for clarification. If you don't mind I would repurpose this thread here for discussion about 1967 and this implementation (to avoid spamming all the watchers of that issue).

In your comment on 1967 you wrote:
> forbid injecting the same temp dir twice on the same level.

What is the "same level"? I can currently only think of applying this to the "same scope", including sub-scopes, i.e.:
- Multiple `static` fields must not inject the same temp dir
- `static` field temp dir cannot be injected again in instance field
- Multiple instance fields must not inject the same temp dir
- `static` or instance field temp dir cannot be injected again as parameter
- Multiple parameters of the same callable must not inject the same temp dir
- Parameters of any setup / teardown methods and any test methods must not inject the same temp dir

With these restrictions in place the definition of how `@TempDir` works can then also greatly be simplified (allowing to avoid any IDs at all): "Every element annotated with `@TempDir` creates a separate new temp directory"
This would remove the need to "guess the semantics" (as you expressed in a comment on 1967).

Your ID approach would have one minor advantage though (assuming you indeed disallow multiple `@TempDir`s to inject the same directory): Existing unit tests which use multiple `@TempDir` to refer to the same directory would break cleanly with a concise exception message. With my proposed change above (removal of ID), such unit tests might fail with confusing assertion exceptions (since the tests themselves, instead of JUnit test setup, would fail). However, then again, how likely is it that such unit test code exists? Maybe we are trying to consider a case here which in reality never or extremely rarely exists, at the cost of making the revised `@TempDir` for all future users more complicated.

Additionally your comment "That's exactly how the new solution works" on 1967 is confusing me. I though that this will be forbidden since injecting the same temp directory will not be possible.
And really, that user should (probably) declare a `static` `@TempDir` field instead of reusing the same ID. With a field you get IDE support and can easily see which test methods access this field. When using an ID for this you have no direct IDE support at all, you have to perform plain text searches in your code, which is rather awful and error prone in case formatting of annotations is not consistent (e.g. different amount of whitespaces).

I have yet to see a use case where multiple `@TempDir` refering to the same directory gives any advantage compared to using an instance or `static` field instead. (Though maybe there exist some, feel free to point me to any of them.)

marcphilipp: We shouldn't be discussing this here but on #2219. 😉 

> Parameters of any setup / teardown methods and any test methods must not inject the same temp dir

I agree with your list of restrictions except for this one. I don't think we should break the following valid scenario:

```java
class ValidTest {

	@BeforeAll
	static void beforeAll(@TempDir Path tempDir) {
		// prepare files all tests need but won't change
	}

	@BeforeEach
	void beforeEach(@TempDir Path tempDir) {
		// prepare files that may be destroyed by each test
	}

	@Test
	void testRequiringTempDir(@TempDir Path tempDir) {
		// use tempDir
	}

	@Test
	void anotherTestRequiringTempDir(@TempDir Path tempDir) {
		// use tempDir
	}

	@Test
	void oneOffTestNotRequiringTempDir() {
		System.out.println("b = " + b);
		System.out.println("c = " + c);
	}
}
```

> Additionally your comment "That's exactly how the new solution works" on 1967 is confusing me. I though that this will be forbidden since injecting the same temp directory will not be possible.

see above example

Marcono1234: Personally I think for this specific example it would be cleaner to have one single `tempDir` `static ` field and then have every method access that field instead. This would make the scope of the temp directory clearer and it would be obvious which methods share this directory. The fact that a parameter is somehow linked to a parameter of another method (e.g. `beforeAll`'s `tempDir` and `testRequiringTempDir`'s `tempDir`) goes against the normal scope of parameters which is per method. 

(Though maybe this boils down to personal code style preferences?)

Issue #2594, opened 2021-04-12
Discover tests in suites packaged into jars, e.g. TCKs


t1:
As discussed [in the discussions](https://github.com/junit-team/junit5/discussions/2589), we need a way to package a test suite into a jar and run it for several implementations. This is a common requirement when building Technology Compatibility Kits. There are generally three artifacts:

1. `api` contains some interfaces and a `ServiceLoader` (or similar IoC mechanism) to load an implementation of the api interfaces.
2. `impl1` containing one implementation (sometimes called a reference implementation `ri`) of the `api`; includes the `services` files for the `ServiceLoader`.
3. `tck` the test suite to check that an implementation of the api is compliant.

You can run the tests in the `tck` by adding a `runtime` dependency on `impl1`. But the tck should be usable by any other implementation as well: I want the `impl1` to load the `tck`-jar and run it, not the other way around. The `tck` should be a library of tests that the implementations can include as a `test` dependency. The dependency should be `impl1→tck` not `tck→impl1`.

I assume that scanning all jars for tests would result in unacceptable performance, so we probably need some way to specify the jars to scan, or even better to mark a jar to be a test-suite.

I currently need this for an open source project: https://github.com/t1/smallrye-graphql It's about the TCK in the client. But I've also built a minimal, reproducible example that may be easier to play around with: https://github.com/t1/junit-tck-mre.

## Deliverables

- [ ] A mechanism to specify the jars to scan for tests
- [ ] Update to the User Manual

mpkorstanje:
> I assume that scanning all jars for tests would result in unacceptable performance, so we probably need some way to specify the jars to scan, or even better to mark a jar to be a test-suite.

I would expect that running the `ConsoleLauncher` with either  `--select-module`, `--select-package` or `--scan-classpath` should suffice or alternatively a declarative test suite from https://github.com/junit-team/junit5/pull/2416 with the equivalent selectors.

t1:
Is there a plan as to when this will be released?

mpkorstanje:
#2415 has been released in the latest preview and will presumably be generally available once the 5.8.0 backlog is cleared out.

t1:
I got it working with 5.8.0-M1. Thanks a lot!

jbduncan:
@t1 Ooh, nice! Am I right to think that you got it working over at https://github.com/smallrye/smallrye-graphql/issues/796?

[...]

PR #2618, opened 2021-05-16
Add DisableParentConfigurationParameters annotation to junit-platform-suite-api

mpkorstanje:
## Overview

Where practical the capabilities of the Junit Platform Launcher API and the
JUnit Platform Suite API should retain parity. Currently the the Suite API is
missing the ability to control whether or not implicit parameters are used.
This PR aims to correct that omission.

A potential concrete use case would be the execution of tcks (e.g. #2594)
where the execution of the imported test suite should not be influenced by
implicit properties.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/EnableImplicitConfigurationParameters.java
+++ b/junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/EnableImplicitConfigurationParameters.java
@@ -13,7 +13,6 @@ package org.junit.platform.suite.api;
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Inherited;
-import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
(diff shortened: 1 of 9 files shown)


junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/EnableImplicitConfigurationParameters.java
```diff
@@ -0,0 +1,41 @@
...
+@Target(ElementType.TYPE)
+@Inherited
+@Documented
+@API(status = Status.EXPERIMENTAL, since = "1.8")
+public @interface EnableImplicitConfigurationParameters {
+
+	/**
+	 * By default implicit configuration parameters are considered.
+	 */
+	boolean enabled() default true;
```
mpkorstanje, line 40: This is not consistent with `UseTechnicalNames` which are not used by default. The alternative would have been to create a `DisableImplicitConfigurationParameters` instead. However  `@UseTechnicalNames` is specific to the `JUnitPlatformRunner` only and I think it is better to mirror the discovery api like all other annotations do.

platform-tests/src/test/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilderTest.java
```diff
@@ -375,11 +378,68 @@ void metaAnnotations() {
...
+			assertTrue(canary.isEmpty());
+		});
+	}
+
 	private static <T> T exactlyOne(List<T> list) {
 		assertEquals(1, list.size());
 		return list.get(0);
 	}
 
+	private static void withTestServices(Runnable runnable) {
```
mpkorstanje, line 430: Duplicated from `LauncherFactoryTests`. TODO: Deduplicate.

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/EnableImplicitConfigurationParameters.java
sbrannen: I have not reviewed this PR other than this comment, but I would like to point out that an annotation with a single attribute with a default value (that represents the default behavior) is typically not something we would introduce since the annotation can be declared without any attributes and would zero effect on the behavior.

Normally we would introduce a marker annotation for such a purpose, something like `@DisableImplicitConfigurationParameters` that declares no attributes.

Of course, if it's foreseeable that one may later need to configure other behaviors related to configuration parameters, `@ConfigurationParameters` with use case specific attributes might be a better choice -- for example, `@ConfigurationParameters(disableImplicitParameters = true)`.

mpkorstanje: Cheers. I do have an other use case in mind that is related to suite configuration parameters. I'll add that to the MR so we have something concrete to discuss.

sbrannen:
- Superseded by #2618.

juliette-derancourt:
@sbrannen Is it superseded by... itself? 🤯

sbrannen:
> @sbrannen Is it superseded by... itself? 🤯

Oh. No. Sorry. I guess I got the open tabs confused.

I meant...

* Superseded by #2623.

Issue #2619, opened 2021-05-17
Introduce a Unique ID tracking TestExecutionListener


sbrannen:
## Overview

Inspired by the simple [`TestClassTrackingTestExecutionListener`](https://github.com/sbrannen/spring-framework/blob/77a115f7172b5b8874391ccfb4ffd800c0b4ba1e/spring-core/src/testFixtures/java/org/springframework/core/testfixture/junit/platform/TestClassTrackingTestExecutionListener.java) that I introduced as a proof of concept in the test suite for the Spring Framework, we should introduce a `UniqueIdTrackingListener` as an implementation of the `TestExecutionListener` API for the JUnit Platform.

`UniqueIdTrackingListener` would track the unique IDs of all "tests" (not containers) that were executed during the execution of test plan. It would track all tests, regardless of the success/failure status of the test, and write the unique IDs to a text file.

The text file can then be used by a build tool or plugin to execute the same set of tests again without having to query the user configuration for the test plan and without having to perform classpath scanning again to discover tests.

This can be useful for test environments such as within a native image -- for example, a GraalVM native image -- in order to rerun the exact same tests from a standard JVM test run within a native image.

## Deliverables

- [x] Introduce `UniqueIdTrackingListener` in the `junit-platform-launcher` module.
- [x] Register the `UniqueIdTrackingListener` by default via Java's `ServiceLoader` mechanism.
- [x] Make the `UniqueIdTrackingListener` an opt-in feature. If a specific JVM system property has not been set to `true`, the registered `UniqueIdTrackingListener` would serve as a no-op listener and not do anything.
- [x] Make the output directory and output file configurable.
- [x] Make the output directory convention-based by default, choosing an appropriate build directory for Maven or Gradle, depending on the presence of `pom.xml` or `*.gradle`/`*.gradle.kts` respectively.
- [x] Document `UniqueIdTrackingListener` in Javadoc.
- ❌ Document `UniqueIdTrackingListener` in the User Guide.
  - Since the User Guide currently has no section dedicated to `TestExecutionListener` implementations, this will be addressed in a separate issue.
- [x] Document `UniqueIdTrackingListener` in the Release Notes.

sbrannen:
Reopening this issue in order to track _skipped_ tests as well.

sbrannen:
Reopening to support concurrent use cases. See https://github.com/graalvm/native-build-tools/issues/84 for details.

PR #2630, opened 2021-05-28
Introduce a Unique ID tracking TestExecutionListener

sbrannen:
## Overview

This is a draft PR for #2619.

The feature should be functionally complete.

This PR is "draft" status for review of the feature in general but more importantly to gain consensus on the names of the system properties that are used to configure the listener.

Documentation will be added once there is consensus on the implementation and naming for the system properties.

All commits will be squashed prior to merging with an appropriate commit message.

---

### Definition of Done


sbrannen:
@lazar-mitrovic and @sdeleuze, feel free to provide feedback on this PR.

marcphilipp:
**Team decision:** The names of the system properties look good.

Another idea that came up was instead of using system properties directly, we should make the `ConfigurationParameters` available to listeners as part of the `TestPlan`. @sbrannen WDYT?

sbrannen:
> **Team decision:** The names of the system properties look good.

Great. Thanks for reviewing them.

> Another idea that came up was instead of using system properties directly, we should make the `ConfigurationParameters` available to listeners as part of the `TestPlan`. @sbrannen WDYT?

Well, that was going to be a follow up PR, but I see that you all already came to the same conclusion. 😉 

I was thinking of either introducing a new default method in `TestExecutionListener` or introducing a new interface, something like `ConfigurationParametersConsumer` or `ConfigurationParametersAware` (inspired by similar interfaces from the Spring Framework).

The new method could be `void setConfigurationParameters(ConfigurationParameters)`.

marcphilipp:
Why not make it part of the `TestPlan` since it has the same scope?

sbrannen:
> Why not make it part of the `TestPlan` since it has the same scope?

Ahhhh.... you're proposing we simply introduce `ConfigurationParameters getConfigurationParameters()` in the `TestPlan` itself.

Yes, I think that's a better idea. 👍 

I created #2631 to address that.

sbrannen:
This has been merged into `main` in a34d435f7bd2c8bbb5e30cc8ad72371df2404451.

PR #2644, opened 2021-06-16
Optionally import org.apiguardian.api in OSGi metadata

### LINKED ISSUE #2547: API Guardian should be optionally imported in OSGi metadata
From: https://github.com/junit-team/junit5/pull/2095#pullrequestreview-314557992:

> 2\. API guardian must not be imported (as previously discussed in #1353).

From #1353:

> > As https://github.com/apiguardian-team/apiguardian is also a compile-time dependency, does it also be OSGi-fied?
> 
> No, that should not be necessary since it's not a runtime dependency.

My question: if API guardian is not intended to be a runtime dependency, then why is it defined with `RUNTIME` retention? I feel that if it has runtime retention, users would be entitled to expect it to be available at runtime.

Therefore, I feel that for consistency reasons, we should do one of the following:
* either API Guardian should be osgi-ified and it should be imported by bundles that use it (can be an optional import as the bundle will run without it), or
* we should consider changing the retention of `@API` to `CLASS`.

Thoughts?

kriegfrj:
## Overview

Fixes #2547.

A "regression" in the context of this issue (compared with the current situation) would be if importing API Guardian suddenly became required, as that would break existing systems. To guard against such a regression, I modified the `verifyOSGi` task so that it excludes the API Guardian jar/bundle from consideration by the resolver. Thus the test will prove that API Guardian is not required for resolving JUnit 5 bundles.

---

---

### Definition of Done

### DIFF
--- a/buildSrc/settings.gradle.kts
+++ b/buildSrc/settings.gradle.kts
@@ -1,1 +1,7 @@
-// intentionally left blank
+dependencyResolutionManagement {
+	versionCatalogs {
+		create("libs") {
+			from(files("../gradle/libs.versions.toml"))
+		}
+	}
+}
--- a/buildSrc/src/main/kotlin/osgi-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/osgi-conventions.gradle.kts
@@ -6,6 +6,12 @@ plugins {
 	`java-library`
 }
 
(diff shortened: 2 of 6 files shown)


marcphilipp (approved):
Thanks! 👍

PR #2650, opened 2021-06-21
Bump API Guardian version to 1.1.2-SNAPSHOT

### LINKED ISSUE #2547: API Guardian should be optionally imported in OSGi metadata

kriegfrj:
## Overview

Because API Guardian 1.1.2-SNAPSHOT now has OSGi metadata, bumping to this dependency allows Bnd to automatically calculate the correct version of the package to import. This allows removing the extra gymnastics that were inserted to calculate it manually.

Fixes #2547 (again!)

---

---

### Definition of Done

### DIFF
--- a/buildSrc/settings.gradle.kts
+++ b/buildSrc/settings.gradle.kts
@@ -1,7 +1,1 @@
-dependencyResolutionManagement {
-	versionCatalogs {
-		create("libs") {
-			from(files("../gradle/libs.versions.toml"))
-		}
-	}
-}
+// intentionally left blank
--- a/buildSrc/src/main/kotlin/osgi-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/osgi-conventions.gradle.kts
@@ -6,11 +6,7 @@ plugins {
 	`java-library`
 }
 
-// Because the current build version of apiguardian (1.1.1) doesn't have OSGi metadata, bnd
-// cannot automatically determine the correct version to use in the Import-Package statement.
-// Fetch the current version out of the libs.versions.toml file.
-val apiguardianVersion = project.extensions.getByType(VersionCatalogsExtension::class).named("libs").findVersion("apiguardian").get()
-val importAPIGuardian = "org.apiguardian.*;version=\"\${range;[==,+);${apiguardianVersion}}\";resolution:=\"optional\""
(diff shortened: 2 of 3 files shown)


kriegfrj:
I'm not sure what is causing the CI to fail. The same error was happening when I ran Gradle locally - even when I ran it on a clean checkout of `main` without any of my own changes. This did not happen when this PR was based on an earlier revision of main. Suspect something changed in 55721b258a0d64505dc4fc65f8a98cbabc2f0868 (I rebased onto the commit just before this one and it builds fine locally).

marcphilipp (approved):

marcphilipp:
> I'm not sure what is causing the CI to fail.

repo.gradle.org is having issues. I've fixed it on `main`. Could you please rebase?

kriegfrj:
So do we need to do a full release of API Guardian to Maven Central? Or change those tests to allow snapshot dependencies?

marcphilipp:
I think we should change those tests.

@sormuras Could you please take a look?

marcphilipp:
I had a brief chat with @sormuras about this. I'll rework those tests to have the local jar injected instead of downloading them. I'm hopeful I can get to it tonight. 🤞

marcphilipp:
@kriegfrj This has now been solved on `main`. Please rebase!

kriegfrj:
> @kriegfrj This has now been solved on `main`. Please rebase!

Done!

marcphilipp:
@kriegfrj Thanks (again!) 😉

PR #2653, opened 2021-06-28
Support AutoCloseable for arguments provided to a @ParameterizedTest

### LINKED ISSUE #2597: Support AutoCloseable for arguments provided to a @ParameterizedTest
## Overview

Sometimes the arguments provided to a `@ParameterizedTest` are resources that need to be cleaned up after each invocation of the parameterized test -- for example, when testing against various implementations of an API. However, it is currently not possible to clean up such resources without duplicated logic in each `@ParameterizedTest` method that makes use of the arguments source.

One _could_ store the argument injected into the `@ParameterizedTest` method in a field and then access that field within an `@AfterEach` method, but that is more work than simply invoking the cleanup method directly from within each `@ParameterizedTest` method. The following simplified example demonstrates the latter approach.

```java
class ServerTests {

	@ServerTest
	void json(Server server) {
		server.handleRequest("JSON");
		server.shutdown();
	}

	@ServerTest
	void xml(Server server) {
		server.handleRequest("XML");
		server.shutdown();
	}


	private interface Server {

		default void handleRequest(String data) {
			System.out.println("Request: " + data + " for " + getClass().getSimpleName());
		}

		default void shutdown() {
			System.out.println("Shutting down " + getClass().getSimpleName());
		}
	}

	private static class MockServer implements Server {
	}

	private static class RealServer implements Server {
	}


	@ParameterizedTest
	@MethodSource("servers")
	@interface ServerTest {
	}

	static Stream<Server> servers() {
		return Stream.of(new MockServer(), new RealServer());
	}

}
```

The following are the results of executing the above test class.

```
Request: XML for MockServer
Shutting down MockServer
Request: XML for RealServer
Shutting down RealServer
Request: JSON for MockServer
Shutting down MockServer
Request: JSON for RealServer
Shutting down RealServer
```

If we want to automate the shutdown of each server, we can implement a custom `ArgumentsProvider` that stores the arguments in the `ExtensionContext.Store` and modify the `Server` interface so that it extends `CloseableResource`. The following example demonstrates this approach.

```java
class ServerTests {

	@ServerTest
	void json(Server server) {
		server.handleRequest("JSON");
	}

	@ServerTest
	void xml(Server server) {
		server.handleRequest("XML");
	}


	private interface Server extends CloseableResource {

		default void handleRequest(String data) {
			System.out.println("Request: " + data + " for " + getClass().getSimpleName());
		}

		default void shutdown() {
			System.out.println("Shutting down " + getClass().getSimpleName());
		}

		@Override
		default void close() {
			shutdown();
		}
	}

	private static class MockServer implements Server {
	}

	private static class RealServer implements Server {
	}

	@Target(ElementType.METHOD)
	@ParameterizedTest
	@ArgumentsSource(ServerArgumentsProvider.class)
	@interface ServerTest {
	}

	static class ServerArgumentsProvider implements ArgumentsProvider {

		private static final Namespace namespace = Namespace.create(ServerArgumentsProvider.class);

		@Override
			MockServer mockServer = new MockServer();
			RealServer realServer = new RealServer();
			context.getStore(namespace).put("mockServer", mockServer);
			context.getStore(namespace).put("realServer", realServer);
			return Stream.of(arguments(mockServer), arguments(realServer));
		}
	}
}
```

The following are the results of executing the above test class.

```
Request: XML for MockServer
Request: XML for RealServer
Shutting down RealServer
Shutting down MockServer
Request: JSON for MockServer
Request: JSON for RealServer
Shutting down RealServer
Shutting down MockServer
```

Here we see that the `Server` instances are in fact automatically shut down; however, there are two undesired side effects of this approach:

1. Each `Server` instance is held onto until after _all_ invocations of a given `@ParameterizedTest` have completed.
2. The `Server` instances are closed/shut down in reverse order.

The latter is due to a feature of the `ExtensionContext.Store` that supports "wrapping" of closable resources, so that behavior is to be expected in this scenario, but it may not be acceptable depending on the actual use case.

Aside from the above runtime side effects, the implementation requires a lot of boilerplate code to get working.

## Proposal

To simplify such use cases, we should invoke the `close()` method on any `CloseableResource` (or `java.lang.AutoCloseable`) that is provided as an argument to a `@ParameterizedTest`, and that should occur after each invocation of the `@ParameterizedTest` method. Doing so would allow the original `ServerTest` example to be implemented using `@MethodSource` and having the `Server` extend `CloseableResource` without the need for a custom `ArgumentsProvider` and avoiding direct interaction with the `ExtensionContext.Store`.

## Deliverables

- [ ] Support `java.lang.AutoCloseable` for `@ParameterizedTest` arguments, enabled by default but configurable via a new `autoCloseArguments` attribute in the `@ParameterizedTest` annotation.

juliette-derancourt:
## Overview

Closes #2597.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M2.adoc
@@ -76,7 +76,7 @@ on GitHub.
   directory and any contained directories instead of failing to delete them.
 * New `assertThrowsExactly` method which is a more strict version of `assertThrows`
   that allows you to assert that the thrown exception has the exact specified class.
-* New `autoCloseArguments` attribute on `@ParameterizedTest` to close `AutoCloseable`
+* New `autoCloseArguments` attribute in `@ParameterizedTest` to close `AutoCloseable`
   arguments at the end of the test. This attribute defaults to true.
 
 
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M2.adoc
@@ -76,6 +76,8 @@ on GitHub.
   directory and any contained directories instead of failing to delete them.
 * New `assertThrowsExactly` method which is a more strict version of `assertThrows`
   that allows you to assert that the thrown exception has the exact specified class.
+* New `autoCloseArguments` attribute on `@ParameterizedTest` to close `AutoCloseable`
+  arguments at the end of the test. This attribute defaults to true.
 
 
 [[release-notes-5.8.0-M2-junit-vintage]]
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1404,6 +1404,7 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsSource_examp
 include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsProvider_example]
 ----
 
+
 [[writing-tests-parameterized-tests-argument-conversion]]
 ==== Argument Conversion
 
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1105,6 +1105,10 @@ parameterized method at the same index in the method's formal parameter list. An
 _aggregator_ is any parameter of type `ArgumentsAccessor` or any parameter annotated with
 `@AggregateWith`.
 
+NOTE: Arguments of type `AutoCloseable` will be closed once `@AfterEach` methods and
+`AfterEachCallback` implementations have been called. To prevent this from happening, set
+the `autoCloseArguments` attribute in `@ParameterizedTest` to `false`.
+
 [[writing-tests-parameterized-tests-sources]]
 ==== Sources of Arguments
 
@@ -1399,11 +1403,6 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsSource_examp
 ----
 include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsProvider_example]
 ----
-
-NOTE: Arguments of type `AutoCloseable` will be closed once `@AfterEach` methods and
-`AfterEachCallback` implementations have been called. To prevent this from happening, set
-the `autoCloseArguments` attribute in `@ParameterizedTest` to `false`.
-
 
 [[writing-tests-parameterized-tests-argument-conversion]]
 ==== Argument Conversion
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1402,7 +1402,7 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsProvider_exa
 
 NOTE: Arguments of type `AutoCloseable` will be closed once `@AfterEach` methods and
 `AfterEachCallback` implementations have been called. To prevent this from happening, set
-attribute `autoCloseArguments` of `ParameterizedTest` to false.
+the `autoCloseArguments` attribute in `@ParameterizedTest` to `false`.
 
 
 [[writing-tests-parameterized-tests-argument-conversion]]
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -1400,6 +1400,10 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsSource_examp
 include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsProvider_example]
 ----
 
+NOTE: Arguments of type `AutoCloseable` will be closed once `@AfterEach` methods and
+`AfterEachCallback` implementations have been called. To prevent this from happening, set
+attribute `autoCloseArguments` of `ParameterizedTest` to false.
+
 
 [[writing-tests-parameterized-tests-argument-conversion]]
 ==== Argument Conversion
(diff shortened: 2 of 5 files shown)


sbrannen (approved):

documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M2.adoc
```diff
@@ -76,6 +76,8 @@ on GitHub.
   directory and any contained directories instead of failing to delete them.
 * New `assertThrowsExactly` method which is a more strict version of `assertThrows`
   that allows you to assert that the thrown exception has the exact specified class.
+* New `autoCloseArguments` attribute on `@ParameterizedTest` to close `AutoCloseable`
```
sbrannen, line 79:
```suggestion
* New `autoCloseArguments` attribute in `@ParameterizedTest` to close `AutoCloseable`
```

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1400,6 +1400,10 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsSource_examp
 include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsProvider_example]
 ----
 
+NOTE: Arguments of type `AutoCloseable` will be closed once `@AfterEach` methods and
+`AfterEachCallback` implementations have been called. To prevent this from happening, set
+attribute `autoCloseArguments` of `ParameterizedTest` to false.
```
sbrannen, line 1405:
```suggestion
the `autoCloseArguments` attribute in `@ParameterizedTest` to `false`.
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
```diff
@@ -207,4 +207,15 @@
...
 
+	/**
+	 * If true, all arguments of the parameterized test implementing {@link AutoCloseable}
+	 * will be closed after {@code @AfterEach} methods and {@code AfterEachCallbacks}
+	 * have been called.
+	 *
+	 * @since 5.8
+	 * @see java.lang.AutoCloseable
+	 * @see ParameterizedTestParameterResolver
+	 */
```
sbrannen, line 218: Please add the `@API` declaration.

sbrannen (changes requested):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
juliette-derancourt: Do we put `@API` declarations on fields? The other ones don't have that :thinking: 
If so, would it be `STABLE`?

sbrannen: > Do we put `@API` declarations on fields?

Only if such fields were part of the public API. For example, we would use `@API` for public constants.

However, `autoCloseArguments` is not technically a field. It's an annotation attribute which is technically a method in an interface, and it's a new part of the public API.

> If so, would it be `STABLE`?

We normally label new features like this `EXPERIMENTAL` until the next minor release (i.e., until it's been battle tested for a few months in the wild).

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
```diff
@@ -65,6 +72,42 @@ public Object resolveParameter(ParameterContext parameterContext, ExtensionConte
...
+		}
+
+		Store store = context.getStore(NAMESPACE);
+		AtomicInteger argumentIndex = new AtomicInteger();
+
+		Arrays.stream(arguments) //
+				.filter(AutoCloseable.class::isInstance) //
+				.map(AutoCloseable.class::cast) //
+				.map(CloseableArgument::new) //
+				.forEach(closeable -> store.put("closeableArgument" + argumentIndex.getAndIncrement(), closeable));
```
sbrannen, line 93: That turned out _nice & concise_. 👍

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTest.java
marcphilipp: I also think we should go with `EXPERIMENTAL` here.

marcphilipp (approved):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestParameterResolver.java
juliette-derancourt: Yay, thanks again for the tip! :ok_hand:

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -1400,6 +1400,10 @@ include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsSource_examp
 include::{testDir}/example/ParameterizedTestDemo.java[tags=ArgumentsProvider_example]
 ----
 
+NOTE: Arguments of type `AutoCloseable` will be closed once `@AfterEach` methods and
+`AfterEachCallback` implementations have been called. To prevent this from happening, set
+the `autoCloseArguments` attribute in `@ParameterizedTest` to `false`.
+
```
sbrannen, line 1406: I just realized this note is in the `@ArgumentSource` section; whereas, it applies to all arguments from any _source_.

I know that all other `@*Source` annotations are meta-annotated with `@ArgumentSource`, but the average user doesn't.

So I think we should add this note higher up in the chapter -- for example, in the _Consuming Arguments_ section.

@juliette-derancourt / @marcphilipp, thoughts?

sbrannen (changes requested):
The PR looks very good now!

I only added one request to move the _NOTE_ in the user guide.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
juliette-derancourt: I put it here as "at the end of the whole _Sources of Arguments_ section" more than in the `@ArgumentSource` section particularly, but you're right it might be better the _Consuming Arguments_ section.

PR #2654, opened 2021-06-28
Resolvable.osgi.result

rotty3000:
## Overview

---

---

### Definition of Done


rotty3000:
I'm not sure how I could have possible reduced code coverage. Any insights anyone can offer here?

junit-platform-commons/junit-platform-commons.gradle.kts
```diff
@@ -11,11 +13,23 @@ dependencies {
...
 
 tasks.jar {
 	val release9ClassesDir = sourceSets.mainRelease9.get().output.classesDirs.singleFile
 	inputs.dir(release9ClassesDir).withPathSensitivity(PathSensitivity.RELATIVE)
+
+	withConvention(BundleTaskConvention::class) {
+		bnd("""
+			Require-Capability:\
+				org.junit.platform.engine;\
+					filter:='(&(org.junit.platform.engine=*)(version>=${'$'}{version_cleanup;${rootProject.property("version")!!}})(!(version>=${'$'}{versionmask;+;${'$'}{version_cleanup;${rootProject.property("version")!!}}})))'
```
kriegfrj, line 29: I don't think that this is the way to go, for a couple of reasons:
1. `commons` is a utility library. It doesn't require an engine to be present in order to provide the functionality that it is designed to provide. So this doesn't seem right the right place to put this. If anything, this requirement belongs on the launcher bundle, but even there I think it is out of place (by way of comparison, `org.apache.felix.scr` does not have a `Require-Capability: osgi.service;filter:='(&(objectClass=*)` directive to wildcard include some stuff).
2. I'm not quite sure what the resolver does with the filter "*" - will it pull in exactly one match (indiscriminately), or all matches? Neither of these seem correct:
   1. If it's pulling in all of them, then it may pull in engines that aren't needed.
   2. If it's only pulling in one, then there is no guarantee that it is the right one.

In the case, where you've only got the Jupiter engine and Jupiter tests in your repo, it will all work by default. But if you've got mixed tests in your workspace you might not get the right bundle assembly. You might even end up with a set of `-runbundles` that resolves at runtime but which silently skips a bunch of tests (maybe even all of the tests!) because the engine those tests require has not been included in `-runbundles`.

Which brings us back to the fundamental point (which this PR is designed to address): the question "which engines are required?" is a question that applies to tests. If it is a JUnit 3/4 test, then the test bundle requires the Vintage engine; if it a Jupiter test then it requires the Jupiter engine. As these requirements only make sense when considered with reference to a particular test, then the most natural place to declare them is in the manifest of the bundle that contains the test (ie, the subject of bndtools/bnd#4719).

junit-jupiter-engine/junit-jupiter-engine.gradle.kts
```diff
@@ -39,6 +41,9 @@ tasks {
 					org.junit.platform.engine;\
 						org.junit.platform.engine='junit-jupiter';\
 						version:Version="${'$'}{version_cleanup;${project.version}}"
+				Require-Capability:\
+					org.junit.platform.launcher;\
+						filter:='(&(org.junit.platform.launcher=junit-platform-launcher)(version>=${'$'}{version_cleanup;${rootProject.property("platformVersion")!!}})(!(version>=${'$'}{versionmask;+;${'$'}{version_cleanup;${rootProject.property("platformVersion")!!}}})))'
```
kriegfrj, line 46: I think I'm ok with the way you've done the Req/Prov on the `org.junit.platform.launcher` namespace (although it's basically the same as a `RequireBundle`). I can see that it might make bndtools/bnd#3795 practically redundant (though I think the theoretical argument for something like bndtools/bnd#3795 is compelling - generally speaking, if you require a particular bundle to be present at runtime, then you must also require its requirements).

kriegfrj (commented):
Here's my thoughts!

marcphilipp:
> I'm not sure how I could have possible reduced code coverage. Any insights anyone can offer here?

There's a race condition in `ExtensionValuesStore` that sometimes caused a line to be covered and sometimes not. Don't worry about it!

I'll let you, the OSGi expert group, discuss amongst yourselves first but I'll then need a few more details to understand the objective of this PR. 🙂

junit-platform-commons/junit-platform-commons.gradle.kts
rotty3000: > commons is a utility library. It doesn't require an engine to be present in order to provide the functionality that it is designed to provide. So this doesn't seem right the right place to put this. If anything, this requirement belongs on the launcher bundle, but even there I think it is out of place (by way of comparison, org.apache.felix.scr does not have a Require-Capability: osgi.service;filter:='(&(objectClass=*) directive to wildcard include some stuff).

The point of it was to make every client lib dependent on having an engine at runtime. I don't think there is a valid argument against that since all client APIs have a hard dependency on commons. Therefore if you make the requirement on commons no matter how many more client APIs you create they will always result in requiring an engine at runtime.

> I'm not quite sure what the resolver does with the filter "*" - will it pull in exactly one match (indiscriminately), or all matches? Neither of these seem correct:
>
>    If it's pulling in all of them, then it may pull in engines that aren't needed.
>    If it's only pulling in one, then there is no guarantee that it is the right one.

It "requires" at least one to exist, which one is dependent on what's available in the resolution repos. I've never used both together so if that's possible, then I guess you're right that we can make each one targetable, but there is no need to change this wildcard. We can pair this concept with a per engine requirement which the dev can use. Let me add that capability.

> Which brings us back to the fundamental point (which this PR is designed to address): the question "which engines are required?" is a question that applies to tests. If it is a JUnit 3/4 test, then the test bundle requires the Vintage engine; if it a Jupiter test then it requires the Jupiter engine.

This I can easily add. Stay tuned...

rotty3000: Actually, you don't even have to add anything to the existing.

All a test developer has to do to add both engines is to write the following in the bnd file of their test bundle (bundle which contains the `Test-Cases` header):

```
Require-Capability: \
    org.junit.platform.engine;filter:='(org.junit.platform.engine=junit-jupiter)',\
    org.junit.platform.engine;filter:='(org.junit.platform.engine=junit-vintage)'
```
OR if they prefer the configuration-as-code approach:
```
@org.osgi.annotation.bundle.Requirement(
    namespace = "org.junit.platform.engine", name = "junit-jupiter")
@org.osgi.annotation.bundle.Requirement(
    namespace = "org.junit.platform.engine", name = "junit-vintage")
```
the resolve process will take care of the rest.

rotty3000: @kriegfrj I do now see your point about the model with commons however. since vintage engine depends on it. Making some tweaks to the model.

rotty3000:
Thanks @marcphilipp I really just wanted to make sure I hadn't done something foolish.

As for the nature of this PR. The idea is to try to eliminate some boiler plate, or worse, custom code in the bnd tooling chain to accomplish a trick that well defined Cap&Req accomplishes on it's own. That is to allow the OSGi resolver to calculate a fully complete system, essentially a closure of bundles (jars) that can satisfy all requirements at runtime.

With the proposed solution, I don't need to do much feeding of configuration into the OSGi resolver to accomplish that. I can merely write a test:

```
class FooTests {
    @Test
    public void aTest() {...}
}
```

and then let the resolver tell me if I have appropriate bundles available in my dependency closure (in my repos) to satisfy the above. If I don't have an engine in my dependencies (say as a testRuntimeOnly dependency) I will get a resolution error telling me which requirement is not satisfied.

junit-platform-commons/junit-platform-commons.gradle.kts
kriegfrj: > Actually, you don't even have to add anything to the existing.
> 
> All a test developer has to do to add both engines is to write the following in the bnd file of their test bundle (bundle which contains the `Test-Cases` header):
> 
> *snip*
>
> the resolve process will take care of the rest.

Correct! That's exactly what is needed. (Btw, the solution proposed in bndtools/bnd#4719 is for Bnd to do exactly this on behalf of the developer to relieve them of the burden of needing to do it manually.)

These Req-Caps make the Req-Cap on commons redundant though. At best, the requirement on commons will pull in the same engine; at worst, it will pull in a different one that's not actually required.

junit-jupiter-engine/junit-jupiter-engine.gradle.kts
kriegfrj: Yes, I can see the value in decoupling the capability namespace from the bsn.

junit-platform-commons/junit-platform-commons.gradle.kts
rotty3000: @kriegfrj I'm going to resolve this convo for now in favour of going over the updates :)

rotty3000:
This is what the new changes consist of:
- `junit-jupiter-api` & `junit-jupiter-params` are targeted specifically at `junit-jupiter-engine`. Should test code use either of those they should transitively require the `junit-jupiter-engine`
- if at the same time, or separately a developer writes Junit 4or3 tests AND want to use junit-platform, all they need to do is add
  ```java
  @Requirement(
	namespace = "org.junit.platform.engine",
	name = "junit-vintage"
  )
  ```
  
BOTH resolve equally well together or separately with the proposed changes.

junit-platform-commons/junit-platform-commons.gradle.kts
rotty3000: > Btw, the solution proposed in bndtools/bnd#4719 is for Bnd to do exactly this on behalf of the developer to relieve them of the burden of needing to do it manually.

First off, with this PR in the case of jupiter tests there's nothing for the dev or bnd to do. It's already working transparently. Add dependencies -> write tests -> resolve -> run tests.

Now, the hard case is JUnit 3&4 where you have two possible runtimes from bnd's point of view, the legacy tester or junit-platform. From the test developer point of view they are still just writing tests. They then have to have opted-into one of the two runtimes. Now in the case of opting to use junit-platform I don't see it as being possible for bnd to optimize away the decision to pull in the vintage engine. I just think that the one extra line for the proper requirement fits nicely beside the existing opt-in as pretty low friction since it's done exactly once. Furthermore, the frequency that we'll see both junit-platform AND the vintage engine used together is no worth optimizing that even further in my opinion.

kriegfrj:
> This is what the new changes consist of:
> 
> * `junit-jupiter-api` & `junit-jupiter-params` are targeted specifically at `junit-jupiter-engine`. Should test code use either of those they should transitively require the `junit-jupiter-engine` (which transitively requires a launcher)

This is definitely an improvement over putting the requirement on `commons`, but I'm still not 100% convinced that it is the appropriate way to go.

Consider the classic DS example, where you have an API bundle, client bundle and impl bundle. I see this as analogous to the case we have now (junit-jupiter-api and junit-junit-params are the API bundles, junit-jupiter-engine is the impl bundle and the test bundle is the client bundle). In classic DS example, the API doesn't have any prov/reqs - it's the impl bundle that has the Provide-Cap and the client bundle that has the Require-Cap.

Likewise, it feels wrong that the API bundles should be requiring the implementation bundle.

Which is why, at this stage, I'm still leaning towards something like bndtools/bnd#4719. But I think I could be convinced, especially if @bjhargrave weighs in on your side.

> * if at the same time, or separately a developer writes Junit 4or3 tests AND want to use junit-platform, all they need to do is add
>   ```java
>   @Requirement(
>     namespace = "org.junit.platform.engine",
>     name = "junit-vintage"
>   )
>   ```
>

As noted above, I know you're concerned about the case where you're using the old tester - my idea for fixing that was to add Provide-Cap for junit-vintage-engine to `biz.aQute.tester`, as effectively the `biz.aQute.tester` and `biz.aQute.junit` bundles have the equivalent capability to junit-vintage-engine. This seems to be a natural way of handling that. Then it wouldn't matter if JUnit 3/4 test bundles where automatically annotated with a Require-Cap.

rotty3000:
FYI, I've updated to Required capabilities to be `effective:=active` which is to say that the requirements are only considered by tooling (i.e. assembly) rather than being considered at runtime.

rotty3000:
> In classic DS example, the API doesn't have any prov/reqs - it's the impl bundle that has the Provide-Cap and the client bundle that has the Require-Cap.

Why doesn't the API have a requirement on an impl? That's just a design decision not a hard and fast rule. But that isn't the scenario we have here anyway.

> Likewise, it feels wrong that the API bundles should be requiring the implementation bundle.

I can't see why! First of all all Jupiter annotations are RUNTIME retention, not only that their intent is to be processed by a runtime actor, the engine. So they logically have a requirement that an engine be present at runtime. Now, we haven't coupled them to a specific implementation but rather we have merely required that there be a provider of the `junit-jupiter` capability. That could legitimately be provided by any bundle.

bjhargrave:
> Likewise, it feels wrong that the API bundles should be requiring the implementation bundle.

This is not wrong. It is the actual design point for the [osgi.implementation namespace](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.namespaces.html#service.namespaces-osgi.implementation.namespace).

See https://github.com/osgi/osgi/blob/b5e212ddf80748a68ce7773b4deaba37b6c219eb/org.osgi.service.component/src/org/osgi/service/component/annotations/Component.java#L42, for example, where the DS API requires a DS runtime.

bjhargrave (approved):

buildSrc/src/main/kotlin/publishing-conventions.gradle.kts
```diff
@@ -23,6 +23,7 @@ tasks.withType<PublishToMavenLocal>().configureEach {
 }
 
 signing {
+	useGpgCmd()
```
marcphilipp, line 26: ❓ Why did you add this?

junit-jupiter-api/junit-jupiter-api.gradle.kts
```diff
@@ -13,4 +15,20 @@ dependencies {
 	compileOnlyApi(libs.apiguardian)
 
 	compileOnly(kotlin("stdlib"))
+
+	osgiVerification(projects.junitJupiterEngine)
+	osgiVerification(projects.junitPlatformLauncher)
```
marcphilipp, line 20: ❓ What is the `osgiVerification` configuration used for?

junit-jupiter-api/junit-jupiter-api.gradle.kts
```diff
@@ -13,4 +15,20 @@ dependencies {
...
+}
+
+tasks {
+	jar {
+		withConvention(BundleTaskConvention::class) {
+			bnd("""
+				Require-Capability:\
+					org.junit.platform.engine;\
+						filter:='(&(org.junit.platform.engine=junit-jupiter)(version>=${'$'}{version_cleanup;${rootProject.property("version")!!}})(!(version>=${'$'}{versionmask;+;${'$'}{version_cleanup;${rootProject.property("version")!!}}})))';\
+						effective:=active
```
marcphilipp, line 30: Is the intention of this to fail at test runtime when no bundle with this capability, i.e. the Jupiter engine, is not present?

buildSrc/src/main/kotlin/publishing-conventions.gradle.kts
rotty3000: In order to test I had to `releaseToMavenLocal` which triggers signing. The signing plugin has a long standing bug on Linux. The official fix is to apply the above line. As referenced in the commit message see https://docs.gradle.org/current/userguide/signing_plugin.html#sec:using_gpg_agent. Probably better to just drop the commit from this PR. Stand by...

junit-jupiter-api/junit-jupiter-api.gradle.kts
rotty3000: This configuration is used to add dependencies to the artifacts visible to the resolver, in this case in order to calculate a complete closure of "bundles" needed a runtime we need providers for the new requirements.

junit-jupiter-api/junit-jupiter-api.gradle.kts
rotty3000: Exactly, yet only when resolving (an assembly) using an OSGi resolver. A regular classpath application will ignore all of this. Furthermore, even an OSGi runtime resolver will ignore this since the requirement is not enforceable except during assembly.

junit-jupiter-api/junit-jupiter-api.gradle.kts
rotty3000: Further, this is specially define in your environment to enable the ability to tweak the OSGiVerification task without polluting any other aspects of the project.

marcphilipp (approved):

rotty3000:
@marcphilipp forgive me if I don't recall where to go from here. Are we just in holding until some milestone or am I required to perform an action?

marcphilipp (changes requested):
Sorry for the delay! I asked the team to take a look but everyone is busy currently. Could you please add an entry to the 5.8.0-M2 release notes?

marcphilipp:
I'm glad to hear that! 😊

PR #2667, opened 2021-07-20
Added before and after all classes callback

SergKhram:
Now we can add extensions for executing before all test classes and after all test classes
https://github.com/junit-team/junit5/issues/456


alexp1917:
why was this closed?

Issue #864, opened 2017-05-25
Support declarative extension registration on fields and parameters


nipafx:
## Overview

Original title: "Support ParameterResolver registration via @ResolveWith on parameter"

While experimenting with parameter resolvers I realized that they can not be applied like argument converters can, i.e. by annotating the parameter directly. This complicates their application and implementation if they are "annotation based".

An example is the `@Mock` proof of concept. The resolver has to be applied to the test method or class and because of that has to check the individual parameter for the presence of the `@Mock` annotation. This indirection causes unnecessary complexity for implementers and users of the extension.

## Original Proposal

I propose an annotation similar to `@ConvertWith`, e.g. `@ResolveWith`, that takes precedence over all registered parameter resolvers. It does not need a `supports` method because it is assumed to support the parameters it is applied to. If it turns out during `resolve` that that's not the case, the test should be aborted by an exception.

With this mechanism, `@Mock` (and I am planning to write something like `@Random`) would get simpler to use because they only need to be applied in one place.

More broadly, I think parameter resolvers and argument converters are conceptually very similar and the API would hence improve considerably if both would operate similarly as well. In that regard I already opened #853, which proposes an extension point to register converters at the method and class level. Together with this issue, the two aim to align both mechanisms.

If there is interest in this, I would give it a shot, preferably together with #853.

## Related Issues

- #416
- #497

sormuras:
Like?
```
  @RepeatedTest(10)
  void testSomeGeneratedSudokus(@Random int seed) {
    assertTrue(Sudoku.generate(seed).isSolvable());
  }
```

nicolaiparlog:
Yes. But without registering any parameter resolver on the class level and without the chance of `@Random` conflicting with them.

sbrannen:
Related discussion on Twitter about being able to register extension on parameters:

https://twitter.com/olivergierke/status/1022802925603758080

sbrannen:
Renamed issue to align with the title of #497 and features discussed in the aforementioned Twitter thread.

sbrannen:
Updated scope to include registration via fields _and_ parameters.

sbrannen:
Current work on this issue can be viewed in the following branch:

https://github.com/junit-team/junit5/compare/issues/864-declarative-extension-registration-params-fields

sbrannen:
_in progress_

sbrannen:
[...]

Issue #2651, opened 2021-06-21
Support ordering between @ExtendWith and @RegisterExtension


oehme:
I was trying to write a [Micronaut test](https://micronaut-projects.github.io/micronaut-test/latest/guide/#junit5) using the `@MicronautTest` annotation. This annotaiton is a meta-annotation that includes `@ExtendWith(MicronautJunit5Extension)`. However, I also needed to start up some other services with random port numbers and add those ports to Micronaut's configuration. I tried doing so with `@Order(1) @RegisterExtension MyOtherService service`, but that didn't work. The MicronautExtension is still called first and fails because calling `getPort` on my service can't be done until it has started.

It would be great if `@Order` worked together with `@ExtendWith`, so that I could push my service in front of Micronaut's startup.

Workaround:

Don't use `MicronautTest` with all its convenience and instead use `ApplicationContext.run(EmbeddedServer.class, properties)`.

## Deliverables

- [ ] ...

oehme:
The same issue has been reported previously for [MicronautTest in Spock](https://github.com/micronaut-projects/micronaut-test/issues/80) and got closed as "not a Micronaut problem", which is why I'm looking for a solution in JUnit instead.

marcphilipp:
Related to #506.

sbrannen:
Related to #2680

PR #2680, opened 2021-08-03
Support declarative extension registration on fields and parameters

sbrannen:
## Overview

This is a PR for #864.

The implementation is essentially feature complete and tested (following the 80/20 rule); however, there are some points that should be discussed before merging this PR into the `main` branch.

## Topics for Discussion

When discussing any of these topics, I recommend that you review the current behavior by analyzing the tests in `ExtensionRegistrationViaParametersAndFieldsTests` before spending time studying the actual implementation details.

1. **Registration Order**: I have done my best to register extensions in the order in which the fields or parameters are used in the lifecycle of a test class and test method. This can be inspected in the `registrationOrder()` test method. Implementation details can be found in `ClassBasedTestDescriptor` and `TestMethodTestDescriptor`.
2. **`@Order` Support**: `ExtensionUtils.registerExtensionsFromFields()` already supported `@Order` for extensions registered via `@RegisterExtension`, and `@Order` is now also applied to extensions registered via fields using `@ExtendWith`. But... `@ExtendWith` extensions are registered _before_ `@RegisterExtension` extensions.
3. **`@ExtendWith` target**: `@ExtendWith` can currently be declared on a `TYPE` or `METHOD`. It cannot be declared directly on a `PARAMETER` or `FIELD`.
4. **Private fields**: `@RegisterExtension` fields are not allowed to be `private`, but in this PR fields meta-annotated with `@ExtendWith` are allowed to be `private`.

I will share my thoughts on these topics as comments and invite feedback.

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M2.adoc
@@ -56,6 +56,9 @@ on GitHub.
 * Generating Java Flight Recorder events via module `org.junit.platform.jfr` is now also
   supported on Java 8 Update 262 or higher, in addition to Java 11 or later. See
   <<../user-guide/index.adoc#running-tests, Flight Recorder Support>> for details.
+* Suites executed by the `junit-platform-suite` module will now inherit the configuration
+  parameters from the parent discovery request. This behavior can be disabled via the
+  `@DisableParentConfigurationParameters` annotation.
 
 
 [[release-notes-5.8.0-M2-junit-jupiter]]
@@ -82,6 +85,9 @@ on GitHub.
 
 ==== New Features and Improvements
 
+* New `assertThrowsExactly()` method in `Assertions` which is a more strict version of
+  `assertThrows()` that allows you to assert that the exception thrown is of the exact
+  type specified.
 * `@TempDir` can now be used to create multiple temporary directories. Instead of creating
   a single temporary directory per context (i.e. test class or method) every declaration
   of the `@TempDir` annotation on a field or method parameter now results in a separate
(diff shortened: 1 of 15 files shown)


sbrannen:
> **Private fields**: `@RegisterExtension` fields are not allowed to be `private`, but in this PR fields meta-annotated with `@ExtendWith` are allowed to be `private`.

From one perspective, this is inconsistent treatment of fields; however, I wonder if we haven't been too strict with regard to the current non-private policy for `@RegisterExtension` fields.

If JUnit or a third party provides an extension that is implicitly registered by a custom composed annotation (which is meta-annotated with `@ExtendWith`), then it is really up to the user of that composed annotation to decide if the annotated field is `private`.

As long as the extension can work with a `private` field, there should be no unnecessary constraint on the visibility of the field.

In other words, why should JUnit not allow the user to decide if a field is `private`?

I recommend that we allow `@RegisterExtension` and `@ExtendWith` fields to be `private`.

sbrannen:
> **`@ExtendWith` target**: `@ExtendWith` can currently be declared on a `TYPE` or `METHOD`. It cannot be declared directly on a `PARAMETER` or `FIELD`.

Although the original intention of #864 was to allow extension authors to provide custom composed annotations that are meta-annotated with `@ExtendWith`, after having spent more time with the issue I think there are likely valid use cases for allowing `@ExtendWith` to be declared directly on a field or parameter.

In other words, I don't believe we should force users or library implementers to create custom composed annotations when all they need is `@ExtendWith` on a field or parameter.

sbrannen:
> **`@Order` Support**: `ExtensionUtils.registerExtensionsFromFields()` already supported `@Order` for extensions registered via `@RegisterExtension`, and `@Order` is now also applied to extensions registered via fields using `@ExtendWith`. But... `@ExtendWith` extensions are registered _before_ `@RegisterExtension` extensions.

The above describes the status quo in this PR, but we can certainly change the semantics.

My rationale for implementing the feature with these semantics was based on the fact that extensions registered via `@ExtendWith` on a class or interface are always registered before any extensions registered via `@RegisterExtension`. This is the current behavior in JUnit Jupiter, and that default behavior will not change.

So, the status quo in this PR is consistent in that regard.

However, we have an opportunity (for the first time) to make it possible for extensions registered via `@ExtendWith` to be ordered by the user **relative** to extensions registered via `@RegisterExtension`.

Achieving that is straight forward: I'd just need to collapse the two iterations in `ExtensionUtils.registerExtensionsFromFields()` into a single iteration over the fields.

After having spent more time with this issue, I am in favor of allowing `@RegisterExtension` and `@ExtendWith` fields to be ordered relative to each other, and I imagine the community would likely expect all such fields to be sorted together when using `@Order`.

sbrannen:
> **Registration Order**: I have done my best to register extensions in the order in which the fields or parameters are used in the lifecycle of a test class and test method. This can be inspected in the `registrationOrder()` test method. Implementation details can be found in `ClassBasedTestDescriptor` and `TestMethodTestDescriptor`.

Aside from the discussion about `@Order` support, I feel pretty comfortable with the current extension registration order, but feedback is certainly welcome!

sbrannen:
> However, we have an opportunity (for the first time) to make it possible for extensions registered via `@ExtendWith` to be ordered by the user **relative** to extensions registered via `@RegisterExtension`.

Related issues:

- #506 
- #2651

sbrannen:
**Team Decision**:

- Allow `@RegisterExtension` and `@ExtendWith` fields to be `private`.
- Allow `@ExtendWith` to be declared directly on fields and parameters.
- Allow `@RegisterExtension` and `@ExtendWith` fields to be sorted via `@Order` relative to each other.
- Throw an exception if a field is annotated with both `@RegisterExtension` and `@ExtendWith`.

marcphilipp (approved):

sbrannen:
Closed via commit 8b5387c5bef050730d8feb52a5292a32530f65cd

PR #2681, opened 2021-08-06
Change `@TempDir` to create a separate dir per declaration

### LINKED ISSUE #1967: Multiple @TempDirs should result in multiple directories

marcphilipp:
Resolves #1967.

---

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/JupiterConfiguration.java
```diff
@@ -41,6 +41,7 @@
 	String DEFAULT_DISPLAY_NAME_GENERATOR_PROPERTY_NAME = "junit.jupiter.displayname.generator.default";
 	String DEFAULT_TEST_METHOD_ORDER_PROPERTY_NAME = "junit.jupiter.testmethod.order.default";
 	String DEFAULT_TEST_CLASS_ORDER_PROPERTY_NAME = "junit.jupiter.testclass.order.default";
+	String TEMP_DIR_BEHAVIOR_PROPERTY_NAME = "junit.jupiter.tempdir.behavior";
```
marcphilipp, line 44:
```suggestion
	String TEMP_DIR_BEHAVIOR_PROPERTY_NAME = "junit.jupiter.tempdir.scope";
```

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryPerDeclarationTests.java
```diff
@@ -0,0 +1,981 @@
...
+	@Test
+	@DisplayName("can be used via static field inside nested test classes")
+	void canBeUsedViaStaticFieldInsideNestedTestClasses() {
+		executeTestsForClass(StaticTempDirUsageInsideNestedClassTestCase.class).testEvents()//
+				.assertStatistics(stats -> stats.started(2).succeeded(2));
+	}
+
+	@Nested
+	@DisplayName("reports failure")
+	@TestMethodOrder(OrderAnnotation.class)
```
juliette-derancourt, line 191: Why the ordering here? Just to display related tests together, or is there another reason?

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryPerDeclarationTests.java
```diff
@@ -0,0 +1,981 @@
...
+			assertNotNull(tempDir);
+			assertNotNull(ref);
+			assertNotNull(this.fileTempDir);
+			assertNotNull(this.pathTempDir);
+		}
+
+	}
+
+	// https://github.com/junit-team/junit5/issues/1748
+	static class TempDirectoryDoesNotPreventConstructorParameterResolutionTestCase {
```
juliette-derancourt, line 401: Isn't this one actually missing a `@TempDir` field or a `@ExtendWith(TempDirectory.class)` annotation?

juliette-derancourt (approved):
I know most of them were already in place but... these test cases are _thorough_! 👌

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryPerDeclarationTests.java
marcphilipp: @sbrannen introduced this originally in https://github.com/junit-team/junit5/commit/ab1211121927cba4837444bca35983331cd59e71.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryPerDeclarationTests.java
marcphilipp: It worked anyway since the extension is registered globally but I think it's clearer if it actually used a `@TempDir` and added one in e79d45a65e60dfe6e4c089bdc26424d261942d58.

sormuras (approved):

PR #2684, opened 2021-08-11
Inline lambdas passed to assertDoesNotThrows to support suspending functions

### LINKED ISSUE #2674: Support assertDoesNotThrow() with suspending functions in Kotlin
Support for suspending functions was added for the `assertThrows()` counterpart (see #1851), but it seems like calling suspending functions in an `assertDoesNotThrow()` block is currently not supported.

## Deliverables

- [ ] Add support for calling suspending functions for all variants of `assertDoesNotThrow()`.

williamboman:
Resolves #2674.

## Overview

Support for suspending functions was added for the assertThrows() counterpart (see #1851), but it seems like calling suspending functions in an assertDoesNotThrow() block is currently not supported.

---

---

### Definition of Done


junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -158,8 +158,12 @@ inline fun <reified T : Throwable> assertThrows(noinline message: () -> String,
...
  */
 @API(status = EXPERIMENTAL, since = "5.5")
-fun <R> assertDoesNotThrow(executable: () -> R): R =
-    Assertions.assertDoesNotThrow(ThrowingSupplier(executable))
+inline fun <reified R> assertDoesNotThrow(executable: () -> R): R = try {
+    val result = executable()
+    Assertions.assertDoesNotThrow(ThrowingSupplier { result })
+} catch (throwable: Throwable) {
+    Assertions.assertDoesNotThrow(ThrowingSupplier { throw throwable })
+}
```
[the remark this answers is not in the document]
williamboman: This might be neater?

```diff
diff --git a/junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt b/junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
index cf420e7f3f..abe4e23769 100644
--- a/junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
+++ b/junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
@@ -158,11 +158,19 @@ inline fun <reified T : Throwable> assertThrows(noinline message: () -> String,
  * @param R the result type of the [executable]
  */
 @API(status = EXPERIMENTAL, since = "5.5")
-inline fun <reified R> assertDoesNotThrow(executable: () -> R): R = try {
-    val result = executable()
-    Assertions.assertDoesNotThrow(ThrowingSupplier { result })
-} catch (throwable: Throwable) {
-    Assertions.assertDoesNotThrow(ThrowingSupplier { throw throwable })
+inline fun <reified R> assertDoesNotThrow(executable: () -> R): R {
+    val (result, throwable) = try {
+        Pair(executable(), null)
+    } catch (throwable: Throwable) {
+        Pair(null as R, throwable)
+    }
+
+    return Assertions.assertDoesNotThrow(ThrowingSupplier {
+        if (throwable != null) {
+            throw throwable
+        }
+        result
+    })
 }
 
 /**
@@ -190,15 +198,20 @@ inline fun <reified R> assertDoesNotThrow(message: String, executable: () -> R):
  * @param R the result type of the [executable]
  */
 @API(status = EXPERIMENTAL, since = "5.5")
-inline fun <reified R> assertDoesNotThrow(noinline message: () -> String, executable: () -> R): R = try {
-    val result = executable()
-    Assertions.assertDoesNotThrow(
-        ThrowingSupplier { result },
-        Supplier(message)
-    )
-} catch (throwable: Throwable) {
-    Assertions.assertDoesNotThrow(
-        ThrowingSupplier { throw throwable },
+inline fun <reified R> assertDoesNotThrow(noinline message: () -> String, executable: () -> R): R {
+    val (result, throwable) = try {
+        Pair(executable(), null)
+    } catch (throwable: Throwable) {
+        Pair(null as R, throwable)
+    }
+
+    return Assertions.assertDoesNotThrow(
+        ThrowingSupplier {
+            if (throwable != null) {
+                throw throwable
+            }
+            result
+        },
         Supplier(message)
     )
 }
```

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
```diff
@@ -96,32 +96,26 @@ class KotlinAssertionsTests {
         dynamicContainer("fails when an exception is thrown", Stream.of(
             dynamicTest("for no arguments variant") {
                 val exception = assertThrows<AssertionError> {
-                    assertDoesNotThrow {
+                    assertDoesNotThrow<Unit> {
```
williamboman, line 99: Oh and, as-is, I believe this is a breaking change. Reified type parameters does not play nicely with `Nothing`

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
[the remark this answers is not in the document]
juliette-derancourt: I agree that the current implementation is not ideal because it calls `assertDoesNotThrow` twice, but I still find it more readable... might be a matter of personal preference though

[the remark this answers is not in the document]
juliette-derancourt: @williamboman Would something like this make sense?

```kotlin
inline fun <reified R> assertDoesNotThrow(executable: () -> R): R {
    Assertions.assertDoesNotThrow(getThrowingSupplier { executable })
}

inline fun <reified R> assertDoesNotThrow(noinline message: () -> String, executable: () -> R): R {
    Assertions.assertDoesNotThrow(
        getThrowingSupplier { executable },
        Supplier(message)
    )
}

inline fun <reified R> getThrowingSupplier(executable: () -> R): ThrowingSupplier<R> = try {
    val result = executable()
    ThrowingSupplier { result }
} catch (throwable: Throwable) {
    ThrowingSupplier { throw throwable }
}
```

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
juliette-derancourt: Does this mean current usages of `assertDoesNotThrow` will break?

williamboman: > Does this mean current usages of `assertDoesNotThrow` will break?

Only the ones where the provided `executable` returns a type that is not [compatible with reified type parameters](https://github.com/JetBrains/kotlin/blob/master/spec-docs/reified-type-parameters.md). Currently I believe is only the fictitious `Nothing` type.

> Could this not break if we "artificially" change the return type to Unit? (genuine question, I'm not much of a Kotlin master)

Do you mean by overloading the functions with a variant that takes `() -> Nothing` and returns `Unit`? Should work, I think :D! I'll give it a try!

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -147,6 +147,14 @@ inline fun <reified T : Throwable> assertThrows(noinline message: () -> String,
     }, Supplier(message))
 }
 
+@PublishedApi
```
williamboman, line 150: https://stackoverflow.com/a/41905907

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
juliette-derancourt: I just read a bit more about `reified` and now I'm confused... why do we even need it, since we don't manipulate the `R` type? 🤔

juliette-derancourt: Oh, is it to keep the type of the returned object?

williamboman: You're 100% right, in this case having a reified type does not make any sense. I was just looking at the `assertThrows` signature and blindly copied it.

williamboman: > Do you mean by overloading the functions with a variant that takes `() -> Nothing` and returns `Unit`? Should work, I think :D! I'll give it a try!

To answer this, yes seems possible in later versions (seems to still be experimental):
- https://discuss.kotlinlang.org/t/overloaded-function-with-lambda-parameter-of-different-return-types/6053/4
- https://youtrack.jetbrains.com/issue/KT-11265

williamboman: I removed the reified type parameters in https://github.com/junit-team/junit5/pull/2684/commits/80aeb8da903a46074e756f5c5c23c97b7c2d4e05

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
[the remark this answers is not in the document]
williamboman: Yep I was imagining something similar ! Added in https://github.com/junit-team/junit5/pull/2684/commits/14407dd63b71b4b3b38b62da37ae548845bcda39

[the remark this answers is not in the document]
juliette-derancourt: Looks good! 👌

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
juliette-derancourt: Ahah, glad I could help! (Can I call myself a kotlin expert now? 😎)

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
juliette-derancourt: Again, possibly dumb question here, but why do we need this method to be `inline`?

juliette-derancourt: Also, could you please move this method below the other ones calling it? :)

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
```diff
@@ -62,12 +62,21 @@ class KotlinAssertionsTests {
...
+    fun `assertThrows with coroutines`() = runBlocking<Unit> {
         assertThrows<AssertionError>("Should fail async") {
             suspend { fail("Should fail async") }()
         }
     }
 
+    @Test
+    fun `assertDoesNotThrow with coroutines`() = runBlocking {
+        val result = assertDoesNotThrow {
+            suspend { "Result" }()
```
juliette-derancourt, line 74: We might want to have another test with a suspending function that actually throws an exception (thus making the test fail), since we want to make sure that the assertion is actually waiting for the "throwing"

juliette-derancourt: Just noticed the test factory below, you could even add one dynamic test in each `dynamicContainer` for this new suspending function use case :)

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
williamboman: > Again, possibly dumb question here, but why do we need this method to be `inline`?

Because this function is ultimately the one responsible for calling the provided `executable`. We want to run the `executable` in the same context as the callsite of `assertDoesNotThrow` (so that it can suspend, etc.), so we need to make sure the full call stack from `assertDoesNotThrow` until here uses inlined functions. If we were to remove the `inline` modifier on this function, we'll get a compiler error in all of the `assertDoesNotThrow` functions saying that we're violating the inlineability <sup>(I totally just made that word up)</sup> of the `executable` argument, and that we need to change the function signature to `assertDoesNotThrow(noinline executable: () -> R)`. Doing so breaks the tests that now calls a suspending function, because the callback is no longer inlined.

junit-jupiter-engine/src/test/kotlin/org/junit/jupiter/api/KotlinAssertionsTests.kt
williamboman: Done! https://github.com/junit-team/junit5/pull/2684/commits/d48eeb4afb14064648247f9f271ed4864b10285c

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
juliette-derancourt: I see! Thanks for the explanation 😄

juliette-derancourt (approved):

juliette-derancourt:
@williamboman could you please add an entry in the release notes?

williamboman:
> @williamboman could you please add an entry in the release notes?

Done! Struggled a bit with the wording (for example, not sure if need to clarify it's Kotlin API or if that's implied), feel free to amend!

documentation/src/docs/asciidoc/release-notes/release-notes-5.8.0-M2.adoc
```diff
@@ -82,6 +82,7 @@ on GitHub.
 
 ==== New Features and Improvements
 
+* `assertDoesNotThrow` now supports calling suspending functions.
```
marcphilipp, line 85:
```suggestion
* `assertDoesNotThrow` now supports suspending functions when called from Kotlin.
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -186,8 +186,19 @@ fun <R> assertDoesNotThrow(message: String, executable: () -> R): R =
...
-fun <R> assertDoesNotThrow(message: () -> String, executable: () -> R): R =
-    Assertions.assertDoesNotThrow(ThrowingSupplier(executable), Supplier(message))
+inline fun <R> assertDoesNotThrow(noinline message: () -> String, executable: () -> R): R =
+    Assertions.assertDoesNotThrow(
+        getThrowingSupplier(executable),
+        Supplier(message)
+    )
+
+@PublishedApi
+internal inline fun <R> getThrowingSupplier(executable: () -> R): ThrowingSupplier<R> = try {
```
marcphilipp, line 196: Not sure about this method name which is "effectively public" due to `@PublishedApi`. Maybe `evaluateAndWrap` or sth. similar would be clearer.

marcphilipp (approved):

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
williamboman: Yeah that's a good call, agreed. 

https://github.com/junit-team/junit5/pull/2684/commits/742d7af3258469c30e6a997d5c372d436cf7a762

PR #2707, opened 2021-09-04
Provide access to the current ExecutionMode in ExtensionContext

### LINKED ISSUE #2396: Provide information about the current ExecutionMode in ExtensionContext
I am contributing to a bunch of JUnit 5 extensions that rely on a form of field injection to achieve their goals.

Field injection, if done in a `@BeforeEach` callback, works well for both parallel and sequential testing. It also works well for both `PER_CLASS` and `PER_METHOD` lifecycle semantics. However, it does not work very well if you're combining parallel execution with `PER_CLASS` lifecycle semantics, as the parallel execution threads for each test method will each inject their own instances into the single test class instance and clobber each other, interfering with the tests running in other threads. In such situations, it is necessary to inject into a `ThreadLocal` container instead.

Accordingly, I would like my extensions to be able to give fail-fast-and-loud behaviour in these circumstances - throwing an `ExtensionConfigurationException` if it finds it is being asked to run in an environment where it is being asked to run `CONCURRENT` and `PER_CLASS` at the same time, and informing the user that they should use a `ThreadLocal` in such a case.

The problem: the `ExtensionContext` that is passed to the extension's lifecycle methods has a method for telling you which class instance lifecycle mode is in force. However, it doesn't seem to give you a way to easily determine the execution mode. This can be inferred by manually examining the annotations and the global execution mode, but I feel that this breaks encapsulation a little bit - I don't have to do the same for the lifecycle.

## Deliverables

Method `ExecutionMode getExecutionMode()` in `ExtensionContext`.

Possibly a pair of methods - one that returns the value annotated on the class, and one that returns the *actual* execution mode in force (taking into account any global setting). The latter would be more directly useful to my use case.

juliette-derancourt:
## Overview

Resolves #2396.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -390,9 +390,9 @@ public interface ExtensionContext {
 	 * @return the {@code ExecutionMode} of the test; never {@code null}
 	 *
 	 * @see org.junit.jupiter.api.parallel.ExecutionMode {@code @ExecutionMode}
-	 * @since 5.8
+	 * @since 5.8.1
 	 */
-	@API(status = STABLE, since = "5.8")
+	@API(status = STABLE, since = "5.8.1")
 	ExecutionMode getExecutionMode();
 
 	/**
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
@@ -26,6 +26,7 @@ import java.util.function.Function;
 
 import org.apiguardian.api.API;
 import org.junit.jupiter.api.TestInstance.Lifecycle;
+import org.junit.jupiter.api.parallel.ExecutionMode;
 import org.junit.platform.commons.PreconditionViolationException;
 import org.junit.platform.commons.support.ReflectionSupport;
 import org.junit.platform.commons.util.Preconditions;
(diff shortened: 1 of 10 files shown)


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.java
```diff
@@ -127,4 +130,21 @@ public Store getStore(Namespace namespace) {
 	public <V> Optional<V> getConfigurationParameter(String key, Function<String, V> transformer) {
 		return this.configuration.getRawConfigurationParameter(key, transformer);
 	}
+
+	@Override
+	public ExecutionMode getExecutionMode() {
+		return toJupiterExecutionMode(getPlatformExecutionMode());
+	}
+
+	protected abstract Node.ExecutionMode getPlatformExecutionMode();
```
juliette-derancourt, line 139: I'm not sure how to call this one...

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.java
```diff
@@ -127,4 +130,21 @@ public Store getStore(Namespace namespace) {
...
+
+	private ExecutionMode toJupiterExecutionMode(Node.ExecutionMode mode) {
```
juliette-derancourt, line 141: It's the reverse of this: https://github.com/junit-team/junit5/blob/main/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java#L171

Should we perhaps move both somewhere, or leave it here?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -383,6 +384,17 @@ default void publishReportEntry(String value) {
...
 
+	/**
+	 * Get the {@link ExecutionMode} associated with the current test or container.
+	 *
+	 * @return the {@code ExecutionMode} of the test; never {@code null}
+	 *
+	 * @see org.junit.jupiter.api.parallel.ExecutionMode {@code @ExecutionMode}
+	 * @since 5.8
+	 */
+	@API(status = STABLE, since = "5.8")
```
marcphilipp, line 395:
```suggestion
	 * @since 5.8.1
	 */
	@API(status = STABLE, since = "5.8.1")
```

marcphilipp (approved):

PR #2721, opened 2021-09-17
Introduce support for text blocks in `@CsvSource`

sbrannen:
## Overview

With the introduction of support for _text blocks_ as a first-class language feature in recent JDKs (preview feature in Java SE 15), we can improve the user experience with `@CsvSource` by allowing the user to provide a text block instead of an array of strings.

## Example

Given the following parameterized test using a text block...

```java
@ParameterizedTest
@CsvSource(textBlock = """
    apple,         1
    banana,        2
    'lemon, lime', 0xF1
    strawberry,    700_000
""")
void csvSourceWithTextBlock(String fruit, int rank) {
	System.out.println(fruit + " : " + rank);
}
```
... the output is:

```
apple : 1
banana : 2
lemon, lime : 241
strawberry : 700000
```

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
@@ -17,6 +17,7 @@ import java.lang.annotation.Annotation;
 import java.util.Arrays;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import com.univocity.parsers.csv.CsvParser;
@@ -31,6 +32,8 @@ import org.junit.platform.commons.util.UnrecoverableExceptions;
  * @since 5.0
  */
 class CsvArgumentsProvider implements ArgumentsProvider, AnnotationConsumer<CsvSource> {
+
+	private static final Pattern NEW_LINE_REGEX = Pattern.compile("\\n");
 
 	private static final String LINE_SEPARATOR = "\n";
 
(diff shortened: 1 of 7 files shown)


sbrannen:
This feature is complete and ready for review.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -47,9 +47,20 @@ public void accept(CsvSource annotation) {
 
 	@Override
 	public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+		Preconditions.condition(this.annotation.value().length > 0 ^ !this.annotation.textBlock().isEmpty(),
+			() -> "@CsvSource must be declared with either `value` or `textBlock` but not both");
+
+		String[] lines;
+		if (!this.annotation.textBlock().isEmpty()) {
+			lines = this.annotation.textBlock().split("\\n");
```
marcphilipp, line 55: We could use `String.lines()` if we're running on JDK 11 here. Not sure it's worth it, though.

marcphilipp (approved):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
sbrannen: Yeah, I actually used `.lines()` first and then realized it wouldn't work on JDK 8. So I reverted that change.

I know we use reflection like that in `JRE.determineCurrentVersion()` to use `java.lang.Runtime.Version`, but I don't think this particular use case warrants the use of reflection just to use `String.lines()`.

We can, however, precompile the regular expression for a slight performance boost. I'll go ahead and do that.

If people complain that they notice a considerable performance hit when using `textBlock`, we can always implement an improved algorithm at a later date.

sbrannen: > We can, however, precompile the regular expression for a slight performance boost. I'll go ahead and do that.

I just force pushed that change.

PR #2728, opened 2021-09-24
Use JDK 17 toolchain for compilation and tests

marcphilipp:
Please refer to individual commits.


sormuras (approved):
#BrewYourBeansOnJDK17

sbrannen:
I tried out your branch locally and encountered build failures.

https://ge.junit.org/s/2ccivq5ukhjt2

sbrannen:
> I tried out your branch locally and encountered build failures.

The build failures go away when running the Gradle build with JDK 11. 👍

sbrannen (approved):
I didn't review the changes for the CI pipelines, but I verified that the build works locally with JDK 11 at the command line and with JDK 17 in Eclipse IDE 2021-09 and the Java 17 patch from the Eclipse Marketplace.

So it looks good to me.

...except that the Groovy and Kotlin plugins do not yet work with Eclipse IDE 2021-09 and the Java 17 patch, as far as I can tell. But that's not a show stopper.

Issue #2734, opened 2021-10-01
Support text blocks as files in `@CsvSource`


sbrannen:
## Overview

PR #2721 introduced experimental support for _text blocks_ in `@CsvSource`; however, there is room for improvement.

A CSV line within a text block cannot currently contain a new line (`\n`), even if it is within a quoted string; whereas, this is supported when using `@CsvSource(value = ...)`.

Comments do not make sense in a single string in `@CsvSource(value = ...)`, but they do make sense within `@CsvSource(textBlock = ...)`.

## Deliverables

- [ ] Support new lines within "CSV lines" in text blocks
- [ ] Support comment lines beginning with `#` in text blocks

sbrannen:
Current work on this issue can be viewed in the following feature branch:

https://github.com/junit-team/junit5/compare/main...sbrannen:issues/2734-treat-text-blocks-as-files-in-CsvSource

sbrannen:
FYI: PR #2737 is ready for technical review.

PR #2737, opened 2021-10-03
Treat text blocks as files in `@CsvSource`

sbrannen:
## Overview

This PR for #2734 is feature complete in terms of implementation and ready for review.

Documentation updates will follow, once this PR has been approved based on technical merit.

One topic for discussion is the use of reflection to invoke `java.lang.String#trimIndent()` on Java 15+. An alternative proposed by @sormuras would be to make use of the MR JAR format (multi-release JAR) as was done with `ModuleUtils`. If the team decides to go that route, we could introduce a new `TextBlockUtils` class in a `java15` source folder in `junit-platform-commons` that uses `java.lang.String#trimIndent()`, with a simplified version of `TextBlockUtils` in the standard source folder that does not actually trim indentation.

### Definition of Done


sbrannen:
Let me point out the implications of not always being able to invoke `String#trimIndent()`.

- `trimIndent()` is available starting with Java 15, just like text blocks (in Java). So anyone using Java text blocks will always have `trimIndent()` support.
- Anyone using multi-line strings in Groovy or Kotlin should be able to use the `textBlock` support in `@CsvSource` as well. If they do that using a JRE prior to Java 15, `trimIndent()` will not be applied. If they do that using a JRE 15+, they'll get the full feature set just like with Java 15+.

What this means is that a Groovy or Kotlin user who's not using JRE 15+ will face some restrictions for how they format their multi-line strings.

For example, this (pseudo-code):

```java
@CsvSource(quoteCharacter = '"', textBlock = """
			# This is a comment
			apple,         1
			banana,        2
			"lemon
			lime", 0xF1
			strawberry,    700_000
		""")
```

... may have to be replaced with the following:

```java
@CsvSource(quoteCharacter = '"', textBlock = """
# This is a comment
apple,         1
banana,        2
"lemon
lime", 0xF1
strawberry,    700_000
""")
```

... depending on whether they use `#` comment lines (which must not have any whitespace before the `#` character) or new lines `\n` within quoted text.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -54,47 +72,86 @@ public void accept(CsvSource annotation) {
...
 			throw (PreconditionViolationException) throwable;
 		}
 		throw new CsvParsingException("Failed to parse CSV input configured via " + annotation, throwable);
 	}
 
+	private static String sanitizeTextBlock(String textBlock) {
+		// Split and remove trailing blank lines.
+		String[] lines = NEW_LINE_REGEX.split(textBlock, 0);
+		// Reassemble the lines.
+		textBlock = String.join(LINE_SEPARATOR, lines);
```
sbrannen, line 145:
```suggestion
		// Strip all trailing whitespace at the end of the text block.
		textBlock = textBlock.replaceAll("\\s+?$", "");
```

FWIW, I considered this approach, and the tests still pass, but I'm not convinced that it's as robust as the implementation in this PR.

Thoughts?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -54,22 +69,35 @@ public void accept(CsvSource annotation) {
...
+			return parseTextBlock(this.annotation.textBlock()).stream().map(Arguments::of);
 		}
 
-		AtomicInteger index = new AtomicInteger(1);
+		AtomicInteger index = new AtomicInteger(0);
 		// @formatter:off
-		return Arrays.stream(lines)
-				.map(line -> parseLine(line, index.getAndIncrement()))
+		return Arrays.stream(this.annotation.value())
+				.map(line -> parseLine(line, index.incrementAndGet()))
```
juliette-derancourt, line 79: Just a question, what's the reason for switching to `incrementAndGet()`?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -26,17 +28,30 @@
...
  */
 class CsvArgumentsProvider implements ArgumentsProvider, AnnotationConsumer<CsvSource> {
 
-	private static final Pattern NEW_LINE_REGEX = Pattern.compile("\\n");
-
 	private static final String LINE_SEPARATOR = "\n";
 
+	private static final Method stripIndentMethod;
+
+	static {
```
juliette-derancourt, line 43: Not a necessity, but maybe this block and `stripIndent(String textBlock)` could be moved in a different file, for the sake of readability?

juliette-derancourt (approved):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
sbrannen: The only reason was to be consistent with the use of the `AtomicInteger` in `parseTextBlock()`.

Otherwise, there's no overwhelming reason.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
sbrannen: Good point. I'll go ahead and create a local version of the previously discussed `TextBlockUtils` class and move those there.

sbrannen: done ✅

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -54,22 +54,35 @@ public void accept(CsvSource annotation) {
...
+		return Arrays.stream(this.annotation.value())
+				.map(line -> parseLine(line, index.incrementAndGet()))
 				.map(Arguments::of);
 		// @formatter:on
 	}
 
+	private List<String[]> parseTextBlock(String textBlock) {
+		try {
+			AtomicInteger index = new AtomicInteger(0);
+			List<String[]> csvRecords = this.csvParser.parseAll(new StringReader(stripIndent(textBlock)));
```
marcphilipp, line 72: Is there an API that returns an iterator or sth. like that? It's not a big concern since these strings shouldn't be huge, though.

marcphilipp (approved):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
sbrannen: > Is there an API that returns an iterator or sth. like that?

There's `parse(Reader)` which delegates to a `RowProcessor`. The other option is to use `beginParsing(Reader)` and `parseNext()` (like in `CsvFileArgumentsProvider`).

So, we could use `beginParsing(new StringReader(stripIndent(textBlock))` and then iterate the same as in `CsvFileArgumentsProvider` with our `CsvParserIterator` (or investigate how to do that with a `RowProcessor`), but I think that's overkill since the entire text block string is already in memory.

> It's not a big concern since these strings shouldn't be huge, though.

Yes, that was my rationale for using `parseAll(Reader)`.

sormuras (approved):
LGTM

PR #2769, opened 2021-11-02
Use value references in Javadoc to eliminate duplication

marcphilipp:
Triggered by discussion in fd93eeca94479b71ce0d06d61df874a167a3fbf7.

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/ClassOrderer.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/ClassOrderer.java
@@ -12,6 +12,7 @@ package org.junit.jupiter.api;
 
 import static java.util.Comparator.comparingInt;
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+import static org.apiguardian.api.API.Status.STABLE;
 
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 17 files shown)


sbrannen:
Oh my.... this PR is a little bigger than I anticipated. 😲

But at a quick glance, I think it looks good: big improvement.

I'll try to find the time in the coming days to review it in detail.

marcphilipp (commented):
TODO: Add `@API` annotations and Javadoc for all properties. 😵

sbrannen:
not stale, just needs to be reviewed (or merged)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/ClassOrderer.java
```diff
@@ -61,6 +62,22 @@
...
+	 *
+	 * <h4>Supported Values</h4>
+	 *
+	 * <p>Supported values include fully qualified class names for types that
+	 * implement {@link org.junit.jupiter.api.ClassOrderer}.
+	 *
+	 * <p>If not specified, test classes are not ordered unless test classes are
+	 * annotated with {@link TestClassOrder @TestClassOrder}.
+	 *
+	 * @since 5.8
```
KostyaSha, line 76: `since 5.9`

sbrannen: `@since` denotes when the feature was introduced.

`@API(since = ...)` denotes "since" when the feature was assigned the API _status_.

sbrannen: So, it's correct as-is: introduced in 5.8 and stable since 5.9.

PR #2782, opened 2021-11-16
Add an example of inner class fully qualified method name

VirtualTim:
## Overview

This took me a while to figure out the syntax, so I thought improving the documentation could be beneficial.
As an aside, there are also no tests for `MethodSource` with an inner class.

---

---

Doco-only change, so DoD is not relevant.

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodSource.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodSource.java
@@ -109,7 +109,7 @@ public @interface MethodSource {
 	 *
 	 * <p>Factory methods in external classes must be referenced by <em>fully
 	 * qualified method name</em> &mdash; for example,
-	 * {@code com.example.StringsProviders#blankStrings}.
+	 * {@code com.example.StringsProviders#blankStrings} or {@code com.example.OuterClass$InnerClass#classMethod}.
 	 *
 	 * <p>If no factory method names are declared, a method within the test class
 	 * that has the same name as the test method will be used as the factory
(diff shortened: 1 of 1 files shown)


marcphilipp (approved):
Thanks!

VirtualTim:
It seems a bit odd that the coverage report includes comments in it's lines of code. It's weird that adding a comment makes the code coverage _worse_.

marcphilipp:
Don't worry about that line. It's not actually a comment but a line that's covered or not base on whether concurrent threads manage to trigger the race condition:

image

It's unrelated to your changes.

marcphilipp (approved):

marcphilipp:
Thanks! 👍

sbrannen:
Thanks for the PR!

> This took me a while to figure out the syntax, so I thought improving the documentation could be beneficial.

It definitely helps to have an example in the documentation. 👍 

The example `InnerClass` is technically a nested class, since "inner" classes in Java are non-static (according to the terminology in the Java Language Specification). 

I realize that's very specific, but in order to avoid any confusion I'm going to change the name of that class.

> As an aside, there are also no tests for `MethodSource` with an inner class.

We actually have several tests for `@MethodSource` factory methods in nested classes. All of the tests in [MethodArgumentsProviderTests](https://github.com/junit-team/junit5/blob/main/junit-jupiter-params/src/test/java/org/junit/jupiter/params/provider/MethodArgumentsProviderTests.java) that use `ExternalFactoryMethods` cover these use cases.

Though, now that I've taken a look at those tests again, the name of the `providesArgumentsUsingExternalFactoryMethodFromStaticNestedClass()` test method should probably be changed to "DoubleNested" since all of the methods are in a static nested class. 😉

VirtualTim:
Thanks for the clarification @sbrannen. I think I missed those tests in my search because the names were concatenated.
Good idea about renaming things. But at least it looks like I wasn't the only one who made that mistake.

PR #2784, opened 2021-11-23
Support searches for repeated annotations declared as meta-annotations on container annotations for different repeated annotations

gordeevnm:
## Overview

I created annotation like this:

```java
@Target(ElementType.METHOD)
@Documented
@TestTemplate
@ExtendWith(DirTestCasesExtension.class)
@interface DirArgSource {
    /**
     * Describe specific files for test case args
     */
    TestCase[] value() default {};

    /**
     * Parse test cases from specified directory
     */
    String[] directory() default {};
}
```

And I want to use it with only filled `directory`, but current implementation of `org.junit.platform.commons.util.AnnotationUtils#findRepeatableAnnotations(...)` does not find `@ExtendWith` in this case

_I haven't tested this patch, it is more like demo of what i mean, but i think it must work._ 🙂

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/AnnotationUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/AnnotationUtils.java
@@ -351,6 +351,11 @@ public final class AnnotationUtils {
 				}
 				// Nested container annotation?
 				else if (isRepeatableAnnotationContainer(candidateAnnotationType)) {
+					
+					// Find annotations that are directly present or meta-present on directly present annotations.
+					findRepeatableAnnotations(candidateAnnotationType.getDeclaredAnnotations(), annotationType, containerType, inherited, found,
+						visited);
+					
 					Method method = ReflectionUtils.tryToGetMethod(candidateAnnotationType, "value").toOptional().get();
 					Annotation[] containedAnnotations = (Annotation[]) ReflectionUtils.invokeMethod(method, candidate);
 


marcphilipp:
@gordeevnm Could you please provide a failing test case that demonstrates the issue?

gordeevnm:
@marcphilipp something like this:
https://gist.github.com/gordeevnm/0616504bc2946c15afd230408d53aa0f

sorry, i can't provide fully worked(compilable) demo right now, so I hope the example above is informative enough

sbrannen:
JUnit's annotation search algorithm that finds `@ExtendWith` is not affected by the attributes declared in your `@DirArgSource` annotation.

In addition, the search algorithm in question works for `@ParameterizedTest` as well as one of the test methods in the Gist you posted.

I assume it actually works for both of your test methods.

The error message you have posted in your Gist (_"You must register at least one TestTemplateInvocationContextProvider that supports @TestTemplate method"_) is generated when none of the registered `TestTemplateInvocationContextProvider`s
responds with `true` for `supportsTestTemplate(...)`.

It therefore appears that your `DirectoryArgsTestExtension` does claim to support a `@TestTemplate` method annotated with `@DirectoryArgsTestTemplate(directory = "test_case_data")`.

However, if you are able to create a minimal sample that reproduces the error, feel free to post back here with additional information.

gordeevnm:
Below is a link to the project with the demo. The problem is that for the test with `@FileArgsTemplate` (in the demo project) `ext.FileArgsExtension#supportsTestTemplate` does not even run(and the constructor too)

https://github.com/gordeevnm/junit-bug-demo

gordeevnm:
In debugger I tried to apply my patch (attached to this issue) and it worked as I want it to. I'm not sure if it's a successful patch because I understand that this annotation search algorithm is used in other places and my commit might break something. But please pay attention to this point. Thanks 🙂

gordeevnm:
@sbrannen should i create issue instead of this PR?

sbrannen:
Thanks for the supplying the demo project.

I'm reopening this PR so that the team can investigate if there is need for improvement in the annotation search algorithm.

Please note, however, that the recommended solution to the problem may involve a change in the way you have structured your annotations.

After taking a glance at the demo project, I noticed that the container for your repeated annotation declares attributes other than `value`.

Although that seems to be technically possible, in my experience that is nonstandard. 

When designing annotations I recommend that containers for repeatable annotations never be looked up explicitly. Their existence is only necessary for the compiler. Along those lines, I would not recommend that a container annotation declare any attributes other than `value`.

In summary, your users should never have to know about the container annotation.

If you want to allow the users to provide additional information via an annotation, it's best to create a dedicated annotation for that use case which is not a container annotation.

sbrannen:
FYI: I have changed the title of this PR to reflect the scope of the proposal as I currently understand it.

@gordeevnm, does the new title reflect your intention?

gordeevnm:
Yes, thank you, new title seems better. 

And I like your point about not adding extra attributes to the container annotation. You may be absolutely right, so if you decide to close this pr I won't be upset 🙂

marcphilipp:
**Team decision:** close with the above rationale

Issue #2787, opened 2021-12-05
Extend ParallelExecutionConfiguration to specify a predicate for determing if a pool is saturated


klease:
I'm working on the Apache Camel project which uses parallel execution on a module with a large number of test cases (over 2400 classes and over 6000 tests.) After upgrading to 5.7.2 or later we were getting RejectedExecutionException which we solved by using a custom configuration with a large value for maxPoolSize.

The cause appears to be that the Tasks for the test classes all call `ForkJoinPool.managedBlock` and the tests themselves can take up to a few seconds to run. Fairly quickly all threads in the ForkJoinPool are busy and therefore each new call to managedBock causes a new thread to be allocated up to the maxPoolSize. Each of these new threads usually runs only 1 or 2 tests.

I found that by providing a predicate for _saturated_ when creating the ForkJoinPool, and having it return true, it was possible to run all the tests with a much smaller maxPoolSize.

Currently it's not possible to control this parameter using a custom implementation of `ParallelExecutionConfiguration`.

So I propose to add a method to the interface such as the following:

```java
/**
 * Get the predicate called when there the ForkJoinPool has reached MaxPoolSize
 * and there are no threads available, or null. The predicate should return true if the
 * blocking thread should try again. If the predicate is null or returns false a
 * RejectedExcecutionException is thrown.
 * @return a Predicate accepting a ForkJoinPool as the parameter or null.
 */
Predicate<? super ForkJoinPool> getSaturatePredicate();
```

The `DefaultParallelExecutionConfiguration` would return null.

The class `ForkJoinPoolHierarchicalTestExecutorService` would use `configuration.getSaturatePredicate()` when creating the `ForkJoinPool`.

sbrannen:
How would one supply the `Predicate`?

klease:
By implementing the new method in a custom implementation of ParallelExecutionConfiguration, for example:
```
        @Override
    	public Predicate<? super ForkJoinPool> getSaturatePredicate() {
        	return (ForkJoinPool p) -> {return true;};
        }
```

sbrannen:
Thanks for the clarification.

Wouldn't `getSaturatedPredicate()` make more sense (_saturated_)?

marcphilipp:
**Team decision:** Add `ParallelExecutionConfiguration.getSaturatedPredicate()`

@klease Would you be interested in submitting a PR for this?

klease:
> @klease Would you be interested in submitting a PR for this?

[...]

PR #2792, opened 2021-12-10
Allow specifying ForkJoinPool saturate predicate

klease:
Add the method getSaturatePredicate to the interface ParallelExecutionConfiguration.
This allows a custom configuration to supply a predicate to handle the case
where the ForkJoinPool has reached the maximum pool size but all threads
are currently blocked.
Add a constructor to DefaultParallelExecutionConfiguration allowing to
speciy the predicate and modify the existing constructor to specify
null for the predicate.
Modify DefaultParallelExecutionConfigurationStrategyTests to check the
saturate predicate value in the default and custom configurations.

Issue: #2787
---


documentation/src/docs/asciidoc/release-notes/release-notes-5.9.0-M1.adoc
```diff
@@ -26,6 +26,8 @@ on GitHub.
 * `ConfigurationParameters` has now a `keySet()` method allowing to retrieve
   all configuration parameter keys.
 * New `IterationSelector` for selecting a subset of a test's or container's iterations.
+* `ParallelExecutionConfiguration` has a new method allowing to supply a predicate to
+   be called when the ForkJoinPool is saturated.
```
juliette-derancourt, line 30:
```suggestion
   be called when the `ForkJoinPool` is saturated.
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
```diff
@@ -60,4 +61,16 @@
...
+	 * Get the predicate called when there the ForkJoinPool is saturated.
+	 * This occurs when the ForkJoinPool has reached MaxPoolSize
+	 * and there are no threads available. The predicate
+	 * should return true if the blocking thread should try again.
+	 * If the predicate is null or returns false a RejectedExcecutionException
+	 * is thrown.
+	 * @return a Predicate accepting a ForkJoinPool as the parameter or null.
+	 * @since 5.9
+	 */
+	Predicate<? super ForkJoinPool> getSaturatePredicate();
```
juliette-derancourt, line 74: Please add an `@API` indication here as well

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
```diff
@@ -60,4 +61,16 @@
 	 */
 	int getKeepAliveSeconds();
 
+	/**
+	 * Get the predicate called when there the ForkJoinPool is saturated.
```
juliette-derancourt, line 65:
```suggestion
	 * Get the predicate called when the {@code ForkJoinPool} is saturated.
```
(Same remark for classes below)

juliette-derancourt (changes requested):
Hello @klease, I made a few comments and launched the CI builds :)

documentation/src/docs/asciidoc/release-notes/release-notes-5.9.0-M1.adoc
sbrannen: After having re-read the documentation, it's my understanding that the predicate is actually used to decide if the pool should be allowed "to saturate".

In other words, I don't think the predicate is called when the pool is already "saturated".

klease: I interpreted "saturated" as meaning the number of threads in the pool has reached the maximum and they are all busy or blocked. In this case, the predicate is used to allow it to be saturated (true) or not (false, in which case an exception is thrown.) But I'm happy to rephrase it if you think it makes more sense.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
klease: Like this: `@API(since = "1.9")`, or 5.9? In addition to the `@since` or instead of it?

juliette-derancourt: In addition to the `@since` ;)

And you're right, the version should be `1.9` because this is in the platform module.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
```diff
@@ -60,4 +61,17 @@
...
+	 * Get the predicate called when the {@code ForkJoinPool} is saturated.
+	 * This occurs when the ForkJoinPool has reached MaxPoolSize
+	 * and there are no worker threads available. The predicate
+	 * should return true if the calling thread should block.
+	 * If the predicate is null or returns false a RejectedExcecutionException
+	 * is thrown.
+	 * @return a Predicate accepting a ForkJoinPool as the parameter or null.
+	 * @since 1.9
+	 * @API(since = "1.9")
+	 */
```
marcphilipp, line 74:
```suggestion
	 * @since 1.9
	 */
	@API(status = EXPERIMENTAL, since = "1.9")
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
```diff
@@ -60,4 +61,17 @@
...
+	Predicate<? super ForkJoinPool> getSaturatePredicate();
```
marcphilipp, line 75: This needs to be a `default` method that returns `null` to preserve backwards compatibility.
```suggestion
	default Predicate<? super ForkJoinPool> getSaturatePredicate() {
		return null;
	}
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
```diff
@@ -60,4 +61,17 @@
 	 */
 	int getKeepAliveSeconds();
 
+	/**
+	 * Get the predicate called when the {@code ForkJoinPool} is saturated.
+	 * This occurs when the ForkJoinPool has reached MaxPoolSize
+	 * and there are no worker threads available. The predicate
+	 * should return true if the calling thread should block.
+	 * If the predicate is null or returns false a RejectedExcecutionException
+	 * is thrown.
```
marcphilipp, line 70: I wouldn't go into details here but link to `ForkJoinPool`.

```suggestion
	 * Get the saturate predicate to be used for the execution's {@link ForkJoinPool}.
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfiguration.java
```diff
@@ -20,14 +23,21 @@
...
 	private final int keepAliveSeconds;
+	private final Predicate<? super ForkJoinPool> saturatePredicate;
 
 	DefaultParallelExecutionConfiguration(int parallelism, int minimumRunnable, int maxPoolSize, int corePoolSize,
 			int keepAliveSeconds) {
+		this(parallelism, minimumRunnable, maxPoolSize, corePoolSize, keepAliveSeconds, null);
+	}
+
+	DefaultParallelExecutionConfiguration(int parallelism, int minimumRunnable, int maxPoolSize, int corePoolSize,
+			int keepAliveSeconds, Predicate<? super ForkJoinPool> saturatePredicate) {
```
marcphilipp, line 34: I don't think we need to change this class at all.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
```diff
@@ -60,4 +61,17 @@
...
+	 * @return a Predicate accepting a ForkJoinPool as the parameter or null.
+	 * @since 1.9
```
marcphilipp, line 72:
```suggestion
	 * @since 1.9
	 * @see ForkJoinPool#ForkJoinPool(int, ForkJoinPool.ForkJoinWorkerThreadFactory, Thread.UncaughtExceptionHandler, boolean, int, int, int, Predicate, long, TimeUnit)
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
```diff
@@ -60,4 +61,17 @@
...
 	int getKeepAliveSeconds();
 
+	/**
+	 * Get the predicate called when the {@code ForkJoinPool} is saturated.
+	 * This occurs when the ForkJoinPool has reached MaxPoolSize
+	 * and there are no worker threads available. The predicate
+	 * should return true if the calling thread should block.
+	 * If the predicate is null or returns false a RejectedExcecutionException
+	 * is thrown.
+	 * @return a Predicate accepting a ForkJoinPool as the parameter or null.
```
marcphilipp, line 71:
```suggestion
	 * @return the saturate predicate to be passed to the ForkJoinPool constructor; may be {@code null}
```

documentation/src/docs/asciidoc/release-notes/release-notes-5.9.0-M1.adoc
```diff
@@ -26,6 +26,8 @@ on GitHub.
 * `ConfigurationParameters` has now a `keySet()` method allowing to retrieve
   all configuration parameter keys.
 * New `IterationSelector` for selecting a subset of a test's or container's iterations.
+* `ParallelExecutionConfiguration` has a new method allowing to supply a predicate to
+   be called when the `ForkJoinPool` is about to be saturated.
```
marcphilipp, line 30:
```suggestion
* `ParallelExecutionConfiguration` allows configuring a predicate to be called when
  the `ForkJoinPool` is about to be saturated.
```

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategyTests.java
```diff
@@ -177,7 +181,9 @@ void customStrategyThrowsExceptionWhenClassDoesNotExist() {
 	static class CustomParallelExecutionConfigurationStrategy implements ParallelExecutionConfigurationStrategy {
 		@Override
 		public ParallelExecutionConfiguration createConfiguration(ConfigurationParameters configurationParameters) {
-			return new DefaultParallelExecutionConfiguration(1, 2, 3, 4, 5);
+			return new DefaultParallelExecutionConfiguration(1, 2, 3, 4, 5, (p) -> {
+				return true;
+			});
```
marcphilipp, line 186:
```suggestion
			return new DefaultParallelExecutionConfiguration(1, 2, 3, 4, 5, __ -> true);
```

marcphilipp (changes requested):

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategyTests.java
klease: Done.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
klease: Done

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
klease: Done

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ParallelExecutionConfiguration.java
klease: Done.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfiguration.java
klease: I added this to allow a custom configuration to use the DefaultParallelExecutionConfiguration with a non-null saturate , as done in the unit test. Otherwise it has to create a subclass.

marcphilipp: This class is package-private, i.e., not usable by custom config classes.

klease: Duh, you're right. In fact, I had noticed this when I tried to implement a custom strategy for Camel and I had to create a class with all interface methods.
I suppose the interface is not so large or the use case so common to merit having an extendable base class?

christophs78:
Also helps to address following issue related to Java 17 changes to ForkJoinPool: https://stackoverflow.com/questions/69768534/why-does-java-17-throw-a-rejectedexecutionexception-when-adding-tasks-to-a-forkj

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfiguration.java
klease: So, would you suggest that I remove the changes in the DefaultParallelExecutionConfiguration class and modify the unit test accordingly? I think that's the only remaining open point of the suggested changes.

juliette-derancourt: I think that's what he meant, yes :+1:

klease:
All requested changes are now committed.

marcphilipp (approved):

klease:
Thanks @marcphilipp!

marcphilipp:
Thanks to you and sorry for the delay!

PR #2880, opened 2022-04-08
Use registered ParameterResolver(s) when invoking @MethodSource factory methods

### LINKED ISSUE #2191: Use registered ParameterResolver(s) when invoking @MethodSource factory methods
Currently, the `MethodSource` methods are supposed to not have any parameters, however, it'll help to have parameters to be resolved in these methods if registered similar to Before/After methods.

Along with #2190, this will give access to `MethodSource` methods to possibly use `Extension.Store` and store any initialized details, which could be purged in `@AfterAll` methods

Also, in case of creating dynamic tests using details from a resolved Object, `ParameterResolver` can enhance abilities of `MethodSource` by a fair bit.

Currently, the way to achieve the same is to have a class implementing `ArgumentsProvider` and possibly manage the value of the object there, or have it in the `MethodSource` method itself statically initializing it in test class.

juliette-derancourt:
## Overview

This PR adds the possibility to provide arguments to a `@MethodSource` factory method.
They will be resolved by any registered `ParameterResolver` supporting this type of parameter.

Closes #2191

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
@@ -68,7 +68,8 @@ class MethodArgumentsProvider implements ArgumentsProvider, AnnotationConsumer<M
 		String methodParameters = methodParts[2];
 
 		return ReflectionUtils.findMethod(loadRequiredClass(className), methodName, methodParameters).orElseThrow(
-			() -> new JUnitException(format("Could not find method [%s] in class [%s]", methodName, className)));
+			() -> new JUnitException(
+				format("Could not find method [%s(%s)] in class [%s]", methodName, methodParameters, className)));
 	}
 
 	private Method getMethodByShortName(Class<?> testClass, String methodName) {
@@ -76,7 +77,7 @@ class MethodArgumentsProvider implements ArgumentsProvider, AnnotationConsumer<M
 		Preconditions.condition(methods.size() > 0,
 			() -> format("Could not find method [%s] in class [%s]", methodName, testClass.getName()));
 		Preconditions.condition(methods.size() <= 1,
-			() -> format("Several factory method named [%s] were found in class [%s]", methodName,
+			() -> format("Several factory methods named [%s] were found in class [%s]", methodName,
 				testClass.getName()));
 		return methods.get(0);
 	}
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
@@ -15,6 +15,7 @@ import static org.junit.jupiter.params.provider.Arguments.arguments;
 
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.List;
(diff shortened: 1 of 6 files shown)


junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
```diff
@@ -43,23 +44,20 @@ public void accept(MethodSource annotation) {
...
 	private Method getMethod(ExtensionContext context, String factoryMethodName) {
-		if (StringUtils.isNotBlank(factoryMethodName)) {
-			if (factoryMethodName.contains("#")) {
-				return getMethodByFullyQualifiedName(factoryMethodName);
-			}
-			else {
-				return ReflectionUtils.getRequiredMethod(context.getRequiredTestClass(), factoryMethodName);
-			}
+		if (StringUtils.isBlank(factoryMethodName)) {
+			return getMethodByShortName(context.getRequiredTestMethod().getName(), context.getRequiredTestClass());
```
marcphilipp, line 55: I think in this case we need to check for `method.getParameterCount() == 0` to preserve the existing behavior. Maybe pass a `Predicate<Method> additionalCondition` to `getMethodByShortName`?

marcphilipp (changes requested):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
juliette-derancourt: It depends, do we want to support parameters in this case too? If so, we could apply the "search" logic by name, and exclude the actual test method (unless we don't have a way to differentiate it from the others)

But now that I think about it, it would be a problem is the test method and factory method want to declare the exact same parameters

Maybe it's simpler to not support parameters for factory method named after the test method

juliette-derancourt: I reverted to previous behavior for now in this case (i.e. only searching for method with no parameters)

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
```diff
@@ -68,10 +67,18 @@ private Method getMethodByFullyQualifiedName(String fullyQualifiedMethodName) {
 		String methodName = methodParts[1];
 		String methodParameters = methodParts[2];
 
-		Preconditions.condition(StringUtils.isBlank(methodParameters),
-			() -> format("factory method [%s] must not declare formal parameters", fullyQualifiedMethodName));
+		return ReflectionUtils.findMethod(loadRequiredClass(className), methodName, methodParameters).orElseThrow(
+			() -> new JUnitException(format("Could not find method [%s] in class [%s]", methodName, className)));
```
sbrannen, line 71:
```suggestion
			() -> new JUnitException(format("Could not find method [%s(%s)] in class [%s]", methodName, methodParameters, className)));
```

sbrannen (changes requested):
I think this looks good. 👍 

juliette-derancourt:
@sbrannen Changes applied, thanks for the review! :ok_hand:

marcphilipp (approved):

PR #2882, opened 2022-04-09
Use user's console charset, if available

sormuras:
## Overview

Prior to this commit the charset returned by `Charset.defaultCharset()` was always used to determine the default theme used by the JUnit Console Launcher application.

Now, on Java 17 and later, Java's Console API is queried in order to retrieve the user-specified of the charset used by the system console -- before falling back to the default charset.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ConsoleUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ConsoleUtils.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2015-2022 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.util;
+
+import static org.apiguardian.api.API.Status.INTERNAL;
+
+import java.nio.charset.Charset;
+
+import org.apiguardian.api.API;
+
+/**
+ * Collection of utilities for working with {@code java.io.Console}
+ * and friends.
+ *
+ * <h3>DISCLAIMER</h3>
+ *
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.9
+ */
+@API(status = INTERNAL, since = "1.9")
+public class ConsoleUtils {
+
+	/**
+	 * {@return the charset of the console}
+	 */
+	public static Charset charset() {
+		return Charset.defaultCharset();
+	}
+}
--- a/junit-platform-commons/src/main/java17/org/junit/platform/commons/util/ConsoleUtils.java
+++ b/junit-platform-commons/src/main/java17/org/junit/platform/commons/util/ConsoleUtils.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2015-2022 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.util;
+
+import static org.apiguardian.api.API.Status.INTERNAL;
+
+import java.io.Console;
+import java.nio.charset.Charset;
+
+import org.apiguardian.api.API;
+
+/**
+ * Collection of utilities for working with {@code java.io.Console}
+ * and friends.
+ *
+ * <h3>DISCLAIMER</h3>
+ *
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.9
+ */
+@API(status = INTERNAL, since = "1.9")
+public class ConsoleUtils {
+
+	/**
+	 * {@return the charset of the console}
+	 */
+	public static Charset charset() {
+		Console console = System.console();
+		return console != null ? console.charset() : Charset.defaultCharset();
+	}
+}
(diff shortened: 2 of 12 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ConsoleUtils.java
```diff
@@ -0,0 +1,40 @@
...
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.9
+ */
+@API(status = INTERNAL, since = "1.9")
+public class ConsoleUtils {
+
+	/**
+	 * {@return the charset of the console}
```
marcphilipp, line 35: Shouldn't this be the following?

```suggestion
	 * @return the charset of the console
```

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ConsoleUtils.java
```diff
@@ -0,0 +1,40 @@
...
+ * <h3>DISCLAIMER</h3>
+ *
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.9
+ */
+@API(status = INTERNAL, since = "1.9")
+public class ConsoleUtils {
```
marcphilipp, line 32: I don't think this needs to be in junit-platform-commons but would move it to junit-platform-console and make it package-private.

marcphilipp (changes requested):

junit-platform-commons/src/main/java17/org/junit/platform/commons/util/ConsoleUtils.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+@API(status = INTERNAL, since = "1.9")
+public class ConsoleUtils {
+
+	/**
+	 * {@return the charset of the console}
+	 */
+	public static Charset charset() {
+		Console console = System.console();
+		return console != null ? console.charset() : Charset.defaultCharset();
```
marcphilipp, line 40: I think we should add a unit test for this.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ConsoleUtils.java
sormuras: See https://bugs.openjdk.java.net/browse/JDK-8075778 how to "avoid duplication of return information in simple situations."

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ConsoleUtils.java
sormuras: Thought of this too, but a having a source set that targets Java 17 directly in `junit-platform-commons` seems more future-proof to me, especially when it comes to overloading other common functionality.

junit-platform-commons/src/main/java17/org/junit/platform/commons/util/ConsoleUtils.java
sormuras: True.
At least, I will add an integration test.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ConsoleUtils.java
marcphilipp: Nice! TIL

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ConsoleUtils.java
marcphilipp: I don't see why we couldn't add it then, when we need it.

sormuras: Fine. Moving it to `junit-platform-console`.

sormuras: Moved via https://github.com/junit-team/junit5/pull/2882/commits/e187063e3de2891c31ed23931297f696c34d588d

junit-platform-commons/src/main/java17/org/junit/platform/commons/util/ConsoleUtils.java
sormuras: Added unit test that asserts Java 17+ code path works.

platform-tests/src/test/java/org/junit/platform/console/options/ConsoleUtilsTests.java
```diff
@@ -0,0 +1,28 @@
...
+import java.nio.charset.Charset;
+
+import org.junit.jupiter.api.Test;
+
+class ConsoleUtilsTests {
+
+	@Test
+	void consolesCharsetIsEither() {
+		var console = System.console();
+		var expected = console != null ? console.charset() : Charset.defaultCharset();
```
marcphilipp, line 24: Isn't `System.console` always `null` here when running tests?

platform-tests/src/test/java/org/junit/platform/console/options/ConsoleUtilsTests.java
```diff
@@ -0,0 +1,28 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.nio.charset.Charset;
+
+import org.junit.jupiter.api.Test;
+
+class ConsoleUtilsTests {
+
+	@Test
+	void consolesCharsetIsEither() {
```
marcphilipp, line 22: This test name sounds incomplete to me.

marcphilipp (approved):

platform-tests/src/test/java/org/junit/platform/console/options/ConsoleUtilsTests.java
sormuras: ...it depends. ;-)

platform-tests/src/test/java/org/junit/platform/console/options/ConsoleUtilsTests.java
sormuras: Or...

sormuras: Completed via https://github.com/junit-team/junit5/pull/2882/commits/cbf4b613efcc0dd2520a251a1e9c137b9bc36ab9

Issue #2883, opened 2022-04-10
TempDir deletion throws StackOverflowError


andrewparmet:
I have a unit test that creates some files in a temporary directory. Notably it's a Docker container that is creating the files. The test passes locally on my machine (running on a Mac) but fails in GitHub actions: https://github.com/andrewparmet/buf-gradle-plugin/runs/5959677186?check_suite_focus=true with a StackOverflowError as it's trying to clean up:

```
java.lang.StackOverflowError
	at java.base/sun.nio.fs.UnixNativeDispatcher.unlink0(Native Method)
	at java.base/sun.nio.fs.UnixNativeDispatcher.unlink(UnixNativeDispatcher.java:156)
	at java.base/sun.nio.fs.UnixFileSystemProvider.implDelete(UnixFileSystemProvider.java:236)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
	at java.base/java.nio.file.Files.delete(Files.java:1142)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath$1.resetPermissionsAndTryToDeleteAgain(TempDirectory.java:250)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath$1.deleteAndContinue(TempDirectory.java:238)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath$1.visitFile(TempDirectory.java:218)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath$1.visitFile(TempDirectory.java:199)
	at java.base/java.nio.file.Files.walkFileTree(Files.java:2725)
	at java.base/java.nio.file.Files.walkFileTree(Files.java:2797)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath$1.resetPermissionsAndTryToDeleteAgain(TempDirectory.java:247)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath$1.deleteAndContinue(TempDirectory.java:238)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath$1.postVisitDirectory(TempDirectory.java:223)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath$1.postVisitDirectory(TempDirectory.java:199)
	at java.base/java.nio.file.Files.walkFileTree(Files.java:2743)
	at java.base/java.nio.file.Files.walkFileTree(Files.java:2797)

[...]

PR #2884, opened 2022-04-10
Remove BOM dependencies from artifacts defined in the BOM to enable participation in Gradle source dependencies

andrewparmet:
## Overview

While investigating solutions to https://github.com/junit-team/junit5/issues/2883 I attempted to use a Gradle [source dependency](https://blog.gradle.org/introducing-source-dependencies) to depend on the latest version of this project since I couldn't find a published snapshot.

The build with a source dependency failed due to what looks like a circular dependency: most artifacts have a dependency on a BOM that declares them (and hence depends on them).

```
Could not determine the dependencies of task ':compileTestKotlin'.
> Could not resolve all task dependencies for configuration ':testCompileClasspath'.
   > Could not find any version that matches org.junit.jupiter:junit-jupiter:5.9.0-SNAPSHOT.
     Searched in the following locations:
       - Git repository at https://github.com/junit-team/junit5.git
     Required by:
         project : > project :junit5:junit-jupiter > project :junit5:junit-bom
```

When I remove the circular dependency with this change the source dependency build succeeds.

Open questions:
- Is there a reason the BOM is used in this way?
- Should I try to add a test to cover this? Could be tricky.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.9.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.9.0-M1.adoc
@@ -74,7 +74,8 @@ on GitHub.
 
 ==== Bug Fixes
 
-* ❓
+* JUnit Jupiter can now participate in builds using a Gradle
+[source dependency](https://blog.gradle.org/introducing-source-dependencies).
 
 ==== Deprecations and Breaking Changes
 
--- a/junit-jupiter-api/junit-jupiter-api.gradle.kts
+++ b/junit-jupiter-api/junit-jupiter-api.gradle.kts
@@ -6,7 +6,6 @@ plugins {
 description = "JUnit Jupiter API"
 
 dependencies {
-	api(platform(projects.junitBom))
 	api(libs.opentest4j)
 	api(projects.junitPlatformCommons)
 
--- a/junit-jupiter-engine/junit-jupiter-engine.gradle.kts
+++ b/junit-jupiter-engine/junit-jupiter-engine.gradle.kts
@@ -10,7 +10,6 @@ plugins {
 description = "JUnit Jupiter Engine"
 
 dependencies {
-	api(platform(projects.junitBom))
 	api(projects.junitPlatformEngine)
 	api(projects.junitJupiterApi)
 
--- a/junit-jupiter-migrationsupport/junit-jupiter-migrationsupport.gradle.kts
+++ b/junit-jupiter-migrationsupport/junit-jupiter-migrationsupport.gradle.kts
@@ -7,7 +7,6 @@ plugins {
 description = "JUnit Jupiter Migration Support"
 
 dependencies {
-	api(platform(projects.junitBom))
 	api(libs.junit4)
 	api(projects.junitJupiterApi)
 
--- a/junit-jupiter-params/junit-jupiter-params.gradle.kts
+++ b/junit-jupiter-params/junit-jupiter-params.gradle.kts
@@ -7,7 +7,6 @@ plugins {
 description = "JUnit Jupiter Params"
 
 dependencies {
-	api(platform(projects.junitBom))
 	api(projects.junitJupiterApi)
 
 	compileOnlyApi(libs.apiguardian)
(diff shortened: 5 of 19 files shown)


marcphilipp:
This is actually on purpose so versions automatically get aligned in consuming Gradle builds.

Have you tried using it as an [included build](https://docs.gradle.org/current/userguide/composite_builds.html) instead, e.g. using [this plugin](https://melix.github.io/includegit-gradle-plugin/latest/index.html)?

marcphilipp:
Closing with the above rationale. Please let me know if you have any questions.

andrewparmet:
Aren't versions are automatically aligned via standard project dependencies? For example, depending on `junit-jupiter-engine` in a project without the BOM dependency will pull in `junit-jupiter-api` of the same version because of default dependency resolution from https://github.com/junit-team/junit5/blob/main/junit-jupiter-engine/junit-jupiter-engine.gradle.kts#L15.

For example, using my fork published locally from my plugin project:
``` kotlin
// build.gradle.kts
dependencies {
    testImplementation("org.junit.jupiter:junit-jupiter:5.9.0-SNAPSHOT")
}
```

Partial output of `./gradlew dependencies`:
```
testCompileClasspath - Compile classpath for compilation 'test' (target  (jvm)).
+--- org.junit.jupiter:junit-jupiter:5.9.0-SNAPSHOT
|    +--- org.junit.jupiter:junit-jupiter-api:5.9.0-SNAPSHOT
|    |    +--- org.opentest4j:opentest4j:1.2.0
|    |    +--- org.junit.platform:junit-platform-commons:1.9.0-SNAPSHOT
|    |    |    \--- org.apiguardian:apiguardian-api:1.1.2
|    |    \--- org.apiguardian:apiguardian-api:1.1.2
|    \--- org.junit.jupiter:junit-jupiter-params:5.9.0-SNAPSHOT
|         +--- org.junit.jupiter:junit-jupiter-api:5.9.0-SNAPSHOT (*)
|         \--- org.apiguardian:apiguardian-api:1.1.2
```

I believe the standard pattern for using a BOM dependency is to declare your module dependencies without a version and then declare a platform dependency on the BOM in the consuming project. See e.g. https://github.com/grpc/grpc-java for the configuration, and then a consumer would do:

``` kotlin
// build.gradle.kts
dependencies {
    testImplementation(platform("org.junit.jupiter:junit-bom:5.9.0-SNAPSHOT"))
    implementation("org.junit.jupiter:junit-jupiter")
}
```

marcphilipp:
For example when using multiple engines or adding a dependency to a platform artifact that isn't a transitive dependency of Jupiter (e.g. to implement a `LauncherSessionListener`) that falls short.

marcphilipp:
See also #2068 for more details.

andrewparmet:
I see. Should a consumer using multiple unrelated dependencies be required to specify the BOM themselves though?

Comparing configurations, here's what it would have to look like with this change and no BOM:

```
dependencies {
    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.9.0-SNAPSHOT")
    testRuntimeOnly("org.junit.jupiter:junit-vintage-engine:5.9.0-SNAPSHOT")
}
```

whereas with the BOM included you could do this, which looks very strange to me:

```
dependencies {
    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.9.0-SNAPSHOT")
    testRuntimeOnly("org.junit.jupiter:junit-vintage-engine")
}
```

In theory a user could do:

```
dependencies {
    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine:5.9.0-SNAPSHOT")
    testRuntimeOnly("org.junit.jupiter:junit-vintage-engine:5.8.2")
}
```

If the modules are truly unrelated, is this actually something you want override by forcing a BOM? I can imagine a (very) weird scenario where someone may really want to use a different version of an engine. To me forcing a BOM violates the principle of least astonishment since it's not the standard way dependencies are managed; I expect to specify one myself if I want to force version alignment.

marcphilipp:
Yes, that scenario is sth. we want to prevent because mixing versions yields confusing errors at runtime. I'd still recommend importing the BOM explicitly so versions can be omitted in all places in consumer projects that need more than single artifact.

andrewparmet:
Sure, I've seen my fair share of strange runtime errors caused by version misalignment. But if the modules are actually independent in the sense that they don't share any common dependency, is it possible to get those kinds of errors? At that point they're independent projects coincidentally handed the same version number. If they are actually related then such a situation suggests they should share a common dependency and then transitive versioning would work properly.

PR #2890, opened 2022-04-12
Support ParallelExecutionConfigurationStrategy with Turkish locale

eocak:
Since Turkish have special characters as "ı" and "İ", uppercase convertion here causes all test methods to finish with exit code -1. Because uppercase of "i" is "İ" in Turkish and that breaks the code. "dynamic" > "DYNAMİC"

There is a special comment for Turkish characters on reference documentation of java.lang.String:

https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#toUpperCase(java.util.Locale)

---

---

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
@@ -13,7 +13,7 @@ package org.junit.platform.engine.support.hierarchical;
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
 import java.math.BigDecimal;
-
+import java.util.Locale;
 import org.apiguardian.api.API;
 import org.junit.platform.commons.JUnitException;
 import org.junit.platform.commons.util.Preconditions;
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
@@ -135,7 +135,7 @@ public enum DefaultParallelExecutionConfigurationStrategy implements ParallelExe
 
 	static ParallelExecutionConfigurationStrategy getStrategy(ConfigurationParameters configurationParameters) {
 		return valueOf(
-			configurationParameters.get(CONFIG_STRATEGY_PROPERTY_NAME).orElse("dynamic").toUpperCase(Locale.ENGLISH));
+			configurationParameters.get(CONFIG_STRATEGY_PROPERTY_NAME).orElse("dynamic").toUpperCase(Locale.ROOT));
 	}
 
 }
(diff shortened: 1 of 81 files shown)


sbrannen (dismissed):
Hi @eocak,

Thanks for submitting your "first pull request ever"! 👍 

It looks like you forgot to `import java.util.Locale;`.

Please address that and push your change.

Also, please add an entry to the 5.9 M1 Release Notes succinctly explaining the fix.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
```diff
@@ -13,7 +13,7 @@
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
 import java.math.BigDecimal;
-
+import java.util.Locale;
```
sbrannen, line 16: It looks like the deletion of that blank line will fail the build. 

Be sure to run `./gradlew check` before pushing commits.

If there are errors with Spotless, perform the action it suggests to fix formatting.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
```diff
@@ -133,7 +134,8 @@ public ParallelExecutionConfiguration createConfiguration(ConfigurationParameter
 	public static final String CONFIG_CUSTOM_CLASS_PROPERTY_NAME = "custom.class";
 
 	static ParallelExecutionConfigurationStrategy getStrategy(ConfigurationParameters configurationParameters) {
-		return valueOf(configurationParameters.get(CONFIG_STRATEGY_PROPERTY_NAME).orElse("dynamic").toUpperCase(Locale.ENGLISH));
+		return valueOf(
+			configurationParameters.get(CONFIG_STRATEGY_PROPERTY_NAME).orElse("dynamic").toUpperCase(Locale.ENGLISH));
```
sormuras, line 138: Shouldn't it be `Locale.ROOT` instead of `Locale.ENGLISH`?

> Useful constant for the root locale. The root locale is the locale whose language, country, and variant are empty ("") strings. This is regarded as the base locale of all locales, and is used as the language/country neutral locale for the locale sensitive operations.

https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#ROOT

eocak: I did not know the existence of Locale.ROOT and I was sure that uppercase of **i** is **I** in English Locale.  That can be changed to ROOT, dont know which is the best choice.

jbduncan: `Locale.ROOT` is used internally in Kotlin's equivalent [`uppercase()`](https://github.com/JetBrains/kotlin/blob/769f4c630f7b3ab326cd32f643cb1def6663f8c5/libraries/stdlib/jvm/src/kotlin/text/StringsJVM.kt#L147) function, so that's what I'd be tempted to use.

marcphilipp (approved):

PR #2892, opened 2022-04-14
Set merge strategy to `union` for release note files

juliette-derancourt:
This would avoid merge conflicts on release note files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,4 @@
 *.key binary
 *.jar binary
 *.ttf binary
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -4,3 +4,4 @@
 *.key binary
 *.jar binary
 *.ttf binary
+release-notes-* merge=union
```
juliette-derancourt, line 7: Could also match the whole folder with:
```suggestion
**/release-notes/* merge=union
```

PR #2913, opened 2022-05-10
Add "abort" methods to Assumptions

ZeroOne3010:
## Overview

Greetings! I have added `abort()`, `abort(String)`, and `abort(Supplier<String>)` methods into the `Assumptions` class, along with the relevant tests and release notes. These are analogous to the `fail` methods in the `Assertions` class in that they do not validate anything but just abort the current test. 

I found myself looking for a method like this a while back and was surprised they did not exist already. These now allow you to write some more complex business logic around the assumption and to fail it with one neat `abort()` call, instead of an ugly `assumeTrue(false)` or `assumeFalse(true)` call.

---

---

### Definition of Done

About this -- the pre-existing assumptions and failure methods did not do any precondition validation either, so I did not add such a thing in these `abort` methods either. It definitely will crash to a NullPointerException if someone passes a `null` `Supplier<String>` to it, but hey, the person writing the tests needs to bear some responsibility too, right?

The User Guide I haven't edited, but the release notes ("release-notes-5.9.0-M1.adoc") I did edit. I think it would be enough for these methods to appear in the JavaDocs and the Release Notes, would you agree?


ZeroOne3010:
I'll just mark the final checkbox, method preconditions, as done as it is in line with the current implementation.

ZeroOne3010:
@sbrannen Hi, any idea on the future of this PR? It has been a while. I've just rebased this on `main` and resolved the conflicts again.

marcphilipp:
@ZeroOne3010 Thanks for your contribution! 👍

ZeroOne3010:
Thank you! 😊

Issue #2915, opened 2022-05-13
assertIterableEquals: StackOverflow for self-iterating Iterable


thomaszub:
`assertIterableEquals`can lead to a StackOverflowError for `Path` or general for `Iterable` that iterates over itself. This is similar to the issue #2157. The PR to this issue fixes the success case only. 

I created a PR with a possible fix and test.

## Steps to reproduce

```java
@Test
void assertIterableEqualsWithDifferentListOfPath() {
	try {
		var expected = listOf(Path.of("1"), Path.of("2").resolve("3"), Path.of("5"));
		var actual = listOf(Path.of("1"), Path.of("2").resolve("4"), Path.of("5"));
		assertIterableEquals(expected, actual);
		expectAssertionFailedError();
	}
	catch (AssertionFailedError ex) {
		assertMessageEquals(ex, "iterable contents differ at index [1][1][0], expected: <3> but was: <4>");
	}
}
```
## Context

 - Used versions (Jupiter/Vintage/Platform): junit-jupiter-api/junit-jupiter-engine 5.8.2
 - Build Tool/IDE: Maven

## Deliverables

- [x] Provide fix and test for comparing `Iterable<? extends Iterable<?>>` instances.

---

I hereby agree to the terms of the JUnit Contributor License Agreement.

PR #2916, opened 2022-05-13
Track already visited pairs of Iterables to avoid stack overflow

thomaszub:
Issue: #2915

## Overview

The visited `Iterable` are tracked and the assert fails if a already visited `Iterable` is visited again. This avoids a StackOverflowError.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
@@ -17,11 +17,11 @@ import static org.junit.jupiter.api.AssertionUtils.formatValues;
 import static org.junit.jupiter.api.AssertionUtils.nullSafeGet;
 
 import java.util.ArrayDeque;
-import java.util.Collection;
 import java.util.Deque;
-import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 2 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
```diff
@@ -69,24 +79,39 @@ private static void assertIterableEquals(Iterable<?> expected, Iterable<?> actua
...
+	private static void assertIterablesNotAlreadyVisited(Iterable<?> expected, Iterable<?> actual,
+			Deque<Integer> indexes, Object messageOrSupplier, Collection<Iterable<?>> visitedIterablesExpected,
+			Collection<Iterable<?>> visitedIterablesActual) {
+		if (visitedIterablesExpected.contains(expected) || visitedIterablesActual.contains(actual)) {
+			failIterablesNotEqual(expected, actual, indexes, messageOrSupplier);
+		}
+		else {
+			visitedIterablesExpected.add(expected);
+			visitedIterablesActual.add(actual);
+		}
```
sbrannen, line 112: Can't this be simplified as follows?

```suggestion
		if (!visitedIterablesExpected.add(expected) && !visitedIterablesActual.add(actual)) {
			failIterablesNotEqual(expected, actual, indexes, messageOrSupplier);
		}
```

sbrannen (changes requested):

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertIterableEquals.java
thomaszub: Good point, will change it (the check must be "or": `!visitedIterablesExpected.add(expected) || !visitedIterablesActual.add(actual)`

thomaszub: 😁

thomaszub: Hi @sbrannen, could you please mark the request as solved. I force pushed the changes (maybe not best idea from my side). Thanks!

sbrannen (approved):

marcphilipp:
I added a test with a counterexample to the proposed tracking of input values in e196f8d and added a new implementation that keeps track of pairs of iterables that are under investigation and potentially reuses the result should the same iterables be encountered again.

marcphilipp (approved):

sbrannen (changes requested):
I think this looks good, but it would beneficial to add some inline documentation for how the algorithm currently works with regard to the "investigated elements" tracking. 

Also, the `SAME_ELEMENTS` enum constant could be renamed to something like `CONTAIN_SAME_ELEMENTS` or similar.

marcphilipp:
> I think this looks good, but it would beneficial to add some inline documentation for how the algorithm currently works with regard to the "investigated elements" tracking.

Done in 6cf3516ff209c980aa63e65960847e820b939c0c.



> Also, the `SAME_ELEMENTS` enum constant could be renamed to something like `CONTAIN_SAME_ELEMENTS` or similar.

Done in 0c175abde477b92ba8ee69bca0b0b2e4a8093155.

marcphilipp:
@thomaszub Thanks for the PR! 👍

Issue #2757, opened 2021-10-27
Provide a MethodOrderer which enforces unique order values


OLibutzki:
When a test is annoated with `@TestMethodOrder( MethodOrderer.OrderAnnotation.class )` we want the test methods to executed in the order of the `@Order` annotation.
When twei methods have the same order value the Javadoc states they "will be sortedarbitrarily adjacent to each other". That makes perfect sense.

Imagine this scenario: We have a tets class which contains 6 steps - each of them depends on its predecessor. Each of them is annotated by `@Order` with an incremented number. Later on we add  step between 2 and 3 and accidently copy the `@Order` annotation from another method.

This might lead to a flaky test as the same order value is used twice.

For sure this is the developer's mistake, but what about defining that each test method has to have an unique order value as we want to control the execution order deterministically?

We could have a special MethodOrderer for this purpose like
`@TestMethodOrder( MethodOrderer.EnforcingOrderAnnotation.class )`

The whole test class failes if there are at least two methods with the same order value.

marcphilipp:
I can see that being useful as well. If we add it, I think there should also be a configuration parameter for changing the default.

XIANY1203:
I am interested in this issue, may I have a try?

marcphilipp:
> Falling back to the "JUnit ordering magic" is something you rarely want to have when using `@TestMethodOrder`.

How would we detect that was the case? I think we can only make the `OrderAnnotation` orderer strict, e.g. via an annotation attribute, a configuration parameter, or both but not solve this in general.

OLibutzki:
We implemented to custom `OrderAnnotation` to solve it. As you can see we just log a warning in case of duplicated order values. Maybe this helps to get some inspiration.

By having a new dedicated `MethodOrderer`, you won't have issues regarding backward compatibility as the clients can decide which `MethodOrderer` they choose.

```
/**
 * This is a drop-in replacement for the {@link OrderAnnotation} method orderer. Unlike the original orderer, this class
 * checks and warns about non unique orders.
 */
@Log4j2
public final class UniqueOrderAnnotation extends OrderAnnotation {

	@Override
	public void orderMethods( final MethodOrdererContext context ) {
		final Map<Integer, List<MethodDescriptor>> groupedMethodDescriptors = context.getMethodDescriptors( )
				.stream( )
				.collect( groupingBy( UniqueOrderAnnotation::getOrder, toList( ) ) );
[...]

PR #2931, opened 2022-05-29
Try to give a switch that forces a unique order

### LINKED ISSUE #2930: Add a parameter to allow @TestMethodOrder to enforce unique order
I encountered the same problem as #2757 , in some scenarios I hope that the order I fill in is different, but in fact when faced with a large number of tests, sometimes I can't guarantee this, so I hope to get a boolean switch to detect all of this. At the same time I don't want to do it by adding a class as described in the previous issue, because it actually means that if I have other sorting methods in the future that I also want to have this functionality, I have to add an extra for this sorting method A class for the enforce state. So I tried adding a boolean enforceUniqueOrder switch to do it, I guess it's easier to manage and use

fll02020:
## Overview

---

---
I try to give a switch called enforceUniqueOrder for @TestMethodOder, so that it can be used in some cases
This is mainly to resolve #2930 and #2757
### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
@@ -57,290 +57,292 @@ import org.junit.platform.commons.util.ClassUtils;
 @API(status = STABLE, since = "5.7")
 public interface MethodOrderer {
 
-    /**
-     * Property name used to set the default method orderer class name: {@value}
-     *
-     * <h4>Supported Values</h4>
-     *
-     * <p>Supported values include fully qualified class names for types that
-     * implement {@link org.junit.jupiter.api.MethodOrderer}.
-     *
-     * <p>If not specified, test methods will be ordered using an algorithm that
-     * is deterministic but intentionally non-obvious.
-     *
-     * @since 5.7
-     */
-    @API(status = STABLE, since = "5.9")
-    String DEFAULT_ORDER_PROPERTY_NAME = "junit.jupiter.testmethod.order.default";
+	/**
+	 * Property name used to set the default method orderer class name: {@value}
+	 *
+	 * <h4>Supported Values</h4>
+	 *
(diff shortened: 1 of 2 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -235,6 +238,22 @@ public OrderAnnotation() {
...
+					removeDuplicate.add(getOrder(descriptor));
+				}
+			}
+
+			long distinctOrderCnt = removeDuplicate.size();
+			long methodCnt = descriptors.size();
+
+			if (methodCnt != distinctOrderCnt && distinctOrderCnt != 0)
+				throw new IllegalArgumentException(
+					"duplicate order" + "expected: <" + methodCnt + "> but was: <" + distinctOrderCnt + ">");
```
vlsi, line 255: Whould you please make the error message human-readable and actionable?
In other words, the error message should tell which tests collided, and it should suggest the way to fix the issue

fll02020: I know about it. I will try to fix it maybe Thursday or Friday

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MethodOrderer.java
```diff
@@ -235,6 +238,22 @@ public OrderAnnotation() {
 		 */
 		@Override
 		public void orderMethods(MethodOrdererContext context) {
+			List<? extends MethodDescriptor> descriptors = context.getMethodDescriptors();
+			Map<Integer, String> removeDuplicate = new LinkedHashMap<>();
```
vlsi, line 242: I believe the hash set/map is not needed. JUnit would sort the method descriptions anyway, so it would be better to scan over the sorted list and see if there's a set of consequent elements sharing the same `order`.

It would be nice to print all the violations rather than just the first for faster turnaround.

PR #2943, opened 2022-06-11
Add a public factory method to instantiate ExecutionRequest

### LINKED ISSUE #2912: Open up ExecutionRequest creation for external usage
Currently I'm working on a test engine that delegates work to other test engines.

Everything works great, except I have to use `ExecutionRequest`'s constructor which is marked as internal.

https://github.com/junit-team/junit5/blob/80bf93fd3cb841e76fc017dc759d8ef0df158d49/junit-platform-engine/src/main/java/org/junit/platform/engine/ExecutionRequest.java#L39-L41

So please add a _public_ way to instantiate `ExecutionRequest` or replace it with an interface like `EngineDiscoveryRequest`.

This will make variations of suite engine possible.

The other possibility is to directly test the `TestEngine#execute` method without using an internal API.

juliette-derancourt:
Resolves #2912

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/ExecutionRequest.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/ExecutionRequest.java
@@ -53,7 +53,9 @@ public class ExecutionRequest {
 	 * @param configurationParameters {@link ConfigurationParameters} that the
 	 * engine may use to influence test execution
 	 * @return a new {@code ExecutionRequest}; never {@code null}
+	 * @since 1.9
 	 */
+	@API(status = STABLE, since = "1.9")
 	public static ExecutionRequest create(TestDescriptor rootTestDescriptor,
 			EngineExecutionListener engineExecutionListener, ConfigurationParameters configurationParameters) {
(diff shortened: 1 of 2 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/ExecutionRequest.java
```diff
@@ -44,6 +44,21 @@ public ExecutionRequest(TestDescriptor rootTestDescriptor, EngineExecutionListen
...
+	 * Factory for creating an execution request.
+	 *
+	 * @param rootTestDescriptor the engine's root {@link TestDescriptor}
+	 * @param engineExecutionListener the {@link EngineExecutionListener} to be
+	 * notified of test execution events
+	 * @param configurationParameters {@link ConfigurationParameters} that the
+	 * engine may use to influence test execution
+	 * @return a new {@code ExecutionRequest}; never {@code null}
+	 */
+	public static ExecutionRequest create(TestDescriptor rootTestDescriptor,
```
sbrannen, line 57: Needs `@since` and `@API` declarations.

PR #2956, opened 2022-06-25
Allow scanning classpath concurrently

### LINKED ISSUE #2952: java.nio.file.FileSystemAlreadyExistsException during test discovery
When using JUnit 5 (and also 4) with version `1.8.2` for the launcher and `5.8.2` for `jupiter` I have been observing the presence of the following warning: `WARNING: Error scanning files for URI jar: ...` followed immediately by `java.nio.file.FileSystemAlreadyExistsException` emitted from the JUnit code `jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.newFileSystem(ZipFileSystemProvider.java:102)` at `org.junit.platform.commons.util.CloseablePath.createForJarFileSystem(CloseablePath.java:57)`. 

I am making a program to run a suite of JUnit tests across multiple threads, and using the JUnit launcher to perform test execution. I suspect this issue is caused by the JUnit `.exectue()` method running test discovery each time I invoke the method, which causes a race condition on the java filesystem object that JUnit creates. For reference, I believe that this class is not thread safe https://github.com/junit-team/junit5/blob/main/junit-platform-commons/src/main/java/org/junit/platform/commons/util/CloseablePath.java. I expect JUnit to produce no warnings when I execute tests across threads.  

Since JUnit no longer supports performing test discovery once for multiple executions (which is ideal for my use case), I am forced to perform repeated test discovery each time I invoke the JUnit launcher. Moreover, it seems that the decision to repeatedly do test discovery on each execution was made by JUnit recently https://github.com/junit-team/junit5/issues/2379. 

The environment I run my program in requires the tests to be in `.jar` files. When the tests are not in a `.jar` file the error coming from `ZipFileSystemProvider` does not happen. 

## Steps to reproduce

I created a simple program https://github.com/oscargolding/junit-load-test that shows the presence of the bug. 

To reproduce the warning clone the project then run the following commands.

First build the project and get the resulting `.jar` file. Then place the file into a relevant directory (I picked `/tmp/1/lib`). 
```
$ ./gradlew build
$ mkdir -p /tmp/1/lib
$ cp ./build/distributions/junit-load-tst-1.0-SNAPSHOT.jar /tmp/1/lib/
```

Then copy all the junit dependency `.jar` files (listed below) that are cached under the directory `~/.gradle/` into `/tmp/1/lib/`. 

```
junit-jupiter-api-5.8.2.jar
junit-jupiter-engine-5.8.2.jar
junit-platform-commons-1.8.2.jar
junit-platform-engine-1.8.2.jar
junit-platform-launcher-1.8.2.jar
junit-platform-runner-1.8.2.jar
junit-platform-suite-api-1.8.2.jar
junit-platform-suite-commons-1.8.2.jar
```

Finally run the commands below to reproduce the warning. 

```
$ cd /tmp/1
$ java -cp "./lib/*" "org.example.Main"
```

## Context

 - Used versions (Jupiter/Vintage/Platform): 5.8.2
 - Build Tool/IDE: Gradle, Java 11

## Deliverables

- [ ] ...

marcphilipp:
Resolves #2952.

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.9.0-M2.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.9.0-M2.adoc
@@ -16,6 +16,7 @@ on GitHub.
 ==== Bug Fixes
 
 * Absolute path entries are now supported in JUnit's Platform Console Launcher on Windows.
+* Fix concurrency issue in classpath scanning.
 
 ==== Deprecations and Breaking Changes
 
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/CloseablePath.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/CloseablePath.java
@@ -14,12 +14,16 @@ import static java.util.Collections.emptyMap;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 
 /**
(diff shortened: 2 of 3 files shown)


sormuras (approved):
Counting and working as expected.

Issue #2889, opened 2022-04-11
Set a custom extension for `@TempDir` via a configuration parameter


scordio:
Test suites using `@TempDir` usually pay a performance penalty due to the file system interaction. While this might be ok for a few tests, the penalty becomes more visible when the number of tests increases, like with AssertJ's [`Path`](https://github.com/assertj/assertj-core/tree/main/src/test/java/org/assertj/core/internal/paths) and [`File`](https://github.com/assertj/assertj-core/tree/main/src/test/java/org/assertj/core/internal/files) assertions tests.

Hitting the file system is probably not always needed, especially when working locally. An in-memory file system implementation might provide sufficient confidence and would run much faster, while the CI could verify the actual integration with a real file system.

For this reason, it would be nice if a custom extension could be set globally for the `@TempDir` annotation via a configuration parameter, similarly to the [Default Display Name Generator](https://junit.org/junit5/docs/current/user-guide/#writing-tests-display-name-generator-default). Also, I'm not sure if this might be a different way to address #2400.

Supporting a custom extension would allow the use of in-memory file systems like [memoryfilesystem](https://github.com/marschall/memoryfilesystem) or [jimfs](https://github.com/google/jimfs), and setting that with a configuration parameter would not require the rewriting of existing test classes and methods which already rely on `@TempDir`.

I tried alternative strategies based on inheritance or test interfaces but I couldn't come up with something really satisfying, hence this feature request. 

## Deliverables

- [ ] A custom extension for the `@TempDir` annotation can be set globally via a configuration parameter.

marcphilipp:
> Should I add a few lines in #2400 about this point?

Please do! 👍

Issue #2400, opened 2020-09-08
Support custom file system for `@TempDir`


nipafx:
Jupiter's `@TempDir` does not allow the configuration of custom file systems. It would be nice if it would.

A key motivation for me to open this issue is that custom file systems are the only thing that Pioneer's `@TempDir` can do that Jupiter's can't. Since Jupiter's is quite a bit more powerful than ours now, we'd like to remove our extension (this also prevents confusion for users), but we don't feel comfortable doing that as long as there's a feature we offer that you don't.

If Jupiter isn't fundamentally opposed to such a feature, we could implement it and open a PR.

marcphilipp:
**Team decision:** Since the `Extension` backing Jupiter's `@TempDir` is currently registered globally, implementing this is not straightforward. In addition, at this time, we have not yet received great interest in such a feature within Jupiter itself. In light of that, we'd like to focus our resources on some of the more popular outstanding issues in Jupiter. If it later turns out that there's significant community interest to include it in Jupiter we may revisit this issue at that time.

nicolaiparlog:
Me:

> If Jupiter isn't fundamentally opposed to such a feature, we could implement it and open a PR.

You:

> we'd like to focus our resources on some of the more popular outstanding issues in Jupiter

Not sure how they relate. :smiley: Should we (Pioneer) consider working on this to create a PR or would you prefer if we didn't?

marcphilipp:
Well, even if you put in the work to submit a PR we'd have to review it. Moreover, we would discuss how the API and implementation should roughly look like before that. So, for now, we'd prefer that you didn't. I hope you understand. 🙂

nicolaiparlog:
> I hope you understand.

Of course, no worries. Just wanted to make sure.

jbduncan:
I think this issue is valuable.

marcphilipp:
Now that `@TempDir` is created per declaration by default, that might make an approach that configures it locally feasible.

scordio:
Although the custom file system could/will be specified per declaration, it'd still be useful to have a way to set it globally, for example with a configuration property (rationale at https://github.com/junit-team/junit5/issues/2889#issue-1200641741).

scordio:
@marcphilipp @sbrannen is this up for grabs? I would like to work on it if that's fine with you.

Issue #2088, opened 2019-10-29
Provide support for setting a custom parent directory for `@TempDir`


rkrisztian:
My project requires that I do not write to `/tmp` or whichever location it is when I use the [`TempDirectory` extension](https://junit.org/junit5/docs/current/user-guide/#writing-tests-built-in-extensions-TempDirectory). So right now I have no choice but to still use [`junit-pioneer`'s implementation](https://junit-pioneer.org/docs/temp-directory/), e.g. like this:

 ```java
@RegisterExtension
Extension tempDirectory = TempDirectory.createInCustomDirectory(() -> Paths.get("build"));
 
@BeforeEach
void setup(@TempDir Path testProjectDir) {
    // Do something with "testProjectDir".
}
```

But it would be nicer if I could just use stock JUnit 5 for this. JUnit 4 supported setting the parent dir, so I find the JUnit 5 way a bit too limiting.

Thanks in advance.

**Comment by sbrannen** [comment]
Tentatively slated for 5.6 M2 solely for the purpose of _team discussion_

**Comment by marcphilipp** [comment]
If your project requires you not to write to the default location of temporary files, you should set the `java.io.tmpdir` system property.

@rkrisztian Could you please try that?

**Comment by rkrisztian** [comment]
Worked, thanks! I wish I knew about this. :)

**Comment by sbrannen** [comment]
Assigned to the general backlog and labeled as waiting for interest.

**Comment by rkrisztian** [comment]
FYI: A co-worker of mine suggested the use of `AbstractTask#getTemporaryDir()`:

```groovy
tasks.withType(Test) {
  systemProperty 'java.io.tmpdir', temporaryDir
}
```

dariocutillas:
I'm also interested in this functionality.

In my case I only want to affect the parent directory of one `@TempDir` and therefore it does not make sense to use `java.io.tmpdir`.

bes1002t:
in my case junit always set it tempDir to the working directory of the project. This is annoying because it always generates a mess into my source code directories.

Additionally this mess is not deleted after the tests were executed. I mean in this case this is good, because I don't want my source code directories to be deleted. Do you think I forget to configure something, or is this function as intended?

sbrannen:
@bes1002t, are you using the `@TempDir` support in JUnit Jupiter?

If so, then the behavior you are describing sounds like a bug, and we would appreciate it if you would open a separate issue describing your setup and providing a failing example.

bes1002t:
[...]

Issue #1786, opened 2019-02-18
Support @TempDir tags in extensions


sijie123:
I have a bunch of tests that require me to initialise a temporary directory with some files, so that the methods in my tests can read and write to it.
Currently, the solution is to implement a BeforeAll/BeforeEach tag and initialise the temporary directory for each test class.
As this is a common routine, it would be more reuse friendly if we could write an extension that encapsulates this routine, and which only needs to receive an input of a Path to initialise a temporary directory with all the files.
I envision the extension to look something like

```java
public class DirectoryInitExtension implements BeforeEachCallback {

    private Path TEST_DATA_FOLDER;

    @TempDir
    public Path tempDir;

    public static DirectoryInitExtension initializeWith(Path folder) {
        return new DirectoryInitExtension(folder);
    }

    @Override
    public void beforeEach(ExtensionContext extensionContext) throws IOException {
        // Copy files from TEST_DATA_FOLDER into tempDir, then return this tempDir as part of the test context.
    }
}
```
However, as I soon learnt (thanks to a fast response [here](https://stackoverflow.com/questions/54647577/how-to-use-tempdir-in-extensions)), TempDirs are not supported in extensions.

I wonder if it is a possibility to enable the same capabilities of TempDirs in extensions.

## Deliverables

- [ ] Support TempDir in extensions? (Sorry, not too sure what to write here)

sormuras:
Extensions should not depend on other extension - not even on those built-in extensions that Jupiter provides.

Why don't you just use `Files.createTempDirectory(...)` in `beforeEach` and delete that directory in a `afterEach` callback? The latter may profit from exposing the "delete directory tree recursively" helper by Jupiter in some way.

sormuras:
A generic resource supplier SPI (https://github.com/junit-team/junit5/pull/1672) could help here, methinks. With such an extension in place, test authors could write:

```java
@Test
void test4711(@New(TestDataFolder.class) Path path) { ... }
```

Will try to implement and publish a spike in JUnit Pioneer, soon.

sbrannen:
> Extensions should not depend on other extension - not even on those built-in extensions that Jupiter provides.

That's typically true; however, there are times when interactions between extensions is desirable. For example, the `SpringExtension` has a `public static ApplicationContext getApplicationContext(ExtensionContext)` method that is used internally by the `SpringExtension` but intentionally public for use by other extensions.

[...]

PR #2958, opened 2022-07-02
Add factory SPI for `@TempDir`

scordio:
## TL;DR

This could solve:

* #2088
* #2400
* #2889
* #3168

## Overview

This adds a factory SPI to `@TempDir`, allowing to define how the temporary directory is created.

My initial target was to support a custom file system (#2400), but it may also be a solution for the issues mentioned above.

The intended usage is to extend `TempDirFactory` and provide the class to the new `factory` attribute of `@TempDir`:

```java
...
import java.nio.file.Files;
...

class CustomFactory implements TempDirFactory {

    @Override
    public Path createTempDirectory(ExtensionContext context) throws IOException {
        return Files.createTempDirectory("junit");
    }

}
```

```java
@Test
void test(@TempDir(factory = CustomFactory.class) tempDir) {
    ...
}
```

#### Example with Custom Parent Directory

```java
...
import java.nio.file.Files;
...

class CustomParentFactory implements TempDirFactory {

    @Override
    public Path createTempDirectory(ExtensionContext context) throws IOException {
        return Files.createTempDirectory(Paths.get("/tmp-dir-root"), "junit");
    }

}
```

#### Example with Custom File System

Using [marschall/memoryfilesystem](https://github.com/marschall/memoryfilesystem):

```java
...
import java.nio.file.Files;
...

class MemoryFileSystemFactory implements TempDirFactory {

    private FileSystem fileSystem;

    @Override
    public Path createTempDirectory(ExtensionContext context) throws IOException {
        fileSystem = MemoryFileSystemBuilder.newEmpty().build();
        return Files.createTempDirectory(fileSystem.getPath("/"), "junit");
    }

    @Override
    public void close() throws IOException {
        fileSystem.close();
    }
}
```

Using [google/jimfs](https://github.com/google/jimfs):

```java
...
import java.nio.file.Files;
...

class JimfsFactory implements TempDirFactory {

    private FileSystem fileSystem;

    @Override
    public Path createTempDirectory(ExtensionContext context) throws IOException {
        FileSystem fileSystem = Jimfs.newFileSystem(Configuration.unix());
        return Files.createTempDirectory(fileSystem.getPath("/"), "junit");
    }

    @Override
    public void close() throws IOException {
        fileSystem.close();
    }
}
```

I would be happy to receive any feedback on the direction I have taken.

If this feature gets accepted, I would like to add a global configuration parameter like `junit.jupiter.tempdir.factory.default` in a separate PR.

### Open points
- [x] Behavior in case of `junit.jupiter.tempdir.scope=PER_CONTEXT`: the extension fails if a custom factory is set
- [x] Both in-memory file system libraries currently fail due to an unsupported `Path::toFile` call:
```
java.lang.UnsupportedOperationException: memory file system does not support #toFile()
	at com.github.marschall.memoryfilesystem.AbstractPath.toFile(AbstractPath.java:74)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath.resetPermissions(TempDirectory.java:353)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath.deleteAllFilesAndDirectories(TempDirectory.java:282)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath.close(TempDirectory.java:268)
```

```
java.lang.UnsupportedOperationException
	at com.google.common.jimfs.JimfsPath.toFile(JimfsPath.java:387)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath.resetPermissions(TempDirectory.java:353)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath.deleteAllFilesAndDirectories(TempDirectory.java:282)
	at org.junit.jupiter.engine.extension.TempDirectory$CloseablePath.close(TempDirectory.java:268)
```

To solve this, I catched the `UnsupportedOperationException` in `TempDirectory.CloseablePath::resetPermissions` and ignored it (b28450594c576d0433801fe73c0b9d377ea893a9). In support of this, the `Path::toFile` [javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#toFile--) mentions that the exception can be thrown if the `Path` is not associated with the default file system, which can be a potential use case once this SPI is offered.

---

---

### Definition of Done


scordio:
> Both in-memory file system libraries currently fail due to an unsupported `Path::toFile` call

After b28450594c576d0433801fe73c0b9d377ea893a9, both in-memory file system libraries work correctly.

lacasseio:
I like what you did with this PR. I wonder if passing `prefix` is necessary given that all new examples ignore it and the standard implementation always uses the same prefix.

I also see this PR as a good way to customize the behaviour of the temporary directory, such as including a space. We have a custom test directory provider that namespace the directory per-class/method, i.e. `test-files/Foo/myTestMethod`, and the idea in this PR could help with that. However, we would need access to some information from the `ExecutionContext`, mainly the test method and/or test class. I won't say to include this feature in this PR, but I think it adds additional use cases to the `TempDirFactory` design. I don't think passing the `ExecutionContext` instead of the `prefix` is wise, but maybe something like `TempDirContext` could double for the `prefix` (if required). What do you think?

Finally, I think having a way to pass a global `TempDirFactory` implementation via system property would be useful. Users could still override the factory per `@TempDir` declaration. Maybe something like `junit.jupiter.tempdir.factory=com.example.MyTempDirFactory`.

I hope this PR gets a be more traction because it provides the last bit of customization that `TempDir` needs.

scordio:
> I like what you did with this PR. 

@lacasseio thanks for the feedback! I have to admit I've put this topic a bit aside lately, there are still some points to clarify and test coverage to improve... I'll get it finalized sooner or later 🙂 

> I wonder if passing prefix is necessary given that all new examples ignore it and the standard implementation always uses the same prefix.

Indeed, the prefix doesn't seem so useful right now but I'd let the JUnit team judge what to do there.

> I also see this PR as a good way to customize the behaviour of the temporary directory, such as including a space. We have a custom test directory provider that namespace the directory per-class/method, i.e. `test-files/Foo/myTestMethod`, and the idea in this PR could help with that. However, we would need access to some information from the `ExecutionContext`, mainly the test method and/or test class. I won't say to include this feature in this PR, but I think it adds additional use cases to the `TempDirFactory` design. I don't think passing the `ExecutionContext` instead of the `prefix` is wise, but maybe something like `TempDirContext` could double for the `prefix` (if required). What do you think?

I agree `TempDirFactory` could have more input/control but I didn't feel the silver bullet yet, so I preferred starting with the smallest scope possible and for sure it could be extended with concrete use cases like yours. Especially here, I'd wait for some feedback from the core team before introducing something more complex.

> Finally, I think having a way to pass a global `TempDirFactory` implementation via system property would be useful. Users could still override the factory per `@TempDir` declaration. Maybe something like `junit.jupiter.tempdir.factory=com.example.MyTempDirFactory`.

That was exactly my idea, I've mentioned it in the middle of the description:

> If this feature gets accepted, I would like to add a global configuration parameter like `junit.jupiter.tempdir.factory` in a separate PR.

lacasseio:
Looking at implementing something similar as part of my current "TempDir" extension, I think it would be important to ensure  `Factory` can be `AutoCloseable` as, for example, `Jimfs` are `AutoCloseable` which signals that the file system should be closed once created (potentially after a test).

scordio:
Thanks for your feedback, @lacasseio! I need to update this topic and will check how to apply what you mentioned, as what the factory produces is already wrapped in a `CloseablePath`.

lacasseio:
Great, I just noticed the example didn't close the FileSystem object and though about mentioning it for completeness. You are right that it won't change the current architecture.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirFactory.java
```diff
@@ -0,0 +1,94 @@
...
+
+	/**
+	 * Return the {@code TempDirFactory} instance corresponding to the
+	 * given {@code Class}.
+	 *
+	 * @param factoryClass the factory's {@code Class}; never {@code null},
+	 * has to be a {@code TempDirFactory} implementation
+	 * @return a {@code TempDirFactory} implementation instance
+	 */
+	static TempDirFactory getTempDirFactory(Class<?> factoryClass) {
```
marcphilipp, line 84: Please move this to `TempDirectory` since it's an implementation detail.

marcphilipp:
**Team Decision:** Check whether `TempDirFactory` implements `AutoCloseable` and, if so, close it in `CloseablePath` after deleting the temp dir.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirFactory.java
scordio: Here I mimicked what `DisplayNameGenerator` has today:

https://github.com/junit-team/junit5/blob/ac8a5bd6c70244a0a5ba8073469452d20af3f671/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java#L365-L382

Happy to take a different direction.

olamy:
very good idea! This would be fantastic to have this available soon.
A global configuration would be good as well such `junit.jupiter.execution.tmpdir.factory` or any similar name.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirFactory.java
```diff
@@ -0,0 +1,94 @@
...
+	/**
+	 * Create a new temporary directory, using the given prefix to generate its name.
+	 * Depending on the implementation, the resulting {@code Path} may or may not be
+	 * associated with the default {@code FileSystem}.
+	 *
+	 * @param prefix the prefix string that can be used in generating the directory's name; never {@code null} or blank
+	 * @return the path to the newly created directory that did not exist before this method was invoked; never {@code null}
+	 * @throws Exception in case of failures
+	 */
+	Path createTempDirectory(String prefix) throws Exception;
```
olamy, line 54: could we have something such the current test class name in order to create directory depending on the name.
Use case is to debug content of created temp directories and/or files in case of test failures.

olamy: @scordio do you need a PR for this enhancement

scordio: Hi @olamy, I'm slowly catching up with the open points I have in this PR.

I'm ok with your proposal. I would also wait for feedback from the team before going in this direction as it requires a bit of refactoring to get the context as a method parameter. Once we get it, I'm happy to handle those changes.

scordio: Triggered by https://github.com/junit-team/junit5/pull/2958#issuecomment-145341381, I replaced the `prefix` parameter of `createTempDirectory` with the `ExtensionContext`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -250,6 +281,19 @@ public void close() throws IOException {
...
 			}
+
+			closeFactory();
+		}
+
+		private void closeFactory() throws IOException {
+			try {
+				if (factory instanceof AutoCloseable) {
+					((AutoCloseable) factory).close();
+				}
```
scordio, line 292: Would it make sense for `TempDirFactory` to extend `AutoCloseable` with an empty default implementation of `close()`?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -330,12 +380,16 @@ private void resetPermissionsAndTryToDeleteAgain(Path path, IOException exceptio
...
+		private static void tryToResetPermissions(Path path) {
+			try {
+				File file = path.toFile();
+				file.setReadable(true);
+				file.setWritable(true);
+				if (Files.isDirectory(path)) {
+					file.setExecutable(true);
+				}
+			}
+			catch (UnsupportedOperationException ignore) {
```
scordio, line 392: I couldn't find an easy way to test this change.

I tried to create a test with an anonymous subtype of `Path` where `toFile` throws an `UnsupportedOperationException` and all the other methods are delegated to a `Path` created with `Files.createTempDirectory`, but the subtype causes failures before reaching this part of the code where `sun.nio.fs.UnixPath` is expected (I'm on Mac OS).

Any suggestions or should I leave this uncovered?

Another option could be to have one of the in-memory file system libraries as a test dependency and use it to cover this change. Would that make sense?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirFactory.java
```diff
@@ -0,0 +1,77 @@
...
+@API(status = EXPERIMENTAL, since = "5.10")
+public interface TempDirFactory {
+
+	/**
+	 * Create a new temporary directory, using the given prefix to generate its name.
+	 * Depending on the implementation, the resulting {@code Path} may or may not be
+	 * associated with the default {@code FileSystem}.
+	 *
+	 * @param context the current extension context; never {@code null}
+	 * @return the path to the newly created directory that did not exist before this method was invoked; never {@code null}
```
scordio, line 50: Similar to #2969, should we allow returning `null` for falling back to the default factory?

scordio:
@marcphilipp @sormuras may I ask for an initial review?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -332,12 +380,16 @@ private void resetPermissionsAndTryToDeleteAgain(Path path, IOException exceptio
...
 			}
```
marcphilipp, line 393: I think the `try-catch` block should only be around `path.toFile()` and we should add a comment to the catch block explaining the well-known cases when this happens.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -187,13 +209,22 @@ private void assertSupportedType(String target, Class<?> type) {
...
-	private Object getPathOrFile(AnnotatedElement sourceElement, Class<?> type, CleanupMode cleanupMode,
-			ExtensionContext extensionContext) {
-		Namespace namespace = getScope(extensionContext) == Scope.PER_DECLARATION //
+	@SuppressWarnings("deprecation")
+	private Object getPathOrFile(AnnotatedElement sourceElement, Class<?> type, TempDirFactory factory,
+			CleanupMode cleanupMode, ExtensionContext extensionContext) {
+		Scope scope = getScope(extensionContext);
+
+		if (scope == Scope.PER_CONTEXT && factory != TempDirFactory.Standard.INSTANCE) {
+			throw new ExtensionConfigurationException("Custom @TempDir factory is not supported with "
```
marcphilipp, line 218: Why is it not supported with that scope?

marcphilipp (changes requested):
Looks very promising! Could you please add an integration test using jimfs or memoryfilesystem by introducing an additional test dependency?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
scordio: It's basically connected to your comment at https://github.com/junit-team/junit5/issues/2400#issuecomment-1096279042 (I guess).

For example, if `@TempDir` is declared on two fields in the same test class but each annotation has its own factory, I assume the PER_CONTEXT scope would provide the same instance to both, making one of the factories useless. I haven't demonstrated this with a test case yet, but it's something I can try 👍 

If my guess is correct, a local factory configuration should be disallowed due to unpredictable behavior, but a global configuration via the `junit.jupiter.tempdir.factory.default` configuration property (not part of this PR but I'm already working on the changes) could be allowed.

Or do you see it differently?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
scordio: Done in 24b6a0b.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
scordio: Solved in 24b6a0b with integration tests for both MemoryFileSystem and Jimfs.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
scordio: Triggered by the integration tests added in 24b6a0b5, I went ahead with this change. Please, let me know if you see anything wrong with it.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirFactory.java
```diff
@@ -0,0 +1,85 @@
...
+	 * @throws Exception in case of failures
+	 */
+	Path createTempDirectory(ExtensionContext context) throws Exception;
+
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	default void close() throws IOException {
+	}
```
scordio, line 61: What's the policy for documenting default overridden methods? Is `{@inheritDoc}` enough, should I leave it empty, or should I write a dedicated Javadoc?

Looking around, I see that `ResourceLock::close` has no Javadoc defined:

https://github.com/junit-team/junit5/blob/6b12a22dd30bfcb9012be456895cad8df1ea9366/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java#L41-L44

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
scordio: @marcphilipp I was able to confirm my assumption via this example in `TempDirectoryPerContextTests` (without the restriction in `TempDirectory`):

multiple_temp-dir-factories

If you check the path injected into `tempDir2` (shown in gray by the debugger), it's the one created by `Factory1` which I wouldn't consider to be the expected behavior.

For this reason, I would keep the restriction. What do you think?

marcphilipp: Yes, that makes sense. 👍

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirFactory.java
marcphilipp: We don't have a documented guideline for this case. I think using `{@inheritDoc}` is fine here.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryPerDeclarationTests.java
```diff
@@ -1131,4 +1150,46 @@ public Path createTempDirectory(ExtensionContext context) throws Exception {
...
+				return Files.createTempDirectory(fileSystem.getPath("/"), "prefix");
+			}
+		}
+
+	}
+
+	static class FactoryWitJimfsTestCase {
+
+		@Test
+		void test(@TempDir(factory = Factory.class) Path tempDir) {
```
marcphilipp, line 1182: What should happen in the following case:
```java
@Test
void test(@TempDir(factory = Factory.class) Path a, @TempDir(factory = Factory.class) Path b)
```

Without specifying `factory`, two different paths from the same (default) file system will be injected. With `factory` they will be from different file systems. I wonder if we need to support that use case somehow. 🤔

marcphilipp (approved):

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryPerDeclarationTests.java
scordio: Not sure I understood your concern.

In case the same factory is specified on both annotations, how the underlying file system is handled should be an implementation detail of the factory.

Maybe you're referring to the fact that each annotation will instantiate a different factory. If something like the file system has to be shared, it should be declared as static (with additional care in the `close()` method).

Is that what you were thinking?

The Javadoc could maybe highlight such cases. What do you think?

marcphilipp: Yes, I was wondering if there's going to be a need to use the same `FileSystem`. However, one could even solve this by using the `ExtensionContext` and putting a `CloseableResource` in the root `Store`. For now, I don't think we have to document anything in particular given we have no real-world use case.

marcphilipp:
@scordio This PR is a great next step but I don't think it solves #1786, does it?

scordio:
> @scordio This PR is a great next step but I don't think it solves #1786, does it?

You're right, I guess I had something else in mind when I mentioned it. I take it out from the description.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirFactory.java
scordio: @marcphilipp What do you think of this open point? Should I leave it aside, to be added only if there is a concrete request?

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -2621,5 +2621,19 @@ The default cleanup mode is `ALWAYS`. You can use the
...
+
+Factories can be created by implementing `TempDirFactory`. The default implementation
+available in Jupiter delegates the directory creation to
+`java.nio.file.Files::createTempDirectory`, passing `junit` as the prefix string
+to be used in generating the directory's name.
+
+[source,java,indent=0]
+.A test class with a temporary directory created by a custom factory
+----
+include::{testDir}/example/TempDirectoryDemo.java[tags=user_guide_factory]
```
scordio, line 2638: Not sure if this is enough or if there should be more. For example, should I mention `close()` for managing underlying resources like a custom `FileSystem`?

Would it make sense to add an example with an in-memory file system?

(If yes, any preference between [google/jimfs](https://github.com/google/jimfs) and [marschall/memoryfilesystem](https://github.com/marschall/memoryfilesystem)?)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirFactory.java
marcphilipp: I don't think we should allow that.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
marcphilipp: Yes, I think that would make a better example. Maybe use jimfs because it seems more popular?

scordio: Done in d92f151.

scordio:
@marcphilipp I think I'm done with my changes, all the open points should be covered. Please, let me know if there is anything to adjust.

scordio:
> If this feature gets accepted, I would like to add a global configuration parameter like `junit.jupiter.tempdir.factory.default` in a separate PR.

Initial draft at https://github.com/scordio/junit5/compare/2400_temp_dir_factory...scordio:junit5:gh-2400_temp_dir_factory_config_property.

I'll raise a PR once this one is merged.

documentation/documentation.gradle.kts
```diff
@@ -54,6 +54,10 @@ dependencies {
 		because("ApiReportGenerator needs it")
 	}
 
+	testImplementation(libs.jimfs) {
+		because("Jimfs is used in src/main/test")
```
marcphilipp, line 58:
```suggestion
		because("Jimfs is used in src/test/java")
```

marcphilipp (approved):

documentation/documentation.gradle.kts
scordio: Whoops, good catch 😅 thanks!

marcphilipp:
@scordio Thanks a lot for the productive collaboration on this one! 🎉

PR #2966, opened 2022-07-05
Fix display name and improve assertions for `@TempDir` test case

scordio:
I guess this:

https://github.com/junit-team/junit5/blob/7416e56237d3eefc890d3b36164a1b4a22fd5941/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryPerDeclarationTests.java#L412-L419

was originally inspired by:

https://github.com/junit-team/junit5/blob/7416e56237d3eefc890d3b36164a1b4a22fd5941/junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/TempDirectoryPerContextTests.java#L786-L797

but the former is not really verifying what the display name says.

This is my attempt to fix it.


sbrannen:

Thanks

PR #2972, opened 2022-07-08
Introduce AssertionFailureBuilder

### LINKED ISSUE #2967: The class org.junit.jupiter.api.AssertionUtils and its methods should be public
For our own purposes, we have implemented additional assertion methods (`assert...(...)`) in a test utility class. Motivated by reusing existing functionality, this class makes use of the static methods of  `org.junit.jupiter.api.AssertionUtils`; mostly for throwing assertion errors and for formatting error messages. `AssertionUtils` and most of its methods, however, are package-protected. In order to make them visible to our class, we have put it in the same package `org.junit.jupiter.api`. Now that we are using Java's module system, we can no longer proceed this way since Java's module system does not allow to have the same package in more than one module. Therefore I would like to propose to make `org.junit.jupiter.api.AssertionUtils` public as well as all its methods that are currently package-protected.

Note that there are a couple of other utility classes in `org.junit.platform.commons.util` that are already public; though `org.junit.jupiter.api.AssertionUtils` is in `junit-jupiter-api` while `org.junit.platform.commons.util` is in `junit-platform-commons`. I'm not necessarily proposing to also relocate `AssertionUtils` into `junit-platform-commons`: But if that makes sense, it seems better. `junit-platform-commons` is already a compile dependency of `junit-jupiter-api`.

## Deliverables

- [ ] Pull request?

marcphilipp:
## Overview

Resolves #2967.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java
@@ -63,6 +63,7 @@ class AssertThrows {
 						.expected(expectedType) //
 						.actual(actualException.getClass()) //
 						.reason("Unexpected exception type thrown") //
+						.cause(actualException) //
 						.build();
 			}
 		}
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java
@@ -10,16 +10,14 @@
 
 package org.junit.jupiter.api;
 
-import static org.junit.jupiter.api.AssertionUtils.buildPrefix;
-import static org.junit.jupiter.api.AssertionUtils.format;
+import static java.lang.String.format;
+import static org.junit.jupiter.api.AssertionFailureBuilder.assertionFailure;
 import static org.junit.jupiter.api.AssertionUtils.getCanonicalName;
-import static org.junit.jupiter.api.AssertionUtils.nullSafeGet;
 
 import java.util.function.Supplier;
 
 import org.junit.jupiter.api.function.Executable;
 import org.junit.platform.commons.util.UnrecoverableExceptions;
-import org.opentest4j.AssertionFailedError;
 
 /**
  * {@code AssertThrows} is a collection of utility methods that support asserting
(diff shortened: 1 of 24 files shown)


junit-jupiter-engine/src/test/java/org/junit/jupiter/api/AssertInstanceOfAssertionsTests.java
```diff
@@ -93,7 +93,7 @@ private <T> void assertInstanceOfSucceeds(Class<T> expectedType, Object actualVa
 
 	private void assertInstanceOfFails(Class<?> expectedType, Object actualValue, String unexpectedSort) {
 		String valueType = actualValue == null ? "null" : actualValue.getClass().getCanonicalName();
-		String expectedMessage = String.format("Unexpected %s ==> expected: <%s> but was: <%s>", unexpectedSort,
+		String expectedMessage = String.format("Unexpected %s, expected: <%s> but was: <%s>", unexpectedSort,
```
sbrannen, line 96: Please don't change the prefix delimiter to a comma.

This has been in place since 5.0 GA, and I don't think we should be changing the formatting this late in the game.

sbrannen: Never mind. As you pointed out elsewhere, this is not a change of the prefix delimiter.

sbrannen (approved):

mamadaliev (approved):

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrowsExactly.java
```diff
@@ -60,15 +58,19 @@ private static <T extends Throwable> T assertThrowsExactly(Class<T> expectedType
...
 				UnrecoverableExceptions.rethrowIfUnrecoverable(actualException);
-				String message = buildPrefix(nullSafeGet(messageOrSupplier))
-						+ format(expectedType, actualException.getClass(), "Unexpected exception type thrown");
-				throw new AssertionFailedError(message, actualException);
+				throw assertionFailure() //
+						.message(messageOrSupplier) //
+						.expected(expectedType) //
+						.actual(actualException.getClass()) //
+						.reason("Unexpected exception type thrown") //
+						.build();
```
bechte, line 66: Shouldn't `actualException` be handed over using `cause(actualException)` or was it intended to omit it here?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrows.java
```diff
@@ -60,15 +58,18 @@ private static <T extends Throwable> T assertThrows(Class<T> expectedType, Execu
...
 				UnrecoverableExceptions.rethrowIfUnrecoverable(actualException);
-				String message = buildPrefix(nullSafeGet(messageOrSupplier))
-						+ format(expectedType, actualException.getClass(), "Unexpected exception type thrown");
-				throw new AssertionFailedError(message, actualException);
+				throw assertionFailure() //
+						.message(messageOrSupplier) //
+						.expected(expectedType) //
+						.actual(actualException.getClass()) //
+						.reason("Unexpected exception type thrown") //
+						.build();
```
bechte, line 66: Shouldn't `actualException` be handed over using `cause(actualException)` or was it intended to omit it here?

bechte (commented):
I added two questions, most likely the missing calls to `cause` should be added. Otherwise, I approve this PR.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertThrowsExactly.java
marcphilipp: Good catch! Fixed in 1e7da1152253a289c963c1d234f3494ffe172d34.

Issue #3051, opened 2022-10-04
`EngineDiscoveryResultValidator` should produce an actionable error message for cyclic TestDescriptor structures


tribbloid:
This is the root cause of the problem reported in this ticket:

https://github.com/helmethair-co/scalatest-junit-runner/issues/83

Due to the flexibility of ScalaTest, A nested suite may cause several test cases to use the same ID, ScalaTest native runner can run it, but JUnit cannot. Instead it will throw a very brief error like this:

```
failed to execute tests

org.gradle.api.internal.tasks.testing.TestSuiteExecutionException: Could not complete execution for Gradle Test Executor 1.
	at org.gradle.api.internal.tasks.testing.SuiteTestClassProcessor.stop(SuiteTestClassProcessor.java:63)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.dispatch.ContextClassLoaderDispatch.dispatch(ContextClassLoaderDispatch.java:33)
	at org.gradle.internal.dispatch.ProxyDispatchAdapter$DispatchingInvocationHandler.invoke(ProxyDispatchAdapter.java:94)
	at com.sun.proxy.$Proxy2.stop(Unknown Source)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker$3.run(TestWorker.java:193)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.executeAndMaintainThreadName(TestWorker.java:129)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:100)
	at org.gradle.api.internal.tasks.testing.worker.TestWorker.execute(TestWorker.java:60)
	at org.gradle.process.internal.worker.child.ActionExecutionWorker.execute(ActionExecutionWorker.java:56)
[...]

PR #3052, opened 2022-10-05
Make error message for cyclic graphs in test discovery more actionable

tribbloid:
## Overview

#3051 error message now contains a list of detected cyclic test UIDs

---

---

### Definition of Done

getCyclicUIDs function will return a list containing the first UID that emits a cycle (can be changed to all UIDs, albeit slower)

may still need test case(s)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidator.java
```diff
@@ -37,30 +38,39 @@ void validate(TestEngine testEngine, TestDescriptor root) {
...
 	}
 
 	/**
 	 * @return {@code true} if the tree does <em>not</em> contain a cycle; else {@code false}.
 	 */
 	boolean isAcyclic(TestDescriptor root) {
+		return getCyclicUIDs(root).isEmpty();
+	}
+
+	ArrayList<UniqueId> getCyclicUIDs(TestDescriptor root) {
```
mpkorstanje, line 55: Consider using an `Optional` instead.

tribbloid: sure, totally make sense for current implementation, but again emitting all UIDs is still a tempting alternative

tribbloid: fixed now, thanks a lot!

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidator.java
```diff
@@ -37,30 +38,39 @@ void validate(TestEngine testEngine, TestDescriptor root) {
...
+			() -> String.format("The discover() method for TestEngine with ID '%s' returned a cyclic graph: "
+					+ "Unique ID(s) '%s' enter a cycle",
+				testEngine.getId(), cyclicUIDs));
 	}
 
 	/**
 	 * @return {@code true} if the tree does <em>not</em> contain a cycle; else {@code false}.
 	 */
 	boolean isAcyclic(TestDescriptor root) {
+		return !getFirstCyclicUID(root).isPresent();
```
mpkorstanje, line 52: Is this method still used by non-test code?

tribbloid:
Good to go. Can we merge and close it?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidator.java
```diff
@@ -37,30 +38,39 @@ void validate(TestEngine testEngine, TestDescriptor root) {
 			() -> String.format(
 				"The discover() method for TestEngine with ID '%s' must return a non-null root TestDescriptor.",
 				testEngine.getId()));
-		Preconditions.condition(isAcyclic(root),
-			() -> String.format("The discover() method for TestEngine with ID '%s' returned a cyclic graph.",
-				testEngine.getId()));
+		Optional<UniqueId> cyclicUIDs = getFirstCyclicUID(root);
+		Preconditions.condition(!cyclicUIDs.isPresent(),
+			() -> String.format("The discover() method for TestEngine with ID '%s' returned a cyclic graph: "
+					+ "Cycle found at unique ID '%s'",
```
marcphilipp, line 44: I think it would be even more helpful to print all unique IDs in the cycle or at least the two between which the cycle was detected (`detectInnerClassCycle` in `ReflectionUtils` could be used for inspiration).

@tribbloid Would you have time to look into that?

tribbloid: Absolutely, that's why I opted to use List instead of Optional in the first place.

I just don't know if it can halt every time, should we add a some kind of depth check?

marcphilipp: Here's an article outlining how to find the shortest cycle in an undirected graph (tests should be a tree, i.e. an acyclic undirected graph): https://www.baeldung.com/cs/shortest-cycle-undirected-graph

tribbloid: @marcphilipp thanks a lot, I'll add the code to print the first detected path of the cycle (using `node.parent()`).

But I'm reluctant to change the algorithm from breadth first (the status quo) to depth first (one used in your link), the search will hang for diverging tree without loop (e.g. if you generate a UUID suffix for each new sub-test). Existing code probably chose breadth first for this reason (and some other reasons)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidator.java
```diff
@@ -37,30 +38,39 @@ void validate(TestEngine testEngine, TestDescriptor root) {
...
 	/**
 	 * @return {@code true} if the tree does <em>not</em> contain a cycle; else {@code false}.
 	 */
 	boolean isAcyclic(TestDescriptor root) {
+		return !getFirstCyclicUID(root).isPresent();
+	}
+
+	Optional<UniqueId> getFirstCyclicUID(TestDescriptor root) {
 		Set<UniqueId> visited = new HashSet<>();
+		Optional<UniqueId> cyclic = Optional.empty();
```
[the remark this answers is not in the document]
tribbloid: it no longer exist in the latest version

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidator.java
tribbloid: No, and after the latest change it has now become getCyclicGraphInfo, which contains a short visualisation of the cycle:

```
				"Test [engine:root] exists in at least two paths:\n" +
						"\t[engine:root]\n" +
						"\t[engine:root] <- [engine:root]"
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineDiscoveryResultValidator.java
tribbloid: OK the latest  version now print 2 paths that leads to the same node. There are some strange styling errors which shouldn't affect implementation.

All 3 test cases now contains examples of the error messages

tribbloid: LGTM. Let me know if it looks good or only requires minor revision.

tribbloid: > Now the styling error is also fixed. Only get some deprecation error under JVM 20 that is irrelevant to the scope of this change.

Can we close it now?

mpkorstanje:
This has been made obsolete by https://github.com/junit-team/junit5/pull/3116.

tribbloid:
@marcphilipp I don't see any objection on this patch, can we merge and close it any time soon? It has already become stale once

marcphilipp:
@tribbloid Sorry for the delay! We've deactivated Stalebot for PRs now. This PR needs a proper review from someone on the team.

marcphilipp (approved):

marcphilipp:
@tribbloid Thank you! 👍

Issue #3028, opened 2022-09-09
Introduce extension API for customizing the ClassLoader in Jupiter


ledoyen:
Following #201 and #3014, this issue will focus on creating a new extension API in JUnit Jupiter that allows for `ClassLoader` customization.

This extension should be able to work in conjunction with others, such as `@TestTemplate`.

How I see it designed in JUnit Jupiter:

* a new interface `ClassLoaderModifier` with a unique method `ClassLoader modify(ClassLoader current)`
    * the documentation should precise that the returned _class loader_ should have the same parent as the given one

## What needs to be checked :

* Is this model sufficient for other class loading modifying tools (powermocks, robolectric, etc.) ?

# 🙏  Please do not troll, ask for due date or complain on this ticket, the original ticket is cluttered with these unhelpful comments and is really painful to read when it comes to what features should be implemented and how

marcphilipp:
While we might have to add some hooks to the platform to make this happen, I think we should focus this initial issue on providing a Jupiter extension like the one you have in [fridujo/classpath-junit-extension](https://github.com/fridujo/classpath-junit-extension).

@ledoyen Could you please change the issue description accordingly?

ledoyen:
I updated the description as asked.

@marcphilipp Do not hesitate to modify it yourself, if you want to improve it or even change the objective / direction.

geoand:
To expand a little on what is mentioned in https://github.com/junit-team/junit5/issues/2579, Quarkus has to use a hack where the test methods are intercepted and "redirected" to a different test instance that is loaded and instantiated from a custom ClassLoader. 
The reason this is necessary is because Quarkus performs various transformations on application classes which would otherwise be lost.
This mostly works, but does create problems, most importantly with method arguments which also need to be loaded from the same ClassLoader.

If we had an API where we could specify the ClassLoader to be used for loading the test, I am pretty sure all these problems would go away.

realdadfish:
Cc @robolectric/robolectric (hope this works)

atsiporu:
Hello all,
I would like to refer to my comment in a related issue that presents my usecause.
https://github.com/junit-team/junit5/issues/201#issuecomment-1308040725

Here is the body of the comment:

[...]

Issue #2579, opened 2021-03-23
Add ability for InvocationInterceptor to intercept parameter suppliers


stuartwdouglas:
Quarkus needs to run tests in its own ClassLoader, and it achieves this by intercepting all methods through InvocationInterceptor and redirecting them to a new test instance loaded from the Quarkus ClassLoader.
 
This works really well with the exception of @MethodSource based parameters. As there is no way to intercept this the method is invoked on the original instance, and the parameters may be loaded from the wrong ClassLoader.

It would be great if we could have some additional methods added to InvocationInterceptor to intercept this invocation as well.

juliette-derancourt:
Tentatively slated for 5.8-M2 solely for the purpose of _team discussion_

sbrannen:
One of the challenges here is that `MethodArgumentsProvider` (which invokes the `@MethodSource` factory method on the test instance) is an implementation of `org.junit.jupiter.params.provider.ArgumentsProvider` which is **not** an `Extension` API in `junit-jupiter-api`.

Whereas, `InvocationInterceptor` **is** an `Extension` API in `junit-jupiter-api`.

Thus, `InvocationInterceptor` cannot know about `ArgumentsProvider`.

marcphilipp:
We could add a new `interceptExtensionMethod(Invocation<T> invocation, ReflectiveInvocationContext<Method> invocationContext, ExtensionContext extensionContext)` method to `InvocationInterceptor`, and add a `MethodInvoker getMethodInvoker()` to `ExtensionContext` with `interface MethodInvoker { Object invokeMethod(Method method, Object target, Object... args); }`. Then, `MethodArgumentsProvider` could use `MethodInvoker` to call the configured method while `InvocationInterceptor` would be able to do what it needs.

sbrannen:
@marcphilipp, your proposal is somewhat related to #2191.

If we implement a solution for #2191, that would likely result in a means for invoking any method using parameter resolvers, analogous to what `org.junit.jupiter.engine.execution.ExecutableInvoker` does internally.

So there may be room for synergies here.

sbrannen:
Also related to #1139.

morhei:
Ran into this issue today.

marcphilipp:
I think we should address #3028 first as it might solve the problem more cleanly.

@stuartwdouglas Could you please point us to Quarkus' `InvocationInterceptor` implementation for reference?

geoand:
[...]

PR #3091, opened 2022-11-14
Introduce `LauncherInterceptor` API

marcphilipp:
## Overview

In order to support use cases where test discovery and execution needs to be wrapped by third-party code, e.g. for using the Eclipse OSGi runtime this PR introduces `LauncherInterceptor` which can be registered using Java's ServiceLoader mechanism. The feature is opt-in via a new configuration parameter than can be set as a system property in order to avoid accidental usage via transitive dependencies.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/LauncherInterceptor.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/LauncherInterceptor.java
@@ -11,9 +11,6 @@
 package org.junit.platform.launcher;
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
-import static org.apiguardian.api.API.Status.INTERNAL;
-
-import java.util.List;
 
 import org.apiguardian.api.API;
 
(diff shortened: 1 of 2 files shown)


sormuras:
Nice work, Marc! Will take a closer look later this week.

I already see a way to supersed this project, or at least make [sormuras/junit-platform-isolator](https://github.com/sormuras/junit-platform-isolator) use this new feature. 🤔

jbduncan:
Interesting! I'm guessing this feature could also be used to run logic that needs to run only once per test suite invocation?

marcphilipp:
Yes, but `LauncherSessionListener` might be a better fit for that.

holly-cummins:
Hi @marcphilipp, thanks for this. I'm having a play with the new capability for the Quarkus use case (as discussed in https://github.com/junit-team/junit5/issues/2579 and https://github.com/junit-team/junit5/issues/3028). Using sophisticated debug (system out printlns), I can see that the launcher interceptor is being invoked, and is calling ` currentThread.setContextClassLoader` in the `intercept` method. 

I have an extension which implements `InvocationInterceptor`. In its `interceptBeforeAllMethod`, I check the classloader of the test class with `extensionContext.getRequiredTestClass().getClassLoader(). I would expect it to be the classloader I set on the TCCL in the interceptor, but it's not. Similarly, if I check `invocationContext.getExecutable().getParameters()` and look at their classloaders, they're the same as they would be if my interceptor wasn't there. 

Am I doing something wrong, or was I just over-optimistic about how much would be changed by doing an interception? I do hope it's "I am doing something wrong," because we've been quite excited about  in the Quarkus team about how this would allow us to eliminate some classloader hacks.

holly-cummins:
Actually, it's an `interceptTestMethod` override where I'm checking the test class's classloader.

PR #3123, opened 2023-01-03
Make inner class static to avoid code smell

MartinWitt:
# Repairing Code Style Issues

## InnerClassMayBeStatic
Inner classes that do not reference their enclosing instances can be made static.
This prevents a common cause of memory leaks and uses less memory per instance of the class.

# Repairing Code Style Issues
* InnerClassMayBeStatic (1)


marcphilipp (approved):

marcphilipp:
@MartinWitt Thanks for the PR! 👍

PR #3143, opened 2023-02-05
Model parameters using org.gradlex.build-parameters

britter:
Introduces the org.gradlex.build-parameters plugin to model parameters
that can be supplied to this build. For that an included build is added
that contains the parameter configurations. It's necessary to put this
into an included build instead of buildSrc because buildSrc can not
build plugins that can be applied to the settings script.

The details on how to conncet to the remote build cache as well as the
enablement of test distribution are then read from the
BuildParametersExtension in settings.gradle.kts. Injection of the
parameters on CI now uses auto-generated environment variable look up
provided by the build parameters plugin.

The next step would be to also model parameters used in other parts of
the build logic using the build parameters plugin. This however requires
a larger refactoring of buildSrc because buildSrc can only depend on
included build modules that are included via Settings.includeBuild, but
we need to include the build-logic build via
pluginManagement.includeBuild due to the reasons stated in the
beginning.

### DIFF
--- a/build-logic/build-parameters/build.gradle.kts
+++ b/build-logic/build-parameters/build.gradle.kts
@@ -1,11 +1,10 @@
 plugins {
-	id("org.gradlex.build-parameters") version "1.4"
+	id("org.gradlex.build-parameters") version "1.4.1"
 }
 
 group = "org.junit.gradle"
 
 buildParameters {
-	enableValidation.set(false) // see https://github.com/gradlex-org/build-parameters/issues/80
 	bool("ci") {
 		description.set("Whether or not this build is running a CI environment")
 		defaultValue.set(false)
--- a/build-logic/build-parameters/build.gradle.kts
+++ b/build-logic/build-parameters/build.gradle.kts
@@ -15,7 +15,7 @@ buildParameters {
 			description.set("Username to authenticate with the remote build cache")
 			fromEnvironment()
 		}
-		string("Password") {
+		string("password") {
 			description.set("Password to authenticate with the remote build cache")
 			fromEnvironment()
 		}
--- a/build-logic/build-parameters/build.gradle.kts
+++ b/build-logic/build-parameters/build.gradle.kts
@@ -1,0 +1,36 @@
+plugins {
+	id("org.gradlex.build-parameters") version "1.4"
+}
+
+group = "org.junit.gradle"
+
+buildParameters {
+	enableValidation.set(false) // see https://github.com/gradlex-org/build-parameters/issues/80
+	bool("ci") {
+		description.set("Whether or not this build is running a CI environment")
+		defaultValue.set(false)
+		fromEnvironment()
+	}
+	group("buildCache") {
+		string("username") {
+			description.set("Username to authenticate with the remote build cache")
+			fromEnvironment()
+		}
+		string("Password") {
+			description.set("Password to authenticate with the remote build cache")
+			fromEnvironment()
+		}
+		string("url") {
+			description.set("URL to the remote build cache")
+			fromEnvironment()
+		}
+	}
+	group("enterprise") {
+		description.set("Parameters controlling Gradle Enterprise features")
+		bool("enableTestDistribution") {
+			description.set("Whether or not to use Test Distribution for executing tests")
+			defaultValue.set(false)
+			fromEnvironment()
+		}
+	}
+}
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,6 +1,8 @@
+import buildparameters.BuildParametersExtension
 import com.gradle.enterprise.gradleplugin.internal.extension.BuildScanExtensionWithHiddenFeatures
 
 pluginManagement {
+	includeBuild("build-logic")
 	repositories {
 		gradlePluginPortal()
 	}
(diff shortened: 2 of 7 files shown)


britter:
This is how the help looks like for the configuration we have in place now: https://ge.junit.org/s/7ahdtkfogcwdi/console-log?page=1#L22

sormuras:
> Introduces the `org.gradlex.build-parameters` plugin to model parameters that can be supplied to this build.

To "model" or to "list"/"show"? What does "model" mean in this context?

settings.gradle.kts
```diff
@@ -57,33 +56,30 @@ gradleEnterprise {
...
 				hostname { null }
 				ipAddresses { emptyList() }
 			}
 		}
 
-		val enableTestDistribution = providers.gradleProperty("enableTestDistribution")
-			.map(String::toBoolean)
-			.getOrElse(false)
-		if (enableTestDistribution) {
+		if (buildParameters.enterprise.enableTestDistribution) {
```
marcphilipp, line 67: ❌ I'm afraid this isn't the only place this parameter is used. It's also used in `buildSrc`:
https://github.com/junit-team/junit5/blob/fe1997419d4fd5b0424a5e992a26f1826d2250c2/buildSrc/src/main/kotlin/testing-conventions.gradle.kts#L24

marcphilipp (commented):
> The next step would be to also model parameters used in other parts of
the build logic using the build parameters plugin. This however requires
a larger refactoring of buildSrc because buildSrc can only depend on
included build modules that are included via Settings.includeBuild, but
we need to include the build-logic build via
pluginManagement.includeBuild due to the reasons stated in the
beginning.

Can other build logic plugins use parameters modeled this way? Would migrating off of `buildSrc` and changing moving the plugins it contains to an included build work?

marcphilipp:
The build fails with the following error message. Does the plugin require all parameters to be modeled?

```

* Where:
Settings file '/home/runner/work/junit5/junit5/settings.gradle.kts' line: 24

* What went wrong:
An exception occurred applying plugin request [id: 'build-parameters']
> Failed to apply plugin 'build-parameters'.
   > Unknown build parameter: enableJaCoCo
```

build-logic/build-parameters/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+		description.set("Whether or not this build is running a CI environment")
+		defaultValue.set(false)
+		fromEnvironment()
+	}
+	group("buildCache") {
+		string("username") {
+			description.set("Username to authenticate with the remote build cache")
+			fromEnvironment()
+		}
+		string("Password") {
```
marcphilipp, line 18: ❔ Does lower/upper case matter here?

britter: Yes, it is. This should be lowercase.
```suggestion
		string("password") {
```

britter:
@sormuras 

> To "model" or to "list"/"show"? What does "model" mean in this context?

To model in this context means, that the `org.gradlex.build-parmeters` plugin provides a DSL for describing parameters that can be supplied to a build. Based on that description it generates compile-safe accessors that build authors can use to access these parameters. We believe that this is a much better way of dealing with parameters than Gradle's built-in features, e.g. reading from extra properties or `project.findProperty`.

Another benefit of using the plugin is the `:parameters` tasks that works much like that tasks task, but for build parameters. It will print out useful information about the available parameters, see https://ge.junit.org/s/7ahdtkfogcwdi/console-log?page=1#L22

@marcphilipp 

> Can other build logic plugins use parameters modeled this way? Would migrating off of buildSrc and changing moving the plugins it contains to an included build work?

Other plugins can use parameters modeled this way, but they have to be built from an included build as well. I tried to explain this in the description, apologies if it was unclear. I did not yet migrated all parameters, e.g. the ones used in `testing-convention`, because I wanted to have some feedback about the general approach before moving all of the build logic around.

So if this way of modeling build parameters is something you would like to integrate into the build, my proposal is that I create another PR the moves the build from buildSrc into an included build first. I would then rebase this change on top of that.

> The build fails with the following error message. Does the plugin require all parameters to be modeled?

The plugin fails the build if a parameter is supplied via the command line that is not modeled in the plugin. It's an opt-out feature that can be [disabled](https://gradlex.org/build-parameters/#disable-build-parameter-validation). We added this because we had build users in mind that are supplying parameters that contain a typo. We think this makes it easier for them to figure out that they had a mistake in the command line.

Unfortunately this is a little bit problematic because there are core plugins like signing and plugin publish that expect configuration values to be passed as project properties. I guess it would be better to opt-out of this at first and then sit down to discuss all use cases where a JUnit developer would supply -P parameters to a build. We are confident that all of this can always be correctly modeled using the build-parameters plugin.

settings.gradle.kts
britter: I'm aware of this but wanted to have feedback on the direction first before fixing this. See below :o)

marcphilipp:
> So if this way of modeling build parameters is something you would like to integrate into the build, my proposal is that I create another PR the moves the build from buildSrc into an included build first. I would then rebase this change on top of that.

If you have time for that, that'd be awesome!

britter:
He friends, I recorded a demo to give a better impression what this can do for you :)

[asciicast](https://asciinema.org/a/AgSsIS3lnHuOyTiDqSW9gx741)

britter:
Replaced by #3170

PR #3151, opened 2023-02-09
Replace buildSrc with an included build

### LINKED ISSUE #3149: Move build logic from `buildSrc` to an included build
Using buildSrc to manage build logic is semi-deprecated by the Gradle team because it has several issues, e.g. it's used as an input for all modules, so each change causes a full rebuild of the whole projects. The recommended way of organizing build logic nowadays is to use an included build instead. Here's the proposal for how to do it in the JUnit project:

- `buildSrc` directory is renamed to `gradle/plugins`.
- The new location is defined as an included build in `settings.gradle.kts` by adding an `includeBuild("gradle/plugins")` statement.
- All pre-compiled script plugins are renamed and prefixed with `junitbuild`. This will result in all plugins build from the build logic build having the group prefix `junitbuild`. For example, the `kotlin-library-conventions.gradle.kts` file will be renamed to `junitbuild.kotlin-library-conventions.gradle.kts`. When the plugin is applied the new plugin ID will be `junitbuild.kotlin-library-conventions`. This makes it easier to distinguish custom plugins from core plugins. I propose `junitbuild` instead of `org.junit.gradle` to also indicate that this code is not public facing or meant to be published to an external repository.
- If the team agrees with the above I would also change the root package for code in `gradle/plugins` from `org.junit.gradle` to `junitbuild`.

This is a prerequisite for delivering #3143.

## Deliverables

- [ ] `buildSrc` is migrated to an included build

britter:
Resolves #3149

## Overview

- moved `buildSrc` to `gradle/plugins`
- added `junitbuild` as group to all plugins
- moved build logic classes from `org.junit.gradle` package to `junitbuild` package

---

---

### Definition of Done


britter:
Looking at the errors on CI:

- The CodeQL job started failing when the Gradle wrapper was updated by @marcphilipp on main: https://github.com/junit-team/junit5/actions/workflows/codeql-analysis.yml
- The problem with CI / Windows I can't reproduce locally and unfortunately the build did not produce a build scan so it's hard to tell what's wrong.

.gitignore
```diff
@@ -1,7 +1,7 @@
 # Gradle
 .gradle
 /build/
-/*/build/
+/**/build/
```
marcphilipp, line 4: I think this can now be simplified to `build` which we couldn't use before because we had a `org.junit.gradle` package.

marcphilipp (approved):

marcphilipp:
> The CodeQL job started failing when the Gradle wrapper was updated by @marcphilipp on main: [junit-team/junit5/actions/workflows/codeql-analysis.yml](https://github.com/junit-team/junit5/actions/workflows/codeql-analysis.yml)

Yep, my bad. I've disabled that job for the time being in https://github.com/junit-team/junit5/commit/1711f1c7b9fca4d269c46046624e9b5b3136e861.

> The problem with CI / Windows I can't reproduce locally and unfortunately the build did not produce a build scan so it's hard to tell what's wrong.

The problem was caused by the `nohttp` plugin scanning the `gradle/plugins/.gradle` folder. For `buildSrc` it had a built-in exclusion:
https://github.com/spring-io/nohttp/blob/3fd247d3e909bfcec25bd005b4da28c384a51322/nohttp-gradle/src/main/java/io/spring/nohttp/gradle/NoHttpCheckstylePlugin.java#L107

I've add `**/.gradle/**` as an exclusion in the root `build.gradle.kts` script.

marcphilipp:
@britter Thanks a lot for your contribution! 👍

Issue #2063, opened 2019-10-14
Revise module declarations


sormuras:
Revise module declarations for wrong, unnecessary and superfluous directives and modifiers.

Related issue: #1939 

## Deliverables
- [x] Check if `transitive` is really needed for each `requires` directive that declares it
- [x] All internal packages are hidden, i.e. not exported
- [x] All modules are `required` explicitly when an exported package is used directly

sormuras:
Module `org.junit.platform.engine` does **not** export package `org.junit.platform.engine.support.filter`  - although it contains the published `ClasspathScanningSupport` class. That class, deprecated, is used by 3rd-party engines. Like: https://github.com/testng-team/testng-junit5/blob/master/src/main/java/org/testng/junit5/TestNGine.java#L3

We should export the package in the upcoming 5.6 release.

sormuras:
#### Is `transitive` needed for `requires transitive org.apiguardian.api;`?

Yes. E.g `org.apiguardian.api.API` is part of "all" JUnit 5 API and it is annotated with `@Retention(RUNTIME)` -- so it leaks to all downstream modules.

#### Is `transitive` needed for `requires transitive org.opentest4j;`?

Yes. E.g. `Assertions.assertAll(...) throws org.opentest4j.MultipleFailuresError`

sormuras:
### org.junit.jupiter
```
org.junit.jupiter (junit-jupiter-5.6.0-M1.jar)
  [Module descriptor]
    requires mandated java.base;
    requires transitive org.junit.jupiter.api;
    requires transitive org.junit.jupiter.engine;
    requires transitive org.junit.jupiter.params;
  [Suggested module descriptor for org.junit.jupiter]
    requires mandated java.base;
  [Transitive reduced graph for org.junit.jupiter]
    requires mandated java.base;
    requires org.junit.jupiter.engine;
    requires org.junit.jupiter.params;
```

sormuras:
### org.junit.jupiter.api

```
org.junit.jupiter.api (file:///home/runner/work/JDK-8225773/JDK-8225773/lib/junit-jupiter-api-5.6.0-M1.jar)
  [Module descriptor]
    requires mandated java.base;
    requires transitive org.apiguardian.api (@1.1.0);
    requires transitive org.junit.platform.commons;
    requires transitive org.opentest4j (@1.2.0);
  [Suggested module descriptor for org.junit.jupiter.api]
    requires mandated java.base;
[...]

PR #3193, opened 2023-03-15
Drop transitive dependency on apiguardian

mkoncek:
## Overview
Annotations should stay as `static` dependencies and not `transitive` this was explained in https://mail.openjdk.org/pipermail/jigsaw-dev/2021-June/014674.html.

Currently the problem we are facing is that when building `assertj-core`, we need to have `apiguardian` as well because `assertj-core` requires `junit5` which transitively requires `apiguardian` even though `assertj-core` does not use `apiguardian` at all.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/module/org.junit.jupiter.api/module-info.java
+++ b/junit-jupiter-api/src/module/org.junit.jupiter.api/module-info.java
@@ -12,7 +12,7 @@
  * Defines JUnit Jupiter API for writing tests.
  */
 module org.junit.jupiter.api {
-	requires static transitive org.apiguardian.api;
+	requires static org.apiguardian.api;
 	requires transitive org.junit.platform.commons;
 	requires transitive org.opentest4j;
 
--- a/junit-jupiter-migrationsupport/src/module/org.junit.jupiter.migrationsupport/module-info.java
+++ b/junit-jupiter-migrationsupport/src/module/org.junit.jupiter.migrationsupport/module-info.java
@@ -15,7 +15,7 @@
  */
 module org.junit.jupiter.migrationsupport {
 	requires transitive junit; // 4
-	requires static transitive org.apiguardian.api;
+	requires static org.apiguardian.api;
 	requires transitive org.junit.jupiter.api;
 	requires org.junit.platform.commons;
 
--- a/junit-jupiter-params/src/module/org.junit.jupiter.params/module-info.java
+++ b/junit-jupiter-params/src/module/org.junit.jupiter.params/module-info.java
@@ -14,7 +14,7 @@
  * @since 5.0
  */
 module org.junit.jupiter.params {
-	requires static transitive org.apiguardian.api;
+	requires static org.apiguardian.api;
 	requires transitive org.junit.jupiter.api;
 	requires transitive org.junit.platform.commons;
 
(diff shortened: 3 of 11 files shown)


sormuras:
Dropping `transitive` from:
```java
	requires java.logging;
	requires java.management; // needed by RuntimeUtils to determine input arguments
	requires static transitive org.apiguardian.api;
	// ...
}
```
generates warnings like:
```
/home/runner/work/junit5/junit5
  /junit-platform-commons/build/module/org.junit.platform.commons
  /org/junit/platform/commons/annotation/Testable.java:79:
warning: [exports] class API in module org.apiguardian.api is not indirectly exported using requires transitive
 ^
```
Suppressing those 30+ warning doesn't seem to be the correct, neither. 🤔

sormuras:
We are talking about module `org.assertj.core`, right?

- https://github.com/assertj/assertj/blob/main/assertj-core/src/main/java9/module-info.java

```java
module org.assertj.core {
  // AssertJ Core's package API
  exports org.assertj.core.annotations;
  //  [...]
  exports org.assertj.core.util.xml;

  requires static java.logging; // required when printThreadDump is true
  requires static java.management;
  requires static java.xml; // used for XML pretty print
  requires static junit;
  requires static net.bytebuddy;
  requires static org.hamcrest;
  requires static org.junit.jupiter.api;
  requires static org.opentest4j; // to throw AssertionFailedError which is IDE friendly

  // Services loaded by org.assertj.core.configuration.ConfigurationProvider
  uses org.assertj.core.configuration.Configuration;
  uses org.assertj.core.presentation.Representation;
}
```

From `requires static junit;` follows:
- `requires static org.hamcrest;`

From `requires static org.junit.jupiter.api;` follows:
- `requires static org.opentest4j;`
- `requires static org.apiguardian.api;`

mkoncek:
Yes, `org.assertj.core`. What you said is true but it shouldn't be that way. Build dependencies are not transitive and `apiguardian` should not be explicitly `transitive`.
Also I know you talked about this in https://github.com/junit-team/junit5/issues/2063. But I would still put more weight on openjdk folks.

sormuras:
How unfortunate it is, `org.apiguardian.api` is a compile time and run-time dependency of `org.junit.jupiter.api` and many (all?) other modules published by the JUnit team. Including modules of the JUnit Platform, JUnit Jupiter, and JUnit Vintage.

This is due to the fact that we added an `enum`, namely `API.Status`, to the public API of the `org.apiguardian.api.API` annotation. If we'd chose type `String` for the `status()` accessor back then, we weren't be in such "transitive troubles" as we are now, IIRC.

> Also I know you talked about this in https://github.com/junit-team/junit5/issues/2063.

Great. I knew I wrote something similar before. Technically, nothing changed since 2019.

> But I would still put more weight on openjdk folks.

Done. 🏋️

marcphilipp:
In the context of https://github.com/apiguardian-team/apiguardian/issues/29, I've just verified that `transitive` is still required to avoid the [compiler warning](https://github.com/junit-team/junit5/pull/3193#issuecomment-1470151055). I agree that it would be great if the Java compiler could be made to not emit it for classes only referenced from annotations. For the time being, we have to assume that's not going to happen. Therefore, I'm closing this PR.

Issue #3205, opened 2023-03-24
Getting out of memory error while running test in parallel 


bachhavdipak:
Getting out of memory error while running test in parallel .  I am using below configuration along with annotation**@Execution(SAME_THREAD)** in the class.

```
junit.jupiter.execution.parallel.enabled = true
junit.jupiter.execution.parallel.mode.default = same_thread
junit.jupiter.execution.parallel.mode.classes.default = same_thread
junit.jupiter.execution.parallel.config.strategy = dynamic

```
Error log:
```

                  2023-03-24T10:08:17.6764760Z UnknownClass.JUnit Jupiter > UnknownClass.executionError FAILED
2023-03-24T10:08:17.6765471Z     org.junit.platform.commons.JUnitException: TestEngine with ID 'junit-jupiter' failed to execute tests
2023-03-24T10:08:17.6765805Z         at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:113)
2023-03-24T10:08:17.6766208Z         at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:88)
2023-03-24T10:08:17.6766499Z         at org.junit.platform.launcher.core.EngineExecutionOrchestrator.lambda$execute$0(EngineExecutionOrchestrator.java:54)
2023-03-24T10:08:17.6766817Z         at org.junit.platform.launcher.core.EngineExecutionOrchestrator.withInterceptedStreams(EngineExecutionOrchestrator.java:67)
2023-03-24T10:08:17.6767143Z         at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:52)
2023-03-24T10:08:17.6767424Z         at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:114)
2023-03-24T10:08:17.6767662Z         at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:86)
2023-03-24T10:08:17.6767942Z         at org.junit.platform.launcher.core.DefaultLauncherSession$DelegatingLauncher.execute(DefaultLauncherSession.java:86)
2023-03-24T10:08:17.6768226Z         at org.junit.platform.launcher.core.SessionPerRequestLauncher.execute(SessionPerRequestLauncher.java:53)
[...]

PR #3044, opened 2022-09-24
Limit max pool size for parallel execution via config param

mpkorstanje:
## Overview

With Java 9+ the `ForkJoinPool` used by JUnit can be configured with a maximum pool size via the `junit.jupiter.execution.parallel.config.fixed.max-pool-size` property.

While the number of concurrently executing tests may exceed the configured parallelism when tests become blocked, it will not exceed the maximum pool size in these circumstances.

With the following configuration:

```properties
junit.jupiter.execution.parallel.enabled=true
junit.jupiter.execution.parallel.config.strategy=fixed
junit.jupiter.execution.parallel.mode.default=concurrent
junit.jupiter.execution.parallel.config.fixed.parallelism=2
```

This example will report between 2-5 tests running concurrently.

```java
class ExampleTest {

    private static final AtomicInteger running = new AtomicInteger();

    @BeforeEach
    void increment() {
        System.out.println("Running " + running.incrementAndGet());
    }

    @AfterEach
    void decrement() {
        running.decrementAndGet();
    }

    static IntStream numbers() {
        return IntStream.range(0, 1000);
    }

    @ParameterizedTest
    @MethodSource("numbers")
    void test(int i) throws ExecutionException, InterruptedException {
        Runnable sleep = () -> {
            try {
                Thread.sleep(600);
            } catch (InterruptedException e) {
                throw new RuntimeException(e);
            }
        };
        ForkJoinPool.commonPool().submit(sleep).get();
    }
}
```

By also configuring the `max-pool-size` we can ensure the concurrently executing test does not exceed the configured 2.

```properties
junit.jupiter.execution.parallel.config.fixed.max-pool-size=2
```

Additionally, because the `ForkJoinPool` will by default reject tasks that would exceed the maximum pool size the
`junit.jupiter.execution.parallel.config.fixed.saturate` property has been added and will default to `true`. There appears to be no reason to ever set this `false` but it is there should someone depend on the old behavior.

These changes were intentionally not made to the `dynamic` strategy to limit the scope of this pull request. While I can reasonably predict what behaviour users of the `fixed` strategy might expect, I can not say the same about the `dynamic` strategy.

Fixes: #2545
Fixes: #1858
Fixes: #3026

This PR replaces #3027.

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement).

---

[...]

PR #3206, opened 2023-03-24
Limit max pool size for dynamic parallel execution

mpkorstanje:
## Overview
This is a followup of #3044 for the `dynamic` strategy.

With Java 9+ the `ForkJoinPool` used by JUnit can be configured with a maximum pool size via the `junit.jupiter.execution.parallel.config.dynamic.max-pool-size-factor` property.

While the number of concurrently executing tests may exceed the configured parallelism factor when tests become blocked, it will not exceed the maximum pool size factor in these circumstances.

With the following configuration:

```properties
junit.jupiter.execution.parallel.enabled=true
junit.jupiter.execution.parallel.config.strategy=dynamic
junit.jupiter.execution.parallel.mode.default=concurrent
junit.jupiter.execution.parallel.config.dynamic.factor=0.1667
```

This example will report between 2 and 4 tests running concurrently on a system with 12 available cores while only (12 * 0.1667 =) 2 would be expected.

```java
class ExampleTest {

    private static final AtomicInteger running = new AtomicInteger();

    @BeforeEach
    void increment() {
        System.out.println("Running " + running.incrementAndGet());
    }

    @AfterEach
    void decrement() {
        running.decrementAndGet();
    }

    static IntStream numbers() {
        return IntStream.range(0, 1000);
    }

    @ParameterizedTest
    @MethodSource("numbers")
    void test(int i) throws ExecutionException, InterruptedException {
        Runnable sleep = () -> {
            try {
                Thread.sleep(600);
            } catch (InterruptedException e) {
                throw new RuntimeException(e);
            }
        };
        ForkJoinPool.commonPool().submit(sleep).get();
    }
}
```

By also configuring the `max-pool-size-factor` we can ensure the concurrently executing test does not exceed the expected 2 .

```properties
junit.jupiter.execution.parallel.config.dynamic.max-pool-size-factor=1
```

Additionally, because the `ForkJoinPool` will by default reject tasks that would exceed the maximum pool size the
`junit.jupiter.execution.parallel.config.dynamic.saturate` property has been added and will default to `true`. There appears to be no reason to ever set this `false` but it is there should someone depend on the old behavior.

Fixes: #3205

---

---

### Definition of Done


junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
```diff
@@ -71,8 +71,20 @@ public ParallelExecutionConfiguration createConfiguration(ConfigurationParameter
...
-			return new DefaultParallelExecutionConfiguration(parallelism, parallelism, 256 + parallelism, parallelism,
-				KEEP_ALIVE_SECONDS, null);
+			int maxPoolSize = configurationParameters.get(CONFIG_DYNAMIC_MAX_POOL_SIZE_FACTOR_PROPERTY_NAME,
+				BigDecimal::new).map(maxPoolSizeFactor -> {
+					Preconditions.condition(maxPoolSizeFactor.compareTo(BigDecimal.ONE) >= 0,
+						() -> String.format(
+							"Factor '%s' specified via configuration parameter '%s' must be greater than or equal to 1",
+							factor, CONFIG_DYNAMIC_FACTOR_PROPERTY_NAME));
+					return maxPoolSizeFactor.multiply(BigDecimal.valueOf(parallelism)).intValue();
+				}).orElseGet(() -> 256 + parallelism);
```
mpkorstanje, line 81: There are two ways to calculate the max pool size:

```
maxPoolSize = dynamic.max-pool-size-factor * dynamic.factor * Runtime.getRuntime().availableProcessors() 
maxPoolSize = dynamic.max-pool-size-factor * Runtime.getRuntime().availableProcessors() 
```

And the choice is a bit arbitrary. But I think reasoning in terms of "threads per degree of parallelism" is a bit easier so I prefer the former.

marcphilipp (approved):

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -2474,7 +2484,7 @@ The following table lists relevant properties for configuring parallel execution
 | ```junit.jupiter.execution.parallel.config.fixed.max-pool-size```
 | Desired maximum pool size of the underlying fork-join pool for the ```fixed```
   configuration strategy
-| a positive integer, must greater than or equal to `junit.jupiter.execution.parallel.config.fixed.parallelism`
+| a positive integer, must be greater than or equal to `junit.jupiter.execution.parallel.config.fixed.parallelism`
 | 256 + the value of `junit.jupiter.execution.parallel.config.fixed.parallelism`
 
 | ```junit.jupiter.execution.parallel.config.fixed.saturate```
```
marcphilipp, line 2490: We should add an entry for the `dynamic` one as well.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -2466,6 +2468,14 @@ The following table lists relevant properties for configuring parallel execution
 | a positive decimal number
 | ```1.0```
 
+| ```junit.jupiter.execution.parallel.config.dynamic.max-pool-size.factor```
```
marcphilipp, line 2471: Why are we using three backticks rather than one in this table?

marcphilipp (changes requested):

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sbrannen: I believe there was a formatting error without the triple backticks.

But maybe that's no longer an issue with more recent versions of Asciidoctor.

So feel free to see what happens if you switch to single backticks.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
mpkorstanje: That one is was `junit.jupiter.execution.parallel.config.dynamic.factor` and was already documented here: https://github.com/junit-team/junit5/blame/07f2bf349e45f5a63228cb6002ae7bdbd3445ce1/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc#L2463-L2467

The full set is:
```
junit.jupiter.execution.parallel.config.dynamic.factor
junit.jupiter.execution.parallel.config.dynamic.max-pool-size-factor

junit.jupiter.execution.parallel.config.fixed.parallelism
junit.jupiter.execution.parallel.config.fixed.max-pool-size
```

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
mpkorstanje: > I believe there was a formatting error without the triple backticks.

This is still the case.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
mpkorstanje: https://github.com/junit-team/junit5/pull/3206/commits/6849f00963c4f33c991cfd2b9dd3009ebab461d9

marcphilipp (approved):

PR #3219, opened 2023-03-31
Document `@MethodSource` as use case for `@TestInstance` per-class lifecycle

mahozad:
## Overview

Related Stack Overflow post: [What use is @TestInstance annotation in JUnit 5?](https://stackoverflow.com/a/66185596/8583692)

Add another use case for `TestInstace.PER_CLASS` in Kotlin, which is, using regular simple class methods as the method factory for `@MethodSource` (which is used for parameterized tests):

```kotlin
@ParameterizedTest
@MethodSource("generateStrings")
fun `Test strings`(argument: String) {
    check(argument.isNotEmpty())
}

private fun generateStrings() = listOf(
    "java",
    "kotlin"
)
```

---

---

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -751,8 +751,8 @@ NOTE: Beginning with Java 16, `@BeforeAll` and `@AfterAll` methods can be declar
 `static` in `@Nested` test classes.
 
 If you are authoring tests using the Kotlin programming language, you may also find it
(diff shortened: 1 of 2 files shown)


sbrannen (changes requested):
Please update the [User Guide](https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-instance-lifecycle) in a similar fashion.

sbrannen (approved):
Thanks for updating the user guide as well.

sbrannen:

Thanks

mahozad:
Thanks for merging.

Issue #3102, opened 2022-12-06
Add a way to inherit ResourceLocks that allows for more parallel execution of tests


Squibbles:
As seen in issue #2038 if a class level resource lock is specified, even if it is only a READ lock it will cause all tests that inherit it to run single threaded if any one of those tests uses a READ_WRITE lock.  It was suggested in a comment by @leonard84 that it might be possible to add an annotation that would apply to a descriptor's children but not to the descriptor itself.  This kind of functionality would be extremely useful in our project because we have 3000+ tests spread over hundreds of test classes.  I would like to apply a READ lock to our commonly used resources in the inheritance chain rather than having to add individual resource locks on each of several thousand tests. However, if I do specify a class level/inherited lock of any kind it makes any class that inherits it run in SAME_THREAD mode if there is any READ_WRITE locks anywhere in the class.

In this below example the first two tests run in parallel and tests 3/4 run individually as expected.  But if you uncomment the class level resource lock then all tests run individually.  I would like there to be a way to specify that all of these tests have a READ lock on resource2 without having to add the lock individually to each test:

```java
   // @ResourceLock(value = "resource2", mode = READ)
   public class SampleClassLevelLocksTest {

    @Test
    @ResourceLock(value = "resource1", mode = READ)
    void test1() throws InterruptedException {
        Thread.sleep(2000L);
        Assertions.assertTrue(true);
    }

    @Test
    @ResourceLock(value = "resource1", mode = READ)
    void test2() throws InterruptedException {
        Thread.sleep(2000L);
        Assertions.assertTrue(true);
    }

    @Test
    @ResourceLock(value = "resource1", mode = READ_WRITE)
    void test3() throws InterruptedException {
        Thread.sleep(2000L);
        Assertions.assertTrue(true);
    }

    @Test
    @ResourceLock(value = "resource1", mode = READ_WRITE)
    void test4() throws InterruptedException {
        Thread.sleep(2000L);
        Assertions.assertTrue(true);
    }
```

## Deliverables

- [ ] Implement a way to specify resource locks that should be inherited to tests without forcing the tests to run in SAME_THREAD mode if there is a READ_WRITE lock anywhere at the same level

marcphilipp:
**Team decision:** Introduce `@ResourceLock(target = SELF | CHILDREN)` where `target` would default to `SELF` to preserve existing behavior. Using `CHILDREN` would have the same effect as declaring `@ResourceLock` on every `@Test` method of a test class.

[...]

PR #3220, opened 2023-04-01
Introduced `@ResourceLock(target = SELF | CHILDREN)`

anatolyD:
Defaults to SELF to preserve existing behavior.
Using CHILDREN has the same effect as declaring @ResourceLock on every @Test method of a test class.

Issue: #3102

## Overview

Introduced `@ResourceLock(target = SELF | CHILDREN)` where the target defaults to `SELF` to preserve existing behavior. Using `CHILDREN` has the same effect as declaring `@ResourceLock` on every `@Test` method of a test class.

I've added a couple of integration tests to verify it works, plus one for NodeWalker. Happy to hear the feedback. 😄 

---

---

### Definition of Done


junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLockTarget.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package org.junit.jupiter.api.parallel;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import org.apiguardian.api.API;
+
+/**
+ *	Indicates the target of a {@link ResourceLock}.
```
marcphilipp, line 18:
```suggestion
 * Indicates the target of a {@link ResourceLock}.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLock.java
```diff
@@ -40,6 +40,9 @@
  * <p>Since JUnit Jupiter 5.4, this annotation is {@linkplain Inherited inherited}
  * within class hierarchies.
  *
+ * <p>Since JUnit Jupiter 5.10, this annotation can be used to specify the target
+ * of the lock with {@link ResourceLockTarget}.
+ *
```
marcphilipp, line 45: What happens if a class has a `@ResourceLock` annotation with target `CHILDREN` and a method has one for the same resource? Does the one from the method override the one from the class? This should be documented here.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ExclusiveResource.java
```diff
@@ -52,18 +52,34 @@ public class ExclusiveResource {
...
+	/**
+	 * Create a new {@code ExclusiveResource}.
+	 *
+	 * @param key the identifier of the resource; never {@code null} or blank
+	 * @param lockMode the lock mode to use to synchronize access to the
+	 * resource; never {@code null}
+	 * @param lockScope the lock scope to use to synchronize access to the
+	 * resource; never {@code null}
+	 */
+	public ExclusiveResource(String key, LockMode lockMode, LockScope lockScope) {
```
marcphilipp, line 79: I don't think this should be a Platform concept. Instead, Jupiter should return `ExclusiveResource` instances from the children `TestDescriptors` if their parent declares a `@ResourceLock` with `target` of `CHILDREN`.

marcphilipp (changes requested):
Sorry for the delay! This looks very useful. 🙂

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ExclusiveResource.java
anatolyD: sounds good, no need to bother Platform with those details :) I drafted how it may look in https://github.com/junit-team/junit5/pull/3220/files#diff-63a01db8dd78b02e05c3888eb786c656af0cd3dc8263c816530003f2d7069daeR188

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -182,15 +181,32 @@ public static ExecutionMode toExecutionMode(org.junit.jupiter.api.parallel.Execu
...
-	Set<ExclusiveResource> getExclusiveResourcesFromAnnotation(AnnotatedElement element) {
+	protected List<ResourceLock> getResourceLocks() {
+		return Collections.emptyList();
+	}
+
+	protected Set<ExclusiveResource> collectExclusiveResourcesFromHierarchy() {
+		Set<ExclusiveResource> resources = mapResourceLocksForTarget(getResourceLocks(), ResourceLockTarget.SELF);
+
+		Optional<TestDescriptor> nextParent = getParent();
+		while (nextParent.isPresent()) {
```
marcphilipp, line 192: IIUC `@ResourceLock(target = CHILDREN)` should only cause them to be applied to _direct_ children.

marcphilipp (changes requested):

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
Squibbles: > IIUC `@ResourceLock(target = CHILDREN)` should only cause them to be applied to _direct_ children.

Is there any reason for this?  Is it harmful in someway for it to not be just direct children?  e.g. in our project our tests inherit from base classes scattered throughout our packages and those base classes tend to inherit from a single source parent.  If I have a resource that I want read locked for everything (main server component for example) it makes more sense to me to have the lock specified in one place instead of having to remember to add it to all the base classes

marcphilipp: @Squibbles This is unrelated to inheritance in class hierarchies. The `@ResourceLock` annotation is inherited to subclasses already. However, the way resource locks are applied is that if a node acquires a lock, its children already hold it while being executed. Therefore it doesn't make sense to recursively declare it for all descendants.

marcphilipp (changes requested):
I think we also need to think about the case where `@ResourceLock(target = CHILDREN)` is applied to methods. I think for `@Test` this should cause a failure. For test templates such as `@ParameterizedTest` and `@RepeatedTest` the invocations should get the `ExclusiveResource` (might already work with the code in this PR but we're missing a test). For `@TestFactory` methods, the direct `DynamicNodeTestDescriptor` children should get the `ExclusiveResource`.

marcphilipp:
@anatolyD Do you have time to continue working on this PR or should someone from the JUnit team take over?

anatolyD:
> @anatolyD Do you have time to continue working on this PR or should someone from the JUnit team take over?

Hey, yes I would love to continue, but if time is a pressuring matter you may want to take it over. I plan to get to it within a week or two

marcphilipp:
I don't think it's not urgent so that's fine. 👍

marcphilipp:
@anatolyD Are you still interested in continuing on this PR?

anatolyD:
Hey, right, it should be closed. Will not progress on it

marcphilipp:
No worries, thanks for letting us know!

Issue #3235, opened 2023-04-07
Apostrophe in display name breaks name generation for `@ParameterizedTest`


TWiStErRob:
## Steps to reproduce

```kotlin
class Tests {
	@BeforeEach 
	fun setUp(info: TestInfo) {
		System.out.println(info.displayName)
	}

	@MethodSource("methodSource")
	@ParameterizedTest(name = "{displayName} - {0}") // BUG
	fun `displayName'Zero`(param: String) {
		// displayNameZero(String) - {0}
		// displayNameZero(String) - {0}
	}

	@MethodSource("methodSource")
	@ParameterizedTest(name = "{0} - {displayName}") 
	fun `zero'DisplayName`(param: String) {
		// foo - zeroDisplayName(String)
		// bar - zeroDisplayName(String)
	}

	@MethodSource("methodSource")
	@ParameterizedTest(name = "{displayName} - {0}") 
	fun displayNameZero(param: String) {
		// displayNameZero(String) - foo
		// displayNameZero(String) - bar
	}

	@MethodSource("methodSource")
	@ParameterizedTest(name = "{0} - {displayName}") 
	fun zeroDisplayName(param: String) {
		// foo - zeroDisplayName(String)
		// bar - zeroDisplayName(String)
	}

	companion object {

		@JvmStatic
		private fun methodSource(): Array<String> =
			arrayOf(
				"foo",
				"bar"
			)
	}
}
```

1. Clone https://github.com/TWiStErRob/repros/tree/main/junit/jupiter-params-displayName-vs-apostrophe
2. `gradlew test`
3. Review report at `build/reports/tests/test/index.html`

Note: the repro project is a bit more complex than the above:
 * `TestK` is a pure Kotlin version
 * `TestJ` is a pure Java version
 * `Test` is the problematic version with Kotlin apostrophes.
I made them really similar so it's easy to compare their outputs.

## Expected
No `{0}` anywhere.
It works correctly without apostrophes:
`build/reports/tests/test/classes/com.example.MyTestK.html`

## Actual
When the method name contains `'`, the `{0}` is not substituted.
It fails formatting the name when apostrophes are added:
`build/reports/tests/test/classes/com.example.MyTest.html`

## Context

 - Used versions (Jupiter/Vintage/Platform): 5.9.2
 - Build Tool/IDE: Gradle / IDEA, but likely irrelevant.

## Deliverables

I think the problem might be around `ParameterizedTestNameFormatter`, but didn't debug it.

[...]

PR #3264, opened 2023-04-21
Do not parse `{displayName}` placeholder for `@ParameterizedTest` using `MessageFormat`

alemendoza-v:
## Overview

The `@ParameterizedTest` annotation was breaking name generation by not allowing `'` in the method names for Kotlin-based tests or in custom display names in geneeral. It also made it so that placeholders were not being replaced correctly.

To solve it, instead of replacing the `DISPLAY_NAME_PLACEHOLDER` before the `MessageFormat` was being evaluated, the `DISPLAY_NAME_PLACEHOLDER` is now replaced with another temporary placeholder, which is then replaced after the `MessageFormat` is evaluated.

Solves: #3235 

---

---

### Definition of Done

- [x] Change is documented in the [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
@@ -36,6 +36,7 @@ class ParameterizedTestNameFormatter {
 	private final String displayName;
 	private final ParameterizedTestMethodContext methodContext;
 	private final int argumentMaxLength;
+	private final String DISPLAY_NAME_TEMPORARY_PLACEHOLDER = "__DISPLAY_NAME__";
 
 	ParameterizedTestNameFormatter(String pattern, String displayName, ParameterizedTestMethodContext methodContext,
 			int argumentMaxLength) {
@@ -62,7 +63,7 @@ class ParameterizedTestNameFormatter {
 		MessageFormat format = new MessageFormat(pattern);
 		Object[] humanReadableArguments = makeReadable(format, namedArguments);
 		String formatted = format.format(humanReadableArguments);
-		return formatted.replace("__DISPLAY_NAME__", this.displayName);
+		return formatted.replace(DISPLAY_NAME_TEMPORARY_PLACEHOLDER, this.displayName);
 	}
 
 	private Object[] extractNamedArguments(Object[] arguments) {
@@ -73,7 +74,7 @@ class ParameterizedTestNameFormatter {
 
 	private String prepareMessageFormatPattern(int invocationIndex, Object[] arguments) {
 		String result = pattern//
-				.replace(DISPLAY_NAME_PLACEHOLDER, "__DISPLAY_NAME__")//
+				.replace(DISPLAY_NAME_PLACEHOLDER, DISPLAY_NAME_TEMPORARY_PLACEHOLDER)//
 				.replace(INDEX_PLACEHOLDER, String.valueOf(invocationIndex));
 
 		if (result.contains(ARGUMENTS_WITH_NAMES_PLACEHOLDER)) {
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
@@ -62,7 +62,7 @@ class ParameterizedTestNameFormatter {
 		MessageFormat format = new MessageFormat(pattern);
 		Object[] humanReadableArguments = makeReadable(format, namedArguments);
 		String formatted = format.format(humanReadableArguments);
-		return formatted.replace("__DISPLAY_NAME__",  this.displayName);
+		return formatted.replace("__DISPLAY_NAME__", this.displayName);
 	}
 
 	private Object[] extractNamedArguments(Object[] arguments) {
@@ -72,9 +72,9 @@ class ParameterizedTestNameFormatter {
 	}
 
 	private String prepareMessageFormatPattern(int invocationIndex, Object[] arguments) {
-			String result = pattern//
-					.replace(DISPLAY_NAME_PLACEHOLDER, "__DISPLAY_NAME__")//
-					.replace(INDEX_PLACEHOLDER, String.valueOf(invocationIndex));
+		String result = pattern//
+				.replace(DISPLAY_NAME_PLACEHOLDER, "__DISPLAY_NAME__")//
+				.replace(INDEX_PLACEHOLDER, String.valueOf(invocationIndex));
 
 		if (result.contains(ARGUMENTS_WITH_NAMES_PLACEHOLDER)) {
 			result = result.replace(ARGUMENTS_WITH_NAMES_PLACEHOLDER, argumentsWithNamesPattern(arguments));
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
@@ -61,7 +61,8 @@ class ParameterizedTestNameFormatter {
 		String pattern = prepareMessageFormatPattern(invocationIndex, namedArguments);
 		MessageFormat format = new MessageFormat(pattern);
 		Object[] humanReadableArguments = makeReadable(format, namedArguments);
-		return format.format(humanReadableArguments);
+		String formatted = format.format(humanReadableArguments);
+		return formatted.replace("__DISPLAY_NAME__",  this.displayName);
 	}
 
 	private Object[] extractNamedArguments(Object[] arguments) {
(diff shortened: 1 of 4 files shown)


junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
```diff
@@ -72,7 +73,7 @@ private Object[] extractNamedArguments(Object[] arguments) {
 
 	private String prepareMessageFormatPattern(int invocationIndex, Object[] arguments) {
 		String result = pattern//
-				.replace(DISPLAY_NAME_PLACEHOLDER, this.displayName)//
+				.replace(DISPLAY_NAME_PLACEHOLDER, "__DISPLAY_NAME__")//
```
sbrannen, line 76: Please extract this to a constant.

junit-jupiter-params/src/test/kotlin/ParameterizedTestNameFormatterIntegrationTests.kt
```diff
@@ -0,0 +1,69 @@
...
+class ParameterizedTestNameFormatterIntegrationTests {
+    @BeforeEach
+    fun setUp(info: TestInfo) {
+        System.out.printf(
+            "Class: %s, method: %s%nDisplay Name: %s%n%n",
+            info.testClass.orElseThrow(),
+            info.testMethod.orElseThrow(),
+            info.displayName
+        )
+    }
```
sbrannen, line 28: We do not print to SYSOUT in tests for debugging purposes.

So, please remove this method.

junit-jupiter-params/src/test/kotlin/ParameterizedTestNameFormatterIntegrationTests.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    @MethodSource("methodSource")
+    @ParameterizedTest
+    fun `implicit'Name`(param: String) {
+        Assertions.assertNotNull(param)
```
sbrannen, line 33: Please static imports for `Assertions`.

junit-jupiter-params/src/test/kotlin/ParameterizedTestNameFormatterIntegrationTests.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @MethodSource("methodSource")
+    @ParameterizedTest(name = "{0}")
+    fun `zero'Only`(param: String) {
+        Assertions.assertNotNull(param)
```
sbrannen, line 39: Please modify all test methods to accept `TestInfo` as an argument and then replace the `Assertions.assertNotNull(param)` assertions with something that asserts the actual expected display name.

sbrannen (changes requested):
Thanks for submitting the PR.

alemendoza-v:
@sbrannen Sure I'll get to work on these changes. Should the Release Notes entry include the basic change? Something like "Bug Fix: `@ParameterizedTest` now accepts single apostrophes inside the method names in Kotlin tests"?

alemendoza-v:
@sbrannen Also,  should the new constant be inside the `Interface` or inside the `ParameterizedTest` class?

junit-jupiter-params/src/test/kotlin/ParameterizedTestNameFormatterIntegrationTests.kt
alemendoza-v: I think static imports in Kotlin do not exist :(

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
```diff
@@ -36,6 +36,7 @@ class ParameterizedTestNameFormatter {
 	private final String displayName;
 	private final ParameterizedTestMethodContext methodContext;
 	private final int argumentMaxLength;
+	private final String DISPLAY_NAME_TEMPORARY_PLACEHOLDER = "__DISPLAY_NAME__";
```
sbrannen, line 39: Please make this `static` and declare it under `ELLIPSIS`.

junit-jupiter-params/src/test/java/org/junit/jupiter/params/ParameterizedTestNameFormatterTests.java
```diff
@@ -60,6 +60,14 @@ void formatsDisplayName() {
...
 	}
 
+	@Test
+	void formatsDisplayNameWithApostrophe() {
+		var formatter = formatter(DISPLAY_NAME_PLACEHOLDER, "display'Zero");
+
+		assertEquals("display'Zero", formatter.format(1));
+		assertEquals("display'Zero", formatter.format(2));
+	}
+
```
sbrannen, line 70:
```suggestion
	@Test
	void formatsDisplayNameWithApostrophe() {
		String displayName = "display'Zero";
		var formatter = formatter(DISPLAY_NAME_PLACEHOLDER, displayName);

		assertEquals(displayName, formatter.format(1));
		assertEquals(displayName, formatter.format(2));
	}

```

Please extract the `displayName` variable as above.

And please introduce the following test which demonstrates that the bug is not solely related to apostrophes.

```java
	@Test
	void formatsDisplayNameContainingFormatElements() {
		String displayName = "{enigma} {0} '{1}'";
		var formatter = formatter(DISPLAY_NAME_PLACEHOLDER, displayName);

		assertEquals(displayName, formatter.format(1));
		assertEquals(displayName, formatter.format(2));
	}
```

junit-jupiter-params/src/test/kotlin/ParameterizedTestNameFormatterIntegrationTests.kt
```diff
@@ -0,0 +1,75 @@
...
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.MethodOrderer
+import org.junit.jupiter.api.TestInfo
+import org.junit.jupiter.api.TestMethodOrder
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.MethodSource
+
+@TestMethodOrder(MethodOrderer.MethodName::class)
```
sbrannen, line 17:
```suggestion
```

You can safely remove this, since it is not necessary for these tests.

junit-jupiter-params/src/test/kotlin/ParameterizedTestNameFormatterIntegrationTests.kt
```diff
@@ -0,0 +1,75 @@
...
+class ParameterizedTestNameFormatterIntegrationTests {
+
+    @MethodSource("methodSource")
```
sbrannen, line 20: Why did you choose the more complicated setup for `@MethodSource` with a companion object and `@JvmStatic`?

I'd likely go with `@ValueSource(strings = ["foo", "bar"])` (assuming that's the correct syntax for Kotlin).

sbrannen (changes requested):
Thanks for making the previously requested changes.

I've requested a few additional changes after the latest review.

sbrannen:
FYI: changes to the Javadoc and User Guide are not needed for this change, so I've updated the _Deliverables_ accordingly.

junit-jupiter-params/src/test/kotlin/ParameterizedTestNameFormatterIntegrationTests.kt
sbrannen: They do, but they don't need the keyword `static` like in Java.

See `org.junit.jupiter.params.aggregator.ArgumentsAccessorKotlinTests` for examples.

sbrannen:
> should the new constant be inside the `Interface` or inside the `ParameterizedTest` class?

It's good as a private constant in `ParameterizedTestNameFormatter` as you have it now, since it is an internal implementation detail local to exactly that one class.

sbrannen:
> Should the Release Notes entry include the basic change? Something like "Bug Fix: `@ParameterizedTest` now accepts single apostrophes inside the method names in Kotlin tests"?

There's no need to prefix it with "Bug Fix:". There's a [dedicated section for bug fixes in JUnit Jupiter](https://github.com/junit-team/junit5/blob/main/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.0-M1.adoc#bug-fixes-1).

Regarding the content, it should be more generic in nature, explaining that the `{displayName}` placeholder is no longer parsed using `MessageFormat`, which means that a `@DisplayName` or Kotlin test method name can contain apostrophes and `MessageFormat` _elements_ (such as `{data}`).

See what you come up with, and if we deem it necessary we will polish the wording when merging the PR.

alemendoza-v:
@sbrannen Thanks for reviewing, I'll get to work on the changes and the release notes entry :)

junit-jupiter-params/src/test/kotlin/ParameterizedTestNameFormatterIntegrationTests.kt
alemendoza-v: The example that the author showed used that setup. But you're right I'll change it to a simpler setup.

sbrannen:
This has been merged into `main` in 7b2688c8cbe6349656ef94492dceff6537d6a87c and revised in 745714463175df7198b018c446ea7a31fb19cd65.

In addition, since this PR was finalized before the `5.9.3` release, I backported this to `5.9.x` for inclusion in the upcoming `5.9.3` release.

Thanks

alemendoza-v:
Thank you!

Issue #3071, opened 2022-10-21
User guide has several accessibility issues


Bukama:
The [JUnit user guide](https://junit.org/junit5/docs/current/user-guide/) has several, repeating accessibility issues, resulting in not fulfilling the "AA" requirements definied by WCAG (and picked up by laws, e.g. the German BITV 2.0).

According to WAVE analysis there are 205 errors and 1174 contrast errors. After a quick view it seems that all of them go down to two things:

* Empty links: Violates level A requirement "2.4.4 Link Purpose (In Context)" of WCAG
* Very low contrast: Violates level AA requirement "1.4.3 Contrast (Minimum) (Level AA)"

![overallreport](https://user-images.githubusercontent.com/4354359/197243718-a04d0068-4870-4c18-840b-23129c58a576.png)

To fix the empty links it's necessary to provide text within the link that describes the functionality and/or target of that link.

To fix the low contrast errors the CSS of the code examples should be adjusted to give a proper contrast. All low contrast errors I randomly checked were located in code examples have a contrast ratio of 4.49:1, but WVAG level AA requires to meet 4.5:1. So only slight adjustments should fix all of these errors. For example changing the current `#008080`  to `#005555` should result in a contrast fulfilling even AAA level.

![lowcontrast](https://user-images.githubusercontent.com/4354359/197243674-5bac56b0-c7c3-4889-aa33-08a085aa63a0.png)

Of the 98 alerts the 85 layout tables warnings are problematic, as screen readers (like NVDA or JAWS) read out the table structure. also the fact that no valid HTML 5 with navigation and content regions / elements is definied is problematic for users who are in need of screen readers etc. as those tools could not identify the page structure.

marcphilipp:
Apart from syntax highlighting we're using the default Asiidoctor template/styles. Is there a corresponding issue there (e.g. for the "empty link" issue)?

Bukama:
When I search in their 700+ open issue for accessibility it comes down to [6 open (and 3 closed)](https://github.com/asciidoctor/asciidoctor/issues?q=is%3Aissue+is%3Aopen++accessibility). One of them ( https://github.com/asciidoctor/asciidoctor/issues/242 ) topics HTML5 but a little bit other to what I've mentioned. There are no issues for the contrast one.

Of course there is none for the links as a style can't know what links you create.

marcphilipp:
@Bukama Could you make the full report available please or provide instructions how to recreate it?

Bukama:
[...]

PR #3267, opened 2023-04-22
Create custom rouge theme

Bukama:
As @marcphilipp  asked me to push my try for a custom asciidoc theme (see #3071 ):

What I tried first was chaning the default theme to one of the other build-in themes, e.g. `colorful`.
This was done easy, as you only have to change the `index.adoc` and add the `:rouge-style: colorful` tag, run the gradle task "asciidoc" and the colorful theme is applied.
It doesn't matter if you set the attribute in the `documentation.gradle.kts` file - set settings in this file are not picked up anyway.

I then tried to create an own theme, by starting to just take the example from [the asciidoc example](https://docs.asciidoctor.org/pdf-converter/latest/theme/source-highlighting-theme/), and just chaning some colors to blue, to make them visual and to see that the theme is picked up. I also added the file to the copy ressources list (`documentation.gradle.kts`) as the guide says you have to do so.

It turns out that the file is copies to `build\docs\asciidoc\resources\themes\` but it does not get picked up, even I figured out that the pure existence of the file, regardless if the new added file / theme is mentioned anyway breaks the theme usage. So even when the `colorful` theme is set, the guide is rendered in default theme. The name / location of the theme file or the theme doesn't change anything about this behavior.

I also tried to run Gradle with "--debug" (as described in the [IntelliJ Gradle Debug docs](https://www.jetbrains.com/help/idea/work-with-gradle-tasks.html)), but no debug information were written.

Well and this whole situation is what blows my mind since hours.

What you see in the PR is that the added file breaks the theme usage. Only thing that works with the current state of this PR is the copying of the file due `include("resources/themes/rogue_junit.rb")` in the Gradle-file. If you comment out those and remove the theme file, you will see that the setting to colorful is used again.


marcphilipp:
@Bukama Thanks! I got it to work. The most important bit was that `asciidoctor -r` is short for "require" (cf. [docs](https://docs.asciidoctor.org/asciidoctor/latest/cli/man1/asciidoctor/#options)) so it had to be added to the asciidoctorj config.

Bukama:
> @Bukama Thanks! I got it to work. The most important bit was that `asciidoctor -r` is short for "require" (cf. [docs](https://docs.asciidoctor.org/asciidoctor/latest/cli/man1/asciidoctor/#options)) so it had to be added to the asciidoctorj config.

hmm. And I thought it was for `resources`, why I tried to added them to that section (where the regular code was in the `source` section` 🤦‍♂️

marcphilipp:
@Bukama We'd like to include this PR in the upcoming 5.10-M1 release. Do you have time to make the changes requested in https://github.com/junit-team/junit5/issues/3071#issuecomment-1536083849 until mid next week?

Bukama:
Removed the dark theme, based on team decision in the linked issue and turned the draft PR into a regular PR.

Proposed commit message.

```
Create custom rouge theme

The rouge code formatting themes are used
by asciidoc, which is used to create the
user guide. But the default theme and all
overs, except the "bw" theme, will create
low color contrast, which can be problematic
for people with visual restrictions.

Instead of using rouge's "bw" theme a custom
theme was created to tackle these issue, because
using the "bw" theme there was no visual switch 
between regular text and code examples.

The created theme is a light themes that 
solves all contrast errors in the user guide.
However it can't fix other accessibility problems
of the user guide, like missing link textes.
Those other problems have to be fixed seperately.

relates to #3071
#PR 3267
```

marcphilipp (approved):

marcphilipp:
@Bukama Thanks for your perseverance in getting this addressed! 👍

PR #3275, opened 2023-04-26
Add `junit.jupiter.tempdir.factory.default` configuration property

scordio:
## Overview

Follow-up to #2958 which adds a `junit.jupiter.tempdir.factory.default` configuration property.

---

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/JupiterConfiguration.java
```diff
@@ -66,4 +67,6 @@ public interface JupiterConfiguration {
 
 	CleanupMode getDefaultTempDirCleanupMode();
 
+	TempDirFactory getDefaultTempDirFactory();
```
marcphilipp, line 70: Doesn't this need to be a `Supplier<TempDirFactory>` so that one instance per declaration can be created?

scordio: Here I've taken inspiration from `getDefaultDisplayNameGenerator`. I didn't write tests yet so you're most likely right 🙂 I'll check it and confirm.

marcphilipp: I think it's different since `DisplayNameGenerator` does not extend `Closeable` and therefore has no defined lifecycle.

scordio: I imagined something like the following for `DefaultJupiterConfiguration`:
```java
@Override
public Supplier<TempDirFactory> getDefaultTempDirFactorySupplier() {
	return () -> tempDirFactoryConverter.get(configurationParameters, DEFAULT_FACTORY_PROPERTY_NAME) //
			.orElse(TempDirFactory.Standard.INSTANCE);
}
```

However, this limits the usefulness of `CachingJupiterConfiguration`:

```java
@Override
public Supplier<TempDirFactory> getDefaultTempDirFactorySupplier() {
	return (Supplier<TempDirFactory>) cache.computeIfAbsent(DEFAULT_FACTORY_PROPERTY_NAME,
		key -> delegate.getDefaultTempDirFactorySupplier());
}
```

as the "heavy" calculation (checking the property + instantiation) is part of the supplier.

Probably I should introduce an alternative to `InstantiatingConfigurationParameterConverter`, i.e., a new `ParameterConverter` that returns the supplier directly.

Would that make sense or do you see a better option?

scordio: Actually, the following in DefaultJupiterConfiguration should be sufficient:

```java
@Override
public Supplier<TempDirFactory> getDefaultTempDirFactorySupplier() {
	return tempDirFactoryConverter.get(configurationParameters, DEFAULT_FACTORY_PROPERTY_NAME) //
			.map(tempDirFactory -> (Supplier<TempDirFactory>) () -> tempDirFactory) //
			.orElse(() -> TempDirFactory.Standard.INSTANCE);
}
```

scordio: Done in a0951ac.

marcphilipp: That didn't quite cut it because `tempDirFactory` was still only instantiated once but closed per declaration. I pushed 48ab4699a78a05c5d257b4cd1d90719da90bd861 to address that.

marcphilipp:
I added another commit that's not strictly part of this PR to ensure the `TempDirFactory` is closed in all cases even if `CleanupMode` is `NEVER` or an exception is thrown during cleanup. 

@scordio Could you please add tests for that? 😇

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/JupiterConfiguration.java
scordio: Cool, thanks!

scordio:
> @scordio Could you please add tests for that? 😇

There were actually (wrong) tests in `CloseablePathCleanupTests` for the previous behavior, which are now broken 🎉 I'll update them accordingly.

marcphilipp:
I think implementation-wise this looks good now. It's missing test coverage for the new config param and documentation. I'd really like to include it in 5.10-M1 which we're planning on releasing by the end of the week. @scordio Do you think you'll have time to add those by then?

scordio:
@marcphilipp yes, definitely!

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/InstantiatingConfigurationParameterConverter.java
```diff
@@ -33,18 +35,23 @@ public InstantiatingConfigurationParameterConverter(Class<T> clazz, String name)
 	}
 
 	Optional<T> get(ConfigurationParameters configurationParameters, String key) {
+		return supply(configurationParameters, key).get();
+	}
+
+	Supplier<Optional<T>> supply(ConfigurationParameters configurationParameters, String key) {
```
scordio, line 41: @marcphilipp although this is implicitly covered by existing tests of `get`, I assume we want to have dedicated tests for `supply` since it's exposed, right?

scordio:
@marcphilipp aside from the coverage of `InstantiatingConfigurationParameterConverter::supply`, I should be ready with all other tests and I plan to work on the documentation and release notes tomorrow.

In case you spot any gaps in coverage or anything else you'd like to do differently, please let me know!

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/InstantiatingConfigurationParameterConverter.java
marcphilipp: Since they would only call `get` I'm not sure if it's worth it.

scordio: I'm not sure either. I leave it aside.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -2643,3 +2643,20 @@ temporary directory. The following example shows how to do it.
...
+annotation specifies a custom factory.
+
+In summary, the factory for a temporary directory is determined according to the
+following precedence rules:
+
+1. by looking at the `factory` attribute of the `@TempDir` annotation, if present
+2. by looking at the default `TempDirFactory` configured via the configuration
+parameter, if present
+
+If none is found, `org.junit.jupiter.api.io.TempDirFactory.Standard` is used.
```
scordio, line 2662: I'm not entirely sure this section is phrased in the best way.

May I ask @sbrannen for native speaker support? 😇

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -177,22 +181,40 @@ private CleanupMode determineCleanupMode(TempDir tempDir) {
...
+
+		if (factory != TempDirFactory.Standard.class && scope == Scope.PER_CONTEXT) {
+			throw new ExtensionConfigurationException("Custom @TempDir factory is not supported with "
+					+ TempDir.SCOPE_PROPERTY_NAME + "=" + Scope.PER_CONTEXT.toString().toLowerCase() + ". Use "
+					+ TempDir.DEFAULT_FACTORY_PROPERTY_NAME + " instead.");
+		}
+
 		return factory == TempDirFactory.Standard.class //
-				? TempDirFactory.Standard.INSTANCE
+				? this.configuration.getDefaultTempDirFactorySupplier().get()
```
sbrannen, line 217: Unless I'm missing something... it appears that the logic here is broken.

If a user specifies their own "default" factory globally via the configuration parameter and then wants to actually use the `Standard` implementation for selected usage of `@TempDir`, then that is impossible with the current setup.

The proper way to handle default values for annotation attributes of type `Class` that allow a user to configure a strategy implementation is to have the attribute _default_ to the interface type instead of a concrete type.

In other words, `org.junit.jupiter.api.io.TempDir.factory()` should be declared as follows.

```java
Class<? extends TempDirFactory> factory() default TempDirFactory.class;
```

The logic for looking up what the user actually wants should then take that into account.

1. If the `factory` attribute is set to `TempDirFactory`, that means the user wants the default factory (either what's been configured globally or `Standard`).
2. If the `factory` attribute is set to anything else, that means the user wants to use that other thing.

sbrannen (changes requested):
I have requested several small changes; however, the most important issue (unless I misunderstood the code) is that a user cannot switch back to using the `Standard` factory implementation.

Please address these issues, and I'll take another look.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sbrannen: See review

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
scordio: Actually, having `default TempDirFactory.class` or not having a default value at all should be the same, right?

sbrannen: No. Not having a default makes it required.

Meaning... the user would always have to specify which factory type to use.

Having a default value for the attribute makes it an "optional" attribute -- which is what we want.

scordio: Yep, mine was a silly question 🤦 😅

sbrannen: FYI: I've encountered this type of use case several times over the years.

Spring's [`@ComponentScan(nameGenerator = ...)`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/ComponentScan.html#nameGenerator()) is a concrete example.

The Javadoc for that attribute describes the same sort of use case we have here with the `factory` for temporary directories.

> The default value of the [BeanNameGenerator](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/support/BeanNameGenerator.html) interface itself indicates that the scanner used to process this `@ComponentScan` annotation should use its inherited bean name generator, e.g. the default [AnnotationBeanNameGenerator](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/AnnotationBeanNameGenerator.html) or any custom instance supplied to the application context at bootstrap time.

scordio: Done in 9eb31a55 and 29df048a.

sbrannen (approved):

sbrannen:
@marcphilipp, do you want to take another look?

marcphilipp (approved):

marcphilipp:
@scordio Thanks once again! 🙇

PR #3278, opened 2023-04-29
Enable Gradle's configuration cache by default

marcphilipp:
To opt-out, use `--no-configuration-cache`


sbrannen:
@marcphilipp, What's the status of this PR?

marcphilipp:
I think we need to wait for https://github.com/diffplug/spotless/issues/1274 to be resolved.

PR #3280, opened 2023-05-03
Load classes using the test's own `ClassLoader`

chrisr3:
Load classes using a `ClassLoader` obtained from the test class or fall back to the default class loader.

This allows JUnit to find the classes while running inside an OSGi framework.

---

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
@@ -57,13 +57,14 @@ abstract class MethodBasedCondition<A extends Annotation> implements ExecutionCo
 	}
 
 	private Method getConditionMethod(String fullyQualifiedMethodName, ExtensionContext context) {
+		final Class<?> testClass = context.getRequiredTestClass();
 		if (!fullyQualifiedMethodName.contains("#")) {
-			return findMethod(context.getRequiredTestClass(), fullyQualifiedMethodName);
+			return findMethod(testClass, fullyQualifiedMethodName);
 		}
 		String[] methodParts = ReflectionUtils.parseFullyQualifiedMethodName(fullyQualifiedMethodName);
 		String className = methodParts[0];
 		String methodName = methodParts[1];
-		Class<?> clazz = ReflectionUtils.tryToLoadClass(className).getOrThrow(
+		Class<?> clazz = ReflectionUtils.tryToLoadClass(className, testClass.getClassLoader()).getOrThrow(
 			cause -> new JUnitException(format("Could not load class [%s]", className), cause));
 		return findMethod(clazz, methodName);
 	}
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
@@ -73,7 +73,7 @@ class MethodArgumentsProvider extends AnnotationBasedArgumentsProvider<MethodSou
 		}
 
 		// Find factory method using fully-qualified name.
-		Method factoryMethod = findFactoryMethodByFullyQualifiedName(testMethod, factoryMethodName);
+		Method factoryMethod = findFactoryMethodByFullyQualifiedName(testClass, testMethod, factoryMethodName);
 
 		// Ensure factory method has a valid return type and is not a test method.
 		Preconditions.condition(isFactoryMethod.test(factoryMethod), () -> format(
@@ -103,12 +103,13 @@ class MethodArgumentsProvider extends AnnotationBasedArgumentsProvider<MethodSou
 		return true;
 	}
 
-	private static Method findFactoryMethodByFullyQualifiedName(Method testMethod, String fullyQualifiedMethodName) {
+	private static Method findFactoryMethodByFullyQualifiedName(Class<?> testClass, Method testMethod,
+			String fullyQualifiedMethodName) {
 		String[] methodParts = ReflectionUtils.parseFullyQualifiedMethodName(fullyQualifiedMethodName);
 		String className = methodParts[0];
 		String methodName = methodParts[1];
 		String methodParameters = methodParts[2];
-		Class<?> clazz = loadRequiredClass(className);
+		Class<?> clazz = loadRequiredClass(className, testClass.getClassLoader());
 
 		// Attempt to find an exact match first.
 		Method factoryMethod = ReflectionUtils.findMethod(clazz, methodName, methodParameters).orElse(null);
@@ -170,8 +171,8 @@ class MethodArgumentsProvider extends AnnotationBasedArgumentsProvider<MethodSou
 				|| isAnnotated(candidate, TestFactory.class);
 	}
 
-	private static Class<?> loadRequiredClass(String className) {
-		return ReflectionUtils.tryToLoadClass(className).getOrThrow(
+	private static Class<?> loadRequiredClass(String className, ClassLoader classLoader) {
+		return ReflectionUtils.tryToLoadClass(className, classLoader).getOrThrow(
 			cause -> new JUnitException(format("Could not load class [%s]", className), cause));
 	}
 
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.0-M1.adoc
@@ -15,7 +15,8 @@ repository on GitHub.
 
 ==== Bug Fixes
 
-* ❓
+* Use a test class's own classloader to search for `@MethodSource` and `MethodBasedCondition` methods.
+  This allows these features to work correctly inside an OSGi framework.
 
 ==== Deprecations and Breaking Changes
 


chrisr3:
> > Also load classes via `Class.forName()`, which is preferred over `ClassLoader.loadClass()` (The latter involves taking class-loading locks, leading to contention in a multi-threaded environment.)
> 
> @chrisr3 Are there any docs stating that?

I would expect so, although I was actually thinking of my own experience profiling an application which did a lot of classloading. We noticed that `ClassLoader.loadClass()` was constantly competing for the Application classloader's class-loading lock:
```java
protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
    synchronized(this.getClassLoadingLock(name)) {
        Class<?> c = this.findLoadedClass(name);
        if (c == null) {
            try {
                if (this.parent != null) {
                    c = this.parent.loadClass(name, false);
                } else {
                    c = this.findBootstrapClassOrNull(name);
                }
            } catch (ClassNotFoundException var10) {
            }

            if (c == null) {
                c = findClass(name);
            }
        }
        ...
    }
```
And all of this contention disappeared when we switched to `Class.forName(String, boolean, ClassLoader)` instead.

The JavaDoc for [`ClassLoader.loadClass()`](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#loadClass-java.lang.String-boolean-) may be relevant here:
> Unless overridden, this method synchronizes on the result of [getClassLoadingLock](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getClassLoadingLock-java.lang.String-) method during the entire class loading process.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java
```diff
@@ -108,7 +108,7 @@ private static Method findFactoryMethodByFullyQualifiedName(Method testMethod, S
 		String className = methodParts[0];
 		String methodName = methodParts[1];
 		String methodParameters = methodParts[2];
-		Class<?> clazz = loadRequiredClass(className);
+		Class<?> clazz = loadRequiredClass(className, testMethod.getDeclaringClass().getClassLoader());
```
sbrannen, line 111:
```suggestion
		Class<?> clazz = loadRequiredClass(className, testClass.getClassLoader());
```

Please pass the `testClass` argument in `findFactoryMethod()` to this method and use that instead of the declaring class of the test method.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -829,7 +829,7 @@ public static Try<Class<?>> tryToLoadClass(String name, ClassLoader classLoader)
 			}
 
 			// Fallback to standard VM class loading
-			return classLoader.loadClass(trimmedName);
+			return Class.forName(trimmedName, false, classLoader);
```
sbrannen, line 832: Please move this to a separate enhancement PR.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -838,7 +838,7 @@ private static Class<?> loadArrayType(ClassLoader classLoader, String componentT
 
 		Class<?> componentType = classNameToTypeMap.containsKey(componentTypeName)
 				? classNameToTypeMap.get(componentTypeName)
-				: classLoader.loadClass(componentTypeName);
+				: Class.forName(componentTypeName, false, classLoader);
```
sbrannen, line 841: Please move this to a separate enhancement PR.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
```diff
@@ -63,11 +64,15 @@ private Method getConditionMethod(String fullyQualifiedMethodName, ExtensionCont
 		String[] methodParts = ReflectionUtils.parseFullyQualifiedMethodName(fullyQualifiedMethodName);
 		String className = methodParts[0];
 		String methodName = methodParts[1];
-		Class<?> clazz = ReflectionUtils.tryToLoadClass(className).getOrThrow(
+		Class<?> clazz = ReflectionUtils.tryToLoadClass(className, getClassLoader(context)).getOrThrow(
```
sbrannen, line 67:
```suggestion
		Class<?> clazz = ReflectionUtils.tryToLoadClass(className, testClass.getClassLoader()).getOrThrow(
```

The test class is required to be present.

In fact, it's already accessed in the above if-block before this line is invoked.

So, please assign the test class to a local variable, use it within the if-block, and use it again here.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/MethodBasedCondition.java
```diff
@@ -63,11 +64,15 @@ private Method getConditionMethod(String fullyQualifiedMethodName, ExtensionCont
...
 			cause -> new JUnitException(format("Could not load class [%s]", className), cause));
 		return findMethod(clazz, methodName);
 	}
 
+	private ClassLoader getClassLoader(ExtensionContext context) {
+		return context.getTestClass().map(Class::getClassLoader).orElseGet(ClassLoaderUtils::getDefaultClassLoader);
+	}
+
```
sbrannen, line 75:
```suggestion
```

This method is no longer needed.

sbrannen (changes requested):
Thanks for the PR!

chrisr3:
> Do you have time to make the requested changes?

I'll try to find time today :+1+:

marcphilipp:
@chrisr3 Could you please also add some tests that use a custom class loader for this (see `OpenTest4JAndJUnit4AwareThrowableCollectorTests` for an example)?

kriegfrj:
> @chrisr3 Could you please also add some tests that use a custom class loader for this (see `OpenTest4JAndJUnit4AwareThrowableCollectorTests` for an example)?

You might also be able to use one like what I used in #2106, if it's helpful. Been a while since I looked at it.

sbrannen:
> You might also be able to use one like what I used in #2106, if it's helpful. Been a while since I looked at it.

You may also find the `CustomTypeClassLoader` that I introduced in [this commit](https://github.com/junit-team/junit5/commit/da6a0cae9beff18227c2dd21a12cd490c7bcfd92#diff-02d2f34d3a384020088ebf7add0c0e09d951747d4e20fadada7d1507db04869dR1934-R1946) useful.

sbrannen:
> _Sigh_, I am _not_ going to have time to look at this today after all.

No worries, @chrisr3.

It seems that we will not be releasing 5.10 M1 over the weekend but rather (hopefully) before the end of next week.

So if you can get to it in the first few days of next week, that would be fine.

sbrannen:
> > You might also be able to use one like what I used in #2106, if it's helpful. Been a while since I looked at it.
> 
> You may also find the `CustomTypeClassLoader` that I introduced in [this commit](https://github.com/junit-team/junit5/commit/da6a0cae9beff18227c2dd21a12cd490c7bcfd92#diff-02d2f34d3a384020088ebf7add0c0e09d951747d4e20fadada7d1507db04869dR1934-R1946) useful.

Actually, please use the `FilteringClassLoader` that I just extracted/introduced in 75c0bd07fd8b7b6bab658c403cbad655e9764a08.

If the `FilteringClassLoader` does not suit your needs, please let me know so that we can improve it.

chrisr3:
> If the `TestClassLoader` does not suit your needs, please let us know so that we can improve it.

Hmmm... I've written a _lot_ of fiddly class-loading Java code over the years, and that experience has taught me that the JVM maintains an internal cache of classes that it has already loaded. The JVM only invokes `ClassLoader.loadClass()` if it cannot find a class within that cache, and if loading class _X_ requires also loading class _Y_, then it will use `X.class.getClassLoader()` to locate class _Y_.

However, no classes returned by `TestClassLoader.loadClass()` can ever satisfy
```java
X.class.getClassLoader() instanceof TestClassloader
```
which means the JVM can bypass `TestClassLoader`'s filter in some cases. And even _that_ is assuming that the JVM hasn't already loaded those classes into its cache in the first place.

I need to think about how to test this :thinking:.

sbrannen:
> However, no classes returned by `TestClassLoader.loadClass()` can ever satisfy
> 
> ```java
> X.class.getClassLoader() instanceof TestClassloader
> ```

When I add this:

```java
System.err.println(customType.getClassLoader() instanceof TestClassLoader);
```

to `org.junit.platform.commons.util.ReflectionUtilsTests.findMethodByParameterNamesWithWithCustomTypeFromDifferentClassLoader()`, I get this:

```
true
```

So I don't understand what you are claiming.

sbrannen:
> I need to think about how to test this 🤔.

I updated one of our tests using `TestClassLoader` to check the `ClassLoader` in 1c9c018007fa25824e43979f0f2a57061fcabbfe.

From that you can see that `TestClassLoader` is the expected `ClassLoader` for the top-level type `CustomType`; however, the `ClassLoader` for `NestedType` is in fact `null`.

So maybe that's what you were hinting at.

In any case, I still think `TestClassLoader` is sufficient for our testing purposes. We don't want to spend a lot of time making this internal testing utility production-ready.

As for how to "test this", statements like the following from that test are sufficient in my opinion.

```java
assertThat(parameterType).isNotEqualTo(CustomType.NestedType.class);
```

chrisr3:
> > However, no classes returned by `TestClassLoader.loadClass()` can ever satisfy
> > ```java
> > X.class.getClassLoader() instanceof TestClassloader
> > ```
> 
> When I add this:
> 
> ```java
> System.err.println(customType.getClassLoader() instanceof TestClassLoader);
> ```
> 
> to `org.junit.platform.commons.util.ReflectionUtilsTests.findMethodByParameterNamesWithWithCustomTypeFromDifferentClassLoader()`, I get this:
> 
> ```
> true
> ```

Oh my, I see what you've done... Try executing this:
```java
var customType1 = customTypeClassLoader.loadClass(customTypeName);
var customType2 = customTypeClassLoader.loadClass(customTypeName);
```

sbrannen:
> Oh my, I see what you've done... Try executing this:
> 
> ```java
> var customType1 = customTypeClassLoader.loadClass(customTypeName);
> var customType2 = customTypeClassLoader.loadClass(customTypeName);
> ```

Indeed... it's not production-ready.

```
java.lang.LinkageError: loader org.junit.platform.commons.test.TestClassLoader @bcec361 attempted duplicate class definition for org.junit.platform.commons.util.classes.CustomType. (org.junit.platform.commons.util.classes.CustomType is in unnamed module of loader org.junit.platform.commons.test.TestClassLoader @bcec361, parent loader 'app')
```

But I'm not worried about that.

None of our tests attempt to load the same class twice. So I can live with that.

`TestClassLoader` is a _quick and dirty_ solution that meets our current needs. We have no plans to publish it, and if we later find it doesn't meet our needs we can then refactor it to meet our needs.

sbrannen:
FWIW, I pushed a _fix_ for that `LinkageError` in 4b45db09a1c048837fac17026fa14deb7e61e341... just in case we end up having a test case that attempts to load the same class via reflection multiple times.

sbrannen:
@chrisr3, if you run into any challenges with testing the changes, feel free to finish up the PR without the tests, and someone in the team will add tests afterwards.

Depending on the use case, you might need to use the `EngineTestKit` combined with a custom `ClassLoader` (maybe the `TestClassLoader`) to actually test the changes in an integration test.

Cheers,

Sam

chrisr3:
> FWIW, I pushed a _fix_ for that `LinkageError` in [4b45db0](https://github.com/junit-team/junit5/commit/4b45db09a1c048837fac17026fa14deb7e61e341)... just in case we end up having a test case that attempts to load the same class via reflection multiple times.

i would actually have suggested using `ClassLoader.loadClass(String, boolean)` as model.

sbrannen:
> i would actually have suggested using `ClassLoader.loadClass(String, boolean)` as model.

Thanks for the tip, @chrisr3. 👍 

I revised the implementation in f16982e56172a51cda838157df412724f53b6196.

sbrannen:
Hi @chrisr3,

I see you've made all of the requested changes. 👍

So, all that's missing is:

- tests
- release notes entry

Since you mentioned that you didn't have time to address #3290 (https://github.com/junit-team/junit5/issues/3290#issuecomment-1540249794), do you think you'll have time to finish this PR this week?

If not, that's absolutely not a problem. Someone in the team can take over the PR and finish it.

Please just let us know about your availability for completing this issue.

Thanks!

chrisr3:
I think this test has an error:
```diff
diff --git a/platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java b/platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
index e8bc60ff01..877139f3cd 100644
--- a/platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/commons/util/ReflectionUtilsTests.java
@@ -1049,10 +1049,9 @@ class ReflectionUtilsTests {
 
                                var parameterTypes = method.getParameterTypes();
                                assertThat(parameterTypes).extracting(Class::getName).containsExactly(nestedTypeName);
-                               Class<?> parameterType = parameterTypes[0].getClass();
+                               Class<?> parameterType = parameterTypes[0];
                                assertThat(parameterType).isNotEqualTo(CustomType.NestedType.class);
-                               // The ClassLoader is null because of the behavior of our TestClassLoader.
-                               assertThat(parameterType.getClassLoader()).isNull();
+                               assertThat(parameterType.getClassLoader()).isEqualTo(customTypeClassLoader);
                        });
                }
        }
```

sbrannen:
> I think this test has an error:

Oops... that's embarrassing. 😮 

Good catch!

I'll fix it right away.

chrisr3:
It's no good - I've run out of time to look at this PR further :disappointed:.

sbrannen:
> It's no good - I've run out of time to look at this PR further 😞.

Don't worry about it, @chrisr3.

We'll try to take it from here.

Thanks for bringing the issue to our attention, creating the PR, assisting with `TestClassLoader` improvements, etc.! 👍

sbrannen:
**Update**:

Work on this issue is complete; however, I cannot push to `main` due to a `protected branch hook` blocking me.

We will investigate the cause for that. In the interim, the completed work can be viewed here:

https://github.com/junit-team/junit5/compare/main...sbrannen:junit5:issues/3279_and_3292_custom_ClassLoader_arrangements

PR #3293, opened 2023-05-09
[params] Use test class' classloader to load parameters

### LINKED ISSUE #3291: `@ParameterizedTest` fails to convert `String` to `Class` with custom `ClassLoader`
## Overview

This is a follow up to #3279.

See also: https://github.com/junit-team/junit5/issues/3279#issuecomment-1531569327

The assumption is that the following test case fails when running within OSGi.

```java
package example;

import static org.assertj.core.api.Assertions.assertThat;

import org.junit.jupiter.params.ParameterizedTest;

class StringToClassConversionTests {

	@ParameterizedTest
	@ValueSource(strings = "example.ClassFromBundle")
	void test(Class<?> clazz) {
		assertThat(clazz).isNotNull();
	}
}

class ClassFromBundle {
}
```

kriegfrj:
Fixes #3291.

## Overview

As noted in #3291, to support varied classloader environments, the argument converter/loader logic should ideally be classloader aware. The best thing to do is to use the test class' classloader to load any classes. This PR attempts to do this.

The declaring class (and hence its classloader) is available through the `ParameterContext` (`getParameter().getDeclaringExecutable().getDeclaringClass()`). However, the reference to the `ParameterContext` is lost in the `SimpleArgumentConverter` abstraction, from which `DefaultArgumentConverter` currently inherits. Hence I have changed `DefaultArgumentConverter` to implement `ArgumentConverter` directly. This technical represents a major (breaking) change to the package `org.junit.jupiter.params.converter`, because `DefaultArgumentConverter` is a public class in that package. However, because it's been annotated `INTERNAL` I don't think that that matters (any downstream users that have inherited from it probably deserve what they get... :smile:).

I have tried to implement it as best as I can so that if (for whatever reason) the `DefaultArgumentConverter` doesn't have access to the parameter context, then it will fall back to its current behaviour. If one is present, then it will use the test class' classloader.

The implementation was complicated by the fact that `DefaultArgumentsAccessor` invokes `DefaultArgumentsConverter.convert()` directly, and  `DefaultArgumentsAccessor` (and the `ArgumentsAccessor` API provides no `ParameterContext`. So there was no way to pass the necessary information in to `convert()`. I have added the `ParameterContext` as a constructor parameter to `DefaultArgumentsAccessor` to work around this. It might have been cleaner (less changes to the tests) to add a second constructor instead and have the original delegate to the new one, 

The current implementation passes all current tests (at least, it did on my machine!) but I haven't added new tests for the enhanced class-loader-sensitive functionality as yet. I want to make sure that we're happy with this direction before I invest that time. There are a couple of different ways that this cat could be skinned (no offence to cats intended) and the interaction between the various default implementations is a bit complicated.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestMethodContext.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestMethodContext.java
@@ -253,7 +253,7 @@ class ParameterizedTestMethodContext {
 
 		@Override
 		public Object resolve(ParameterContext parameterContext, Object[] arguments, int invocationIndex) {
-			ArgumentsAccessor accessor = new DefaultArgumentsAccessor(invocationIndex, arguments);
+			ArgumentsAccessor accessor = new DefaultArgumentsAccessor(invocationIndex, parameterContext, arguments);
 			try {
 				return this.argumentsAggregator.aggregateArguments(accessor, parameterContext);
 			}
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
@@ -18,6 +18,7 @@ import java.util.Collections;
 import java.util.List;
 
 import org.apiguardian.api.API;
+import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.params.converter.DefaultArgumentConverter;
 import org.junit.platform.commons.util.ClassUtils;
 import org.junit.platform.commons.util.Preconditions;
@@ -37,12 +38,14 @@ public class DefaultArgumentsAccessor implements ArgumentsAccessor {
 
 	private final int invocationIndex;
 	private final Object[] arguments;
+	private final DefaultArgumentConverter converter;
 
-	public DefaultArgumentsAccessor(int invocationIndex, Object... arguments) {
+	public DefaultArgumentsAccessor(int invocationIndex, ParameterContext context, Object... arguments) {
 		Preconditions.condition(invocationIndex >= 1, () -> "invocation index must be >= 1");
 		Preconditions.notNull(arguments, "Arguments array must not be null");
 		this.invocationIndex = invocationIndex;
 		this.arguments = arguments;
+		this.converter = new DefaultArgumentConverter(context);
 	}
 
 	@Override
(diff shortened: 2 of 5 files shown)


sbrannen:
Superseded by commit f6e73acdbe9133bdd4286241c88f707a3a3228cc

sbrannen:
Hi @kriegfrj,

Thanks for the PR!

That's quite an extensive write up, so thanks for sharing your thought process and rationale as well.

> Hence I have changed `DefaultArgumentConverter` to implement `ArgumentConverter` directly. 

That was the same conclusion I came to.

> This technical represents a major (breaking) change to the package `org.junit.jupiter.params.converter`, because `DefaultArgumentConverter` is a public class in that package. However, because it's been annotated `INTERNAL` I don't think that that matters (any downstream users that have inherited from it probably deserve what they get... 😄).

Exactly.

> a couple of different ways that this cat could be skinned

Indeed, I took a different approach for some of the aspects. Feel free to take a look at [my implementation](https://github.com/junit-team/junit5/commit/f6e73acdbe9133bdd4286241c88f707a3a3228cc), and if you have any questions or comments don't hesitate to speak up.

> The current implementation passes all current tests (at least, it did on my machine!) 

Did you run a full Gradle build?

I'm asking because "updating" `ArgumentsAccessorKotlinTests` was arguably the toughest challenge for me, requiring research into Kotlin features like declaring `Object...` var-args in Kotlin and passing them to a Java method that accepted them. Getting Mockito to work in Kotlin was also a challenge (for me), because `when` is a keyword in Kotlin. 😱 

On the plus side, I learned a few things about Kotlin syntax, etc.

Cheers,

Sam

kriegfrj:
> That's quite an extensive write up, so thanks for sharing your thought process and rationale as well.
> 

I apologise - it happened by accident. I started with a brief comment, then I checked the code to see if there might be any hiccups, added a couple more comments, check the code again... before I knew it I had a complete implementation *and* the doc explaining how I got there...

> > Hence I have changed `DefaultArgumentConverter` to implement `ArgumentConverter` directly.
> 
> That was the same conclusion I came to.
>

Encouraging to know we're both on the same page! :smile: 

> I took a different approach for some of the aspects. Feel free to take a look at [my implementation](https://github.com/junit-team/junit5/commit/f6e73acdbe9133bdd4286241c88f707a3a3228cc), and if you have any questions or comments don't hesitate to speak up.

LGTM. I'm glad you went with your own implementation - it seemed more compact than mine.

The only difference of any real consequence (that I can see) the default behaviour - I tried to use `tryToLoadClass(String)` as the default, instead of manually calling `ClassLoaderUtils.getDefaultClassLoader()`. My reasoning was that if ever you want to change the default behaviour, then you only have to change `tryToLoadClass(String)`. Admittedly, also made the code a bit more ugly, so it was a tradeoff.

Issue #3012, opened 2022-08-24
Introduce `@SelectMethod` for selecting methods in `@Suite` classes


dominikozi:
Feature request to add support to filtering test methods by name in test suites. This could be done by annotation similar to @SelectClasses where user could specify list of method names. This could then be used when tests to run are determined during runtime and there is no possibility to add tags.

dominikozi:
Example: 

There are two test classes

```
public class A(){
	@Test
	public void firstA(){...}
	@Test
	public void secondA(){...}
}

public class B(){
	@Test
	public void firstB(){...}
	@Test
	public void secondB(){...}
}
```

I need to generate a test suite during execution of my program that will only run some of avabile methods(which will be chosen during runtime), for example : firstA() and secondB() or secondA(), firstB(). In this scenario tagging will not work as I can not modify test classes code.

If this gets implemented it could for **example** look like this: 

@SelectMethods ({"com.example.junit5.examples.A.firstA", 
"com.example.junit5.examples.B.secondB"})

Mohammed-2002:
Hey guys! I'm a college student an I really want to solve  issue 3012  "Allow filtering methods by name in test suites" to improve my skills in Java and Kotlin, I just don't know which class(es) is/are responsible for filtering classes during running suits test, in other terms which classes be called when looking for the selected classes @selectclasses()? and running the selected tests?, I want to see it as an example so I can start on that issue, any one who can give me some explanation? I would be very thankful!

mpkorstanje:
@Mohammed-2002 if you do a usage search for the `SelectClasses` annotation you can find all the code that directly touches it. Then sprinkle a few breakpoints around and run the tests. That should get  you most if not all of the interactions.

So in this case you would have to look here:

https://github.com/junit-team/junit5/blob/3ba144efdb38181d13132a552988e69c35c619a2/junit-platform-suite-commons/src/main/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilder.java#L149-L151

Mohammed-2002:
thank you for your answer! 
i found a solution for this problem but i don't know if it is the best. see the code below. 
import org.junit.platform.engine.TestExecutionResult;
import org.junit.platform.engine.TestExecutionResult.Status;
import org.junit.platform.engine.TestExecutionListener;
import org.junit.platform.engine.TestIdentifier;

import java.util.Arrays;
import java.util.List;

public class CustomTestExecutionListener implements TestExecutionListener {

[...]

PR #3296, opened 2023-05-11
Allow filtering methods by name in test suites

Mohammed-2002:
Allowing selecting methods in test suites
    
    This commit adds two annotations to the ‘junit-platform-suite-api’,
    and also includes three types of tests to ensure that the annotations
    functionality is working correctly.
    
    The two annotations allow the user to select methods in suite testing
    based on the name, the class where the method is located (optional)
    and the input parameters (optional).
    
    The three added tests test the operation of the two added annotations,
    examining three possible situations for using the added annotations.
   
   Resolve: #3012
---


junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/SelectMethod.java
```diff
@@ -0,0 +1,38 @@
...
+
+	/**
+	 * The parameter types of the method to select, in the form "int,String".
+	 */
+	String parameterTypes() default "";
+
+	/**
+	 * The return type of the method to select.
+	 */
+	Class<?> returnType() default void.class;
```
mpkorstanje, line 36: This name and java doc comment don't match with it's function. If you're in need of inspiration, have a look at the existing Javadoc on `DiscoverySelectors.selectMethod(String className, String methodName, String methodParameterTypes)`. 

This field also can not have a default value. And should be listed first so that the fields match order of the element in a fully qualified method name.

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/SelectMethod.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+
+public @interface SelectMethod {
```
mpkorstanje, line 21: Missing and elsewhere: Javadoc and API Guardian annotation. All public classes should have these. Have a look at the surrounding classes for hints about the style and language to use.

junit-platform-suite-commons/src/main/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilder.java
```diff
@@ -201,6 +207,16 @@ public LauncherDiscoveryRequest build() {
...
 	}
 
+	private List<DiscoverySelector> selectMethods(SelectMethod... methods) {
+		List<DiscoverySelector> selectors = new ArrayList<>();
+
+		for (SelectMethod selectMethod : methods) {
+			selectors.add(DiscoverySelectors.selectMethod(selectMethod.returnType().getName(), selectMethod.name(),
+				selectMethod.parameterTypes()));
+		}
+		return selectors;
```
mpkorstanje, line 217: Consider using the streaming API to clean up this mapping operation.

The returned list should also not contain duplicates. The `.distinct()` can help with this.

mpkorstanje, line 217: For each class you should also add an entry to `this.selectedClassNames`. Otherwise the classes may be filtered out.

You should also write a test for this.

platform-tests/src/test/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilderTests.java
```diff
@@ -214,6 +217,88 @@ class Suite {
...
+	void selectMethodsOnlyByName() {
+		class TestClass {
+			public void testMethod() {
+			}
+
+			public void anotherTestMethod() {
+			}
+		}
+		@SelectMethods({ @SelectMethod(name = "testMethod"), @SelectMethod(name = "anotherTestMethod") })
+		class Suite {
```
mpkorstanje, line 230: You can repeat repeatable annotations directly without the outer `@SelectMethods`.

platform-tests/src/test/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilderTests.java
```diff
@@ -214,6 +217,88 @@ class Suite {
...
+		}
+		LauncherDiscoveryRequest request = builder.suite(Suite.class).build();
+		List<MethodSelector> selectors = request.getSelectorsByType(MethodSelector.class);
+		assertEquals(2, selectors.size());
+		assertTrue(selectors.stream().anyMatch(selector -> selector.getMethodName().equals("testMethod")));
+		assertTrue(selectors.stream().anyMatch(selector -> selector.getMethodName().equals("anotherTestMethod")));
+	}
```
mpkorstanje, line 237: You've got two problems going on here.

1. Your assertions are quite complex. They're quite complex because your test case is quite complex.
  
   Start with the simplest case of having a single `@SelectMethod` annotation for a single class with a single zero-arg method. Then from that expand in different directions of complexity. Multiple methods, single method with one argument, two arguments. 

   To make this easier you can use the `exactlyOne` to unpack the lists into a single object.

2. Your assertions do not verify that all 3 attributes were copied over into the `MethodSelector`. You should verify the class name, method name, and parameter types.

platform-tests/src/test/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilderTests.java
```diff
@@ -214,6 +217,88 @@ class Suite {
 		assertEquals(TestCase.class, exactlyOne(selectors).getJavaClass());
 	}
 
+	@Test
+	void selectMethodsOnlyByName() {
```
mpkorstanje, line 221: It is not possible to select methods only by name. It would be prudent to build a snapshot version of your branch and try it out in a test project. You should try to use (and break) the code you're writing.

mpkorstanje (changes requested):
You've got the right structure down. Now it is time for details, like making sure the code actually works. :wink: 

You can build a snapshot version of your project (see the CONTRIBUTING.md) and use that in a test project. You'll see that things won't quite work as you expect them to.

Also do pay attention to the guidance in the CONTRIBUTING.md and the template that was shown when you created the PR. You may want to create a second PR without submitting it to read it in full.

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/SelectMethod.java
```diff
@@ -0,0 +1,38 @@
...
+
+	/**
+	 * The name of the method to select.
+	 */
+	String name();
+
+	/**
+	 * The parameter types of the method to select, in the form "int,String".
+	 */
+	String parameterTypes() default "";
```
mpkorstanje, line 31: As an after thought. Since this is going to be a list of parameters, perhaps we can use  `Class<?>[]` and use `ClassUtils.nullSafeToString` to create the `methodParameterTypes` for the `MethodSelector`. One of the `MethodSelector` constructors does the same thing.

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/SelectMethod.java
Mohammed-2002: API has a default input called `since`, in most select annotations is  `since = 1.10`, but I'm not sure which version I have to type for the annotations I made.

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/SelectMethod.java
Mohammed-2002: I've only found two methodSelector constructors that take methodParameterTypes as input, both take it as `string`, not as `Class<?>[]`. Am I supposed to convert `Class<?>[]` to `String` myself with `ClassUtils.nullSafeToString`?

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/SelectMethod.java
mpkorstanje: The current version of the JUnit Platform is 1.9.3. So the next will be 1.10.

https://github.com/junit-team/junit5#dependency-metadata

mpkorstanje: And it looks like it's 1.10 because some experimental features got made stable. Yours should be marked as experimental. That way we can make breaking changes should we get it wrong.

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/SelectMethod.java
mpkorstanje: For now I think we can use `nullSafeToString`. Once everything works properly and is tested we can look at adding a new overloaded method to `MethodSelector`.

junit-platform-suite-commons/src/main/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilder.java
Mohammed-2002: It is possible that the name of the class is already added to selectedClassNames, in this case a class would appear twice, is that a bad thing?

mpkorstanje: No. Because the patterns are used in an include filter duplicates aren't a problem. It would be better to use a set to store the selected class name though.

Mohammed-2002:
@mpkorstanje Please have a look at the code, I've updated the two interfaces @SelectMethod and @SelectMethods. i also updated the class but the first test (only one too) fails! no selectors are detected.

platform-tests/src/test/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilderTests.java
```diff
@@ -214,6 +217,22 @@ class Suite {
...
+		@SelectMethod(methodClass = TestClass.class, methodName = "testMethod")
+		class Suite {
+		}
+		LauncherDiscoveryRequest request = builder.suite(Suite.class).build();
+		List<MethodSelector> selectors = request.getSelectorsByType(MethodSelector.class);
+		exactlyOne(selectors);
+		assertTrue(selectors.stream().anyMatch(selector -> selector.getMethodName().equals("testMethod")
+				&& selector.getClassName().equals("TestClass") && selector.getMethodParameterTypes().isEmpty()));
+	}
+
```
Mohammed-2002, line 235: I removed all the tests to start writing new tests but this test fails and I don't really understand why.

Mohammed-2002: @after debugging I notice that if I create and select two methods this test passes, the test fails only if there is just one '@SelectMethod' annotation.

Mohammed-2002: After changing the assert_function to 
`assertTrue(selectors.stream().anyMatch(selector -> selector.getMethodName().equals("testMethod")
				&& selector.getClassName().equals(TestClass.class.getName()) && selector.getMethodParameterTypes().equals("")));` 
the test now only fails if there is just one annotation of `@SelectMethod`

Mohammed-2002: After debugging for almost an hour,  I finally know what is going on here, the code below just checks if there are multiple annotations of `@SelectMethod`, if there is only one it returns an `emptyList.`

`findAnnotationValues(suiteClass, SelectMethods.class, SelectMethods::value)
				.map(this::selectMethods)
				.ifPresent(this::selectors);`

To solve this problem, I added this code to the suite constructor: 

`		findAnnotation(suiteClass, SelectMethod.class)
				.map(this::selectMethods)
				.ifPresent(this::selectors);`

Is this the best solution?

mpkorstanje: I think using `findRepeatableAnnotations` would be better. The documentation on `findAnnotation` and `findRepeatableAnnotations`  should explain why. 

As an exercise you could also continue with `findAnnotation`. When using `findAnnotation` the next test case you implement, one with multiple `@SelectMethod` annotations will break.

mpkorstanje: As an aside, the `MethodSelector` implements `hasCode()` and `equals()`. So you can use `assertEquals`. This will also give you a nicer error message than using `anyMatch`.

For example:
```
assertEquals(DiscoverySelectors.selectMethod(TestClass.class, "testMethod"), exactlyOne(selectors));

assertEquals(DiscoverySelectors.selectMethod(TestClass.class, "firstMethod"), selectors.get(0));
assertEquals(DiscoverySelectors.selectMethod(TestClass.class, "secondMethod"), selectors.get(1));
```

Mohammed-2002: The test case I implemented, one with multiple @SelectMethod annotations passed tough, this is because I used 
```
findAnnotationValues(suiteClass, SelectMethods.class, SelectMethods::value)
				.map(this::selectClasses)
				.ifPresent(this::selectors);
```
In combination with 
`findAnnotation(suiteClass, SelectMethod.class) .map(this::selectMethods) .ifPresent(this::selectors);`

But now I just have this and it works!
		```
findRepeatableAnnotations(suiteClass,SelectMethod.class)
				.stream()
				.map(this::selectMethods)
				.forEach(this::selectors);
```
PS: `findRepeatableAnnotations` does not return an `optional`, instead it returns an empty list if there are no annotations found.

**Comment by Mohammed-2002** [comment]
```
	@Test
	void selectOneMethodWithOneParameter() {
		class TestClass {
			public void testMethod(int i) {
			}
		}
		@SelectMethod(methodClass = TestClass.class, methodName = "testMethod")
		class Suite {
		}
		LauncherDiscoveryRequest request = builder.suite(Suite.class).build();
		List<MethodSelector> selectors = request.getSelectorsByType(MethodSelector.class);
		exactlyOne(selectors);
		assertEquals(DiscoverySelectors.selectMethod(TestClass.class, "testMethod"), exactlyOne(selectors));
	}
```
a test like this passed even if the selected method does not exist (there is a parameter type), is the responsibility to check if a method really exists for `SuiteLauncherDiscoveryRequestBuilder`  or for another class such as the 'LauncherDiscoveryRequest'?

**Comment by mpkorstanje** [comment]
The purpose of the `SuiteLauncherDiscoveryRequestBuilder` is to map the annotations to discovery selectors. So it doesn't really matter if the method exists or not. 

I do suppose it is prudent to use the actual method when constructing the expected method selector. This leaves less room for error.

```
        class TestClass {
            public void test(String a, String b) {
            }
        }
        @SelectMethod(methodClass = TestClass.class, methodName = "testMethod", methodParameterTypes = {String.class, String.class})
        class Suite {
        }
        var request = builder.suite(Suite.class).build();
        var selectors = request.getSelectorsByType(MethodSelector.class);
        var expectedMethod = TestClass.class.getMethod("test", String.class, String.class);
        assertEquals(DiscoverySelectors.selectMethod(TestClass.class, expectedMethod), exactlyOne(selectors));
```

**Comment by Mohammed-2002** [comment]
@mpkorstanje  I think everything is completed now and the pull request is ready to be merged but I may have forgotten-misunderstood some details. Let me pls know if there is anything left to do.

File: junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/SelectMethod.java [hunk 9]
```diff
@@ -0,0 +1,50 @@
...
+	/**
+	 * the name of the method to select; never null or blank
+	 */
+	String methodName();
+
+	/**
+	 * The method parameter types as an array of Class objects;
+	 * never null, not to be entered if the function has no input parameters.
+	 */
+	Class<?>[] methodParameterTypes() default {};
```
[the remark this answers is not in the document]
mpkorstanje: Unfortunately I also just found a rather annoying note in the documentation of `DiscoverySelectors.selectMethod`:

https://github.com/junit-team/junit5/blob/0ee6e09a41e944feecb2df2f4845c86f4aea1baa/junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java#L516-L518

Because the exact syntax depends on the underlying test engine, we can not use a list of classes here.

mpkorstanje (commented):
So this looks good. Now I it is time for some polishing!

 * Carefully review [CONTRIBUTING.md](https://github.com/junit-team/junit5/blob/main/CONTRIBUTING.md) for anything you may have missed. 
  * Update your commit message. It is currently out of date.
  * Update your PR description on Github to follow the same format as every other PR. You can find [the Markdown template to use here](https://raw.githubusercontent.com/junit-team/junit5/main/.github/PULL_REQUEST_TEMPLATE.md). This will help the JUnit Team see that you've checked your work against the [Definition of Done](https://github.com/junit-team/junit5/wiki/Definition-of-Done).
  * I've left more code comments below.
    
Polishing is unfortunately the most tedious part of writing code. But it is also necessary to make sure that people after us can contribute just as smoothly.

Also note: I'm not part of the JUnit 5eam so I will not merge your PR. But I did contribute the Suite engine to JUnit 5 so once this PR is done I will be happy to recommend it to the JUnit team.

junit-platform-suite-commons/src/main/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilder.java
```diff
@@ -201,6 +210,17 @@ public LauncherDiscoveryRequest build() {
 		return delegate.build();
 	}
 
+	private List<MethodSelector> selectMethods(SelectMethod... methods) {
+		Arrays.stream(methods).map(method -> method.methodClass().getName()).distinct().forEach(
+			this.selectedClassNames::add);
```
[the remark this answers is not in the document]
Mohammed-2002: This method does not exist anymore, I changed it to a static method in class `AdditionalDiscoverySelectors` as follows
```

	static MethodSelector selectMethod(Class className, String methodName, String methodParameterTypes) {
		Preconditions.notBlank(methodName, "Method name must not be null or blank");
		Preconditions.notNull(methodName, "Method parameter types must not be null");
		return DiscoverySelectors.selectMethod(className,methodName,methodParameterTypes);
	}

```
Which specific test case do u mean here?

platform-tests/src/test/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilderTests.java
```diff
@@ -214,6 +217,107 @@ class Suite {
...
+		class Suite {
+		}
+		LauncherDiscoveryRequest request = builder.suite(Suite.class).build();
+		List<MethodSelector> selectors = request.getSelectorsByType(MethodSelector.class);
+		exactlyOne(selectors);
+		assertEquals(DiscoverySelectors.selectMethod(TestClass.class, "testMethod", "int"), exactlyOne(selectors));
+	}
+
+	@Test
+	void selectTwoMethodsWithOneParameter() {
```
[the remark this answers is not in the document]
Mohammed-2002: You are right! it's an unnecessary test!!

mpkorstanje:
@Mohammed-2002 I think you pushed the wrong code.

Mohammed-2002:
Im going to make a new PR and delete this one, it will be easier for me.


Op za 13 mei 2023 20:36 schreef M.P. Korstanje ***@***.***>:

mpkorstanje:
If it helps you can use the reflog to find the commit you lost. I've also got it on my machine from the last review if you can't get it back with the reflog.

https://rewind.com/blog/how-to-restore-deleted-branch-commit-git-reflog/

sbrannen:
- superseded by #3303

PR #3321, opened 2023-05-23
Add thread-safety details to TempDirFactory documentation

### LINKED ISSUE #3312: Improve documentation for `TempDirFactory` API contract
The API Javadoc of `TempDirFactory` does not define what the threading model is. This forces the implementer to make assumptions. See Effective Java, 3rd Edition Item 82: Document thread safety.

I would assume something like this:
- An instance is created whenever a `@TempDir` value is requested for the first time.
- `#createTempDirectory` can be called concurrently by multiple threads
- `#close` will not be called before the last call to `#createTempDirectory` has returned
- `#close` will be called only once
- `#close` will be called when the engine is shut down

An alternative would be to define a general threading model to be supported by [extensions](https://junit.org/junit5/docs/snapshot/user-guide/#extensions) and link to this.

## Deliverables

- [ ] explicitly define the threading model to be supported by `TempDirFactory` implementations

scordio:
Closes #3312.

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.0-RC1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.0-RC1.adoc
@@ -39,10 +39,10 @@ JUnit repository on GitHub.
 
 ==== New Features and Improvements
 
-* ❓
+* Add thread-safety details to `TempDirFactory` documentation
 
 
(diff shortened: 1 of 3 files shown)


marschall:
@scordio this helps a lot and tells me everything I need to know to write an implementation

marcphilipp (approved):

marcphilipp:
@scordio Thank you!

Issue #3326, opened 2023-05-26
Introduce `selectMethod` & `selectedNestedMethod` discovery selectors that accept parameter types as var-args


sbrannen:
## Overview

Introduce overloaded `selectMethod()` and `selectNestedMethod()` variants in `DiscoverySelectors` that accept a var-args `Class<?>...` argument of parameter types.

These new methods are type-safe alternatives to providing the names of parameter types as a comma-delimited string like in `DiscoverySelectors.selectMethod(Class<?>, String, String)`.

For background information, see https://github.com/junit-team/junit5/issues/3012#issuecomment-1564320887.

## Related Issues

- #3012 
- #3324
- #3334 
- #3335 

## Deliverables

- [x] Introduce `DiscoverySelectors.selectMethod()` variants.
- [x] Introduce `DiscoverySelectors.selectNestedMethod()` variants.
- [x] Document in Release Notes.

PR #3327, opened 2023-05-27
Introduce `DiscoverySelectors.selectMethod(Class<?>, String, Class<?>...)`

sbrannen:
## Overview

This is a _draft_ PR for #3326.

Still need to introduce a `selectNestedMethod()` overload that accepts var-arg parameter types.

---

### Definition of Done


sbrannen:
@marcphilipp, as discussed, here is the draft PR for #3326.

It combines several sets of changes which I included in a single commit thus far. 

For the sake of time, I opted not to introduce separate commits since doing so is a bit cumbersome. Though, if that's too cumbersome to review, let me know, and I can introduce separate commits for the following (which this PR currently contains).

- Restructure `DiscoverySelectorsTests`: use `@Nested` test classes for logical grouping.
- Refactor `MethodSelector` constructors and `DiscoverySelectors.selectMethod()` methods so that the `ClassLoader` argument is first.
- Introduce `getParameterTypeNames()` and deprecate `getMethodParameterTypes()` in `MethodSelector` and `NestedMethodSelector`.
- Introduce `DiscoverySelectors.selectMethod(Class<?>, String, Class<?>...)`, `DiscoverySelectors.selectMethod(String, String, Class<?>...)`, and `DiscoverySelectors.selectMethod(ClassLoader, String, String, Class<?>...)`.

sbrannen:
Related discussion on the actual commit instead of this PR: https://github.com/sbrannen/junit5/commit/b31465d3ca099f453fbdddf2605cf1e06ca164cf#r115749361

/cc @kriegfrj

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -1361,7 +1361,7 @@ public static Optional<Method> findMethod(Class<?> clazz, String methodName, Str
 		return findMethod(clazz, methodName, resolveParameterTypes(clazz, methodName, parameterTypeNames));
 	}
 
-	private static Class<?>[] resolveParameterTypes(Class<?> clazz, String methodName, String parameterTypeNames) {
+	public static Class<?>[] resolveParameterTypes(Class<?> clazz, String methodName, String parameterTypeNames) {
```
marcphilipp, line 1364: Shouldn't this get `@API(status = INTERNAL, since = "1.10")`?

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -592,6 +592,68 @@ public static MethodSelector selectMethod(Class<?> javaClass, String methodName,
...
+	 * is declared, or a subclass thereof; never {@code null} or blank
+	 * @param methodName the name of the method to select; never {@code null} or blank
+	 * @param parameterTypes the formal parameter types of the method; never
+	 * {@code null} though potentially empty if the method does not declare parameters
+	 * @since 1.10
+	 * @see MethodSelector
+	 */
+	@API(status = EXPERIMENTAL, since = "1.10")
+	public static MethodSelector selectMethod(ClassLoader classLoader, String className, String methodName,
+			Class<?>... parameterTypes) {
```
marcphilipp, line 628: I'm unsure about the overloads that take a `ClassLoader`, a `String` class name and a `Class<?>...` argument (such as here and below). Since the class is expected to be loaded by the supplied class loader it seems inconsistent. Did you add these overloads to supported mixed mode when indirectly called via the `@SelectMethod` annotation (in a future PR)?

marcphilipp (approved):

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
sbrannen: > Since the class is expected to be loaded by the supplied class loader it seems inconsistent. 

Yes, the `Class` will be loaded by the supplied `ClassLoader`.

But I don't understand what you think is inconsistent.

Can you please expound?

> Did you add these overloads to supported mixed mode when indirectly called via the `@SelectMethod` annotation (in a future PR)?

I added these overloads to align with other changes we have made in 1.10 to support the general "explicit ClassLoader for selectors" theme.

If we don't have these overloads, the user has no control over how the `Class` is loaded for the supplied `className`.

In other words, this aligns with the impetus for introducing `DiscoverySelectors.selectMethod(ClassLoader, String, String, String)` which has also been introduced in 1.10.

Having said that, I do concede that these particular overloads may not be used by many people. I added them for the sake of completeness, since I figured it wouldn't hurt. Although it does increase the number of `selectMethod(...)` overloads. 

If you feel that we should not introduce these overloads, I can remove them, and we can add them later if requested by the community.

sbrannen: Thanks for the clarification.

I'll remove those variants.

If we ever decide we need them, they can be recovered from the commit history in this PR.

sbrannen: Done in 74e78639ccc83d9bcaac600fb5563c9d5054172e

sbrannen:
@marcphilipp, I made the requested changes.

Do you want to take a final look before I merge this PR?

marcphilipp:
@sbrannen Sorry, I was offline for much of the weekend and didn't see this until now.

sbrannen:
No worries.... I just merged it. 🙂

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
kriegfrj: > I meant that the test class will be loaded by the supplied class loader while the parameter types won't. If we don't have a use case for these right now I'd therefore omit them.

Sorry for being late to this discussion... this is a little bit complicated, but for posterity/future reference:

While it *can* be a problem if the parameters are loaded by a different classloader, it not *necessarily* a problem. In fact, in a multiple-classloader environment like OSGi, there are common circumstances where the parameter types *must* be loaded by a different classloader. Consider, for example, if you have a test class in a custom bundle - it will be loaded by the bundle's classloader. However, if it has a test method which takes a Java primitive as a parameter like `java.lang.String`, `java.lang.int` or `java.lang.Integer`, then the parameter will be loaded by the system classloader.

The more important question rather than "is it a different classloader" is "are the classes visible to the test class' classloader". The answer to this question in general depends the specific classloader hierarchy/environment, but it is certainly possible with the `(ClassLoader, String, String, Class<?>...)` variant to end up passing `Class` instances as parameters that are not visible to the test class' classloader and hence won't load. The same is also true of the `(ClassLoader, String, String, String...)` variant though, as you could pass in FQN of parameter types that are not visible to the `ClassLoader` you've provided.

sbrannen: Thanks for chiming in, @kriegfrj.

> The answer to this question in general depends the specific classloader hierarchy/environment

I think that pretty much sums it up: _it depends_. 😉 

In any case, we'll go with what we have in `main` for the time being, and we can revisit this later if necessary.

PR #3336, opened 2023-05-30
Spike API for attaching files to test executions

marcphilipp:


documentation/src/test/java/example/TestReporterDemo.java
```diff
@@ -41,5 +46,18 @@ void reportMultipleKeyValuePairs(TestReporter testReporter) {
...
+		testReporter.publishFile("test1.txt", file -> Files.write(file, singletonList("Test 1")));
+
+		Path existingFile = tempDir.resolve("test2.txt");
+		testReporter.publishFile(Files.write(existingFile, singletonList("Test 2")));
+
+		testReporter.publishFile("test3", dir -> {
+			Path nestedFile = Files.createDirectory(dir).resolve("nested.txt");
+			Files.write(nestedFile, singletonList("Nested content"));
+		});
+	}
```
marcphilipp, line 61: This demonstrates what the programming model for test written with JUnit Jupiter would look like.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestReporter.java
```diff
@@ -77,4 +82,34 @@ default void publishEntry(String value) {
...
+	 * the supplied action.
+	 *
+	 * @param fileName the name of the file to be attached; never {@code null} or blank
+	 *                 and must not contain any path separators
+	 * @param action   the action to be executed to write the file; never {@code null}
+	 * @since 5.11
+	 */
+	@API(status = EXPERIMENTAL, since = "5.11")
+	default void publishFile(String fileName, ThrowingConsumer<Path> action) {
+		throw new UnsupportedOperationException();
```
marcphilipp, line 112: This interface shouldn't be implemented by clients but is annotated with `@FunctionalInterface` so adding another abstract method would, at least theoretically, break backward compatibility. We could decide to ignore that in this case since nobody should have implemented this, go with this pattern, or introduce a separate `FileReporter` interface that can be injected alongside `TestReporter` or could extend it and replace it.

junit-platform-testkit/src/main/java/org/junit/platform/testkit/engine/EngineTestKit.java
```diff
@@ -249,8 +250,8 @@ private static void executeDirectly(TestEngine testEngine, EngineDiscoveryReques
 			EngineExecutionListener listener) {
 		UniqueId engineUniqueId = UniqueId.forEngine(testEngine.getId());
 		TestDescriptor engineTestDescriptor = testEngine.discover(discoveryRequest, engineUniqueId);
-		ExecutionRequest request = new ExecutionRequest(engineTestDescriptor, listener,
-			discoveryRequest.getConfigurationParameters());
+		ExecutionRequest request = ExecutionRequest.create(engineTestDescriptor, listener,
+			discoveryRequest.getConfigurationParameters(), OutputDirProvider.NOOP);
```
marcphilipp, line 254: For testing with `EngineTestKit` a special implementation of `OutputDirProvider` should be provided that uses a temporary directory that is cleaned up after the surrounding test is finished.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestReporter.java
```diff
@@ -77,4 +82,34 @@ default void publishEntry(String value) {
...
+	@API(status = EXPERIMENTAL, since = "5.11")
+	default void publishFile(Path file) {
+		publishFile(file.getFileName().toString(), path -> Files.copy(file, path, REPLACE_EXISTING));
+	}
+
+	/**
+	 * Publish a file with the supplied name written by the supplied action and
+	 * attach it to the current test or container.
+	 * <p>
+	 * The file will be created in the report output directory prior to invoking
```
dsaff, line 102: Is this true?  In the motivating example above, the path is used to create a file or folder in the action lambda.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/JupiterEngineExecutionContext.java
```diff
@@ -37,9 +38,9 @@ public class JupiterEngineExecutionContext implements EngineExecutionContext {
 	private boolean beforeAllCallbacksExecuted = false;
 	private boolean beforeAllMethodsExecuted = false;
 
-	public JupiterEngineExecutionContext(EngineExecutionListener executionListener,
-			JupiterConfiguration configuration) {
-		this(new State(executionListener, configuration));
+	public JupiterEngineExecutionContext(EngineExecutionListener executionListener, JupiterConfiguration configuration,
+			OutputDirProvider outputDirProvider) {
```
dsaff, line 42: You know this design much better than I do, so this is just a question: would it be better to make OutputDirProvider an attribute of JupiterConfiguration rather than passing it alongside?

junit-platform-engine/src/main/java/org/junit/platform/engine/reporting/FileEntry.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.platform.commons.util.ToStringBuilder;
+
+/**
+ * {@code FileEntry} encapsulates a file to be published to the reporting infrastructure.
+ *
+ * @since 1.11
+ * @see #from(Path)
+ */
+@API(status = EXPERIMENTAL, since = "1.11")
+public final class FileEntry {
```
dsaff, line 29: Debating whether we'd want some kind of MIME type, or if we'd trust report users to use extensions or content analysis to know which files were, for example, displayable as images...

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/HierarchicalOutputDirProvider.java
```diff
@@ -0,0 +1,78 @@
...
+			rootDir = rootDirSupplier.get();
+		}
+		return rootDir;
+	}
+
+	private static String sanitizeName(String value) {
+		StringBuilder result = new StringBuilder();
+		for (int i = 0; i < value.length(); i++) {
+			char c = value.charAt(i);
+			if (!FORBIDDEN_CHARS.contains(c) || Character.isISOControl(c)) {
```
dsaff, line 69: Are you sure this doesn't want to be (!FORBIDDEN_CHARS.contains(c) && !Character.isISOControl(c))

dsaff (commented):
This definitely feels like it's a good direction!  My primary design question is whether we'd want a way to attach "types" (MIME types, or something else?) to these files.

In a separate thread of thought, how could we make this available for people still using JUnit 4?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/TestReporter.java
marcphilipp: You're right, this isn't true. Clarified in 6256b39a65ac70a2f6b392c213f8bdce78249e51

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/JupiterEngineExecutionContext.java
marcphilipp: It was passed alongside because `JupiterConfiguration` is created during discovery while `OutputDirProvider` was only passed in later as part of `ExecutionRequest`. However, since `Launcher` has an `execute` method that takes a discovery request, I've moved it there in 170bfa606f63d103e561a94f537e01295a74e668 which did simplify the code considerably.

junit-platform-engine/src/main/java/org/junit/platform/engine/reporting/FileEntry.java
marcphilipp: I can see a case for an optional mime type but making it mandatory is probably too heavy a burden on users. I'd expect one would browse the output dir in a file explorer or web browser at some point which will rely mostly on file extensions anyway. We could add a mime type later or right away. WDYT?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/HierarchicalOutputDirProvider.java
marcphilipp: Good catch! I think I meant to put that in parentheses. Fixed in fc073e3dec335504d443938aed6d39b2d6b76fea

marcphilipp:
> This definitely feels like it's a good direction! My primary design question is whether we'd want a way to attach "types" (MIME types, or something else?) to these files.

see https://github.com/junit-team/junit5/pull/3336#discussion_r1278280333

> In a separate thread of thought, how could we make this available for people still using JUnit 4?

I was wondering that as well. I'm hesitant to change JUnit 4's `RunListener` API to add a method that can publish a file. However, we could provide a `TestRule` implementation that has an API similar to the methods added in this PR to `TestReporter`. The rule would only work when used via the Vintage engine. Otherwise publishing files would be a no-op. The Vintage engine could provide a backing reporter mechanism via a `ThreadLocal`. If you think that might be viable, I can take a stab at spiking it.

marcphilipp:
@leonard84 How would you expose this to Spock tests?

marcphilipp:
> The Vintage engine could provide a backing reporter mechanism via a `ThreadLocal`.

On second thought, maybe a `ServiceLoader`-based approach would be better.

marcphilipp:
@dsaff I pushed a POC for how it could work for JUnit 4

documentation/src/test/java/example/vintage/VintageTestReportingDemo.java
```diff
@@ -0,0 +1,36 @@
...
+
+	@Rule
+	public TestReporting testReporting = new TestReporting();
+
+	@Test
+	public void reportFiles() throws Exception {
+		Path existingFile = Files.write(temporaryFolder.getRoot().toPath().resolve("test.txt"), singletonList("Test"));
+		testReporting.publishFile(existingFile);
+	}
+}
```
marcphilipp, line 36: JUnit 4 example

junit-platform-engine/src/main/java/org/junit/platform/engine/reporting/FileEntry.java
dsaff: Leaving it for later seems fine.

documentation/src/test/java/example/vintage/VintageTestReportingDemo.java
dsaff: Looks really clean!

junit-platform-engine/src/main/java/org/junit/platform/engine/reporting/FileEntry.java
```diff
@@ -0,0 +1,67 @@
...
+	 * Factory for creating a new {@code FileEntry} from the supplied file.
+	 *
+	 * @param file the file to publish; never {@code null}
+	 */
+	public static FileEntry from(Path file) {
+		return new FileEntry(file);
+	}
+
+	private final LocalDateTime timestamp = LocalDateTime.now();
+	private final Path file;
```
leonard84, line 41: Adding a `Map<String, String>` for metadata/annotations would be useful.
For example, you could store the URL for the screenshot's current page and other values.

PR #3348, opened 2023-06-10
Document replacement for `@Test(expected = ...)` in JUnit 4 migration tips

pbodnar:
## Overview

I believe that the `expected` property is widely used in JUnit 4, so it would deserve a note in the migration guide.

---

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
@@ -63,6 +63,10 @@ tests to JUnit Jupiter.
 * `@Rule` and `@ClassRule` no longer exist; superseded by `@ExtendWith` and
   `@RegisterExtension`.
   - See also <<migrating-from-junit4-rule-support>>.
+* `@Test(expected = ...)` and the `ExpectedException` rule no longer exist; use
+  `Assertions.assertThrows(...)` instead.
+  See <<migrating-from-junit4-rule-support>> if you still do need to use
+  `ExpectedException`.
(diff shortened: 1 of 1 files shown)


sbrannen:
The same can be said about the `ExpectedException` rule.

pbodnar:
OK, I'm only not sure how to incorporate it with regards to the point about rules above and the chapter linked from there.

Maybe st like the following?

```diff
* `@Rule` and `@ClassRule` no longer exist; superseded by `@ExtendWith` and
  `@RegisterExtension`.
  - See also <<migrating-from-junit4-rule-support>>.
-* `@Test.expected` no longer exists; use `assertThrows` method instead.
+* `@Test.expected` and `ExpectedException` no longer exist; use `assertThrows` method instead.
+  See <<migrating-from-junit4-rule-support>> if you still do need to use `ExpectedException`.
```

sbrannen:
> OK, I'm only not sure how to incorporate it with regards to the point about rules above and the chapter linked from there.
> 
> Maybe st like the following?

Seems good to me.

How about this?

> `@Test(expected = ...)` and the `ExpectedException` rule no longer exist; use `Assertions.assertThrows(...)` instead.

pbodnar:
OK, so here we go? :)

pbodnar:
@sbrannen, @marcphilipp, can you see some gotcha, or ...? :)

sbrannen:
> @sbrannen, @marcphilipp, can you see some gotcha, or ...? :)

I think that looks good now. 

I'll get this merged into `main` before 5.10 RC1 gets released.

sbrannen:

Thanks

PR #3364, opened 2023-06-19
Revise stacktrace pruning

### LINKED ISSUE #3299: Revise stack trace pruning
## Context

Stack trace pruning was introduced in #3247.
We want this feature to be disabled for tests inside this repository (otherwise all stack traces would be almost empty), and it is when the tests are launched via gradle (due to [this setting](https://github.com/junit-team/junit5/blob/69e74953eacb174842a65a2d7bf67c4d633de6e9/gradle/plugins/common/src/main/kotlin/junitbuild.testing-conventions.gradle.kts#L44)). But when launched in a different way, this setting have no impact and the feature is therefore enabled when it shouldn’t.

## Solution

A potential solution could be to not prune stack traces of tests specifically located inside the `org.junit` package, even when the feature is enabled.

## Related

- https://github.com/junit-team/junit5/pull/3277#discussion_r1181233970

## Deliverables

- [ ] Collect `MethodSource` and `ClassSource` instances from the current test descriptor and its parents, find the bottom-most stackframe with one of these classes, and include all stackframe upward from there
- [ ] Don't prune everything from `java.*` and `jdk.*`, just the reflection packages from both but only below the test method
- [ ] Remove package pattern configuration parameter introduced in M1

juliette-derancourt:
## Overview

Resolves #3299

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ExceptionUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ExceptionUtils.java
@@ -42,7 +42,8 @@ public final class ExceptionUtils {
 
 	private static final String JUNIT_PLATFORM_LAUNCHER_PACKAGE_PREFIX = "org.junit.platform.launcher.";
 
-	private static final String STACK_TRACE_ELEMENTS_TO_EXCLUDE = "org.junit.*,jdk.internal.reflect.*,sun.reflect.*";
+	private static final Predicate<String> STACK_TRACE_ELEMENT_FILTER = ClassNamePatternFilterUtils //
+			.excludeMatchingClassNames("org.junit.*,jdk.internal.reflect.*,sun.reflect.*");
 
 	private ExceptionUtils() {
 		/* no-op */
@@ -114,9 +115,6 @@ public final class ExceptionUtils {
 		Preconditions.notNull(throwable, "Throwable must not be null");
 		Preconditions.notNull(testClassNames, "List of test class names must not be null");
 
-		Predicate<String> stackTraceElementFilter = ClassNamePatternFilterUtils //
-				.excludeMatchingClassNames(STACK_TRACE_ELEMENTS_TO_EXCLUDE);
-
 		List<StackTraceElement> stackTrace = Arrays.asList(throwable.getStackTrace());
 		List<StackTraceElement> prunedStackTrace = new ArrayList<>();
 
@@ -134,7 +132,7 @@ public final class ExceptionUtils {
 			else if (className.startsWith(JUNIT_PLATFORM_LAUNCHER_PACKAGE_PREFIX)) {
 				prunedStackTrace.clear();
 			}
-			else if (stackTraceElementFilter.test(className)) {
+			else if (STACK_TRACE_ELEMENT_FILTER.test(className)) {
 				prunedStackTrace.add(element);
 			}
 		}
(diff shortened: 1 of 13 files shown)


platform-tests/src/test/java/org/junit/platform/commons/util/ExceptionUtilsTests.java
```diff
@@ -64,15 +67,16 @@ void readStackTraceForLocalJUnitException() {
 		}
 	}
 
-	@Test
-	void pruneStackTraceOfCallsFromSpecificPackage() {
+	@ParameterizedTest
+	@ValueSource(strings = { "org.junit.", "jdk.internal.reflect.", "sun.reflect." })
+	void pruneStackTraceOfCallsFromSpecificPackage(String shouldBePruned) {
```
sbrannen, line 72: Is there a concrete need to throw and catch these exceptions?

If not, just rely on the fact that `fillInStackTrace()` is called by the exceptions constructor (technically in `java.lang.Throwable.Throwable(String)`).

platform-tests/src/test/java/org/junit/platform/commons/util/ExceptionUtilsTests.java
```diff
@@ -64,15 +67,16 @@ void readStackTraceForLocalJUnitException() {
...
 		try {
 			throw new JUnitException("expected");
 		}
 		catch (JUnitException e) {
-			pruneStackTrace(e, element -> !element.startsWith("org.junit."));
+			pruneStackTrace(e, Collections.emptyList());
```
sbrannen, line 77: In tests, feel free to use `List.of()` instead of `Collections.emptyList()`.

sbrannen (changes requested):
I've requested a few minor changes and will hopefully find time to perform a more thorough review after the PR has been updated.

sbrannen (approved):

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ExceptionUtils.java
```diff
@@ -92,35 +94,44 @@ public static String readStackTrace(Throwable throwable) {
...
 	 */
 	@API(status = INTERNAL, since = "5.10")
-	public static void pruneStackTrace(Throwable throwable, Predicate<String> stackTraceElementFilter) {
+	public static void pruneStackTrace(Throwable throwable, List<String> testClassNames) {
 		Preconditions.notNull(throwable, "Throwable must not be null");
-		Preconditions.notNull(stackTraceElementFilter, "Predicate must not be null");
+		Preconditions.notNull(testClassNames, "List of test class names must not be null");
+
+		Predicate<String> stackTraceElementFilter = ClassNamePatternFilterUtils //
+				.excludeMatchingClassNames(STACK_TRACE_ELEMENTS_TO_EXCLUDE);
```
marcphilipp, line 118: I think we should extract this into a constant.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/StackTracePruningEngineExecutionListener.java
```diff
@@ -10,49 +10,60 @@
...
+			ExceptionUtils.findNestedThrowables(throwable).forEach(
+				t -> ExceptionUtils.pruneStackTrace(t, testClassNames));
 		}
 		super.executionFinished(testDescriptor, testExecutionResult);
 	}
 
-	private void pruneStackTrace(Throwable throwable) {
-		ExceptionUtils.pruneStackTrace(throwable, stackTraceElementFilter);
+	private static List<String> getTestClassNames(TestDescriptor testDescriptor) {
+		return Stream.of(testDescriptor.getSource(), testDescriptor.getParent().flatMap(TestDescriptor::getSource)) //
```
marcphilipp, line 51: I think we should do this up to the engine descriptor. For example, to handle an exception thrown from a top-level `@BeforeEach` method with multiple levels of `@Nested` classes.

marcphilipp (changes requested):

marcphilipp (approved):

sormuras (approved):

PR #2613, opened 2021-05-11
Fix package path computation in `ClasspathScanner`

sormuras:
## Overview

Fix `getRootUrisForPackage()` in class `ClasspathScanner` by looking for two "wordings" of a package name.

For example, package `foo.bar` is expanded to `foo/bar` and `foo/bar/`. The latter allows find package `foo.bar` in a module called `foo.bar`, and not `foo`.

Fixes #2500
Fixes #2600
Fixes #2612

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement).

---

### Definition of Done

- [x] There are no TODOs left in the code
- [x] Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.commons/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [x] [Coding conventions](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [x] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling
- [x] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [x] Change is documented in the [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)
- [x] All [continuous integration builds](https://github.com/junit-team/junit5#continuous-integration-builds) pass

PR #3415, opened 2023-08-06
Fix console-related test failures on JDK 22

rybak:
## Overview

Details are in the commit message.

I'm not sure if this is a good approach, or how JUnit project likes to handle compatibility issues like this. I tried to follow the example of surrounding code in `junitbuild.testing-conventions.gradle.kts` as best as I could.

---

### DIFF
--- a/gradle/plugins/common/src/main/kotlin/junitbuild.testing-conventions.gradle.kts
+++ b/gradle/plugins/common/src/main/kotlin/junitbuild.testing-conventions.gradle.kts
@@ -51,6 +51,9 @@ tasks.withType<Test>().configureEach {
 			"-XX:FlightRecorderOptions=stackdepth=1024"
 		)
 	}
+	if (buildParameters.javaToolchainVersion.get() >= 22) {
+		jvmArgs("-Djdk.console=java.base")
+	}
 
 	// Track OS as input so that tests are executed on all configured operating systems on CI
 	trackOperationSystemAsInput()


sormuras:
Thanks for working on this, Andrei. I created #3419 to track the underlying issue.

My hope is that we'll find a solution that does not need the `-Djdk.console=java.base` to restore older behaviour.
Main question is: where is `stty: /dev/tty: No such device or address` coming from? Is it only a problem on CI systems, like GitHub Actions?

rybak:
@sormuras thank you for review and creating the issue. I don't remember for sure, but I think I saw the tty problem locally too, so it might be related to JLine.

Please also see related https://github.com/junit-team/junit5/pull/3416 which is ready for review (unlike this PR). It is fixing an issue introduced in #2613, so you are probably the best reviewer for #3416.

sormuras:
> I don't remember for sure, but I think I saw the tty problem locally too, so it might be related to JLine.

I was not able to reproduce directly, meaning on a standard shell - I can reproduce it on GitHub Actions though. See https://github.com/sormuras/JDK-8313893 for details. The issue originates from Picocli: https://github.com/remkop/picocli/issues/1104

sormuras:
> Please also see related https://github.com/junit-team/junit5/pull/3416 which is ready for review (unlike this PR). It is fixing an issue introduced in https://github.com/junit-team/junit5/pull/2613, so you are probably the best reviewer for https://github.com/junit-team/junit5/pull/3416.

I plan to take a look next week. Thanks for also working on this!

sormuras:
Superseded by #3423

PR #3416, opened 2023-08-06
Fix StandaloneTests specific for Java 8

rybak:
## Overview

While working on JDK 22 compatibility in #3415, I've discovered a problem with tests specific for Java 8 – details in the last commit.

I've also reproduced this issue on commit c62cd6ab11, where the affected tests were introduced. The slight build script change needed to run Gradle on this commit are in [a separate branch in my fork](https://github.com/junit-team/junit5/compare/main...rybak:junit5:java8-issue-origin), along with some debug output.

---


rybak:
From 2357c82:

> Tests that claim by their name to run on Java 8 don't actually run on Java 8. 

Re-ran the tests on `main` to double-check:

Screenshot_20230806 java 8 tests

rybak:
I have a suspicion that this might be caused by a discrepancy in bytecode version in class files. In JDK 17, `javac` produces bytecode of version 61.0, while in JDK 8, `javac` produces bytecode of version 52.0.

In class `StandaloneTests`, test `compile()` launches `javac` with the current JDK, thus creating class files, which cannot be picked up – see new commit 795e01d043420c6dbcbfcedd5f3c84eb80d666d0.

With commit 795e01d043420c6dbcbfcedd5f3c84eb80d666d0, tests executeOnJava8() and executeOnJava8SelectPackage() now fail for more obvious reasons: output of `-showversion` goes to stderr instead of stdout, as expected by `expected-out.txt`.

rybak:
Force pushed to squash fixup commits and to reword commit messages.

rybak:
I hope my explanation of the issue isn't too long and makes sense :-)

marcphilipp (approved):
Thanks for spotting and fixing this! 👍

rybak:
Removed two commits from the top:

```
$ git range-diff e53f93482a...060f2c4bec
1:  0b1a0029b3 < -:  ---------- Remove -Xlint:-options
2:  e53f93482a < -:  ---------- Revert "Remove -Xlint:-options"
```

PR #3500, opened 2023-10-09
Apply method predicate before searching type hierarchy

### LINKED ISSUE #3498: `@BeforeAll` method in super class skipped when it has same name as a `@BeforeEach` method in subclass
## Steps to reproduce

Given:

```java
package a;

import org.junit.jupiter.api.BeforeAll;

public abstract class A {

    @BeforeAll
    static void before() {
        System.out.println("A");
    }
}
```

and

```java
package b;

import a.A;
import org.junit.jupiter.api.BeforeEach;

public class B extends A {

    @BeforeEach
    void before() {
        System.out.println("B");
    }

    @Test
    void test() {
        System.out.println("Hi there");
    }
}
```

Executing the test in `B` prints: `B` and `Hi there`.  Note the absence of  `A`: the `@BeforeAll` method was not called.  Note that the non-static `before` method in `B` does not override the static method (it is in a different package), nor does it conflict with it.  Renaming the `before` method in `B` or `A` fixes the problem.

This can easily happen when you have an `abstract` base class in some utility package, and you are extending it to build tests.  If you happen to name a method in this test class that conflicts with a package-private method in the base class, things do not work as expected.

## Workaround:

Ensure the names in the base class are sufficiently unique that they're unlikely to have similar names in a subclass.

## Deliverables

- A fix for this problem.  It probably is a problem with annotation scanning not being thorough enough and not taking this special case into account.

sbrannen:
- fixes #3498

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.10.1.adoc
@@ -15,7 +15,10 @@ JUnit repository on GitHub.
 
 ==== Bug Fixes
 
-* ❓
+* Method predicates are now applied while searching the type hierarchy. This fixes bugs
+  in `findMethods(...)` and `streamMethods(...)` in `ReflectionSupport` as well as
+  `findAnnotatedMethods(...)` in `AnnotationSupport`.
+  - See link:https://github.com/junit-team/junit5/issues/3498[issue 3498] for details.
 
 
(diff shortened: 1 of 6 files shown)


sbrannen:
Dear @pixeebot, please **do not spam** issue trackers about things you may or may not do in the future.

hjohn (commented):
LGTM

marcphilipp:
@nahsra Could you please let us know when the new version is live? I've disabled Pixeebot in the meantime.

PR #3507, opened 2023-10-17
Introduce `StringConversionSupport` in `junit-platform-commons`

### LINKED ISSUE #3449: Introduce `StringConversionSupport` in `junit-platform-commons`
[`DefaultArgumentConverter`](https://junit.org/junit5/docs/5.9.1/api/org.junit.jupiter.params/org/junit/jupiter/params/converter/DefaultArgumentConverter.html) used by JUnit is marked as internal. I'm not sure why that is but I think it'd be great if we got some support for argument conversion, so I'm proposing partially or entirely exposing or mirroring it in the `support` package (or somewhere else where it makes sense).

Or if the JUnit team doesn't want to support it, maybe Pioneer can host something?

## Deliverables

- [x] All or partial functionality of `DefaultArgumentConverter` exposed.

juliette-derancourt:
## Overview

- Resolves #3449

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
@@ -26,6 +26,7 @@ import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.platform.commons.support.ConversionException;
 import org.junit.platform.commons.support.StringConversionSupport;
 import org.junit.platform.commons.util.ClassLoaderUtils;
+import org.junit.platform.commons.util.ReflectionUtils;
 
 /**
  * {@code DefaultArgumentConverter} is the default implementation of the
@@ -61,15 +62,32 @@ public class DefaultArgumentConverter implements ArgumentConverter {
 	}
 
 	public final Object convert(Object source, Class<?> targetType, ParameterContext context) {
-		Class<?> declaringClass = context.getDeclaringExecutable().getDeclaringClass();
-		ClassLoader classLoader = ClassLoaderUtils.getClassLoader(declaringClass);
+		if (source == null) {
+			if (targetType.isPrimitive()) {
+				throw new ArgumentConversionException(
+					"Cannot convert null to primitive value of type " + targetType.getTypeName());
+			}
+			return null;
+		}
 
-		try {
-			return StringConversionSupport.convert(source, targetType, classLoader);
+		if (ReflectionUtils.isAssignableTo(source, targetType)) {
+			return source;
 		}
-		catch (ConversionException ex) {
-			throw new ArgumentConversionException(ex.getMessage(), ex);
+
+		if (source instanceof String) {
+			Class<?> declaringClass = context.getDeclaringExecutable().getDeclaringClass();
+			ClassLoader classLoader = ClassLoaderUtils.getClassLoader(declaringClass);
+			try {
+				return StringConversionSupport.convert((String) source, targetType, classLoader);
+			}
+			catch (ConversionException ex) {
+				throw new ArgumentConversionException(ex.getMessage(), ex);
+			}
 		}
+
+		throw new ArgumentConversionException(
+			String.format("No built-in converter for source type %s and target type %s",
+				source.getClass().getTypeName(), targetType.getTypeName()));
 	}
 
 }
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
@@ -23,8 +23,8 @@ import java.util.UUID;
 
 import org.apiguardian.api.API;
 import org.junit.jupiter.api.extension.ParameterContext;
-import org.junit.platform.commons.support.ConversionException;
-import org.junit.platform.commons.support.StringConversionSupport;
+import org.junit.platform.commons.support.conversion.ConversionException;
+import org.junit.platform.commons.support.conversion.StringConversionSupport;
 import org.junit.platform.commons.util.ClassLoaderUtils;
 import org.junit.platform.commons.util.ReflectionUtils;
 
@@ -44,7 +44,7 @@ import org.junit.platform.commons.util.ReflectionUtils;
  *
  * @since 5.0
  * @see org.junit.jupiter.params.converter.ArgumentConverter
- * @see org.junit.platform.commons.support.StringConversionSupport
+ * @see org.junit.platform.commons.support.conversion.StringConversionSupport
  */
 @API(status = INTERNAL, since = "5.0")
 public class DefaultArgumentConverter implements ArgumentConverter {
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
@@ -10,10 +10,7 @@
 
 package org.junit.jupiter.params.converter;
 
-import static java.util.Arrays.asList;
-import static java.util.Collections.unmodifiableList;
 import static org.apiguardian.api.API.Status.INTERNAL;
-import static org.junit.platform.commons.util.ReflectionUtils.getWrapperType;
 
 import java.io.File;
 import java.math.BigDecimal;
@@ -21,15 +18,14 @@ import java.math.BigInteger;
 import java.net.URI;
 import java.net.URL;
 import java.util.Currency;
-import java.util.List;
 import java.util.Locale;
-import java.util.Optional;
 import java.util.UUID;
 
 import org.apiguardian.api.API;
 import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.platform.commons.support.ConversionException;
+import org.junit.platform.commons.support.StringConversionSupport;
 import org.junit.platform.commons.util.ClassLoaderUtils;
-import org.junit.platform.commons.util.ReflectionUtils;
 
 /**
  * {@code DefaultArgumentConverter} is the default implementation of the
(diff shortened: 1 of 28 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Locale;
+import java.util.Optional;
+import java.util.UUID;
+
+import org.apiguardian.api.API;
+import org.junit.platform.commons.util.ClassLoaderUtils;
+import org.junit.platform.commons.util.ReflectionUtils;
+
+/**
+ * The {@code DefaultArgumentConverter} is able to convert from strings to a
```
sbrannen, line 34:
```suggestion
 * {@code StringConversionSupport} is able to convert from strings to a
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringToObjectConverter.java
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright 2015-2023 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.support;
```
sbrannen, line 11: I'm wondering if it wouldn't be better to introduce a `conversion` subpackage here.

@marcphilipp, WDYT?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
```diff
@@ -75,47 +61,15 @@ public final Object convert(Object source, ParameterContext context) {
...
-						// simply rethrow it
-						throw (ArgumentConversionException) ex;
-					}
-					// else
-					throw new ArgumentConversionException(
-						"Failed to convert String \"" + source + "\" to type " + targetType.getTypeName(), ex);
-				}
-			}
+		catch (ConversionException e) {
+			throw new ArgumentConversionException(e.getMessage(), e);
```
sbrannen, line 71:
```suggestion
		catch (ConversionException ex) {
			throw new ArgumentConversionException(ex.getMessage(), ex);
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,106 @@
...
+ * number of primitive types and their corresponding wrapper types (Byte, Short,
+ * Integer, Long, Float, and Double), date and time types from the
+ * {@code java.time} package, and some additional common Java types such as
+ * {@link File}, {@link BigDecimal}, {@link BigInteger}, {@link Currency},
+ * {@link Locale}, {@link URI}, {@link URL}, {@link UUID}, etc.
+ *
+ * <p>If the source and target types are identical the source object will not
+ * be modified.
```
sbrannen, line 42:
```suggestion
 * <p>If the target type is {@code String} the source {@code String} will not
 * be modified.
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,106 @@
...
+		new StringToCommonJavaTypesConverter(), //
+		new FallbackStringToObjectConverter() //
+	));
+
+	private StringConversionSupport() {
+		/* no-op */
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <T> T convert(Object source, Class<T> targetType, ClassLoader classLoader) {
```
sbrannen, line 63:
```suggestion
	public static <T> T convert(String source, Class<T> targetType, ClassLoader classLoader) {
```

For this new public API, the `source` should always be a `String`.

So we need to change the parameter type and update the rest of the code accordingly.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,106 @@
...
+			if (targetType.isPrimitive()) {
+				throw new ConversionException(
+					"Cannot convert null to primitive value of type " + targetType.getTypeName());
+			}
+			return null;
+		}
+
+		if (ReflectionUtils.isAssignableTo(source, targetType)) {
+			return (T) source;
+		}
```
sbrannen, line 74: I _think_ we might want to move this logic back to `DefaultArgumentConverter` and change this logic here to something specific to `String`.

But I'd need to analyze the code in an IDE to make a concrete decision on that.

Perhaps you can investigate our options in your IDE.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,106 @@
...
+
+		if (source instanceof String) {
```
sbrannen, line 76: After we change the parameter type for `source` to `String`, this if-block should no longer be necessary.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,106 @@
...
+						throw (ConversionException) ex;
+					}
+					// else
+					throw new ConversionException(
+						"Failed to convert String \"" + source + "\" to type " + targetType.getTypeName(), ex);
+				}
+			}
+		}
+		throw new ConversionException(String.format("No built-in converter for source type %s and target type %s",
+			source.getClass().getTypeName(), targetType.getTypeName()));
```
sbrannen, line 98: Similarly, this can also be simplified since the `source` will always be a `String`.

sbrannen (changes requested):
Thanks for working on this one, @juliette-derancourt!

Assuming you copied all of the `StringToObjectConverter` infrastructure verbatim ("assuming" because I did not verify it 😇), I think this looks good.

And I've requested a few minor changes to focus this API on `String` for the `source` type.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringToObjectConverter.java
juliette-derancourt: I think it would look tidier with a dedicated subpackage 👍

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
juliette-derancourt: I moved it back to `DefaultArgumentConverter`, and only checked equality between `targetType` and `String.class`. Does this seem ok to you?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,104 @@
...
+			return (T) source;
+		}
+
+		Class<?> targetTypeToUse = toWrapperType(targetType);
+		Optional<StringToObjectConverter> converter = stringToObjectConverters.stream().filter(
+			candidate -> candidate.canConvert(targetTypeToUse)).findFirst();
+		if (converter.isPresent()) {
+			try {
+				ClassLoader classLoaderToUse = Optional.ofNullable(classLoader) //
+						.orElseGet(ClassLoaderUtils::getDefaultClassLoader);
```
sbrannen, line 81:
```suggestion
				ClassLoader classLoaderToUse = classLoader != null ? classLoader //
						: ClassLoaderUtils.getDefaultClassLoader();
```

Let's go with a ternary expression here.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,104 @@
...
+		new StringToCommonJavaTypesConverter(), //
+		new FallbackStringToObjectConverter() //
+	));
+
+	private StringConversionSupport() {
+		/* no-op */
+	}
+
+	@SuppressWarnings("unchecked")
+	public static <T> T convert(String source, Class<T> targetType, ClassLoader classLoader) {
```
sbrannen, line 62: Please add complete method-level Javadoc, since this is the core of the new API.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
```diff
@@ -0,0 +1,104 @@
...
+ * {@code StringConversionSupport} is able to convert from strings to a number
+ * of primitive types and their corresponding wrapper types (Byte, Short,
+ * Integer, Long, Float, and Double), date and time types from the
+ * {@code java.time} package, and some additional common Java types such as
+ * {@link File}, {@link BigDecimal}, {@link BigInteger}, {@link Currency},
+ * {@link Locale}, {@link URI}, {@link URL}, {@link UUID}, etc.
+ *
+ * <p>If the target type is {@code String} the source {@code String} will not
+ * be modified.
+ */
```
sbrannen, line 42: Please add `@since` tag.

sbrannen (changes requested):

It's looking even better!

I've now requested only some minor polishing.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
sbrannen: I think it's fine like that. 

Merci

junit-platform-commons/src/main/java/org/junit/platform/commons/support/StringConversionSupport.java
juliette-derancourt: I tried my best, but I might need your advice on this 😁

sbrannen (approved):
Thanks for making all of the requested changes.

Looks good!

Issue #1877, opened 2019-05-13
Missing opens directives in module descriptors


sormuras:
### Description

Having declared

```java
@Test
@EnabledOnJre(JRE.JAVA_11) void test() {}
```

or other usages of Jupiter API that internally needs to access _package private_ types yields:

```
Unable to make org.junit.jupiter.api.condition.EnabledOnJreCondition() accessible:
module org.junit.jupiter.api does not "opens org.junit.jupiter.api.condition"
                                   to module org.junit.platform.commons
```

Module `org.junit.jupiter.api` is declared like follows:
```java
module org.junit.jupiter.api {
	requires transitive org.apiguardian.api;
	requires transitive org.junit.platform.commons;
	requires transitive org.opentest4j;

	exports org.junit.jupiter.api;
	exports org.junit.jupiter.api.condition;
	exports org.junit.jupiter.api.extension;
	exports org.junit.jupiter.api.function;
	exports org.junit.jupiter.api.io;
	exports org.junit.jupiter.api.parallel;
}
```

Exporting _public_ types from all 6 packages doesn't allow deep reflective access to non-public types. Thus, directives like `opens org.junit.jupiter.api.condition to org.junit.platform.commons;` need to added to this module descriptor. This goes also for all other modules that make use of the reflection support provided by module `org.junit.platform.commons`.

#### Reflecting code uses
* Usages of `@ExtendWith` in JUnit "Project Production Java Files"
![image](https://user-images.githubusercontent.com/2319838/57631781-76ceeb00-75a0-11e9-90ad-0ac736af099f.png)

### Deliverables
- [ ] Write integration test that shows the issue.
- [ ] Fix by adding appropriate `opens to org.junit.platform.commons` directives.
- [ ] Find other places where reflection support from `org.junit.platform.commons` is used.

sormuras:
Updated initial issue description.

sormuras:
### JUnit 5 - `documentation` as modular test
```
java
--show-version
--show-module-resolution
--module-path {...}
--add-modules documentation
--module org.junit.platform.console
--scan-modules
--include-classname .*Tests
--include-classname .*Demo
--exclude-tag exclude
```
- test run module resolution https://gist.github.com/sormuras/9cb94d59805b1435f6e80de2776cf128
- test run tree https://gist.github.com/sormuras/9ce9155988cf2166b0ac28299aa18d48

```
28 containers failed
20 tests failed
```

Mamadou59:
[...]

PR #3592, opened 2023-12-05
Introduce `@AutoClose`

### LINKED ISSUE #3367: Introduce `@AutoClose` mechanism for closing field resources after test execution
Occasionally I create and use resources in unit tests that should be released after test execution. You usually do this in tear-down methods. But in a lot of cases a simple `@AutoClose` annotated field would be sufficient for that, something similar to `@TempDir` for temporary folders.

This would spare boilerplate code like:

```java
Service service = new ClosableService();

@AfterEach
void cleanup() {
    service.close();
}
```

Allowing it to be replaced by:

```java
@AutoClose
Service service = new ClosableService();
```

Static fields would be closed after all tests have run and instance fields after each test.

Other annotation names could be `@Destroy` or `@AutoDestroy`.

If there is no `close()` method, it should be possible to specify a custom method name -- for example, `destroy`, `dispose`, `shutdown`, etc.

## Deliverables

- A new `@AutoClose` annotation
- Update User Guide Writing Tests -> Annotations

bjmi:
## Overview

Closes #3367

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/link-attributes.adoc
+++ b/documentation/src/docs/asciidoc/link-attributes.adoc
@@ -153,7 +153,7 @@ endif::[]
 // Jupiter Engine
 :junit-jupiter-engine:                       {javadoc-root}/org.junit.jupiter.engine/org/junit/jupiter/engine/package-summary.html[junit-jupiter-engine]
 // Jupiter Extension Implementations
-:AutoCloseExtension:                         {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java[AutoCloseExtension]
+:AutoCloseExtension:                         {current-branch}/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java[AutoCloseExtension]
 :DisabledCondition:                          {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/DisabledCondition.java[DisabledCondition]
 :RepetitionExtension:                        {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/RepetitionExtension.java[RepetitionExtension]
 :TempDirectory:                              {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java[TempDirectory]
--- a/documentation/src/docs/asciidoc/link-attributes.adoc
+++ b/documentation/src/docs/asciidoc/link-attributes.adoc
@@ -153,6 +153,7 @@ endif::[]
 // Jupiter Engine
 :junit-jupiter-engine:                       {javadoc-root}/org.junit.jupiter.engine/org/junit/jupiter/engine/package-summary.html[junit-jupiter-engine]
 // Jupiter Extension Implementations
+:AutoCloseExtension:                         {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java[AutoCloseExtension]
 :DisabledCondition:                          {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/DisabledCondition.java[DisabledCondition]
 :RepetitionExtension:                        {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/RepetitionExtension.java[RepetitionExtension]
 :TempDirectory:                              {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java[TempDirectory]
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -2710,3 +2710,26 @@ following precedence rules:
 2. The default `TempDirFactory` configured via the configuration
 parameter, if present
 3. Otherwise, `org.junit.jupiter.api.io.TempDirFactory$Standard` will be used.
+
+[[writing-tests-built-in-extensions-AutoClose]]
+==== The AutoClose Extension
+
+The built-in `{AutoCloseExtension}` is used to automatically close resources used in
+tests. Therefore, the `@AutoClose` annotation is applied to fields within the
+test class to indicate that the annotated resource should be automatically closed after
+the test execution.
+
+By default, the `@AutoClose` annotation expects the annotated resource to provide
+a `close()` method that will be invoked for closing the resource. However, developers
+can customize the closing behavior by providing a different method name through the
+`value` attribute. For example, setting `value = "shutdown"` will look
+for a method named `shutdown()` to close the resource.
+
+For example, the following test declares a database connection field annotated with
+`@AutoClose` that is automatically closed afterward.
+
+[source,java,indent=0]
+.A test class using @AutoClose annotation to close used resource
+----
+include::{testDir}/example/AutoCloseDemo.java[tags=user_guide_example]
+----
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -45,6 +45,7 @@ in the `junit-jupiter-api` module.
 | `@ExtendWith`            | Used to <<extensions-registration-declarative,register extensions declaratively>>. Such annotations are _inherited_.
 | `@RegisterExtension`     | Used to <<extensions-registration-programmatic,register extensions programmatically>> via fields. Such fields are _inherited_ unless they are _shadowed_.
 | `@TempDir`               | Used to supply a <<writing-tests-built-in-extensions-TempDirectory,temporary directory>> via field injection or parameter injection in a lifecycle method or test method; located in the `org.junit.jupiter.api.io` package.
+| @AutoClose               | Indicates that a field in a JUnit 5 test class represents a resource that should be automatically closed after test execution.
 |===
 
 WARNING: Some annotations may currently be _experimental_. Consult the table in
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -45,7 +45,7 @@ in the `junit-jupiter-api` module.
 | `@ExtendWith`            | Used to <<extensions-registration-declarative,register extensions declaratively>>. Such annotations are _inherited_.
 | `@RegisterExtension`     | Used to <<extensions-registration-programmatic,register extensions programmatically>> via fields. Such fields are _inherited_ unless they are _shadowed_.
 | `@TempDir`               | Used to supply a <<writing-tests-built-in-extensions-TempDirectory,temporary directory>> via field injection or parameter injection in a lifecycle method or test method; located in the `org.junit.jupiter.api.io` package.
-| @AutoClose               | Indicates that a field in a JUnit 5 test class represents a resource that should be automatically closed after test execution.
+| `@AutoClose`             | Denotes that the annotated field represents a resource that should be automatically closed after test execution.
 |===
 
 WARNING: Some annotations may currently be _experimental_. Consult the table in
--- a/documentation/src/test/java/example/AutoCloseDemo.java
+++ b/documentation/src/test/java/example/AutoCloseDemo.java
@@ -18,13 +18,15 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 
 import org.junit.jupiter.api.AutoClose;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
+@Disabled
 // tag::user_guide_example[]
 class AutoCloseDemo {
 
 	@AutoClose
-	Connection connection = DriverManager.getConnection("jdbc:mysql://localhost/testdb");
+	Connection connection = getJdbcConnection("jdbc:mysql://localhost/testdb");
 
 	@Test
 	void usersTableHasEntries() throws SQLException {
@@ -33,8 +35,15 @@ class AutoCloseDemo {
 		assertTrue(resultSet.next());
 	}
 
-	AutoCloseDemo() throws SQLException {
+	// ...
+	// end::user_guide_example[]
+	private static Connection getJdbcConnection(String url) {
+		try {
+			return DriverManager.getConnection(url);
+		}
+		catch (SQLException ex) {
+			throw new RuntimeException(ex);
+		}
 	}
 
 }
-// end::user_guide_example[]
--- a/documentation/src/test/java/example/AutoCloseDemo.java
+++ b/documentation/src/test/java/example/AutoCloseDemo.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2015-2023 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package example;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+import org.junit.jupiter.api.AutoClose;
+import org.junit.jupiter.api.Test;
+
+// tag::user_guide_example[]
+class AutoCloseDemo {
+
+	@AutoClose
+	Connection connection = DriverManager.getConnection("jdbc:mysql://localhost/testdb");
+
+	@Test
+	void usersTableHasEntries() throws SQLException {
+		ResultSet resultSet = connection.createStatement().executeQuery("SELECT * FROM users");
+
+		assertTrue(resultSet.next());
+	}
+
+	AutoCloseDemo() throws SQLException {
+	}
+
+}
+// end::user_guide_example[]
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
@@ -29,8 +29,9 @@ import org.apiguardian.api.API;
  * <p>
  * By default, the {@code AutoClose} annotation expects the annotated resource to provide a {@code close()} method
  * that will be invoked for closing the resource. However, developers can customize the closing behavior by providing
- * a different method name through the {@code value} attribute. For example, setting {@code value = "destroy"} will
- * look for a method named {@code destroy()} to close the resource.
+ * a different method name through the {@code value} attribute. For example, setting {@code value = "shutdown"} will
+ * look for a method named {@code shutdown()} to close the resource.
+ * When multiple annotated resources exist the order of closing them is unspecified.
  * </p>
  *
  * <p>
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2015-2023 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+
+/**
+ * The {@code AutoClose} annotation is used to automatically close resources used in JUnit 5 tests.
+ *
+ * <p>
+ * This annotation should be applied to fields within JUnit 5 test classes. It indicates that the annotated
+ * resource should be automatically closed after the test execution. The annotation targets
+ * {@link java.lang.annotation.ElementType#FIELD} elements, allowing it to be applied to instance variables.
+ * </p>
+ *
+ * <p>
+ * By default, the {@code AutoClose} annotation expects the annotated resource to provide a {@code close()} method
+ * that will be invoked for closing the resource. However, developers can customize the closing behavior by providing
+ * a different method name through the {@code value} attribute. For example, setting {@code value = "destroy"} will
+ * look for a method named {@code destroy()} to close the resource.
+ * </p>
+ *
+ * <p>
+ * The {@code AutoClose} annotation is retained at runtime, allowing it to be accessed and processed during test execution.
+ * </p>
+ *
+ * @see java.lang.annotation.Retention
+ * @see java.lang.annotation.Target
+ * @since 5.11
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@API(status = API.Status.EXPERIMENTAL, since = "5.11")
+public @interface AutoClose {
+
+	/**
+	 * Specifies the name of the method to invoke for closing the resource.
+	 * The default value is "close".
+	 *
+	 * @return the method name for closing the resource
+	 */
+	String value() default "close";
+
+}
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
@@ -18,29 +18,24 @@ import java.lang.annotation.Target;
 import org.apiguardian.api.API;
 
 /**
- * The {@code AutoClose} annotation is used to automatically close resources used in JUnit 5 tests.
+ * The {@code AutoClose} annotation is used to automatically close resources
+ * used in tests.
  *
- * <p>
- * This annotation should be applied to fields within JUnit 5 test classes. It indicates that the annotated
- * resource should be automatically closed after the test execution. The annotation targets
- * {@link java.lang.annotation.ElementType#FIELD} elements, allowing it to be applied to instance variables.
- * </p>
+ * <p>This annotation should be applied to fields within test classes. It
+ * indicates that the annotated resource should be automatically closed after
+ * the test execution.
  *
- * <p>
- * By default, the {@code AutoClose} annotation expects the annotated resource to provide a {@code close()} method
- * that will be invoked for closing the resource. However, developers can customize the closing behavior by providing
- * a different method name through the {@code value} attribute. For example, setting {@code value = "shutdown"} will
- * look for a method named {@code shutdown()} to close the resource.
- * When multiple annotated resources exist the order of closing them is unspecified.
- * </p>
+ * <p>By default, the {@code AutoClose} annotation expects the annotated
+ * resource to provide a {@code close()} method that will be invoked for closing
+ * the resource. However, developers can customize the closing behavior by
+ * providing a different method name through the {@link #value} attribute. For
+ * example, setting {@code value = "shutdown"} will look for a method named
+ * {@code shutdown()} to close the resource. When multiple annotated resources
+ * exist the order of closing them is unspecified.
  *
- * <p>
- * The {@code AutoClose} annotation is retained at runtime, allowing it to be accessed and processed during test execution.
- * </p>
- *
+ * @since 5.11
  * @see java.lang.annotation.Retention
  * @see java.lang.annotation.Target
- * @since 5.11
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
@@ -49,7 +44,8 @@ public @interface AutoClose {
 
 	/**
 	 * Specifies the name of the method to invoke for closing the resource.
-	 * The default value is "close".
+	 *
+	 * <p>The default value is {@code close}.
 	 *
 	 * @return the method name for closing the resource
 	 */
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
@@ -10,6 +10,7 @@
 
 package org.junit.jupiter.api;
 
+import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
@@ -38,10 +39,11 @@ import org.junit.jupiter.api.extension.ExtendWith;
  * @see java.lang.annotation.Retention
  * @see java.lang.annotation.Target
  */
-@Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
-@API(status = API.Status.EXPERIMENTAL, since = "5.11")
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
 @ExtendWith(AutoCloseExtension.class)
+@API(status = API.Status.EXPERIMENTAL, since = "5.11")
 @SuppressWarnings("exports")
 public @interface AutoClose {
 
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
@@ -42,6 +42,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
 @Target(ElementType.FIELD)
 @API(status = API.Status.EXPERIMENTAL, since = "5.11")
 @ExtendWith(AutoCloseExtension.class)
+@SuppressWarnings("exports")
 public @interface AutoClose {
 
 	/**
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
@@ -16,6 +16,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.apiguardian.api.API;
+import org.junit.jupiter.api.extension.ExtendWith;
 
 /**
  * The {@code AutoClose} annotation is used to automatically close resources
@@ -40,6 +41,7 @@ import org.apiguardian.api.API;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.FIELD)
 @API(status = API.Status.EXPERIMENTAL, since = "5.11")
+@ExtendWith(AutoCloseExtension.class)
 public @interface AutoClose {
 
 	/**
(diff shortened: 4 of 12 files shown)


bjmi:
Unit tests pending.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
```diff
@@ -0,0 +1,101 @@
...
+			}
+			getAndDestroy(field, toBeClosed);
+		};
+	}
+
+	private static void getAndDestroy(Field field, Object toBeClosed) {
+		String methodName = field.getAnnotation(AutoClose.class).value();
+		Method destroyMethod = ReflectionUtils.findMethod(toBeClosed.getClass(), methodName).orElseThrow(
+			() -> new ExtensionConfigurationException("@AutoClose: Cannot resolve the destroy method " + methodName
+					+ "() at " + getQualifiedFieldName(field) + ": " + field.getType().getSimpleName()));
```
marcphilipp, line 93: I think the error message should be:
> "@AutoClose failed to close object for field <declaring-class>.<name> because the <method-name>() method could not be resolved.

@sbrannen Does that sound good to you?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
```diff
@@ -0,0 +1,101 @@
...
+	public void afterEach(ExtensionContext context) {
+		Store contextStore = context.getStore(NAMESPACE);
+
+		for (Object instance : context.getRequiredTestInstances().getAllInstances()) {
+			registerCloseables(contextStore, instance.getClass(), instance);
+		}
+	}
+
+	private void registerCloseables(Store contextStore, Class<?> testClass, /* @Nullable */ Object testInstance) {
+		Predicate<Field> isStatic = testInstance == null ? ReflectionUtils::isStatic : ReflectionUtils::isNotStatic;
```
marcphilipp, line 66: I think we need to make sure the closing of non-static fields is tied to the instance lifecycle, so this works:

```java
@TestInstance(PER_CLASS)
static class Tests {
    
    static boolean CLOSED;
    
    @AutoClose
    AutoCloseable field = () -> CLOSED = true;
    
    @Test
    void test1() {
        assertFalse(CLOSED);
    }
    
    @Test
    void test2() {
        assertFalse(CLOSED);
    }
}
```

To do this, I think we should be able to implement `TestInstancePreDestroyCallback` instead of `AfterEachCallback`.

marcphilipp (changes requested):

ibrahimesseddyq:
@bjmi still working on this?

bjmi:
> @bjmi still working on this?

Yes, I am. I'm just waiting for the confirmation of @sbrannen :)

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
```diff
@@ -0,0 +1,191 @@
...
+		EngineExecutionResults engineExecutionResults = executeTestsForClass(AutoCloseTestCase.class);
+
+		Events tests = engineExecutionResults.testEvents();
+		tests.assertStatistics(stats -> stats.succeeded(2));
+		// @formatter:off
+		assertEquals(asList(
+				"afterEach-close()", "afterEach-run()",
+				"afterEach-close()", "afterEach-run()",
+					"afterAll-close()"), recorder);
+		// @formatter:onf
```
sbrannen, line 59:
```suggestion
		assertThat(recorder).containsExactly(
				"afterEach-close()",
				"afterEach-run()",
				"afterEach-close()",
				"afterEach-run()",
				"afterAll-close()");
		// @formatter:on
```

For newly introduced functionality, we typically prefer the use of AssertJ assertions.

It would be great if you could switch to AssertJ for assertions in this test class. If  not, I'll likely do it after the work has been merged.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
```diff
@@ -0,0 +1,191 @@
...
+
+			assertNotNull(closable);
+			assertNotNull(runnable);
+			assertNull(nullField);
+		}
+
+		@BeforeAll
+		static void setup() {
+			staticClosable = new AutoCloseSpy("afterAll-");
+		}
```
sbrannen, line 130: Please move this to the top of the class above the test methods.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
```diff
@@ -0,0 +1,191 @@
...
+
+		@Override
+		public void close() {
+			checkIfAlreadyInvoked();
+			recordInvocation("close()");
+		}
+
+		private void checkIfAlreadyInvoked() {
+			if (!invokedMethod.isEmpty())
+				throw new IllegalStateException();
```
sbrannen, line 181: We do not use if-statements without blocks (`{...}`) in JUnit 5.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
```diff
@@ -0,0 +1,58 @@
...
+ * </p>
+ *
+ * @see java.lang.annotation.Retention
+ * @see java.lang.annotation.Target
+ * @since 5.11
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@API(status = API.Status.EXPERIMENTAL, since = "5.11")
+public @interface AutoClose {
```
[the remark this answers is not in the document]
bjmi: Great idea but unfortunately `AutoCloseExtension` resides in junit-jupiter-engine and there is no compile-time dependency to it.  
Additionally it is proposed to make `AutoCloseExtension.class` package-private. This also contradicts to the usage in `@ExtendWith` as `@AutoClose` and `AutoCloseExtension` reside in different packages.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	static class AutoCloseTestCase {
+
+		private static @AutoClose AutoCloseable staticClosable;
+		private static @AutoClose AutoCloseable nullStatic;
+
+		private final @AutoClose AutoCloseable closable = new AutoCloseSpy("afterEach-");
+		private final @AutoClose("run") Runnable runnable = new AutoCloseSpy("afterEach-");
+		private @AutoClose AutoCloseable nullField;
```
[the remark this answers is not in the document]
bjmi: Applied.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
bjmi: Applied. But...
What about single return? 😁 
```
if (invokedMethod.isEmpty()) return;
throw new IllegalStateException();
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/MutableExtensionRegistry.java
```diff
@@ -52,7 +52,8 @@ public class MutableExtensionRegistry implements ExtensionRegistry, ExtensionReg
 		new TimeoutExtension(), //
 		new RepeatedTestExtension(), //
 		new TestInfoParameterResolver(), //
-		new TestReporterParameterResolver()));
+		new TestReporterParameterResolver(), //
+		new AutoCloseExtension()));
```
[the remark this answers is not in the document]
bjmi: The `@ExtendWith(AutoCloseExtension.class)` approach doesn't work atm thus I'll leave the auto-registration in `MutableExtensionRegistry`.  
I think it can be changed to the on demand approach later as well.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
bjmi: @sbrannen do you agree?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
bjmi: Applied.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
sbrannen: Yes, I agree that the lifecycle of an auto-closeable instance field should be scoped to the test instance lifecycle.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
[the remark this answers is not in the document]
sbrannen: Sorry. I overlooked that in the GitHub UI.

The `AutoCloseExtension` needs to be package-private and reside in the same package/module as `@AutoClose`. So.. in `junit-jupiter-api`.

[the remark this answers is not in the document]
sbrannen: This approach is analogous to all of the `ExecutionCondition` implementations in the  `org.junit.jupiter.api.condition` package.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
```diff
@@ -0,0 +1,101 @@
...
+						+ "  because it was null.");
+				return;
+			}
+			invokeCloseMethod(field, toBeClosed);
+		};
+	}
+
+	private static void invokeCloseMethod(Field field, Object toBeClosed) {
+		String methodName = field.getAnnotation(AutoClose.class).value();
+		Method destroyMethod = ReflectionUtils.findMethod(toBeClosed.getClass(), methodName).orElseThrow(
```
sbrannen, line 90:
```suggestion
		Method closeMethod = ReflectionUtils.findMethod(toBeClosed.getClass(), methodName).orElseThrow(
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
```diff
@@ -0,0 +1,101 @@
...
+			() -> new ExtensionConfigurationException(
+				"@AutoClose failed to close object for field " + getQualifiedFieldName(field) + " because the "
+						+ methodName + "() method could not be " + "resolved."));
+		ReflectionUtils.invokeMethod(destroyMethod, toBeClosed);
```
sbrannen, line 94:
```suggestion
		ReflectionUtils.invokeMethod(closeMethod, toBeClosed);
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/MutableExtensionRegistry.java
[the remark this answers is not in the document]
sbrannen: Please see my other comments about moving the extension to the `junit-jupiter-api` module.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
```diff
@@ -0,0 +1,197 @@
...
+	@BeforeEach
+	void resetRecorder() {
+		recorder.clear();
+	}
+
+	@Test
+	void fieldsAreProperlyClosed() {
+		EngineExecutionResults engineExecutionResults = executeTestsForClass(AutoCloseTestCase.class);
+
+		Events tests = engineExecutionResults.testEvents();
```
sbrannen, line 53:
```suggestion
		Events tests = executeTestsForClass(AutoCloseTestCase.class).testEvents();
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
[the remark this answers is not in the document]
bjmi: After moving `AutoCloseExtension` in the `org.junit.jupiter.api` package alongside `@AutoClose` in the junit-jupiter-api module the compiler emits following error:
```
junit5\junit-jupiter-api\build\module\org.junit.jupiter.api\org\junit\jupiter\api\AutoClose.java:44: warning: [exports] class AutoCloseExtension in module org.junit.jupiter.api is not accessible to clients that require this module
@ExtendWith(AutoCloseExtension.class)
            ^
error: warnings found and -Werror specified
1 error
1 warning

error: warnings found and -Werror specified
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
bjmi: Swichted to TestInstancePreDestroyCallback.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
bjmi: Applied.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java
bjmi: Applied.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
bjmi: Inlined.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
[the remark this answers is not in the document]
bjmi: Can we leave it as a built-in extension for now?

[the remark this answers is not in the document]
sbrannen: Before we go back to the "built-in extension", please let me know what happens when you add this:

```java
opens org.junit.jupiter.api to org.junit.platform.commons;
```

to:

https://github.com/junit-team/junit5/blob/main/junit-jupiter-api/src/module/org.junit.jupiter.api/module-info.java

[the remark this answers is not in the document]
sbrannen: For details,

- see #1877

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
sbrannen: Nope, we don't do that either.

sbrannen: TBH, for stuff like that we often use `Preconditions` to achieve the goal with a one-liner.

documentation/src/test/java/example/AutoCloseDemo.java
```diff
@@ -0,0 +1,40 @@
...
+
+	@Test
+	void usersTableHasEntries() throws SQLException {
+		ResultSet resultSet = connection.createStatement().executeQuery("SELECT * FROM users");
+
+		assertTrue(resultSet.next());
+	}
+
+	AutoCloseDemo() throws SQLException {
+	}
```
sbrannen, line 37: No need for that constructor.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * {@code AutoCloseExtension} is a JUnit Jupiter extension that closes resources
+ * if a field in a test class is annotated with {@link AutoClose @AutoClose}.
+ *
+ * <p>Consult the Javadoc for {@link AutoClose @AutoClose} for details on the
+ * contract.
+ *
+ * @since 5.11
+ * @see AutoClose
+ * @see AutoCloseable
```
sbrannen, line 39: Reference to `AutoCloseable` is now obsolete.

documentation/src/test/java/example/AutoCloseDemo.java
bjmi: Unfortunately `DriverManager.getConnection(url)` throws a checked SQLException and the used ctor + throws allows writing the given test.
Maybe I can elaborate a better example.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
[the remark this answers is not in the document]
bjmi: Put `opens org.junit.jupiter.api to org.junit.platform.commons;` one line above the existing `opens org.junit.jupiter.api.condition to org.junit.platform.commons;`.
Result:
```
Task :junit-jupiter-api:compileModule
C:\dev\git-repos\junit5\junit-jupiter-api\build\module\org.junit.jupiter.api\org\junit\jupiter\api\AutoClose.java:44: warning: [exports] class AutoCloseExtension in module org.junit.jupiter.api is not accessible to clients that require this module
@ExtendWith(AutoCloseExtension.class)
            ^
error: warnings found and -Werror specified
error: warnings found and -Werror specified


> Task :junit-jupiter-api:compileModule FAILED
1 error
1 warning
```
Back to built-in extension?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java
bjmi: Removed.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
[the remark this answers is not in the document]
sbrannen: If we cannot get it to work, we can switch back to the built-in extension and follow up on `main` after the PR has been merged.

@sormuras, do you know why this works for `ExecutionCondition` implementations but not for the `AutoCloseExtension`?

documentation/src/test/java/example/AutoCloseDemo.java
sbrannen: Ahhh... OK. The reason wasn't immediately clear in the GitHub UI, but I get it now.

How about this instead?

```java
class AutoCloseDemo {

	@AutoClose
	Connection connection = getJdbcConnection("jdbc:mysql://localhost/testdb");

	@Test
	void usersTableHasEntries() throws SQLException {
		ResultSet resultSet = connection.createStatement().executeQuery("SELECT * FROM users");

		assertTrue(resultSet.next());
	}

	private static Connection getJdbcConnection(String url) {
		try {
			return DriverManager.getConnection(url);
		}
		catch (SQLException ex) {
			throw new RuntimeException(ex);
		}
	}

}
```

That at least makes the code clear when viewed in the User Guide.

Also, we could _hide_ the implementation of `getJdbcConnection()` in the User Guide.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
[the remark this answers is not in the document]
sbrannen: OK, I figured it out.

You just need to annotate `@AutoClose` with the following.

```java
@SuppressWarnings("exports")
```

Note that `@EnabledOnJre` and all other `ExecutionCondition` annotations also suppress the `exports` warning.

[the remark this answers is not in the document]
sbrannen: @marcphilipp apparently added that suppression for execution condition annotations in https://github.com/junit-team/junit5/commit/a76e1a5da74f97e0ac335fc874265bb5454b2866.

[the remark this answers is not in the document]
bjmi: Applied. `@SuppressWarnings("exports")`

documentation/src/test/java/example/AutoCloseDemo.java
bjmi: Parameterless `Constructor() throws Exception { }` is kind a "pattern" to allow writing field declarations easily without catching all the thrown exceptions, having extracting them into methods or put everything in setup methods which declare `throws Exception` instead.

sbrannen: Yes, I am aware of the language feature; however, the point of an example in the User Guide is to demonstrate a specific feature _without_ superfluous noise. 

For this feature, our goal is to demonstrate how to use `@AutoClose` on a field whose type should be closed. The reader does not need to be concerned with any checked exceptions thrown by the code that we choose to use to create such a field.

If you don't want to make the changes, that's fine. I can address this while merging the PR.

In any case, my current plan is to move `getConnection()` to a `JdbcUtils` class (which is not shown in the example) and invoke that via `JdbcUtils.getConnection(...)`.

This is somewhat analogous to the `WebClient` examples we have, where we do not show the implementation of `WebClient`. So I might even refactor the `@AutoClose` example to use `WebClient` instead of a JDBC `Connection` and point out that the `WebClient` has a `close()` method that needs to be invoked. Anyway, I'm just brainstorming, but hopefully you get the point.

sbrannen: FWIW, the `Statement` and `ResultSet` in the current example also need to be closed.

In light of that, we definitely need to rework the example.

documentation/src/docs/asciidoc/link-attributes.adoc
```diff
@@ -153,6 +153,7 @@ endif::[]
 // Jupiter Engine
 :junit-jupiter-engine:                       {javadoc-root}/org.junit.jupiter.engine/org/junit/jupiter/engine/package-summary.html[junit-jupiter-engine]
 // Jupiter Extension Implementations
+:AutoCloseExtension:                         {current-branch}/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/AutoCloseExtension.java[AutoCloseExtension]
```
sbrannen, line 156:
```suggestion
:AutoCloseExtension:                         {current-branch}/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java[AutoCloseExtension]
```

This line also needs to be relocated to the Jupiter API section.

documentation/src/test/java/example/AutoCloseDemo.java
sbrannen: I just noticed you already incorporated the `getJdbcConnection()` approach. Thanks.

sbrannen (approved):
Thanks for making all of the requested changes! 👍 

All that's left are the changes to the link attributes and the demo.

Though we can take care of those for you if you don't have time.

Note, however, that there is no rush. I don't plan to merge this until 2024. 😉 

Enjoy your end-of-year celebrations!

sbrannen:
One last thing: this also needs an entry in `release-notes-5.11.0-M1.adoc`.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -45,6 +45,7 @@ in the `junit-jupiter-api` module.
 | `@ExtendWith`            | Used to <<extensions-registration-declarative,register extensions declaratively>>. Such annotations are _inherited_.
 | `@RegisterExtension`     | Used to <<extensions-registration-programmatic,register extensions programmatically>> via fields. Such fields are _inherited_ unless they are _shadowed_.
 | `@TempDir`               | Used to supply a <<writing-tests-built-in-extensions-TempDirectory,temporary directory>> via field injection or parameter injection in a lifecycle method or test method; located in the `org.junit.jupiter.api.io` package.
+| `@AutoClose`             | Denotes that the annotated field represents a resource that should be automatically closed after test execution.
```
hertzsprung, line 48: Suggest (if I've understood correctly):
```suggestion
| `@AutoClose`             | Denotes that the annotated field represents a resource that will be automatically closed after each test execution.
```

sbrannen: Not necessarily. It depends on the scope of the field and the test instance lifecycle.

A closeable instance field will be closed after each test execution when the test instance lifecycle is "per test method".

But for test instance lifecycle "per test class", the closeable field will only be closed when the test instance is destroyed.

sbrannen: Though we can improve the wording to hint at "scope" rules and ensure the scope rules are clearly covered in the Javadoc.

documentation/src/test/java/example/AutoCloseDemo.java
bjmi: > FWIW, the `Statement` and `ResultSet` in the current example also need to be closed.

I'm quite sure that the statement and resultset are also closed when the owning connection is closed and makes this a concise unit test. But there are definitely better examples for the user guide e.g. your proposed WebClient example.
I just want this PR merged that we can get rid of our own AutoCloseExtension ;)
Merry Christmas!

bjmi:
> One last thing: this also needs an entry in `release-notes-5.11.0-M1.adoc`.

Pushed.
Are the commits squashed before the merge or should I do it in advance?

sbrannen:
> Pushed.

Thanks

> Are the commits squashed before the merge or should I do it in advance?

We typically allow uses to submit multiple commits for a PR in order to maintain the history in the conversation, and we then squash all commits into a single commit when merging.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
```diff
@@ -0,0 +1,57 @@
...
+ * @since 5.11
+ * @see java.lang.annotation.Retention
+ * @see java.lang.annotation.Target
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@API(status = API.Status.EXPERIMENTAL, since = "5.11")
+@ExtendWith(AutoCloseExtension.class)
+@SuppressWarnings("exports")
+public @interface AutoClose {
```
sbrannen, line 46: Just noticed this is missing `@Documented`, but we can add that after merging if necessary.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java
```diff
@@ -0,0 +1,99 @@
...
+		Class<?> testClass = context.getRequiredTestClass();
+
+		registerCloseables(contextStore, testClass, null);
+	}
+
+	@Override
+	public void preDestroyTestInstance(ExtensionContext context) {
+		Store contextStore = context.getStore(NAMESPACE);
+
+		for (Object instance : context.getRequiredTestInstances().getAllInstances()) {
```
sbrannen, line 57: I'm not sure we need to get all test instances here.

Rather, I think it should suffice to get only the current test instance.

In other words, if an enclosing class has any `@AutoClose` resources, those should be managed for the test instance of that enclosing class.

Or am I missing something?

sbrannen: In order to ensure we have the correct semantics here, we'll need to introduce various `@Nested` test cases.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoClose.java
bjmi: Applied.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java
sbrannen: For example, all tests in `AutoCloseTests` still pass if you change the method body to the following.

```java
		Store contextStore = context.getStore(NAMESPACE);
		Object testInstance = context.getRequiredTestInstance();
		registerCloseables(contextStore, testInstance.getClass(), testInstance);
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java
```diff
@@ -0,0 +1,99 @@
...
+			}
+		});
+	}
+
+	private static Store.CloseableResource asCloseableResource(Object testInstance, Field field) {
+		return () -> {
+			Object toBeClosed = ReflectionUtils.tryToReadFieldValue(field, testInstance).get();
+			if (toBeClosed == null) {
+				logger.warn(() -> "@AutoClose couldn't close object for field " + getQualifiedFieldName(field)
+						+ "  because it was null.");
```
sbrannen, line 79:
```suggestion
						+ " because it was null.");
```

Please also verify that a `WARNING` is logged in `AutoCloseTests` using our `@TrackLogRecords` support.

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
```diff
@@ -0,0 +1,224 @@
...
+		assertThrows(IllegalStateException.class, spy::close);
+		assertThrows(IllegalStateException.class, spy::run);
+		assertEquals(asList("close()"), recorder);
+	}
+
+	@Test
+	void instancePerClass() {
+		Events tests = executeTestsForClass(AutoCloseInstancePerClassTestCase.class).testEvents();
+		tests.assertStatistics(stats -> stats.succeeded(2));
+	}
```
sbrannen, line 99:
```suggestion
		assertThat(AutoCloseInstancePerClassTestCase.closed).isTrue();
	}
```

junit-jupiter-engine/src/test/java/org/junit/jupiter/engine/extension/AutoCloseTests.java
```diff
@@ -0,0 +1,224 @@
...
+ * Copyright 2015-2023 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.engine.extension;
```
sbrannen, line 11: Please move the test class to `org.junit.jupiter.api`.

sbrannen:
Hey @bjmi,

Since I started investigating the support for `@Nested` test classes, I ended up putting in a fair amount of work locally on my machine.

Consequently, I have decided to mark this PR as a "draft", and I will continue the rest of the work on my branch, which you can see here: https://github.com/junit-team/junit5/compare/main...sbrannen:junit5:issues/3592-AutoClose.

In light of that, there's no need to make further modifications to this PR. All of your hard work is contained in the first commit on the above branch! 👍

sbrannen:
### Update

I force-pushed my latest work to the aforementioned [branch](https://github.com/junit-team/junit5/compare/main...sbrannen:junit5:issues/3592-AutoClose).

I added several precondition checks (primitive, array, blank method name) and corresponding tests.

I also added tests for `@Nested` behavior, and I'm wondering if we have a bug 🐛 in our `TestInstancePreDestroyCallback` support. Specifically, it appears that the test instance of the enclosing class gets _destroyed_ **before** the instance of the `@Nested` test class (or at least the `TestInstancePreDestroyCallback` gets invoked in such a fashion).

```java
@Test
void fieldsAreProperlyClosedWithNestedTestClasses() {
	Events tests = executeTestsForClass(EnclosingTestCase.class).testEvents();
	tests.assertStatistics(stats -> stats.succeeded(2));
	assertThat(recorder).containsExactly(//
		"EnclosingTestCase-afterTestInstance-close()", //
		"NestedTestCase-afterTestInstance-close()", //
		"NestedTestCase-afterAll-close()", //
		"EnclosingTestCase-afterAll-close()"//
	);
}
``` 

I would actually expect that `"EnclosingTestCase-afterTestInstance-close()"` comes **after** 
`"NestedTestCase-afterAll-close()"` and **before** `"EnclosingTestCase-afterAll-close()"`.

@junit-team/junit-5, does that ring a bell to anyone?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java
```diff
@@ -0,0 +1,99 @@
...
+ * Copyright 2015-2023 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api;
```
sbrannen, line 11: FYI: having this in the `api` package unfortunately introduces a package cycle between `api` and `api.extension`. 😞 

So I've reverted to the "built-in extension" approach in my branch.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AutoCloseExtension.java
```diff
@@ -0,0 +1,99 @@
...
+		Store contextStore = context.getStore(NAMESPACE);
+
+		for (Object instance : context.getRequiredTestInstances().getAllInstances()) {
+			registerCloseables(contextStore, instance.getClass(), instance);
+		}
+	}
+
+	private void registerCloseables(Store contextStore, Class<?> testClass, Object testInstance) {
+		Predicate<Field> predicate = testInstance == null ? ReflectionUtils::isStatic : ReflectionUtils::isNotStatic;
+		findAnnotatedFields(testClass, AutoClose.class, predicate).forEach(field -> {
```
sbrannen, line 64: The search for fields needs to use `HierarchyTraversalMode.BOTTOM_UP` semantics.

Otherwise, `@AutoClose` fields in superclasses are closed _before_ `@AutoClose` fields in subclasses -- which is undesirable.

So I've made this change in my branch.

sbrannen:
### Update

I sorted out all open issues and force-pushed to the aforementioned [branch](https://github.com/junit-team/junit5/compare/main...sbrannen:junit5:issues/3592-AutoClose).

At this point, I believe the work is "feature complete" and ready for review in case anyone wants to take a look.

All that's left is to clarify the "scope" of such resource cleanup in the documentation (see https://github.com/junit-team/junit5/pull/3592#discussion_r1435809802).

documentation/src/test/java/example/AutoCloseDemo.java
sbrannen: I went with `WebClient`.

You can see the result in my [branch](https://github.com/junit-team/junit5/compare/main...sbrannen:junit5:issues/3592-AutoClose).

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sbrannen: This has been addressed in my [branch](https://github.com/junit-team/junit5/compare/main...sbrannen:junit5:issues/3592-AutoClose).

sbrannen:
This has been merged into `main` in e32c2e3b75aa26d763db8f32e00e93879377d14d and revised in 89a0201073efd12b0732ceaeba52c6d7d7029f3a.

Thanks again for the contribution! 👍

PR #3621, opened 2023-12-30
Measure elapsed time with `System.nanoTime()` in tests

bjmi:
## Overview

java.lang.System.nanoTime() isn't affected by system clock changes as opposed to `currentTimeMillis()`.

An example can be found in [nanoTime() javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--).

---


sbrannen:

Thanks

PR #3629, opened 2024-01-04
Try out v3-beta of gradle-build-action

marcphilipp:


marcphilipp:
Works as expected but we should wait for the official release

Issue #3153, opened 2023-02-10
Document benefits of `messageSupplier` in `Assertions`


manueljordan:
About JUnit 5 documentation in the [2.5. Assertions](https://junit.org/junit5/docs/current/user-guide/#writing-tests-assertions) section, exists the following `@Test` method

```java
    @Test
    void standardAssertions() {
        assertEquals(2, calculator.add(1, 1));
        assertEquals(4, calculator.multiply(2, 2),
                "The optional failure message is now the last parameter");
        assertTrue('a' < 'b', () -> "Assertion messages can be lazily evaluated -- "
                + "to avoid constructing complex messages unnecessarily.");
    }
```

Is not clear what is the advantage to use `() -> "Assertion messages can be lazily evaluated -- ..."` and not simply `"Assertion messages can be lazily evaluated -- ..."` It means: `() -> ""` vs `""` 

Therefore:

```java
        assertTrue('a' < 'b', () -> "Assertion messages can be lazily evaluated -- "
                + "to avoid constructing complex messages unnecessarily.");
        assertTrue('a' < 'b', "Assertion messages can be lazily evaluated -- "
                + "to avoid constructing complex messages unnecessarily.");
```

What does _Assertion messages can be lazily evaluated_ mean?  The explanation would be valuable for a better understanding and to know explicitly when use mandatorily an approach over the other. Currently in the complete/full documentation only appears through a sample code the `() -> "something"` approach just twice, but not more. Some example(s) is/are appreciate showing the two approaches together and to understand clearly the difference of use of one over the other. Therefore I want to know when would be mandatory use the _lambda_ approach. So far, the current official documentation does not cover this explicitly

I already wrote a question on SO about this question at:

* [JUnit 5: What does "Assertion messages can be lazily evaluated" mean?](https://stackoverflow.com/questions/75348941/junit-5-what-does-assertion-messages-can-be-lazily-evaluated-mean)

The answer has sense, but if it can be covered and _expanded_ from the official source documentation would be nice for the community

Thanks for your understanding

marcphilipp:
**Team decision:** Change the documentation to contain a more meaningful example.

marcphilipp:
@manueljordan Would you like to take a stab at this?

manueljordan:
[...]

PR #3635, opened 2024-01-10
Document benefits of `messageSupplier` in `Assertions`

ankitwasankar:
## Overview

Documented the benefits of MessageSupplier in assertion, as per the discussion on issue #3153 

---

---

### Definition of Done

### DIFF
--- a/documentation/src/test/java/example/AssertionsDemo.java
+++ b/documentation/src/test/java/example/AssertionsDemo.java
@@ -39,10 +39,14 @@ class AssertionsDemo {
 	@Test
 	void standardAssertions() {
 		assertEquals(2, calculator.add(1, 1));
+		// When assertion failure message is used as a plain expression,
+		// the expression will be evaluated irrespective of the assertion result.
 		assertEquals(4, calculator.multiply(2, 2),
-				"The optional failure message is now the last parameter");
-		assertTrue('a' < 'b', () -> "Assertion messages can be lazily evaluated -- "
-				+ "to avoid constructing complex messages unnecessarily.");
+				"This expression will always evaluate irrespective of assertion result.");
+		// When assertion failure message is created as a lambda expression,
+		// expression will only evaluate when assertion result is failed (lazy evaluation)
+		// and can help us save unnecessary computations when they are not needed.
+		assertTrue('a' < 'b', () -> "This lambda will evaluate only if assertion fails.");
 	}
 
 	@Test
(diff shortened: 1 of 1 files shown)


sbrannen (changes requested):
Let's take a different approach.

I actually like the original _text_ in the assertion failure messages (perhaps because I wrote them), and I think we should keep that content.

So, let's move that text to a new paragraph before the example code in the User Guide, and let's incorporate some of the text you wrote in that paragraph -- in other words, combine the existing text with your text to come up with something more comprehensive.

After doing that, let's go back to an example similar to the one you proposed in https://github.com/junit-team/junit5/issues/3153#issuecomment-1880076320. Having plain strings in both variants (`String` and `Supplier<String>`) is what was wrong with the example to begin with. Using a lambda expression or method reference for the `messageSupplier` will make it much easier to explain the benefits.

ankitwasankar:
@sbrannen - thank you for the review comment, I have updated the code.
Please let me know if you think something needs to be added or removed.

sbrannen (changes requested):
Hi @ankitwasankar,

I think you misunderstood what I was requesting, so I'll restate that here.

> I like the original text in the assertion failure messages (perhaps because I wrote them), and I think we should keep that content.

By "content" I mean the text/explanation which should retain mostly intact but in the "Assertions" section of the User Guide, not in the included `AssertionsDemo` code as inline comments.

> So, let's move that text to a new paragraph before the example code in the User Guide, and let's incorporate some of the text you wrote in the same paragraph -- in other words, combine the existing text with your text to come up with something more comprehensive.

I'd still like to see this.

> After doing that, let's go back to an example similar to the one you proposed in https://github.com/junit-team/junit5/issues/3153#issuecomment-1880076320. Having plain strings in both variants (String and Supplier<String>) is what was wrong with the example to begin with. 

That last part is the most important. There is no point in having a static string returned from a lambda expression.

The whole reason we support a `messageSupplier` is to avoid unnecessary creation of _dynamic_ strings.

Please note as well that the concatenation of two static strings does not constitute a dynamic strings since the compiler will likely convert them to a single static string during compilation.

> Using a lambda expression or method reference for the `messageSupplier` will make it much easier to explain the benefits.

Again, that's what we're trying to convey here, and your `getMessage()` example in https://github.com/junit-team/junit5/issues/3153#issuecomment-1880076320 achieves that.

I think we can also remove the `assertWithMessageSupplier()` test method and include all "standard assertions" in the existing `standardAssertions()` method.

Please make the necessary changes to achieve the above.

Thanks!

marcphilipp:
Closing in favor of #3153.

PR #3667, opened 2024-01-30
Ignore `NoSuchFileException` in `TempDir` cleanup

gilday:
## Overview

In `TempDirectory`'s cleanup logic, while walking the temp directory, the file walker may fail to visit a file that has been removed by another thread or process. In this case, JUnit should ignore the `NoSuchFileException`, because there's nothing to clean up.

I have observed such failures when using JGit to manage repositories in temporary directories. JGit uses lock files that it cleans up in background threads. JGit and JUnit will race to delete the lock file, and the JUnit `TempDirectory` cleanup fails when JGit wins that race.

I considered what an automated test for this new behavior would look like. I could add a clear-box test that uses high-level concurrency objects to make sure that the test can delete a file between the time that the `SimpleFileVisitor` lists the temp directory and the time that the `SimpleFileVisitor` attempts to delete the file, but this would require adding new seams to `TempDirectory` that are inconsistent with the existing tests for that type. Would you accept this change without adding new test coverage?

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.11.0-M1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.11.0-M1.adoc
@@ -67,6 +67,8 @@ repository on GitHub.
     *** The same applies to other types of test methods (`@TestFactory`,
       `@ParameterizedTest`, etc.) as well as lifecycle methods (`@BeforeAll`,
       `@AfterAll`, `@BeforeEach`, and `@AfterEach`).
+* `TempDir` suppresses `NoSuchFileException` when deleting files that may have been deleted
+  by another thread or process.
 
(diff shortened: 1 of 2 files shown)


marcphilipp (approved):

marcphilipp:
> I considered what an automated test for this new behavior would look like. I could add a clear-box test that uses high-level concurrency objects to make sure that the test can delete a file between the time that the `SimpleFileVisitor` lists the temp directory and the time that the `SimpleFileVisitor` attempts to delete the file, but this would require adding new seams to `TempDirectory` that are inconsistent with the existing tests for that type. Would you accept this change without adding new test coverage?

Given that such a test would likely be flaky I don't think adding it would be justified. I'd be okay with making an exception in this case.

marcphilipp:
@gilday Thank you for your contribution! :+1:

Issue #3108, opened 2022-12-15
JUnit parallel executor running too many tests with a fixed policy.


OPeyrusse:
JUnit only parallel executor service relies on a ForkJoinPool. Unfortunately, this does not play well with code also using ForkJoinPools.

The observed issue is that, when activating parallel tests, JUnit uses `ForkJoinPoolHierarchicalTestExecutorService`. However, our tests are also using `ForkJoinPool`s and `ForkJoinTask`s. The orchestration of the test awaits for the completion of those tasks before moving on in the tests.
But the issue is that `ForkJoinTask` and `ForkJoinWorkerThread` are capable of detecting the use of the FJP framework ([here](https://github.com/openjdk/jdk/blob/98fa48c330941efe6588a907b383802a11ed0e6b/src/java.base/share/classes/java/util/concurrent/ForkJoinTask.java#L411-L415) for example) and react to it. As JUnit tasks and the project tasks are in different ForkJoinPools, they cannot help each other. This only results in more tests being started by already running and **incomplete** tests.

This can be an issue when tests are resource sensitive. For example, it may not be possible to open too many connections to a Database.
Tough not explicitly illustrated in this project, we also faced `StackOverflowError` because of recursive test executions in a single worker. In the following logs, produced by the reproducing project, we can see that two workers are recursively starting tests before finishing any:
```console
ForkJoinPool-1-worker-1 starting a new test. Running: 2
ForkJoinPool-1-worker-2 starting a new test. Running: 1
ForkJoinPool-1-worker-1 starting a new test. Running: 3
ForkJoinPool-1-worker-1 starting a new test. Running: 4
ForkJoinPool-1-worker-1 starting a new test. Running: 5
ForkJoinPool-1-worker-2 starting a new test. Running: 6
ForkJoinPool-1-worker-1 starting a new test. Running: 7
ForkJoinPool-1-worker-1 starting a new test. Running: 8
ForkJoinPool-1-worker-1 starting a new test. Running: 9
ForkJoinPool-1-worker-2 starting a new test. Running: 10
ForkJoinPool-1-worker-1 starting a new test. Running: 11
ForkJoinPool-1-worker-1 starting a new test. Running: 12
ForkJoinPool-1-worker-1 starting a new test. Running: 14
ForkJoinPool-1-worker-2 starting a new test. Running: 13
ForkJoinPool-1-worker-1 starting a new test. Running: 15
ForkJoinPool-1-worker-1 starting a new test. Running: 16
[...]

PR #3685, opened 2024-02-10
Add support for non-fork-join test executor

asardaes:
## Overview

Since there wasn't much feedback in #3108, I figured I'd open a PR with what I have to see if it helps. I reiterate some comments from the linked issue and add some more info.

The idea is to allow a different executor with a fixed thread pool to execute actual tests in parallel in order to avoid potential conflicts with user code if they also rely on the JVM's fork-join APIs. The default would remain as before. Users could then change the executor via configuration.

I wrote a couple of TODOs, but they're both related to my understanding of `CONCURRENT` execution mode semantics. The current implementation checks execution mode in both `submit` and `forkConcurrentTasks` + `executeNonConcurrentTasks`, but in both scenarios, if the task is *not* concurrent, it executes it in the current thread, and that doesn't know if some other thread in the pool has stolen work and is executing another test, so is it really guaranteeing that no other test is running at the same time?

I added a method `ExclusiveTask#await` and use it instead of `join` in `joinConcurrentTasksInReverseOrderToEnableWorkStealing` - if the task should execute in a `ForkJoinWorkerThread`, `await` delegates to `join`, otherwise it delegates to `compute`. If I don't do it like this, the tests execute fine and are even show as PASSED in the terminal, but after that everything stalls and all threads in the fork-join pool remain parked, and I cannot figure out why. I tried to reproduce this outside of a JUnit 5 session and I couldn't.

I added `FixedThreadsParallelExecutionIntegrationTests` as an almost identical copy of `ParallelExecutionIntegrationTests` but using the fixed thread pool for test execution, and I can see that there are a lot of failures. I spent some time looking through the logic to see if I could figure out how it's set up, but I couldn't make much progress, I would appreciate feedback if I'm understanding the tests incorrectly. I will point out something more about this in the comments below.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
@@ -13,6 +13,7 @@ package org.junit.platform.engine.support.hierarchical;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static org.apiguardian.api.API.Status.STABLE;
 import static org.junit.platform.engine.support.hierarchical.Node.ExecutionMode.CONCURRENT;
+import static org.junit.platform.engine.support.hierarchical.Node.ExecutionMode.SAME_THREAD;
 
 import java.lang.Thread.UncaughtExceptionHandler;
 import java.lang.reflect.Constructor;
@@ -21,6 +22,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ForkJoinPool.ForkJoinWorkerThreadFactory;
 import java.util.concurrent.ForkJoinTask;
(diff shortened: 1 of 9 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
```diff
@@ -112,38 +122,30 @@ private static Callable<ForkJoinPool> sinceJava7ConstructorInvocation(ParallelEx
...
-		if (!isAlreadyRunningInForkJoinPool()) {
-			// ensure we're running inside the ForkJoinPool so we
-			// can use ForkJoinTask API in invokeAll etc.
-			return forkJoinPool.submit(exclusiveTask);
+	@SuppressWarnings("try")
+	private static void executeWithLock(TestTask testTask) {
+		ResourceLock resourceLock = testTask.getResourceLock();
+		if (!(Thread.currentThread() instanceof ForkJoinWorkerThread)) {
+			resourceLock.release();
+		}
```
asardaes, line 130: The `canRunTestsIsolatedFromEachOtherWithNestedCases` test from `FixedThreadsParallelExecutionIntegrationTests` was stuck in a deadlock until I added these lines. I spent some time debugging and it seems to me that all of the `org.junit.platform.engine.support.hierarchical.SingleLock` instances created for the tasks are somehow locked by a fork-join thread before the tasks are scheduled in the fixed thread pool, but no matter how many breakpoints I added, I couldn't figure out where that's happening.

marcphilipp:
@asardaes Thank you for your PR! It will be some time before we can get to it but I plan to look into this as part of the work I'm doing for the [Sovereign Tech Fund](https://github.com/junit-team/junit5/discussions/4014) so there is hope. :wink:

marcphilipp:
Thanks for submitting this PR, @asardaes! It was a good starting point for me to explore this topic. I've just created #5027 which provides a less intrusive alternative. I'd appreciate your feedback!

Closing in favor of #5027.

Issue #3693, opened 2024-02-16
Suite configuration is not applied to tests selected by unique ID


mpkorstanje:
## Steps to reproduce

Given a suite with some configuration:

```java
@Suite
@ConfigurationParameter(key = Constants.DEFAULT_TEST_INSTANCE_LIFECYCLE_PROPERTY_NAME, value = "per_class")
@SelectClasses(ExampleTest.class)
public class SuiteTest {
}
```

And a test that is sensitive to that configuration

```java
class ExampleTest {

    boolean shared = false;

    @Test
    void test1(){
        if(!shared) {
            System.out.println("Hello world 1");
        } else {
            System.out.println("World was already greeted");
        }
        shared = true;

    }

    @Test
    void test2(){
        if(!shared) {
            System.out.println("Hello world 2");
        } else {
            System.out.println("World was already greeted");
        }
        shared = true;
    }
}
```

Then executing the tests either by their uniqueId or the `SuiteTest` class should not make a difference in execution:

```java
public class TestRunner {

    public static void main(String[] args) {
        System.out.println("By suite");

        LauncherDiscoveryRequest requestA = request()
                .selectors(
                        DiscoverySelectors.selectClass(SuiteTest.class)
                )
                .build();

        LauncherFactory.create().execute(requestA);

        System.out.println("By unique id");

        LauncherDiscoveryRequest requestB = request()
                .selectors(
                        DiscoverySelectors.selectUniqueId(
                                UniqueId.forEngine("junit-platform-suite")
                                        .append("suite", "com.example.SuiteTest")
                                        .appendEngine("junit-jupiter")
                                        .append("class", "com.example.ExampleTest" )
                                        .append("method", "test1()")
                        ),
                        DiscoverySelectors.selectUniqueId(
                                UniqueId.forEngine("junit-platform-suite")
                                        .append("suite", "com.example.SuiteTest")
                                        .appendEngine("junit-jupiter")
                                        .append("class", "com.example.ExampleTest" )
                                        .append("method", "test2()")
                        )
                )
                .build();

        LauncherFactory.create().execute(requestB);
    }
}
```

But when executed we see that this is the case.

```
By suite
Hello world 1
World was already greeted

By unique id
Hello world 1
Hello world 2
```

## Context

 - JUnit 5.10.1.

[...]

PR #3694, opened 2024-02-16
Apply suite configuration to tests selected by unique ID

mpkorstanje:
## Overview

Applies suite configuration to tests selected by unique ID.

Tests in a suite selected by id did not receive their configuration because this was only applied to the discovery request builder when a suite class was selected. By splitting the processing of configuration annotations and select and filter annotations, the configuration can now always be applied prior to discovery.

Fixes: #3693

---

---

### Definition of Done


junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
```diff
@@ -75,10 +75,14 @@ SuiteTestDescriptor addDiscoveryRequestFrom(Class<?> suiteClass) {
...
 	}
 
-	SuiteTestDescriptor addDiscoveryRequestFrom(UniqueId uniqueId) {
+	SuiteTestDescriptor addDiscoveryRequestFrom(Class<?> suiteClass, UniqueId uniqueId) {
 		Preconditions.condition(launcherDiscoveryResult == null,
 			"discovery request can not be modified after discovery");
-		discoveryRequestBuilder.selectors(DiscoverySelectors.selectUniqueId(uniqueId));
+		// @formatter:off
+		discoveryRequestBuilder
+				.suiteConfigurationParameters(suiteClass)
```
mpkorstanje, line 83: JUnit Vintage probably needs an equivalent fix.

documentation/src/docs/asciidoc/release-notes/release-notes-5.11.0-M1.adoc
```diff
@@ -18,6 +18,9 @@ repository on GitHub.
 
 * `ReflectionSupport.findFields(...)` now returns a distinct set of fields.
 
+* The `junit-platform-suite-engine` will now include configuration provided by
```
sbrannen, line 21:
```suggestion
* The `junit-platform-suite-engine` now includes configuration provided by
```

junit-platform-suite-commons/src/main/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilder.java
```diff
@@ -198,6 +196,16 @@ public SuiteLauncherDiscoveryRequestBuilder suite(Class<?> suiteClass) {
 		return this;
 	}
 
+	public SuiteLauncherDiscoveryRequestBuilder suiteConfigurationParameters(Class<?> suiteClass) {
```
sbrannen, line 199: I just realized we don't have any Javadoc for `SuiteLauncherDiscoveryRequestBuilder`. 😱 

So we should definitely address that, but rather in a different issue.

In any case, I'd like to see some Javadoc on this new method.

In addition, let's come up with a more descriptive name, perhaps something like one of the following.

- `configurationParametersForSuite(Class<?>)`
- `configurationParametersForSuiteClass(Class<?>)`
- `applyConfigurationParametersFromSuite(Class<?>)`

I know the last one is a bit long and doesn't start with `configurationParameter` like the existing methods; however, I _do_ think it aptly describes what it does (which is a bit special compared to the existing `configurationParameter()` and `configurationParameters()` methods).

Thoughts?

sbrannen (changes requested):
Thanks for the super fast PR! 👍 

I think it looks pretty solid, but I've requested a few minor enhancements.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
sbrannen: Good point.

However, since the `JUnitPlatform` is officially deprecated in favor of the `@Suite` support, I don't think we'll make any modifications to the `JUnitPlatform` runner.

junit-platform-suite-commons/src/main/java/org/junit/platform/suite/commons/SuiteLauncherDiscoveryRequestBuilder.java
mpkorstanje: I skipped the Javadoc because most classes marked as internal were lacking it. I'll add some for new methods.

Namingwise, I am considering deprecating the`suite` method and replace it with two methods:
 
 * `applyConfigurationParametersFromSuite`
 * `applySelectorsAndFiltersFromSuite`

And then also look at a way to avoid invoking `applyConfigurationParametersFromSuite` over and over again for evey unique id.

sbrannen: > I skipped the Javadoc because most classes marked as internal were lacking it. I'll add some for new methods.

Indeed, it is marked as internal even though it's `public`. I had overlooked that.

In any case, I think it's still worth documenting (even if it's "just for us").

I raised #3695 to address that.

Speaking of which, would you like to address #3695?

I'm asking because someone else has already volunteered to do that, and I want to avoid duplicate work.

sbrannen: > Namingwise, I am considering deprecating the existing `suite(Class<?>)` method and replacing it with two methods:
> 
>     * `applyConfigurationParametersFromSuite(Class<?>)`
> 
>     * `applySelectorsAndFiltersFromSuite(Class<?>)`

I like that. 👍 

I'm also wondering if we should add a precondition check which requires that the supplied `Class` is annotated with `@Suite`.

Or... perhaps that's over-engineering. 🤷 

> And then also look at a way to avoid invoking `applyConfigurationParametersFromSuite()` over and over again for every unique ID.

Yes, please do that.

> In hindsight these are two rather distinct operations. It just didn't occur to me when I wrote the suite engine.

No worries. It's a learning process, and APIs are meant to evolve. 😉

mpkorstanje: > I'm asking because someone else has already volunteered to do that, and I want to avoid duplicate work.

That's quick! This looks like a good issue for a first time contributor, so I'm happy to see someone else try.

mpkorstanje: > I'm also wondering if we should add a precondition check which requires that the supplied Class is annotated with @Suite.
>
> Or... perhaps that's over-engineering. 🤷

That would be a breaking change. The `@Suite` parameter is optional for the `JUnitPlatform` runner.

marcphilipp (approved):
Looks great! 👍

sbrannen:
> Looks great! 👍

I second that.

Thanks again for raising the issue **and** for the well-executed PR. 👍 

p.s. I polished a bit in 2a5c6b238e87384d27336d8353fb689b1b0275e1.

mpkorstanje:
Cheers! You're welcome. I think I've said before, it's nice to work on a codebase where everything is clear and well organized.

Issue #3630, opened 2024-01-04
Support scanning for classpath resources


mpkorstanje:
Currently `ReflectionUtils` supports resolving class based selectors such as `PackageSelector` and  `ClasspathRootSelector`.  These can be resolved using `findAllClassesInPackage` and `findAllClassesInClasspathRoot` respectively. 

Unfortunately there are no resource based equivalents to help resolve resources contained in a class path directory targeted by a `ClasspathResourceSelector` or `ClasspathRootSelector` e.g:

```java
List<Resource> findAllResourcesInPackage(String basePackageName, ResourceFilter resourceFilter)

List<Resource> findAllResourcesInClasspathRoot(URI root, ResourceFilter resourceFilter) 
```

As this functionality is backed by the `ClasspathScanner` I believe it would be prudent to add this functionality prior to making the `ClasspathScanner` part of the public API as proposed in https://github.com/junit-team/junit5/issues/3628. This would avoid breaking changes by adding more methods to an interface, or potential bug reports caused by a default implementation that always returns an empty collection of resources. 

And I would be happy to provide a PR that facilitates this.

## Deliverables

- [ ] Support for scanning class path resources through `ReflectionUtils` and `ClasspathScanner`.

sbrannen:
We never implemented that, because we never needed to scan for resources (other than `.class` files).

What's the concrete use case you're thinking of -- scanning for Cucumber files or similar?

mpkorstanje:
Yup. Cucumber uses `.feature` files which are typically located on the classpath. 

The benefits would be two fold
 * Eventually replace Cucumbers classpath scanning with the JUnit Platform
 * Benefit from implementations that support classpath scanning on Android.

marcphilipp:
**Team decision:** Let's add the two extra methods to `ClasspathScanner` and make them available via `ReflectionSupport` and `ReflectionUtils`.

As a follow-up task, we should add a `addResourceContainerSelectorResolver()` method to `EngineDiscoveryRequestResolver.Builder` analogous to `addClassContainerSelectorResolver()`.

marcphilipp:
@mpkorstanje A PR would be most welcome! 🙂

mpkorstanje:
I am still working on this, unfortunately due to circumstances it is going a bit slower than usual.

sbrannen:
Thanks for letting us know, @mpkorstanje.

This is currently assigned to the 5.11 M2 milestone, so you've got a bit more time.

Issue #3628, opened 2024-01-04
Allow third parties to provide a custom `ClasspathScanner` implementation


yuuki3655:
The Jupiter and Vintage engines use `org.junit.platform.commons.util.ClasspathScanner` to discover test classes in the classpath. The `ClasspathScanner` scans directories and looks for class files. It works fine for a standard JVM process; however, it needs to be customized for a specific environment such as Android.

A possible solution for this is to make the `ClasspathScanner` an interface and load the implementation via the `ServiceLoader` mechanism.

```java
public interface ClasspathScanner {
	static ClasspathScanner getInstance() {
		ServiceLoader<ClasspathScanner> serviceLoader = ServiceLoader.load(ClasspathScanner.class);
		for (ClasspathScanner scanner : serviceLoader) {
			return scanner;
		}
		return new DefaultClasspathScanner(ClassLoaderUtils::getDefaultClassLoader, ReflectionUtils::tryToLoadClass);
	}

	List<Class<?>> scanForClassesInPackage(String basePackageName, ClassFilter classFilter);

	List<Class<?>> scanForClassesInClasspathRoot(URI root, ClassFilter classFilter);
}
```

Then, we could provide the Android-specific implementation with [`androidx.test.internal.runner.ClassPathScanner`](https://github.com/android/android-test/blob/main/runner/android_junit_runner/java/androidx/test/internal/runner/ClassPathScanner.java), for example.

This change would make the Jupiter and Vintage engines work in Android.

marcphilipp:
Thanks for the proposal! It'd be awesome to get Jupiter and Vintage working out-of-the-box on Android. I'll discuss it with the team and will get back to you.

sbrannen:
SGTM.

Let's put that interface in the `org.junit.platform.commons.support` package.

marcphilipp:
**Team decision:** Let's follow the proposal but wait for #3630 to be done first.

yuuki3655:
Thank you very much for implementing this feature!

marcphilipp:
@yuuki3655 Did it help solve your use case?

PR #3705, opened 2024-02-25
Support scanning for classpath resources

mpkorstanje:
## Overview

Adds several new methods to `ReflectionSupport` to find or stream resource objects in the classpath root, package, and modules. 

A resource is represented as:

```java
public interface Resource {
	String getName();

	URI getUri();

	default InputStream getInputStream() throws IOException {
		return getUri().toURL().openStream();
	}
}
```

Fixes: #3630

---

---

### Definition of Done


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ResourceFileVisitor.java
```diff
@@ -0,0 +1,66 @@
...
+
+	private final Consumer<Path> classFileConsumer;
+
+	ResourceFileVisitor(Consumer<Path> classFileConsumer) {
+		this.classFileConsumer = classFileConsumer;
+	}
+
+	@Override
+	public FileVisitResult visitFile(Path file, BasicFileAttributes attributes) {
+		if (isNotClassFile(file)) {
```
mpkorstanje, line 42: I'm working with the assumption that when searching for test resources, we are not interested in classes and can exclude from discovery as soon as possible. I'm not sure if this assumption is correct.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/Resource.java
```diff
@@ -0,0 +1,61 @@
...
+
+	/**
+	 * Get the resource name.
+	 * <p>
+	 * The resource name is a {@code /}-separated path. The path is relative to
+	 * the classpath root in which the resource is located.
+	 *
+	 * @return the resource name; never {@code null}
+	 */
+	String getName();
```
mpkorstanje, line 43: I have intentionally expanded the original definition of `Resource` somewhat. 

A few reasons:
 * To make it possible apply the package filter in the `EngineDiscoveryRequestResolver`, `ReflectionUtils`. needs a `Predicate<String> resourceNameFilter`. The `Predicate<Resource> resourceFilter` should have access to the at least as much information as the `Predicate<String>`.
  
 * While the URI does identify a resource uniquely, it is rooted in the file system. This makes it a less convenient as a human readable identifier. For example in Cucumber, feature files locations are identified by their classpath resource name  or the path relative to the current working directory. For example
 
 ```
 Scenario: Addition                    # io/cucumber/examples/calculator/basic_arithmetic.feature:7    <<<<<<<<<<<<
  Given a calculator I just turned on # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.a_calculator_I_just_turned_on()
  When I add 4 and 5                  # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.adding(int,int)
  Then the result is 9                # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.the_result_is(double)

Scenario Outline: Many additions      # io/cucumber/examples/calculator/basic_arithmetic.feature:30     <<<<<<<<<<<<
  Given a calculator I just turned on # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.a_calculator_I_just_turned_on()
  Given the previous entries:         # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.thePreviousEntries(java.util.List<io.cucumber.examples.calculator.RpnCalculatorStepDefinitions$Entry>)
    | first | second | operation |
    | 1     | 1      | +         |
    | 2     | 1      | +         |
  When I press +                      # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.I_press(java.lang.String)
  And I add 2 and 3                   # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.adding(int,int)
  And I press +                       # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.I_press(java.lang.String)
  Then the result is 10               # io.cucumber.examples.calculator.RpnCalculatorStepDefinitions.the_result_is(double)
```

File: junit-platform-commons/src/main/java9/org/junit/platform/commons/util/ModuleUtils.java [hunk 3]
```diff
@@ -180,6 +221,7 @@ List<Class<?>> scan(ModuleReference reference) {
 							.filter(name -> !name.equals("module-info"))
 							.filter(classFilter::match)
 							.map(this::loadClassUnchecked)
+							// TODO: The ClasspathScanner invokes Predicate.test here
 							.filter(classFilter::match)
```
mpkorstanje, line 225: I think this may be a small bug.

platform-tests/src/test/java/org/junit/platform/commons/util/ClasspathScannerTests.java
```diff
@@ -166,14 +192,39 @@ private void scanForClassesInClasspathRootWithinJarFile(String resourceName) thr
...
+		try (var classLoader = new URLClassLoader(new URL[] { jarfile })) {
+			var classpathScanner = new ClasspathScanner(() -> classLoader, ReflectionUtils::tryToLoadClass);
+
+			var resources = classpathScanner.scanForResourcesInClasspathRoot(jarfile.toURI(), allResources);
+			assertThat(resources).extracting(Resource::getName) //
+					.containsExactlyInAnyOrder("org/junit/platform/jartest/notincluded/not-included.resource",
+						"org/junit/platform/jartest/included/included.resource",
+						"org/junit/platform/jartest/included/recursive/recursively-included.resource",
+						// TODO: This is interesting. Would we also scan classes in META-INF/versions?
+						"META-INF/MANIFEST.MF");
```
mpkorstanje, line 224: I'm not sure if this is surfacing a bug in the class-path scanner or if this is expected behavior.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathFilters.java
```diff
@@ -0,0 +1,45 @@
...
+	static final String CLASS_FILE_SUFFIX = ".class";
+	private static final String PACKAGE_INFO_FILE_NAME = "package-info" + CLASS_FILE_SUFFIX;
+	private static final String MODULE_INFO_FILE_NAME = "module-info" + CLASS_FILE_SUFFIX;
+
+	static Predicate<Path> classFiles() {
+		return file -> isNotPackageInfo(file) && isNotModuleInfo(file) && isClassFile(file);
+	}
+
+	static Predicate<Path> resourceFiles() {
+		return file -> !isClassFile(file);
```
mpkorstanje, line 30: I am explicitly excluding class files when scanning for resources. Tests that are located in class files will always be selected with a class selector, so it doesn't make sense to all consider them when scanning for resources.

sbrannen:
Thanks for the PR, @mpkorstanje! 👍 

We'll try to find some time to properly review it in the coming weeks (months?).

mpkorstanje:
Cheers! No problem.

platform-tests/src/test/java/org/junit/platform/commons/util/ClasspathScannerTests.java
marcphilipp: Yes, looks like we would also find classes in `META-INF/versions` but then fail to load them in `ClasspathScanner.processClassFileSafely`. Let's ignore this for this PR, though.

marcphilipp (commented):
> * To make it possible to use the package filter in the `EngineDiscoveryRequestResolver` in combination with the discovery to resources, `ReflectionUtils`. needs a `Predicate<String> resourceNameFilter`. The `Predicate<Resource> resourceFilter` should have access to the at least as much information as the `Predicate<String>`

I'm fine with adding `getName()` to `Resource`. However, the reason for having a `Predicate<String>` filter when scanning for classes is to avoid the overhead of loading all classes. I think the overhead for instantiating a `ClasspathResource` instance for every resource is negligible in comparison. Therefore, we could probably avoid having both `Predicate<String>` and `Predicate<Resource>` as parameters and only use the latter. Or did I miss anything?

mpkorstanje:
I don't think it is necessary for performance. 

But it also doesn't hurt and keeps the API similar to the class version. And in turn that would allow the a future `ResourceContainerSelectorResolver` implementation to be similar to the current `ClassContainerSelectorResolver` implementation.

I'll have a look and see if I run into anything unexpected with a potential `ResourceContainerSelectorResolver` implementation. If we don't need the `Predicate<String>`, then what about the `ResourceFilter`, is that stlil needed?

marcphilipp:
> I'll have a look and see if I run into anything unexpected with a potential `ResourceContainerSelectorResolver` implementation. If we don't need the `Predicate<String>`, then what about the `ResourceFilter`, is that stlil needed?

Thanks! Please see my comments on #3718.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/ReflectionSupport.java
```diff
@@ -137,6 +157,30 @@ public static List<Class<?>> findAllClassesInClasspathRoot(URI root, Predicate<C
...
+	 * @param resourceFilter the resource type filter; never {@code null}
+	 * @param resourceNameFilter the resource name filter; never {@code null}
+	 * @return an immutable list of all such resources found; never {@code null}
+	 * but potentially empty
+	 * @see #findAllResourcesInPackage(String, Predicate, Predicate)
+	 * @see #findAllResourcesInModule(String, Predicate, Predicate)
+	 */
+	@API(status = EXPERIMENTAL, since = "1.11")
+	public static List<Resource> findAllResourcesInClasspathRoot(URI root, Predicate<Resource> resourceFilter,
+			Predicate<String> resourceNameFilter) {
```
marcphilipp, line 179: :question: Why are the variants with two `Predicate` parameters back?

mpkorstanje: Because of https://github.com/junit-team/junit5/pull/3718#discussion_r1639617633. To ensure resources are found using the same semantics as `ClassLoader.getResource`, the best solution is to use `ClassLoader.getResource`. But the class loader is slow, so to prevent calling the classloader unnecessarily we end up with the same solution as the `findAllClasses*` variants. 

I've put an illustrative example of the expected behavior for shadowed resources in https://github.com/junit-team/junit5/pull/3705/commits/794793a9f396f60acd5f451e8305deecbd71b062#diff-67ad34a98151fe1159d3dbaa3b89cf23f691b6a8eac64231424ee48bd72c14d2R234 .

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java
```diff
@@ -166,11 +242,34 @@ private String determineFullyQualifiedClassName(Path baseDir, String basePackage
...
+	/**
+	 * The fully qualified resource name is a {@code /}-separated path.
+	 * <p>
+	 * The path is relative to the classpath root in which the resource is located.
+
+	 * @return the resource name; never {@code null}
+	 */
+	private String determineFullyQualifiedResourceName(Path baseDir, String basePackageName, Path resourceFile) {
+		// @formatter:off
+		return Stream.of(
```
mpkorstanje, line 254: While this very readable, 29% of the time in this is spend on joining strings.

image

mpkorstanje: Doing this with a string builder only removes ~30ms, not quite worth it.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -889,6 +892,27 @@ public static Try<Class<?>> tryToLoadClass(String name, ClassLoader classLoader)
...
+	 *
+	 * @param classpathResourceName the name of the resource to load; never {@code null} or blank
+	 * @param classLoader the {@code ClassLoader} to use; never {@code null}
+	 * @since 1.11
+	 */
+	@API(status = INTERNAL, since = "1.11")
+	public static Try<Resource> tryToLoadResource(String classpathResourceName, ClassLoader classLoader) {
+		Preconditions.notBlank(classpathResourceName, "Resource name must not be null or blank");
+		boolean startsWithSlash = classpathResourceName.startsWith("/");
+		URL resource = classLoader.getResource(startsWithSlash ? "/" + classpathResourceName : classpathResourceName);
```
marcphilipp, line 909: Why are we adding another slash if the resource name already starts with one?
```suggestion
		URL resource = classLoader.getResource(startsWithSlash ? classpathResourceName : "/" + classpathResourceName);
```

marcphilipp (changes requested):

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
mpkorstanje: Mmh. Looks like I dropped the tests when pulling this in from #3718.

mpkorstanje: Okay. Looks like I got the definition the wrong way around. The leading `/` should be removed prior to class loading because we're using `ClassLoader.getResource()` and not  `Class.getResource()`. Derp.

mpkorstanje:
I'll have a look at the coverage later. I'm out of time for now.

marcphilipp:
> I'll have a look at the coverage later. I'm out of time for now.

That's my bad (again). Rebasing should fix it.

marcphilipp (approved):

mpkorstanje:
Cheers! Might be good to document the switches for coverage and predictive test selection in the `CONTRIBUTING.md`.

marcphilipp:
> Cheers! Might be good to document the switches for coverage and predictive test selection in the `CONTRIBUTING.md`.

Done in c8beae0aaf30aacc9fa1f1673ff73afa3d24f836. Do you think that's clear enough?

mpkorstanje:
> Done in https://github.com/junit-team/junit5/commit/c8beae0aaf30aacc9fa1f1673ff73afa3d24f836. Do you think that's clear enough?

Yes. Though I now seem to have some problems generating the coverage report at all now. I'll have to look into this later.

>  We think restricting the resources to remove "duplicates" is too limiting. Instead, a filter that checks that Resource.getUri() equals ClassLoader.getResource could be used with a default implementation provided as a static method in ResourceFilter. WDYT?

Seems reasonable. I'll have a look at the plumbing.

mpkorstanje:
I've pushed the plumbing changes.

I'm mostly stuck on #3718. I think there are some subtleties that  could maybe propagate backwards into this PR. But I suppose we can keep evolving the API as needed. So we can consider this done.

marcphilipp:
Agreed. Let's finish this one and then focus on #3718. I can also take a more active role there if you don't have the time.

mpkorstanje:
That would be appreciated!

marcphilipp (approved):
@mpkorstanje Do you think it would make sense to add a `Predicate<Resource>` as a constant that allows filtering out duplicates?

```java
Predicate<Resource> DEFAULT_RESOURCE = resource -> {
	try {
		URL defaultResourceUrl = ClassLoaderUtils.getDefaultClassLoader().getResource(resource.getName());
		return defaultResourceUrl != null && resource.getUri().equals(defaultResourceUrl.toURI());
	}
	catch (URISyntaxException e) {
		throw new JUnitException("Failed to convert URL of resource to URI", e);
	}
};
```

mpkorstanje:
> @mpkorstanje Do you think it would make sense to add a Predicate<Resource> as a constant that allows filtering out duplicates?

Not in this iteration I'd think. You're about to go down [the same rabbit hole I spend a week in](https://github.com/junit-team/junit5/pull/3718#issuecomment-2224257764). :wink: 

A predicate would not work. It would not remove duplicates, but rather remove all shadowed resources from the list. To match the semantics of `ClassLoader.getResource` you'd have to create a new `Resource` by loading the given resource by name. Then de-duplicate those resources.

In https://github.com/junit-team/junit5/pull/3718 I wrote a few functions to just that, a good starting point would be the `ResourceContainerSelectorResolver` from there look at `ResourceFilter.loadClasspathResource()` and `ReflectionSup
port.tryToLoadResource`.

marcphilipp:
> Not in this iteration I'd think. You're about to go down [the same rabbit hole I spend a week in](https://github.com/junit-team/junit5/pull/3718#issuecomment-2224257764). 😉

Thanks for writing that up! I'll reply on the other PR.

> A predicate would not work. It would not remove duplicates, but rather remove all shadowed resources.

Isn't that what you were after? Wouldn't it only allow canonical resources to pass the predicate?

mpkorstanje:
> Isn't that what you were after? Wouldn't it only allow canonical resources to pass the predicate?

That depends. When discovering a class in classpath root a, we currently load a class from classpath root b, if b shadows a.

Do we want the same behaviour for resources?

marcphilipp:
I see what you mean now... :thinking: 

I'm pondering whether it makes sense to include this PR in 5.11 _without_ #3718 and, if so, when we would ship the latter. WDYT?

mpkorstanje:
I think it would make sense to ship this PR as it would unlock #3628. That was [the original motivation](https://github.com/junit-team/junit5/issues/3630#issuecomment-1878719592) for splitting these PRs.

PR #3718, opened 2024-03-08
Add resource container selector resolver

mpkorstanje:
## Overview

Resolve resource container selectors

As a follow up for #3630 and #3705 this adds a
`addResourceContainerSelectorResolver()` 
method to `EngineDiscoveryRequestResolver.Builder` analogous to
`addClassContainerSelectorResolver()`.

Points of note:

 * As classpath resources can be selected from packages, the package
   filter should also be applied. To make this possible the base path of
   a resource is rewritten to a package name prior to being filtered.

 * The `ClasspathResourceSelector` now has a `getClasspathResources`
   method. This method will lazily try to load the resources if  not
   already provided when discovering resources in a container.
   
 * `selectClasspathResource(Set<Resource>)` was added to short circuit the
    need to resolve resources twice. And to make it possible to use 
    this method as part of the public API,
    `ReflectionSupport.tryToLoadResources` was also added.

---

---

### Definition of Done


junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/ClasspathResourceSelector.java
```diff
@@ -65,6 +76,26 @@ public String getClasspathResourceName() {
...
+	 * method attempts to lazily load the {@link Resource} based on its name and
+	 * throws a {@link PreconditionViolationException} if the resource cannot
+	 * be loaded.
+	 */
+	@API(status = EXPERIMENTAL, since = "1.11")
+	public Resource getClasspathResource() {
+		if (this.classpathResource == null) {
+			// @formatter:off
+            Try<Resource> tryToLoadClass = ReflectionUtils.tryToLoadResource(this.classpathResourceName);
+            this.classpathResource = tryToLoadClass.getOrThrow(cause ->
```
marcphilipp, line 92:
```suggestion
            Try<Resource> tryToLoadResource = ReflectionUtils.tryToLoadResource(this.classpathResourceName);
            this.classpathResource = tryToLoadResource.getOrThrow(cause ->
```

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ResourceFilter.java
```diff
@@ -0,0 +1,51 @@
...
+ * <h2>DISCLAIMER</h2>
+ *
+ * <p>These utilities are intended solely for usage within the JUnit framework
+ * itself. <strong>Any usage by external parties is not supported.</strong>
+ * Use at your own risk!
+ *
+ * @since 1.11
+ */
+@API(status = INTERNAL, since = "1.11")
+public class ResourceFilter implements Predicate<Resource> {
```
marcphilipp, line 32: I think we don't need this class and can use `Predicate<Resource>` directly instead.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/ResourceContainerSelectorResolver.java
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	private static String packageName(String classpathResourceName) {
+		int lastIndexOf = classpathResourceName.lastIndexOf(CLASSPATH_RESOURCE_PATH_SEPARATOR);
+		if (lastIndexOf < 0) {
+			return DEFAULT_PACKAGE_NAME;
+		}
+		// classpath resource names do not start with /
+		String resourcePackagePath = classpathResourceName.substring(0, lastIndexOf);
+		return resourcePackagePath.replace(CLASSPATH_RESOURCE_PATH_SEPARATOR, PACKAGE_SEPARATOR_CHAR);
```
marcphilipp, line 59: I'm a bit worried about doing this for every candidate resource but apart from a limited-size cache I don't have a better idea.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java
```diff
@@ -154,6 +203,22 @@ private void processClassFileSafely(Path baseDir, String basePackageName, ClassF
 		}
 	}
 
+	private void processResourceFileSafely(Path baseDir, String basePackageName, Predicate<Resource> resourceFilter,
+			Path resourceFile, Consumer<Resource> resourceConsumer) {
+		try {
+			String fullyQualifiedResourceName = determineFullyQualifiedResourceName(baseDir, basePackageName,
+				resourceFile);
+			Resource resource = new ClasspathResource(fullyQualifiedResourceName, resourceFile.toUri());
```
mpkorstanje, line 211: This might be a problem.

1. Resources (and classes) with the same name can appear multiple times on the classpath.
2. So while walking the file tree, we can encounter the same resource twice
3. For classes the class loader ensures the same class is loaded in `processClassFileSafely` and eventually these duplicates are deduped by `.distinct()`. But in `processResourceFileSafely` we do not have this.
4. So we can end up with resource `<classpath-root-a>/com/example/resource.txt` and `<classpath-root-b>/com/example/resource.txt`.
5. This does not match the behaviour of `ClassLoader.getResource()`.

marcphilipp: Do you think it would it be sufficient to document that multiple resources with the same name (but different URLs) may be returned?

mpkorstanje: No. I don't think that would be sufficient. 

I would expect the behavior of the `ClasspathResourceSelector` for a given resource name to be consistent with `ClassLoader.getResource()`. And because the `ResourceContainerSelectorResolver` produces class path resource selectors with a resource through `ReflectionUtils.findAllResourcesIn*` the resources discovered this way should also be consistent with `ClassLoader.getResource()`.

In essence given this structure:
``` 
<classpath-root-a>
  - com/example/resource.txt 
 <classpath-root-b>
 - com/example/resource.txt.
```

Then using either
 * `DiscoverySelectors.selectPackage("com.example")`
 * `DiscoverySelectors.selectResource("com/example/resource.txt")`

Should result in a single test derived from the resource with `ClassLoader.getResource("com/example/resource.txt").toUri()`. As is, using the package selector would result in two tests.

mpkorstanje: I'll try to add some tests for this. Shadowing resources on the class path is not obvious.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/ResourceContainerSelectorResolver.java
mpkorstanje: We could short circuit this if the predicate accepts all packages.

I'll see if I can performance tests this.

mpkorstanje: Okay. Performance impact is pretty negligible.

Profiling `ResourceContainerSelectorResolver.resolve` with 160k resource files in a small directory structure, takes  ~4600ms, most of which is spend in either `ClassLoader.getResource` or `FileTreeWalker.next` The time spend in `packageName` is ~10ms. Adding caching increases that to ~20ms. Were you perhaps thinking of `String.replaceAll`? 

image

marcphilipp: Thanks for checking! Looks like my worry was unfounded.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ClasspathScanner.java
marcphilipp: I agree that in most cases, a single resource should be returned. I don't have a concrete use case for finding all of them, i.e. in different classpath roots. It came up recently in the context of potentially supporting multiple `junit-platform.properties` files.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/ResourceContainerSelectorResolver.java
```diff
@@ -0,0 +1,88 @@
...
+	@Override
+	public Resolution resolve(ClasspathRootSelector selector, Context context) {
+		return resourceSelectors(
+			findAllResourcesInClasspathRoot(selector.getClasspathRoot(), resourceFilter, resourceNameFilter));
+	}
+
+	@Override
+	public Resolution resolve(ModuleSelector selector, Context context) {
+		return resourceSelectors(
+			findAllResourcesInModule(selector.getModuleName(), resourceFilter, resourceNameFilter));
```
mpkorstanje, line 72: Currently, I'm assuming that all resources are class path resources, even if they're on the module path. Is that a valid assumption?

marcphilipp: I think that's reasonable. @sormuras WDYT?

sormuras: Resources are resources - running on Java 9+ all of them are in modules. Some of those modules are unnamend (loaded via the class-path), some are named (loaded via the module-path). So, maybe a bit late in this PR, but dropping "Classpath" from type and method names might be suited. For example `getClasspathResource` should be named `getResource` and `selectClasspathResource(Resource)` should be named `selectResource(Resource)`.

When loading resources in a named module the access is guarded by the module graph configuration supplied by the runtime. So yes, the implementation proposed should work as-is. A test confirming that would be nice to have, though.

marcphilipp: Since there's already `selectClasspathResource`, I wouldn't introduce a new name.

@mpkorstanje We'll discuss internally and will let you know before asking you to make any changes.

marcphilipp: **Team decision:** Stick with `DiscoverySelectors.selectClasspathResource` but document that the module path will also be checked.

mpkorstanje:
So the problems I'm running into is that:

1. Resources are only uniquely identified by their URI.
2. Resources are not uniquely identified by their name. Searching a container may return multiple results that need to be de-duplicated in some fashion. 
3. Resources are uniquely identified by their name and module name. So searching a module yields theoretically unique resources.
4. Resources discovered in a container should also be discoverable using the resource selector.

So then to identify resources we have either the option to

 a) Use URIs. This would allow multiple resources. But item 4 could only be satisfied using the URI selector. And as URIs include file system details, they're not very portable. Ideally the same project on different file systems should be able to use the same unique ids.
 
  b)  Use the resource name exclusively. This means resources need to be de-duplicated. The `ClassLoader.getResource` is a decent mechanism to find a "canonical" resource. This would also satisfy item 4. But it also seems rather restrictive when taking the module system into account.  
 
  c) Use the resource name + an optional module name. The module name would only be set if the resources was discovered through a module selector. This too would satisfy item 4. It is possible to provide a module name to `ClassLoader.getResource` so it is possible to use `DiscoverySelectors.slectResource("path/to.feature", "example-test-module")`
  
  De-duplicating would however be quite tricky. Discovering tests in a package and module could now results in discovery selectors that are not equal, but still target the same resource. And because the resource container selector resolver resolves each container selector in turn, there is no good way to de-duplicate. 
  
  So logically, only option b seems reasonable. Though quite restrictive.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/ResourceContainerSelectorResolver.java
mpkorstanje: Resolved 27c367113cfb96b7663c737120f8e9aad3b9fdad

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ResourceUtils.java
```diff
@@ -0,0 +1,111 @@
...
+
+	/**
+	 * Match resources against a package filter.
+	 *
+	 * <p>The {@code /} separated path of a resource is rewritten to a
+	 * {@code .} separated package names. The package filter is applied to that
+	 * package name.
+	 */
+	public static Predicate<Resource> packageName(Predicate<String> packageFilter) {
+		// TODO: Filter out invalid package names? META-INF and such?
```
mpkorstanje, line 93: Open TODO:

mpkorstanje:
> We briefly discussed the problem in the team. Do you have a concrete use case that would require solving the problem of having a machine-independent identifier for re-running a test based on a non-unique classpath resource? Otherwise, maybe we can move forward as-is and see if it actually becomes a problem?

Haven't got any concrete examples in mind right now. So that seems reasonable.

marcphilipp (commented):
Thanks for updating this PR! I'll discuss it with the team and will get back to you.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -893,6 +895,57 @@ public static Try<Class<?>> tryToLoadClass(String name, ClassLoader classLoader)
...
+		Preconditions.notNull(classLoader, "Class loader must not be null");
+		boolean startsWithSlash = classpathResourceName.startsWith("/");
+		String canonicalClasspathResourceName = (startsWithSlash ? classpathResourceName.substring(1)
+				: classpathResourceName);
+
+		return Try.call(() -> {
+			List<URL> resources = Collections.list(classLoader.getResources(canonicalClasspathResourceName));
+			if (resources.isEmpty()) {
+				throw new NullPointerException("classLoader.getResource returned no resources");
+			}
```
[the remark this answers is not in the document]
mpkorstanje: Like `tryToLoadClass` this method is used in context where something is expected to exist. So I reckon it is better for JUnit to complain when the class path is not as the user expects it to be. 

For example: without it it would be quite hard to notice that this Suite is not properly configured. 

```java
@SelectResources("com/example/features1")
@SelectResources("com/exampӏe/features2") // Spot the typo
```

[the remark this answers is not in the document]
marcphilipp: Being strict for `ClasspathResourceSelector` there makes sense! :+1: 

Nevertheless, I think we should return an empty set here as their might be other use cases for loading resources where their absence is acceptable.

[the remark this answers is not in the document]
mpkorstanje: Done. 2c49146.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -310,17 +314,51 @@ public static ClasspathResourceSelector selectClasspathResource(String classpath
...
+	 * @see #selectClasspathResource(String, FilePosition)
+	 * @see #selectClasspathResource(String)
+	 * @see ClasspathResourceSelector
+	 * @see ReflectionSupport#tryToGetResources(String)
+	 */
+	@API(status = EXPERIMENTAL, since = "1.12")
+	public static ClasspathResourceSelector selectClasspathResource(Set<Resource> classpathResources) {
+		Preconditions.notEmpty(classpathResources, "classpath resources must not be null or empty");
+		Preconditions.containsNoNullElements(classpathResources, "individual classpath resources must not be null");
+		List<String> resourceNames = classpathResources.stream().map(Resource::getName).distinct().collect(toList());
```
marcphilipp, line 356: Does `distinct()` work if `Resource.getName()` returns `null`?

mpkorstanje: Yes it does.

```
        List<String> list = Stream.of("a", "b", null, null).distinct().toList();
        System.out.println(list); // prints [a, b, null]
```        

The Javadoc on `.distinct` is a bit ambiguous when it suggests that the definition of `Object.equals(Object)` is used. That definition does not include null values.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/ClasspathResourceSelector.java
```diff
@@ -65,6 +83,32 @@ public String getClasspathResourceName() {
...
+	@API(status = EXPERIMENTAL, since = "1.12")
+	public Set<Resource> getClasspathResources() {
+		if (this.classpathResources == null) {
+			Try<Set<Resource>> tryToGetResource = ReflectionUtils.tryToGetResources(this.classpathResourceName);
+			Set<Resource> classpathResources = tryToGetResource.getOrThrow( //
+				cause -> new PreconditionViolationException( //
+					"Could not load resource(s) with name: " + this.classpathResourceName, cause));
+			if (classpathResources.isEmpty()) {
+				throw new PreconditionViolationException(
+					"Could not load resource(s) with name: " + this.classpathResourceName);
```
marcphilipp, line 105: I think it could be helpful to provide a different error message in this case. WDYT?

```suggestion
					"Could not find any resources with name: " + this.classpathResourceName);
```

documentation/src/docs/asciidoc/release-notes/release-notes-5.12.0-M1.adoc
```diff
@@ -27,6 +27,9 @@ JUnit repository on GitHub.
 [[release-notes-5.12.0-M1-junit-platform-new-features-and-improvements]]
 ==== New Features and Improvements
 
+* New `addResourceContainerSelectorResolver()` in `EngineDiscoveryRequestResolver.Builder` to
+support the discovery of class path resource based tests, analogous to the
+`addClassContainerSelectorResolver()`.
```
marcphilipp, line 32:
```suggestion
  support the discovery of class path resource based tests, analogous to the
  `addClassContainerSelectorResolver()`.
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/ResourceContainerSelectorResolver.java
```diff
@@ -0,0 +1,77 @@
...
+ * @since 1.12
+ */
+class ResourceContainerSelectorResolver implements SelectorResolver {
+	private final Predicate<Resource> packageFilter;
+	private final Predicate<Resource> resourceFilter;
+
+	ResourceContainerSelectorResolver(Predicate<Resource> resourceFilter, Predicate<String> resourcePackageFilter) {
+		this.packageFilter = packageName(resourcePackageFilter);
+		this.resourceFilter = resourceFilter;
+	}
```
marcphilipp, line 44: Why are there two `Predicate<Resource>` but only one is passed to the methods in `ReflectionSupport` and one is used to filter in `resourceSelectors`?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/ResourceContainerSelectorResolver.java
```diff
@@ -0,0 +1,77 @@
...
+
+	@Override
+	public Resolution resolve(ModuleSelector selector, Context context) {
+		return resourceSelectors(findAllResourcesInModule(selector.getModuleName(), packageFilter));
+	}
+
+	@Override
+	public Resolution resolve(PackageSelector selector, Context context) {
+		return resourceSelectors(findAllResourcesInPackage(selector.getPackageName(), packageFilter));
```
marcphilipp, line 58: Wouldn't it be possible avoid calling `findAllResourcesInPackage` if the package doesn't pass the package name filter? If so, we should also do that in `ClassContainerSelectorResolver`.

mpkorstanje: No, that would not be possible.  For example when selecting the root package with `selectPackage("")`, the root package would not match `includePackageNames("org.junit.platform")`, but its sub packages do match. This is tested in `shouldFilterPackages`.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/ResourceContainerSelectorResolver.java
mpkorstanje: Ah good one. We only need one predicate to pass to the `ReflectionSupport` methods  with no further filtering.

Done in [1d13082](https://github.com/junit-team/junit5/pull/3718/commits/1d13082aa6eac12388feb7284d0ca2db7afb885c).

marcphilipp (approved):

marcphilipp:
@mpkorstanje Thanks a lot! Do you have time to try out adopting this in Cucumber from a snapshot?

mpkorstanje:
Cheers! I've already had a work in progress. But it will need some updates. Shouldn't be too difficult.

https://github.com/cucumber/cucumber-jvm/pull/2835

Issue #2154, opened 2020-01-16
Method `Assertions.assertThrows(expectedType, executable, message)` doesn't check the contents of the exception message.


borisilic:
**JUnit Version:** 5.4.0

**For this unit test**

```
@Test
public void verifyErrorMessageWhenBranchCannotBeFound() {
        given(restTemplate.getForObject(any(), any(), (Object) any())).willThrow(NotFoundException.class);

        assertThrows(NotFoundException.class, () -> testInstance.retrieveBranchDetails("3001"),
                "I don't think this error messaging thing checker works");
}
```

**Testing this code**
``` 
public BranchDetailsResponse retrieveBranchDetails(String branchNumber) {
        try {
            return restTemplate.getForObject(url, BranchDetailsResponse.class, branchNumber);
        } catch (HttpStatusCodeException e) {
            throw new NotFoundException("");
        }
   }
```

I expect that the test will fail as the error message passed to the `assertThrows` method does not match the error message that is returned from the thrown exception. 

I believe the offending piece of code is here in the class `AssertThrows.java`

```
@SuppressWarnings("unchecked")
private static <T extends Throwable> T assertThrows(Class<T> expectedType, Executable executable,
	Object messageOrSupplier) {
	try {
		executable.execute();
	}
	catch (Throwable actualException) {
                // This part here just returns so long as the two exceptions match without checking the messages
		if (expectedType.isInstance(actualException)) {
			return (T) actualException;
		}
		else {
			BlacklistedExceptions.rethrowIfBlacklisted(actualException);
			String message = buildPrefix(nullSafeGet(messageOrSupplier))
					+ format(expectedType, actualException.getClass(), "Unexpected exception type thrown");
			throw new AssertionFailedError(message, actualException);
		}
	}

	String message = buildPrefix(nullSafeGet(messageOrSupplier))
			+ String.format("Expected %s to be thrown, but nothing was thrown.", getCanonicalName(expectedType));
	throw new AssertionFailedError(message);
}
```

The if statement which is bolded will return the actualException if it matches the expectedException but nowhere does it do any message error checking.

sormuras:
The (optional) failure message you pass to any `Assertions.assertXyz(..., message)` method is never checked. For example, the [assertThrows](https://junit.org/junit5/docs/current/api/org/junit/jupiter/api/Assertions.html#assertThrows(java.lang.Class,org.junit.jupiter.api.function.Executable,java.lang.String)) API documentation reads:

[...]

PR #3727, opened 2024-03-14
Make it clearer that `assertThrows` message is not the 'expected message'

Marcono1234:
## Overview

Tries to adjust the documentation ~~and name of parameters~~ of `assertThrows` and `assertThrowsExactly` to make it clearer that the provided 'message' is not the 'expected message' for the thrown exception.
This seems to be a somewhat common misunderstanding, see #2154.

In case you are familiar with CodeQL you can also try running [this custom query](https://github.com/Marcono1234/codeql-java-queries/blob/a574ff4e97cc383686eef6c8dfa71ea8c27d1068/codeql-custom-queries-java/queries/unit-tests/assertThrows-expected-message.ql) on your project or on GitHub projects. For example there seems to be incorrect `assertThrows` usage in apache/kafka ([example](https://github.com/apache/kafka/blob/5dd382ccbd1fada55cc44a5223dcae56e0e8811d/connect/transforms/src/test/java/org/apache/kafka/connect/transforms/MaskFieldTest.java#L252C92-L252C118)), apache/commons-lang ([example](https://github.com/apache/commons-lang/blob/5a9cc5f20be095efb9eb50cb456d00ce706daffa/src/test/java/org/apache/commons/lang3/StringUtilsTest.java#L3109C87-L3109C115)) and netty/netty ([example](https://github.com/netty/netty/blob/a09cc1620c9f8ac0cb17851b85a051459ee02964/codec/src/test/java/io/netty/handler/codec/compression/Lz4FrameDecoderTest.java#L62C12-L62C41)).

Any suggestions for better approaches, and in general any feedback regarding this is welcome!

---

---

### Definition of Done


junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
Marcono1234, line 1: Not completely sure how useful the changes here for the Kotlin API are.

And potentially this parameter renaming would be a source-incompatible change since Kotlin supports [named arguments](https://kotlinlang.org/docs/functions.html#named-arguments).

sbrannen: Good point.

Since it's technically a breaking change for existing Kotlin code, I think we probably should not change the parameter names.

Marcono1234:
> However, I'm not convinced that we need to change the names of the parameters.
> 
> For example, `failureMessage` can also easily be misinterpreted to mean the failure message for the thrown exception.

I think renaming the parameter does add some value (but only for these methods, not in general for all `Assertions` methods), for example when the IDE of a user hasn't downloaded the Javadoc / sources yet or when the user at a glance just looks at the parameter name. Especially when users are unaware that this 'message' parameter is common for all assertion methods, it is easy to confuse since 'message' is also the naming used by `Throwable` (e.g. `Throwable.getMessage`).

I chose "failure message" because that term seems to be used by the documentation for these and other assertion methods already, but I am also open for any other name.


> Furthermore, if we're going to update the Javadoc for the `message` variants, we'll need to update the Javadoc for the `messageSupplier` variants as well.

Do you mean for `assertThrows` and `assertThrowsExactly` with `Supplier` parameter? I think I updated those, but maybe I overlooked something, or I am misunderstanding you.

marcphilipp (changes requested):
I also think we should address this in the Javadoc rather than renaming the parameter which (for Kotlin) would be a breaking change.

Marcono1234:
> which (for Kotlin) would be a breaking change

Do you mean this only for `Assertions.kt` or also for `Assertions.java`?

I can revert the parameter name change for both classes, but I am afraid just changing the documentation might not be effective enough.

marcphilipp:
I think the names should be consistent across Kotlin and Java. Therefore, I think we should keep the old names in both places. When writing tests, you should see each assertion fail in the right way at least once. If you follow that guideline, it will be hard to confuse the two messages. Therefore, I think improving the documentation is the right move here.

sbrannen:
> Do you mean for `assertThrows` and `assertThrowsExactly` with `Supplier` parameter? I think I updated those, but maybe I overlooked something, or I am misunderstanding you.

@Marcono1234, yes. I apologize for having overlooked that somehow.

> When writing tests, you should see each assertion fail in the right way at least once. If you follow that guideline, it will be hard to confuse the two messages. Therefore, I think improving the documentation is the right move here.

Exactly!

As I stated previously, `failureMessage` can also be misinterpreted to mean the failure message for the thrown exception. In addition, changing the names of parameters is a breaking change for Kotlin and would result in inconsistency in the APIs and documentation.

Thus, if we do anything at all about this, it has to be limited to improving the documentation.

Marcono1234:
Have renamed the parameters.

sbrannen (changes requested):
Thanks for reverting the changes to the parameter names.

After further consideration, I decided that we should provide additional advice regarding the expected exception message.

See latest review comment.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -3070,7 +3070,8 @@ public static <T extends Throwable> T assertThrowsExactly(Class<T> expectedType,
 	 * <p>If you do not want to perform additional checks on the exception instance,
 	 * ignore the return value.
 	 *
-	 * <p>Fails with the supplied failure {@code message}.
+	 * <p>Fails with the supplied failure {@code message}. That message is <em>not</em>
+	 * the expected message for the thrown exception.
```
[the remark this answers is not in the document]
Marcono1234: Ok, I have adjusted the documentation with this now. For the methods with `Supplier` I slightly modified the first sentence since there is no `message` parameter.

marcphilipp (approved):

sbrannen (approved):

sbrannen:
This has been merged into `main` in 5abe0e7688cd1ddf6bac69ebe0a1b46c13afaec2 and revised in d67e5c021a2c120bafc9c7e50e1731cce8d027a2.

Thanks

PR #3737, opened 2024-03-18
Introduce parseable DiscoverySelector representations

leonard84:
## Overview

Before this PR, it was not possible to use the more advanced `DiscoverySelectors` without writing code. Now, we have a generic way to specify all supported `DiscoverySelectors` as strings and parse them. 
This will allow advanced usage scenarios. For example, Gradle can add a `Test` task parameter to pass a custom selector directly to the platform. This would allow using the `IterationSelector` when delegating to Gradle in IDEA.

---

---

### Definition of Done


junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/NestedMethodSelector.java
```diff
@@ -239,4 +243,41 @@ public String toString() {
 				.toString();
 	}
 
+	@Override
+	public Optional<String> toSelectorString() {
+		return nestedClassSelector.toSelectorString().map(parent -> {
+			StringBuilder sb = new StringBuilder(parent)
+					// Not totally happy with how we have to change the prefix here.
+					// Alternativly, we could duplicate the logic of the NestedClassSelector
```
mpkorstanje, line 251: Just an idea, you could consider making the scheme recursive, i.e:

`method:nested-class:com.example.Example#!/test(String,String)`.

It works for Jar URIs.

marcphilipp: I've since reworked this part which resolved the issue.

leonard84:
@marcphilipp, it looks good, what is missing? Or, are you waiting for me to switch it from draft to rfr?

marcphilipp:
@leonard84 Yes, it does look good and I will finish it for 5.11 M2.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -937,4 +942,44 @@ public static IterationSelector selectIteration(DiscoverySelector parentSelector
...
+	 * {@link DiscoverySelectorIdentifier}.
+	 *
+	 * @param identifier the string representation of a {@code DiscoverySelectorIdentifier};
+	 * never {@code null} or blank
+	 * @since 1.11
+	 * @see DiscoverySelectorIdentifierParser
+	 */
+	@API(status = EXPERIMENTAL, since = "1.11")
+	public static Optional<? extends DiscoverySelector> parse(String identifier) {
+		return new DiscoverySelectorIdentifierParsers().parse(identifier);
```
marcphilipp, line 956: Creating a new instance of `DiscoverySelectorIdentifierParsers` here which leads to a new `ServiceLoader` lookup and instantiation does not feel right here. I don't think we should make it `static`, though. Thoughts?

leonard84: Is this really a problem? I agree that making it `static` feels dirty.
 
Maybe we should add a note to prefer using either `parseAll` when processing more than one, and/or consider making `DiscoverySelectorIdentifierParsers` public so that its `parse` method can be used repeatedly, e.g., when working with streams.

marcphilipp: I've changed it to be lazily load them only once via a singleton in 9967df695015b9b81a85127f7ce2c854965bccd2

junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java
```diff
@@ -39,7 +40,9 @@ class TestDiscoveryOptionsMixin {
 	@ArgGroup(validate = false, order = 2, heading = "%n@|bold SELECTORS|@%n%n")
 	SelectorOptions selectorOptions;
 
-	@ArgGroup(validate = false, order = 3, heading = "%n@|bold FILTERS|@%n%n")
+	@ArgGroup(validate = false, order = 3, heading = "%n  For more information on selectors including syntax examples, see"
+			+ "%n  @|underline https://junit.org/junit5/docs/current/user-guide/#running-tests-discovery-selectors|@"
```
marcphilipp, line 44:
```suggestion
			+ "%n  @|underline https://junit.org/junit5/docs/${junit.docs.version}/user-guide/#running-tests-discovery-selectors|@"
```

I tried using a dynamic version here but hit https://github.com/remkop/picocli/issues/2281

marcphilipp (approved):

Issue #3748, opened 2024-03-31
@ParameterizedTest is not recognized by the method selector using Junit Platform Console Standalone


rcasia:
Junit Standalone cannot discover a test method that has `@ParameterizedTest` annotation when runned using the method selector. When I run other test methods it works as usual. Also it discovers the parameterized tests when using the class selector. 

## Steps to reproduce

1. Wrote this test:

```java
  @ParameterizedTest
  @CsvSource({ "1", "2", "3" })
  void testNumbers(int i) {
    assertEquals(i, 3);
  }
``` 
2. Run this command
```shell
./mvnw -q dependency:build-classpath -Dmdep.outputFile=/tmp/classpath.txt

javac -d /tmp/outputDir -cp $(cat /tmp/classpath.txt) src/main/**/*.java

javac -d /tmp/outputDir -cp $(cat /tmp/classpath.txt):/tmp/outputDir src/test/**/*.java

java -jar ~/junit-platform-console-standalone-1.10.2.jar execute -cp $(cat /tmp/classpath.txt):/tmp/outputDir -m=com.rcasia.demo.DemoApplicationTests#testNumbers --fail-if-no-tests --reports-dir=/tmp/reportDir
``` 

### Expected
To discover the parameterized test method.

### Actual

Does not discover the parameterized test method.

<details><summary>Logs :spiral_notepad: </summary>
<p>

org.junit.platform.commons.JUnitException: TestEngine with ID 'junit-jupiter' failed to discover tests
        at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverEngineRoot(EngineDiscoveryOrchestrator.java:160)
        at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discoverSafely(EngineDiscoveryOrchestrator.java:132)
        at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:107)
        at org.junit.platform.launcher.core.EngineDiscoveryOrchestrator.discover(EngineDiscoveryOrchestrator.java:78)
        at org.junit.platform.launcher.core.DefaultLauncher.discover(DefaultLauncher.java:99)
        at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:85)
        at org.junit.platform.launcher.core.DelegatingLauncher.execute(DelegatingLauncher.java:47)
[...]

PR #3752, opened 2024-04-01
Clarify test method discovery and supported formats in User Guide (CLI)

rcasia:
Issue: #3748

## Overview

This PR adds clarity to the usage of test methods with the --select-method argument for CLI users.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java
@@ -112,7 +112,7 @@ class TestDiscoveryOptionsMixin {
 		private final List<ClassSelector> selectedClasses2 = new ArrayList<>();
 
 		@Option(names = { "-m",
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Okay, I'm not part of the JUnit Team, so I can't say anything with authority. Usually they'll ask for contributions, or if it is small stuff, to avoid the back and forth they'll do it directly.

That said, I think it would help if:
 * The formatting in the issue was fixed.
 * There was an actual changelog item.

sbrannen:
If we are going to be more explicit about the syntax for method selectors, we should also be more explicit about other selectors: packages, classes, etc.

For example, those all have to be "fully-qualified".

sbrannen:
> That said, I think it would help if:
> 
>     * The formatting in the issue was fixed.
> 
>     * There was an actual changelog item.

I agree with both of those points.

The formatting should align with the existing formatting (line breaks, etc.) in the same file.

This would warrant a minimal entry in the release notes.

marcphilipp:
I'll address this as part of #3737

Issue #3730, opened 2024-03-15
Document test exception handling


Marcono1234:
It seems there is currently not one single place which describes how exceptions in tests are handled. This can lead to confusion regarding behavior for unexpected exceptions, especially when they are checked exceptions, see for example #3508.
This information is currently split across the documentation, such as the [`assertDoesNotThrow` method documentation](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable)), which says:
> Although any exception thrown from a test method will cause the test to fail

JUnit 4 has a separate [FAQ entry for this](https://junit.org/junit4/faq.html#atests_8)[^1].

It might therefore be good to have one place (probably a new User Guide section?) where the exception handling is documented:
- Uncaught exceptions cause a test error by default
  - for checked exceptions a `throws` clause has to be added to the test method
  - usage of `throws` clause has no effect (e.g. it does not indicate an "expected exception")
  - (maybe) recommend using `throws Exception` instead of `throws CheckedExceptionA, CheckedExceptionB, CheckedExceptionC` since the latter introduces a lot of unnecessary clutter[^2]
  - can use `assertDoesNotThrow` to explicitly indicate unexpected exceptions, but it is not required
  - (maybe) mention special exception types which cause test to be skipped or aborted
- Expected exceptions can be tested for using `assertThrows` / `assertThrowsExactly`
- Difference between test _failure_ (i.e. assertion failure) and test _error_ (i.e. unexpected exception)[^3]

But this section should probably be targeted at regular users, so it should not go into detail regarding how extensions can handle exceptions. That seems to already be covered in the User Guide section about extensions.

What do you think?

## Deliverables

- [ ] A section in the User Guide about test exception handling

[^1]: Though that entry is slightly confusing because it recommends using `throws ...` for `IndexOutOfBoundsException`, which is an unchecked exception.
[^2]: It seems not everyone [agrees that using `throws Exception` there is fine](https://stackoverflow.com/a/2496602). But personally I am not sure if developers really deliberately list all checked exceptions, or if most often they just use the IDE action for adding a `throws` clause.
[^3]: I am not sure though how strong this separation still exists for JUnit 5. I thought I head read about it in the JUnit 5 documentation before, but I cannot find it right now. And in some cases there seems to be no such differentiation, e.g. #3436.

marcphilipp:
**Team decision:**

[...]

PR #3780, opened 2024-04-16
Document test exception handling

driptaroop:
## Overview

Documents how exceptions in tests are handled. This PR contains documentation that describes,

1. Uncaught exceptions cause test failures by default.
2. How assertions are implemented using exceptions.
3. JUnit 5 does not differentiate between failed assertions and other exceptions but IDEs and other tools may choose to do so by checking for instances of AssertionError.
4. Adds assertDoesNotThrow example for Java.

Solves Issue: #3730
---

---

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -367,6 +367,40 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
 `AssumptionViolatedException` to signal that a test should be aborted instead of marked
 as a failure.
 
+[[writing-tests-exceptions]]
+=== Other Exceptions
+
+JUnit Jupiter provides robust support for handling test exceptions. This includes the built-in mechanisms for managing test failures due to exceptions, the role of exceptions in implementing assertions, and how to specifically assert non-throwing conditions in code.
+
+[[writing-tests-uncaught-exceptions]]
+==== Uncaught Exceptions and Test Failures
+
+In JUnit Jupiter, if an exception is thrown and not caught within the test method, the framework will mark the test as failed.
+
+[source,java,indent=0]
+----
+include::{testDir}/example/exception/UncaughtExceptionHandlingDemoTestCase.java[tags=user_guide]
+----
+
+In the above example, the `testThatFailsDueToUncaughtException` method throws a `RuntimeException`. Since the exception is not caught within the test method, JUnit Jupiter will automatically fail the test.
+
+[[writing-tests-assertions-using-exceptions]]
+==== Assertions Using Exceptions
+
+Assertions in JUnit Jupiter are implemented using exceptions. The framework provides a set of assertion methods in the `org.junit.jupiter.api.Assertions` class, which throw `AssertionError` when an assertion fails. This mechanism is a core aspect of how JUnit handles assertion failures as exceptions.
+
+NOTE: JUnit Jupiter itself does not differentiate between failed assertions (`AssertionError`) and other types of exceptions. All uncaught exceptions lead to a test failure. However, Integrated Development Environments (IDEs) and other tools may distinguish between these two types of failures by checking whether the thrown exception is an instance of AssertionError.
+
+[[writing-tests-assertions-using-assertDoesNotThrow]]
+==== Asserting Non-Exception Conditions with `assertDoesNotThrow`
+
+Although any exception thrown from a test method will cause the test to fail, there are certain use cases where it can be beneficial to explicitly assert that an exception is not thrown for a given code block within a test method. The `assertDoesNotThrow` assertion can be used when you want to verify that a particular piece of code does not throw any exceptions.
+
+[source,java,indent=0]
+----
+include::{testDir}/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java[tags=user_guide]
+----
+
 [[writing-tests-disabling]]
 === Disabling Tests
 
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -370,36 +370,96 @@ as a failure.
 [[writing-tests-exceptions]]
 === Exception Handling
 
-JUnit Jupiter provides robust support for handling test exceptions. This includes the built-in mechanisms for managing test failures due to exceptions, the role of exceptions in implementing assertions, and how to specifically assert non-throwing conditions in code.
+JUnit Jupiter provides robust support for handling test exceptions. This includes the
+built-in mechanisms for managing test failures due to exceptions, the role of exceptions
+in implementing assertions, and how to specifically assert non-throwing conditions in code.
 
-[[writing-tests-uncaught-exceptions]]
+[[writing-tests-exceptions-uncaught]]
 ==== Uncaught Exceptions and Test Failures
 
-In JUnit Jupiter, if an exception is thrown and not caught within the test method, the framework will mark the test as failed.
+In JUnit Jupiter, if an exception is thrown and not caught within the test method, the
+framework will mark the test as failed.
 
 [source,java,indent=0]
 ----
 include::{testDir}/example/exception/UncaughtExceptionHandlingDemoTestCase.java[tags=user_guide]
 ----
 
-In the above example, the `testThatFailsDueToUncaughtException` method throws a `RuntimeException`. Since the exception is not caught within the test method, JUnit Jupiter will automatically fail the test.
+In the above example, the `testThatFailsDueToUncaughtException` method throws a
+`RuntimeException`. Since the exception is not caught within the test method, JUnit
+Jupiter will automatically fail the test.
 
-[[writing-tests-assertions-using-exceptions]]
+NOTE: It's important to clarify that specifying a `throws` clause in the test method
+has no effect on the outcome of the test. JUnit Jupiter does not interpret it as
+an expectation or assertion about what exceptions the test method should throw. A test
+fails only if an exception is thrown unexpectedly or if an assertion fails.
+
+[[writing-tests-exceptions-assertions]]
 ==== Assertions Using Exceptions
 
-Assertions in JUnit Jupiter are implemented using exceptions. The framework provides a set of assertion methods in the `org.junit.jupiter.api.Assertions` class, which throw `AssertionError` when an assertion fails. This mechanism is a core aspect of how JUnit handles assertion failures as exceptions.
+Assertions in JUnit Jupiter are implemented using exceptions. The framework provides
+a set of assertion methods in the `org.junit.jupiter.api.Assertions` class, which throw
+`AssertionError` when an assertion fails. This mechanism is a core aspect of how JUnit
+handles assertion failures as exceptions.
 
-NOTE: JUnit Jupiter itself does not differentiate between failed assertions (`AssertionError`) and other types of exceptions. All uncaught exceptions lead to a test failure. However, Integrated Development Environments (IDEs) and other tools may distinguish between these two types of failures by checking whether the thrown exception is an instance of AssertionError.
+NOTE: This failed assertion behaviour (i.e. throwing `AssertionError`) may be different
+for third party assertion libraries. See also: <<writing-tests-assertions-third-party>>.
 
-[[writing-tests-assertions-using-assertDoesNotThrow]]
-==== Asserting Non-Exception Conditions with `assertDoesNotThrow`
+NOTE: JUnit Jupiter itself does not differentiate between failed assertions
+(`AssertionError`) and other types of exceptions. All uncaught exceptions lead to a test
+failure. However, Integrated Development Environments (IDEs) and other tools may
+distinguish between these two types of failures by checking whether the thrown exception
+is an instance of `AssertionError`.
 
-Although any exception thrown from a test method will cause the test to fail, there are certain use cases where it can be beneficial to explicitly assert that an exception is not thrown for a given code block within a test method. The `assertDoesNotThrow` assertion can be used when you want to verify that a particular piece of code does not throw any exceptions.
+[[writing-tests-exceptions-expected]]
+==== Asserting Expected Exception Conditions
+
+JUnit Jupiter offers specialized assertions for testing that specific exceptions are
+thrown under expected conditions. The `assertThrows` and `assertThrowsExactly` assertions
+are critical tools for validating that your code responds correctly to error conditions
+by throwing the appropriate exceptions.
+
+[[writing-tests-exceptions-expected-assertThrows]]
+===== Using `assertThrows`
+
+The `assertThrows` method is used to verify that a particular type of exception is thrown
+during the execution of a provided executable block. It not only checks for the type of
+the thrown exception but also its subclasses, making it suitable for more generalized
+exception handling tests.
+
+[source,java,indent=0]
+----
+include::{testDir}/example/exception/ExceptionAssertionDemoTestCase.java[tags=user_guide]
+----
+
+[[writing-tests-exceptions-expected-assertThrowsExactly]]
+===== Using `assertThrowsExactly`
+
+`assertThrowsExactly` is used when you need to assert that an exception is not only
+thrown but is exactly of a specific type, not allowing for subclasses of the exception.
+This is useful when precise exception handling behavior needs to be validated.
+
+[source,java,indent=0]
+----
+include::{testDir}/example/exception/ExceptionAssertionExactDemoTestCase.java[tags=user_guide]
+----
+
+[[writing-tests-exceptions-using-assertDoesNotThrow]]
+==== Using `assertDoesNotThrow` to Verify Absence of Exceptions
+
+Although any exception thrown from a test method will cause the test to fail, there are
+certain use cases where it can be beneficial to explicitly assert that an exception is
+not thrown for a given code block within a test method. The `assertDoesNotThrow`
+assertion can be used when you want to verify that a particular piece of code does not
+throw any exceptions.
 
 [source,java,indent=0]
 ----
 include::{testDir}/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java[tags=user_guide]
 ----
+
+NOTE: Other third party assertion libraries also provide similar support. For example,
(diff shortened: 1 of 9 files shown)


documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -367,6 +367,40 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
...
+
+Assertions in JUnit Jupiter are implemented using exceptions. The framework provides a set of assertion methods in the `org.junit.jupiter.api.Assertions` class, which throw `AssertionError` when an assertion fails. This mechanism is a core aspect of how JUnit handles assertion failures as exceptions.
+
+NOTE: JUnit Jupiter itself does not differentiate between failed assertions (`AssertionError`) and other types of exceptions. All uncaught exceptions lead to a test failure. However, Integrated Development Environments (IDEs) and other tools may distinguish between these two types of failures by checking whether the thrown exception is an instance of AssertionError.
+
+[[writing-tests-assertions-using-assertDoesNotThrow]]
+==== Asserting Non-Exception Conditions with `assertDoesNotThrow`
+
+Although any exception thrown from a test method will cause the test to fail, there are certain use cases where it can be beneficial to explicitly assert that an exception is not thrown for a given code block within a test method. The `assertDoesNotThrow` assertion can be used when you want to verify that a particular piece of code does not throw any exceptions.
```
sbrannen, line 397: Add note that other assertion libraries provide similar support, such as AssertJ's `assertThatNoException().isThrownBy(() -> ...)`.

documentation/src/test/java/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.jupiter.api.Test;
+
+// @formatter:off
+// tag::user_guide[]
+
+class AssertDoesNotThrowExceptionDemoTestCase {
+
+	@Test
+	void assertDoesNotThrowExceptionTest() {
+		Assertions.assertThrows(RuntimeException.class, () -> {
```
sbrannen, line 23: We use static imports for `Assertions`.

documentation/src/test/java/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java
```diff
@@ -0,0 +1,39 @@
...
+			failsDueToException();
+			Assertions.assertDoesNotThrow(() -> {
+				shouldNotThrowException();
+			});
```
sbrannen, line 27: Please move this to a separate example.

sbrannen (changes requested):
Thanks for the PR.

I've requested a few changes.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -367,6 +367,40 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
...
+
+[source,java,indent=0]
+----
+include::{testDir}/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java[tags=user_guide]
+----
+
```
Marcono1234, line 403: Question to the maintainers: Would it make sense to also mention `assertThrows` in a small section?

I know that `assertThrows` is used in the existing code samples, for example in the "Assertions" section. But users might first have a look at this new section and are then confused because there is no information about handling expected exceptions, only about unexpected exceptions. This might be especially a problem when this section is named "Exception Handling" (as proposed in https://github.com/junit-team/junit5/pull/3780#discussion_r1567605519).

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -367,6 +367,40 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
...
+==== Uncaught Exceptions and Test Failures
+
+In JUnit Jupiter, if an exception is thrown and not caught within the test method, the framework will mark the test as failed.
+
+[source,java,indent=0]
+----
+include::{testDir}/example/exception/UncaughtExceptionHandlingDemoTestCase.java[tags=user_guide]
+----
+
+In the above example, the `testThatFailsDueToUncaughtException` method throws a `RuntimeException`. Since the exception is not caught within the test method, JUnit Jupiter will automatically fail the test.
```
Marcono1234, line 385: Question to the maintainers: Should this also explicitly mention that a `throws` clause has no effect, to avoid any misunderstanding?
For example
> [...] It does not matter whether or not the test method has a `throws` clause which lists the exception (or a superclass of it).

Otherwise users might assume `throws SomeException` means that it is expected that the test fails with `SomeException`.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sbrannen: Yes, I think we should cover `assertThrows` (and mention `assertThrowsExactly`) in this new section.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sbrannen: Standard Java semantics apply here, but that's not always readily apparent to someone new to testing.

So, yes, it would be beneficial for some users to point that out.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -367,6 +367,40 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
 `AssumptionViolatedException` to signal that a test should be aborted instead of marked
 as a failure.
 
+[[writing-tests-exceptions]]
+=== Other Exceptions
```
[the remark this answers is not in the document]
driptaroop: I agree. changed it.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -367,6 +367,40 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
...
+----
+include::{testDir}/example/exception/UncaughtExceptionHandlingDemoTestCase.java[tags=user_guide]
+----
+
+In the above example, the `testThatFailsDueToUncaughtException` method throws a `RuntimeException`. Since the exception is not caught within the test method, JUnit Jupiter will automatically fail the test.
+
+[[writing-tests-assertions-using-exceptions]]
+==== Assertions Using Exceptions
+
+Assertions in JUnit Jupiter are implemented using exceptions. The framework provides a set of assertion methods in the `org.junit.jupiter.api.Assertions` class, which throw `AssertionError` when an assertion fails. This mechanism is a core aspect of how JUnit handles assertion failures as exceptions.
```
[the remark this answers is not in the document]
driptaroop: Added the "see also" section with "Third-party Assertion Libraries"

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -367,6 +367,40 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
 `AssumptionViolatedException` to signal that a test should be aborted instead of marked
 as a failure.
 
+[[writing-tests-exceptions]]
+=== Other Exceptions
+
+JUnit Jupiter provides robust support for handling test exceptions. This includes the built-in mechanisms for managing test failures due to exceptions, the role of exceptions in implementing assertions, and how to specifically assert non-throwing conditions in code.
```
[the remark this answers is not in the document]
driptaroop: sorry I missed this one first because I had soft wrapping enabled. did it now.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -367,6 +367,40 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
...
+
+[[writing-tests-uncaught-exceptions]]
```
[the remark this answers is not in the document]
driptaroop: made the anchors more hierarchical

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
driptaroop: added assertJ's non-exception assertion in this section

documentation/src/test/java/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java
driptaroop: done

documentation/src/test/java/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java
driptaroop: I created separate sections for `assertThrows`, `assertThrowsExactly`, and `assertDoesNotThrow` and moved individual examples accordingly.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
driptaroop: created separate sections for `assertThrows`, `assertThrowsExactly`, and `assertDoesNotThrow`

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
driptaroop: added a section describing the throws clause in junit exception context

documentation/src/test/java/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java
```diff
@@ -0,0 +1,33 @@
...
+package example.exception;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+import org.junit.jupiter.api.Test;
+
+// @formatter:off
+// tag::user_guide[]
+
+class AssertDoesNotThrowExceptionDemoTestCase {
```
sbrannen, line 20: I realize this is not officially documented, but test classes in the `documentation` module must be named `*Demo`.

https://github.com/junit-team/junit5/blob/3e03c97f83d1ff35dcb2b3a7cc48b54943eb1b69/documentation/documentation.gradle.kts#L160

So, please rename all `*TestCase` example classes to `*Demo` so that they are run with the Gradle build.

sbrannen (changes requested):
Thanks for making all of the requested changes.

This is looking very good. 👍 

I just added one (hopefully final) request to rename the example test classes.

documentation/src/test/java/example/exception/AssertDoesNotThrowExceptionDemoTestCase.java
driptaroop: Thanks for the suggestion. I initially named the classes as `*Demo` but that was failing the build because of `UncaughtExceptionHandlingDemoTestCase` (which was designed to fail). So I renamed it and others (to keep parity) to `*TestCase` to exclude from build. If I rename then to `*Demo` then the build will fail. 

One thing I can do is I can rename all the tests except `UncaughtExceptionHandlingDemoTestCase` to make the builds pass. do you think that's a good idea? do you have any other suggestions?

sbrannen: > do you have any other suggestions?

Yes. 😉 

Please name them all `*Demo` and make use of our `@ExpectToFail` annotation/extension.

See `example.AssertionsDemo.timeoutExceeded()` for details.

sbrannen: https://github.com/junit-team/junit5/blob/3e03c97f83d1ff35dcb2b3a7cc48b54943eb1b69/documentation/src/test/java/example/AssertionsDemo.java#L131-L143

driptaroop: ah this is nice and exactly what I wanted. Thanks for the suggestion. I have changed all the classes name to `*Demo`

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -367,6 +367,100 @@ NOTE: As of JUnit Jupiter 5.4, it is also possible to use methods from JUnit 4's
...
+is an instance of `AssertionError`.
+
+[[writing-tests-exceptions-expected]]
+==== Asserting Expected Exception Conditions
+
+JUnit Jupiter offers specialized assertions for testing that specific exceptions are
+thrown under expected conditions. The `assertThrows` and `assertThrowsExactly` assertions
+are critical tools for validating that your code responds correctly to error conditions
+by throwing the appropriate exceptions.
+
```
Marcono1234, line 421: Maybe it would be useful to mention that `assertThrows` and `assertThrowsExactly` return the thrown exception to allow performing additional assertions on it, such as checking its message?
And then maybe include an example for this in `ExceptionAssertionDemo` and `ExceptionAssertionExactDemo` (or adjust the existing ones)?

documentation/src/test/java/example/exception/ExceptionAssertionExactDemo.java
```diff
@@ -0,0 +1,38 @@
...
+public class ExceptionAssertionExactDemo {
+
+    // end::user_guide[]
+    @extensions.ExpectToFail
+    // tag::user_guide[]
+    @Test
+    void testExpectedExceptionIsThrown() {
+        assertThrowsExactly(IllegalArgumentException.class, () -> {
+            throw new IllegalArgumentException("Expected exception");
+        }, "This will fail because because we expect IllegalArgumentException to be thrown");
```
Marcono1234, line 29: This message looks incorrect? The test will actually pass because exactly the same exception type is thrown.

documentation/src/test/java/example/exception/ExceptionAssertionExactDemo.java
```diff
@@ -0,0 +1,38 @@
...
+
+        assertThrowsExactly(RuntimeException.class, () -> {
+            throw new IllegalArgumentException("Expected exception");
+        }, "This will fail because expected exactly RuntimeException to be thrown, not subclasses");
```
Marcono1234, line 33: To the maintainers: Would it make sense to rather have this "This will fail because ..." as comment instead of as message? Because as message it seems a bit weird.

sbrannen: Good point, @Marcono1234.

Like with the timeout assertion examples in the [Assertions](https://junit.org/junit5/docs/current/user-guide/#writing-tests-assertions) section, let's move all "comments" to actual code comments preceding the corresponding assertions, and let's do that consistently across all examples introduced in this PR.

Consequently, let's remove the custom failure messages completely. If we later come up with succinct failure messages that make sense in the examples (which is likely difficult due to the simplicity of the examples), we can add them after this PR, but I'd like to get this PR finished and merged soon (potentially in time for 5.11 M1).

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
sbrannen: Yes, I think it's worth mentioning that explicitly.

The `exceptionTesting()` example in the Assertions section already demonstrates how to check the exception message in the returned exception, but I agree that it's worth repeating that in the `assertThrows` and `assertThrowsExactly` examples in these new sections.

documentation/src/test/java/example/exception/ExceptionAssertionDemo.java
```diff
@@ -18,15 +19,23 @@
...
-        assertThrows(IllegalArgumentException.class, () -> {
+        // The following assertion succeeds because the code under assertion throws
+        // the expected IllegalArgumentException
+        // The assertion also returns the thrown exception which can be used for
+        // further assertions like asserting the exception messages
+        IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> {
             throw new IllegalArgumentException("Expected exception");
-        }, "Expected IllegalArgumentException to be thrown");
+        });
+        assertEquals(exception.getMessage(), "Expected exception");
```
Marcono1234, line 32: The arguments should be switched; these methods define left as _expected_ and right as _actual_, see the [`assertEquals` documentation](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertEquals(java.lang.Object,java.lang.Object)).

Otherwise the assertion error thrown on mismatch will be quite confusing.
```suggestion
        assertEquals("Expected exception", exception.getMessage());
```

documentation/src/test/java/example/exception/ExceptionAssertionExactDemo.java
```diff
@@ -24,13 +25,20 @@ public class ExceptionAssertionExactDemo {
...
-        assertThrowsExactly(IllegalArgumentException.class, () -> {
+        // The following assertion succeeds because the code under assertion throws
+        // IllegalArgumentException which is exactly equal to the expected type
+        // The assertion also returns the thrown exception which can be used for
+        // further assertions like asserting the exception messages
+        IllegalArgumentException exception = assertThrowsExactly(IllegalArgumentException.class, () -> {
             throw new IllegalArgumentException("Expected exception");
-        }, "This will fail because because we expect IllegalArgumentException to be thrown");
+        });
+        assertEquals(exception.getMessage(), "Expected exception");
```
Marcono1234, line 35:
```suggestion
        assertEquals("Expected exception", exception.getMessage());
```

sbrannen (approved):
Thanks for making all of the requested changes.

This looks good now.

I'll perform any further polishing after merging this PR.

driptaroop:
thank you for the suggestions and changes. Afaik, you will merge this and there is nothing else for me to do right?

sbrannen:
> Afaik, you will merge this and there is nothing else for me to do right?

Correct

sbrannen:
This has been merged into `main` in f915d67f186693b9a98a0cbead6707af3da6d8ea and revised in 27f3b7cf68d268721ca32edd11cdd175a51afa35.

Thanks for the PR and the great collaboration! 👍

Issue #3040, opened 2022-09-20
Provide built-in support for GraalVM native images


sbrannen:
## Overview

The GraalVM [Native Build Tools](https://graalvm.github.io/native-build-tools/latest/index.html) (NBT) project provides support for running tests within a native image using the JUnit Platform. Since JUnit 5 has never provided built-in support for running in a native image, the NBT team implemented an internal `PluginConfigProvider` mechanism for contributing native image configuration for the JUnit Platform, JUnit Jupiter, and JUnit Vintage projects.

- [`PlatformConfigProvider`](https://github.com/graalvm/native-build-tools/blob/master/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/platform/PlatformConfigProvider.java)
- [`JupiterConfigProvider`](https://github.com/graalvm/native-build-tools/blob/master/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/jupiter/JupiterConfigProvider.java)
- [`VintageConfigProvider`](https://github.com/graalvm/native-build-tools/blob/master/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/config/vintage/VintageConfigProvider.java)

The `JUnitPlatformFeature` also [registers reflection for all test classes](https://github.com/graalvm/native-build-tools/blob/3cf049769e879840dd182a0985c3ecc55a936cc7/common/junit-platform-native/src/main/java/org/graalvm/junit/platform/JUnitPlatformFeature.java#L150-L158).

## Rationale

The code in the above classes is very specific to JUnit 5 and can therefore be better maintained by the JUnit team. In addition, having the native image configuration directly in JUnit 5 artifacts will allow the JUnit team to proactively integration test support for GraalVM native images, thereby avoiding issues such as #3035.

## Deliverables

- [ ] Add a new section on GraalVM native image support in the User Guide -- for example, in the "Running Tests" section.
- [ ] Introduce integration tests that run on the CI server which verify support for GraalVM native images without using Native Build Tools (as a companion to recently introduced integration tests that use NBT).
- [ ] Avoid build-time initialization in JUnit 5 wherever feasible.
- [ ] Determine how much of the functionality in the `PluginConfigProvider` implementations in NBT can be _migrated_ to JUnit 5.
- [ ] Determine how much of the functionality in the `JUnitPlatformFeature` and its support classes can be _migrated_ to JUnit 5.

wilkinsona:
[...]

PR #3785, opened 2024-04-18
Fix class-level execution conditions on GraalVM

### LINKED ISSUE #3745: `@EnabledInNativeImage` is not properly supported at the class level
## Steps to reproduce

- `@EnabledInNativeImage` is only valid when marked on the function body and has no effect when marked on the class. This means that the following snippet has no problem.
```java
import org.junit.jupiter.api.condition.EnabledInNativeImage;

import static org.junit.jupiter.api.Assertions.assertEquals;

public class FunctionLevelTest {
    @Test
    @EnabledInNativeImage
    void testOnFunction() {
        assertEquals(1 + 2, 3);
    }
}
```
- But in the following snippet, `@EnabledInNativeImage` is invalid.
```java
import org.junit.jupiter.api.condition.EnabledInNativeImage;

import static org.junit.jupiter.api.Assertions.assertEquals;

@EnabledInNativeImage
public class ClassLevelTest {

    @Test
    void testOnClass(){
        assertEquals(1 + 2, 3);
    }
}
```
- I created https://github.com/linghengqian/junit5-native-image-annotation-test to demonstrate. Here are the steps to reproduce.
```shell
sdk install java 21.0.2-graalce
sdk use java 21.0.2-graalce

git clone git@github.com:linghengqian/junit5-native-image-annotation-test.git
cd ./junit5-native-image-annotation-test/
./mvnw -T1C -e clean test
./mvnw -PnativeTestInJunit -T1C -e clean test
```

- In the final Log, we can see that `com.lingh.ClassLevelTest.testOnClass()` is not executed in nativeTest.
```shell
Finished generating 'native-tests' in 25.1s.
[INFO] Executing: /home/linghengqian/IdeaProjects/junit5-native-image-annotation-test/target/native-tests --xml-output-dir /home/linghengqian/IdeaProjects/junit5-native-image-annotation-test/target/native-test-reports -Djunit.platform.listeners.uid.tracking.output.dir=/home/linghengqian/IdeaProjects/junit5-native-image-annotation-test/target/test-ids
JUnit Platform on Native Image - report
----------------------------------------

com.lingh.FunctionLevelTest > testOnFunction() SUCCESSFUL


Test run finished after 1 ms
[         2 containers found      ]
[         0 containers skipped    ]
[         2 containers started    ]
[         0 containers aborted    ]
[         2 containers successful ]
[         0 containers failed     ]
[         1 tests found           ]
[         0 tests skipped         ]
[         1 tests started         ]
[         0 tests aborted         ]
[         1 tests successful      ]
[         0 tests failed          ]
```

- More early investigations are from https://github.com/apache/shardingsphere/pull/30031 and https://github.com/apache/shardingsphere/issues/29052 .

## Context

 - Used versions (Jupiter/Vintage/Platform): `Jupiter`
 - Build Tool/IDE: `Maven Wrapper 3.9.5`/`IntelliJ IDEA 2023.3.6 (Ultimate Edition)`

## Deliverables

- Null.

marcphilipp:
- **Update to latest version of native build tools**
- **Add failing integration test**
- **Track descendants of skipped containers**
- **Add to release notes**

Fixes #3745.

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/UniqueIdTrackingListener.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/UniqueIdTrackingListener.java
@@ -167,7 +167,8 @@ public class UniqueIdTrackingListener implements TestExecutionListener {
 
 	private boolean trackTestUid(TestIdentifier testIdentifier) {
 		if (testIdentifier.isTest()) {
-			return this.uniqueIds.add(testIdentifier.getUniqueId());
+			this.uniqueIds.add(testIdentifier.getUniqueId());
+			return true;
 		}
 		return false;
 	}
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/UniqueIdTrackingListener.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/UniqueIdTrackingListener.java
@@ -130,6 +130,7 @@ public class UniqueIdTrackingListener implements TestExecutionListener {
 	private final List<String> uniqueIds = new ArrayList<>();
 
 	private boolean enabled;
+	private TestPlan testPlan;
 
 	public UniqueIdTrackingListener() {
 		// to avoid missing-explicit-ctor warning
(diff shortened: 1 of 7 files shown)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/UniqueIdTrackingListener.java
```diff
@@ -138,22 +139,37 @@ public UniqueIdTrackingListener() {
...
+			this.testPlan.getChildren(testIdentifier).forEach(this::trackTestUidRecursively);
+		}
 	}
 
-	private void trackTestUid(TestIdentifier testIdentifier) {
-		if (this.enabled && testIdentifier.isTest()) {
-			this.uniqueIds.add(testIdentifier.getUniqueId());
+	private boolean trackTestUid(TestIdentifier testIdentifier) {
+		if (testIdentifier.isTest()) {
+			return this.uniqueIds.add(testIdentifier.getUniqueId());
```
sbrannen, line 170: This will always return `true` since it's an `ArrayList`.

Did you perhaps mean to switch to a `LinkedHashSet`?

If not, perhaps we should change that to the following for better clarity.

```suggestion
			this.uniqueIds.add(testIdentifier.getUniqueId());
			return true;
```

platform-tooling-support-tests/projects/graalvm-starter/build.gradle.kts
```diff
@@ -31,8 +31,6 @@ tasks.test {
 graalvmNative {
 	binaries {
 		named("test") {
-			buildArgs.add("--initialize-at-build-time=org.junit.platform.launcher.core.LauncherConfig")
-			buildArgs.add("--initialize-at-build-time=org.junit.jupiter.engine.config.InstantiatingConfigurationParameterConverter")
```
sbrannen, line 35: Even though recent versions of Native Build Tools register those as well, I don't think we should remove this configuration.

- see https://github.com/junit-team/junit5/issues/3040

sbrannen (changes requested):
Thanks for putting together this PR!

I added two comments for consideration.

platform-tooling-support-tests/projects/graalvm-starter/build.gradle.kts
marcphilipp: Why should we keep it? The reason for adding this here was because Native Build Tools didn't already configure those. I think adding a [`native-image.properties`](https://www.graalvm.org/22.1/reference-manual/native-image/BuildConfiguration/#embedding-a-configuration-file) file is a separate issue.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/UniqueIdTrackingListener.java
marcphilipp: I know that it will always return `true`. I kept the `ArrayList` to avoid changing to much. That being said, I think your suggestion does make it clearer so I'll follow it.

platform-tooling-support-tests/projects/graalvm-starter/build.gradle.kts
sbrannen: Ahhh... yes, you're right.

This is just our test build script. I apologize for not thinking that through.

Please just disregard that comment, and I'll _hide_ it. 😉

sbrannen (approved):

Issue #2677, opened 2021-07-29
Add ExclusiveResource programmatically


U1F984:
I have several test classes, each with multiple test methods inside. Some of those classes are injected with a parameter, and when running classes in parallel those classes should not run parallel. If I understand correctly, the `@ResourceLock` annotation can help here, but it would be very inconvenient to manually add that annotation to hundereds of tests. Ideally a extension could take a look at the class, see that it has this parameter and somehow add the ExclusiveResource to it.

My JUnit execution parameters:

```
-Djunit.jupiter.execution.parallel.enabled=true
-Djunit.jupiter.execution.parallel.mode.default=same_thread
-Djunit.jupiter.execution.parallel.mode.classes.default=concurrent
-Djunit.jupiter.execution.parallel.config.strategy=fixed
-Djunit.jupiter.execution.parallel.config.fixed.parallelism=8
```

marcphilipp:
We could add a new extension point for this:

```java
interface ResourceLockProvider extends Extension {
	default Set<ExclusiveResource> provideResourcesForClass(Class<?> testClass, Set<ExclusiveResource> declaredResources) {
		return declaredResources;
	};
	Set<ExclusiveResource> provideResourcesForNestedClass(Class<?> nestedClass, Set<ExclusiveResource> declaredResources) {
		return declaredResources;
	};
	Set<ExclusiveResource> provideResourcesForMethod(Class<?> testClass, Method testMethod, Set<ExclusiveResource> declaredResources) {
		return declaredResources;
	};
}
interface ExclusiveResource {
	String getKey();
	ResourceAccessMode getAccessMode();
	static ExclusiveResource of(String key);
	static ExclusiveResource of(String key, ResourceAccessMode accessMode);
}

```

jbduncan:
This feature might be quite useful for my [PR](https://github.com/junit-pioneer/junit-pioneer/pull/491) on junit-pioneer to allow "resources" - arbitrary objects - to be instantiated and shared amongst tests declaratively. 

Specifically, if each "resource" were to have its own `ResourceLock` that could be created programmatically, then it would simplify making my extension thread-safe.

The proposed API leaves me with one question, though: if my extension were to implement `ResourceLockProvider`, when would the `provideX` methods be called?

marcphilipp:
It would have to be called immediately before starting execution on the engine level.

jbduncan:
Okay, thank you!

[...]

PR #3889, opened 2024-07-16
Allow determining exclusive resources programmatically

VladimirDmitrienko:
Solves: #2677

## Overview

Introduce adding `shared resources` programmatically via `@ResourceLock(providers = ...)` attribute
which accepts an array of one or more classes implementing `ResourceLocksProvider` interface.

It allows to add resources in runtime (immediately before starting execution on the engine level).

Resources can be distributed based on any test class / nested test class / test method attributes (e.g. package name, class / method name etc.) or any other user logic.

This approach serves as a more flexible and less verbose alternative for cases in which:

- adding lots of `@ResourceLock(value, mode)` annotations manually may be inconvenient;
- shared resources are not known until runtime.

---

---

### Definition of Done

- [x] Change is documented in the [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/)
- [x] Change is documented in the [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLocksFrom.java
```diff
@@ -0,0 +1,65 @@
...
+ * @see ResourceAccessMode
+ * @see ResourceLock
+ * @see ResourceLocksProvider
+ * @see ResourceLocksProvider.Lock
+ */
+@API(status = EXPERIMENTAL, since = "5.12")
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+public @interface ResourceLocksFrom {
```
marcphilipp, line 58: After discussion in the team, we'd prefer adding a `Class<? extends ResourceLocksProvider>[] providers() default {}` attribute to the existing `@ResourceLock` annotation (and making `value()` optional by giving it an empty default value).

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLocksProvider.java
```diff
@@ -0,0 +1,187 @@
...
+	 *
+	 * @since 5.12
+	 * @see Isolated
+	 * @see Resources
+	 * @see ResourceAccessMode
+	 * @see ResourceLock
+	 * @see ResourceLocksFrom
+	 * @see ResourceLocksProvider
+	 */
+	final class Lock {
```
marcphilipp, line 121: I wonder if it we should make this an interface and pass a `LockFactory` to the methods of `ResourceLocksProvider` rather than making the implementation part of the API. Having such a parameter object has proven useful in many cases but it might be overengineering here. Thoughts?

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -2930,8 +2930,15 @@ parallel with each other but not while any other test that requires `READ_WRITE`
...
+.Declaring shared resources with `{ResourceLock}` annotation
 ----
-include::{testDir}/example/SharedResourcesDemo.java[tags=user_guide]
+include::{testDir}/example/SharedResourcesWithResourceLockAnnotationDemo.java[tags=user_guide]
+----
+
+[source,java]
+.Declaring shared resources provider with `{ResourceLocksFrom}` annotation
+----
+include::{testDir}/example/SharedResourcesWithResourceLocksFromAnnotationDemo.java[tags=user_guide]
```
marcphilipp, line 2941: We should also mention here (and in the annotations' Javadoc) that "static" and "dynamic" locks are combined.

marcphilipp (changes requested):
Looks very promising! Sorry for the late request to change the API.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLocksProvider.java
VladimirDmitrienko: Actually, I considered `Lock` like an immutable data carrier which has no behavior, that's why I'm not sure about an interface in this case.

Also, the current usage of `Lock` reminds `@ResourceLock`:
```java
@ResourceLock("key");
// both create a 'READ_WRITE' resource identified by 'key':
new Lock("key");
```

Maybe you have in mind some use cases for an interface?

marcphilipp: Not really, let's go with the record-like data class.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
VladimirDmitrienko: Updated. 
Mentioned in Javadoc for `@ResourceLock` and `ResourceLocksProvider`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLocksFrom.java
VladimirDmitrienko: Updated.

VladimirDmitrienko:
Hi @marcphilipp 
Maybe you have any updates regarding this PR?

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -2927,21 +2927,30 @@ In addition to controlling the execution mode using the `{Execution}` annotation
...
-name can be user-defined or one of the predefined constants in `{Resources}`:
+execution. The shared resource is identified by a unique name which is a `String`.
+The name can be user-defined or one of the predefined constants in `{Resources}`:
 `SYSTEM_PROPERTIES`, `SYSTEM_OUT`, `SYSTEM_ERR`, `LOCALE`, or `TIME_ZONE`.
 
-If the tests in the following example were run in parallel _without_ the use of
-{ResourceLock}, they would be _flaky_. Sometimes they would pass, and at other times they
-would fail due to the inherent race condition of writing and then reading the same JVM
-System Property.
+Since Junit Jupiter 5.12 `{ResourceLock}` annotation has 'providers' attribute
```
marcphilipp, line 2934: Since the User Guide is versioned, we usually avoid mentioning specific versions when features were introduced. Thus, please remove that bit.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -2927,21 +2927,30 @@ In addition to controlling the execution mode using the `{Execution}` annotation
...
+Sometimes they would pass, and at other times they would fail due to
+the inherent race condition of writing and then reading the same JVM System Property.
+
+When access to shared resources is declared using the `{ResourceLock}` annotation
+or added with `{ResourceLocksProvider}` interface, the JUnit Jupiter engine uses
+this information to ensure that no conflicting tests are run in parallel.
+This guarantee extends to lifecycle methods of a test class or method.
+For example, if a test method is annotated with a `{ResourceLock}` annotation,
+the "lock" will be acquired before any `@BeforeEach` methods are executed
+and released after all `@AfterEach` methods have been executed.
```
marcphilipp, line 2953: Please revert the changes made to the formatting of these two paragraphs.

documentation/src/test/java/example/sharedresources/DynamicSharedResourcesDemo.java
```diff
@@ -0,0 +1,86 @@
...
+		System.setProperty("my.prop", "banana");
+		assertEquals("banana", System.getProperty("my.prop"));
+	}
+
+	static final class Provider implements ResourceLocksProvider {
+
+		@Override
+		public Set<Lock> provideForMethod(Class<?> testClass, Method testMethod) {
+			ResourceAccessMode mode;
+			if (testMethod.getName().equals("customPropertyIsNotSetByDefault")) {
```
marcphilipp, line 73: Maybe it would make for a better example if we'd check for `testMethod.getName().startsWith("canSet")` and invert the `if`?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
```diff
@@ -81,7 +87,24 @@ public final Set<TestTag> getTags() {
...
+		// @formatter:off
+		Function<ResourceLocksProvider, Set<ResourceLocksProvider.Lock>> providerToLocks =
+				p -> p.provideForMethod(getTestClass(), getTestMethod());
+
+		Set<ExclusiveResource> fromMethodLevelAnnotation = getExclusiveResourcesFromAnnotation(
+				getTestMethod(),
+				providerToLocks
+		);
+		Stream<ExclusiveResource> fromProvidersInClassLevelAnnotation = getExclusiveResourcesFromAnnotationProviders(
+				getTestClass(),
```
marcphilipp, line 99: What about `@Nested` test classes? I think we'll have to collect exclusive resources from providers from the test class and, if it's a `@Nested` one, also from its enclosing class, recursively.

marcphilipp (changes requested):

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
marcphilipp: On second thought, there is no need to apply locks from the class level to methods as it is redundant. When executing the method, the class-level locks are already held.

VladimirDmitrienko: > What about `@Nested` test classes? I think we'll have to collect exclusive resources from providers from the test class and, if it's a `@Nested` one, also from its enclosing class, recursively.

I suppose that the behaviour of `ResourceLocksProvider` should be similar to the plain `@ResourceLock("...")`.
Let's consider such an example:
```java
@ResourceLock("A")
static class TestClass {

    @Nested
    @ResourceLock("B")
    class NestedTestClass {
    }
}
```
`NestedTestClass` gets only `B` but **does not** get `A`.

That's why I'm not sure that `@Nested` classes should invoke providers of enclosing classes.

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
VladimirDmitrienko: Got it. Done

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
VladimirDmitrienko: Done

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
marcphilipp: Please disregard my initial comment. We should neither check the test class nor any of its enclosing classes since that's already being done by the parent test descriptors. Thus, please remove the extra code that sets `fromProvidersInClassLevelAnnotation` etc.

VladimirDmitrienko: If we remove this code then `LocksProvider#provideForMethod()` will not be invoked for `test()`:
```java
@ResourceLock(providers = LocksProvider.class)
static class TestClass {

    @Test
    void test() {
    }
}
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -180,11 +186,38 @@ public static ExecutionMode toExecutionMode(org.junit.jupiter.api.parallel.Execu
 		throw new JUnitException("Unknown ExecutionMode: " + mode);
 	}
 
-	Set<ExclusiveResource> getExclusiveResourcesFromAnnotation(AnnotatedElement element) {
+	Set<ExclusiveResource> getExclusiveResourcesFromAnnotation(AnnotatedElement element,
+			Function<ResourceLocksProvider, Set<ResourceLocksProvider.Lock>> providerToLocks) {
+		// @formatter:off
+		return Stream.concat(
+				getExclusiveResourcesFromAnnotationValue(element),
+				getExclusiveResourcesFromAnnotationProviders(element, providerToLocks)
```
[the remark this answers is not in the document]
VladimirDmitrienko: Done

documentation/src/test/java/example/sharedresources/DynamicSharedResourcesDemo.java
VladimirDmitrienko: Good point.
Done

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
marcphilipp: Hmm, I see. I didn't think about that use case. In that case, we also need to ensure `LocksProvider#provideForMethod()` gets called for nested classes, e.g.:

```java
@ResourceLock(providers = LocksProvider.class)
static class TestClass { // A: `LocksProvider#provideForClass()` is called
	@Nested
	class Inner { // B: `LocksProvider#provideForNestedClass()` is called
    	@Test
    	void test() { // C: `LocksProvider#provideForMethod()` is called
    	}
	}
}
```

If I'm reading the current code correctly, neither B nor C would be called.

marcphilipp (changes requested):

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
VladimirDmitrienko: You're right.
Done

platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/ResourceLockAnnotationTests.java
```diff
@@ -0,0 +1,377 @@
...
+				new ExclusiveResource("b2", LockMode.READ)
+		);
+
+		var nestedClassResources = getNestedClassResources(
+				SharedResourcesViaAnnotationValueAndProvidersTestCase.NestedClass.class
+		);
+		assertThat(nestedClassResources).containsExactlyInAnyOrder(
+				new ExclusiveResource("c1", LockMode.READ_WRITE),
+				new ExclusiveResource("c2", LockMode.READ_WRITE)
+		);
```
marcphilipp, line 156: The provider declared on the outermost class is called as well, right? I think we should change the test to show that.

marcphilipp (changes requested):

platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/ResourceLockAnnotationTests.java
VladimirDmitrienko: Yes, that's right.
But such case is already covered in `ResourceLocksProviderTests#ClassLevelProviderTestCase`.
Do you think we should cover it here as well?

VladimirDmitrienko: Indeed. We didn't test such case.
Added

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -2970,8 +2978,15 @@ parallel with each other but not while any other test that requires `READ_WRITE`
 to the same shared resource is running.
 
 [source,java]
+.Declaring shared resources "statically" with `{ResourceLock}` annotation
+----
+include::{testDir}/example/sharedresources/StaticSharedResourcesDemo.java[tags=user_guide]
+----
+
+[source,java]
+.Adding shared resources "dynamically" with `{ResourceLocksProvider}` interface
```
marcphilipp, line 2987:
```suggestion
.Adding shared resources "dynamically" with `{ResourceLocksProvider}` implementation
```

documentation/src/test/java/example/sharedresources/DynamicSharedResourcesDemo.java
```diff
@@ -0,0 +1,86 @@
...
+			ResourceAccessMode mode;
+			if (testMethod.getName().startsWith("canSet")) {
+				mode = READ_WRITE;
+			}
+			else {
+				mode = READ;
+			}
+			Set<Lock> locks = new HashSet<>();
+			locks.add(new Lock(SYSTEM_PROPERTIES, mode));
+			return locks;
```
marcphilipp, line 81:
```suggestion
			return Collections.singleton(new Lock(SYSTEM_PROPERTIES, mode));
```

documentation/src/test/java/example/sharedresources/DynamicSharedResourcesDemo.java
```diff
@@ -0,0 +1,86 @@
...
+
+		@Override
+		public Set<Lock> provideForMethod(Class<?> testClass, Method testMethod) {
+			ResourceAccessMode mode;
+			if (testMethod.getName().startsWith("canSet")) {
+				mode = READ_WRITE;
+			}
+			else {
+				mode = READ;
+			}
```
marcphilipp, line 78:
```suggestion
			ResourceAccessMode mode = testMethod.getName().startsWith("canSet") ? READ_WRITE : READ;
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLock.java
```diff
@@ -48,10 +49,20 @@
  * <p>Since JUnit Jupiter 5.4, this annotation is {@linkplain Inherited inherited}
  * within class hierarchies.
  *
+ * <p>Since JUnit Jupiter 5.12, this annotation supports adding shared resources
+ * in runtime via {@link ResourceLock#providers}.
```
marcphilipp, line 53:
```suggestion
 * dynamically at runtime via {@link ResourceLock#providers}.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLocksProvider.java
```diff
@@ -0,0 +1,223 @@
...
+import org.apiguardian.api.API;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.platform.commons.util.Preconditions;
+import org.junit.platform.commons.util.ToStringBuilder;
+
+/**
+ * {@code @ResourceLocksProvider} is used to add shared resources
+ * to a test class and / or its test methods in runtime.
```
marcphilipp, line 29:
```suggestion
 * to a test class and / or its test methods dynamically at runtime.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLocksProvider.java
```diff
@@ -0,0 +1,223 @@
...
+	 */
+	default Set<Lock> provideForMethod(Class<?> testClass, Method testMethod) {
+		return emptySet();
+	}
+
+	/**
+	 *
+	 * <p>{@link Lock} represents a shared resource.
+	 *
+	 * <p>Each resource is identified by {@link #key}.
```
marcphilipp, line 121:
```suggestion
	 * <p>Each resource is identified by a {@link #key}.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLocksProvider.java
```diff
@@ -0,0 +1,223 @@
...
+	 * In addition,{@link #accessMode} allows you to specify
```
marcphilipp, line 122:
```suggestion
	 * In addition, the {@link #accessMode} allows to specify
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/ResourceLocksProvider.java
```diff
@@ -0,0 +1,223 @@
...
+	 * In the latter case, the test class or test method
+	 * may be executed concurrently with other tests
+	 * that also require {@code READ} access but not at the same time
+	 * as any other test that requires {@code READ_WRITE} access.
+	 *
+	 * <p>This guarantee extends to lifecycle methods of a test class or method.
+	 * For example, if a {@code Lock} is added to a test method
+	 * then the "lock" will be acquired before any
+	 * {@link BeforeEach @BeforeEach} methods are executed and released after all
+	 * {@link AfterEach @AfterEach} methods have been executed.
```
marcphilipp, line 137: I don't think we need to repeat all that here, in particular since we link to the Javadoc of `@ResourceLock` in the `@apiNote` below.

```suggestion
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -180,11 +200,46 @@ public static ExecutionMode toExecutionMode(org.junit.jupiter.api.parallel.Execu
...
 
-	Set<ExclusiveResource> getExclusiveResourcesFromAnnotation(AnnotatedElement element) {
+	Set<ExclusiveResource> getExclusiveResourcesFromAnnotations(AnnotatedElement element,
+			Function<ResourceLocksProvider, Set<ResourceLocksProvider.Lock>> providerToLocks) {
+		// @formatter:off
+		List<ResourceLock> ownAnnotations = findRepeatableAnnotations(element, ResourceLock.class);
+		List<ResourceLock> enclosingClassesAnnotations = getEnclosingTestClasses().stream()
+				.map(clazz -> findRepeatableAnnotations(clazz, ResourceLock.class))
+				.flatMap(Collection::stream)
+				.collect(toList());
```
marcphilipp, line 210: I think it would be slightly better to get the registered `ResourceLockProviders` from the parent test descriptor rather than collecting them again for every test and reinstantiating them but I'll take care of that in a follow-up PR.

marcphilipp (approved):
Thanks! I'll take it from here.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
VladimirDmitrienko: You're right, many method / nested class descriptors may share the same parent 👍

VladimirDmitrienko:
> Thanks! I'll take it from here.

Thank you! 
This was my first contribution, and it was very interesting and valuable experience.

PR #3890, opened 2024-07-17
Introduce zero invocations in test templates and parameterized tests

### LINKED ISSUE #1477: Introduce support for zero invocations in test templates and parameterized tests
## Overview

Using: JUnit: 5.2.0

Great job on the new `ParameterizedTest` support in v.5. The replacement of the static, one-per-class Parameters annotation with more flexible `MethodSource`, etc. has been like a breath of fresh air and allowed me to remove thousands (!!) of lines of supporting code from my system. I'm really loving it.
However, someone decided to disallow zero parameters with this precondition check in `ParameterizedTestExtension`.

```java
.onClose(() ->
    Preconditions.condition(invocationCount.get() > 0,
    "Configuration error: You must provide at least one argument for this @ParameterizedTest"));
```

The problem with this is that we have some testing situations where parameterized tests with zero parameters are not exceptional. For example, we run tests against thousands of a certain type of class generically against a database of past production failures, and many of these classes have never experienced a production failure. When the tests are run, we now get failures due the above precondition check. JUnit 4 handled this cleanly: it would simply not run any tests on those classes.

## Workaround

I can get around this by adding a `null` to the method creating the collection of parameters if it is empty, and then return from the beginning of the `@ParameterizedTest` method code if the passed parameter is `null`. That lets us continue to run the parameterized tests against all of the classes, but comes with some disadvantages:

- We must add specific code to the front and back of every parameterized pair just to avoid a failure that doesn't matter to the tests.
- The handling of the `null` causes the test count inflation for these "phantom" tests.
- `null` is now reserved as a signal for no parameters, rather than something wrong in the parameter creation.

## Proposal

If nobody has any strong feelings about disallowing the no-parameter case, can we just have this precondition removed from a future version?

Thanks.

nskvortsov:
Add a flag to ParameterizedTest to control arguments requirement. This allows users to explicitly opt out from validation of arguments set count and silently skip a test if no arguments are provided

In general, support TestTemplateInvocationContextProvider returning zero invocation contexts. Such providers must override new interface method to indicate that the framework should expect "no context returned"

Resolves #1477

## Overview

Implemented the team decision described in [this comment](https://github.com/junit-team/junit5/issues/1477#issuecomment-1919412893)

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
@@ -10,6 +10,7 @@
 
 package org.junit.jupiter.api.extension;
 
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.STABLE;
 
 import java.util.stream.Stream;
@@ -85,5 +86,23 @@ public interface TestTemplateInvocationContextProvider extends Extension {
 	 * @see ExtensionContext
 	 */
 	Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(ExtensionContext context);
+
+	/**
+	 * Signals that in the supplied {@linkplain ExtensionContext context} the provider may return zero
+	 * {@linkplain TestTemplateInvocationContext invocation contexts}.
+	 *
+	 * <p>If provider returns empty stream from {@link #provideTestTemplateInvocationContexts(ExtensionContext)}
+	 * this will be considered an execution error. Override this method to ignore the absence of invocation contexts.
+	 *
+	 * @param context the extension context for the test template method about
+	 * to be invoked; never {@code null}
+	 * @return {@code true} to allow zero contexts, {@code false} (default) to fail execution in case of zero contexts.
+	 *
+	 * @since 5.12
+	 */
+	@API(status = EXPERIMENTAL, since = "5.12")
+	default boolean mayReturnZeroInvocationContexts(ExtensionContext context) {
+		return false;
+	}
 
 }
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
@@ -88,15 +88,19 @@ public interface TestTemplateInvocationContextProvider extends Extension {
 	Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(ExtensionContext context);
 
 	/**
-	 * Signals that in the supplied {@linkplain ExtensionContext context} the provider may return zero
+	 * Signal that in the supplied {@linkplain ExtensionContext context} the
+	 * provider may return zero
 	 * {@linkplain TestTemplateInvocationContext invocation contexts}.
 	 *
-	 * <p>If provider returns empty stream from {@link #provideTestTemplateInvocationContexts(ExtensionContext)}
-	 * this will be considered an execution error. Override this method to ignore the absence of invocation contexts.
+	 * <p>If this method returns {@code false} and the provider returns an empty
+	 * stream from {@link #provideTestTemplateInvocationContexts}, this will be
+	 * considered an execution error. Override this method to ignore the absence
+	 * of invocation contexts for this provider.
 	 *
 	 * @param context the extension context for the test template method about
 	 * to be invoked; never {@code null}
-	 * @return {@code true} to allow zero contexts, {@code false} (default) to fail execution in case of zero contexts.
+	 * @return {@code true} to allow zero contexts, {@code false} (default) to
+	 * fail execution in case of zero contexts.
 	 *
 	 * @since 5.12
 	 */
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
@@ -105,7 +105,7 @@ public interface TestTemplateInvocationContextProvider extends Extension {
 	 * @since 5.12
 	 */
 	@API(status = EXPERIMENTAL, since = "5.12")
-	default boolean mayReturnZeroInvocationContexts(ExtensionContext context) {
+	default boolean mayReturnZeroTestTemplateInvocationContexts(ExtensionContext context) {
 		return false;
 	}
 
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java
@@ -101,18 +101,28 @@ public class TestTemplateTestDescriptor extends MethodBasedTestDescriptor implem
 			context.getExtensionRegistry());
 		AtomicInteger invocationIndex = new AtomicInteger();
 		for (TestTemplateInvocationContextProvider provider : providers) {
-			int initialValue = invocationIndex.get();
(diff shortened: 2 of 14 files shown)


pain64:
Need this feature a lot! Is there a forecast for when it will be ready?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
```diff
@@ -86,4 +87,22 @@ public interface TestTemplateInvocationContextProvider extends Extension {
...
+	 * this will be considered an execution error. Override this method to ignore the absence of invocation contexts.
+	 *
+	 * @param extensionContext the extension context for the test template method about
+	 * to be invoked; never {@code null}
+	 * @return {@code true} to allow zero contexts, {@code false} (default) to fail execution in case of zero contexts.
+	 *
+	 * @since 5.11
+	 */
+	@API(status = EXPERIMENTAL, since = "5.11")
+	default boolean mayReturnEmptyInvocationContext(ExtensionContext extensionContext) {
```
marcphilipp, line 104:
```suggestion
	default boolean mayReturnZeroInvocationContexts(ExtensionContext context) {
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java
```diff
@@ -142,13 +146,16 @@ private void execute(DynamicTestExecutor dynamicTestExecutor, TestDescriptor tes
...
 
-		Preconditions.condition(invocationIndex > 0,
+		boolean allMayReturnEmptyContext = providers.stream().allMatch(
+			p -> p.mayReturnEmptyInvocationContext(extensionContext));
+
+		Preconditions.condition(invocationIndex > 0 || allMayReturnEmptyContext,
 			() -> "None of the supporting " + TestTemplateInvocationContextProvider.class.getSimpleName() + "s "
-					+ providers.stream().map(provider -> provider.getClass().getSimpleName()).collect(
-						joining(", ", "[", "]"))
+					+ providers.stream().filter(p -> !p.mayReturnEmptyInvocationContext(extensionContext)).map(
```
marcphilipp, line 157: I don't think we should filter them out here.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
```diff
@@ -86,4 +87,22 @@ public interface TestTemplateInvocationContextProvider extends Extension {
...
+	 * {@linkplain TestTemplateInvocationContext invocation contexts}
+	 *
+	 * <p>If provider returns empty stream from {@link #provideTestTemplateInvocationContexts(ExtensionContext)}
+	 * this will be considered an execution error. Override this method to ignore the absence of invocation contexts.
+	 *
+	 * @param extensionContext the extension context for the test template method about
+	 * to be invoked; never {@code null}
+	 * @return {@code true} to allow zero contexts, {@code false} (default) to fail execution in case of zero contexts.
+	 *
+	 * @since 5.11
```
marcphilipp, line 101: Should be 5.12 now

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
```diff
@@ -92,11 +94,19 @@ public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContex
...
 								"Configuration error: You must configure at least one set of arguments for this @ParameterizedTest"));
 		// @formatter:on
 	}
 
+	@Override
+	public boolean mayReturnEmptyInvocationContext(ExtensionContext extensionContext) {
+		Method templateMethod = extensionContext.getRequiredTestMethod();
+		return findAnnotation(templateMethod, ParameterizedTest.class).map(
+			parameterizedTest -> !parameterizedTest.requireArguments()).orElse(
+				TestTemplateInvocationContextProvider.super.mayReturnEmptyInvocationContext(extensionContext));
```
marcphilipp, line 107:
```suggestion
		return findAnnotation(templateMethod, ParameterizedTest.class) //
				.map(parameterizedTest -> !parameterizedTest.requireArguments()) //
				.orElse(false);
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
```diff
@@ -86,4 +87,22 @@ public interface TestTemplateInvocationContextProvider extends Extension {
 	 */
 	Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(ExtensionContext context);
 
+	/**
+	 * Signals that in the given {@linkplain ExtensionContext extension context} the provider may return zero
+	 * {@linkplain TestTemplateInvocationContext invocation contexts}
```
marcphilipp, line 92:
```suggestion
	 * Signals that in the supplied {@linkplain ExtensionContext context} the provider may return zero
	 * {@linkplain TestTemplateInvocationContext invocation contexts}.
```

marcphilipp (changes requested):
Looks very promising! :+1:

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java
nskvortsov: @marcphilipp I added this filter to narrow down the list. There is no need for a developer to investigate well-behaving providers, so their presence on the error message would be useless.
Why do you think we should keep them?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
nskvortsov: Addressed.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
nskvortsov: updated

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
nskvortsov: updated

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java
nskvortsov: reworded

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java
jbduncan: > Out of curiosity: where does this format with trailing empty comment come from? (I see it is required by Spotless)

I can speak on this as a former Spotless contributor!

It simply splits long lines into multiple shorter lines. It's a common trick with a few formatters like Eclipse and google-java-format. It works because these formatters leave the empty comments alone.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/TestTemplateTestDescriptor.java
nskvortsov: I agree. Per-provider validation will be most precise. I've added one more commit to the pull request.
Just a side note: this will be a more strict check in case of multiple providers. But sounds like a corner case to me.

nskvortsov:
Let me know if you are ok with the latest change, I will do the rebase.

marcphilipp (approved):

marcphilipp:
> Let me know if you are ok with the latest change, I will do the rebase.

Thanks for making the changes. I'll take it from here!

marcphilipp:
@nskvortsov Thank you! :+1:

PR #3899, opened 2024-07-25
Introduce `@BeforeSuite` and `@AfterSuite` annotations

### LINKED ISSUE #456: Introduce `@BeforeSuite` and `@AfterSuite` annotations

juliette-derancourt:
## Overview

Resolves #456 

---

---

### Definition of Done

### DIFF
--- a/junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/AfterSuite.java
+++ b/junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/AfterSuite.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2015-2024 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.suite.api;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+
+/**
+ * {@code @AfterSuite} is used to signal that the annotated method should be
+ * executed <em>after</em> <strong>all</strong> tests in the current test suite.
+ *
+ * <h2>Method Signatures</h2>
+ *
+ * <p>{@code @AfterSuite} methods must have a {@code void} return type, must
+ * be {@code static} and must not be {@code private}.
+ *
+ * <h2>Inheritance and Execution Order</h2>
+ *
+ * <p>{@code @AfterSuite} methods are inherited from superclasses as long as they
+ * are not <em>overridden</em> according to the visibility rules of the Java
+ * language. Furthermore, {@code @AfterSuite} methods from superclasses will be
+ * executed after {@code @AfterSuite} methods in subclasses.
+ *
+ * <p>JUnit Jupiter does not guarantee the execution order of multiple
+ * {@code @AfterSuite} methods that are declared within a single test class or
+ * test interface. While it may at times appear that these methods are invoked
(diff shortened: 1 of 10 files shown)


junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/AfterSuite.java
```diff
@@ -0,0 +1,75 @@
...
+ * be {@code static} and must not be {@code private}.
+ *
+ * <h2>Inheritance and Execution Order</h2>
+ *
+ * <p>{@code @AfterSuite} methods are inherited from superclasses as long as they
+ * are not <em>overridden</em> according to the visibility rules of the Java
+ * language. Furthermore, {@code @AfterSuite} methods from superclasses will be
+ * executed after {@code @AfterSuite} methods in subclasses.
+ *
+ * <p>JUnit Jupiter does not guarantee the execution order of multiple
```
mpkorstanje, line 39: Should probably say something other than JUnit Jupiter. Maybe Suite Engine or JUnit 5?

Or as this is an API package, perhaps this should be described as a specification of the implementation rather than a description of a specific implementation.

platform-tests/src/test/java/org/junit/platform/suite/engine/BeforeAndAfterSuiteTests.java
```diff
@@ -0,0 +1,231 @@
...
+import org.junit.platform.testkit.engine.EngineExecutionResults;
+import org.junit.platform.testkit.engine.EngineTestKit;
+
+/**
+ * Integration tests that verify support for {@link BeforeSuite} and {@link AfterSuite},
+ * in the {@link SuiteTestEngine}.
+ *
+ * @since 1.11
+ */
+@Nested
```
mpkorstanje, line 64: Is this needed here?

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/AfterSuite.java
sbrannen: > Should probably say something other than JUnit Jupiter.

Good catch: it should not mention JUnit Jupiter.

> Maybe JUnit Platform Suite Engine or JUnit 5?

We don't like to refer to "JUnit 5" in documentation, since it may someday be JUnit 6.

So, perhaps just "JUnit" is sufficient here.

> Or as this is an API package, perhaps this should be described as a specification of the implementation rather than a description of a specific implementation.

Good point. Given that it's an API that could have different implementations, describing it as a spec would be "correct".

However, I don't imagine there will be many implementations going forward beyond the built-in Suite Engine.

So, perhaps we should mention the "JUnit Platform Suite Engine" (as you suggested above) and leave it at that (at least for the 1.11 RC1 release).

platform-tests/src/test/java/org/junit/platform/suite/engine/BeforeAndAfterSuiteTests.java
sbrannen: If you're referring to `@Nested` on a top-level class, I'd say, "No." 😉

junit-platform-suite-api/src/main/java/org/junit/platform/suite/api/AfterSuite.java
marcphilipp: Done in 057d30855848a3b7a70b8ff66df7eff0914d1480

platform-tests/src/test/java/org/junit/platform/suite/engine/BeforeAndAfterSuiteTests.java
marcphilipp: Done in c656fdcc6dbe1fd1c0377a551daaf5d845956ed1

marcphilipp (approved):

PR #3938, opened 2024-08-28
Document benefits of `messageSupplier` in `Assertions`

### LINKED ISSUE #3153: Document benefits of `messageSupplier` in `Assertions`

chris-carneiro:
## Overview

Resolves #3153.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -293,6 +293,13 @@ JUnit Jupiter comes with many of the assertion methods that JUnit 4 has and adds
 that lend themselves well to being used with Java 8 lambdas. All JUnit Jupiter assertions
 are `static` methods in the `{Assertions}` class.
 
+Assertion methods optionally accept the assertion message as their third parameter, which
+can be either a `String` or a `Supplier<String>`.
+
+When using a `Supplier<String>` (e.g., a lambda expression), the message is evaluated
+lazily. This can provide a performance benefit, especially if message construction is
+complex or time-consuming, as it is only evaluated when the assertion fails.
+
 [source,java,indent=0]
 ----
 include::{testDir}/example/AssertionsDemo.java[tags=user_guide]
--- a/documentation/src/test/java/example/AssertionsDemo.java
+++ b/documentation/src/test/java/example/AssertionsDemo.java
@@ -41,8 +41,9 @@ class AssertionsDemo {
 		assertEquals(2, calculator.add(1, 1));
 		assertEquals(4, calculator.multiply(2, 2),
 				"The optional failure message is now the last parameter");
-		assertTrue('a' < 'b', () -> "Assertion messages can be lazily evaluated -- "
-				+ "to avoid constructing complex messages unnecessarily.");
+
+		// Lazily evaluates generateFailureMessage('a','b').
+		assertTrue('a' < 'b', () -> generateFailureMessage('a','b'));
 	}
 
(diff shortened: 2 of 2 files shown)


documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -293,6 +293,14 @@ JUnit Jupiter comes with many of the assertion methods that JUnit 4 has and adds
 that lend themselves well to being used with Java 8 lambdas. All JUnit Jupiter assertions
 are `static` methods in the `{Assertions}` class.
 
+Assertion methods can accept an optional third parameter, which can be either a `String` or a `Supplier<String>`.
+When using a `Supplier<String>` (e.g., a lambda expression) as the message supplier,
+the message is evaluated lazily.
+
+Lazily evaluating the assertion message can provide a performance benefit,
+especially if the message construction is complex or time-consuming,
+as it is only evaluated when the assertion fails.
```
marcphilipp, line 302:
```suggestion
Assertion methods optionally accept the assertion message as their third parameter, which
can be either a `String` or a `Supplier<String>`.

When using a `Supplier<String>` (e.g., a lambda expression), the message is evaluated
lazily. This can provide a performance benefit, especially if message construction is
complex or time-consuming, as it is only evaluated when the assertion fails.
```

marcphilipp (changes requested):
@chris-carneiro Thanks for taking this on! :+1:

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
chris-carneiro: You're welcome

marcphilipp (approved):

PR #3941, opened 2024-08-31
Add thread-per-test-class execution model

### LINKED ISSUE #3939: Add "thread per test class" execution model
I recently ran into an ugly OOM caused by `ThreadLocal`s produced via different class loaders (Quarkus test class loaders) leaking into the singleton `Test worker` thread. Some background on the issue is [here](https://github.com/projectnessie/nessie/issues/9441#issuecomment-2320636765) and on the [Quarkus chat](https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/ThreadLocal.20.22fun.22.20in.20Quarkus.20tests).

Long story short: I was able to work around the `ThreadLocal`-OOM by running each test class on its own dedicated thread, as implemented in [this PR](https://github.com/projectnessie/nessie/pull/9453/files#diff-3d8e942362529e5d4cd36a30d4b8962c4548dc17ef3109b3fbaeecbcbaa4f688).

I wonder whether the JUnit project would be open for a PR to implement such a "thread per test class" model - basically a 3rd implementation in addition to `SameThreadHierarchicalTestExecutorService` and `ForkJoinPoolHierarchicalTestExecutorService`. This new execution model seems useful for other Quarkus users as well.

snazy:
The thread-per-test-class execution model runs tests via one thread per test class. This model is primarily useful in situations when test executions attach `ThreadLocal`s to the test worker thread, which "leak" into other test class executions. This problem can lead to out-of-memory errors, if the `ThreadLocal`s reference (large) object trees, either as its value or via its initial-value `Supplier`, which cannot be cleaned up / garbage collected, because the `ThreadLocal` is referenced by the test worker thread.

The problem becomes even worse, if the test class creates its own class loader and attaches `ThreadLocal`s that reference classes loaded by such class loaders. In such cases the whole class loader including all its loaded classes and (static) state is not eligible for garbage collection, leaking more heap (and non-heap) memory.

Using one thread per test class works around the above problem(s), because once the (ephemeral) thread per test-class finishes, the whole thread and all its `ThreadLocal`s become eligible for garbage collection.

Particularly Quarkus unit tests (`@QuarkusTest` annotated test classes) benefit from this execution model.

This change cannot eliminate other sources of similar leaks, like threads spawned from tests or not removed MBeans. Those kinds of leaks are better handled by the test code or the tested code providing "proper" cleanup mechanisms.

This new execution is implemented via the introduced `ThreadPerClassHierarchicalTestExecutorService`, a 3rd model in addition to `SameThreadHierarchicalTestExecutorService` and `ForkJoinPoolHierarchicalTestExecutorService`. It is enabled if `junit.jupiter.execution.threadperclass.enabled` is set to `true` and `junit.jupiter.execution.parallel.enabled` is `false`.

Resolves #3939.

## Overview

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ThreadPerClassHierarchicalTestExecutorService.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ThreadPerClassHierarchicalTestExecutorService.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2015-2024 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.engine.support.hierarchical;
+
+import static java.lang.String.format;
+import static java.time.Duration.ofMinutes;
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.apiguardian.api.API.Status.STABLE;
+
+import java.time.Duration;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.apiguardian.api.API;
+import org.junit.platform.commons.JUnitException;
+import org.junit.platform.engine.ConfigurationParameters;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.UniqueId;
+
+/**
+ * A {@linkplain HierarchicalTestExecutorService executor service} that creates a new thread for
+ * each test class, all {@linkplain TestTask test tasks}.
+ *
+ * <p>This execution model is useful to prevent some kinds of class / class-loader leaks. For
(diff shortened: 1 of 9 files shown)


snazy:
Marking this PR as draft, I'm reworking the changes. The current state is rather a preparation.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ThreadPerClassHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,133 @@
...
+		tasks.forEach(this::executeTask);
+	}
+
+	protected void executeTask(TestTask testTask) {
+		NodeTestTask<?> nodeTestTask = (NodeTestTask<?>) testTask;
+		TestDescriptor testDescriptor = nodeTestTask.getTestDescriptor();
+
+		UniqueId.Segment lastSegment = testDescriptor.getUniqueId().getLastSegment();
+
+		if ("class".equals(lastSegment.getType())) {
```
mpkorstanje, line 75: This is an implementation detail of the JUnit Jupiter Engine. Perhaps you can use `TestDescriptor.getSource` instead to potentially obtain the class source?

Note: I'm not part of the JUnit Team, just commenting.

marcphilipp (commented):
Thanks for the PR! I'm not convinced whether it's addressing the problem on the right level, though. An alternative would be for Jupiter to provide a new `ClassExecutionInterceptor` extension point (or a new method in `InvocationInterceptor`) that would be called from `ClassTestDescriptor.around`. I'll bring this up for discussion with the team.

marcphilipp:
Thanks for your PR! Closing with rationale in https://github.com/junit-team/junit5/issues/3939#issuecomment-2422437188.

PR #3936, opened 2024-08-26
Fix potential double acquisition in `SingleLock`

leonard84:
`SingleLock` uses a `ForkJoinPool.ManagedBlocker` for efficient blocking in a fork-join pool. In the `isReleasable()` method it calls `tryLock()` without remembering the outcome. The `block` method also unconditionally called `lockInterruptibly()` even if the lock was already acquired. The problem that can arise is that the lock is acquired multiple times, but only released once.

`CompositeLockManagedBlocker` is also updated, so that `block()` checks for `acquired` before locking.

## Overview

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement).

---

### Definition of Done

- [x] There are no TODOs left in the code
- [x] Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.commons/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [x] [Coding conventions](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [ ] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling
- [x] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [x] Change is documented in the [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)

PR #3981, opened 2024-09-11
Fix deadlock in `ForkJoinPoolHierarchicalTestExecutorService`

### LINKED ISSUE #3945: Workstealing can lead to deadlocks
## Steps to reproduce

This does not work every time as it is dependent on multiple factors.

- Checkout https://github.com/spockframework/spock/commit/b69f9f115abfbdb381e8e32a9a9063d93c68588a
- run `./gradlew :spock-specs:test --rerun -Dvariant=3.0 -DjavaVersion=21`
- You might need to try a few times. This depends a bit on your machine, but if it takes longer than 5m then it is most likely stuck.
- You can attach a debugger to inspect the current executing tests

## What is Broken?
As you can see in these two screenshots we switch from executing `MockSpecInfoAnnotationSpec` to `ByteBuddyMockFactoryConcurrentSpec` due to workstealing. The problem is that `MockSpecInfoAnnotationSpec` holds a read-lock for the `org.junit.platform.engine.support.hierarchical.ExclusiveResource.GLOBAL_KEY` but `ByteBuddyMockFactoryConcurrentSpec` needs a write-lock.

image
image

## Context

 - Used versions (Jupiter/Vintage/Platform): 5.11.0/5.12.0-SNAPSHOT
 - Build Tool/IDE: Gradle

## Deliverables

- [ ] ...

leonard84:
## Overview

fixes #3945

This PR actually fixes the problem that was attempted in #3936. 
That PR is still valid on its own, although it didn't fix the actual problem.
The release notes still apply for this as well.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
@@ -52,6 +52,7 @@ import org.junit.platform.engine.ConfigurationParameters;
 public class ForkJoinPoolHierarchicalTestExecutorService implements HierarchicalTestExecutorService {
 
 	private final ForkJoinPool forkJoinPool;
+	private final TaskEventListener taskEventListener;
 	private final int parallelism;
 	private final ThreadLocal<ThreadLock> threadLocks = ThreadLocal.withInitial(ThreadLock::new);
 
@@ -73,7 +74,13 @@ public class ForkJoinPoolHierarchicalTestExecutorService implements Hierarchical
 	 */
 	@API(status = STABLE, since = "1.10")
 	public ForkJoinPoolHierarchicalTestExecutorService(ParallelExecutionConfiguration configuration) {
+		this(configuration, TaskEventListener.NOOP);
+	}
+
+	ForkJoinPoolHierarchicalTestExecutorService(ParallelExecutionConfiguration configuration,
+			TaskEventListener taskEventListener) {
 		forkJoinPool = createForkJoinPool(configuration);
+		this.taskEventListener = taskEventListener;
 		parallelism = forkJoinPool.getParallelism();
 		LoggerFactory.getLogger(getClass()).config(() -> "Using ForkJoinPool with parallelism of " + parallelism);
 	}
@@ -242,20 +249,13 @@ public class ForkJoinPoolHierarchicalTestExecutorService implements Hierarchical
 			// and let the worker thread fork it once it is done with the current task.
 			ResourceLock resourceLock = testTask.getResourceLock();
 			ThreadLock threadLock = threadLocks.get();
-			//				System.out.println(
-			//						"Checking " + this + " lock compatibility: " + resourceLock + " vs " + threadLock.locks + " " + Thread.currentThread());
 			if (!threadLock.areAllHeldLocksCompatibleWith(resourceLock)) {
-				//					System.out.println(
-				//							"Deferring task " + this + " because of lock incompatibility: " + resourceLock + " vs " + threadLock.locks);
-
 				threadLock.addDeferredTask(this);
+				taskEventListener.deferred(testTask);
 				// Return false to indicate that this task is not done yet
 				// this means that .join() will wait.
 				return false;
 			}
-			//			else {
-			//				System.out.println("No existing thread lock for " + this + " " + Thread.currentThread());
-			//			}
 			try (ResourceLock lock = resourceLock.acquire()) {
 				threadLock.incrementNesting(lock);
 				testTask.execute();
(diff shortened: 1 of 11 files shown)


sbrannen:
Is the intention to backport this to `5.10.x` along with #3936 and #3988?

sbrannen:
> Is the intention to backport this to `5.10.x` along with #3936 and #3988?

Never mind: I see this is linked to #3945 which is already assigned to `5.10.4`.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/LockManager.java
```diff
@@ -44,18 +48,27 @@ private static Comparator<String> globalKeyFirst() {
...
 
 	ResourceLock getLockForResource(ExclusiveResource resource) {
-		return new SingleLock(toLock(resource));
+		Lock lock = toLock(resource);
+		if (GLOBAL_READ.equals(resource)) {
+			return new SingleLock.GlobalReadLock(lock);
+		}
+		else if (GLOBAL_READ_WRITE.equals(resource)) {
+			return new SingleLock.GlobalReadWriteLock(lock);
+		}
```
leonard84, line 61: 🤔  We should make fields for these two globals and return the same instance.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/LockManager.java
```diff
@@ -44,18 +48,27 @@ private static Comparator<String> globalKeyFirst() {
...
+		return new SingleLock(lock);
 	}
 
 	private List<Lock> getDistinctSortedLocks(Collection<ExclusiveResource> resources) {
+		if (resources.isEmpty()) {
+			return emptyList();
+		}
+		if (resources.size() == 1) {
+			return singletonList(toLock(getOnlyElement(resources)));
```
leonard84, line 70: :x: Isn't this bypassing the globals handling? The old code would invoke `getLockForResource`.
The new code will still use the same underlying `lock` instance but it won't use the correct `SingleLock` subclass.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
```diff
@@ -43,4 +43,12 @@ default void close() {
 		release();
 	}
 
+	/**
+	 * {@return whether the given lock is compatible with this lock}
+	 * @param other the other lock to check for compatibility
+	 */
+	default boolean isCompatible(ResourceLock other) {
+		return other instanceof NopLock || this instanceof NopLock;
```
leonard84, line 51: nit: We should probably switch this, as the `this` case is more frequent.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/LockManager.java
marcphilipp: Done in c87721f9e60035b60ba4354fdcb4cdfd0c1b7739.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/LockManager.java
marcphilipp: You're right! Done in c87721f9e60035b60ba4354fdcb4cdfd0c1b7739.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
marcphilipp: Done in cd364fc3c3fb4d33f6b946964912166950588718.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
```diff
@@ -51,7 +52,9 @@
 public class ForkJoinPoolHierarchicalTestExecutorService implements HierarchicalTestExecutorService {
 
 	private final ForkJoinPool forkJoinPool;
+	private final TaskEventListener taskEventListener;
 	private final int parallelism;
+	private final ThreadLocal<ThreadLock> threadLocks = ThreadLocal.withInitial(ThreadLock::new);
```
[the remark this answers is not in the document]
marcphilipp: Yes, it would. I think we could remove the `.remove()` call to further simplify it without changing the behavior. Any objections?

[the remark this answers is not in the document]
leonard84: Since we own the FJ pool and close it after the execution, we can get away with "leaking" the thread locals. If we used the common FJ pool, this would be a problem.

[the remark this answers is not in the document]
marcphilipp: I have decided to further simplify it in 5256049af3655c80599e7df839049f4272d16ffc

PR #4004, opened 2024-09-18
Run isolated tasks last

### LINKED ISSUE #3928: Poor thread utilization when using ExclusiveResource.GLOBAL_KEY
When executing in parallel using a fixed number of threads, when several tests attempt to acquire a read/write lock against the `GLOBAL_KEY`, threads are likely to be assigned in a sub-optimal manner.

This issue was originally reported as https://github.com/cucumber/cucumber-jvm/issues/2910. The key to this issue is the low max-pool size. This is a common, if crude, way to limit the number of active web drivers.

## Steps to reproduce

See https://github.com/mpkorstanje/junit5-scheduling for a minimal reproducer.

```
junit.jupiter.execution.parallel.enabled=true
junit.jupiter.execution.parallel.mode.default=concurrent
junit.jupiter.execution.parallel.config.strategy=fixed
junit.jupiter.execution.parallel.config.fixed.parallelism=3
junit.jupiter.execution.parallel.config.fixed.max-pool-size=3
```

```java
@Isolated
class SerialATest {

    @BeforeEach
    void simulateWebDriver() throws InterruptedException {
        System.out.println("Serial A: " + Thread.currentThread().getName() );
        Thread.sleep(1000);
    }

   ... Several tests

}
```

```java
@Isolated
class SerialBTest {

   ... Copy of SerialATest

}
```

```java
class ParallelTest {

    @BeforeEach
    void simulateWebDriver() throws InterruptedException {
        System.out.println("Parallel: " + Thread.currentThread().getName() );
        Thread.sleep(1000);
    }

    ... Several tests

}
```

Executing these tests will likely result in an output similar to this:

```log
Parallel: ForkJoinPool-1-worker-2
Parallel: ForkJoinPool-1-worker-2
Parallel: ForkJoinPool-1-worker-2
Parallel: ForkJoinPool-1-worker-2
Parallel: ForkJoinPool-1-worker-2
Serial A: ForkJoinPool-1-worker-3
Serial A: ForkJoinPool-1-worker-3
Serial B: ForkJoinPool-1-worker-1
Serial B: ForkJoinPool-1-worker-1
```

The output implies that  `worker-1` and `worker-3` are waiting to acquire the `ExclusiveResource.GLOBAL_KEY`, leaving only `worker-2` to process the parallel section on its own. Once done, the other workers can then acquire the lock in turn. In the ideal scenario, the parallel section would be executed with the maximum number of workers.

## Context

 - Jupiter: 5.10.3
 - Java 17

## Deliverables

To be determined.

marcphilipp:
- **Fix/suppress warnings**
- **Add failing test case**
- **Run isolated tasks after all other tasks**
- **Add to release notes**

Resolves #3928.

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
@@ -13,6 +13,7 @@ package org.junit.platform.engine.support.hierarchical;
 import static java.util.concurrent.CompletableFuture.completedFuture;
 import static org.apiguardian.api.API.Status.STABLE;
 import static org.junit.platform.engine.support.hierarchical.Node.ExecutionMode.CONCURRENT;
+import static org.junit.platform.engine.support.hierarchical.Node.ExecutionMode.SAME_THREAD;
 
 import java.lang.Thread.UncaughtExceptionHandler;
 import java.lang.reflect.Constructor;
@@ -38,6 +39,7 @@ import org.junit.platform.commons.function.Try;
 import org.junit.platform.commons.logging.LoggerFactory;
 import org.junit.platform.commons.util.ExceptionUtils;
 import org.junit.platform.engine.ConfigurationParameters;
+import org.junit.platform.engine.support.hierarchical.SingleLock.GlobalReadWriteLock;
 
 /**
  * A {@link ForkJoinPool}-based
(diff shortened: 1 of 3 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
```diff
@@ -155,29 +157,34 @@ public void invokeAll(List<? extends TestTask> tasks) {
 			new ExclusiveTask(tasks.get(0)).execSync();
 			return;
 		}
-		Deque<ExclusiveTask> nonConcurrentTasks = new LinkedList<>();
+		Deque<ExclusiveTask> isolatedTasks = new LinkedList<>();
+		Deque<ExclusiveTask> sameThreadTasks = new LinkedList<>();
```
leonard84, line 161: 🤔 why are we using `LinkedList` instead of `ArrayDeque`?
It will probably not really matter performance wise.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
```diff
@@ -155,29 +157,34 @@ public void invokeAll(List<? extends TestTask> tasks) {
 			new ExclusiveTask(tasks.get(0)).execSync();
 			return;
 		}
-		Deque<ExclusiveTask> nonConcurrentTasks = new LinkedList<>();
+		Deque<ExclusiveTask> isolatedTasks = new LinkedList<>();
```
leonard84, line 160: 🤔 The only thing that is a bit weird is that isolated tasks can only be top-level tasks, yet this is invoked on multiple levels. Probably not worth changing.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
```diff
@@ -155,29 +157,34 @@ public void invokeAll(List<? extends TestTask> tasks) {
...
-				nonConcurrentTasks.add(exclusiveTask);
+				exclusiveTask.fork();
+				concurrentTasksInReverseOrder.addFirst(exclusiveTask);
 			}
 		}
 	}
 
-	private void executeNonConcurrentTasks(Deque<ExclusiveTask> nonConcurrentTasks) {
-		for (ExclusiveTask task : nonConcurrentTasks) {
+	private void executeSync(Deque<ExclusiveTask> tasks) {
```
leonard84, line 186: I guess we are using `Deque` for consistency because here, we treat it more as a list instead of consuming the elements.

leonard84 (approved):
LGTM

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
marcphilipp: I was wondering that, too, but kept using `LinkedList` since it was already in place.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
marcphilipp: We could have a different class for the top-level task and other tasks. :thinking:

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
marcphilipp: Yes, solely for consistency, any `Iterable` would do

PR #4012, opened 2024-09-19
Allow for work stealing when only holding read locks

marcphilipp:
This allows tasks with only read locks to be stolen by threads that are
currently holding only read locks.

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
@@ -87,7 +87,6 @@ class CompositeLock implements ResourceLock {
 	public String toString() {
 		return new ToStringBuilder(this) //
 				.append("resources", resources) //
-				.append("locks", locks) //
 				.toString();
 	}
 
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
@@ -27,11 +27,14 @@ class CompositeLock implements ResourceLock {
 
 	private final List<ExclusiveResource> resources;
 	private final List<Lock> locks;
+	private final boolean exclusive;
 
 	CompositeLock(List<ExclusiveResource> resources, List<Lock> locks) {
 		Preconditions.condition(resources.size() == locks.size(), "Resources and locks must have the same size");
 		this.resources = unmodifiableList(resources);
 		this.locks = Preconditions.notEmpty(locks, "Locks must not be empty");
+		this.exclusive = resources.stream().anyMatch(
+			resource -> resource.getLockMode() == ExclusiveResource.LockMode.READ_WRITE);
 	}
 
 	@Override
@@ -73,6 +76,11 @@ class CompositeLock implements ResourceLock {
 		for (int i = acquiredLocks.size() - 1; i >= 0; i--) {
 			acquiredLocks.get(i).unlock();
 		}
+	}
+
+	@Override
+	public boolean isExclusive() {
+		return exclusive;
 	}
 
 	@Override
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
@@ -10,22 +10,35 @@
 
 package org.junit.platform.engine.support.hierarchical;
 
+import static java.util.Collections.unmodifiableNavigableSet;
+
 import java.util.ArrayList;
 import java.util.List;
+import java.util.NavigableSet;
+import java.util.SortedSet;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.locks.Lock;
 
 import org.junit.platform.commons.util.Preconditions;
+import org.junit.platform.commons.util.ToStringBuilder;
 
 /**
  * @since 1.3
  */
 class CompositeLock implements ResourceLock {
 
+	private final SortedSet<ExclusiveResource> resources;
 	private final List<Lock> locks;
 
-	CompositeLock(List<Lock> locks) {
+	CompositeLock(NavigableSet<ExclusiveResource> resources, List<Lock> locks) {
+		Preconditions.condition(resources.size() == locks.size(), "Resources and locks must have the same size");
+		this.resources = unmodifiableNavigableSet(resources);
 		this.locks = Preconditions.notEmpty(locks, "Locks must not be empty");
+	}
+
+	@Override
+	public SortedSet<ExclusiveResource> getResources() {
+		return resources;
 	}
 
 	// for tests only
@@ -62,6 +75,14 @@ class CompositeLock implements ResourceLock {
 		for (int i = acquiredLocks.size() - 1; i >= 0; i--) {
 			acquiredLocks.get(i).unlock();
 		}
+	}
+
+	@Override
+	public String toString() {
+		return new ToStringBuilder(this) //
+				.append("resources", resources) //
+				.append("locks", locks) //
+				.toString();
 	}
 
 	private class CompositeLockManagedBlocker implements ForkJoinPool.ManagedBlocker {
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
@@ -10,12 +10,10 @@
 
 package org.junit.platform.engine.support.hierarchical;
 
-import static java.util.Collections.unmodifiableNavigableSet;
+import static java.util.Collections.unmodifiableList;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.NavigableSet;
-import java.util.SortedSet;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.locks.Lock;
 
@@ -27,17 +25,17 @@ import org.junit.platform.commons.util.ToStringBuilder;
  */
 class CompositeLock implements ResourceLock {
 
-	private final SortedSet<ExclusiveResource> resources;
+	private final List<ExclusiveResource> resources;
 	private final List<Lock> locks;
 
-	CompositeLock(NavigableSet<ExclusiveResource> resources, List<Lock> locks) {
+	CompositeLock(List<ExclusiveResource> resources, List<Lock> locks) {
 		Preconditions.condition(resources.size() == locks.size(), "Resources and locks must have the same size");
-		this.resources = unmodifiableNavigableSet(resources);
+		this.resources = unmodifiableList(resources);
 		this.locks = Preconditions.notEmpty(locks, "Locks must not be empty");
 	}
 
 	@Override
-	public SortedSet<ExclusiveResource> getResources() {
+	public List<ExclusiveResource> getResources() {
 		return resources;
 	}
 
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/LockManager.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/LockManager.java
@@ -10,23 +10,25 @@
 
 package org.junit.platform.engine.support.hierarchical;
 
-import static java.util.Collections.emptyList;
-import static java.util.Collections.singletonList;
-import static java.util.Comparator.comparing;
-import static java.util.Comparator.naturalOrder;
+import static java.util.Collections.emptyNavigableSet;
+import static java.util.Collections.unmodifiableNavigableSet;
 import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.toCollection;
 import static java.util.stream.Collectors.toList;
 import static org.junit.platform.commons.util.CollectionUtils.getOnlyElement;
-import static org.junit.platform.engine.support.hierarchical.ExclusiveResource.GLOBAL_KEY;
+import static org.junit.platform.commons.util.CollectionUtils.toUnmodifiableList;
 import static org.junit.platform.engine.support.hierarchical.ExclusiveResource.GLOBAL_READ;
 import static org.junit.platform.engine.support.hierarchical.ExclusiveResource.GLOBAL_READ_WRITE;
 import static org.junit.platform.engine.support.hierarchical.ExclusiveResource.LockMode.READ;
+import static org.junit.platform.engine.support.hierarchical.ExclusiveResource.singleton;
 
 import java.util.Collection;
-import java.util.Comparator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableSet;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
@@ -40,14 +42,6 @@ import org.junit.platform.engine.support.hierarchical.SingleLock.GlobalReadWrite
  */
 class LockManager {
 
-	private static final Comparator<ExclusiveResource> COMPARATOR //
-		= comparing(ExclusiveResource::getKey, globalKeyFirst().thenComparing(naturalOrder())) //
-				.thenComparing(ExclusiveResource::getLockMode);
-
-	private static Comparator<String> globalKeyFirst() {
-		return comparing(key -> !GLOBAL_KEY.equals(key));
-	}
-
 	private final Map<String, ReadWriteLock> locksByKey = new ConcurrentHashMap<>();
 	private final GlobalReadLock globalReadLock;
 	private final GlobalReadWriteLock globalReadWriteLock;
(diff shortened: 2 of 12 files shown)


marcphilipp:
Actually, I think we can get rid of the special types for the global locks now. I'll add that to this PR before merging.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
```diff
@@ -43,12 +45,18 @@ default void close() {
...
 	 * @param other the other lock to check for compatibility
 	 */
 	default boolean isCompatible(ResourceLock other) {
-		return this instanceof NopLock || other instanceof NopLock;
+		if (this.getResources().isEmpty() || other.getResources().isEmpty()) {
+			return true;
+		}
+		return other.getResources().stream() //
+				.allMatch(it -> this.getResources().contains(it)
+						|| ExclusiveResource.COMPARATOR.compare(this.getResources().last(), it) < 0);
```
leonard84, line 60: 🧐 I would extract the `this.getResources().last()` to a local variable outside the stream.

leonard84, line 60: Can't this be simplified to
```suggestion
				.allMatch(it -> ExclusiveResource.COMPARATOR.compare(this.getResources().last(), it) <= 0);
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/LockManager.java
```diff
@@ -10,105 +10,98 @@
...
+		return toResourceLock(toDistinctSortedResources(resources));
 	}
 
 	ResourceLock getLockForResource(ExclusiveResource resource) {
-		return toResourceLock(toLock(resource));
+		return toResourceLock(singleton(resource));
 	}
 
-	private List<Lock> toDistinctSortedLocks(Collection<ExclusiveResource> resources) {
+	private NavigableSet<ExclusiveResource> toDistinctSortedResources(Collection<ExclusiveResource> resources) {
```
leonard84, line 59: 🤔 Do we really need a comparatively expensive `TreeSet` over a cheap list? We already ensure uniqueness with `.distinct()` in L69

We are fine with using `List` for the actual locks.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
```diff
@@ -67,44 +93,128 @@ void defersTasksWithIncompatibleLocks() throws Exception {
...
+				Set.of(GLOBAL_READ, new ExclusiveResource("b", LockMode.READ_WRITE)) //
+			), //
+			arguments(//
+				Set.of(GLOBAL_READ, new ExclusiveResource("a", LockMode.READ_WRITE)), //
+				Set.of(GLOBAL_READ, new ExclusiveResource("a", LockMode.READ)) //
+			), //
+			arguments(//
+				Set.of(GLOBAL_READ_WRITE), //
+				Set.of(GLOBAL_READ) //
+			)//
```
leonard84, line 129: :x: these are wrong and break the guarantees we want to ensure with our `ExclusiveResource` in the first place.
You could argue that `a`,`b` RW is fine, but I would say we should not mix RW locks.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
```diff
@@ -48,17 +59,32 @@ void exceptionsFromInvalidConfigurationAreNotSwallowed() {
...
+	static List<Arguments> incompatibleLockCombinations() {
+		return List.of(//
+			arguments(//
+				Set.of(GLOBAL_READ), //
+				Set.of(GLOBAL_READ_WRITE) //
+			), //
+			arguments(//
+				Set.of(new ExclusiveResource("a", LockMode.READ)), //
+				Set.of(new ExclusiveResource("a", LockMode.READ_WRITE)) //
+			), //
```
leonard84, line 71:
```suggestion
			arguments(//
				Set.of(new ExclusiveResource("a", LockMode.READ)), //
				Set.of(new ExclusiveResource("a", LockMode.READ_WRITE)) //
			), //
			arguments(//
				Set.of(new ExclusiveResource("a", LockMode.READ_WRITE)), //
				Set.of(new ExclusiveResource("a", LockMode.READ_WRITE)) //
			), //
```

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ResourceLockTests.java
```diff
@@ -12,68 +12,78 @@
...
+		assertTrue(nopLock.isCompatible(new SingleLock(anyResource(), anyReentrantLock())));
+		assertTrue(nopLock.isCompatible(new CompositeLock(singleton(anyResource()), List.of(anyReentrantLock()))));
 	}
 
 	@Test
-	void singleLocksAreIncompatibleWithNonNopLocks() {
-		var singleLock = new SingleLock(anyReentrantLock());
+	void singleLocksAreIncompatibleWithOtherLocksThatCanPotentiallyCauseDeadlocks() {
+		ExclusiveResource resourceB = resource("B", LockMode.READ);
+		var singleLock = new SingleLock(resourceB, anyReentrantLock());
```
leonard84, line 42: 🧐 The name suggests that we check for incompatibility, yet there are compatible assertions here. I would split it accordingly, even making it data-driven for easier comprehension.

Applies to the other tests as well.

leonard84 (changes requested):

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
```diff
@@ -10,24 +10,37 @@
...
 /**
  * @since 1.3
  */
 class CompositeLock implements ResourceLock {
 
+	private final SortedSet<ExclusiveResource> resources;
 	private final List<Lock> locks;
 
-	CompositeLock(List<Lock> locks) {
+	CompositeLock(NavigableSet<ExclusiveResource> resources, List<Lock> locks) {
```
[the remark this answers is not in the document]
marcphilipp: We only require the methods in `SortedSet` but there's no `Collections.unmodifiableSortedSet` method.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
marcphilipp: Why would that be incompatible?

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
marcphilipp: Which guarantees do these break?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/LockManager.java
marcphilipp: Changed in 9edf0e05e016a555d0eb01ad011a5c8113739ee2

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ResourceLockTests.java
marcphilipp: I think it's easier to see compatible vs. incompatible this way because the example are directly next to each other.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
marcphilipp: No, since that would disallow "common prefixes", e.g. both having the global read lock but `other` having another (read) lock.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
marcphilipp: As discussed, this would break the exclusiveness guarantee of a test (class) holding a RW lock.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
marcphilipp: As discussed elsewhere, as soon as one side is holding a RW lock, we can't allow any additional locks to avoid breaking the exclusiveness guarantees.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
```diff
@@ -43,12 +46,47 @@ default void close() {
 		release();
 	}
 
+	/**
+	 * {@return the exclusive resources this lock represents}
+	 */
+	List<ExclusiveResource> getResources();
+
+	/**
+	 * {@return whether this lock requires exclusiveness}
```
leonard84, line 55: 😁 I see you've readily adapted the new Javadoc tag.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
```diff
@@ -43,12 +46,47 @@ default void close() {
...
+
+		// Whenever there's a READ_WRITE lock, it's incompatible with any other lock
+		// because we guarantee that all children will have exclusive access to the
+		// resource in question. In practice, whenever a READ_WRITE lock is present,
+		// NodeTreeWalker will force all children to run in the same thread so that
+		// it should never attempt to steal work from another thread, and we shouldn't
+		// actually reach this point.
+		boolean isGlobalReadLock = ownResources.size() == 1
+				&& ExclusiveResource.GLOBAL_READ.equals(ownResources.get(0));
+		if ((!isGlobalReadLock && other.isExclusive()) || this.isExclusive()) {
```
leonard84, line 80: Why not just
```suggestion
		if (other.isExclusive() || this.isExclusive()) {
```

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
```diff
@@ -67,44 +108,128 @@ void defersTasksWithIncompatibleLocks() throws Exception {
...
+				Set.of(GLOBAL_READ, new ExclusiveResource("a", LockMode.READ)) //
+			), //
+			arguments(//
+				Set.of(GLOBAL_READ, new ExclusiveResource("a", LockMode.READ)), //
+				Set.of(GLOBAL_READ, new ExclusiveResource("a", LockMode.READ)) //
+			), //
+			arguments(//
+				Set.of(GLOBAL_READ, new ExclusiveResource("a", LockMode.READ)), //
+				Set.of(GLOBAL_READ, new ExclusiveResource("b", LockMode.READ)) //
+			)//
```
leonard84, line 144: Shouldn't you also test some with more than a single additional resource?

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
```diff
@@ -48,17 +58,48 @@ void exceptionsFromInvalidConfigurationAreNotSwallowed() {
...
+	static List<Arguments> incompatibleLockCombinations() {
+		return List.of(//
+			arguments(//
+				Set.of(GLOBAL_READ), //
+				Set.of(GLOBAL_READ_WRITE) //
+			), //
+			arguments(//
+				Set.of(new ExclusiveResource("a", LockMode.READ)), //
+				Set.of(new ExclusiveResource("a", LockMode.READ_WRITE)) //
+			), //
```
leonard84, line 70: I'd also include examples of invalid multi-lock combinations.

e.g.
- `GR, aR, bR, dR` - `GR, aR, cR`

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
```diff
@@ -67,44 +108,128 @@ void defersTasksWithIncompatibleLocks() throws Exception {
...
+				System.out.println("Interrupted while waiting for work to be stolen");
+			}
+		});
+
+		assertNull(deferredTask.get());
+		assertEquals(tasks.get("nestedTask").threadName, tasks.get("leafTask2").threadName);
+		assertNotEquals(tasks.get("leafTask1").threadName, tasks.get("leafTask2").threadName);
+	}
+
+	private static Map<String, DummyTestTask> runWithAttemptedWorkStealing(TaskEventListener taskEventListener,
```
leonard84, line 176: Right now we only try one level of nesting, but shouldn't we also check multiple levels?

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ResourceLockTests.java
```diff
@@ -12,68 +12,120 @@
...
 	void nopLocksAreCompatibleWithEverything() {
 		var nopLock = NopLock.INSTANCE;
 
+		assertTrue(nopLock.isCompatible(nopLock));
 		assertTrue(nopLock.isCompatible(NopLock.INSTANCE));
-		assertTrue(nopLock.isCompatible(new SingleLock(anyReentrantLock())));
-		assertTrue(nopLock.isCompatible(new SingleLock.GlobalReadLock(anyReentrantLock())));
-		assertTrue(nopLock.isCompatible(new SingleLock.GlobalReadWriteLock(anyReentrantLock())));
-		assertTrue(nopLock.isCompatible(new CompositeLock(List.of(anyReentrantLock()))));
+		assertTrue(nopLock.isCompatible(new SingleLock(anyResource(), anyReentrantLock())));
```
leonard84, line 32: 🤔 introduce helper methods like you already did for `resource` to further reduce the noise

```
SingleLock singleLock(ExclusiveResource er) {
    return new SingleLock(er, anyReentrantLock());
}

CompositeLock compositeLock(List<ExclusiveResource> ers) {
    return new CompositeLock(er, ers.stream().map(__ -> anyReentrantLock()).toList());
}
```

You could also introduce assertion methods.
```
void compatible(ResourceLock a, ResourceLock b) {
    assertTrue(a.isCompatible(b));
}

void incompatible(ResourceLock a, ResourceLock b) {
    assertFalse(a.isCompatible(b));
}
```

Then this line would be

```
compatible(nopLock, singleLock(anyResource()));
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/CompositeLock.java
```diff
@@ -64,6 +78,19 @@ private void release(List<Lock> acquiredLocks) {
...
+	@Override
+	public boolean isExclusive() {
+		return exclusive;
+	}
+
+	@Override
+	public String toString() {
+		return new ToStringBuilder(this) //
+				.append("resources", resources) //
+				.append("locks", locks) //
```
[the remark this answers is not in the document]
marcphilipp: Not really. Removed in dc3135dc43a4141dc1452696535bc3e61a65a3f5.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
marcphilipp: Indeed! :slightly_smiling_face:

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
marcphilipp: Because we need to make the global read lock (which is always on direct children of the engine node) compatible with the first write lock (which may be on a test method). 

This was already the case before this set of changes:
https://github.com/junit-team/junit5/blob/81d347986208573a654196dfcb3e2ed09ab493a5/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/SingleLock.java#L63-L72

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
marcphilipp: You mean with additional locks on every level?

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
marcphilipp: Done in 5a88d210b7ad79bcd9935e5179fa97da50b16a31.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
marcphilipp: Done in 5a88d210b7ad79bcd9935e5179fa97da50b16a31.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ResourceLockTests.java
marcphilipp: Polished in f18f048196d82ad98b43ad96a6cbcfca436f63f3.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
leonard84: Ideally, we'd test this; the parentheses mean that they are deferred at that level.

A -> (B) -> C -> (D)

Can we achieve it in an easier scenario?

leonard84 (approved):
The implementation looks good. Only some test suggestions and style bikeshedding left.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ResourceLock.java
marcphilipp: Done in ec118028e190703a0f59cf77cfc9c1ae52b23da1.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ResourceLockTests.java
```diff
@@ -12,71 +12,135 @@
...
+		return NopLock.INSTANCE;
+	}
+
+	private static SingleLock singleLock(ExclusiveResource resource) {
+		return new SingleLock(resource, anyLock());
+	}
+
+	private static CompositeLock compositeLock(ExclusiveResource... resources) {
+		return new CompositeLock(List.of(resources),
+			IntStream.range(0, resources.length).mapToObj(__ -> anyLock()).toList());
```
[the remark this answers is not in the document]
marcphilipp: Simplified in b6309e3e1e89b3652bcfe2f28a9fc4a2b94e7a56.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ResourceLockTests.java
```diff
@@ -12,71 +12,135 @@
...
 	@Test
-	void singleLocksAreIncompatibleWithNonNopLocks() {
-		var singleLock = new SingleLock(anyReentrantLock());
+	void readOnlySingleLocksAreIncompatibleWithOtherLocksThatCanPotentiallyCauseDeadlocks() {
+		ExclusiveResource bR = readOnlyResource("b");
+
+		assertCompatible(singleLock(bR), nopLock());
+		assertCompatible(singleLock(bR), singleLock(bR));
+		assertIncompatible(singleLock(bR), singleLock(readWriteResource("b")));
+		assertIncompatible(singleLock(bR), singleLock(readOnlyResource("a")));
```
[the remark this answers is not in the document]
marcphilipp: Done in c205e5eae97e65f7fc4090d7da27af5aea81986d.

platform-tests/src/test/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorServiceTests.java
marcphilipp: Done in 7c654e450e7ff5dd4ebd950498895f190fcaf455.

PR #4025, opened 2024-09-25
Add constructor injection support for ParameterizedTest SPIs

### LINKED ISSUE #4018: Add constructor injection support for `ArgumentsProvider`, `ArgumentConverter`, and `ArgumentsAggregator`
It is possible to provide parameters to the constructor of a test class, and also to a method used by `@ParameterizedTest`/`@MethodSource`.

Request: Be able to provide parameters to the constructor of an `ArgumentsProvider` implementation.

Sample:

```java
@ExtendWith(MyExtension.class) // this extension provides values for type 'Pojo' parameters
public class WithArgumentsProvider {

    @Nested
    public class NestedTest1 {

        @ParameterizedTest
        @ArgumentsSource(MyArgumentsProvider.class)
        public void test1(String a) {
            System.out.println("NestedTest1 - Test1: " + a);
        }

    }

    @Nested
    public class NestedTest2 {

        @ParameterizedTest
        @ArgumentsSource(MyArgumentsProvider.class)
        public void test2(String a) {
            System.out.println("NestedTest2 - Test2: " + a);
        }

    }

    public static class MyArgumentsProvider implements ArgumentsProvider {

        public MyArgumentsProvider(Pojo pojo) {
            // ...
        }

        @Override
            return Stream.of("a", "b", "c").map(Arguments::of);
        }

    }

}
```

Dev hint: https://github.com/junit-team/junit5/blob/e8c92f8cdeafe80ae485da4b4d8bd77f7b33e345/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestExtension.java#L87

marcphilipp:
## Overview

- **Allow ArgumentsProvider implementations to use constructor injection**
- **Allow ArgumentConverter implementations to use constructor injection**
- **Allow ArgumentsAggregator implementations to use constructor injection**
- **Document constructor injection support in release notes**

Resolves #4018.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestSpiInstantiator.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestSpiInstantiator.java
@@ -18,7 +18,7 @@ import java.util.Optional;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.platform.commons.JUnitException;
 import org.junit.platform.commons.PreconditionViolationException;
-import org.junit.platform.commons.support.ModifierSupport;
+import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.commons.util.ReflectionUtils;
 
 /**
@@ -26,8 +26,10 @@ import org.junit.platform.commons.util.ReflectionUtils;
  */
 class ParameterizedTestSpiInstantiator {
 
-	static <T> T instantiate(Class<T> spiClass, Class<? extends T> clazz, ExtensionContext extensionContext) {
-		return extensionContext.getExecutableInvoker().invoke(findConstructor(spiClass, clazz));
+	static <T> T instantiate(Class<T> spiInterface, Class<? extends T> implementationClass,
+			ExtensionContext extensionContext) {
+		return extensionContext.getExecutableInvoker() //
+				.invoke(findConstructor(spiInterface, implementationClass));
 	}
 
 	/**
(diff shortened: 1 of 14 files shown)


junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestSpiInstantiator.java
```diff
@@ -0,0 +1,63 @@
...
+		Optional<Constructor<?>> defaultConstructor = getFirstElement(
+			ReflectionUtils.findConstructors(clazz, it -> it.getParameterCount() == 0));
+		if (defaultConstructor.isPresent()) {
+			return (Constructor<? extends T>) defaultConstructor.get();
+		}
+		if (ModifierSupport.isNotStatic(clazz)) {
+			String message = String.format("The %s [%s] must be either a top-level class or a static nested class",
+				spiClass.getSimpleName(), clazz.getName());
+			throw new JUnitException(message);
+		}
```
sbrannen, line 51: Shouldn't this "precondition" check be placed at the top of this method?

And shouldn't it rather use `ReflectionUtils.isInnerClass(clazz)`?

sbrannen (changes requested):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestSpiInstantiator.java
marcphilipp: You're right! Done in 126ab50a063558e3cb61c2c5e4d7e8524ff89dd4.

sbrannen (approved):
Thanks for making the changes in `ParameterizedTestSpiInstantiator`.

I might propose a different approach to that algorithm, but in the interim I'm approving this PR. 👍

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestSpiInstantiator.java
```diff
@@ -0,0 +1,73 @@
...
+	private static <T, V extends T> Constructor<? extends V> findConstructor(Class<T> spiInterface,
+			Class<V> implementationClass) {
+
+		Preconditions.condition(!ReflectionUtils.isInnerClass(implementationClass),
+			() -> String.format("The %s [%s] must be either a top-level class or a static nested class",
+				spiInterface.getSimpleName(), implementationClass.getName()));
+
+		return findDefaultConstructor(implementationClass) //
+				.orElseGet(() -> findSingleConstructor(spiInterface, implementationClass));
+	}
```
sbrannen, line 51: Since `Class.getDeclaredConstructors()` clones the constructors array **and** makes "child copies" of the constructors, I usually try to avoid repeated calls to `getDeclaredConstructors()` for a single use case. I also like to avoid the use of `Optional` and multiple `Stream`s if it's easily avoidable.

Instead, I typically go with something like this:

```java
	/**
	 * Find the "best" constructor for the supplied implementation class.
	 *
	 * <p>For backward compatibility, it first checks for a single constructor
	 * and returns that. If there are multiple constructors, it checks for a
	 * default constructor which takes precedence over any other constructors.
	 * Otherwise, this method throws an exception stating that it failed to
	 * find a suitable constructor.
	 */
	@SuppressWarnings("unchecked")
	static <T, V extends T> Constructor<? extends V> findConstructor(Class<T> spiInterface,
			Class<V> implementationClass) {

		Preconditions.condition(!ReflectionUtils.isInnerClass(implementationClass),
			() -> String.format("The %s [%s] must be either a top-level class or a static nested class",
				spiInterface.getSimpleName(), implementationClass.getName()));

		Constructor<?>[] constructors = implementationClass.getDeclaredConstructors();

		// Single constructor?
		if (constructors.length == 1) {
			return (Constructor<V>) constructors[0];
		}
		// Find default constructor.
		for (Constructor<?> constructor : constructors) {
			if (constructor.getParameterCount() == 0) {
				return (Constructor<V>) constructor;
			}
		}
		// Otherwise...
		String message = String.format(
			"Failed to find constructor for %s [%s]. "
					+ "Please ensure that a no-argument or a single constructor exists.",
			spiInterface.getSimpleName(), implementationClass.getName());
		throw new JUnitException(message);
	}
```

But now I see you've already merged the PR.

Shall I make those changes on `main`, or...?

marcphilipp: > Shall I make those changes on `main`, or...?

Please do!

sbrannen: Done! 

https://github.com/junit-team/junit5/commit/daea6f12c21ac664016087c1c5949dd3f06da232

PR #4032, opened 2024-09-29
Fix ExtensionContext on instance creation

### LINKED ISSUE #3445: Optionally provide test-scoped `ExtensionContext` for test-specific callbacks in extensions
I am currently writing a small extension which implements `ParameterResolver` to provide a few parameters which need to be shut down after the test. After reading the documentation, my first intuition was to write the extension like this:

```java
final class MyExtension implements ParameterResolver {
    private static final Namespace NAMESPACE = Namespace.create(MyExtension.class);

    @Override
    public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
        return parameterContext.getParameter().getType() == Parameter.class;
    }

    @Override
    public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
        return extensionContext
                .getStore(NAMESPACE)
                .getOrComputeIfAbsent(ParameterWrapper.class)
                .parameter;
    }

    private static final class ParameterWrapper implements CloseableResource {
        private final Parameter parameter;

        private ParameterWrapper() {
            this.parameter = new Parameter();
        }

        @Override
        public void close() {
            parameter.close();
        }
    }
}
```

The problem with that solution is that the `ExtensionContext` is scoped for the whole test class when resolving arguments for the test class constructor, although the test uses `TestInstance.Lifecycle.PER_METHOD`. As a consequence, my example above will use the same parameter for all tests, instead of creating a new one per test. The documentation does not describe a lot of details about the lifecycle of each `ExtensionContext`, therefore I cannot declare it as a bug, but this behavior is very unexpected in my opinion. Since the callbacks are on the level of the individual tests, I expected the `ExtensionContext` to be scoped to the test as well. The same problem applies to `TestInstancePreConstructCallback`. I haven't tested `TestInstanceFactory`, but I guess it is also affected.

The obvious workaround is to also add a `TestInstancePreDestroyCallback` which closes the parameter and removes it from the store. Note that the implementation of that method is not trivial because in contrast to the methods above, this callback receives a child store which is scoped to the test. Unfortunately, this means that calling remove on the given store directly instead of iterating over the parents would not work.

```java
    @Override
    public void preDestroyTestInstance(ExtensionContext context) {
        ParameterWrapper wrapper = null;
        do {
            wrapper = context.getStore(NAMESPACE).remove(ParameterWrapper.class, ParameterWrapper.class);
            context = context.getParent().orElse(null);
        } while (context != null && wrapper == null);
        if (wrapper != null) wrapper.close();
    }
```

## Deliverables

- [ ] Test-scoped `ExtensionContext` for `InvocationInterceptor.interceptTestClassConstructor(…)`
- [ ] Test-scoped `ExtensionContext` for `ParameterResolver` while resolving constructor parameters
- [ ] Test-scoped `ExtensionContext` for `TestInstancePreConstructCallback`
- [ ] Test-scoped `ExtensionContext` for `TestInstancePostProcessor`
- [ ] Test-scoped `ExtensionContext` for `TestInstanceFactory`
- [ ] Consider #3670 in the implementation

**-- or --**

- [ ] Documentation of the limitation in the Javadoc of the interfaces or callbacks.

## Related issues

- #3323
- #3670

JojOatXGME:
## Overview

Resolve #3445. This PR moves the instance creation into the execution context of the test method. This affects the following callbacks:

* `TestInstancePreConstructCallback`
* `TestInstanceFactory`
* `ParameterResolver` (when resolving parameters for the constructor)
* `TestInstancePostProcessor`

Assuming the test is using `TestInstance.Lifecycle.PER_METHOD` (the default), these callbacks will now receive the `ExtensionContext` for the currently executed test method. This has the following effects:

* The callbacks can now access `ExtensionContext.getTestMethod()`
* `ExtensionContext.getTestClass()` and `TestInstanceFactoryContext.getTestClass()` are no-longer interchangeable
* `ExtensionContext.getTestClass()` now returns the (nested) class of the currently executed tests
* Instances of `CloseableResource` add to the store will now be closed at the end of the test execution

If the test is using `TestInstance.Lifecycle.PER_CLASS`, the behavior is unaltered.

Not sure if the change is considered API breaking. The previous behavior was rather unintuitive and not explicitly documented. However, the example at `TestInfoDemo` is no-longer valid and has been changed as part of the PR. I also think there is a good chance that there are extensions in the wild which rely on the previous behavior, since the behavior was in place for multiple years. I would therefore like to know how I should proceed with the PR.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/test/java/example/TestInfoDemo.java
+++ b/documentation/src/test/java/example/TestInfoDemo.java
@@ -24,7 +24,8 @@ import org.junit.jupiter.api.TestInfo;
 class TestInfoDemo {
 
 	TestInfoDemo(TestInfo testInfo) {
-		assertEquals("TestInfo Demo", testInfo.getDisplayName());
+		String displayName = testInfo.getDisplayName();
+		assertTrue(displayName.equals("TEST 1") || displayName.equals("test2()"));
 	}
 
 	@BeforeEach
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
@@ -201,8 +201,7 @@ public abstract class ClassBasedTestDescriptor extends JupiterTestDescriptor {
 			// and store the instance in the ExtensionContext.
 			ClassExtensionContext extensionContext = (ClassExtensionContext) context.getExtensionContext();
 			throwableCollector.execute(() -> {
-				TestInstances testInstances = context.getTestInstancesProvider().getTestInstances(
-					context.getExtensionRegistry(), throwableCollector);
+				TestInstances testInstances = context.getTestInstancesProvider().getTestInstances(context);
 				extensionContext.setTestInstances(testInstances);
 			});
 		}
@@ -273,32 +272,30 @@ public abstract class ClassBasedTestDescriptor extends JupiterTestDescriptor {
 	}
 
 	private TestInstancesProvider testInstancesProvider(JupiterEngineExecutionContext parentExecutionContext,
-			ClassExtensionContext extensionContext) {
+			ClassExtensionContext ourExtensionContext) {
 
-		return (registry, registrar, throwableCollector) -> extensionContext.getTestInstances().orElseGet(
-			() -> instantiateAndPostProcessTestInstance(parentExecutionContext, extensionContext, registry, registrar,
-				throwableCollector));
+		// For Lifecycle.PER_CLASS, ourExtensionContext.getTestInstances() is used to store the instance.
+		// Otherwise, extensionContext.getTestInstances() is always empty and we always create a new instance.
+		return (registry, context) -> ourExtensionContext.getTestInstances().orElseGet(
+			() -> instantiateAndPostProcessTestInstance(parentExecutionContext, registry, context));
 	}
 
 	private TestInstances instantiateAndPostProcessTestInstance(JupiterEngineExecutionContext parentExecutionContext,
-			ExtensionContext extensionContext, ExtensionRegistry registry, ExtensionRegistrar registrar,
-			ThrowableCollector throwableCollector) {
+			ExtensionRegistry registry, JupiterEngineExecutionContext context) {
 
-		TestInstances instances = instantiateTestClass(parentExecutionContext, registry, registrar, extensionContext,
-			throwableCollector);
-		throwableCollector.execute(() -> {
-			invokeTestInstancePostProcessors(instances.getInnermostInstance(), registry, extensionContext);
+		TestInstances instances = instantiateTestClass(parentExecutionContext, registry, context);
+		context.getThrowableCollector().execute(() -> {
+			invokeTestInstancePostProcessors(instances.getInnermostInstance(), registry, context.getExtensionContext());
 			// In addition, we initialize extension registered programmatically from instance fields here
 			// since the best time to do that is immediately following test class instantiation
 			// and post-processing.
-			registrar.initializeExtensions(this.testClass, instances.getInnermostInstance());
+			context.getExtensionRegistry().initializeExtensions(this.testClass, instances.getInnermostInstance());
 		});
 		return instances;
 	}
 
 	protected abstract TestInstances instantiateTestClass(JupiterEngineExecutionContext parentExecutionContext,
-			ExtensionRegistry registry, ExtensionRegistrar registrar, ExtensionContext extensionContext,
-			ThrowableCollector throwableCollector);
+			ExtensionRegistry registry, JupiterEngineExecutionContext context);
 
 	protected TestInstances instantiateTestClass(Optional<TestInstances> outerInstances, ExtensionRegistry registry,
 			ExtensionContext extensionContext) {
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassTestDescriptor.java
@@ -20,16 +20,13 @@ import java.util.Optional;
 import java.util.Set;
 
 import org.apiguardian.api.API;
-import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.TestInstances;
 import org.junit.jupiter.engine.config.JupiterConfiguration;
 import org.junit.jupiter.engine.execution.JupiterEngineExecutionContext;
-import org.junit.jupiter.engine.extension.ExtensionRegistrar;
 import org.junit.jupiter.engine.extension.ExtensionRegistry;
 import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.TestTag;
 import org.junit.platform.engine.UniqueId;
-import org.junit.platform.engine.support.hierarchical.ThrowableCollector;
 
 /**
  * {@link TestDescriptor} for tests based on Java classes.
@@ -74,9 +71,8 @@ public class ClassTestDescriptor extends ClassBasedTestDescriptor {
 
 	@Override
 	protected TestInstances instantiateTestClass(JupiterEngineExecutionContext parentExecutionContext,
-			ExtensionRegistry registry, ExtensionRegistrar registrar, ExtensionContext extensionContext,
-			ThrowableCollector throwableCollector) {
-		return instantiateTestClass(Optional.empty(), registry, extensionContext);
+			ExtensionRegistry registry, JupiterEngineExecutionContext context) {
+		return instantiateTestClass(Optional.empty(), registry, context.getExtensionContext());
 	}
 
 }
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/NestedClassTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/NestedClassTestDescriptor.java
@@ -21,16 +21,13 @@ import java.util.Optional;
 import java.util.Set;
 
 import org.apiguardian.api.API;
-import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.TestInstances;
 import org.junit.jupiter.engine.config.JupiterConfiguration;
 import org.junit.jupiter.engine.execution.JupiterEngineExecutionContext;
-import org.junit.jupiter.engine.extension.ExtensionRegistrar;
 import org.junit.jupiter.engine.extension.ExtensionRegistry;
 import org.junit.platform.engine.TestDescriptor;
 import org.junit.platform.engine.TestTag;
 import org.junit.platform.engine.UniqueId;
-import org.junit.platform.engine.support.hierarchical.ThrowableCollector;
 
 /**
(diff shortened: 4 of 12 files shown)


marcphilipp:
see https://github.com/junit-team/junit5/issues/3445#issuecomment-2387853283

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/EnableTestScopedConstructorContext.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @since 5.12
+ * @see TestInstancePreConstructCallback
+ * @see TestInstancePostProcessor
+ * @see TestInstanceFactory
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@API(status = MAINTAINED, since = "5.12")
```
JojOatXGME, line 74: I wasn't sure which status to use in the `API` annotation. Which status do you suggest?

If we want to us `EXPERIMENTAL` for now, then I should probably also reformulate my note in the Javadoc. (Suggesting to switch to an experimental API for future compatibility seems a bit odd. 😄)

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/EnableTestScopedConstructorContext.java
```diff
@@ -0,0 +1,76 @@
...
+public @interface EnableTestScopedConstructorContext {
```
JojOatXGME, line 75: Feel free to suggest other names. This is the best name that came to mind. I think the name is good enough, at least I like it much more than my other ideas (e.g. `NewTestInstanceConstructionContext`).

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
```diff
@@ -355,24 +368,36 @@ private Object invokeTestInstanceFactory(Optional<Object> outerInstance, Extensi
 	}
 
 	private Object invokeTestClassConstructor(Optional<Object> outerInstance, ExtensionRegistry registry,
-			ExtensionContext extensionContext) {
+			ExtensionContext extensionContext, ExtensionContext ourExtensionContext) {
 
 		Constructor<?> constructor = ReflectionUtils.getDeclaredConstructor(this.testClass);
 		return executableInvoker.invoke(constructor, outerInstance, extensionContext, registry,
 			InvocationInterceptor::interceptTestClassConstructor);
```
JojOatXGME, line 375: Calls to `ParameterResolver` are not yet looking at the new annotation. Shall calls to `ParameterResolver` also consider the annotation?

I just wanted to confirm that I should adjust this method as well. I was a bit reluctant to change `InterceptingExecutableInvoker` for some reason. For now, I wrote the documentation matching the current behavior, I will change the documentation together with the implementation.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
```diff
@@ -295,27 +300,35 @@ private TestInstances instantiateAndPostProcessTestInstance(JupiterEngineExecuti
...
-	private Object invokeTestInstanceFactory(Optional<Object> outerInstance, ExtensionContext extensionContext) {
+	private Object invokeTestInstanceFactory(Optional<Object> outerInstance, ExtensionContext extensionContext,
+			ExtensionContext ourExtensionContext) {
 		Object instance;
 
 		try {
-			instance = this.testInstanceFactory.createTestInstance(
-				new DefaultTestInstanceFactoryContext(this.testClass, outerInstance), extensionContext);
+			if (AnnotationUtils.isAnnotated(this.testInstanceFactory.getClass(),
+				EnableTestScopedConstructorContext.class)) {
```
JojOatXGME, line 324: I am a little bit concerned about performance with all these calls to `AnnotationUtils.isAnnotated`. Do you think this could be an issue?

JojOatXGME (commented):
I introduced the annotation `@EnableTestScopedConstructorContext` to enable the new behavior for `TestInstancePreConstructCallback`, `TestInstancePostProcessor` and `TestInstanceFactory`. I have few remaining questions, which I have added as review comments.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/EnableTestScopedConstructorContext.java
JojOatXGME: I just realized that you have already suggested `@MethodLevelExtensionContextAware`. I don't have a strong preference. Your suggestion is a bit shorter.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
marcphilipp: It does seem unnecessary to check it multiple times (in different places). I'll take care of that in a follow-up PR.

JojOatXGME: Ok, if you have an idea how to solve this, then it is good. My best idea would be to add a cache to the extension registry.

Just want to note that using a marker interface instead of an annotation would also alleviate the problem, as `instanceof` is rather fast as far as I know. However, I agree that a marker interface is probably not the best solution on extensions, which may already be rather heavy on interfaces.

Not sure if that is the right place to discuss it, but I also just had another idea. We could introduce a common method with a default implementation to all the affected extension interfaces.

```java
default boolean supportsTestScopedConstructorContext() {
    // or isMethodLevelExtensionContextAware()
    return false;
}
```

Not entirely sure if I would like such a method. It seems like a very good solution from the technical perspective, but it might feel a bit clunky.

marcphilipp: > Ok, if you have an idea how to solve this, then it is good. My best idea would be to add a cache to the extension registry, which is probably fine.

Yes, something along those lines. But I'd like to get feel for it myself before I can make a concrete recommendation. In any case, I think this can be addressed separately.

> Just want to note that using a marker interface instead of an annotation would also alleviate the problem, as `instanceof` is rather fast as far as I know. However, I agree that a marker interface is probably not the best solution on extensions, which may already be rather heavy on interfaces.
> 
> Not sure if that is the right place to discuss it, but I also just had another idea. We could introduce a common method with a default implementation to all the affected extension interfaces.

Thanks for sharing your ideas! I think both would be equivalent but I still think the annotation is the best option for a "temporary" solution like this because it doesn't muddle the extension interfaces even if it's a little more involved implementation-wise.

documentation/src/test/java/example/TestInfoDemo.java
```diff
@@ -24,8 +24,7 @@
 class TestInfoDemo {
 
 	TestInfoDemo(TestInfo testInfo) {
-		String displayName = testInfo.getDisplayName();
-		assertTrue(displayName.equals("TEST 1") || displayName.equals("test2()"));
+		assertEquals("TestInfo Demo", testInfo.getDisplayName());
```
JojOatXGME, line 27: My last changes (6cda95fcfab7d6b0ca75e8a8f7dc77c246899dbe) also reverted the behavior of `TestInfo`, as the related extension doesn't have our new annotation. I therefore also reverted the `TestInfo`-related changes in `ParameterResolverTests` and `ParameterizedTestIntegrationTests`.

`TestInfo` and `TestReporter` could also benefit from the test-specific information, but these specific cases are probably not very important right now.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
JojOatXGME: You are right, I added `InvocationInterceptor` to the documentation and to the deliverables in the issue. I also addressed the implementation for `InvocationInterceptor`, but there are currently no tests which verify that the correct `ExtensionContext` is used. `ParameterResolver` is only covered indirectly over `TestInfo`.

documentation/src/test/java/example/TestInfoDemo.java
marcphilipp: > `TestInfo` and `TestReporter` could also benefit from the test-specific information, but these specific cases are probably not very important right now.

I've made a note to look at the core extensions after this PR is merged.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
marcphilipp: > I also addressed the implementation for `InvocationInterceptor`, but there are currently no tests which verify that the correct `ExtensionContext` is used.

Noted. I'll create an umbrella issue to sort out the remaining issues when we're done with this PR.

marcphilipp (approved):

marcphilipp:
I'm going ahead with merging this PR. I will address the remaining work in #4057.

marcphilipp:
Thanks a lot, @JojOatXGME! :+1:

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/EnableTestScopedConstructorContext.java
JojOatXGME: I just wanted to note, that after [the idea of a common method](https://github.com/junit-team/junit5/pull/4032#discussion_r1789952664) has lingered in my head over the last day, I have started to prefer it. The idea would be to add a default method to the empty `Extension` interface. At the end, I don't think any of my reasons is that significant, and I am totally happy with the current solution as well. Anyway, here are the reasons why I started to like the idea of a default method, although it is more clunky to use.

* Improved discoverability as the method is shown by the IDE when you look for methods to override (e.g. <kbd>Ctrl</kbd>+<kbd>I</kbd> in IntelliJ). However, the Javadoc also makes it rather easy to find already.
* Extensions can use arbitrary logic for the flag. For example, a wrapper or proxy which delegates all method calls to another extension would be able to preserve the flag from the delegate. However, it is probably a very special use case and not that relevant.
* Less likely to impact the performance without having to introduce any cache.
* Maybe more consistent, as I think there are currently no other annotations used in the extension logic.
* A default method could technically be removed again without breaking binary compatibility (ABI). However, I assume removing it from the API again is not really an option anyway, so it is not really relevant.

Feel free to pick whatever you prefer. I just wanted to communicate my thoughts.

JojOatXGME:
Also thanks from my site. I am happy to remove my workaround in my projects once it is released. 😄 Feel free to reach out if you have any questions, otherwise I will just leave it to you now.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/EnableTestScopedConstructorContext.java
marcphilipp: Those are good points! To get a better feel for it, I've given it a try in #4062. I may be subject to the [IKEA effect](https://en.wikipedia.org/wiki/IKEA_effect) now, but it like the solution better for the reasons you've given above.

PR #4049, opened 2024-10-06
Support `void` and `Void` class lookups in `ReflectionUtils`

### LINKED ISSUE #4048: Support `void` class lookups in `ReflectionUtils`
Running following parametrized test in JUnit 5.11
```java
@ValueSource(strings = {
    "void", "boolean", "byte", "char", "short", "int", "long", "float", "double" })
@ParameterizedTest
void isPrimitive(Class<?> candidate) {
    assertTrue(candidate.isPrimitive());
}
```
only makes the "void" case fail.
```
org.junit.jupiter.api.extension.ParameterResolutionException:
    Error converting parameter at index 0: Failed to convert String "void" to type java.lang.Class
```
It looks like conversion of "void" to Class isn't supported.
Both `classNameToTypeMap` and `primitiveToWrapperMap` of `org.junit.platform.commons.util.ReflectionUtils` are lacking `void.class` and `Void.class`, but maybe this is by intention.

bjmi:
Fix #4048.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
@@ -225,7 +225,6 @@ public final class ReflectionUtils {
 			Long[].class,
 			Float[].class,
 			Double[].class,
-			Void[].class,
 			String[].class,
 
 			Boolean[][].class,
@@ -236,7 +235,6 @@ public final class ReflectionUtils {
 			Long[][].class,
 			Float[][].class,
 			Double[][].class,
-			Void[][].class,
 			String[][].class
 		);
 		// @formatter:on
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
@@ -186,6 +186,7 @@ public final class ReflectionUtils {
 			long.class,
 			float.class,
 			double.class,
+			void.class,
 
 			boolean[].class,
 			byte[].class,
@@ -213,6 +214,7 @@ public final class ReflectionUtils {
 			Long.class,
 			Float.class,
 			Double.class,
+			Void.class,
 			String.class,
 
 			Boolean[].class,
@@ -223,6 +225,7 @@ public final class ReflectionUtils {
 			Long[].class,
 			Float[].class,
 			Double[].class,
+			Void[].class,
 			String[].class,
 
 			Boolean[][].class,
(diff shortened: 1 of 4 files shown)


marcphilipp:
Could you please add an "integration test" demonstrating the scenario from #4048 is now supported?

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -223,6 +225,7 @@ public enum HierarchyTraversalMode {
 			Long[].class,
 			Float[].class,
 			Double[].class,
+			Void[].class,
```
sbrannen, line 228:
```suggestion
```

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -233,6 +236,7 @@ public enum HierarchyTraversalMode {
 			Long[][].class,
 			Float[][].class,
 			Double[][].class,
+			Void[][].class,
```
sbrannen, line 239:
```suggestion
```

sbrannen (changes requested):
Thanks for the PR!

I would not consider `Void[]` and `Void[][]` "common types", so please remove those from the `classNameToTypeMap` cache.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
bjmi: Applied.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
bjmi: Applied.

marcphilipp (changes requested):
> > Could you please add an "integration test" demonstrating the scenario from #4048 is now supported?
> 
> This is now implicitly tested by `DefaultArgumentConverterTests.convertsStringToClass()`. Should the newly introduced `DefaultArgumentConverterTests.IntegrationTests` still be retained?

Thanks! No, I don't think we need `DefaultArgumentConverterTests.IntegrationTests`.

Please add an entry to the 5.12.0-M1 release notes.

bjmi:
> Thanks! No, I don't think we need `DefaultArgumentConverterTests.IntegrationTests`.

Removed.
> Please add an entry to the 5.12.0-M1 release notes.

As I'm not a native speaker feel free to change the text in the release notes.

marcphilipp (approved):

marcphilipp:
Thanks, @bjmi! :+1:

PR #4120, opened 2024-11-12
Implement include and exclude filtering when calling registerAutoDetectedExtensions in MutableExtensionRegistry

YongGoose:
fix #3717 

## Overview

This PR enables the activation of **specific global extensions** in JUnit Jupiter.

I've already worked on this in my personal repository, and since there’s a large amount of code changes, I decided to split it into two PRs.

https://github.com/YongGoose/junit5/commit/a2ef07f02a0fd7e4980931bd2ea2f390f53962fa
I plan to proceed in the following two ways.

1. Add an include method to ClassNamePatternFilterUtils.
2. Implement include and exclude filtering when calling registerAutoDetectedExtensions in MutableExtensionRegistry.

---

I hereby agree to the terms of the [JUnit Contributor License Agreement](https://github.com/junit-team/junit5/blob/002a0052926ddee57cf90580fa49bc37e5a72427/CONTRIBUTING.md#junit-contributor-license-agreement).

---

### Definition of Done

- [ ] There are no TODOs left in the code
- [ ] Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.commons/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [ ] [Coding conventions](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [ ] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling
- [ ] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [ ] Change is documented in the [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)

PR #4135, opened 2024-11-18
Support parallelization in junit-vintage-engine

### LINKED ISSUE #2229: Support some level of parallelization in junit-vintage-engine
## Goal
Add support for some level of parallelization inside the Junit Vintage engine.
Ideally the level of configuration should be similar to maven surefire parallel options.

## Why
When including junit 5 tests in a project containing junit4 tests the natural path is to use junit 5 platform with the jupiter and vintage engines.

However this causes problem with parallelization. 
In a project with Junit4 tests only maven surefire level parallelization works fine. (See https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html).
For Junit5 tests the jupiter engine [parallel execution works great](https://junit.org/junit5/docs/snapshot/user-guide/#writing-tests-parallel-execution). 

However when a project contains a mix of Junit4 and Junit 5 there is no great way to configure parallelism:
* Surefire parallel does not work with Junit Platform Provider so it's not an option
* Surefire forkCount or gradle maxParallelForks can be used but:
  * It's not available at all when using  [tycho surefire](https://www.eclipse.org/tycho/sitedocs/tycho-surefire-plugin/plugin-info.html)
  * It introduces some memory and execution time overhead compared to surefire parallel or junit 5 parallel.
  * It provides less control than Junit 5 parallel execution options.
* Combining fork + junit 5 parallel execution leads to more parallel executions than expected  when executing junit 5 tests because parallelization is done at two level.

Another option is to configure two distinct test tasks one for junit 4 using surefire parallel and one for junit 5 but it kind of defeats the point of junit platform.

## Deliverables

- [x] Not sure

YongGoose:
## Overview
Resolves #2229 

This PR implements parallel execution support for the JUnit Vintage engine, addressing issue #2229.
- Implemented `concurrent execution` of test descriptors when parallel execution is enabled.
- Updated the JUnit Vintage engine to read parallel configuration parameters. (`planned`)

## Open Question

I haven’t yet worked on the part that handles reading configuration parameters, such as whether `parallel execution` should be **enabled** or the **number of threads** to be used. 
Is there a specific configuration format or approach you would recommend for this? 

---

---

### Definition of Done


junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +88,64 @@ public void execute(ExecutionRequest request) {
...
+
+	private void executeInParallel(VintageEngineDescriptor engineDescriptor,
+			EngineExecutionListener engineExecutionListener) {
+		ExecutorService executorService = Executors.newFixedThreadPool(getThreadPoolSize());
+		RunnerExecutor runnerExecutor = new RunnerExecutor(engineExecutionListener);
+
+		List<TestDescriptor> children = new CopyOnWriteArrayList<>(engineDescriptor.getModifiableChildren());
+
+		CompletableFuture<Void> allOf = CompletableFuture.allOf(
+			children.stream().map(descriptor -> CompletableFuture.runAsync(() -> {
```
marcphilipp, line 109: We should also remove the descriptors from the original list to allow them to be garbage collected once they're done. I think we should be able to use an iterator on `engineDescriptor.getModifiableChildren()`, convert each into a future, call `iterator.remove` and then wait for the combined future.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +88,64 @@ public void execute(ExecutionRequest request) {
...
+				try {
+					runnerExecutor.execute(testDescriptor);
+				}
+				catch (Exception e) {
+					engineExecutionListener.executionSkipped(testDescriptor, e.getMessage());
+				}
+			}, executorService)).toArray(CompletableFuture[]::new));
+
+		try {
+			allOf.get(SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
marcphilipp, line 120: Doesn't this limit the overall test execution duration to 60 seconds? I don't think we can make that assumption. Instead we need to wait without a timeout.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +88,64 @@ public void execute(ExecutionRequest request) {
...
+		}
+		catch (Exception e) {
+			logger.warn(e, () -> "Timeout or interruption while waiting for parallel test execution to finish");
+			Thread.currentThread().interrupt();
```
marcphilipp, line 124: We should only do this for `InterruptedException`. `ExecutionExceptions` should be unwrapped and rethrown using `ExceptionUtils.throwAsUncheckedException`.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +88,64 @@ public void execute(ExecutionRequest request) {
...
+		}
+		finally {
+			executorService.shutdown();
```
marcphilipp, line 127: Additionally, we should call `awaitTermination` with a timeout here.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -37,6 +45,11 @@
 @API(status = INTERNAL, since = "4.12")
 public final class VintageTestEngine implements TestEngine {
 
+	private static final Logger logger = LoggerFactory.getLogger(VintageTestEngine.class);
+
+	private static final int DEFAULT_THREAD_POOL_SIZE = 10;
```
marcphilipp, line 50: I think this should default to `Runtime.availableProcessors()`

marcphilipp (commented):
Thanks for submitting the draft! I left a bunch of comments. :slightly_smiling_face:

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +88,64 @@ public void execute(ExecutionRequest request) {
...
+			executorService.shutdown();
+		}
+	}
+
+	private void executeSequentially(VintageEngineDescriptor engineDescriptor,
+			EngineExecutionListener engineExecutionListener) {
 		RunnerExecutor runnerExecutor = new RunnerExecutor(engineExecutionListener);
-		for (Iterator<TestDescriptor> iterator = engineDescriptor.getModifiableChildren().iterator(); iterator.hasNext();) {
+		Iterator<TestDescriptor> iterator = engineDescriptor.getModifiableChildren().iterator();
+		while (iterator.hasNext()) {
```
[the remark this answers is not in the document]
YongGoose: I forgot to revert the changes after making modifications. 😅
I will restore them to their original state now.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +88,64 @@ public void execute(ExecutionRequest request) {
...
+			executeSequentially(engineDescriptor, engineExecutionListener);
+		}
+	}
+
+	private void executeInParallel(VintageEngineDescriptor engineDescriptor,
+			EngineExecutionListener engineExecutionListener) {
+		ExecutorService executorService = Executors.newFixedThreadPool(getThreadPoolSize());
+		RunnerExecutor runnerExecutor = new RunnerExecutor(engineExecutionListener);
+
+		List<TestDescriptor> children = new CopyOnWriteArrayList<>(engineDescriptor.getModifiableChildren());
```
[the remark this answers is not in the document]
YongGoose: I agree. I will make the change :)

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
YongGoose: I've implemented it like this. 

```
finally {
	try {
		if (!executorService.awaitTermination(SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
			logger.warn(() -> "Executor service did not terminate within the specified timeout");
			executorService.shutdownNow();
		}
	}
	catch (InterruptedException e) {
		logger.warn(e, () -> "Interruption while waiting for executor service to shut down");
		Thread.currentThread().interrupt();
	}
}
```

However, one concern I have is that declaring a `try-catch` inside the `finally` block might reduce readability."

YongGoose: I've implemented it like this. 

```
finally {
	try {
		if (!executorService.awaitTermination(SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
			logger.warn(() -> "Executor service did not terminate within the specified timeout");
			executorService.shutdownNow();
		}
	}
	catch (InterruptedException e) {
		logger.warn(e, () -> "Interruption while waiting for executor service to shut down");
		Thread.currentThread().interrupt();
	}
}
```

However, one concern I have is that declaring a `try-catch` inside the `finally` block might reduce readability.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
YongGoose: I didn't think the test could be longer than 60 seconds... I will make the change!

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
YongGoose: That's a good idea. I'll make the changes as you suggested!

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +90,81 @@ public void execute(ExecutionRequest request) {
...
+				executorService.shutdown();
+				if (!executorService.awaitTermination(SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+					logger.warn(() -> "Executor service did not terminate within the specified timeout");
+					executorService.shutdownNow();
+				}
+			}
+			catch (InterruptedException e) {
+				logger.warn(e, () -> "Interruption while waiting for executor service to shut down");
+				Thread.currentThread().interrupt();
+			}
```
marcphilipp, line 147: Can we extract this into a separate method to improve readability?

YongGoose:
@marcphilipp 

If the parallelization work is completed, I would appreciate it if you could take a look at the Open Question in the PR description 😀

marcphilipp:
@YongGoose The last two weeks were very busy. I should be able to get to your PRs early next week, though. Cheers!

marcphilipp:
> I haven’t yet worked on the part that handles reading configuration parameters, such as whether `parallel execution` should be **enabled** or the **number of threads** to be used.
> Is there a specific configuration format or approach you would recommend for this?

I think it should be similar to the [configuration parameters in Jupiter](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution-config-properties). Thus, I'd suggest we use `junit.vintage.execution.parallel.enabled` and `junit.vintage.execution.parallel.pool-size`. WDYT?

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +90,85 @@ public void execute(ExecutionRequest request) {
...
+		List<CompletableFuture<Void>> futures = new ArrayList<>();
+		for (Iterator<TestDescriptor> iterator = engineDescriptor.getModifiableChildren().iterator(); iterator.hasNext();) {
+			TestDescriptor descriptor = iterator.next();
+			CompletableFuture<Void> future = CompletableFuture.runAsync(() -> {
+				RunnerTestDescriptor testDescriptor = (RunnerTestDescriptor) descriptor;
+				try {
+					runnerExecutor.execute(testDescriptor);
+				}
+				catch (Exception e) {
+					engineExecutionListener.executionSkipped(testDescriptor, e.getMessage());
```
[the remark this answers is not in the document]
YongGoose: I wrote it because I believe that during execution, an exception occurred due to a failure, causing the test to be skipped.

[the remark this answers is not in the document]
marcphilipp: Looking at `RunnerExecutor` it should "never" throw any exceptions (except for `OutOfMemoryError`) so I don't see how that's possible. Could you please elaborate?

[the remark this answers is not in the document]
YongGoose: Upon reviewing the execute method of the RunnerExecutor class, I can confirm that it indeed handles exceptions internally. The method uses a try-catch block to catch any Throwable, which encompasses possible exceptions and errors.
https://github.com/junit-team/junit5/blob/main/junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/RunnerExecutor.java#L44

Therefore, I plan to proceed by removing the try-catch block. WDYT?

Thank you for the constructive feedback.🙂

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -75,11 +90,85 @@ public void execute(ExecutionRequest request) {
...
+	private void shutdownExecutorService(ExecutorService executorService) {
+		try {
+			executorService.shutdown();
+			if (!executorService.awaitTermination(SHUTDOWN_TIMEOUT_SECONDS, TimeUnit.SECONDS)) {
+				logger.warn(() -> "Executor service did not terminate within the specified timeout");
+				executorService.shutdownNow();
+			}
+		}
+		catch (InterruptedException e) {
+			logger.warn(e, () -> "Interruption while waiting for executor service to shut down");
```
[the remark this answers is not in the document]
YongGoose: I made this change in https://github.com/junit-team/junit5/pull/4135/commits/dfeb1cd97ca448839f0d7f9dfa78115dafc0ae37

YongGoose:
> I think it should be similar to the [configuration parameters in Jupiter](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parallel-execution-config-properties). Thus, I'd suggest we use junit.vintage.execution.parallel.enabled and junit.vintage.execution.parallel.pool-size. WDYT?

That's a great idea! I made this change in https://github.com/junit-team/junit5/pull/4135/commits/e0312cc21aaf992cf268573b832bea3c627619fa

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -37,6 +47,13 @@
 @API(status = INTERNAL, since = "4.12")
 public final class VintageTestEngine implements TestEngine {
 
+	private static final Logger logger = LoggerFactory.getLogger(VintageTestEngine.class);
+
+	private static final int DEFAULT_THREAD_POOL_SIZE = Runtime.getRuntime().availableProcessors();
+	private static final int SHUTDOWN_TIMEOUT_SECONDS = 30;
+	private static final String PARALLEL_EXECUTION_ENABLED = "junit.jupiter.execution.parallel.enabled";
```
marcphilipp, line 54:
```suggestion
	private static final String PARALLEL_EXECUTION_ENABLED = "junit.vintage.execution.parallel.enabled";
```

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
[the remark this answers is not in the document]
marcphilipp: Let's remove it :+1:

marcphilipp (changes requested):
As a next step, could you please add some tests?

YongGoose:
While implementing the test, I came across a question and decided to leave a comment :)

In this test case, the return value of `VintageEngineDescriptor.getModifiableChildren` is the `SuccessfulParallelTestCase` class.

- `VintageTestEngine`
``` java
private boolean executeInParallel(VintageEngineDescriptor engineDescriptor,
		EngineExecutionListener engineExecutionListener, ExecutionRequest request) {
	ExecutorService executorService = Executors.newFixedThreadPool(getThreadPoolSize(request));
	RunnerExecutor runnerExecutor = new RunnerExecutor(engineExecutionListener);

	List<CompletableFuture<Void>> futures = new ArrayList<>();
	for (Iterator<TestDescriptor> iterator = engineDescriptor.getModifiableChildren().iterator(); iterator.hasNext();) {
		TestDescriptor descriptor = iterator.next();
		CompletableFuture<Void> future = CompletableFuture.runAsync(() -> {
			runnerExecutor.execute((RunnerTestDescriptor) descriptor);
		}, executorService);

		futures.add(future);
		iterator.remove();
	}
```

- `TestCode`
``` java
@Test
void successfulParallelTest(TestReporter reporter) {
	var events = executeInParallelSuccessfully(3, SuccessfulParallelTestCase.class).list();

	var startedTimestamps = getTimestampsFor(events, event(test(), started()));
	var finishedTimestamps = getTimestampsFor(events, event(test(), finishedSuccessfully()));

	reporter.publishEntry("startedTimestamps", startedTimestamps.toString());
	reporter.publishEntry("finishedTimestamps", finishedTimestamps.toString());

	assertThat(startedTimestamps).hasSize(3);
	assertThat(finishedTimestamps).hasSize(3);
}
```

- `SuccessfulParallelTestCase`
``` java
@RunWith(Enclosed.class)
public class JUnit4ParallelTestCase {

	public static class SuccessfulParallelTestCase {
		static AtomicInteger sharedResource;
		static CountDownLatch countDownLatch;

		@BeforeClass
		public static void initialize() {
			sharedResource = new AtomicInteger();
			countDownLatch = new CountDownLatch(3);
		}

		@Test
		public void firstTest() throws Exception {
			incrementAndBlock(sharedResource, countDownLatch);
		}

		@Test
		public void secondTest() throws Exception {
			incrementAndBlock(sharedResource, countDownLatch);
		}

		@Test
		public void thirdTest() throws Exception {
			incrementAndBlock(sharedResource, countDownLatch);
		}
	}
    ...
```

image


### Question
- If the TestDescriptor has child TestDescriptors as in the case above, could it be improved more efficiently through `recursion`?

> If a test class contains a large number of test methods, I expect that recursion could make the process more efficient.

marcphilipp:
> ### Question
> 
>     * If the TestDescriptor has child TestDescriptors as in the case above, could it be improved more efficiently through `recursion`?
> 
> 
> > If a test class contains a large number of test methods, I expect that recursion could make the process more efficient.

Do you mean that running the children of a `RunnerTestDescriptor` _concurrently_ instead of just the top-level `RunnerTestDescriptors` could speed up the overall execution? If so, I agree with that. However, that's not generally possible as the `Runner` orchestrates the execution of its children, not the `VintageTestEngine`.

JUnit 4 actually has a feature that allows running methods in parallel. It works for all subclasses of `ParentRunner` by [setting a `RunnerScheduler`](https://github.com/junit-team/junit4/blob/df00d5eced3a7737b88de0f6f9e3673f0cf88f88/src/main/java/org/junit/experimental/ParallelComputer.java#L32-L52). We could use that to inject our own scheduler that uses the same `ExecutorService`. However, I think we should do that in a separate PR and focus on running test classes in parallel in this PR. Thus, you'll have to adjust the tests you've added to not check that methods can be run in parallel but classes.

YongGoose:
> Do you mean that running the children of a `RunnerTestDescriptor` _concurrently_ instead of just the top-level `RunnerTestDescriptors` could speed up the overall execution?

Exactly!

> JUnit 4 actually has a feature that allows running methods in parallel. It works for all subclasses of ParentRunner by [setting a RunnerScheduler](https://github.com/junit-team/junit4/blob/df00d5eced3a7737b88de0f6f9e3673f0cf88f88/src/main/java/org/junit/experimental/ParallelComputer.java#L32-L52). We could use that to inject our own scheduler that uses the same ExecutorService. However, I think we should do that in a separate PR and focus on running test classes in parallel in this PR.

I'll work on it in the next PR along with the issue once this PR is complete!

> Thus, you'll have to adjust the tests you've added to not check that methods can be run in parallel but classes.

I was actually considering this issue, so thank you for the detailed suggestion!👍🏻
I'll proceed with the approach you recommended.

YongGoose:
> Thus, you'll have to adjust the tests you've added to not check that methods can be run in parallel but classes.

I implemented it by extracting and comparing thread names when executing test classes. https://github.com/junit-team/junit5/commit/b196388d6d840c377d616eb1005fa4058a239a29

I understand that in `JUnit-Jupiter`, threads can be recorded and verified using `ThreadReporter`. 
Therefore, I would like to `modularize` the code for recording threads. 
- If you have any suggestions on how to achieve this, please let me know :)

marcphilipp:
Thanks! I don't think we can reuse that here since we have to use JUnit 3/4 primitives and can't use Jupiter extensions. I'll be out for about a week for Christmas but will suggest something when I'm back.

YongGoose:
> Thanks! I don't think we can reuse that here since we have to use JUnit 3/4 primitives and can't use Jupiter extensions. I'll be out for about a week for Christmas but will suggest something when I'm back.

Have a good Christmas :)

marcphilipp:
I pushed a commit that does the following:

- Add an assertion that all start timestamps are <= all finish timestamps to verify test classes are executed in parallel
- Remove the test for failures because I didn't think it provided value

WDYT?

junit-vintage-engine/src/test/java/org/junit/vintage/engine/execution/ParallelExecutionIntegrationTests.java
```diff
@@ -0,0 +1,107 @@
...
+import org.junit.platform.testkit.engine.EngineTestKit;
+import org.junit.platform.testkit.engine.Event;
+import org.junit.platform.testkit.engine.Events;
+import org.junit.vintage.engine.VintageTestEngine;
+import org.junit.vintage.engine.samples.junit4.JUnit4ParallelTestCase.SecondTestCase;
+
+class ParallelExecutionIntegrationTests {
+
+	private static final String PARALLEL_EXECUTION_ENABLED = "junit.vintage.execution.parallel.enabled";
+	private static final String PARALLEL_POOL_SIZE = "junit.vintage.execution.parallel.pool-size";
```
marcphilipp, line 46: I think we should make the two constants on `VintageTestEngine` public, add Javadoc and `@API` annotations, and reference them here.

marcphilipp (changes requested):
Please also add an entry to the release notes and a section to the User Guide (as a new section in the "Migrating from JUnit 4" chapter)

YongGoose:
> Add an assertion that all start timestamps are <= all finish timestamps to verify test classes are executed in parallel
> Remove the test for failures because I didn't think it provided value
> 
> WDYT?

I think using timestamps to verify parallel execution is a great idea.

The code looks much cleaner now!
Thank you🙂

YongGoose:
> I pushed a commit that does the following:
> 
> * Add an assertion that all start timestamps are <= all finish timestamps to verify test classes are executed in parallel
> * Remove the test for failures because I didn't think it provided value
> 
> WDYT?

Sounds like a great idea.
The code looks much cleaner now!!

Thank you🙂

junit-vintage-engine/src/test/java/org/junit/vintage/engine/execution/ParallelExecutionIntegrationTests.java
YongGoose: Done in this commit! https://github.com/junit-team/junit5/pull/4135/commits/fc7b9c063fa67e1f6764bab1eb8f7cc62c2b8feb

YongGoose:
> Please also add an entry to the release notes and a section to the User Guide (as a new section in the "Migrating from JUnit 4" chapter)

I hope I did it well this time...

---

I have a question while writing the release notes.  

Shouldn't the details of this PR also be included in the release notes?  
Looking at the issue, it seems like it will be released in 5.12 M1, but I noticed it's not mentioned in the release notes, so I wanted to bring it up!
- https://github.com/junit-team/junit5/pull/4120

YongGoose:
@marcphilipp 

Additionally, once this PR is merged, I will create an issue regarding that comment and start working on it!
- https://github.com/junit-team/junit5/pull/4135#issuecomment-2545714327

marcphilipp:
> Shouldn't the details of this PR also be included in the release notes? Looking at the issue, it seems like it will be released in 5.12 M1, but I noticed it's not mentioned in the release notes, so I wanted to bring it up!
> - https://github.com/junit-team/junit5/pull/4120
>
> If it needs to be added, I will either create a new PR or include it in this one!

Good catch! Indeed, that was an oversight on my part which I've now rectified in 4f926ed576b0290cd739925897ce658174ca5169.

junit-vintage-engine/src/test/java/org/junit/vintage/engine/execution/ParallelExecutionIntegrationTests.java
marcphilipp: Thanks! Good point about `VintageTestEngine` being `INTERNAL`. Let's follow Jupiter's example and move them to a new `Constants` class in the same package and make them `EXPERIMENTAL` instead.

marcphilipp (changes requested):

junit-vintage-engine/src/test/java/org/junit/vintage/engine/execution/ParallelExecutionIntegrationTests.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4135/commits/80c882c645095963037e1af1376f5d2c4b141371 !🙂

marcphilipp:
> > Please also add an entry to the release notes and a section to the User Guide (as a new section in the "Migrating from JUnit 4" chapter)
> 
> I hope I did it well this time...

Thanks, you did! I polished your work a bit anyway because we're quite particular. Don't worry about it! :slightly_smiling_face:

marcphilipp (approved):

PR #4153, opened 2024-11-25
Enable auto-flushing of output to fix `testfeed` details mode

marcphilipp:
## Overview

Prior to this PR, `ConsoleLauncher` created `PrintWriters` that wouldn't flush automatically when `println` is called which stopped the `testfeed` details mode from being useful. Moreover, it didn't initialized them with the right `Charsets`. Now, we rely on Picocli's initialization again which does both correctly.

---

### Definition of Done

### DIFF
--- a/junit-platform-console/src/main/java/org/junit/platform/console/ConsoleLauncher.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/ConsoleLauncher.java
@@ -30,38 +30,23 @@ import org.junit.platform.console.tasks.ConsoleTestExecutor;
 public class ConsoleLauncher {
 
 	public static void main(String... args) {
-		CommandResult<?> result = run(null, null, args);
+		CommandResult<?> result = newCommandFacade().run(args);
 		System.exit(result.getExitCode());
 	}
 
 	@API(status = INTERNAL, since = "1.0")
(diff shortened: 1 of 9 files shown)


junit-platform-console/src/main/java/org/junit/platform/console/ConsoleLauncher.java
```diff
@@ -30,9 +30,7 @@
 public class ConsoleLauncher {
 
 	public static void main(String... args) {
-		PrintWriter out = new PrintWriter(System.out);
-		PrintWriter err = new PrintWriter(System.err);
```
sormuras, line 34: Why not "just" use `new PrintWriter(System.out|.err, true);` here?

marcphilipp: Because theoretically, we'd also have to determine the correct encoding (which Picocli already does using the `sun.stdout.encoding` and `sun.stderr.encoding` system properties).

sormuras (approved):
`record OutputStreamConfig(PrintWriter out, PrintWriter err) { ... }` 🤓

marcphilipp:
It crossed my mind, too. 🙂 But then we couldn't encapsulate the fields.

PR #4159, opened 2024-11-27
Replace Bartholdy with custom Process{Builder} wrappers

marcphilipp:
## Overview

To be more flexible with making changes to the way we run processes from integration tests.


sormuras (approved):
Looks good and green to me.

PR #4180, opened 2024-12-06
Escape whitespace in XML attribute values

### LINKED ISSUE #4174: Legacy XML reports should preserve new line characters in message attributes
Hello!

After switching test execution from using gradle with junit platform to console launcher I noticed that 
value of message attribute in JUnit xml report loosing new line characters.
e.g.
Using gradle xml :
```xml
<error message="line 1&#10; line 2&#10; line 3&#10;"...
```
Using Console launcher xml:
```xml
<error message="
line 1
line 2
line 3
"
```

## Steps to reproduce

1. Run tests using gradle with junit platform. Throw multiline exception in test.
2. Run tests using console launcher, throw multiline exception in test.

## Context

`org.junit.platform:junit-platform-console-standalone:1.11.3`
`org.junit.jupiter:junit-jupiter-engine:5.11.3`
`JRE 23`

As I understand console launcher uses new open/xml report generator and gradle platform launcher uses legacy report generator, right?

marcphilipp:
While XML attribute values may contain whitespace such as line breaks,
the XML spec [1] dictates that XML processors must replace them with
spaces which causes downstream tools to misrepresent the original value.

[1] https://www.w3.org/TR/xml/#AVNormalize

Resolves #4174.


marcphilipp:
This did not have the intended outcome. Replacing whitespace with XML character references before passing them to `XMLStreamWriter#writeAttribute` caused them to be written as `&amp;10;` rather than `&10;`.

PR #4200, opened 2024-12-13
Attribute non-test thread output to most recent test thread

marcphilipp:
When using frameworks or running external processes, test output is
often written on other threads than the test. When tests are executed
sequentially that output can be attributed unambiguously to the current
test. If tests are run in parallel, picking the right test to attribute
output to becomes much harder, though.


mpkorstanje:
>  If tests are run in parallel, picking the right test to attribute output to becomes much harder, though.

When executing in parallel, you could attach any output that can't be attributed to a test to the currently running engine. That would at least preserve some context.

marcphilipp:
> When executing in parallel, you could attach any output that can't be attributed to a test to the currently running engine. That would at least preserve some context.
> 
> It would require some communication between the platform and engine. Not sure where that would go.

Thanks for the idea! I don't think there's a perfect solution for this use case without explicitly attaching threads to tests or sth. like that. The current solution will capture output on the engine descriptor only if no test/container is still running.

Issue #3605, opened 2023-12-16
Support for binary literals in parameterized test arguments


MightyEagle82:
At the moment there is support for hexadecimal literals but not for binary:
- 0xF0 - allowed
- 0b11110000 - disallowed

sbrannen:
We actually provide support for decimal, hexadecimal, and octal numbers simply because that's what `java.lang.Integer.decode(String)` supports.

I had forgotten about binary literals, since I never use them personally, but...

> Starting in J2SE 7.0, as binary numbers, using the form 0b (or 0B) followed by one or more binary digits (0 or 1). For example, 0b101010 is the integer 42. Like octal and hex numbers, binary literals may represent negative numbers.

So, although we could potentially introduce explicit support for binary literals in `org.junit.platform.commons.support.conversion.StringToNumberConverter`, I think it may be a good idea to request direct support in the JDK (in `Integer.decode()`).

@sormuras, thoughts?

sormuras:
https://bugs.openjdk.org/browse/JDK-7181749 "Integer.decode("0b1010101") fails" has:

> Duplicate of functionality requested in comments of 6863378; closing accordingly.

Which in turn https://bugs.openjdk.org/browse/JDK-6863378 "Project Coin: Consider library support for underscores in numbers and binary literals" has:

> "Not essential functionality; closing as will not fix."

So, supporting such conversions is an option for JUnit - if there's enough interest from users and we are willing to maintain it.

MightyEagle82:
Guys, thank you for the replies anyway.

It was the first time when I decide to use `CsvSource` instead of `ValueSource` to provide binary values for test arguments. I haven't faced with `decode` methods until now. After posting this issue I found that they didn't support binary literals and was very surprised. I understand that JUnit relies on JDK API. So it won't be a problem if you just close this issue with **won't be fixed** status.

young0264:
Hi @sbrannen  @sormuras 

Thanks for the detailed discussion above.
I understand that 'Integer.decode()' and related JDK methods do not currently support binary literals ('Ob..'), and that this has been discussed in OpenJDK with a "won't fix" resolution.

I'd be happy to work on a PR that adds support for binary literals(e.g, `Ob1010`) to 'StringToNumberconverter' within JUnit, handling it explicitly in cases where the JDK doesn't.

Would the team be open to such a contribution, assuming it comes with tests and minimal maintenance overhead? 

Thanks again for your work on JUnit.!

[...]

Issue #2702, opened 2021-09-01
Implicit conversion from integral numbers to byte array


scordio:
I'd like to propose the addition of implicit conversion from integral numbers (`byte`, `short`, `int`, `long`) to byte array. The concrete use case is code dealing with bitwise operations, mostly related to cryptography.

My wish is to write something like this:

```java
  @ParameterizedTest
  @ValueSource(ints = {Integer.MIN_VALUE, 0, 0xFF, 0xFFFF, 0xFFFFFF, Integer.MAX_VALUE})
  void test(byte[] byteArray) {
    assertThat(byteArray).hasSize(4);
  }
```

where I can leverage binary and hexadecimal literals which are automatically converted to a byte array using a [`ByteBuffer`](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html) with Big Endian order.

I put together [a small poc](https://github.com/scordio/junit5-playground/tree/master/src/test/java/com/github/scordio/junit/jupiter/params/converter) with a custom argument converter. If the proposal is accepted, I'm happy to raise a PR about it.

## Deliverables

- [ ] A new implicit conversion from `byte` / `short` / `int` / `long` to `byte[]`

sbrannen:
For that use case, you wouldn't use an `ArgumentConverter` since that can only convert a single argument.

Rather, you would use an `ArgumentsAggregator` to aggregate all arguments into a single array.

sbrannen:
In addition, I don't see how we could introduce _implicit_ support for that without breaking existing use cases, but perhaps I'm overlooking something.

scordio:
@sbrannen no, I meant something different. Probably a more detailed example would've explained it better.

I don't want to aggregate all the arguments in a single array, I want each argument to be converted into a dedicated array, with array size depending on the primitive type.

Following my example:
```java
@ValueSource(ints = {Integer.MIN_VALUE, 0, 0xFF, 0xFFFF, 0xFFFFFF, Integer.MAX_VALUE})
```

each integer would be converted to a separate byte array of size 4, being the size in bytes of the integer type (i.e., `Integer.BYTES`). In case of byte values, each array would have size 1; shorts would have arrays of size 2, and longs would have arrays of size 8.

[...]

PR #4219, opened 2024-12-23
Introduce `Converter` in `junit-platform-commons`

### LINKED ISSUE #853: Create extension point for ArgumentConverter
As it stands argument converters need to be applied (with `@ConvertWith`) either to a custom annotation (at least it looks like that; couldn't make it work) or to the argument. When the former is not an option (for example because you can't make it work :wink:), the latter quickly becomes repetitive.

An extension point that allows registering argument converters as the class and method level (much like parameter resolvers) would remedy this problem.

Beyond ease of use, this extension point would also enable creating more wholesome extensions which, applied to a test class, can register parameter resolvers, argument converters, post instance extensions, and more _all in one annotation_. It would also make it possible to do something like `@ConvertByCalling("fromString")` (either on class, method, or parameter), which would try to call a static method `fromString(String)` on the target type.

If you think this is a valuable feature, I'd like to take a shot at implementing it.

### LINKED ISSUE #3141: `Locale` argument conversion not setting `language` and `country` properly
I am trying to write a parameterized test for a service that uses a resource bundle under the hood, loading the resource bundle with [`ResourceBundle.getBundle(String, Locale)`](https://docs.oracle.com/javase/8/docs/api/java/util/ResourceBundle.html#getBundle-java.lang.String-java.util.Locale-).

The test gets several `Locale` values in a `@CsvSource` annotation: 

The [documentation](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-argument-conversion-implicit) does not have an example for a country-based locale so I initially tried with:

```java
@CsvSource({
  "en_FR, value1",
  "en_GB, value2",
  "fr_FR, value3",
  "fr_MC, value4",
  "it_IT, value5",
  ...
})
void test(Locale locale, String expected) {
  // call service using the resource bundle

  // compare result with `expected`
}
```

All worked fine on Windows (local environment) but failed on Linux (CI).

After some digging, on Windows everything works because it's a case-insensitive file system, so for example the `en_FR` value is converted to a `en_fr` language-only locale (i.e., without country) and it successfully matches the file ending with `en_FR` because of case insensitivity.

Looking at the code, I think the problem is here:

https://github.com/junit-team/junit5/blob/4288cf1233b4af6002e922cb9ddd1b159fe2c9b3/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java#L265

which uses the [`Locale(String)`](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#Locale-java.lang.String-) constructor. The constructor Javadoc also explains the different behavior based on case sensitivity:

> This constructor normalizes the language value to lowercase. 

Looking at the test cases:

https://github.com/junit-team/junit5/blob/4288cf1233b4af6002e922cb9ddd1b159fe2c9b3/junit-jupiter-params/src/test/java/org/junit/jupiter/params/converter/DefaultArgumentConverterTests.java#L222-L223

Probably the second one is invalid. I would have expected that to be:

```java
assertConverts("en_us", Locale.class, Locale.US);
```

which does not pass with the current implementation.

## Steps to reproduce

As the documentation does not cover conversion of locales with country, I am not sure what would be the "right" test case to demonstrate the issue.

Assuming the [IETF BCP 47 language tag format](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#forLanguageTag-java.lang.String-) would be the right format for input values, this test fails already at the first assertion:

```java
  @ParameterizedTest
  @ValueSource(strings = { "en-US" })
  void test(Locale locale) {
    assertEquals("en", locale.getLanguage());
    assertEquals("US", locale.getCountry());
    assertEquals(Locale.US, locale);
  }
```

## Workaround

Currently, I'm using a custom `ArgumentConverter` which delegates the conversion to [`Locale::forLanguageTag`](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#forLanguageTag-java.lang.String-).

## Context

 - Used versions (Jupiter/Vintage/Platform): 5.8.2
 - Build Tool/IDE: Intellij IDEA / Maven

## Deliverables

- [ ] `DefaultArgumentConverter` properly converts IETF BCP 47 strings
- [ ] The [documentation](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-argument-conversion-implicit) has an example using the IETF BCP 47 language tag format

scordio:
## Overview

* Resolves #853
* Relates to #2702
* Relates to #3605
* Relates to #4514

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionContext.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionContext.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2015-2025 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.support.conversion;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+import static org.junit.platform.commons.util.ClassLoaderUtils.getDefaultClassLoader;
+
+import org.apiguardian.api.API;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * {@code ConversionContext} encapsulates the <em>context</em> in which the
+ * current conversion is being executed.
+ *
+ * <p>{@link Converter Converters} are provided an instance of
+ * {@code ConversionContext} to perform their work.
+ *
+ * @param sourceType the descriptor of the source type
+ * @param targetType the descriptor of the type the source should be converted into
+ * @param classLoader the {@code ClassLoader} to use
+ *
+ * @since 6.0
+ * @see Converter
+ */
+@API(status = EXPERIMENTAL, since = "6.0")
+public record ConversionContext(TypeDescriptor sourceType, TypeDescriptor targetType, ClassLoader classLoader) {
+
+	/**
+	 * Create a new {@code ConversionContext}, expecting an instance of the
+	 * source instead of its type descriptor.
+	 *
+	 * @param source the source instance; may be {@code null}
+	 * @param targetType the descriptor of the type the source should be converted into
+	 * @param classLoader the {@code ClassLoader} to use; may be {@code null} to
+	 * use the default {@code ClassLoader}
+	 */
+	public ConversionContext(@Nullable Object source, TypeDescriptor targetType, @Nullable ClassLoader classLoader) {
+		this(TypeDescriptor.forInstance(source), targetType,
+			classLoader != null ? classLoader : getDefaultClassLoader());
+	}
+
+}
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
@@ -10,15 +10,16 @@
 
 package org.junit.platform.commons.support.conversion;
 
+import static org.apiguardian.api.API.Status.DEPRECATED;
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.MAINTAINED;
-import static org.junit.platform.commons.util.ReflectionUtils.getWrapperType;
 
-import java.util.List;
-import java.util.Optional;
+import java.util.ServiceLoader;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import org.apiguardian.api.API;
 import org.jspecify.annotations.Nullable;
-import org.junit.platform.commons.util.ClassLoaderUtils;
 
 /**
  * {@code ConversionSupport} provides static utility methods for converting a
@@ -29,17 +30,6 @@ import org.junit.platform.commons.util.ClassLoaderUtils;
 @API(status = MAINTAINED, since = "1.13.3")
 public final class ConversionSupport {
 
-	private static final List<StringToObjectConverter> stringToObjectConverters = List.of( //
-		new StringToBooleanConverter(), //
-		new StringToCharacterConverter(), //
-		new StringToNumberConverter(), //
-		new StringToClassConverter(), //
-		new StringToEnumConverter(), //
-		new StringToJavaTimeConverter(), //
-		new StringToCommonJavaTypesConverter(), //
-		new FallbackStringToObjectConverter() //
-	);
-
 	private ConversionSupport() {
 		/* no-op */
 	}
@@ -47,48 +37,6 @@ public final class ConversionSupport {
 	/**
 	 * Convert the supplied source {@code String} into an instance of the specified
 	 * target type.
-	 *
-	 * <p>If the target type is {@code String}, the source {@code String} will not
-	 * be modified.
-	 *
-	 * <p>Some forms of conversion require a {@link ClassLoader}. If none is
-	 * provided, the {@linkplain ClassLoaderUtils#getDefaultClassLoader() default
-	 * ClassLoader} will be used.
-	 *
-	 * <p>This method is able to convert strings into primitive types and their
-	 * corresponding wrapper types ({@link Boolean}, {@link Character}, {@link Byte},
-	 * {@link Short}, {@link Integer}, {@link Long}, {@link Float}, and
-	 * {@link Double}), enum constants, date and time types from the
-	 * {@code java.time} package, as well as common Java types such as {@link Class},
-	 * {@link java.io.File}, {@link java.nio.file.Path}, {@link java.nio.charset.Charset},
-	 * {@link java.math.BigDecimal}, {@link java.math.BigInteger},
-	 * {@link java.util.Currency}, {@link java.util.Locale}, {@link java.util.UUID},
-	 * {@link java.net.URI}, and {@link java.net.URL}.
-	 *
-	 * <p>If the target type is not covered by any of the above, a convention-based
-	 * conversion strategy will be used to convert the source {@code String} into the
-	 * given target type by invoking a static factory method or factory constructor
-	 * defined in the target type. The search algorithm used in this strategy is
-	 * outlined below.
-	 *
-	 * <h4>Search Algorithm</h4>
-	 *
-	 * <ol>
-	 * <li>Search for a single, non-private static factory method in the target
-	 * type that converts from a {@link String} to the target type. Use the
-	 * factory method if present.</li>
-	 * <li>Search for a single, non-private constructor in the target type that
-	 * accepts a {@link String}. Use the constructor if present.</li>
-	 * <li>Search for a single, non-private static factory method in the target
-	 * type that converts from a {@link CharSequence} to the target type. Use the
-	 * factory method if present.</li>
-	 * <li>Search for a single, non-private constructor in the target type that
-	 * accepts a {@link CharSequence}. Use the constructor if present.</li>
-	 * </ol>
-	 *
-	 * <p>If multiple suitable factory methods or constructors are discovered they
-	 * will be ignored. If neither a single factory method nor a single constructor
-	 * is found, the convention-based conversion strategy will not apply.
 	 *
 	 * @param source the source {@code String} to convert; may be {@code null}
 	 * but only if the target type is a reference type
@@ -101,49 +49,44 @@ public final class ConversionSupport {
 	 * type is a reference type
 	 *
 	 * @since 1.11
+	 * @see DefaultConverter
+	 * @deprecated Use {@link #convert(Object, ConversionContext)} instead.
 	 */
-	@SuppressWarnings("unchecked")
+	@Deprecated
+	@API(status = DEPRECATED, since = "6.0")
 	public static <T> @Nullable T convert(@Nullable String source, Class<T> targetType,
 			@Nullable ClassLoader classLoader) {
-		if (source == null) {
-			if (targetType.isPrimitive()) {
-				throw new ConversionException(
-					"Cannot convert null to primitive value of type " + targetType.getTypeName());
-			}
-			return null;
-		}
-
-		if (String.class.equals(targetType)) {
-			return (T) source;
-		}
-
-		Class<?> targetTypeToUse = toWrapperType(targetType);
-		Optional<StringToObjectConverter> converter = stringToObjectConverters.stream().filter(
-			candidate -> candidate.canConvertTo(targetTypeToUse)).findFirst();
-		if (converter.isPresent()) {
-			try {
-				ClassLoader classLoaderToUse = classLoader != null ? classLoader
-						: ClassLoaderUtils.getDefaultClassLoader();
-				return (T) converter.get().convert(source, targetTypeToUse, classLoaderToUse);
-			}
-			catch (Exception ex) {
-				if (ex instanceof ConversionException conversionException) {
-					// simply rethrow it
-					throw conversionException;
-				}
-				// else
-				throw new ConversionException(
-					"Failed to convert String \"%s\" to type %s".formatted(source, targetType.getTypeName()), ex);
-			}
-		}
-
-		throw new ConversionException(
-			"No built-in converter for source type java.lang.String and target type " + targetType.getTypeName());
+		ConversionContext context = new ConversionContext(source, TypeDescriptor.forClass(targetType), classLoader);
+		return convert(source, context);
 	}
 
-	private static Class<?> toWrapperType(Class<?> targetType) {
-		Class<?> wrapperType = getWrapperType(targetType);
-		return wrapperType != null ? wrapperType : targetType;
+	/**
+	 * Convert the supplied source object into an instance of the specified
+	 * target type.
+	 *
+	 * @param source the source object to convert; may be {@code null}
+	 * but only if the target type is a reference type
+	 * @param context the context for the conversion
+	 * @param <T> the type of the target
+	 * @return the converted object; may be {@code null} but only if the target
+	 * type is a reference type
+	 * @since 6.0
+	 */
+	@API(status = EXPERIMENTAL, since = "6.0")
+	@SuppressWarnings({ "unchecked", "rawtypes", "TypeParameterUnusedInFormals" })
+	public static <T> @Nullable T convert(@Nullable Object source, ConversionContext context) {
+		ServiceLoader<Converter> serviceLoader = ServiceLoader.load(Converter.class, context.classLoader());
+
+		Converter converter = Stream.concat( //
+			StreamSupport.stream(serviceLoader.spliterator(), false), //
+			Stream.of(DefaultConverter.INSTANCE)) //
+				.filter(candidate -> candidate.canConvert(context)) //
+				.findFirst() //
+				.orElseThrow(() -> new ConversionException(
+					"No registered or built-in converter for source '%s' and target type %s".formatted( //
+						source, context.targetType())));
+
+		return (T) converter.convert(source, context);
 	}
 
 }
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2015-2025 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.support.conversion;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import org.apiguardian.api.API;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * {@code Converter} is an abstraction that allows an input object to
+ * be converted to an instance of a different class.
+ *
+ * <p>Implementations are loaded via the {@link java.util.ServiceLoader} and must
+ * follow the service provider requirements. They should not make any assumptions
+ * regarding when they are instantiated or how often they are called. Since
+ * instances may potentially be cached and called from different threads, they
+ * should be thread-safe.
+ *
+ * <p>Extend {@link SimpleConverter} if your implementation always converts
+ * from a given source type into a given target type and does not need access to
+ * the {@link ClassLoader} to perform the conversion.
+ *
+ * @param <S> the type of the source to convert
+ * @param <T> the type the source should be converted into
+ *
+ * @since 6.0
+ * @see ConversionSupport
+ * @see SimpleConverter
+ */
+@API(status = EXPERIMENTAL, since = "6.0")
+public interface Converter<S, T extends @Nullable Object> {
+
+	/**
+	 * Determine if the supplied conversion context is supported.
+	 *
+	 * @param context the context for the conversion; never {@code null}
+	 * @return {@code true} if the conversion is supported
+	 */
+	boolean canConvert(ConversionContext context);
+
+	/**
+	 * Convert the supplied source object according to the supplied conversion context.
+	 * <p>This method will only be invoked if {@link #canConvert(ConversionContext)}
+	 * returns {@code true} for the same context.
+	 *
+	 * @param source the source object to convert; may be {@code null}
+	 * but only if the target type is a reference type
+	 * @param context the context for the conversion; never {@code null}
+	 * @return the converted object; may be {@code null} but only if the target
+	 * type is a reference type
+	 * @throws ConversionException if an error occurs during the conversion
+	 */
+	T convert(@Nullable S source, ConversionContext context) throws ConversionException;
+
+}
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/SimpleConverter.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/SimpleConverter.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2015-2025 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.support.conversion;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import org.apiguardian.api.API;
+import org.jspecify.annotations.Nullable;
+import org.junit.platform.commons.util.Preconditions;
+
+/**
+ * {@code SimpleConverter} is an abstract base class for {@link Converter}
+ * implementations that always convert objects of a given source type into a
+ * given target type.
+ *
+ * @param <S> the type of the source argument to convert
+ * @param <T> the type of the target object to create from the source
+ * @since 6.0
+ */
+@API(status = EXPERIMENTAL, since = "6.0")
+public abstract class SimpleConverter<S, T extends @Nullable Object> implements Converter<S, T> {
+
+	private final Class<S> sourceType;
+	private final Class<T> targetType;
+
+	/**
+	 * Create a new {@code SimpleConverter}.
+	 *
+	 * @param sourceType the type of the argument to convert; never {@code null}
+	 * @param targetType the type of the target object to create from the source;
+	 * never {@code null}
+	 */
+	protected SimpleConverter(Class<S> sourceType, Class<T> targetType) {
+		this.sourceType = Preconditions.notNull(sourceType, "sourceType must not be null");
+		this.targetType = Preconditions.notNull(targetType, "targetType must not be null");
+	}
+
+	@Override
+	public final boolean canConvert(ConversionContext context) {
+		// FIXME adjust for subtypes
+		return !context.sourceType().equals(TypeDescriptor.NONE) //
+				&& this.sourceType == context.sourceType().getType() //
+				&& this.targetType == context.targetType().getType();
+	}
+
+	@Override
+	public final T convert(@Nullable S source, ConversionContext context) {
+		Preconditions.notNull(source, "source cannot be null");
+		return convert(source);
+	}
+
+	/**
+	 * Convert the supplied {@code source} object of type {@code S} into an object
+	 * of type {@code T}.
+	 *
+	 * @param source the source object to convert; never {@code null}
+	 * @return the converted object; may be {@code null} but only if the target
+	 * type is a reference type
+	 * @throws ConversionException if an error occurs during the conversion
+	 */
+	protected abstract T convert(S source) throws ConversionException;
+
+}
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToClassConverter.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToClassConverter.java
@@ -12,24 +12,21 @@ package org.junit.platform.commons.support.conversion;
 
 import org.jspecify.annotations.Nullable;
 import org.junit.platform.commons.support.ReflectionSupport;
+import org.junit.platform.commons.util.Preconditions;
 
-class StringToClassConverter implements StringToObjectConverter {
+class StringToClassConverter implements Converter<String, Class<?>> {
 
 	@Override
-	public boolean canConvertTo(Class<?> targetType) {
-		return targetType == Class.class;
+	public boolean canConvert(ConversionContext context) {
+		return !context.sourceType().equals(TypeDescriptor.NONE) && context.targetType().getType() == Class.class;
 	}
 
 	@Override
-	public Object convert(String source, Class<?> targetType) throws Exception {
-		throw new UnsupportedOperationException("Invoke convert(String, Class<?>, ClassLoader) instead");
-	}
-
-	@Override
-	public @Nullable Object convert(String className, Class<?> targetType, ClassLoader classLoader) throws Exception {
+	public Class<?> convert(@Nullable String className, ConversionContext context) {
+		Preconditions.notNull(className, "className cannot be null");
 		// @formatter:off
-		return ReflectionSupport.tryToLoadClass(className, classLoader)
-				.getOrThrow(cause -> new ConversionException(
+		return ReflectionSupport.tryToLoadClass(className, context.classLoader())
+				.getNonNullOrThrow(cause -> new ConversionException(
 						"Failed to convert String \"" + className + "\" to type java.lang.Class", cause));
 		// @formatter:on
 	}
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2015-2025 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.commons.support.conversion;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Parameter;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Optional;
+
+import org.apiguardian.api.API;
+import org.jspecify.annotations.Nullable;
+import org.junit.platform.commons.util.Preconditions;
+import org.junit.platform.commons.util.ReflectionUtils;
+
+/**
+ *
+ *
+ * @since 6.0
+ */
+@API(status = EXPERIMENTAL, since = "6.0")
+public final class TypeDescriptor {
+
+	/**
+	 * {@code TypeDescriptor} returned when no value is available.
+	 */
+	public static final TypeDescriptor NONE = new TypeDescriptor(null);
+
+	private final @Nullable Class<?> type;
+
+	public static TypeDescriptor forClass(Class<?> clazz) {
+		Preconditions.notNull(clazz, () -> "clazz must not be null");
+		return new TypeDescriptor(clazz);
+	}
+
+	public static TypeDescriptor forInstance(@Nullable Object instance) {
+		return instance != null ? forClass(instance.getClass()) : NONE;
+	}
+
+	public static TypeDescriptor forField(Field field) {
+		Preconditions.notNull(field, "field must not be null");
+		return forClass(field.getType());
+	}
+
+	public static TypeDescriptor forParameter(Parameter parameter) {
+		Preconditions.notNull(parameter, "parameter must not be null");
+		return forClass(parameter.getType());
+	}
+
+	private TypeDescriptor(@Nullable Class<?> type) {
+		this.type = type;
+	}
+
+	public Class<?> getType() {
+		if (type == null) {
+			throw new NoSuchElementException("No type present");
+		}
+		return type;
+	}
+
+	/**
+	 * Get the wrapper type of this type descriptor, if available.
+	 *
+	 * <p>If this type descriptor represents a primitive type, this method
+	 * returns the corresponding wrapped type. Otherwise, this method returns
+	 * {@link Optional#empty() empty()}.
+	 *
+	 * @return an {@code Optional} containing the wrapper type; never
+	 * {@code null} but potentially empty
+	 */
+	// FIXME [NullAway] parameter type of referenced method is @NonNull, but parameter in functional interface method java.util.function.Function.apply(T) is @Nullable
+	@SuppressWarnings("NullAway")
+	public Optional<Class<?>> getWrapperType() {
+		return Optional.ofNullable(type).map(ReflectionUtils::getWrapperType);
+	}
+
+	public boolean isPrimitive() {
+		return type != null && type.isPrimitive();
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+		TypeDescriptor that = (TypeDescriptor) o;
+		return Objects.equals(this.type, that.type);
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hashCode(type);
+	}
+
+	@Override
+	public String toString() {
+		return type != null ? type.getName() : "'null'";
+	}
+
+}
--- a/jupiter-tests/src/test/java/org/junit/jupiter/params/converter/LocaleConverter.java
+++ b/jupiter-tests/src/test/java/org/junit/jupiter/params/converter/LocaleConverter.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2015-2025 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.params.converter;
+
+import java.util.Locale;
+
+import org.junit.platform.commons.support.conversion.SimpleConverter;
+
+// FIXME move to ConversionSupportIntegrationTests
+public class LocaleConverter extends SimpleConverter<String, Locale> {
+
+	public LocaleConverter() {
+		super(String.class, Locale.class);
+	}
+
+	@Override
+	protected Locale convert(String source) {
+		return Locale.forLanguageTag(source);
+	}
+
+}
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/DefaultArgumentsAccessor.java
@@ -20,6 +20,7 @@ import java.util.function.BiFunction;
 import org.apiguardian.api.API;
 import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.params.converter.DefaultArgumentConverter;
+import org.junit.platform.commons.support.conversion.TypeDescriptor;
 import org.junit.platform.commons.util.ClassUtils;
 import org.junit.platform.commons.util.Preconditions;
 
@@ -45,7 +46,8 @@ public class DefaultArgumentsAccessor implements ArgumentsAccessor {
 		Preconditions.notNull(classLoader, "ClassLoader must not be null");
 
 		BiFunction<@Nullable Object, Class<?>, @Nullable Object> converter = (source,
-				targetType) -> DefaultArgumentConverter.INSTANCE.convert(source, targetType, classLoader);
+				targetType) -> DefaultArgumentConverter.INSTANCE.convert(source, TypeDescriptor.forClass(targetType),
+					classLoader);
 		return new DefaultArgumentsAccessor(converter, invocationIndex, arguments);
 	}
 
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/DefaultArgumentConverter.java
@@ -26,8 +26,10 @@ import org.apiguardian.api.API;
 import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.params.support.FieldContext;
+import org.junit.platform.commons.support.conversion.ConversionContext;
 import org.junit.platform.commons.support.conversion.ConversionException;
 import org.junit.platform.commons.support.conversion.ConversionSupport;
+import org.junit.platform.commons.support.conversion.TypeDescriptor;
 import org.junit.platform.commons.util.ReflectionUtils;
 
 /**
@@ -41,7 +43,7 @@ import org.junit.platform.commons.util.ReflectionUtils;
  * {@link File}, {@link BigDecimal}, {@link BigInteger}, {@link Currency},
  * {@link Locale}, {@link URI}, {@link URL}, {@link UUID}, etc.
  *
- * <p>If the source and target types are identical the source object will not
+ * <p>If the source and target types are identical, the source object will not
  * be modified.
  *
  * @since 5.0
@@ -58,49 +60,42 @@ public class DefaultArgumentConverter implements ArgumentConverter {
 
 	@Override
 	public final @Nullable Object convert(@Nullable Object source, ParameterContext context) {
-		Class<?> targetType = context.getParameter().getType();
 		ClassLoader classLoader = getClassLoader(context.getDeclaringExecutable().getDeclaringClass());
-		return convert(source, targetType, classLoader);
+		return convert(source, TypeDescriptor.forParameter(context.getParameter()), classLoader);
 	}
 
 	@Override
 	public final @Nullable Object convert(@Nullable Object source, FieldContext context)
 			throws ArgumentConversionException {
-
-		Class<?> targetType = context.getField().getType();
 		ClassLoader classLoader = getClassLoader(context.getField().getDeclaringClass());
-		return convert(source, targetType, classLoader);
+		return convert(source, TypeDescriptor.forField(context.getField()), classLoader);
 	}
 
-	public final @Nullable Object convert(@Nullable Object source, Class<?> targetType, ClassLoader classLoader) {
+	public final @Nullable Object convert(@Nullable Object source, TypeDescriptor targetType, ClassLoader classLoader) {
 		if (source == null) {
 			if (targetType.isPrimitive()) {
 				throw new ArgumentConversionException(
-					"Cannot convert null to primitive value of type " + targetType.getTypeName());
+					"Cannot convert null to primitive value of type " + targetType.getType().getTypeName());
 			}
 			return null;
 		}
 
-		if (ReflectionUtils.isAssignableTo(source, targetType)) {
+		if (ReflectionUtils.isAssignableTo(source, targetType.getType())) {
 			return source;
 		}
 
-		if (source instanceof String string) {
-			try {
-				return convert(string, targetType, classLoader);
-			}
-			catch (ConversionException ex) {
-				throw new ArgumentConversionException(ex.getMessage(), ex);
-			}
+		try {
+			ConversionContext context = new ConversionContext(source, targetType, classLoader);
+			return delegateConversion(source, context);
 		}
-
-		throw new ArgumentConversionException("No built-in converter for source type %s and target type %s".formatted(
-			source.getClass().getTypeName(), targetType.getTypeName()));
+		catch (ConversionException ex) {
+			throw new ArgumentConversionException(ex.getMessage(), ex);
+		}
 	}
 
 	@Nullable
-	Object convert(@Nullable String source, Class<?> targetType, ClassLoader classLoader) {
-		return ConversionSupport.convert(source, targetType, classLoader);
+	Object delegateConversion(@Nullable Object source, ConversionContext context) {
+		return ConversionSupport.convert(source, context);
 	}
 
 }
--- a/junit-platform-commons/src/main/java/module-info.java
+++ b/junit-platform-commons/src/main/java/module-info.java
@@ -56,5 +56,6 @@ module org.junit.platform.commons {
 			org.junit.platform.suite.engine,
 			org.junit.platform.testkit,
 			org.junit.vintage.engine;
+	uses org.junit.platform.commons.support.conversion.Converter;
 	uses org.junit.platform.commons.support.scanning.ClasspathScanner;
 }
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionException.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionException.java
@@ -15,6 +15,7 @@ import static org.apiguardian.api.API.Status.MAINTAINED;
 import java.io.Serial;
 
 import org.apiguardian.api.API;
+import org.jspecify.annotations.Nullable;
 import org.junit.platform.commons.JUnitException;
 
 /**
@@ -33,7 +34,7 @@ public class ConversionException extends JUnitException {
 		super(message);
 	}
 
-	public ConversionException(String message, Throwable cause) {
+	public ConversionException(String message, @Nullable Throwable cause) {
 		super(message, cause);
 	}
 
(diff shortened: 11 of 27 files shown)


scordio:
There is plenty of work to do 🙃

The current highlights:
* New `ConversionService` interface, with `ClassLoader` param instead of [`ConversionContext`](https://github.com/junit-team/junit5/issues/853#issuecomment-2545123955) (will come back to it)
* New `TypedConversionService` base class
* The implementation of the existing [`ConversionSupport.convert(String, Class, ClassLoader)`](https://github.com/junit-team/junit5/blob/efc375d763ac83019523ac40497b4301df3d9bb2/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java#L101-L137) is now part of the new `DefaultConversionService`
* A test setup to demonstrate how to solve #3141 (I'll shape it better as an integration test)

Any feedback would be highly appreciated!

junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/ArgumentConverter.java
```diff
@@ -40,7 +40,6 @@
  * the {@link ParameterContext} to perform the conversion.
  *
  * @since 5.0
- * @see SimpleArgumentConverter
```
scordio, line 43: This was a duplicate:

https://github.com/junit-team/junit5/blob/529da15aa840ba4d063af94db915b7e81d5fa8d4/junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/ArgumentConverter.java#L43-L47

marcphilipp:
Thanks for the draft! :+1: 

The tests are failing due to:
> org.junit.platform.commons.support.conversion.ConversionService: module org.junit.platform.commons does not declare `uses`

That's because `junit-platform-commons/src/module/org.junit.platform.commons/module-info.java` is missing
```
uses org.junit.platform.commons.support.conversion.ConversionService;
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -97,48 +47,49 @@ private ConversionSupport() {
 	 * type is a reference type
 	 *
 	 * @since 1.11
+	 * @see DefaultConversionService
```
marcphilipp, line 50: We should not link to a package-private type here as it will not be included in the published Javadoc.

```suggestion
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -97,48 +47,49 @@ private ConversionSupport() {
...
+	 * @deprecated Use {@link #convert(Object, Class, ClassLoader)} instead.
 	 */
 	@SuppressWarnings("unchecked")
+	@Deprecated
```
marcphilipp, line 54:
```suggestion
	@Deprecated
	@API(status = DEPRECATED, since = "5.12")
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -97,48 +47,49 @@ private ConversionSupport() {
...
 
-		throw new ConversionException(
-			"No built-in converter for source type java.lang.String and target type " + targetType.getTypeName());
+		return (T) Stream.concat( //
+			StreamSupport.stream(serviceLoader.spliterator(), false), //
+			Stream.of(DefaultConversionService.INSTANCE)) //
+				.filter(candidate -> candidate.canConvert(source, targetType, classLoader)) //
+				.findFirst() //
+				.map(candidate -> candidate.convert(source, targetType, classLoaderToUse)) //
+				.orElseThrow(() -> new ConversionException("No built-in converter for source type "
```
marcphilipp, line 87:
```suggestion
				.orElseThrow(() -> new ConversionException("No registered or built-in converter for source type "
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/DefaultConversionService.java
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		if (String.class.equals(targetType)) {
+			return source;
+		}
+
+		// FIXME move/copy next three lines to canConvert?
+		Class<?> targetTypeToUse = toWrapperType(targetType);
+		Optional<StringToObjectConverter> converter = stringToObjectConverters.stream().filter(
+			candidate -> candidate.canConvertTo(targetTypeToUse)).findFirst();
```
marcphilipp, line 148: I was also wondering about that but am not sure. I guess we should only move it if it makes things simpler or helps us produce a better error message.

jupiter-tests/src/test/java/org/junit/jupiter/params/converter/LocaleConversionService.java
```diff
@@ -0,0 +1,29 @@
...
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.params.converter;
+
+import java.util.Locale;
+
+import org.junit.platform.commons.support.conversion.TypedConversionService;
+
+// FIXME delete
```
marcphilipp, line 17: This would make a good test case, though. We have existing tests that register services for tests using an extra class loader:

https://github.com/junit-team/junit5/blob/16c6f72c1c728c015e35cb739ea75884f19f990c/platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java#L353-L366

We could generalize and move that method to a test utility class (e.g. in `junit-jupiter-api/src/testFixtures`) so it can be reused here.

marcphilipp (commented):
Looks very promising! :+1:

marcphilipp:
> That's because `junit-platform-commons/src/module/org.junit.platform.commons/module-info.java` is missing
> 
> ```
> uses org.junit.platform.commons.support.conversion.ConversionService;
> ```

When you add that, you'll also have to add it to `platform-tooling-support-tests/projects/jar-describe-module/junit-platform-commons.expected.txt` to adjust the integration test.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
scordio: `DefaultConversionService` now takes over the detailed explanation in the Javadoc about the conversion strategy for string parameters. Would it make sense to make the class `public` and mark it with `@API(status = INTERNAL, since = "5.12")`?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/DefaultConversionService.java
scordio: I think I found a good reason to move the logic to `canConvert`, I'll dig a bit more later.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
marcphilipp: IIRC `DefaultArgumentConverter` needs to be `public` because it's accessed from other packages. I can see your point and I'd be okay with doing this for `DefaultConversionService`.

scordio:
I've been lagging behind with this one but I should be able to spend time on it in the upcoming weekend.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/DefaultConversionService.java
scordio: Done in 8ed6eb6.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -30,17 +29,6 @@
 @API(status = EXPERIMENTAL, since = "1.11")
 public final class ConversionSupport {
```
scordio, line 30: I realized now that there are no tests for `ConversionSupport` in `junit-platform-commons`, but `DefaultArgumentConverterTests` from `jupiter-tests` provide the corresponding coverage.

Is that intended or should I add something to `platform-tests`?

scordio: No worries, I can take care of that!

scordio: Raised #4305.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/converter/TypedArgumentConverter.java
```diff
@@ -46,6 +46,9 @@ protected TypedArgumentConverter(Class<S> sourceType, Class<T> targetType) {
 		this.targetType = Preconditions.notNull(targetType, "targetType must not be null");
 	}
 
+	/**
+	 * {@inheritDoc}
```
sbrannen, line 50: Isn't this superfluous if you're not adding additional Javadoc yourself?

scordio:
Would you like to include this in 5.13 too? I should have enough time in the upcoming days to finalize it.

marcphilipp:
@scordio Sorry for the late reply! I was out for a few days. If you still have time to finish it in the next two weeks, we could still include it in 5.13.0-RC1. WDYT?

scordio:
Sounds good!

sbrannen:
Sorry, for chiming in so late in the game, but...

Instead of `ConversionService`, I would prefer that we call this new public SPI something more like `Converter`.

And I'm also wondering if it should be parameterized.

FYI: I'm taking inspiration from the conversion APIs in the Spring Framework, which have been evolved over the years.

For example, I see our `Converter` as basically a combination of the [`Converter`](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/converter/Converter.java) and [`ConditionalConverter`](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/converter/ConditionalConverter.java) APIs from Spring.

Thoughts?

marcphilipp:
I'd be okay with `Converter` instead of `ConversionService` as well. I'm not sure about parameterizing it. Where we use it (for example, to dynamically resolve method paramters), we won't be able to benefit from the type arguments. It might be a bit nicer for `ConversionService`/`Converter` authors to implement but we could also provide a convenience base class similar to `TypedArgumentConverter`. One thing we should consider is using a parameter object (like Spring's `TypeDescriptor`) instead of the raw `Class`. This could, for example, be used to later add support for nested classes with _runtime_ enclosing types. Thoughts?

sbrannen:
> I'd be okay with `Converter` instead of `ConversionService` as well.

Then let's change it to `Converter`. 👍 

> I'm not sure about parameterizing it. Where we use it (for example, to dynamically resolve method parameters), we won't be able to benefit from the type arguments. It might be a bit nicer for `ConversionService`/`Converter` authors to implement but we could also provide a convenience base class similar to `TypedArgumentConverter`.

Although we (the framework itself) might not benefit from generics at the moment, I would rather introduce them now than later discover that we actually need them ourselves to support additional (more dynamic) use cases.

In other words, what downside would there be to supporting generics from the start?

> One thing we should consider is using a parameter object (like Spring's `TypeDescriptor`) instead of the raw `Class`. This could, for example, be used to later add support for nested classes with _runtime_ enclosing types. Thoughts?

Yes, as you know... I am a big fan of "parameter context objects" in APIs that need to stand the test of time.

As I mentioned before, the conversion subsystem in Spring Framework has evolved organically over time, and most modern implementations actually implement for [GenericConverter](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/converter/GenericConverter.java) _and_ [`ConditionalConverter`](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/converter/ConditionalConverter.java), which both deal with `TypeDescriptor` instead of `Class` (except for the direct `Class` usage in `ConvertiblePair`).

In fact, there's even a dedicated [`ConditionalGenericConverter`](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/converter/ConditionalGenericConverter.java) interface which combines those two, which I implemented in the latest converter that I implemented: [`OptionalToObjectConverter`](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/support/OptionalToObjectConverter.java).

I am of course not suggesting that we introduce mechanisms as complex as those in Spring; however, I do hope that we can design in the API in such a manner that we don't prevent ourselves from being able to evolve the APIs and feature set over time as necessary.

And... supporting generics and something like a `TypeDescriptor` parameter object would be a good step in that direction.

scordio:
> @sbrannen Sorry, for chiming in so late in the game, but...

Nothing has been merged, so it's definitely not late! 🙃 

> @sbrannen Instead of `ConversionService`, I would prefer that we call this new public SPI something more like `Converter`.

I like `Converter`, too!

> @sbrannen  And I'm also wondering if it should be parameterized.

> @marcphilipp we could also provide a convenience base class similar to `TypedArgumentConverter`.

I introduced [`TypedConversionService`](https://github.com/scordio/junit5/blob/conversion-service/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypedConversionService.java). Do you mean that, to be renamed to `TypedConverter` maybe?

> @marcphilipp One thing we should consider is using a parameter object (like Spring's `TypeDescriptor`) instead of the raw Class. This could, for example, be used to later add support for nested classes with runtime enclosing types. Thoughts?

What about a `Type` parameter? We had a similar case in AssertJ and decided to use that, with some remarks in the Javadoc contract:

https://github.com/assertj/assertj/blob/b600591dbf480e1311b19b9eae6104ed4ae87442/assertj-core/src/main/java/org/assertj/core/api/AssertFactory.java#L78-L92

Of course, this puts the burden of dealing with `Type` subtypes on the implementation. Who's on Java 21+ can make it nicer with `instanceof` pattern matching for `switch`.

marcphilipp:
I agree with @sbrannen. Let's go with the generic version and the parameter object.

scordio:
I think `TypeDescriptor` shouldn't be specific to the conversion feature, so I plan to define it in `org.junit.platform.commons`.

Any better ideas?

sbrannen:
> I think `TypeDescriptor` shouldn't be specific to the conversion feature, so I plan to define it in `org.junit.platform.commons`.

I think it's OK for `TypeDescriptor` to reside in `org.junit.platform.commons.support.conversion`.

scordio:
@sbrannen I'm still trying to get a proper background before jumping on the actual changes, so here's a dumb question 🙂 

Specifically about parameterized types, is the [`ResolvableType`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/ResolvableType.html) abstraction closer to what we would need, from your perspective? Or maybe you imagine that too, alongside a `TypeDescriptor`?

marcphilipp:
To start with, I had thought a relatively simple interface should suffice for `TypeDescriptor`.

```java
interface TypeDescriptor {
	Class<?> getType();
}
```

@sbrannen WDYT?

scordio:
My suggestion is not to close the door for `Type getType()` so maybe `Class<?> getRawClass()` or similar?

Mostly to be ready for cases where `TypeDescriptor` might wrap a `ParameterizedType`.

marcphilipp:
We could do what Java did (e.g. for `Method.getReturnType()`) and have `Class<?> getType()` now and potentially `Type getGenericType()` later.

scordio:
After renaming to `Converter`, I realized `StringToObjectConverter` is pretty much a similar abstraction.

I propose to refactor the `StringToObjectConverter` hierarchy and rewrite all classes as `Converter` implementations (still with package-private visibility and used by `DefaultConverter` only).

WDYT?

sbrannen:
> I propose to refactor the `StringToObjectConverter` hierarchy and rewrite all classes as `Converter` implementations (still with package-private visibility and used by `DefaultConverter` only).
> 
> WDYT?

That's pretty much the vision I had all along: to have single abstraction/API.

So, yes, that sounds good to me. 👍 

Though... the devil is in details. 😉

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
```diff
@@ -0,0 +1,77 @@
...
+		return new TypeDescriptor(parameter.getType());
+	}
+
+	private TypeDescriptor(Class<?> type) {
+		this.type = type;
+	}
+
+	public Class<?> getType() {
+		return type;
+	}
```
scordio, line 49: For now, all the code relies on `getType()`, but I plan to check if dedicated APIs could better encapsulate these use cases.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -97,48 +48,50 @@ private ConversionSupport() {
...
+	@API(status = DEPRECATED, since = "5.13")
 	public static <T> T convert(String source, Class<T> targetType, ClassLoader classLoader) {
-		if (source == null) {
-			if (targetType.isPrimitive()) {
-				throw new ConversionException(
-					"Cannot convert null to primitive value of type " + targetType.getTypeName());
-			}
-			return null;
-		}
+		return convert(source, TypeDescriptor.forType(targetType), getClassLoader(classLoader));
```
scordio, line 57: I wonder if this should delegate to the other `convert` method or should invoke `DefaultConverter` directly, skipping the new service loader logic. For now, I went with the former.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
```diff
@@ -0,0 +1,68 @@
...
+	 * Determine if the supplied source object can be converted into an instance
+	 * of the specified target type.
+	 *
+	 * @param source the source object to convert; may be {@code null} but only
+	 * if the target type is a reference type
+	 * @param targetType the descriptor of the type the source should be converted into;
+	 * never {@code null}
+	 * @return {@code true} if the supplied source can be converted
+	 */
+	boolean canConvert(Object source, TypeDescriptor targetType);
```
scordio, line 48: Initially, this was:

```suggestion
	boolean canConvert(Object source, TypeDescriptor targetType, ClassLoader classLoader);
```

However, I couldn't see the benefit of the `ClassLoader` parameter here, so I deleted it.

Plus, I'm not entirely sure if this should rather be:

```suggestion
	boolean canConvert(TypeDescriptor sourceType, TypeDescriptor targetType);
```

The way `Converter` is currently integrated into `ConversionSupport` makes it easy to provide the source instance, but I wonder if this would be flexible enough with a bit of foresight.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypedConverter.java
```diff
@@ -0,0 +1,68 @@
...
+	/**
+	 * Convert the supplied {@code source} object of type {@code S} into an object
+	 * of type {@code T}.
+	 *
+	 * @param source the source object to convert; may be {@code null}
+	 * @return the converted object; may be {@code null} but only if the target
+	 * type is a reference type
+	 * @throws ConversionException if an error occurs during the conversion
+	 */
+	protected abstract T convert(S source) throws ConversionException;
```
scordio, line 66: Should this carry the target type too?

```suggestion
	protected abstract T convert(S source, TypeDescriptor targetType) throws ConversionException;
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToObjectConverter.java
```diff
@@ -14,39 +14,33 @@
...
 
-	/**
-	 * Convert the supplied {@link String} to the supplied target type (which is
-	 * guaranteed to be a wrapper type for primitives &mdash; for example,
-	 * {@link Integer} instead of {@code int}).
-	 *
-	 * <p>This method will only be invoked in {@link #canConvertTo(Class)}
-	 * returned {@code true} for the same target type.
-	 */
-	Object convert(String source, Class<?> targetType) throws Exception;
```
scordio, line 34: Being an internal API, I deleted this `convert(String, Class<?>)` variant to reduce complexity. The downside is that the `ClassLoader` parameter is unused on average.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypedConverter.java
marcphilipp: Yes, I think so as well as `sourceType` and `classLoader`.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
```diff
@@ -0,0 +1,68 @@
...
+ * <p>Extend {@link TypedConverter} if your implementation always converts
+ * from a given source type into a given target type and does not need access to
+ * the {@link ClassLoader} to perform the conversion.
+ *
+ * @since 1.13
+ * @see ConversionSupport
+ * @see TypedConverter
+ */
+@API(status = EXPERIMENTAL, since = "1.13")
+public interface Converter {
```
marcphilipp, line 36: Didn't we decide to make this `generic` as in `Converter<S, T>`?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
marcphilipp: I think we should pass an _additional` `TypeDescriptor sourceType` to both `canConvert` and `convert`.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
marcphilipp: I guess for that to make sense, we'd have to parameterize `TypeDescriptor` as well.

@scordio @sbrannen WDYT?

scordio: I had the same in mind but wasn't entirely confident about this direction yesterday, especially before refactoring `StringToObjectConversion`.

But now the refactoring is done so I can sketch it and see what happens 🙂

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
scordio: Done in 5190e37.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+		return new TypeDescriptor(instance.getClass());
+	}
+
+	public static TypeDescriptor forField(Field field) {
+		return new TypeDescriptor(field.getType());
+	}
+
+	public static TypeDescriptor forParameter(Parameter parameter) {
+		return new TypeDescriptor(parameter.getType());
+	}
```
scordio, line 45: General `null` handling is currently missing

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
```diff
@@ -0,0 +1,81 @@
...
+			return false;
+		}
+		TypeDescriptor that = (TypeDescriptor) o;
+		return this.type.equals(that.type);
+	}
+
+	@Override
+	public int hashCode() {
+		return this.type.hashCode();
+	}
```
scordio, line 79: I noticed `EqualsAndHashCodeAssertions` under `junit-jupiter-api` to test `equals`/`hashCode`. 

Would you like me to do something similar in `junit-platform-commons`?

Was [EqualsVerifier](https://jqno.nl/equalsverifier/) ever considered for such tests?

jupiter-tests/src/test/java/org/junit/jupiter/params/converter/LocaleConversionService.java
scordio: > We could generalize and move that method to a test utility class (e.g. in `junit-jupiter-api/src/testFixtures`) so it can be reused here.

@marcphilipp fine if I do it in a separate PR? Mostly to keep the size of this one under control 🙃

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
marcphilipp: > I noticed `EqualsAndHashCodeAssertions` under `junit-jupiter-api` to test `equals`/`hashCode`. It seems I can use the same in `junit-platform-commons`.

:+1: 

> Out of curiosity, was [EqualsVerifier](https://jqno.nl/equalsverifier/) ever considered for such tests?

I don't think it was discussed but it might be a good addition or even replacement but probably in a separate PR.

jupiter-tests/src/test/java/org/junit/jupiter/params/converter/LocaleConversionService.java
scordio: Raised #4544.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
```diff
@@ -0,0 +1,93 @@
...
+		Preconditions.notNull(field, "field must not be null");
+		return new TypeDescriptor(field.getType());
+	}
+
+	public static TypeDescriptor forParameter(Parameter parameter) {
+		Preconditions.notNull(parameter, "parameter must not be null");
+		return new TypeDescriptor(parameter.getType());
+	}
+
+	private TypeDescriptor(@Nullable Class<?> type) {
```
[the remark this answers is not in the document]
marcphilipp: The use case is converting from `null` to a target type, right? I wonder if we can use `Object` as the source type in those cases. WDYT?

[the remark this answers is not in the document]
scordio: I was thinking about `Object`, but:

```java
Object o = null;
System.out.println(o instanceof Object)
```

prints `false`, so I thought it could be misleading. Do you think `Void` could be better?

Another option I see is to make `TypeDescriptor` an interface and have two private implementations, e.g., `ClassBasedTypeDescriptor` and `NullBasedTypeDescriptor`, so that the constructor of `ClassBasedTypeDescriptor` doesn't need to be annotated with `@Nullable` anymore. WDYT?

(better proposals about naming are always welcome 😄)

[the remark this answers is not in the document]
scordio: Let me know what you think about 9fe5b3b.

If you like it, I'll remove the annotation from #4571 and update the Javadoc instead.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -98,49 +49,52 @@ private ConversionSupport() {
...
+	 * never {@code null}
+	 * @param classLoader the {@code ClassLoader} to use; may be {@code null} to
+	 * use the default {@code ClassLoader}
+	 * @param <T> the type of the target
+	 * @return the converted object; may be {@code null} but only if the target
+	 * type is a reference type
+	 *
+	 * @since 1.13
 	 */
+	@API(status = EXPERIMENTAL, since = "1.13")
```
scordio, line 77: If I remember correctly, the platform will jump to version 6, right?

```suggestion
	@API(status = EXPERIMENTAL, since = "6.0")
```

marcphilipp: Yep, that's correct. :+1:

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
scordio: Started in 8b78923, there is still work to do.

@marcphilipp @sbrannen could you please check if this is the direction you have in mind?

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

marcphilipp:
@scordio Thanks! I've changed the setting.

scordio:
Now that `Converter` has type parameters, would it make sense to implement that in `TypedArgumentConverter` so that the existing `TypedArgumentConverter` implementations could also be used with the new SPI?

Before exploring this in detail, I wanted to check your opinion first.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypedConverter.java
scordio: There is now `ConversionContext`.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToObjectConverter.java
scordio: `StringToObjectConverter` has been split into `StringToTargetTypeConverter` and `StringToWrapperTypeConverter`, plus `StringToClassConverter` now implements `Converter` directly.

All the above types are package-private, being an implementation detail of `DefaultConverter`.

marcphilipp:
I think that could be useful and worth giving a try. :+1:

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToObjectConverter.java
```diff
@@ -16,40 +16,34 @@
  * Internal API for converting arguments of type {@link String} to a specified
  * target type.
  */
-interface StringToObjectConverter {
+abstract class StringToObjectConverter implements Converter<String, Object> {
+
+	@Override
+	public final boolean canConvert(TypeDescriptor sourceType, TypeDescriptor targetType) {
+		return canConvert(targetType.getType());
```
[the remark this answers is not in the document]
scordio: Since there is `TypeDescriptor.NONE` as the constant to represent when there is no value, I'm thinking that `getType()` should be non-nullable and fail with an exception if invoked on the `NONE` instance.

This means that `Converter` implementations should guard such invocations with a check, like how `DefaultConverter` does at the moment:

https://github.com/scordio/junit-framework/blob/e4d293e46b1c319a15da63f0a292d9469d10b46e/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/DefaultConverter.java#L75-L77 

I'll update the PR accordingly and see if corner cases arise.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
```diff
@@ -0,0 +1,101 @@
...
+		Class<?> wrapperType = ReflectionUtils.getWrapperType(type);
+		return wrapperType != null ? wrapperType : getType();
+	}
+
+	public boolean isPrimitive() {
+		return type.isPrimitive();
+	}
+
+	public String getTypeName() {
+		return type.getName();
```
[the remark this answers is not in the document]
scordio: Yes, especially when taking the direction mentioned at https://github.com/junit-team/junit-framework/pull/4219#discussion_r2217979779.

jupiter-tests/src/test/java/org/junit/jupiter/params/converter/LocaleConverter.java
```diff
@@ -0,0 +1,30 @@
...
+// FIXME move to ConversionSupportIntegrationTests
+public class LocaleConverter extends TypedConverter<String, Locale> {
+
+	public LocaleConverter() {
+		super(String.class, Locale.class);
+	}
+
+	@Override
+	protected @Nullable Locale convert(@Nullable String source) {
+		return source != null ? Locale.forLanguageTag(source) : null;
```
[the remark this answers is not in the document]
scordio: It's possible that you commented before I refactored the `Converter` hierarchy, so what I'm about to say might not align with what you had in mind 🙂 

Currently, `LocaleConverter` extends `TypedConverter`, and I assume the latter should maintain support for null values, similar to `TypedArgumentConverter`.

However, for the sake of this test, `LocaleConverter` could also extend `Converter` directly and have a more fine-grained `canConvert` implementation, given that the underlying `Locale.forLanguageTag` API is not supposed to accept null values.

Anyway, I think this test lost its value as `DefaultConverter` uses `Locale.forLanguageTag` natively (#4751), so I'll probably rewrite it differently or replace it with a more meaningful one.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
```diff
@@ -0,0 +1,113 @@
...
+	 * {@link Optional#empty() empty()}.
+	 *
+	 * @return an {@code Optional} containing the wrapper type; never
+	 * {@code null} but potentially empty
+	 */
+	// FIXME [NullAway] parameter type of referenced method is @NonNull, but parameter in functional interface method java.util.function.Function.apply(T) is @Nullable
+	@SuppressWarnings("NullAway")
+	public Optional<Class<?>> getWrapperType() {
+		return Optional.ofNullable(type).map(ReflectionUtils::getWrapperType);
+	}
```
scordio, line 85: I think this is a NullAway bug. I'll compose a small reproducer and raise an issue with the project.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToClassConverter.java
```diff
@@ -12,23 +12,20 @@
...
 	}
 
 	@Override
-	public Object convert(String source, Class<?> targetType) throws Exception {
-		throw new UnsupportedOperationException("Invoke convert(String, Class<?>, ClassLoader) instead");
-	}
-
-	@Override
-	public @Nullable Object convert(String className, Class<?> targetType, ClassLoader classLoader) throws Exception {
+	public @Nullable Class<?> convert(@Nullable String className, ConversionContext context) {
```
scordio: I think this should be:


```suggestion
	public Class<?> convert(@Nullable String className, ConversionContext context) {
```

However, JSpecify disagrees with me, I guess due to `Try`... I need to work on it a bit more.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
marcphilipp: Hmm, that is weird. How does `NullAway` know about the nullability of the `Function` type parameters of `Optional.map`? Maybe it doesn't?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToClassConverter.java
marcphilipp: `Try` may contain `null` but you can use `Try.getNonNullOrThrow` instead.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
scordio: I'm still playing with it to understand it better. For example, the error is slightly different with a lambda:

```java
public Optional<Class<?>> getWrapperType() {
	return Optional.ofNullable(type).map(value -> ReflectionUtils.getWrapperType(value));
}
```

```
warning: [NullAway] passing @Nullable parameter 'value' where @NonNull is required
		return Optional.ofNullable(type).map(value -> ReflectionUtils.getWrapperType(value));
		                                                                             ^
    (see http://t.uber.com/nullaway )
```

I would have expected NullAway to infer that `value` is non-nullable, but maybe it has no way to do that.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToObjectConverter.java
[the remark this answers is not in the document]
scordio: `getType()` is no longer `@Nullable` 🥳

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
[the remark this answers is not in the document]
scordio: `getTypeName()` no longer exists and I preferred overriding `toString()` instead.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
marcphilipp: It looks to me like it thinks `Optional.map` is typed as 

`<U> Optional<U> map(Function<? super @Nullable T, ? extends @Nullable U> mapper)`

when it should be

`<U> Optional<U> map(Function<? super @NonNull T, ? extends @Nullable U> mapper)`

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/StringToClassConverter.java
scordio: Done in 2b180cd.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @param <S> the type of the source to convert
+ * @param <T> the type the source should be converted into
+ *
+ * @since 6.0
+ * @see ConversionSupport
+ * @see TypedConverter
+ */
+@API(status = EXPERIMENTAL, since = "6.0")
+public interface Converter<S, T> {
```
scordio: Taking inspiration from `Try`, I wonder if this should be:


```suggestion
public interface Converter<S, T extends @Nullable Object> {
```

scordio:
@marcphilipp @sbrannen I think this is now ready for another round of reviews!

A few points that still require work:
* Better test coverage (unit and integration tests)
* Documentation (Javadoc and user guide)
* The unresolved comments above

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
marcphilipp: In that case, `convert` should not return `@Nullable T` but just `T`, shouldn't it?

scordio: Yes, exactly. I'm trying it out.

scordio: Done in a57f5c9.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypedConverter.java
```diff
@@ -0,0 +1,69 @@
...
+ * {@code TypedConversionService} is an abstract base class for
+ * {@link Converter} implementations that always convert objects of a
+ * given source type into a given target type.
+ *
+ * @param <S> the type of the source argument to convert
+ * @param <T> the type of the target object to create from the source
+ * @since 6.0
+ */
+@API(status = EXPERIMENTAL, since = "6.0")
+public abstract class TypedConverter<S, T extends @Nullable Object> implements Converter<S, T> {
```
marcphilipp: Since `Converter` is now "typed", I'm wondering if we should rename this class to `SimpleConverter` or sth. similar?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypedConverter.java
```diff
@@ -0,0 +1,69 @@
...
+	@Override
+	public final boolean canConvert(ConversionContext context) {
+		// FIXME TypeDescriptor.NONE handling?
+		// FIXME add test cases with subtypes
+		return this.sourceType == context.sourceType().getType() && this.targetType == context.targetType().getType();
+	}
+
+	@Override
+	public final T convert(@Nullable S source, ConversionContext context) {
+		return convert(source);
```
marcphilipp: Is this extra indirection worth it?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionContext.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @param sourceType the descriptor of the source type
+ * @param targetType the descriptor of the type the source should be converted into
+ * @param classLoader the {@code ClassLoader} to use
+ *
+ * @since 6.0
+ * @see Converter
+ */
+@API(status = EXPERIMENTAL, since = "6.0")
+public record ConversionContext(TypeDescriptor sourceType, TypeDescriptor targetType, ClassLoader classLoader) {
```
marcphilipp, line 34: I guess the constructors need to be public so they can be used in unit tests for `Converter` implementations, right?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
```diff
@@ -0,0 +1,64 @@
...
+@API(status = EXPERIMENTAL, since = "6.0")
+public interface Converter<S, T extends @Nullable Object> {
+
+	/**
+	 * Determine if the supplied conversion context is supported.
+	 *
+	 * @param context the context for the conversion; never {@code null}
+	 * @return {@code true} if the conversion is supported
+	 */
+	boolean canConvert(ConversionContext context);
```
marcphilipp, line 48: Should we pass `source` here as well? In other words, could a `Converter` only be able to convert `S` only partially (for example, excluding `null`)?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
```diff
@@ -0,0 +1,64 @@
...
+	 * returned {@code true} for the same context.
+	 *
+	 * @param source the source object to convert; may be {@code null}
+	 * but only if the target type is a reference type
+	 * @param context the context for the conversion; never {@code null}
+	 * @return the converted object; may be {@code null} but only if the target
+	 * type is a reference type
+	 * @throws ConversionException if an error occurs during the conversion
+	 */
+	T convert(@Nullable S source, ConversionContext context) throws ConversionException;
```
marcphilipp, line 62: Should we really allow `source` to be `null` here? I think it would simplify things a lot if we didn't, wouldn't it? Maybe we'd restrict ourselves too much if we forbade `null` here, but maybe `TypedConverter` or a new `NonNullConverter` base class could forbid `null` and make it easier to implement a `Converter` with that returns `@NonNull T`?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionContext.java
scordio: I kept them public for any code invoking `ConversionSupport.convert()`, which is not only `DefaultArgumentConverter` but potentially user code as `ConversionSupport` is a [public API](https://github.com/junit-team/junit-framework/issues/3449).

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
scordio: I initially tried this direction but ended up with class cast exceptions due to `Converter` having type parameters, so I eventually removed it.

I also checked how Spring does it in [ConditionalConverter](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/convert/converter/ConditionalConverter.html#matches(org.springframework.core.convert.TypeDescriptor,org.springframework.core.convert.TypeDescriptor)), and the strategy seemed to be the same, so I thought this was a common pitfall 🙂

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypedConverter.java
scordio: The advantage I see is that classes extending this one will have a simpler surface, without `ConversionContext` appearing in the code.

Also, after the updates mentioned at https://github.com/junit-team/junit-framework/pull/4219#discussion_r2223858778, this will prevent `null` source.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
scordio: > Maybe we'd restrict ourselves too much if we forbade `null` here

I think so, especially considering [this](https://github.com/scordio/junit-framework/blob/a57f5c9ba657f2f22f11aaae97493e09d74f7324/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/DefaultConverter.java#L75-L77) and [this](https://github.com/scordio/junit-framework/blob/a57f5c9ba657f2f22f11aaae97493e09d74f7324/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/DefaultConverter.java#L133-L138) logic in `DefaultConverter`. Part of the `null` handling could be pulled to `ConversionSupport`, but I have the feeling it would spread the responsibility of rejecting `null` in case of primitive types.

As an extreme case, I considered the scenario where users would like to completely replace the built-in logic of `DefaultConverter` with a custom `Converter` that always returns `true` on `canConvert`. Such a converter might decide that a `null` source is always converted to a placeholder object. If `Converter` were to enforce a non-null source, such a use case wouldn't be possible. As I mentioned, this is likely to be unrealistic, but I considered it anyway 🙂 

> maybe `TypedConverter` or a new `NonNullConverter` base class could forbid `null` and make it easier to implement a `Converter` with that returns `@NonNull T`?

All considered, I think having `TypedConverter` to enforce a non-null source makes sense, and implementors can always fall back to `Converter` if they want to have control of `null` values.

Initially, I wanted to mimic the behavior of `TypedArgumentConverter` in `TypedConverter`. However, after digesting https://github.com/junit-team/junit-framework/pull/4219#issuecomment-2927458333 a bit more, `Converter` is the actual counterpart of `TypedArgumentConverter`, and renaming `TypedConverter` to `SimpleConverter` would also help to remove the potential ambiguity.

I'll adjust `TypedConverter` accordingly.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypedConverter.java
scordio: Done in 7903c2b.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
scordio: Done in 7903c2b.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -96,49 +49,44 @@ private ConversionSupport() {
...
+	@API(status = EXPERIMENTAL, since = "6.0")
+	@SuppressWarnings({ "unchecked", "rawtypes", "TypeParameterUnusedInFormals" })
+	public static <T> @Nullable T convert(@Nullable Object source, ConversionContext context) {
+		ServiceLoader<Converter> serviceLoader = ServiceLoader.load(Converter.class, context.classLoader());
 
-		throw new ConversionException(
-			"No built-in converter for source type java.lang.String and target type " + targetType.getTypeName());
-	}
+		Converter converter = Stream.concat( //
+			StreamSupport.stream(serviceLoader.spliterator(), false), //
```
marcphilipp, line 81:
```suggestion
			serviceLoader.stream().map(ServiceLoader.Provider::get), //
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
marcphilipp: Thanks! I still think it would be preferable if `canConvert` could return `false` for `null` instead of `convert` throwing an exception (if the source type is not `TypeDescriptor.NONE`. I can see the problem with the `ClassCastExceptions` and don't have a great idea how to achieve this. One option could be to pass `@Nullable Object source` (instead of `S`) to `canConvert`. Another would be to introduce an extra `canConvertNull` method. WDYT?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -96,49 +49,44 @@ private ConversionSupport() {
...
+			Stream.of(DefaultConverter.INSTANCE)) //
+				.filter(candidate -> candidate.canConvert(context)) //
+				.findFirst() //
```
marcphilipp, line 84: We don't have to solve this now, but I'm fully expecting us to get requests for some kind of ordering/precedence support in the future.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/SimpleConverter.java
```diff
@@ -0,0 +1,71 @@
...
+	 * never {@code null}
+	 */
+	protected SimpleConverter(Class<S> sourceType, Class<T> targetType) {
+		this.sourceType = Preconditions.notNull(sourceType, "sourceType must not be null");
+		this.targetType = Preconditions.notNull(targetType, "targetType must not be null");
+	}
+
+	@Override
+	public final boolean canConvert(ConversionContext context) {
+		// FIXME adjust for subtypes
```
marcphilipp, line 48: What did you have in mind here? Allow subtypes of `S` and supertypes of `T`?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
```diff
@@ -96,49 +49,44 @@ private ConversionSupport() {
...
+	@SuppressWarnings({ "unchecked", "rawtypes", "TypeParameterUnusedInFormals" })
+	public static <T> @Nullable T convert(@Nullable Object source, ConversionContext context) {
+		ServiceLoader<Converter> serviceLoader = ServiceLoader.load(Converter.class, context.classLoader());
 
-		throw new ConversionException(
-			"No built-in converter for source type java.lang.String and target type " + targetType.getTypeName());
-	}
+		Converter converter = Stream.concat( //
+			StreamSupport.stream(serviceLoader.spliterator(), false), //
+			Stream.of(DefaultConverter.INSTANCE)) //
```
marcphilipp, line 82: I'm not sure whether this is better. From its Javadoc and a quick look at the code the `Iterator` backing `ServiceLoader.spliterator` is also lazy. I'll leave it up to you.

```suggestion
		Converter converter = Stream.<Supplier<Converter>> concat( //
			serviceLoader.stream(), //
			Stream.of(() -> DefaultConverter.INSTANCE)) //
				.map(Supplier::get) //
```

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
scordio: Maybe an `int order()` default method in `Converter` could be done already now?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/SimpleConverter.java
scordio: Yes, exactly. I just haven't had the time to provide proper testing coverage here yet 🙂 that's next on the list.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
scordio: If I recall correctly, I originally drew inspiration from other parts of JUnit here. 

I'll double-check both options.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java
marcphilipp: Let's hold off doing that for now to keep the PR size down.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java
```diff
@@ -0,0 +1,113 @@
...
+	public static TypeDescriptor forParameter(Parameter parameter) {
+		Preconditions.notNull(parameter, "parameter must not be null");
+		return forClass(parameter.getType());
+	}
+
+	private TypeDescriptor(@Nullable Class<?> type) {
+		this.type = type;
+	}
+
+	public Class<?> getType() {
```
marcphilipp, line 64: The discussion in #5004 made me think that we should consider supporting `Type` instead of just `Class<?>`, i.e. making this class more like Guava's [`TypeToken`](https://guava.dev/releases/21.0/api/docs/com/google/common/reflect/TypeToken.html) or Spring's [`ParameterizedTypeReference`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/ParameterizedTypeReference.html).

Thoughts?

scordio: > we should consider supporting `Type` instead of just `Class<?>`

Yes, I also think it would be more future-proof. I mentioned something similar at https://github.com/junit-team/junit-framework/pull/4219#issuecomment-2851078799, but I can see now I didn't make it prominent enough.

> Re `TypeToken`, there's a related PR introducing a similar class.

The current shape of `TypeDescriptor` probably makes it closer to Spring's [`ResolvableType`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/ResolvableType.html)... or, let's say, I was heavily inspired by it!

Do I understand it correctly that you're considering introducing a type token? Is there a use case related to this PR that you imagine would benefit from it, or is this mostly related to the discussion in #5004?

If we were to follow the Spring design, there could be a dedicated `TypeToken` class and a new `TypeDescriptor.forType(TypeToken)` factory method: I think such an enhancement could come at any point in time and not necessarily when `TypeDescriptor` is introduced.

([here](https://github.com/assertj/assertj/issues/3368#issuecomment-1962310392) is a slightly related discussion about type tokens in AssertJ, which we eventually decided to postpone, waiting for user demand)

Or maybe I misunderstood what you have in mind 🙃

marcphilipp: > Is there a use case related to this PR that you imagine would benefit from it, or is this mostly related to the discussion in https://github.com/junit-team/junit-framework/issues/5004?

For example, a `String` to `List<T>` converter, could inspect the type arguments of `List` and decide how to convert the individual elements.

Example: `"1,2,3"` as `source` and `List<Integer>` vs. `List<Long>` as `targetType`

> If we were to follow the Spring design, there could be a dedicated `TypeToken` class and a new `TypeDescriptor.forType(TypeToken)` factory method: I think such an enhancement could come at any point in time and not necessarily when `TypeDescriptor` is introduced.

Agreed, let's not do it right away but keep it in mind when naming things, for example, here:

```suggestion
	public Class<?> getRawClass() {
```

scordio: > Example: `"1,2,3"` as `source` and `List<Integer>` vs. `List<Long>` as `targetType`

I got a similar use case covered in scordio/junit-converters#4 (see the corresponding test case [here](https://github.com/scordio/junit-converters/blob/c8268bf9f5045fe8722ff4dd35857de8002e6236/src/test/java/io/github/scordio/junit/converters/tests/SpringConversionIntegrationTests.java#L49)).

However, that's at the level of `ArgumentConverter` where `Field` or `Parameter` is directly available and can be fed into the Spring classes to perform the conversion.

I'll keep this use case in mind and see how the Spring integration can work with the changes from this PR, especially to determine whether the current `TypeDescriptor` APIs are sufficient for this purpose.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/Converter.java
scordio: I'm unresolving this comment as I still have second thoughts about it.

> Should we really allow `source` to be `null` here? I think it would simplify things a lot if we didn't, wouldn't it?

After digesting this point for some time, I reevaluated where to position the responsibility for `null` conversion. That responsibility currently lies with `DefaultConverter`, which I moved from the original `ConversionSupport` implementation, but maybe that was a mistake:
* `ConversionSupport` should keep the `null` conversion responsibility, outside the `Converter` SPI
* `Converter#convert` should always expect non-null values

The limitations I see are that custom converters don't have control over the `null` conversion process, and the fact that the `Converter` contract deviates from the `ArgumentConverter` contract, the latter expecting `null` values.

Still, I believe restricting to non-null values would simplify things a lot (and Spring [used the same strategy](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/convert/converter/Converter.html#convert(S))).

Thoughts?

marcphilipp: @sbrannen Assuming you were involved on the Spring side, any experience you can share?

scordio:
This PR has fallen behind `main` and the comment history has become difficult to follow, so I'll rebase its branch and create a new PR, summarizing all the topics that haven't been closed.

PR #4225, opened 2025-01-01
Add workflow to verify reproducibility of binary release artifacts

marcphilipp:
This workflow is supposed to be triggered manually _after_ release
artifacts have been uploaded to a staging repository and the
corresponding Git tag has been pushed to GitHub.

To ensure their integrity and reproducibility, the workflow rebuilds
all binary artifacts and verifies that their checksums are identical to
those in the staging repository.


marcphilipp:
Successfully tested: https://github.com/junit-team/junit5/actions/runs/12572583304

Issue #4238, opened 2025-01-08
Implement Parallel Method Execution in JUnit-Vintage engine


YongGoose:
## Description

JUnit 4 allows running methods in parallel by setting a `RunnerScheduler` for ParentRunner subclasses. 
This issue aim to implement this feature to improve test execution performance.

## Objectives

- Implement parallel execution of test methods
- Adjust existing tests to verify method parallelization

## Deliverables

- [ ]  Implement custom RunnerScheduler using ExecutorService
- [ ]  Update tests to verify parallel method execution
- [ ]  Document usage and performance implications

---

This issue is a follow-up task to the issue below.

- https://github.com/junit-team/junit5/issues/2229

YongGoose:
@marcphilipp 

I plan to add a property to the **JUnit-Vintage engine**, similar to the two property values in the **JUnit-Jupiter engine**, to select whether only classes should be run in parallel or only methods should be run in parallel.

- `junit.jupiter.execution.parallel.mode.default`
- `junit.jupiter.execution.parallel.mode.classes.default`


plan to implement it similar to JUnit4's `ParallelComputer`
- https://github.com/junit-team/junit4/blob/df00d5eced3a7737b88de0f6f9e3673f0cf88f88/src/main/java/org/junit/experimental/ParallelComputer.java#L15-L17

I will complete the implementation by tomorrow and create a PR. 
If you have any better suggestions, please let me know in the comments! 😁

marcphilipp:
I think we should have two independent configuration parameters: one for classes and one for methods.

* `junit.vintage.execution.parallel.classes`
* `junit.vintage.execution.parallel.methods`

YongGoose:
> I think we should have two independent configuration parameters: one for classes and one for methods.
> 
> * `junit.vintage.execution.parallel.classes`
> * `junit.vintage.execution.parallel.methods`

Thank you for the suggestion 👍🏻

I’ll implement the two configuration parameters as you suggested :D

PR #4242, opened 2025-01-11
Implement Parallel Method Execution in JUnit-Vintage engine

YongGoose:
Resolves: #4238

## Overview

Implements parallel execution of test methods in the junit-vintage-engine, inspired by the parallelization support available in JUnit 4 through `ParallelComputer`. 
- This feature allows for enhanced test execution performance by leveraging modern multi-threaded environments.

## Changes
1. Implemented Custom RunnerScheduler 

- Created a RunnerScheduler using ExecutorService.
- Ensures proper scheduling and execution of test methods in parallel.

2. Updated Existing Tests 
- Adjusted tests to verify the parallel execution of methods.
- Added new test cases to ensure backward compatibility and correctness.

3. Documentation Updates **(Not yet)**
- Added documentation on how to enable and configure parallel method execution.
- Explained performance implications and scenarios where parallelization is beneficial.

---

---

### Definition of Done


YongGoose:
@marcphilipp 

I created this Draft PR to ask a few questions.
First, I’ve added two independent configuration parameters as you suggested. 

However, I have a question regarding their behavior:  

- What happens if the user sets the `junit.vintage.execution.parallel.enabled` parameter to `true`, but both `junit.vintage.execution.parallel.classes` and `junit.vintage.execution.parallel.methods` are set to `false`?

- Additionally, what occurs if `junit.vintage.execution.parallel.enabled` is set to `false`, while `junit.vintage.execution.parallel.classes` or `junit.vintage.execution.parallel.methods` set to `true`?

marcphilipp:
> What happens if the user sets the `junit.vintage.execution.parallel.enabled` parameter to `true`, but both `junit.vintage.execution.parallel.classes` and `junit.vintage.execution.parallel.methods` are set to `false`?

I think we should check for that, log a warning ([like we do](https://github.com/junit-team/junit5/blob/ee52c64336ad3a079d598bd1c3884bafbc285047/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/EnumConfigurationParameterConverter.java#L63-L66) for invalid configuration parameters in Jupiter), and continue as if `junit.vintage.execution.parallel.enabled` was `false`.

> Additionally, what occurs if `junit.vintage.execution.parallel.enabled` is set to `false`, while `junit.vintage.execution.parallel.classes` or `junit.vintage.execution.parallel.methods` set to `true`?

I think turning off parallel execution (using the `...enabled=false` parameter) while keeping the rest of the parallel configuration in place is a valid use case. Therefore, I think we should ignore the `...classes` and `...methods` parameters in this case without logging a warning.

YongGoose:
> > What happens if the user sets the `junit.vintage.execution.parallel.enabled` parameter to `true`, but both `junit.vintage.execution.parallel.classes` and `junit.vintage.execution.parallel.methods` are set to `false`?
> 
> I think we should check for that, log a warning ([like we do](https://github.com/junit-team/junit5/blob/ee52c64336ad3a079d598bd1c3884bafbc285047/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/EnumConfigurationParameterConverter.java#L63-L66) for invalid configuration parameters in Jupiter), and continue as if `junit.vintage.execution.parallel.enabled` was `false`.
> 
> > Additionally, what occurs if `junit.vintage.execution.parallel.enabled` is set to `false`, while `junit.vintage.execution.parallel.classes` or `junit.vintage.execution.parallel.methods` set to `true`?
> 
> I think turning off parallel execution (using the `...enabled=false` parameter) while keeping the rest of the parallel configuration in place is a valid use case. Therefore, I think we should ignore the `...classes` and `...methods` parameters in this case without logging a warning.

Done in https://github.com/junit-team/junit5/pull/4242/commits/be5234f9ba6820ecf5b84a58482483e6bdbf550d !
If you have a better approach, please let me know in the comments 🙂

YongGoose:
@marcphilipp 

I implemented this to get feedback on parallelization at the class and method levels. https://github.com/junit-team/junit5/pull/4242/commits/b09c55ec593bd9ac7028c3ab6a013f25c4be7af9

> The code is intended to facilitate understanding, so it hasn't been fully polished yet❕

There are three scenarios when running parallel tests:  

1. Running methods in parallel.  
2. Running classes in parallel.  
3. Running both methods and classes in parallel.  

> Currently facing challenges with designing for these scenarios.

I want to design it similarly to JUnit 4's ParallelComputer, but I need help..🥹
- https://github.com/junit-team/junit4/blob/df00d5eced3a7737b88de0f6f9e3673f0cf88f88/src/main/java/org/junit/experimental/ParallelComputer.java#L54-L66

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -138,6 +156,41 @@ private boolean executeInParallel(VintageEngineDescriptor engineDescriptor,
...
+			@Override
+			public void schedule(Runnable childStatement) {
+				executorService.submit(childStatement);
+			}
+
+			@Override
+			public void finished() {
+				try {
+					executorService.shutdown();
+					executorService.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
```
marcphilipp, line 171: I think shutting down the thread pool here is not correct because this will be called for each runner plus we're also using it if `classes` is `true` as well.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/RunnerTestDescriptor.java
```diff
@@ -161,6 +171,10 @@ public boolean isIgnored() {
 		return ignored;
 	}
 
+	public void setScheduler(RunnerScheduler scheduler) {
+		this.scheduler = scheduler;
+	}
```
marcphilipp, line 176: > I want to design it similarly to JUnit 4's ParallelComputer, but I need help..🥹
> 
>     * [junit-team/junit4@`df00d5e`/src/main/java/org/junit/experimental/ParallelComputer.java#L54-L66](https://github.com/junit-team/junit4/blob/df00d5eced3a7737b88de0f6f9e3673f0cf88f88/src/main/java/org/junit/experimental/ParallelComputer.java#L54-L66)

Wouldn't this suffice?

```suggestion
	public void setScheduler(org.junit.runners.model.RunnerScheduler scheduler) {
		Runner runner = getRunnerToReport(); 
		if (runner instanceof ParentRunner) {
			((ParentRunner<?>) runner).setScheduler(scheduler);
		}
	}
```

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/RunnerScheduler.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apiguardian.api.API;
+
+/**
+ * Represents a strategy for scheduling when individual test methods
+ * should be run (in serial or parallel)
+ *
+ * @since 5.13
+ */
+@API(status = INTERNAL, since = "5.13")
+public interface RunnerScheduler {
```
marcphilipp, line 24: Can't we use `org.junit.runners.model.RunnerScheduler` directly?

marcphilipp (changes requested):

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
YongGoose: Done in https://github.com/junit-team/junit5/commit/8831f949f961d5dff640b0bf0a043ad540914a5a :)

YongGoose: I agree with the comment you provided.  

Based on that, I removed the `shutdown` and only used the `awaitTermination` method. 
Then, I added a conditional statement to check if it completes within the given time.  https://github.com/junit-team/junit5/commit/e92e834db7b0ce6f960f35ece05404108c05f110

Upon testing, I found that not all tests finished within the specified time.  

As a result, I implemented the `finished` method to do nothing, but this caused the assertion comparing the start and end times to fail.  

Could you provide a hint about what kind of work might need to be done inside the `finished` method?

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/RunnerScheduler.java
YongGoose: Directly using `org.junit.runners.model.RunnerScheduler` doesn’t seem like a good approach.  

However, when running the `junit-vintage-engine:testWithoutJUnit4` task, an exception occurs because it fails to find `org.junit.runners.model.RunnerScheduler`.

- Exception message

```
* What went wrong:
Execution failed for task ':junit-vintage-engine:testWithoutJUnit4'.
> Failure during test discovery: Forked test JVM terminated unexpectedly with exit value 1
  Test Distribution has stopped executing any remaining tests and silently skips them!
  For potential reasons why the test JVM exited, check the troubleshooting section at https://gradle.com/help/test-distribution-troubleshooting.
  Standard error from JVM:
      Terminating due to fatal error
      java.lang.NoClassDefFoundError: org/junit/runners/model/RunnerScheduler
      	at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
```

marcphilipp: I pushed a commit that fixes that. Instead of referencing `RunnerScheduler` in `VintageTestEngine`, it passes `ExecutorService` to `RunnerTestDescriptor` and creates the `RunnerScheduler` in there.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
marcphilipp: I think we need to wait for all futures scheduled for a particular runner. I have addressed that in the commit I've pushed to your branch just now.

YongGoose:
@marcphilipp 

While writing test code, I discovered an issue caused by using the same `ExecutorService`.

Assumption: Lets assume the thread pool size of the ExecutorService is `2`.

Scenario:

1. Two tests, `A` and `B`, are initiated in parallel from an outer loop.
2. `Test A` internally attempts to execute two subtasks (A1, A2).
3. Similarly, `Test B` internally attempts to execute two subtasks (B1, B2).
4. `A` starts `A1`, and `B` starts `B1`. At this point, all threads in the pool are occupied.
5. `A` waits for `A2` to complete, and `B` waits for `B2` to complete.
6. However, `A2` and `B2` cannot start because there are no available threads in the pool.
7. As a result, both `A` and `B` enter a deadlock, waiting indefinitely.

To prevent this issue, we could either set a timeout on `allOf.get()` (as a way to resolve the deadlock even if the test doesn't execute successfully) or separate the `ExecutorService` instances.  

WDYT?

YongGoose:
> While writing test code, I discovered an issue caused by using the same `ExecutorService`.
> 
> * https://github.com/junit-team/junit5/actions/runs/12841358007/job/35811074310?pr=4242
> 
> > I concluded that the test did not complete due to a deadlock.
> 
> **Assumption**: Lets assume the thread pool size of the ExecutorService is `2`.
> 
> **Scenario**:
> 
> 1. Two tests, `A` and `B`, are initiated in parallel from an outer loop.
> 2. `Test A` internally attempts to execute two subtasks (A1, A2).
> 3. Similarly, `Test B` internally attempts to execute two subtasks (B1, B2).
> 4. `A` starts `A1`, and `B` starts `B1`. At this point, all threads in the pool are occupied.
> 5. `A` waits for `A2` to complete, and `B` waits for `B2` to complete.
> 6. However, `A2` and `B2` cannot start because there are no available threads in the pool.
> 7. As a result, both `A` and `B` enter a deadlock, waiting indefinitely.
> 
> To prevent this issue, we could either set a timeout on `allOf.get()` (as a way to resolve the deadlock even if the test doesn't execute successfully) or separate the `ExecutorService` instances.
> 
> WDYT?
> 
> I’m also curious whether this issue will be included in version 5.12M1 or 5.13.

@marcphilipp 

I am eager to resolve this issue😀
If you have time, may I kindly ask for your response? Thank you in advance!

marcphilipp:
@YongGoose Sorry for the slow response! I was traveling for the last few days.

That's a good catch! For Jupiter, we use a `ForkJoinPool` to avoid this problem. Are you familiar with that?

YongGoose:
> @YongGoose Sorry for the slow response! I was traveling for the last few days.

Apologies if I came across as rude.
I didn’t know you were traveling. 😅
Hope you had a great trip!

> That's a good catch! For Jupiter, we use a `ForkJoinPool` to avoid this problem. Are you familiar with that?

I didn’t know that!
I’ll look into the part where ForkJoinPool is used in Jupiter.

marcphilipp:
> Apologies if I came across as rude.
> I didn’t know you were traveling. 😅

No worries!

> I’ll look into the part where ForkJoinPool is used in Jupiter.

Jupiter uses [`ForkJoinPoolHierarchicalTestExecutorService`](https://github.com/junit-team/junit5/blob/main/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java) from `junit-platform-engine`.

YongGoose:
> Jupiter uses [`ForkJoinPoolHierarchicalTestExecutorService`](https://github.com/junit-team/junit5/blob/main/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java) from `junit-platform-engine`.

I’m running into some challenges trying to use `ForkJoinPoolHierarchicalTestExecutorService` in Vintage.

Before moving forward, I’d like to get your feedback on the best approach here!

1. Update `VintageTestEngine` to use `ForkJoinPoolHierarchicalTestExecutorService`.
2. Create a `custom ForkJoinPool` and use that as the thread pool instead.

Which direction do you think makes more sense?

marcphilipp:
I think option 1 won't work because Vintage can't implement `HierarchicalTestEngine` so I'd go with option 2.

YongGoose:
> I think option 1 won't work because Vintage can't implement `HierarchicalTestEngine` so I'd go with option 2.

Thank you! 

I’ll proceed as you suggested.  

As for another PR I’m working on, I accidentally included commits from the main branch while resolving git conflicts. So, I’m planning to create a new PR 😅

YongGoose:
> I think option 1 won't work because Vintage can't implement `HierarchicalTestEngine` so I'd go with option 2.

I have changed the Thread pool to ForkJoinPool.  

As a result, the deadlock issue seems to be resolved. However, there is one thing I would like to discuss.  

In the case of `ForkJoinPoolHierarchicalTestExecutorService`, it efficiently processes tasks using control statements such as `ForkJoinTask.getSurplusQueuedTaskCount() < parallelism`.  

I would like to apply a similar approach to `JUnit Vintage Engine`, but it seems that a large-scale refactoring would be required. Also, I am concerned that this might lead to `over-engineering`.  

I am still not very confident in making such decisions.  
To what extent do you think optimization should be applied?

YongGoose:
@marcphilipp 

Congratulations on the release of JUnit 5.12.0-M1! 🥳

When you have time, could you please leave comments on my two PRs?
- https://github.com/junit-team/junit5/pull/4242#issuecomment-2620723295
- https://github.com/junit-team/junit5/pull/4281#issuecomment-2615571021

Thank you as always.

PS. #4281  is still far from perfect... If my code is still wrong, I would appreciate any constructive feedback and some reference code to help me improve. 🙂

marcphilipp:
> When you have time, could you please leave comments on my two PRs?

Yes, I'll get to it in the coming days. Sorry for the wait!

YongGoose:
> > I think option 1 won't work because Vintage can't implement `HierarchicalTestEngine` so I'd go with option 2.
> 
> I have changed the Thread pool to ForkJoinPool.
> 
> As a result, the deadlock issue seems to be resolved. However, there is one thing I would like to discuss.
> 
> In the case of `ForkJoinPoolHierarchicalTestExecutorService`, it efficiently processes tasks using control statements such as `ForkJoinTask.getSurplusQueuedTaskCount() < parallelism`.
> 
> I would like to apply a similar approach to `JUnit Vintage Engine`, but it seems that a large-scale refactoring would be required. Also, I am concerned that this might lead to `over-engineering`.
> 
> I am still not very confident in making such decisions. To what extent do you think optimization should be applied?

@marcphilipp 

I feel that my previous comment might not have conveyed enough information, so I'd like to provide some additional clarification.  

In `ForkJoinPoolHierarchicalTestExecutorService`, `ThreadLocal` is used to manage `ThreadLock` objects, which helps prevent deadlocks. Each `ThreadLock` object keeps track of the `ResourceLock` held by the current thread and a list of deferred tasks.  

This mechanism allows each thread to manage resources independently. Additionally, the logic for deferring tasks and executing them at the right moment ensures efficient processing.  

I originally wanted to implement a similar class in the JUnit Vintage engine to leverage these features. However, I started to feel that introducing such a complex class in JUnit Vintage might be over-engineering, which led me to leave this comment.  

To be honest, I’m also not entirely confident that I can execute this well. Any thoughts or suggestions would be greatly appreciated!

marcphilipp:
> In `ForkJoinPoolHierarchicalTestExecutorService`, `ThreadLocal` is used to manage `ThreadLock` objects, which helps prevent deadlocks. Each `ThreadLock` object keeps track of the `ResourceLock` held by the current thread and a list of deferred tasks.
> 
> This mechanism allows each thread to manage resources independently. Additionally, the logic for deferring tasks and executing them at the right moment ensures efficient processing.
> 
> I originally wanted to implement a similar class in the JUnit Vintage engine to leverage these features. However, I started to feel that introducing such a complex class in JUnit Vintage might be over-engineering, which led me to leave this comment.

It's quite a complex mechanism (implementation-wise). Moreover, it requires declaring exclusive resources "statically" (as part of the `TestDescriptor` tree that is returned from `TestEngine.discover`). The last bit is problematic for the Vintage engine, as the `TestDescriptor` tree depends on `Description` objects provided by the runner specified using `@RunWith`. Since we can't extend that API we'd have to find some other means to determine which class/method a `Descriptor` represents. We do this to determine the `TestSource` for each `TestDescriptor` but it's not as reliable as we'd need it to implement resource-based locking. Therefore, I think we should refrain from implementing that for the Vintage engine, at least at this time.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -92,32 +95,46 @@ public void execute(ExecutionRequest request) {
...
-		else {
+
+		if (!classes && !methods) {
+			logger.warn(() -> "Parallel execution is enabled but no scope is defined. "
+					+ "Falling back to sequential execution.");
 			executeSequentially(engineDescriptor, engineExecutionListener);
+			return;
+		}
+
+		if (executeInParallel(engineDescriptor, engineExecutionListener, request)) {
```
marcphilipp, line 113: I think it's more readable to give the returned `boolean` a name

```suggestion
		boolean wasInterrupted = executeInParallel(engineDescriptor, engineExecutionListener, request);
		if (wasInterrupted) {
```

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -92,32 +95,46 @@ public void execute(ExecutionRequest request) {
...
-			CompletableFuture<Void> future = CompletableFuture.runAsync(() -> {
-				runnerExecutor.execute((RunnerTestDescriptor) descriptor);
-			}, executorService);
+		List<RunnerTestDescriptor> runnerTestDescriptors = collectRunnerTestDescriptors(engineDescriptor,
+			executorService);
 
+		if (!classes) {
+			for (RunnerTestDescriptor runnerTestDescriptor : runnerTestDescriptors) {
+				runnerExecutor.execute(runnerTestDescriptor);
+			}
```
marcphilipp, line 129: Extract a method for readability?

```suggestion
			executeClassesSequentially(runnerTestDescriptors, runnerExecutor);
```

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -92,32 +95,46 @@ public void execute(ExecutionRequest request) {
...
+			return false;
+		}
+
+		List<CompletableFuture<Void>> futures = new ArrayList<>();
```
marcphilipp, line 133: I'd extract a method called `executeClassesInParallel` starting from this line and ending at the end of the method for readability.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -138,6 +155,27 @@ private boolean executeInParallel(VintageEngineDescriptor engineDescriptor,
...
+		for (TestDescriptor descriptor : engineDescriptor.getModifiableChildren()) {
+			RunnerTestDescriptor runnerTestDescriptor = (RunnerTestDescriptor) descriptor;
+
+			if (methods) {
+				runnerTestDescriptors.add(parallelMethodExecutor(runnerTestDescriptor, executorService));
+				continue;
+			}
+			runnerTestDescriptors.add(runnerTestDescriptor);
+		}
+		return runnerTestDescriptors;
```
marcphilipp, line 170:
```suggestion
		return engineDescriptor.getModifiableChildren().stream() //
				.map(RunnerTestDescriptor.class::cast) //
				.map(it -> methods ? parallelMethodExecutor(it, executorService) : it) //
				.collect(toList());
```

junit-vintage-engine/src/test/java/org/junit/vintage/engine/execution/ParallelExecutionIntegrationTests.java
```diff
@@ -67,6 +75,57 @@ void executesTestClassesInParallel(TestReporter reporter) {
...
+		reporter.publishEntry("finishedClassesTimestamps", finishedClassesTimestamps.toString());
+		reporter.publishEntry("startedMethodsTimestamps", startedMethodsTimestamps.toString());
+		reporter.publishEntry("finishedMethodsTimestamps", finishedMethodsTimestamps.toString());
+
+		assertThat(startedClassesTimestamps).hasSize(3);
+		assertThat(finishedClassesTimestamps).hasSize(3);
+		assertThat(startedMethodsTimestamps).hasSize(9);
+		assertThat(finishedMethodsTimestamps).hasSize(9);
+
+		assertThat(threadNames).hasSizeGreaterThanOrEqualTo(2);
```
marcphilipp, line 126: I think the problem here is that the `ForkJoinPool` is not stealing work when we're waiting for the `RunnerDescriptors` children (see next comment).

```suggestion
		assertThat(threadNames).hasSize(3);
```

junit-vintage-engine/src/main/java/org/junit/vintage/engine/descriptor/RunnerTestDescriptor.java
```diff
@@ -161,6 +168,35 @@ public boolean isIgnored() {
...
+						CompletableFuture.allOf(futures.toArray(new CompletableFuture<?>[0])).get();
+					}
+					catch (ExecutionException e) {
+						throw ExceptionUtils.throwAsUncheckedException(e.getCause());
+					}
+					catch (InterruptedException e) {
+						logger.warn(e, () -> "Interrupted while waiting for runner to finish");
+						Thread.currentThread().interrupt();
+					}
+				}
```
marcphilipp, line 195:
```suggestion
				private final List<Future<?>> futures = new CopyOnWriteArrayList<>();

				@Override
				public void schedule(Runnable childStatement) {
					futures.add(executorService.submit(childStatement));
				}

				@Override
				public void finished() {
					ThrowableCollector collector = new OpenTest4JAwareThrowableCollector();
					AtomicBoolean wasInterrupted = new AtomicBoolean(false);
					for (Future<?> future : futures) {
						collector.execute(() -> {
							// We're calling `Future.get()` individually to allow for work stealing
							// in case `ExecutorService` is a `ForkJoinPool`
							try {
								future.get();
							}
							catch (ExecutionException e) {
								throw e.getCause();
							}
							catch (InterruptedException e) {
								wasInterrupted.set(true);
							}
						});
					}
					collector.assertEmpty();
					if (wasInterrupted.get()) {
						logger.warn(() -> "Interrupted while waiting for runner to finish");
						Thread.currentThread().interrupt();
					}
				}
```

marcphilipp (changes requested):

junit-vintage-engine/src/test/java/org/junit/vintage/engine/execution/ParallelExecutionIntegrationTests.java
YongGoose: I did notice that the `ForkJoinPool` isn't stealing work, but I didn't realize that it was because of waiting for the `RunnerDescriptors`' children.  

Running the Futures separately seems like a great approach.  

I learned a lot from this—thank you so much! 🙇🏻‍♂️

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4242/commits/cd31e487ee17f2da0ec4780d9e4c5f062e5c0c8b !

YongGoose:
> It's quite a complex mechanism (implementation-wise). Moreover, it requires declaring exclusive resources "statically" (as part of the `TestDescriptor` tree that is returned from `TestEngine.discover`). The last bit is problematic for the Vintage engine, as the `TestDescriptor` tree depends on `Description` objects provided by the runner specified using `@RunWith`. Since we can't extend that API we'd have to find some other means to determine which class/method a `Descriptor` represents. We do this to determine the `TestSource` for each `TestDescriptor` but it's not as reliable as we'd need it to implement resource-based locking. Therefore, I think we should refrain from implementing that for the Vintage engine, at least at this time.

Thank you for the detailed explanation.  

I wasn't very familiar with the internal structure of the `Vintage engine`, so it was difficult to make a judgment, but your explanation helped me understand it better.  

As you advised, I will not implement this feature in the Vintage engine for now.  
If I determine that implementation is necessary in the future, I will address it through a separate issue.

YongGoose:
@marcphilipp 

I have completed the documentation and changed the PR status from `Draft` to `Ready for Review`. 
If there's anything I haven't done yet or any additional work needed, please feel free to let me know.

PS. I don't think I could have completed this work without your help.  
I sincerely appreciate it.

documentation/src/docs/asciidoc/release-notes/release-notes-5.12.0-RC1.adoc
```diff
@@ -41,4 +41,9 @@ repository on GitHub.
 [[release-notes-5.12.0-RC1-junit-vintage]]
 === JUnit Vintage
 
-No changes.
+[[release-notes-5.12.0-RC1-junit-vintage-new-features-and-improvements]]
+==== New Features and Improvements
+
+* Support for executing test methods in parallel. Please refer to the
+<<../user-guide/index.adoc#migrating-from-junit4-parallel-execution, User Guide>> for
+more information.
```
marcphilipp, line 49: RC1 is already released. I've scheduled 5.12.0-RC2 for this Friday. Could you please move this entry to the RC2 release notes? I'll then merge it with the M1 entry for parallel execution when preparing the release notes for 5.12.0.

marcphilipp (changes requested):

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -92,32 +96,64 @@ public void execute(ExecutionRequest request) {
 
 	private void executeAllChildren(VintageEngineDescriptor engineDescriptor,
```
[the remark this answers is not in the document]
YongGoose: Done in https://github.com/junit-team/junit5/pull/4242/commits/2b4fbeebbf1edeef990d29bddbfac7e6a141224c !

documentation/src/docs/asciidoc/release-notes/release-notes-5.12.0-RC1.adoc
YongGoose: I don't see the 5.12.0-RC2 release notes. 
Should I create them first and then move?

sbrannen: Yes, we create the new release notes file from a template first.

You can mimic what was done for 5.12.0-RC1: https://github.com/junit-team/junit5/commit/b04fe73d2fd5dafb88c9ef9a5c79b685e310cb0d

This is the template: https://github.com/junit-team/junit5/blob/main/documentation/src/docs/asciidoc/release-notes/release-notes-TEMPLATE.adoc

junit-vintage-engine/src/main/java/org/junit/vintage/engine/VintageTestEngine.java
```diff
@@ -92,32 +96,64 @@ public void execute(ExecutionRequest request) {
...
+			return;
+		}
 
-		if (parallelExecutionEnabled) {
-			if (executeInParallel(engineDescriptor, engineExecutionListener, request)) {
-				Thread.currentThread().interrupt();
-			}
+		if (!classes && !methods) {
+			logger.warn(() -> "Parallel execution is enabled but no scope is defined. "
+					+ "Falling back to sequential execution.");
```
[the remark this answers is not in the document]
YongGoose: Done in https://github.com/junit-team/junit5/pull/4242/commits/9c7e81a56d4763c0628aec8f2a121115a25754bf !

documentation/src/docs/asciidoc/release-notes/release-notes-5.12.0-RC1.adoc
YongGoose: Done in https://github.com/junit-team/junit5/pull/4242/commits/5bfdd18c84e6b763c7b809e104b313f9b79b2088 !

YongGoose: @sbrannen 

It looks like Marc has already created it. 
I didn’t check the main branch. 😅

I’ll delete it and reflect the changes again.

documentation/src/docs/asciidoc/release-notes/release-notes-5.12.0-RC2.adoc
```diff
@@ -0,0 +1,69 @@
+[[release-notes-5.12.0-RC2]]
+== 5.12.0-RC2
+
+*Date of Release:* February 14, 2025
+
+*Scope:* Minor enhancements since JUnit 5.12 RC1.
+
+For a complete list of all _closed_ issues and pull requests for this release, consult the
+link:{junit5-repo}+/milestone/2?closed=1+[5.12.0-RC2] milestone page in the
```
sbrannen, line 9:
```suggestion
link:{junit5-repo}+/milestone/90?closed=1+[5.12.0-RC2] milestone page in the
```

sbrannen (changes requested):

YongGoose:
@marcphilipp  @sbrannen 

Both of you, thank you!  🙂
Thanks to you, the document editing process went smoothly.

documentation/src/docs/asciidoc/release-notes/release-notes-5.12.0-RC1.adoc
marcphilipp: I should have said that I just pushed that commit to `main`. Sorry about that!

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -39,29 +39,37 @@ discovered tests before executing them (see <<running-tests>> for details).
...
 
-The JUnit Vintage test engine supports parallel execution of top-level test classes,
+The JUnit Vintage test engine supports parallel execution of top-level test classes and test methods,
 allowing existing JUnit 3 and JUnit 4 tests to benefit from improved performance through
 concurrent test execution. It can be enabled and configured using the following
 <<running-tests-config-params, configuration parameters>>:
 
 `junit.vintage.execution.parallel.enabled=true|false`::
-  Enable/disable parallel execution (defaults to `false`).
+Enable/disable parallel execution (defaults to `false`).
```
marcphilipp, line 48:
```suggestion
  Enable/disable parallel execution (defaults to `false`). Requires opt-in for `classes`
  or `methods` to be executed in parallel using the configuration parameters below.
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -39,29 +39,37 @@ discovered tests before executing them (see <<running-tests>> for details).
...
 ----
 junit.vintage.execution.parallel.enabled=true
+junit.vintage.execution.parallel.classes=true
+junit.vintage.execution.parallel.methods=true
 junit.vintage.execution.parallel.pool-size=4
 ----
 
-With these properties set, the `VintageTestEngine` will execute tests in parallel,
-potentially significantly reducing the overall test suite execution time.
-
```
marcphilipp, line 64: I think this paragraph should stay _after_ the snippet. Otherwise, the `:` at the end of the paragraph above is in the wrong place.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/VintageExecutor.java
```diff
@@ -0,0 +1,184 @@
...
+	private static final Logger logger = LoggerFactory.getLogger(VintageExecutor.class);
+
+	private static final int DEFAULT_THREAD_POOL_SIZE = Runtime.getRuntime().availableProcessors();
+	private static final int SHUTDOWN_TIMEOUT_SECONDS = 30;
+
+	private boolean classes;
+	private boolean methods;
+
+	public void executeAllChildren(VintageEngineDescriptor engineDescriptor,
+			EngineExecutionListener engineExecutionListener, ExecutionRequest request) {
```
marcphilipp, line 54: I would pass these three parameters to the constructor so they don't have to be passed to the other methods in this case thereby simplifying the code.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/VintageExecutor.java
```diff
@@ -0,0 +1,184 @@
...
+		boolean wasInterrupted = executeInParallel(engineDescriptor, engineExecutionListener, request);
+		if (wasInterrupted) {
+			Thread.currentThread().interrupt();
+		}
+	}
+
+	private void initializeParallelExecutionParameters(ExecutionRequest request) {
+		classes = request.getConfigurationParameters().getBoolean(Constants.PARALLEL_CLASS_EXECUTION).orElse(false);
+		methods = request.getConfigurationParameters().getBoolean(Constants.PARALLEL_METHOD_EXECUTION).orElse(false);
+	}
```
marcphilipp, line 79: This should be done in the constructor so the fields can be `final`.

junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/VintageExecutor.java
```diff
@@ -0,0 +1,184 @@
...
+
+	private boolean getParallelExecutionEnabled(ExecutionRequest request) {
+		return request.getConfigurationParameters().getBoolean(PARALLEL_EXECUTION_ENABLED).orElse(false);
```
marcphilipp, line 82: I would also do this in the constructor.

marcphilipp (changes requested):

junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/VintageExecutor.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4242/commits/9f6670e9d4a5ee72ae3e9538b8d37c71e8940440 !
Additionally, I added a log for when `poolSize` is negative and created a corresponding test case.

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
YongGoose: Done in https://github.com/junit-team/junit5/pull/4242/commits/21e8cd13c3e54c090e628233121c31500780f0e2 !

marcphilipp (approved):

marcphilipp:
@YongGoose Once again, thanks for all your work! :+1:

PR #4279, opened 2025-01-27
Support `@TempDir` constructor injection for Java record classes

marcphilipp:
The `@TempDir` annotation on a constructor parameter is copied to the
generated `final` instance field. This caused it top be picked up by
`TempDirectory` for instance field injection which then failed because
the field was `final`. This change now checks if a test class is an
instance of a record type and skips instance field injection. Since
records cannot have any instance fields besides the ones generated from
their constructor, there's no need to inject values into any of them.

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
@@ -366,6 +366,24 @@ public final class ReflectionUtils {
 	}
 
 	/**
+	 * {@return whether the supplied {@code object} is an instance of a record class}
+	 * @since 1.12
+	 */
+	@API(status = INTERNAL, since = "1.12")
+	public static boolean isRecordObject(Object object) {
+		return object != null && isRecordClass(object.getClass());
+	}
+
+	/**
+	 * {@return whether the supplied {@code clazz} is a record class}
+	 * @since 1.12
+	 */
+	@API(status = INTERNAL, since = "1.12")
+	public static boolean isRecordClass(Class<?> clazz) {
+		return "java.lang.Record".equals(clazz.getSuperclass().getName());
+	}
+
+	/**
 	 * Determine if the return type of the supplied method is primitive {@code void}.
 	 *
 	 * @param method the method to test; never {@code null}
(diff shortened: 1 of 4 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -1993,6 +1993,24 @@ public static Set<Class<?>> getAllAssignmentCompatibleClasses(Class<?> clazz) {
...
+		return object != null && isRecordClass(object.getClass());
+	}
+
+	/**
+	 * {@return whether the supplied {@code clazz} is a record class}
+	 * @since 1.12
+	 */
+	@API(status = INTERNAL, since = "1.12")
+	public static boolean isRecordClass(Class<?> clazz) {
+		return "java.lang.Record".equals(clazz.getSuperclass().getName());
```
sbrannen, line 2011: `clazz.getSuperclass()` can return `null`.

So, you'll need to defensively check for that to avoid a `NullPointerException`.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -1993,6 +1993,24 @@ public static Set<Class<?>> getAllAssignmentCompatibleClasses(Class<?> clazz) {
 		return result;
 	}
 
+	/**
+	 * {@return whether the supplied {@code object} is an instance of a record class}
+	 * @since 1.12
+	 */
+	@API(status = INTERNAL, since = "1.12")
+	public static boolean isRecordObject(Object object) {
```
sbrannen, line 2001: Let's move these new methods to directly after `isInnerClass()`,

sbrannen (changes requested):

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
marcphilipp: Done in 6c020d50294c98fdee506eb768f5765761c9052e

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
marcphilipp: Done in 58cb54143017afb945b4b6e67f5f069f34a1d735

PR #4281, opened 2025-01-27
Provide mechanism for managing resources across engines and executions

### LINKED ISSUE #2816: Provide mechanism for managing resources across engines and executions
The Platform should provide an extensible mechanism (`ServiceLoader`?) so that user code can create "closeable resources" that are injected as part of the context received by `TestEngine` implementations.

This feature is analogous to the `ExtensionContext.Store` concept from Jupiter but takes it to the Platform level. Moreover, it allows to reuse the same resource across multiple executions in a `LauncherSession`. We should advocate that resources stored in such a key-value store be immutable until their destruction. If such resources implement `AutoCloseable` they will get closed immediately before the `LauncherSession` they belong to is closed.

YongGoose:
Resolves #2816 

## Overview

- [x] Designing a class to manage context at the platform level 
- [x] Adding a session-level store to LauncherSession 
- [x] Adding a request-level store to Launcher
- [x] Exposing the request-level store in ~EngineDiscoveryRequest and TestPlan~ ExecutionRequest
- [x] Utilizing the request-level store in ExtensionContext
- [x] Tests for Jupiter, Suite engine, EngineTestKit etc.

---

---

### Definition of Done


YongGoose:
@marcphilipp 

> What's the use case for this method? Test engines don't have access to the Launcher so we'll need to pass this store as part of LauncherDiscoveryRequest and ExecutionRequest, don't we?

Done in https://github.com/junit-team/junit5/pull/4281/commits/02ef45b979b157e681de26626de2db0ece40d6d8 !

PS. While resolving a git conflict, I accidentally included the commit history from the main branch, So i create a new PR. 😁

YongGoose:
@marcphilipp 

I think you must be very busy with the release of version 5.12 M1. 
Please feel free to leave your comments whenever it’s convenient for you.

YongGoose:
@marcphilipp 

This PR is something I’ve personally been eager to work on for a long time.
However, I still don’t have a clear sense of how to proceed with it.

If this feature gets added to JUnit 5, it would be a huge learning opportunity for me and a great addition to my skill set.
That said, I don’t want to cause any fatigue for the JUnit team—especially you, Marc.

So if you feel that I’m not quite ready to take on this PR yet, please let me know, and I’ll close it.  
Either way, I’ll keep looking for other issues and do my best to contribute! 🙂

> But if you’re willing to let me see it through, I’ll do my best as always.
I’ll likely need more guidance than usual compared to other tasks.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -418,6 +418,28 @@ default void publishReportEntry(String value) {
...
+	/**
+	 * Returns the store for session-level data.
+	 *
+	 * <p>This store is used to store data that is scoped to the session level.
+	 * The data stored in this store will be available throughout the entire session.
+	 *
+	 * @return the store for session-level data
+	 * @since 5.13
+	 */
+	Store getSessionLevelStore();
```
marcphilipp, line 430: These also need an `@API` annotation:
```java
@API(status = EXPERIMENTAL, since = "5.13")
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -85,6 +86,18 @@ public NamespacedHierarchicalStore<N> newChild() {
...
 
+	/**
+	 * Returns the parent store of this {@code NamespacedHierarchicalStore}.
+	 *
+	 * <p>If this store does not have a parent, an empty {@code Optional} is returned.
+	 *
+	 * @return an {@code Optional} containing the parent store, or an empty {@code Optional} if there is no parent
+	 * @since 5.13
+	 */
+	public Optional<NamespacedHierarchicalStore<N>> getParent() {
```
marcphilipp, line 97: Also needs an `@API` annotation

junit-platform-engine/src/main/java/org/junit/platform/engine/Namespace.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.apiguardian.api.API;
+import org.junit.platform.commons.util.Preconditions;
+
+@API(status = STABLE, since = "5.13")
```
marcphilipp, line 23: We usually introduce new APIs as `EXPERIMENTAL` in their initial release

```suggestion
@API(status = EXPERIMENTAL, since = "5.13")
```

junit-platform-engine/src/main/java/org/junit/platform/engine/Namespace.java
```diff
@@ -0,0 +1,85 @@
...
+ * Copyright 2015-2025 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.platform.engine;
```
marcphilipp, line 11: Can we move this to the same packaged as `NamespacedHierarchicalStore`?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -59,6 +65,8 @@ class DefaultLauncher implements Launcher {
 			"PostDiscoveryFilter array must not contain null elements");
 		this.discoveryOrchestrator = new EngineDiscoveryOrchestrator(testEngines,
 			unmodifiableCollection(postDiscoveryFilters), listenerRegistry.launcherDiscoveryListeners);
+		this.launcherSession = LauncherFactory.openSession(config);
```
marcphilipp, line 68: We should not open a session here but pass in the `sessionStore` to the constructor.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncherSession.java
```diff
@@ -43,6 +45,7 @@ public void close() {
 	private final LauncherInterceptor interceptor;
 	private final LauncherSessionListener listener;
 	private final DelegatingLauncher launcher;
+	private final NamespacedHierarchicalStore<Namespace> store;
 
 	DefaultLauncherSession(List<LauncherInterceptor> interceptors, Supplier<LauncherSessionListener> listenerSupplier,
 			Supplier<Launcher> launcherSupplier) {
```
marcphilipp, line 51: This probably needs to be changed to the following:

```suggestion
			Function<NamespacedHierarchicalStore<Namespace>, Launcher> launcherFactory) {
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4281/commits/33f5fb82eafedf8a2e29b7ea7f6fe16b78d6fe2d

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4281/commits/33f5fb82eafedf8a2e29b7ea7f6fe16b78d6fe2d

junit-platform-engine/src/main/java/org/junit/platform/engine/Namespace.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4281/commits/9fd6fd8c15a9af9944b64d59db2d5daf767f8143 !

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncherSession.java
marcphilipp: This is to make creating the `Launcher` lazy so it can be intercepted (see `else` branch below).

marcphilipp: Since it now needs to be passed the session-level store to its constructor, I think we'll need to pass that in.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.java
```diff
@@ -192,6 +192,16 @@ public Store getStore(Namespace namespace) {
...
 
+	@Override
+	public Store getSessionLevelStore() {
+		return getStore(Namespace.create("session"));
+	}
+
+	@Override
+	public Store getRequestLevelStore() {
+		return getStore(Namespace.create("request"));
+	}
```
YongGoose, line 203: @marcphilipp 

Is this the correct approach to creating session and request stores?  

I understand that a `Namespace` is distinguished based on its name. However, I'm not sure if it's appropriate to create a `Namespace` by directly using the names of the `session` and `request`.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncherSession.java
YongGoose: Based on the comments you left on the PR and the comments in the issue, I've made some progress on the implementation.  
- https://github.com/junit-team/junit5/issues/2816#issuecomment-2508195159
- https://github.com/junit-team/junit5/issues/2816#issuecomment-2567630332

I've marked the parts where I'm still unsure about how to pass things with `TODOs`.  

Thanks to your help, I'm starting to get a better understanding. I'll keep working hard. Thank you!

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.java
marcphilipp: No, the namespace is an orthogonal concept here. Instead these methods should take a `Namespace` parameter. Sorry for omitting that in https://github.com/junit-team/junit5/issues/2816#issuecomment-2567630332! They should return an adapted (they use a different `Namespace` type and `Store` interface) version of the `NamespacedHierarchicalStore` objects passed to the Jupiter engine. Please let me know if I should provide more details.

YongGoose: > Sorry for omitting that in https://github.com/junit-team/junit5/issues/2816#issuecomment-2567630332!

No, I think I should have asked earlier. 🥹

> No, the namespace is an orthogonal concept here.

May I ask about the concept of `Namespace`?

Also, how does the `Namespace` in `jupiter.api.extension` differ from the one I created?
A more overall detailed explanation would be really helpful for my understanding. 🙂

YongGoose: @marcphilipp 

I studied on my own and found some answers.  

`Namespace` is used to define the scope of data within a `Store`.  
From what I understand, Jupiter uses a different `Store`, which is why it also requires a different type of `Namespace`.  And as you suggested, I’ve modified the method to take `Namespace` as a parameter. Please review it when you have a chance. https://github.com/junit-team/junit5/pull/4281/commits/5d2c2b8ef26c4444d828354c653c3b5a01718b38

However, I still don’t fully understand the `NamespacedHierarchicalStore` that is passed to the Jupiter engine.

marcphilipp: > `Namespace` is used to define the scope of data within a `Store`.

That's correct. The original intention was that Jupiter extensions could store data in the `Store` without worrying about unintended side-effects (for example, another extension storing something else under the same key) while also allowing extensions to share state by using the same `Namespace` on purpose.

> From what I understand, Jupiter uses a different `Store`, which is why it also requires a different type of `Namespace`.

That's correct, but Jupiter's implementation uses `NamespacedHierarchicalStore` internally.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
```diff
@@ -63,6 +65,8 @@
 @API(status = STABLE, since = "1.0")
 public class LauncherFactory {
 
+	private static final NamespacedHierarchicalStore<Namespace> SESSION_STORE = new NamespacedHierarchicalStore<>(null);
```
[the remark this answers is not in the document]
marcphilipp: Since the stores need to be closed, we need to attach them to the `LauncherSession`. I pushed 60f66f54b998aee551274b5aaf9e1ec72472a8bc to illustrate what I mean.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.java
marcphilipp: I pushed 4624808c77709376597d753f6730150c710c0fbd to outline what I mean. `ExecutionRequest` is still using the wrong `NamespacedHierarchicalStore`, though, and I didn't add any tests.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherFactory.java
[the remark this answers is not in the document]
YongGoose: That was exactly the approach I had in mind!

Thank you so much. 🙇🏻

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.java
YongGoose: > `ExecutionRequest` is still using the wrong `NamespacedHierarchicalStore`, though, and I didn't add any tests.

I’ll take care of that task myself!

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/EngineExecutionOrchestrator.java
```diff
@@ -42,13 +44,17 @@
 public class EngineExecutionOrchestrator {
 
 	private final ListenerRegistry<TestExecutionListener> listenerRegistry;
+	private final NamespacedHierarchicalStore<Namespace> requestStore;
 
+	// TODO[Question]: Is it okay to have a null value for requestStore?
 	public EngineExecutionOrchestrator() {
-		this(ListenerRegistry.forTestExecutionListeners());
+		this(ListenerRegistry.forTestExecutionListeners(), null);
```
YongGoose, line 51: @marcphilipp 

The tests are currently failing.
I suspect that the issue is caused by null being passed to requestStore, but I'm not sure how to fix it. 

Could you give me a hint?

YongGoose: @marcphilipp 

Whenever you have time, could you give me some feedback on my comment? 🙂

marcphilipp: Sure, I'll take a look as soon as possible. Sorry for the delay!

marcphilipp: > The tests are currently failing.
> I suspect that the issue is caused by null being passed to requestStore, but I'm not sure how to fix it.

I think you're right. The check that fails is here:

https://github.com/junit-team/junit5/blob/6b4ee039c1877640791ee0a84c4530f485c2623f/junit-platform-engine/src/main/java/org/junit/platform/engine/ExecutionRequest.java#L108

I think we have two options:

1. Make both request- and session-level stores `Optional` in `ExtensionContext` etc.
2. Ensure both are always present

I'm leaning towards (2). In `SuiteLauncher` we should pass the request-level store from the `ExecutionRequest` that is passed to `SuiteTestEngine`. In `EngineTestKit`, we should mimick create both session- and request-level stores and close them `TestEngine.execute` returns.

To simplify things, I think we should remove the stores from discovery again (meaning remove them from `EngineDiscoveryRequest` etc.) and only pass them `TestEngine.execute()` as part of the `ExecutionRequest`. Moreover, it should let us deal with `SessionPerRequestLauncher` better where a session-level store doesn't really make sense.

marcphilipp: I pushed e0a43efae107af9a76609546e03b825dcc375c9e that does that. WDYT?

platform-tests/src/test/java/org/junit/platform/launcher/core/DefaultLauncherTests.java
```diff
@@ -152,8 +152,8 @@ public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId
 		assertThat(testPlan.getRoots()).hasSize(1);
 		var engineIdentifier = getOnlyElement(testPlan.getRoots());
 		assertThat(getOnlyElement(testPlan.getRoots()).getDisplayName()).isEqualTo("my-engine-id");
-		verify(discoveryListener).launcherDiscoveryStarted(request);
-		verify(discoveryListener).launcherDiscoveryFinished(request);
+		verify(discoveryListener).launcherDiscoveryStarted(any());
+		verify(discoveryListener).launcherDiscoveryFinished(any());
```
marcphilipp, line 156: I think this change can now be reverted.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
```diff
@@ -100,7 +106,13 @@ private LauncherDiscoveryResult discover(LauncherDiscoveryRequest discoveryReque
...
 
 	private void execute(InternalTestPlan internalTestPlan, TestExecutionListener[] listeners) {
-		executionOrchestrator.execute(internalTestPlan, listeners);
+		try (NamespacedHierarchicalStore<Namespace> requestLevelStore = createRequestLevelStore()) {
+			executionOrchestrator.execute(internalTestPlan, requestLevelStore, listeners);
+		}
+	}
+
+	private NamespacedHierarchicalStore<Namespace> createRequestLevelStore() {
+		return new NamespacedHierarchicalStore<>(sessionLevelStore, closeAutoCloseables());
```
YongGoose, line 115: Was there a specific reason you removed the `Namespace`?

image

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncherSession.java
```diff
@@ -73,11 +83,17 @@ LauncherSessionListener getListener() {
...
+			store.close();
 			listener.launcherSessionClosed(this);
 			interceptor.close();
 		}
 	}
 
+	@Override
+	public NamespacedHierarchicalStore<Namespace> getStore() {
+		return store;
+	}
```
YongGoose, line 95:
```suggestion
	@Override
	public NamespacedHierarchicalStore<Namespace> getSessionLevelStore() {
		return sessionLevelStore;
	}
```

WDYT?

YongGoose (commented):
Removing the store from the discovery process sounds like a good idea.  
As the next step, I will proceed with the following task.
- Tests for Jupiter, Suite engine, EngineTestKit etc.


Thank you, Marc!

platform-tests/src/test/java/org/junit/platform/launcher/core/DefaultLauncherTests.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4281/commits/62bcb8f167dfccb891a2a76ff7c7afd2b3dfe45d !

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
marcphilipp: Since the scope of the request-level store is now only a call to `execute`, I've changed the code to use a try-with-resources block. Therefore, it closes its own store and no longer needs to attach it to the session-level store to ensure it gets closed.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/LauncherSession.java
```diff
@@ -47,4 +49,6 @@ public interface LauncherSession extends AutoCloseable {
 	@Override
 	void close();
 
+	NamespacedHierarchicalStore<Namespace> getStore();
```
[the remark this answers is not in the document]
marcphilipp: Ditto

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncher.java
YongGoose: Thank you for the detailed explanation :)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncherSession.java
YongGoose: The store in `DefaultLauncherSession` is only used for creating the `Launcher`.  
So, setting it as `store` would be fine, but for consistency, I set it as `sessionLevelStore`.  

Hmm... I can't think of any other approach at the moment. 😬
For now, shall we put this on hold and decide once we have a clearer answer?

YongGoose:
@marcphilipp 

While working on the tests, I have a question.  

As per your guidance, I am adding tests for `JupiterEngine`, `SuiteEngine`, and `EngineKit`. However, I am wondering if `SuiteEngine` requires furthermore testing.  

Since the main focus of this PR is adding **request-level and session-level stores**, I am designing the tests around that.  

The `execute` method in `SuiteEngine` utilizes the `requestLevelStore` from `ExecutionRequest` to run multiple tests. In my view, the core role of `execute` is to execute multiple tests, and this functionality is already being validated by existing tests.  

Do you think any additional tests are needed for `SuiteEngine`?

PS. I truly appreciate how you not only explain things but also implement the code to make communication more efficient—it’s been incredibly helpful. 👍🏻
However, this time, I’d like to try implementing it on my own! 🙂

marcphilipp:
> The `execute` method in `SuiteEngine` utilizes the `requestLevelStore` from `ExecutionRequest` to run multiple tests. In my view, the core role of `execute` is to execute multiple tests, and this functionality is already being validated by existing tests.
> 
> Do you think any additional tests are needed for `SuiteEngine`?

I was thinking about a very basic tests that verifies that `SuiteEngine` passes along the `requestLevelStore` to test engines. It's probably sufficient to mock `SuiteLauncher` and check that a non-null store is passed to it.

junit-platform-launcher/src/testFixtures/java/org/junit/platform/launcher/core/NamespacedHierarchicalStoreProviders.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.junit.platform.launcher.core;
+
+import static org.junit.platform.engine.support.store.NamespacedHierarchicalStore.CloseAction.closeAutoCloseables;
+
+import org.junit.platform.engine.support.store.Namespace;
+import org.junit.platform.engine.support.store.NamespacedHierarchicalStore;
+
+public class NamespacedHierarchicalStoreProviders {
+	private static final NamespacedHierarchicalStore<Namespace> store = new NamespacedHierarchicalStore<>(null);
```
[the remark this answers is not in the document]
YongGoose: Done in https://github.com/junit-team/junit5/pull/4281/commits/6b6ca12a366ed653dd48ba721e83d599d13e0d07 !

YongGoose:
> I was thinking about a very basic tests that verifies that `SuiteEngine` passes along the `requestLevelStore` to test engines. It's probably sufficient to mock `SuiteLauncher` and check that a non-null store is passed to it.

As you mentioned in the previous comment, I verified that the store is correctly passed using mock.
- In addition to SuiteTestEngine, I have added tests for other parts that I thought required testing.

What do you think? @marcphilipp

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -744,6 +768,11 @@ public Namespace append(Object... parts) {
 			Collections.addAll(newParts, parts);
 			return new Namespace(newParts);
 		}
+
+		@API(status = INTERNAL, since = "5.13")
+		public Object[] getParts() {
+			return parts.toArray();
```
marcphilipp, line 774: When we switch to JDK 17 in 6.0, returning a `List` here will allow us to avoid copying since we can use `List.copyOf()`. Thus, I think we should do so right away and introduce a `org.junit.platform.engine.support.store.Namespace.create(List)` method.

```suggestion
		public List<Object> getParts() {
			return unmodifiableList(parts);
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/NamespaceAwareStore.java
```diff
@@ -27,12 +26,12 @@
  * @since 5.0
  */
 @API(status = INTERNAL, since = "5.0")
-public class NamespaceAwareStore implements Store {
+public class NamespaceAwareStore<N> implements Store {
```
marcphilipp, line 29: I think the type parameter I introduced here can be removed again. We only ever use the `Namespace` from junit-platform-engine in production code.

marcphilipp (commented):
> As you mentioned in the previous comment, I verified that the store is correctly passed using mock.

Looking good, great work so far! :+1: 

I think what we're missing is tests for the following scenarios:

- A Jupiter extension can read a value a `LauncherSessionListener` stored in the session-level store in its `launcherSessionOpened` method (see `LauncherFactoryTests`).
- Two dummy test engines (see `TestEngineStub` or `TestEngineSpy`) can share values via the request-level store. The first engine stores data, the second one reads it.

Do you have enough information to add those?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -418,6 +419,29 @@ default void publishReportEntry(String value) {
...
+	/**
+	 * Returns the store for request-level data.
+	 *
+	 * <p>This store is used to store data that is scoped to the request level.
+	 * The data stored in this store will be available only for the duration of the current request.
+	 *
+	 * @return the store for request-level data
+	 * @since 5.13
+	 */
+	Store getRequestLevelStore(Namespace namespace);
```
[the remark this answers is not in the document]
YongGoose: Done in https://github.com/junit-team/junit5/pull/4281/commits/bb3a61dbcc55e24d4035355a02cbcaf8d42909ba !

YongGoose:
> Do you have enough information to add those?

I don’t have enough information about the two dummy test engines, but I believe I can do it. ⚡️
If I need any help, I’ll ask with a comment.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4281/commits/78a6023867f039fa78ef75391d5953dda4cdb660 !
Would it be okay to ask if this aligns with your intention? 🧐

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/Namespace.java
```diff
@@ -0,0 +1,101 @@
...
+	 * which use the same sequence of {@code objects} for creating a namespace.
+	 *
+	 * <p>The order of the {@code objects} is significant.
+	 *
+	 * <p>Internally the {@code objects} are compared using {@link Object#equals(Object)}.
+	 */
+	public static Namespace create(List<Object> objects) {
+		Preconditions.notEmpty(objects, "objects list must not be null or empty");
+		Preconditions.containsNoNullElements(objects, "individual objects must not be null");
+		return new Namespace(unmodifiableList(objects));
```
marcphilipp, line 59: We need to create a copy here to avoid being affected by any potential change to the original list.

```suggestion
		return new Namespace(new ArrayList<>(objects));
```

YongGoose: Got it!

I’ll make the same changes to the other classes as well.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/NamespaceAwareStore.java
marcphilipp: When a type parameter is only used by tests, we should rather refactor the tests. In this case, they should be able to work with the `Namespace` from junit-platform-engine.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/Namespace.java
```diff
@@ -0,0 +1,100 @@
...
+import org.junit.platform.commons.util.Preconditions;
+
+@API(status = EXPERIMENTAL, since = "5.13")
+public class Namespace {
+
+	/**
+	 * The default, global namespace which allows access to stored data from
+	 * all extensions.
+	 */
+	public static final Namespace GLOBAL = Namespace.create(new Object());
```
marcphilipp, line 31: I think `ExtensionContext.Namespace.GLOBAL` needs to be created from this in order for the conversion in `LauncherStoreFacade` to work correctly and we should add a test for that.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/LauncherStoreFacade.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	NamespaceAwareStore getStoreAdapter(NamespacedHierarchicalStore<Namespace> valuesStore,
+			ExtensionContext.Namespace namespace) {
+		Preconditions.notNull(namespace, "Namespace must not be null");
+		return new NamespaceAwareStore(valuesStore, convert(namespace));
+	}
+
+	private Namespace convert(ExtensionContext.Namespace namespace) {
+		return Namespace.create(namespace.getParts());
```
marcphilipp, line 54:
```suggestion
		return namespace.equals(ExtensionContext.Namespace.GLOBAL) //
				? Namespace.GLOBAL //
				: Namespace.create(namespace.getParts());
```

jupiter-tests/src/test/java/org/junit/jupiter/engine/descriptor/LauncherStoreFacadeTest.java
```diff
@@ -0,0 +1,104 @@
...
+	private NamespacedHierarchicalStore<Namespace> sessionLevelStore;
+	private ExtensionContext.Namespace extensionNamespace;
+
+	@BeforeEach
+	@SuppressWarnings("unchecked")
+	void setUp() {
+		requestLevelStore = mock(NamespacedHierarchicalStore.class);
+		sessionLevelStore = mock(NamespacedHierarchicalStore.class);
+		extensionNamespace = ExtensionContext.Namespace.create("test");
+		when(requestLevelStore.getParent()).thenReturn(Optional.of(sessionLevelStore));
```
marcphilipp, line 50:
```suggestion
		sessionLevelStore = new NamespacedHierarchicalStore<>(null);
		requestLevelStore = new NamespacedHierarchicalStore<>(sessionLevelStore);
```

jupiter-tests/src/test/java/org/junit/jupiter/engine/descriptor/LauncherStoreFacadeTest.java
```diff
@@ -0,0 +1,104 @@
...
+	@Test
+	void createsInstanceSuccessfullyWithValidStore() {
+		assertDoesNotThrow(() -> new LauncherStoreFacade(requestLevelStore));
+	}
+
+	@Test
+	void throwsExceptionWhenRequestLevelStoreHasNoParent() {
+		when(requestLevelStore.getParent()).thenReturn(Optional.empty());
+
+		assertThrowsExactly(JUnitException.class, () -> new LauncherStoreFacade(requestLevelStore), () -> {
```
marcphilipp, line 62:
```suggestion
		assertThrowsExactly(JUnitException.class, () -> new LauncherStoreFacade(sessionLevelStore), () -> {
```

marcphilipp (changes requested):
This is what I meant regarding the `GLOBAL` namespace. We can't create the Jupiter one from the junit-platform-engine one (due to it not being visible there) but we can check for it when converting the namespaces.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/Namespace.java
marcphilipp: see https://github.com/junit-team/junit5/pull/4281#pullrequestreview-2671378936

marcphilipp:
No worries! Good luck for your presentation!

YongGoose:
@marcphilipp 

Currently, the dummyTest in JupiterTest is failing. This happens because JupiterTest runs in isolation, attempting to retrieve a value from session-level-store without first inserting it, resulting in a NullPointerException.

One possible solution is to modify JupiterExtensionExample to allow null values to pass, but I don’t think this is the best approach.

WDYT?

marcphilipp:
The convention we have adopted after `LauncherFactoryTests` was initially written is to name test classes that should not be run directly (for example, because they're used by a test via `EngineTestKit` or directly via the `Launcher`) with a `TestCase` suffix. So I'd suggest renaming `JupiterTest` to `JupiterTestCase` accordingly. This also applies to the `JUnit4Example` and `JUnit5Example` classes.

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
```diff
@@ -333,6 +338,27 @@ public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult
 		}));
 	}
 
+	@Test
+	void extensionCanReadValueFromSessionStore() {
+		var config = LauncherConfig.builder().addLauncherSessionListeners(
+			new LauncherSessionListenerExample()).enableTestEngineAutoRegistration(true).build();
```
marcphilipp, line 344:
```suggestion
		var config = LauncherConfig.builder() //
				.addLauncherSessionListeners(new LauncherSessionListenerExample()) //
				.build();
```

YongGoose:
> The convention we have adopted after `LauncherFactoryTests` was initially written is to name test classes that should not be run directly (for example, because they're used by a test via `EngineTestKit` or directly via the `Launcher`) with a `TestCase` suffix. This avoids them being picked up due to this config:
> 
> https://github.com/junit-team/junit5/blob/0f585b245fe908358b71471a4981091c3b8e405d/gradle/plugins/common/src/main/kotlin/junitbuild.testing-conventions.gradle.kts#L67
> 
> So I'd suggest renaming `JupiterTest` to `JupiterTestCase` accordingly. This also applies to the `JUnit4Example` and `JUnit5Example` classes.

Done in https://github.com/junit-team/junit5/pull/4281/commits/e56946fb762d0b310e18f88177c0244af4559dbb !

YongGoose:
@marcphilipp 

I'm planning to move on to the documentation work. 🚀
Would it be possible to check if there's anything else that should be added to the tests?

Also, if you could recommend which part of file should be updated or where a new document should be added, that would be really helpful. 

> Writing documentation is always a challenge! 😅

junit-platform-launcher/src/main/java/org/junit/platform/launcher/LauncherSession.java
marcphilipp, line 52: Please add Javadoc for this method.

junit-platform-launcher/src/testFixtures/java/org/junit/platform/launcher/core/NamespacedHierarchicalStoreProviders.java
```diff
@@ -0,0 +1,27 @@
...
+
+import static org.junit.platform.engine.support.store.NamespacedHierarchicalStore.CloseAction.closeAutoCloseables;
+
+import org.junit.platform.engine.support.store.Namespace;
+import org.junit.platform.engine.support.store.NamespacedHierarchicalStore;
+
+public class NamespacedHierarchicalStoreProviders {
+
+	public static NamespacedHierarchicalStore<Namespace> dummyNamespacedHierarchicalStore() {
+		return new NamespacedHierarchicalStore<>(new NamespacedHierarchicalStore<>(null), closeAutoCloseables());
```
marcphilipp, line 21: The parent store should also use `closeAutoCloseables()`.

```suggestion
		return new NamespacedHierarchicalStore<>(dummyNamespacedHierarchicalStoreWithNoParent(), closeAutoCloseables());
```

jupiter-tests/src/test/java/org/junit/jupiter/engine/descriptor/ExtensionContextTests.java
```diff
@@ -78,6 +79,8 @@ public class ExtensionContextTests {
 
 	private final JupiterConfiguration configuration = mock();
 	private final ExtensionRegistry extensionRegistry = mock();
+	private final LauncherStoreFacade launcherStoreFacade = new LauncherStoreFacade(
+		new NamespacedHierarchicalStore<>(new NamespacedHierarchicalStore<>(null)));
```
marcphilipp, line 83: This should use `NamespacedHierarchicalStoreProviders`, right?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/Namespace.java
```diff
@@ -0,0 +1,100 @@
...
+	 * which use the same sequence of {@code parts} for creating a namespace.
+	 *
+	 * <p>The order of the {@code parts} is significant.
+	 *
+	 * <p>Internally the {@code parts} are compared using {@link Object#equals(Object)}.
+	 */
+	public static Namespace create(Object... parts) {
+		Preconditions.notEmpty(parts, "parts array must not be null or empty");
+		Preconditions.containsNoNullElements(parts, "individual parts must not be null");
+		return create(Arrays.asList(parts));
```
marcphilipp, line 44: Calling `create(List)` would check the preconditions again.

```suggestion
		return new Namespace(Arrays.asList(parts));
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/Namespace.java
```diff
@@ -0,0 +1,100 @@
...
+	 * which use the same sequence of {@code objects} for creating a namespace.
+	 *
+	 * <p>The order of the {@code objects} is significant.
+	 *
+	 * <p>Internally the {@code objects} are compared using {@link Object#equals(Object)}.
+	 */
+	public static Namespace create(List<Object> objects) {
+		Preconditions.notEmpty(objects, "objects list must not be null or empty");
+		Preconditions.containsNoNullElements(objects, "individual objects must not be null");
+		return new Namespace(new ArrayList<>(objects));
```
marcphilipp, line 58:
```suggestion
		return new Namespace(objects);
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/Namespace.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	private final List<Object> parts;
+
+	private Namespace(List<Object> parts) {
+		this.parts = parts;
```
marcphilipp, line 64:
```suggestion
		this.parts = new ArrayList<>(parts);
```

jupiter-tests/src/test/java/org/junit/jupiter/engine/JupiterEngineTests.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.platform.engine.ConfigurationParameters;
+import org.junit.platform.engine.EngineExecutionListener;
+import org.junit.platform.engine.ExecutionRequest;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.launcher.core.NamespacedHierarchicalStoreProviders;
+
+/**
+ * @since  5.13
+ */
+public class JupiterEngineTests {
```
marcphilipp, line 32: Please call this class `JupiterTestEngineTests` and merge `JupiterTestEngineBasicTests` into here.

jupiter-tests/src/test/java/org/junit/jupiter/engine/JupiterEngineTests.java
```diff
@@ -0,0 +1,69 @@
...
+
+	private final TestDescriptor rootTestDescriptor = mock(JupiterEngineDescriptor.class);
```
marcphilipp, line 34:
```suggestion
	private final TestDescriptor rootTestDescriptor = mock();
```

jupiter-tests/src/test/java/org/junit/jupiter/engine/JupiterEngineTests.java
```diff
@@ -0,0 +1,69 @@
...
+		assertThat(context).isNotNull();
+	}
+
+	@Test
+	void createExecutionContextWithNoParentsRequestLevelStore() {
+		when(executionRequest.getRequestLevelStore()).thenReturn(
+			NamespacedHierarchicalStoreProviders.dummyNamespacedHierarchicalStoreWithNoParent());
+
+		assertThatThrownBy(() -> engine.createExecutionContext(executionRequest)).isInstanceOf(
+			JUnitException.class).hasMessageContaining("Request-level store must have a parent");
```
marcphilipp, line 66: Please use `//` at the end of each line to make Spotless format this properly.

platform-tests/src/test/java/org/junit/platform/engine/support/store/NamespaceTests.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.jupiter.api.Test;
+
+public class NamespaceTests {
+
+	@Test
+	void namespacesEqualForSamePartsSequence() {
+		Namespace ns1 = Namespace.create("part1", "part2");
+		Namespace ns2 = Namespace.create("part1", "part2");
+
+		assertEquals(ns1, ns2);
```
marcphilipp, line 25: Please use `assertEqualsAndHashCode` from `EqualsAndHashCodeAssertions` here.

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
```diff
@@ -390,6 +418,33 @@ static class JUnit5Example {
...
+		@Override
+		public void launcherSessionOpened(LauncherSession session) {
+			session.getStore().put(Namespace.GLOBAL, "testKey", "testValue");
+		}
+	}
+
+	static class JupiterExtensionExample implements BeforeAllCallback {
+		@Override
+		public void beforeAll(ExtensionContext context) {
+			var value = context.getRoot().getStore(ExtensionContext.Namespace.GLOBAL).get("testKey");
```
marcphilipp, line 443: There should be no need to call `getRoot()` here. Please also check that we can get the value from `context.getSessionLevelStore(ExtensionContext.Namespace.GLOBAL)` as well.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/DefaultLauncherSession.java
```diff
@@ -73,11 +83,17 @@ LauncherSessionListener getListener() {
 	public void close() {
 		if (launcher.delegate != ClosedLauncher.INSTANCE) {
 			launcher.delegate = ClosedLauncher.INSTANCE;
+			store.close();
 			listener.launcherSessionClosed(this);
```
marcphilipp, line 87: To allow the listener to read things from the `store`, we should first call the listener and then close the store.

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
```diff
@@ -333,6 +338,29 @@ public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult
 		}));
 	}
 
+	@Test
+	void extensionCanReadValueFromSessionStore() {
```
marcphilipp, line 342: I think we should also test that an extension can store a value in the session-level store that can be read by a `LauncherSessionListener` in `launcherSessionClosed()`.

platform-tests/src/test/java/org/junit/platform/launcher/core/StoreSharingTests.java
```diff
@@ -0,0 +1,54 @@
...
+				super.execute(request);
+			}
+		};
+
+		ExecutionRequest request = mock(ExecutionRequest.class);
+		when(request.getRequestLevelStore()).thenReturn(
+			NamespacedHierarchicalStoreProviders.dummyNamespacedHierarchicalStore());
+
+		engineWriter.execute(request);
+		engineReader.execute(request);
```
marcphilipp, line 52: I think we should go through `Launcher.execute` here rather than calling `TestEngine.execute` directly. That way, we would also test the `Launcher`'s handling of the store meets our expectations.

platform-tests/src/test/java/org/junit/platform/suite/engine/SuiteEngineTests.java
```diff
@@ -467,6 +476,25 @@ void passesOutputDirectoryProviderToEnginesInSuite() {
...
+
+		EngineExecutionListener listener = mock(EngineExecutionListener.class);
+		NamespacedHierarchicalStore<Namespace> requestLevelStore = NamespacedHierarchicalStoreProviders.dummyNamespacedHierarchicalStore();
+
+		ExecutionRequest request = ExecutionRequest.create(engineDescriptor, listener,
+			mock(ConfigurationParameters.class), mock(OutputDirectoryProvider.class), requestLevelStore);
+
+		new SuiteTestEngine().execute(request);
+
+		verify(mockDescriptor).execute(listener, requestLevelStore);
```
marcphilipp, line 495: I think we should make the assertion a bit stronger to avoid accidentally depending on any `equals()` implementation.

```suggestion
		verify(mockDescriptor).execute(same(listener), same(requestLevelStore));
```

marcphilipp (changes requested):
>  Would it be possible to check if there's anything else that should be added to the tests?

The main thing I'm missing is integration tests that verifies `AutoCloseable` objects put into the request-level store get closed after execution and those put into the session-level store get closed when the `LauncherSession` is closed.

> Also, if you could recommend which part of file should be updated or where a new document should be added, that would be really helpful.

We should update the [usage example](https://junit.org/junit5/docs/current/user-guide/#launcher-api-launcher-session-listeners-tool-example-usage) to store `HttpServer` (or a wrapper, if necessary) in the session-level store, retrieve it from a new `ParameterResolver` extension and inject it into the test rather than using system properties.

Moreover, we should add a new "Managing State Across Test Engines" subsection to [JUnit Platform Launcher API](https://junit.org/junit5/docs/current/user-guide/#launcher-api) where we provide an example of two test engines lazily creating and sharing a resource (maybe the same `HttpServer` or a wrapper thereof as in the other example?) regardless of which is executed first.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassExtensionContext.java
```diff
@@ -35,11 +35,26 @@ final class ClassExtensionContext extends AbstractExtensionContext<ClassBasedTes
 
 	private TestInstances testInstances;
 
+	/**
+	 * Create a new {@code ClassExtensionContext} with {@link Lifecycle#PER_METHOD}.
+	 *
+	 * @see #ClassExtensionContext(ExtensionContext, EngineExecutionListener, ClassBasedTestDescriptor, Lifecycle, JupiterConfiguration, ExtensionRegistry, LauncherStoreFacade, ThrowableCollector)
+	 */
+	ClassExtensionContext(ExtensionContext parent, EngineExecutionListener engineExecutionListener,
```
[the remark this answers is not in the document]
YongGoose: It seems to be unused.  
I'll remove it!

YongGoose:
> The main thing I'm missing is integration tests that verifies `AutoCloseable` objects put into the request-level store get closed after execution and those put into the session-level store get closed when the `LauncherSession` is closed.

Thank you so much for your detailed feedback on my work!  🙇🏻‍♂️🙇🏻‍♂️
Since you've reviewed it thoroughly, I'll make sure to refine my work carefully before requesting a review.

jupiter-tests/src/test/java/org/junit/jupiter/engine/descriptor/ExtensionContextTests.java
YongGoose: Yes, you're right!

YongGoose:
@marcphilipp 

For the session-level store, I implemented the test to call `store.close` when the `LauncherSession` is closed to release resources. For the request-level store, I used an Extension to detect when the execution has ended.  

Could you please provide feedback on my work?

platform-tests/src/test/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStoreTests.java
```diff
@@ -560,6 +570,32 @@ private void assertClosed() {
...
+		var config = LauncherConfig.builder().build();
+
+		try (LauncherSession session = LauncherFactory.openSession(config)) {
+			var launcher = session.getLauncher();
+			var request = request().selectors(selectClass(RequestResourceAutoCloseTestCase.class)).build();
+
+			launcher.execute(request);
+		}
+	}
+
```
marcphilipp, line 598: I think these two tests should be in `LauncherFactoryTests`, too.

marcphilipp (commented):
> For the **request-level store**, I used an Extension to detect when the execution has ended.
> But, I'm not sure if checking the request-level store through Extension is the right approach.

The extension is executed within the request so that's too early. I think the only way to test this is to make the `closed` flag of the resource `static`, set it to `false` before running the test, and checking that it's set to `true` after `Launcher.execute` returns.

platform-tests/src/test/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStoreTests.java
YongGoose: Done in https://github.com/junit-team/junit5/pull/4281/commits/377b3813ecd9d76848c19d073d40902853f3f353 !

marcphilipp (changes requested):
I reworked the User Guide sample so it's stateless (in the listener) and lazy. I hope you don't mind!

I think we're almost there! :+1:

YongGoose:
> I reworked the User Guide sample so it's stateless (in the listener) and lazy. I hope you don't mind!
> 
> 
> 
> I think we're almost there! :+1: 


Thanks for the update! 

I only put together a rough draft before bed and was planning to refine it this morning, but it feels like I woke up to an unexpected gift. 

Really appreciate it! 👍

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
```diff
@@ -497,4 +526,65 @@ public void beforeAll(ExtensionContext context) {
...
+
+	private static class AutoCloseCheckListener implements LauncherSessionListener {
+		@Override
+		public void launcherSessionClosed(LauncherSession session) {
+			session.getStore().close();
+			CloseTrackingResource sessionResource = session //
+					.getStore() //
+					.get(Namespace.GLOBAL, "sessionResource", CloseTrackingResource.class);
+
+			assertThat(sessionResource.isClosed()).isTrue();
```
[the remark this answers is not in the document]
YongGoose: When `assertThat(sessionResource.isClosed()).isFalse();` is called, `sessionResource` has not been defined yet.

Therefore, I believe it would be more appropriate to compare it after `sessionResource` has been defined, as shown in the code below. What do you think?

```
@Override
public void launcherSessionClosed(LauncherSession session) {
	CloseTrackingResource sessionResource = session //
			.getStore() //
			.get(Namespace.GLOBAL, "sessionResource", CloseTrackingResource.class);

	assertThat(sessionResource.isClosed()).isFalse();
	session.getStore().close();
	assertThat(sessionResource.isClosed()).isTrue();
}
```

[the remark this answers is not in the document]
marcphilipp: > When `assertThat(sessionResource.isClosed()).isFalse();` is called, `sessionResource` has not been defined yet.

Yes, sorry about that. :facepalm: 

I think we should instead not close it ourselves here but check that it's been closed in the test that uses this listener.

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
```diff
@@ -333,6 +338,80 @@ public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult
...
+		var config = LauncherConfig.builder() //
+				.addLauncherSessionListeners(new AutoCloseCheckListener()) //
+				.build();
+
+		try (LauncherSession session = LauncherFactory.openSession(config)) {
+			var launcher = session.getLauncher();
+			var request = request().selectors(selectClass(SessionResourceAutoCloseTestCase.class)).build();
+
+			launcher.execute(request);
+		}
```
marcphilipp, line 397:
```suggestion
			assertThat(CloseTrackingResource.closed).isFalse();
		}

		assertThat(CloseTrackingResource.closed).isTrue();
```

platform-tests/src/test/java/org/junit/platform/launcher/core/LauncherFactoryTests.java
```diff
@@ -390,6 +469,126 @@ static class JUnit5Example {
...
+		public void launcherSessionClosed(LauncherSession session) {
+			assertThat(sessionResource.isClosed()).isFalse();
+
+			CloseTrackingResource sessionResource = session //
+					.getStore() //
+					.get(Namespace.GLOBAL, "sessionResource", CloseTrackingResource.class);
+
+			session.getStore().close();
+
+			assertThat(sessionResource.isClosed()).isTrue();
```
marcphilipp, line 591:
```suggestion
			CloseTrackingResource sessionResource = session //
					.getStore() //
					.get(Namespace.GLOBAL, "sessionResource", CloseTrackingResource.class);

			assertThat(sessionResource.isClosed()).isFalse();
```

marcphilipp:
@YongGoose I'll discuss this PR with the team on Friday and will leave it open until then.

documentation/src/test/java/example/session/CloseableServerSocket.java
YongGoose, line 1: Since `CloseableHttpServer` is automatically stored by `GlobalSetupTeardownListener`, it cannot be verified as being created by the test engine. To address this, I created a `CloseableServerSocket` class to implement the verification logic.

YongGoose (commented):
Dcumentation work is always challenging. 🤯 I’ve added comments where I felt additional explanation or improvements might be helpful. 

Thank you, as always, for taking the time to provide feedback on my work! 🙇🏻

YongGoose:
@marcphilipp 
I reviewed the changes https://github.com/junit-team/junit5/pull/4281/commits/677648a7c85de358f898bdbc3dc309d2581b3358, and I think limiting the scope to define the range is a great approach.  👍🏻

In the previous method, we had to add new methods every time we wanted to extend functionality, which was a drawback. However, with these changes, the flexibility for expansion has significantly improved.

Would it be okay to ask if there were any further discussions about this PR during the team meeting?

marcphilipp (approved):

documentation/src/test/java/example/sharedresources/SharedResourceTest.java
```diff
@@ -0,0 +1,46 @@
...
+		LauncherConfig launcherConfig = LauncherConfig //
+				.builder() //
+				.addTestEngines(new FirstCustomEngine(), new SecondCustomEngine()) //
+				.build();
+		LauncherDiscoveryRequest launcherDiscoveryRequest = LauncherDiscoveryRequestBuilder.request().build();
+
+		try (LauncherSession session = LauncherFactory.openSession(launcherConfig)) {
+			session.getLauncher().execute(launcherDiscoveryRequest);
+
+			assertThat(FirstCustomEngine.socket).isSameAs(SecondCustomEngine.socket);
```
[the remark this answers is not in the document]
marcphilipp: I switched to `assertSame` to avoid the dependency to AssertJ in our docs.

documentation/src/docs/asciidoc/user-guide/advanced-topics/launcher-api.adoc
```diff
@@ -285,3 +301,45 @@ execute any tests but will notify registered `{TestExecutionListener}` instances
...
+the platform's store API to lazily initialize and share resources, ensuring they are
+created only once regardless of execution order. The following example demonstrates two
+custom test engines sharing a ServerSocket resource. `FirstCustomEngine` attempts to
+retrieve an existing ServerSocket from the global store or creates a new one if it
+doesn't exist:
+
+[source,java,indent=0]
+----
+include::{testDir}/example/FirstCustomEngine.java[tags=user_guide]
+----
```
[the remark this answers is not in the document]
marcphilipp: I think that's fine. :+1:

documentation/src/test/java/example/session/CloseableServerSocket.java
marcphilipp: Since `ServerSocket` extends `AutoCloseable` (contrary to `HttpServer`), I removed the wrapper class.

marcphilipp:
@YongGoose Thank you for all your hard work! :+1: :bow:

YongGoose:
> Thank you for all your hard work! 👍 🙇

This project was personally very challenging for me. In the beginning, there were moments when I considered giving up, but instead of wasting time on such thoughts, I decided to put in even more effort and give it my all.

As I kept pushing forward, two months flew by, and I was finally able to complete the task.
Thank you so much @marcphilipp. I truly believe that I could never have accomplished this without your help. 🙇🏻‍♂️

I will do my best in the next task as well, just as I did with this one. 💪🏻
I am always grateful to you.

marcphilipp:
> Would it be okay to ask if there were any further discussions about this PR during the team meeting?

Sure! We took a few steps back and discussed whether we need this at all. In the end, we decided we do since it "resource management" was the reason for introducing `LauncherSession{Listener}` in the first place and this mechanism provides a proper way of interacting with such resources in tests without having to use static variables or system properties.

PR #4289, opened 2025-01-30
Order test descriptors children in place

mpkorstanje:
## Overview
Currently, it is not possible to order children
without first removing all children and then
adding them back in order. E.g:

```java
var children = new ArrayList<>(testDescriptor.getChildren());
Collections.shuffle(children);
children.forEach(testDescriptor::removeChild);
children.forEach(testDescriptor::addChild);
```
By adding `orderChildren(UnaryOperator<List<TestDescriptor>> orderer)` it now becomes possible to write:

```java
testDescriptor.orderChildren(children -> {
	Collections.shuffle(children);
	return children;
});
```

---

### Definition of Done

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/AbstractOrderingVisitor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/AbstractOrderingVisitor.java
@@ -11,14 +11,12 @@
 package org.junit.jupiter.engine.discovery;
 
 import static java.util.stream.Collectors.toCollection;
+import static java.util.stream.Collectors.toList;
 
 import java.util.ArrayList;
-import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Function;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor;
@@ -60,9 +58,8 @@ abstract class AbstractOrderingVisitor<PARENT extends TestDescriptor, CHILD exte
 	protected void orderChildrenTestDescriptors(TestDescriptor parentTestDescriptor, Class<CHILD> matchingChildrenType,
 			Function<CHILD, WRAPPER> descriptorWrapperFactory, DescriptorWrapperOrderer descriptorWrapperOrderer) {
 
-		Set<? extends TestDescriptor> children = parentTestDescriptor.getChildren();
-
-		List<WRAPPER> matchingDescriptorWrappers = children.stream()//
+		List<WRAPPER> matchingDescriptorWrappers = parentTestDescriptor.getChildren()//
+				.stream()//
 				.filter(matchingChildrenType::isInstance)//
 				.map(matchingChildrenType::cast)//
 				.map(descriptorWrapperFactory)//
@@ -74,50 +71,48 @@ abstract class AbstractOrderingVisitor<PARENT extends TestDescriptor, CHILD exte
 		}
 
 		if (descriptorWrapperOrderer.canOrderWrappers()) {
-			List<TestDescriptor> nonMatchingTestDescriptors = children.stream()//
-					.filter(childTestDescriptor -> !matchingChildrenType.isInstance(childTestDescriptor))//
-					.collect(Collectors.toList());
+			parentTestDescriptor.orderChildren(children -> {
+				List<TestDescriptor> nonMatchingTestDescriptors = children.stream()//
+						.filter(childTestDescriptor -> !matchingChildrenType.isInstance(childTestDescriptor))//
+						.collect(toList());
 
-			// Make a local copy for later validation
-			Set<WRAPPER> originalWrappers = new LinkedHashSet<>(matchingDescriptorWrappers);
+				// Make a local copy for later validation
+				List<WRAPPER> originalWrappers = new ArrayList<>(matchingDescriptorWrappers);
 
-			descriptorWrapperOrderer.orderWrappers(matchingDescriptorWrappers);
+				descriptorWrapperOrderer.orderWrappers(matchingDescriptorWrappers);
 
-			int difference = matchingDescriptorWrappers.size() - originalWrappers.size();
-			if (difference > 0) {
-				descriptorWrapperOrderer.logDescriptorsAddedWarning(difference);
-			}
-			else if (difference < 0) {
-				descriptorWrapperOrderer.logDescriptorsRemovedWarning(difference);
-			}
+				int difference = matchingDescriptorWrappers.size() - originalWrappers.size();
+				if (difference > 0) {
+					descriptorWrapperOrderer.logDescriptorsAddedWarning(difference);
+				}
+				else if (difference < 0) {
+					descriptorWrapperOrderer.logDescriptorsRemovedWarning(difference);
+				}
 
-			Set<TestDescriptor> orderedTestDescriptors = matchingDescriptorWrappers.stream()//
-					.filter(originalWrappers::contains)//
-					.map(AbstractAnnotatedDescriptorWrapper::getTestDescriptor)//
-					.collect(toCollection(LinkedHashSet::new));
+				List<TestDescriptor> orderedTestDescriptors = matchingDescriptorWrappers.stream()//
+						.filter(originalWrappers::contains)//
+						.map(AbstractAnnotatedDescriptorWrapper::getTestDescriptor)//
+						.collect(toList());
 
-			// There is currently no way to removeAll or addAll children at once, so we
-			// first remove them all and then add them all back.
-			Stream.concat(orderedTestDescriptors.stream(), nonMatchingTestDescriptors.stream())//
-					.forEach(parentTestDescriptor::removeChild);
+				// If we are ordering children of type ClassBasedTestDescriptor, that means we
+				// are ordering top-level classes or @Nested test classes. Thus, the
+				// nonMatchingTestDescriptors list is either empty (for top-level classes) or
+				// contains only local test methods (for @Nested classes) which must be executed
+				// before tests in @Nested test classes. So we add the test methods before adding
+				// the @Nested test classes.
+				if (matchingChildrenType == ClassBasedTestDescriptor.class) {
+					return Stream.concat(nonMatchingTestDescriptors.stream(), orderedTestDescriptors.stream())//
+							.collect(toList());
+				}
+				// Otherwise, we add the ordered descriptors before the non-matching descriptors,
+				// which is the case when we are ordering test methods. In other words, local
+				// test methods always get added before @Nested test classes.
+				else {
+					return Stream.concat(orderedTestDescriptors.stream(), nonMatchingTestDescriptors.stream())//
+							.collect(toList());
+				}
+			});
 
-			// If we are ordering children of type ClassBasedTestDescriptor, that means we
-			// are ordering top-level classes or @Nested test classes. Thus, the
-			// nonMatchingTestDescriptors list is either empty (for top-level classes) or
-			// contains only local test methods (for @Nested classes) which must be executed
-			// before tests in @Nested test classes. So we add the test methods before adding
-			// the @Nested test classes.
-			if (matchingChildrenType == ClassBasedTestDescriptor.class) {
-				Stream.concat(nonMatchingTestDescriptors.stream(), orderedTestDescriptors.stream())//
-						.forEach(parentTestDescriptor::addChild);
-			}
-			// Otherwise, we add the ordered descriptors before the non-matching descriptors,
-			// which is the case when we are ordering test methods. In other words, local
-			// test methods always get added before @Nested test classes.
-			else {
-				Stream.concat(orderedTestDescriptors.stream(), nonMatchingTestDescriptors.stream())//
-						.forEach(parentTestDescriptor::addChild);
-			}
 		}
 
 		// Recurse through the children in order to support ordering for @Nested test classes.
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
@@ -10,12 +10,16 @@
 
 package org.junit.platform.engine;
 
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.STABLE;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.UnaryOperator;
 
 import org.apiguardian.api.API;
 import org.junit.platform.commons.util.Preconditions;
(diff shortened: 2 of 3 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -172,6 +176,30 @@ default Set<? extends TestDescriptor> getDescendants() {
...
+	 */
+	@API(since = "5.12", status = EXPERIMENTAL)
+	default void orderChildren(UnaryOperator<List<TestDescriptor>> orderer) {
+		Preconditions.notNull(orderer, "orderer must not be null");
+		Set<? extends TestDescriptor> children = getChildren();
+		List<? extends TestDescriptor> orderedChildren = orderer.apply(new ArrayList<>(children));
+		boolean unmodified = children.size() == orderedChildren.size() && children.containsAll(orderedChildren);
+		Preconditions.condition(unmodified, "orderer may not add or remove test descriptors");
+		orderedChildren.forEach(this::removeChild);
+		orderedChildren.forEach(this::addChild);
```
mpkorstanje, line 200: This could be optimized by pushing down the implementation, but then the interface still needs a sensible default.

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -172,6 +176,30 @@ default Set<? extends TestDescriptor> getDescendants() {
...
+	 * that an {@code orderer} may not remove children from this descriptor.
+	 *
+	 * @param orderer a unary operator to order the children from this test descriptor.
+	 */
+	@API(since = "5.12", status = EXPERIMENTAL)
+	default void orderChildren(UnaryOperator<List<TestDescriptor>> orderer) {
+		Preconditions.notNull(orderer, "orderer must not be null");
+		Set<? extends TestDescriptor> children = getChildren();
+		List<? extends TestDescriptor> orderedChildren = orderer.apply(new ArrayList<>(children));
+		boolean unmodified = children.size() == orderedChildren.size() && children.containsAll(orderedChildren);
```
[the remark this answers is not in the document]
mpkorstanje: You are correct.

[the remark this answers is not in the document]
mpkorstanje: > If so, what would happen if it compared [a, a, b] and [a, b, b]?

This should not be possible. `children` is a set and may not contain duplicates.

> How about children = [a, b, c] and orderedChildren = [a, b, b]?

That will result in an erroneous check.

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
marcphilipp: I think having this unoptimized version in the interface is fine but we should have an optimized version in `AbstractTestDescriptor`

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -172,6 +176,30 @@ default Set<? extends TestDescriptor> getDescendants() {
...
+	 *
+	 * @param orderer a unary operator to order the children from this test descriptor.
+	 */
+	@API(since = "5.12", status = EXPERIMENTAL)
+	default void orderChildren(UnaryOperator<List<TestDescriptor>> orderer) {
+		Preconditions.notNull(orderer, "orderer must not be null");
+		Set<? extends TestDescriptor> children = getChildren();
+		Set<? extends TestDescriptor> orderedChildren = new LinkedHashSet<>(orderer.apply(new ArrayList<>(children)));
+		boolean unmodified = children.equals(orderedChildren);
+		Preconditions.condition(unmodified, "orderer may not add or remove test descriptors");
```
marcphilipp, line 198: I think we're lacking a test where a misbehaving orderer that removes a descriptor still gets its ordering for the remaining descriptors applied. The check introduced by this line will cause an exception to be thrown in this case which gets caught and logged in `AbstractOrderingVisitor.doWithMatchingDescriptor`.

Could you please add such a test (or extend an existing one) and make sure to include the removed descriptors in the `UnaryOperator` implementation to restore the original behavior?

marcphilipp (changes requested):

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
mpkorstanje: I think the logical conclusion of this request is the remove this check and use the order provided by the orderer as a suggestion for the actual order.

marcphilipp: I've pushed a commit that reinstates the condition because I think it's better to be strict here. @mpkorstanje Could you please take a look?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/AbstractTestDescriptor.java
```diff
@@ -147,6 +150,22 @@ public void removeFromHierarchy() {
...
+		Preconditions.notNull(suggestedOrder, "orderer may not return null");
+
+		Set<? extends TestDescriptor> orderedChildren = new LinkedHashSet<>(suggestedOrder);
+		boolean unmodified = this.children.equals(orderedChildren);
+		Preconditions.condition(unmodified, "orderer may not add or remove test descriptors");
+		Preconditions.condition(this.children.size() == suggestedOrder.size(),
+			"orderer may not add duplicate test descriptors");
+
+		this.children.clear();
+		this.children.addAll(orderedChildren);
```
marcphilipp, line 166: This is now the optimized implementation

marcphilipp (approved):

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -172,6 +176,40 @@ default Set<? extends TestDescriptor> getDescendants() {
 	 */
 	void removeFromHierarchy();
 
+	/**
+	 * Order all children from this descriptor.
+	 *
+	 * <p>The {@code orderer} is provided a modifiable list of child test descriptors in
+	 * this test descriptor. Never {@code null}. The {@code orderer} must return
+	 * a list of descriptors in any order, never {@code null}.
```
marcphilipp, line 184: I think the API would be clearer if the `ordered` received an unmodifiable list and returned the new list. Thoughts?

marcphilipp (changes requested):

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/AbstractOrderingVisitor.java
```diff
@@ -167,7 +149,43 @@ private boolean canOrderWrappers() {
 		}
 
 		private void orderWrappers(List<WRAPPER> wrappers) {
+			// Make a local copy for later validation
+			List<WRAPPER> originalWrappers = new ArrayList<>(wrappers);
+
 			this.orderingAction.accept(wrappers);
+
+			warnAboutAndRemoveAddedWrappers(originalWrappers, wrappers);
+			warnAboutAndReAddRemovedWrappers(originalWrappers, wrappers);
```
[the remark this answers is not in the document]
mpkorstanje: I think a solution could be to (written without a IDE):

```
private void orderWrappers(Set<WRAPPER> originalWrappers) {

  List<WRAPPER> wrappersToOrder = new ArrayList<>(wrappers);
  this.orderingAction.accept(wrappersToOrder);
  Set<WRAPPER> orderedWrappers = new HashSet<>(wrappersToOrder);
  orderedWrappers.stream()
     .filter(originalWrappers::contains)
     .forEach(wrapper -> {
          originalWrappers.remove(wrapper);
          originalWrappers.add(wrapper);
     }
   
     Integer difference = originalWrappers.size() - orderedWrappers.size();
   
     // TODO: Warn about difference
     
}
```

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
mpkorstanje: It has pro's and cons.

To use `Collections.shuffle(children)` or `children.sort(...)` every orderer would have to create a new list first. Which I imagine would be quite annoying when implementing an orderer.

But the `TestDescriptor` could optimize by not doing anything if the orderer returns the same list (by reference).

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
mpkorstanje: Ah. I see what you meant.

Left a few comments.

platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/AbstractTestDescriptorTests.java
```diff
@@ -49,15 +49,22 @@ void initTree() {
 		group11 = new GroupDescriptor(UniqueId.root("group", "group1-1"));
 		group1.addChild(group11);
 
-		group1.addChild(new LeafDescriptor(UniqueId.root("leaf", "leaf1-1")));
-		group1.addChild(new LeafDescriptor(UniqueId.root("leaf", "leaf1-2")));
+		var leaf11 = new LeafDescriptor(UniqueId.root("leaf", "leaf1-1"));
+		group1.addChild(leaf11);
+		var leaf12 = new LeafDescriptor(UniqueId.root("leaf", "leaf1-2"));
+		group1.addChild(leaf12);
```
mpkorstanje, line 55: Note to self: This didn't have to be changed.

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
```diff
@@ -172,6 +176,40 @@ default Set<? extends TestDescriptor> getDescendants() {
...
+		Preconditions.notNull(orderer, "orderer must not be null");
+		Set<? extends TestDescriptor> originalChildren = getChildren();
+		List<TestDescriptor> suggestedOrder = orderer.apply(new ArrayList<>(originalChildren));
+		Preconditions.notNull(suggestedOrder, "orderer may not return null");
+
+		Set<? extends TestDescriptor> orderedChildren = new LinkedHashSet<>(suggestedOrder);
+		boolean unmodified = originalChildren.equals(orderedChildren);
+		Preconditions.condition(unmodified, "orderer may not add or remove test descriptors");
+		Preconditions.condition(originalChildren.size() == suggestedOrder.size(),
+			"orderer may not add duplicate test descriptors");
```
mpkorstanje, line 202: I feel we could express both these requirements more clearly and in a single assertion. `orderer may only change the ordering of test descriptors`.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/AbstractOrderingVisitor.java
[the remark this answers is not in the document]
marcphilipp: Good catch, I think that should work! Looking forward to reviewing your change. :slightly_smiling_face:

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/AbstractOrderingVisitor.java
```diff
@@ -166,8 +149,33 @@ private boolean canOrderWrappers() {
...
+		}
+
+		@SuppressWarnings("unchecked")
+		private void orderWrappersWithOrderFrom(Set<WRAPPER> wrappers, Set<Object> orderedWrappers) {
+			// Avoid ClassCastException if a misbehaving ordering action added a non-WRAPPER
+			for (Object wrapper : orderedWrappers) {
+				//noinspection SuspiciousMethodCalls
+				if (wrappers.remove(wrapper)) {
+					// guaranteed by wrappers.contains
+					wrappers.add((WRAPPER) wrapper);
```
mpkorstanje, line 176: Not happy with this yet. A misbehaving ordering action could add test descriptors with a broken equals/hashcode implementation.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/AbstractOrderingVisitor.java
```diff
@@ -167,7 +150,32 @@ private boolean canOrderWrappers() {
...
+			int difference = orderedWrappers.size() - wrappers.size();
+			int distinctDifference = distinctWrappersToIndex.size() - wrappers.size();
+			if (difference > 0) { // difference >= distinctDifference
+				logDescriptorsAddedWarning(difference);
+			}
+			if (distinctDifference < 0) { // distinctDifference <= difference
+				logDescriptorsRemovedWarning(distinctDifference);
+			}
+
+			wrappers.sort(comparing(wrapper -> distinctWrappersToIndex.getOrDefault(wrapper, -1)));
```
mpkorstanje, line 166: @marcphilipp my solution from 842d554ca1d8c20c3b3f5bfba7c81945a9429efb runs in linear time, this is n log n. But has the advantage of not copying elements touched by the ordering action into `wrappers`.

I'll leave it to you which you'd prefer.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/AbstractTestDescriptor.java
mpkorstanje, line 166: Here too there is the possibility here that `orderedChildren` are impersonating the actual `children`.

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
marcphilipp: I agree that this API is "easiest to work with" and have pushed a commit that documents how to use it in the method's Javadoc.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/AbstractOrderingVisitor.java
marcphilipp: This comment is outdated, right?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/AbstractTestDescriptor.java
marcphilipp: I think we can live with that. It's clearly documented that an orderer must not do that. If anyone does, all bets are off. Besides, one can already call `removeChild` and `addChild` to achieve impersonation. :slightly_smiling_face:

junit-platform-engine/src/main/java/org/junit/platform/engine/TestDescriptor.java
marcphilipp: I think your original error message `orderer may not add or remove test descriptors` is clearer and have merged these two lines in 56dfbfb5ed79f59dad86b25bb0a3090ad77d6035.

marcphilipp (approved):

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/AbstractOrderingVisitor.java
mpkorstanje: Yes!

mpkorstanje (commented):
LGTM!

marcphilipp:
@mpkorstanje Thank you! :bow:

Issue #4246, opened 2025-01-16
Change Java baseline to 17 for 6.0


marcphilipp:
## Motivation

Java 8 is the current baseline of JUnit 5. It was released over 10 years ago and recent versions of Java have introduced many enhancements that make moving to a more recent JDK as the baseline for JUnit alluring. Moreover, there is movement in the ecosystem regarding this issue. For example, Spring has [changed their baseline](https://spring.io/blog/2022/03/28/an-update-on-java-17-adoption) to JDK 17+ with 6.x and free support for 5.x will end in 2024. Similarly, AssertJ will [move to Java 17](https://github.com/assertj/assertj/issues/3442). Therefore, we should consider making a similar move. Doing so would mean bumping the major version number and would allow us to clean up and simplify a number of things. This issue serves as a starting point for collecting those topics and tasks. For users stuck on Java 8 we will extend support for the last 5.x release for at least a year and backport security and important bug fixes to 5.x.

## Todos

- [x] https://github.com/junit-team/junit5/pull/4249
- [x] Remove deprecated code
- [x] Clean up `ConsoleLauncher` CLI (remove single-dash long options and mode without subcommand)
- [x] Look for issues labeled with `type: breaking change` and decide whether they can be addressed in 6.0
- [x] Remove some if not all multi-release JAR code
- [x] https://github.com/junit-team/junit5/pull/4575
- [x] Support `AutoCloseable` instead of `CloseableResource` in `Store`?
- [x] Remove `@TempDir` per context mode?
- [x] ~Change default behavior for `TestInstancePostProcessor` and similar extensions wrt. method-level `ExtensionContext`?~ Too early for that
- [x] https://github.com/junit-team/junit5/issues/4576
- [x] https://github.com/junit-team/junit5/issues/4577
- [x] Bump Kotlin language version to latest stable
- [x] #4579
- [x] Remove `ReflectionUtils.isRecord*` methods and replace with calls to `Class.isRecord()`
- [x] Use `List.of` etc. instead of `Collections.emptyList` etc.
- [x] https://github.com/junit-team/junit5/issues/4578
- [x] Add nullability annotations to public API (use jspecify)
- [x] Replace reflection in `JRE.determineCurrentVersion()`
- [x] Resolve `TODO [#4246]` comments within the code base

jbduncan:
[...]

Issue #2452, opened 2020-10-18
@ParameterizedTest display custom names for arguments


SarathBabu:
When using `@ParamerizedTest` annotation, will it be possible to display a string other than what is supplied by toString()?
When we are using some big objects the toString value is very large and displays lots of informations which are not relevant

It would be helpful if we can give provision to
* display some parameter on the argument like `{0.shortName}` or `{0.route.origin}` etc
* Provide another class name which will generate the string to display for each argument

## Deliverables
Provision to customise the name of the arguments in `@ParamerizedTest`

sbrannen:
Closing this issue since it is effectively a duplicate of  #1154.

I'll link to this issue from  #1154, since this issue targets parameterized tests explicitly; however, a general purpose solution should be implemented in conjunction with #1154 that is usable with display names for parameterized tests as well.

Issue #1154, opened 2017-11-11
Add support for using argument fields in parametrized test names


DcortezMeleth:
## Overview

- ( x) **Feature request.** 

I have found that it would be useful if you could use argument fields as a part of parametrized test name. Right now, when using just `{0}` as described in docs, it results in injecting into a test name string in form of `ClassName(fieldName1=fieldValue1, fieldName2=fieldValue2, ...)`. But that doesn't seen readable at all especially then you have a class with more than just few fields. Now I would like to be able to use some semantics as `{0.name}` or `{0}.name` to use just name field of class used as test method argument, so I can write test name as follows:

``` java

@BeerTest
class BeerConversionServiceTest {

    @Autowired
    private BeerConversionService beerConversionService;

    private static Stream<Arguments> createBeers() {
        return Stream.of(
                Arguments.of(new Beer(...), 25L),
                Arguments.of(new Beer(...), 50L)
        );
    }

    @DisplayName("Get alcohol content of a beer in mg")
    @ParameterizedTest(name = "Alcohol content of {0.name} should be equal to {1}")
    @MethodSource("createBeers")
    void getAlcInMg(Beer beer, long mg) {
        Assertions.assertEquals(beerConversionService.getAlcInMg(beer), mg);
    }

}
```

This would give users possibility to keep test names simple, yet still descriptive, as they could reference objects passed as arguments and use only parts of them which identifies test cases.

I didn't look into the source code so I cannot tell which approach would be easier to implement but I would guess that `{0.name}` seems more reasonable.

marcphilipp:
I agree that this would be useful. Implementation-wise it might be a rabbit hole, though. We should probably check for getters first. If there are none, we could try to find a field traversing the class hierarchy upward.

@junit-team/junit-lambda Thoughts?

VikasSherawat:
hi @marcphilipp may i please work on this?

marcphilipp:
Sure!

sbrannen:
> I agree that this would be useful. Implementation-wise it might be a rabbit hole, though. We should probably check for getters first. If there are none, we could try to find a field traversing the class hierarchy upward.
> 
> @junit-team/junit-lambda Thoughts?

1. accessor method: `name()`
2. getter method: `getName()`
3. field named `name` (local and then traversing the hierarchy)

DcortezMeleth:
>     1. accessor method: name()
>     2. getter method: getName()
>     3. field named name (local and then traversing the hierarchy)

[...]

PR #4293, opened 2025-01-31
Spike for expression language evaluation in display names

jean-andre-gauthier:
## Overview

- Introduces a new `@ExpressionLanguage` annotation, that takes a `Class<ExpressionLanguageAdapter>` as parameter
- Introduces a new `ExpressionLanguageAdapter` interface, that can be implemented to bind in a third-party expression language
- Adds several tests in `ExpressionLanguageMustacheTests`, that demonstrate how `@ExpressionLanguage / ExpressionLanguageAdapter` can be used

I'm marking this as a draft PR, because it's not entirely clear yet how the API should look like. I'll address the items in the Definition of Done once we have more clarity on that.

---

---

### Definition of Done


junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
```diff
@@ -289,4 +303,34 @@ Set<String> placeholders() {
...
+		private final ExpressionLanguageAdapter expressionLanguageAdapter;
+
+		public ExpressionLanguageNonPlaceholderFormatter(ExpressionLanguageAdapter expressionLanguageAdapter) {
+			this.expressionLanguageAdapter = expressionLanguageAdapter;
+
+		}
+
+		@Override
+		public void append(ArgumentsContext argumentsContext, StringBuffer result) {
+			expressionLanguageAdapter.format(argumentsContext.arguments.get()[0], result);
```
jean-andre-gauthier, line 333: This is not correct, but I'm unsure how we should name multiple arguments, so that they can be accessed by the EL?

sbrannen (changes requested):
I've only taken a quick glance, but for starters...

This new feature must be applicable to JUnit Jupiter in general, not just for parameterized tests.

So the APIs need to reside in the `junit-jupiter-api` module.

Please revise your work to take that into account.

See also: https://github.com/junit-team/junit5/issues/1154#issuecomment-345445625

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ExpressionLanguageAdapter.java
```diff
@@ -0,0 +1,9 @@
+
+package org.junit.jupiter.params;
+
+public interface ExpressionLanguageAdapter {
+
+	void compile(String template);
+
+	void format(Object scope, StringBuffer result);
```
marcphilipp, line 8: I think we should split this into two to allow (not necessarily in this PR) registering a global default expression language that would support instantiating the `ExpressionLanguageAdapter` at most once and also provide a hook to clean up resources by extending `AutoCloseable`.

```suggestion
	Expression compile(String template);
```

```java
interface Expression {
	void evaluate(Map<String, Object> context, Appendable result);
}
```

marcphilipp:
> This new feature must be applicable to JUnit Jupiter in general, not just for parameterized tests.

@sbrannen Besides parameterized tests, what other use cases with "dynamic display names" did you have in mind? `@RepeatedTest(name = ...)`? Dynamic tests?

sbrannen:
> @sbrannen Besides parameterized tests, what other use cases with "dynamic display names" did you have in mind? `@RepeatedTest(name = ...)`? Dynamic tests?

There were discussions in #1154, #2452, and other places over the years where people expressed an interest in general-purpose expression language support.

I think at the very least that this should be usable with any `@TestTemplate`. Though, we may find this can be used in other scenarios as well.

And I fear that tying this to `junit-jupiter-param` would unnecessarily limit our options in this area.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedTestNameFormatter.java
sbrannen: Regarding the `Map<String, Object>`, that's also what I suggested in https://github.com/junit-team/junit5/issues/1154#issuecomment-345445625.

> In summary, I believe it would be better to design a new extension API that receives the `String displayName` and a `Map<String,Object> expressionContextMap` and then allow people to pick between Mustache, SpEL, etc.

sbrannen: > That would require compiling with `javac`'s `-parameters` option. We already check for `parameter.isNamePresent()` in `ParameterizedTestMethodContext.getParameterName(int)`, though, so we could populate the `Map` based on that and fail (?) if no parameter names are present?

I don't think I would throw an exception. Rather, I would document the behavior, pointing out that source code parameter names are available when compiled with `-parameters` and that otherwise the arguments are available via `arg0`, `arg1`, etc. (i.e., the default parameter names returned by `Parameter.getName()`).

jean-andre-gauthier:
Hi both, thanks very much for your reviews! I'll be off on holidays for the next two weeks, but will work on implementing https://github.com/junit-team/junit5/issues/1154#issuecomment-2643200828  after that (once it has been discussed with the team)

kaipe:
> > @sbrannen Besides parameterized tests, what other use cases with "dynamic display names" did you have in mind? `@RepeatedTest(name = ...)`? Dynamic tests?
> 
> There were discussions in #1154, #2452, and other places over the years where people expressed an interest in general-purpose expression language support.
> 
> I think at the very least that this should be usable with any `@TestTemplate`. Though, we may find this can be used in other scenarios as well.
> 
> And I fear that tying this to `junit-jupiter-param` would unnecessarily limit our options in this area.

Here I like to remind about my opinion on the subject  https://github.com/junit-team/junit5/issues/4246#issuecomment-2601057637 - as a way to not "limit our options in this area".

I.e. not tying it to `junit-jupiter-param` nor `junit-jupiter-engine` but offering it from `junit-platform-engine` and allow display-name manipulation in midair during test-execution.

It would allow JUnit Jupiter to expose an API for display-name manipulation as part of `ExtensionContext`, therewith allowing any kind of Jupiter extension to manipulate display-name with some expression language or any other mean available. Perhaps with a popup dialog, in which a tester could document how an exploratory test session caused a backend error?

There is a [kind of implementation](https://javadoc.io/static/org.lazyparams/lazyparams/1.0.1/org/lazyparams/LazyParamsCoreUtil.html) for it, because the functionality is required by LazyParams.

Issue #3577, opened 2023-11-29
Change return type to `Stream<? extends TestTemplateInvocationContext>`


sormuras:
https://github.com/junit-team/junit5/blob/f5b78f68c6a90dfa63611c7a10027581c5d913ba/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/TestTemplateInvocationContextProvider.java#L87

Michael wrote:
> I mean I tried to do `Stream.generate(() -> new MyInvocationContext()).limit(/*number*/)`
> This does not work.
> but `Stream.generate(() -> new MyInvocationContext()).limit(/*number*/).map(e -> e)` works...

Michael1993:
I think the return value should be `Stream<? extends TestTemplateInvocationContext>`? I might be missing something.

sormuras:
Question is, can we change the return type without breaking existing code?

marcphilipp:
According to https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_interfaces_-_API_methods that would break binary compatibility. We should have declared the method like that initially, though.

kaipe:
On a related topic ...
Wouldn't it be better to have return type Iterable?
I figure it's a simplier abstraction that is easy to implement and still provides all capabilities that are requied by JUnit.

marcphilipp:
I don't think it makes a big difference. `Stream` nowadays has many factory methods that allow `Iterator`-like implementations. Moreover, we ensure `Stream.close` is also called to support external resources such as files.

PR #4315, opened 2025-02-12
Introduce extension API for container templates

### LINKED ISSUE #871: Create Container Template Extension Point

marcphilipp:
## Overview

Analogous to `@TestTemplate` on the method level, this PR introduces a class-level `@ContainerTemplate` annotation with an accompanying `ContainerTemplateInvocationContextProvider` extension API.

Resolves #871.

---

---

### Definition of Done


Sola-ris:
I have two questions about the container templates, hope it's Ok to ask them here

* Are there plans to make the container template invocations interceptable via an `InvocationInterceptor`
  like test method templates?
* #3577 talks about declaring the return type of `TestTemplateInvocationContextProvider#provideTestTemplateInvocationContexts`
  as `Stream<? extends TestTemplateInvocationContext>` instead of `Stream<TestTemplateInvocationContext>`.
  Was `ContainerTemplateInvocationContextProvider#provideContainerTemplateInvocationContexts`'s return type declared without the upper bound?

marcphilipp:
> - Are there plans to make the container template invocations interceptable via an `InvocationInterceptor`
>       like test method templates?

We don't currently have plans to support intercepting the entire invocation of a container template which would encompass all of its test methods, nested test classes etc. Instead, we will add new lifecycle methods and extension points, specifically for container templates in #4320. 

> - #3577 talks about declaring the return type of `TestTemplateInvocationContextProvider#provideTestTemplateInvocationContexts`
>       as `Stream<? extends TestTemplateInvocationContext>` instead of `Stream<TestTemplateInvocationContext>`.
>       Was `ContainerTemplateInvocationContextProvider#provideContainerTemplateInvocationContexts`'s return type declared without the upper bound to be in sync with `TestTemplateInvocationContextProvider`?

Thanks for reminding me of that issue! I've changed `ContainerTemplateInvocationContextProvider` accordingly in bffb6c79107a8b77d520c3ab7447073661618bc0. `TestTemplateInvocationContextProvider` will have to wait until 6.0 (which is planned for later this year) because changing it is binary incompatible with previous releases.

Sola-ris:
> We don't currently have plans to support intercepting the entire invocation of a container template which would encompass all of its test methods, nested test classes etc. Instead, we will add new lifecycle methods and extension points, specifically for container templates in #4320.

Those should also cover the use case I had in mind, thank you.

> Thanks for reminding me of that issue! I've changed `ContainerTemplateInvocationContextProvider` accordingly in [bffb6c7](https://github.com/junit-team/junit5/commit/bffb6c79107a8b77d520c3ab7447073661618bc0). `TestTemplateInvocationContextProvider` will have to wait until 6.0 (which is planned for later this year) because changing it is binary incompatible with previous releases.

You're welcome :D

marcphilipp:
As discussed in the team, I'll go ahead and merge this to make it easier to work on #878 next. Feedback is still appreciated, though, even post-merge!

bruderj15:
This is awesome! I was awaiting this feature for a long time!
Will I have to wait a little bit longer until 5.13.0 or may you release it earlier?
Either way, great work!

marcphilipp:
> Will I have to wait a little bit longer until 5.13.0 or may you release it earlier?

You will have to wait until 5.13. However, if you're willing to use a milestone version, we will probably release an 5.13.0-M1 early on to make it easier to try out this new feature and provide feedback.

Issue #4321, opened 2025-02-15
Improve documentation on Parallel Test Execution in JUnit Vintage engine


YongGoose:
The current JUnit user guide provides an explanation of test configuration for parallel execution in the JUnit Vintage engine.  
However, I believe it should be **clearer and more detailed**.  

In JUnit Vintage, users can configure parallel execution at both the `class` and `method` levels independently.  
This can lead to various execution scenarios, and it is important to clearly inform users about these possibilities.

> Since I implemented this myself, feel free to reach out to me if you have any questions while working on this issue.  

## Deliverables

- [ ] ~~Add table of relevant configuration parameters (This [PR](https://github.com/junit-team/junit5/pull/2946
) should be helpful)~~
- [ ] Provide clear examples of parallel execution scenarios for both class and method levels

## Related Documents

- https://github.com/junit-team/junit5/issues/2669#issuecomment-2660929400
- https://github.com/junit-team/junit5/pull/4135
- https://github.com/junit-team/junit5/pull/4242
- https://junit.org/junit5/docs/snapshot/user-guide/index.html#migrating-from-junit4-parallel-execution

etrandafir93:
hello @marcphilipp - can i work on this issue?

marcphilipp:
@etrandafir93 Sure! Before you start, let's clarify a few things, though.

> Add table of relevant configuration parameters (This [PR](https://github.com/junit-team/junit5/pull/2946) should be helpful)

I think the [current description list](https://junit.org/junit5/docs/5.12.0-RC2/user-guide/#migrating-from-junit4-parallel-execution) is more readable, especially on small screens, than the table we have for Jupiter's parallel configuration.

@YongGoose Do you think anything is missing from it or what did you have in mind?

YongGoose:
> I think the [current description list](https://junit.org/junit5/docs/5.12.0-RC2/user-guide/#migrating-from-junit4-parallel-execution) is more readable, especially on small screens, than the table we have for Jupiter's parallel configuration.

I didn’t consider smaller screens.
After checking on my phone, I agree that Vintage’s parallel configuration has better readability than Jupiter’s parallel configuration.

> Add table of relevant configuration parameters (This https://github.com/junit-team/junit5/pull/2946 should be helpful)

Now, I don’t think this task is necessary.

marcphilipp:
@etrandafir93 If you're still interested, feel free to go ahead and start working on the remaining "deliverable" in the issue description. 🙂

[...]

PR #4336, opened 2025-02-23
improve junit vintage docs about parallel execution

etrandafir93:
Issue: #4321

## Overview

Improving user guide section about parallel test execution at class/methods level with JUnit Vintage.

---

---

### Definition of Done

- [ ] (n/a) There are no TODOs left in the code
- [ ] (n/a) Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.commons/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc
- [ ] (n/a) [Coding conventions](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed
- [ ] (n/a) Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling
- [ ] (n/a) Public API has [Javadoc](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [ ] (n/a) Change is documented in the [User Guide](https://junit.org/junit5/docs/snapshot/user-guide/) and [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
@@ -45,8 +45,8 @@ concurrent test execution. It can be enabled and configured using the following
 <<running-tests-config-params, configuration parameters>>:
 
 `junit.vintage.execution.parallel.enabled=true|false`::
-  Enable/disable parallel execution (defaults to `false`). Requires opt-in for `classes`
-  or `methods` to be executed in parallel using the configuration parameters below.
+  Enable/disable parallel execution (defaults to `false`).
+Requires opt-in for `classes` or `methods` to be executed in parallel using the configuration parameters below.
 
 `junit.vintage.execution.parallel.classes=true|false`::
   Enable/disable parallel execution of test classes (defaults to `false`).
@@ -55,21 +55,76 @@ concurrent test execution. It can be enabled and configured using the following
   Enable/disable parallel execution of test methods (defaults to `false`).
 
 `junit.vintage.execution.parallel.pool-size=<number>`::
-  Specifies the size of the thread pool to be used for parallel execution. By default, the
-  number of available processors is used.
+  Specifies the size of the thread pool to be used for parallel execution.
+By default, the  number of available processors is used.
 
-Example configuration in `junit-platform.properties`:
+
+Let's assume we have two test classes `FooTest` and `BarTest` with each class containing three unit tests.
+Now, let's enable the parallel execution of test classes:
+
+[source,properties]
+----
+junit.vintage.execution.parallel.enabled=true
+junit.vintage.execution.parallel.classes=true
+----
+
+With this setup, the `VintageTestEngine` will use two different threads,
+one for each test class:
+
+[source,plaintext]
+----
+ForkJoinPool-1-worker-1 - BarTest::test1
+ForkJoinPool-1-worker-2 - FooTest::test1
+ForkJoinPool-1-worker-1 - BarTest::test2
+ForkJoinPool-1-worker-2 - FooTest::test2
+ForkJoinPool-1-worker-1 - BarTest::test3
+ForkJoinPool-1-worker-2 - FooTest::test3
+----
+
+On the other hand, we can enable the parallel test execution at a `method` level,
+rather than the `class` level:
+
+[source,properties]
+----
+junit.vintage.execution.parallel.enabled=true
+junit.vintage.execution.parallel.methods=true
+----
+
+Therefore, the test methods from within a class will be executed in parallel,
+while different test classes will be executed sequentially:
+
+[source,plaintext]
+----
+ForkJoinPool-1-worker-1 - BarTest::test1
+ForkJoinPool-1-worker-2 - BarTest::test2
+ForkJoinPool-1-worker-3 - BarTest::test3
+
+ForkJoinPool-1-worker-3 - FooTest::test1
+ForkJoinPool-1-worker-2 - FooTest::test2
+ForkJoinPool-1-worker-1 - FooTest::test3
+----
+
+Finally, we can also enable the parallelization at both `class` and `method` level:
 
 [source,properties]
 ----
 junit.vintage.execution.parallel.enabled=true
 junit.vintage.execution.parallel.classes=true
 junit.vintage.execution.parallel.methods=true
-junit.vintage.execution.parallel.pool-size=4
 ----
 
-With these properties set, the `VintageTestEngine` will execute tests in parallel,
-potentially significantly reducing the overall test suite execution time.
+With these properties set, the `VintageTestEngine` will execute all the tests classes
+and methods in parallel, potentially significantly reducing the overall test suite execution time:
+
+[source,plaintext]
+----
+ForkJoinPool-1-worker-6 - FooTest::test2
+ForkJoinPool-1-worker-7 - BarTest::test3
+ForkJoinPool-1-worker-3 - FooTest::test1
+ForkJoinPool-1-worker-8 - FooTest::test3
+ForkJoinPool-1-worker-5 - BarTest::test2
+ForkJoinPool-1-worker-4 - BarTest::test1
+----
 
 [[migrating-from-junit4-tips]]
 === Migration Tips


documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -55,21 +55,76 @@ concurrent test execution. It can be enabled and configured using the following
...
+----
+
+On the other hand, we can enable the parallel test execution at a `method` level,
+rather than the `class` level:
+
+[source,properties]
+----
+junit.vintage.execution.parallel.enabled=true
 junit.vintage.execution.parallel.methods=true
-junit.vintage.execution.parallel.pool-size=4
```
YongGoose, line 68: Why did you remove this?

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -45,8 +45,8 @@ concurrent test execution. It can be enabled and configured using the following
 <<running-tests-config-params, configuration parameters>>:
 
 `junit.vintage.execution.parallel.enabled=true|false`::
-  Enable/disable parallel execution (defaults to `false`). Requires opt-in for `classes`
-  or `methods` to be executed in parallel using the configuration parameters below.
+  Enable/disable parallel execution (defaults to `false`).
+Requires opt-in for `classes` or `methods` to be executed in parallel using the configuration parameters below.
```
YongGoose, line 49:
```suggestion
  Requires opt-in for `classes` or `methods` to be executed in parallel using the configuration parameters below.
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -55,21 +55,76 @@ concurrent test execution. It can be enabled and configured using the following
   Enable/disable parallel execution of test methods (defaults to `false`).
 
 `junit.vintage.execution.parallel.pool-size=<number>`::
-  Specifies the size of the thread pool to be used for parallel execution. By default, the
-  number of available processors is used.
+  Specifies the size of the thread pool to be used for parallel execution.
+By default, the  number of available processors is used.
```
YongGoose, line 59:
```suggestion
  By default, the  number of available processors is used.
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -55,21 +55,76 @@ concurrent test execution. It can be enabled and configured using the following
...
 
-Example configuration in `junit-platform.properties`:
+
+Let's assume we have two test classes `FooTest` and `BarTest` with each class containing three unit tests.
+Now, let's enable the parallel execution of test classes:
 
 [source,properties]
 ----
 junit.vintage.execution.parallel.enabled=true
 junit.vintage.execution.parallel.classes=true
```
YongGoose, line 68:
```suggestion
junit.vintage.execution.parallel.enabled=true
junit.vintage.execution.parallel.classes=true
junit.vintage.execution.parallel.pool-size=2
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -55,21 +55,76 @@ concurrent test execution. It can be enabled and configured using the following
...
+ForkJoinPool-1-worker-1 - FooTest::test3
+----
+
+Finally, we can also enable the parallelization at both `class` and `method` level:
+
+[source,properties]
+----
+junit.vintage.execution.parallel.enabled=true
+junit.vintage.execution.parallel.classes=true
+junit.vintage.execution.parallel.methods=true
```
YongGoose, line 113: The default number of threads is `Runtime.getRuntime().availableProcessors()` unless specified otherwise.
So, it would be a good idea to include a configuration setting to define the number of threads.

- https://github.com/junit-team/junit5/blob/main/junit-vintage-engine/src/main/java/org/junit/vintage/engine/execution/VintageExecutor.java#L48

YongGoose (changes requested):
It would be great to add documentation for the following cases as well.

The document looks much better now.
Thank you! 🙂

```
junit.vintage.execution.parallel.enabled=false
junit.vintage.execution.parallel.classes=true
junit.vintage.execution.parallel.methods=true
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
etrandafir93: I added the spaces, as suggested - but do we need them? 

I have added the brake lines to comply with this [sembr](https://sembr.org/) convention, but I won't have an effect on the way the text is displayed. That's why I haven't added any space at the start of the row.

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
etrandafir93: I left this property out to allow the reader to focus on the ones that directly impact the example. 

For the given example with 2 test classes and parallelization enabled per class, we'll always use 2 threads - even if the thread pool size is larger, correct?

YongGoose: > For the given example with 2 test classes and parallelization enabled per class, we'll always use 2 threads - even if the thread pool size is larger, correct?

That’s correct.

While you could explain it by saying that running FooTest and BarTest uses two threads, I believe it is more appropriate to relate it to the size of the thread pool rather than directly linking the number of tests to the number of threads. For example, when running `three tests` in parallel with a thread pool size of `two`, only `two threads` will be used, regardless of the number of tests.

What do you think? @marcphilipp

etrandafir93:
> junit.vintage.execution.parallel.enabled=false
> junit.vintage.execution.parallel.classes=true
> junit.vintage.execution.parallel.methods=true

@YongGoose - thanks for the feedback. I have updated the PR now addressing the comments. 
I have broken down the content into 4 short subsections, as the main section grew. 

I have added the example above as well as the case where parallel execution is enabled but no parallel execution level is defined - to highlight this misconfiguration as well. 

Let me know what you think:)

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
YongGoose: I hope this is helpful to you.

image

image

- https://github.com/junit-team/junit5/pull/4242#discussion_r1948051443

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -45,8 +45,8 @@ concurrent test execution. It can be enabled and configured using the following
 <<running-tests-config-params, configuration parameters>>:
 
 `junit.vintage.execution.parallel.enabled=true|false`::
-  Enable/disable parallel execution (defaults to `false`). Requires opt-in for `classes`
-  or `methods` to be executed in parallel using the configuration parameters below.
+  Enable/disable parallel execution (defaults to `false`).
+ Requires opt-in for `classes` or `methods` to be executed in parallel using the configuration parameters below.
```
YongGoose, line 49: There should definitely be two spaces at the start :)

> I’ve also had more feedback on documentation than on code when contributing. 😂

YongGoose:
> @YongGoose- 피드백 감사합니다. 저는 지금 PR을 업데이트하여 댓글을 해결했습니다. 주요 섹션이 커짐에 따라 콘텐츠를 4개의 짧은 하위 섹션으로 나누었습니다.
> 
> 위의 예와 병렬 실행이 활성화되어 있지만 병렬 실행 수준이 정의되지 않은 경우를 추가하여 이러한 잘못된 구성을 강조했습니다.
> 
> 여러분의 생각을 알려주세요:)

LGTM 👍

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
etrandafir93: oh 😅 fixed now. I wonder if spotless can be configured to catch this

YongGoose: I’m not very familiar with `Spotless`, so I’m not sure if it provides that feature.
If it does, it might be a good idea to discuss it with Marc and consider contributing.
It looks like the file below is used for `Spotless configuration`.

- https://github.com/junit-team/junit5/blob/main/gradle/plugins/common/src/main/kotlin/junitbuild.spotless-conventions.gradle.kts

jbduncan: As a former contributor to Spotless, there's nothing technical stopping us from adopting such a check. It would "just" require implementing a regex (which the Gradle plugin has native support for) or finding a JVM library or CLI tool that does the job.

However, as English uses full stops for ends of sentences, word shortenings (for example, "regarding" to "re.") and inside other words sometimes (for example, "01.01.2025"), this may be tricky to do without heuristics unless we're happy to catch all full stops followed one space or 3+ spaces.

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -45,8 +45,8 @@ concurrent test execution. It can be enabled and configured using the following
 <<running-tests-config-params, configuration parameters>>:
 
 `junit.vintage.execution.parallel.enabled=true|false`::
-  Enable/disable parallel execution (defaults to `false`). Requires opt-in for `classes`
-  or `methods` to be executed in parallel using the configuration parameters below.
+  Enable/disable parallel execution (defaults to `false`).
+  Requires opt-in for `classes` or `methods` to be executed in parallel using the configuration parameters below.
```
marcphilipp, line 49: We currently use a different convention, namely hard-wrapping at 90 characters. Please revert this change and reformat your additions accordingly.

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -55,21 +55,122 @@ concurrent test execution. It can be enabled and configured using the following
...
+
+ForkJoinPool-1-worker-3 - FooTest::test1
+ForkJoinPool-1-worker-2 - FooTest::test2
+ForkJoinPool-1-worker-1 - FooTest::test3
+----
+
+[[migrating-from-junit4-parallel-execution-class-and-method-level]]
+==== Full Parallelization
+
+Finally, we can also enable the parallelization at both `class` and `method` level:
```
marcphilipp, line 117:
```suggestion
Finally, we can also enable the parallelization at both class and method level:
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -55,21 +55,122 @@ concurrent test execution. It can be enabled and configured using the following
...
+[source,plaintext]
+----
+main - BarTest::test1
+main - BarTest::test2
+main - BarTest::test3
+
+main - FooTest::test1
+main - FooTest::test2
+main - FooTest::test3
+----
```
marcphilipp, line 173: This is actually be considered a configuration error and produces a warning. Therefore, we should point that out and omit the example below.

```suggestion
Similarly, tests will be executed sequentially if you enable parallel execution in general
but enable neither class-level nor method-level parallelization.
```

marcphilipp (changes requested):

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
etrandafir93: fixed now

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
etrandafir93: fixed

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
etrandafir93: removed

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,18 +58,100 @@ concurrent test execution. It can be enabled and configured using the following
   Specifies the size of the thread pool to be used for parallel execution. By default, the
   number of available processors is used.
 
-Example configuration in `junit-platform.properties`:
+[[migrating-from-junit4-parallel-execution-class-level]]
+==== Parallelization at Class Level
+
+Let's assume we have two test classes `FooTest` and `BarTest` with each class containing
+three unit tests. Now, let's enable the parallel execution of test classes:
```
marcphilipp, line 65:
```suggestion
three unit tests. Now, let's enable parallel execution of test classes:
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
marcphilipp: I think explaining one concept at a time (`classes`, `methods`, both, pool size, and then sequential execution) would be easier to follow. Therefore, I'd omit it here and introduce a new "Configuring the Pool Size" subsection before the "Sequential Execution" one.

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,18 +58,100 @@ concurrent test execution. It can be enabled and configured using the following
...
+==== Parallelization at Method Level
+
+Alternatively, we can enable the parallel test execution at a `method` level,
+rather than the `class` level:
+
+[source,properties]
+----
+junit.vintage.execution.parallel.enabled=true
 junit.vintage.execution.parallel.methods=true
 junit.vintage.execution.parallel.pool-size=4
```
marcphilipp, line 97: As per my suggestion above.

```suggestion
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,18 +58,100 @@ concurrent test execution. It can be enabled and configured using the following
...
+ForkJoinPool-1-worker-1 - BarTest::test2
+ForkJoinPool-1-worker-2 - FooTest::test2
+ForkJoinPool-1-worker-1 - BarTest::test3
+ForkJoinPool-1-worker-2 - FooTest::test3
+----
+
+[[migrating-from-junit4-parallel-execution-method-level]]
+==== Parallelization at Method Level
+
+Alternatively, we can enable the parallel test execution at a `method` level,
```
marcphilipp, line 90:
```suggestion
Alternatively, we can enable parallel test execution at a `method` level,
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,18 +58,100 @@ concurrent test execution. It can be enabled and configured using the following
...
+[source,properties]
+----
+junit.vintage.execution.parallel.enabled=true
 junit.vintage.execution.parallel.methods=true
 junit.vintage.execution.parallel.pool-size=4
 ----
 
-With these properties set, the `VintageTestEngine` will execute tests in parallel,
-potentially significantly reducing the overall test suite execution time.
+Therefore, the test methods from within a class will be executed in parallel, while
```
marcphilipp, line 100:
```suggestion
Therefore, the test methods within each class will be executed in parallel, while
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,18 +58,100 @@ concurrent test execution. It can be enabled and configured using the following
...
+
+ForkJoinPool-1-worker-3 - FooTest::test1
+ForkJoinPool-1-worker-2 - FooTest::test2
+ForkJoinPool-1-worker-1 - FooTest::test3
+----
+
+[[migrating-from-junit4-parallel-execution-class-and-method-level]]
+==== Full Parallelization
+
+Finally, we can also enable the parallelization at both class and method level:
```
marcphilipp, line 117:
```suggestion
Finally, we can also enable parallelization at both class and method level:
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,18 +58,100 @@ concurrent test execution. It can be enabled and configured using the following
...
+
+[source,properties]
+----
+junit.vintage.execution.parallel.enabled=true
+junit.vintage.execution.parallel.classes=true
+junit.vintage.execution.parallel.methods=true
+junit.vintage.execution.parallel.pool-size=8
+----
+
+With these properties set, the `VintageTestEngine` will execute all the tests classes and
```
marcphilipp, line 127:
```suggestion
With these properties set, the `VintageTestEngine` will execute all tests classes and
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,18 +58,100 @@ concurrent test execution. It can be enabled and configured using the following
...
+ForkJoinPool-1-worker-3 - FooTest::test1
+ForkJoinPool-1-worker-8 - FooTest::test3
+ForkJoinPool-1-worker-5 - BarTest::test2
+ForkJoinPool-1-worker-4 - BarTest::test1
+----
+
+[[migrating-from-junit4-parallel-execution-disabled]]
+==== Sequential Execution
+
+On the other hand, if we disable the parallel execution, the `VintageTestEngine`
```
marcphilipp, line 143:
```suggestion
On the other hand, if we disable parallel execution, the `VintageTestEngine`
```

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,18 +58,100 @@ concurrent test execution. It can be enabled and configured using the following
...
+will execute the tests sequentially, regardless of the other properties:
```
marcphilipp, line 144:
```suggestion
will execute all tests sequentially, regardless of the other properties:
```

marcphilipp (changes requested):

YongGoose:
> I think explaining one concept at a time (classes, methods, both, pool size, and then sequential execution) would be easier to follow. Therefore, I'd omit it here and introduce a new "Configuring the Pool Size" subsection before the "Sequential Execution" one.

Actually, explaining the pool size together was my own idea, as mentioned in the comment below. 😅
- https://github.com/junit-team/junit5/pull/4336#discussion_r1967079410

@etrandafir93 I apologize for any confusion it may have caused.
It seems better to explain it in a separate section, following Marc’s suggestion.

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
etrandafir93: fixed

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
etrandafir93: fixed now

etrandafir93:
> > I think explaining one concept at a time (classes, methods, both, pool size, and then sequential execution) would be easier to follow. Therefore, I'd omit it here and introduce a new "Configuring the Pool Size" subsection before the "Sequential Execution" one.
> 
> Actually, explaining the pool size together was my own idea, as mentioned in the comment below. 😅
> 
> * [Improve JUnit Vintage docs about parallel execution #4336 (comment)](https://github.com/junit-team/junit5/pull/4336#discussion_r1967079410)
> 
> @etrandafir93 I apologize for any confusion it may have caused. It seems better to explain it in a separate section, following Marc’s suggestion.

@YongGoose No worries, it was a good discussion 👍 Also, I've noticed that the pool-size parameter is used internally to define the executor's parallelism level, which is slightly different. I've added a short sentence about this.

@marcphilipp thanks for the feedback, I have applied your suggestions now, let me know what you think.

documentation/src/docs/asciidoc/user-guide/migration-from-junit4.adoc
```diff
@@ -58,7 +58,87 @@ concurrent test execution. It can be enabled and configured using the following
...
+ForkJoinPool-1-worker-2 - FooTest::test2
+ForkJoinPool-1-worker-1 - BarTest::test3
+ForkJoinPool-1-worker-2 - FooTest::test3
+----
+
+[[migrating-from-junit4-parallel-execution-method-level]]
+==== Parallelization at Method Level
+
+Alternatively, we can enable parallel test execution at a `method` level,
+rather than the `class` level:
```
marcphilipp, line 90:
```suggestion
Alternatively, we can enable parallel test execution at a method level,
rather than the class level:
```

marcphilipp (approved):

marcphilipp:
@etrandafir93 Thank you! :+1:

PR #4442, opened 2025-04-02
Introduce feature flag for auto-closing `AutoCloseable` in Jupiter's `ExtensionContext.Store`

### LINKED ISSUE #4434: Introduce feature flag for auto-closing `AutoCloseable` in Jupiter's `ExtensionContext.Store`
Currently, only instances of `CloseableResource` that are stored in an `ExtensionContext.Store` are closed automatically when the `ExtensionContext` is closed. However, items stored in the session-/request-level stores only need to implement `AutoCloseable`. To address this confusing difference in behavior, we should introduce a configuration parameter that allows to treat `AutoCloseable` like `CloseableResource` in Jupiter's stores. It should be "on" by default in 5.13.

## Deliverables

- [x] Introduce configuration parameter for enabling auto-closing `AutoCloseable` objects (`true` by default)
- [x] Deprecate `CloseableResource`
- ~Make `CloseableResource` extend `AutoCloseable` so existing implementations would keep working~ Doesn't work because `CloseableResource#close` is allowed to throw `Throwable`, not just `Exception`
- [x] Log warning if an object is ~put into~ encountered while closing the `Store` that implements `CloseableResource` but not `AutoCloseable` (if the feature is on)
- [x] Document migration path for external extensions (implement both `CloseableResource` and `AutoCloseable` should they need to support pre-5.13 versions)

YongGoose:
## Overview

fix #4434 

---

---

### Definition of Done


YongGoose:
@marcphilipp 

> Treat AutoCloseable like CloseableResource in Jupiter's stores when configuration parameter is `on` (true)

Do you think we should add a test case for this scenario?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -503,16 +504,19 @@ interface Store {
 		 * inverse order they were added in.
 		 *
 		 * @since 5.1
+		 * @deprecated without replacement
```
marcphilipp, line 507:
```suggestion
		 * @deprecated Please extend {@code AutoCloseable} directly.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -503,16 +504,19 @@ interface Store {
...
+		interface CloseableResource extends AutoCloseable {
 
 			/**
 			 * Close underlying resources.
 			 *
-			 * @throws Throwable any throwable will be caught and rethrown
+			 * @throws Exception any exception will be caught and rethrown
 			 */
-			void close() throws Throwable;
+			void close() throws Exception;
```
marcphilipp, line 519: Changing this would be binary-incompatible. Thus, I don't think we can have `CloseableResource` extend `AutoCloseable` after all. I guess we could support both `CloseableResource` and `AutoCloseable` instead as separate interfaces. I'll discuss this with the team.

YongGoose: I agree with you.
Please discuss this with the team and share the outcome! 👍🏻

marcphilipp: @YongGoose We've updated the issue description in #4434.

YongGoose: Got it!
I will close this PR, start a new task, and then create a new PR.

Also, if you have some time, could you briefly explain how the discussion went on the issue below? 🙂
- https://github.com/junit-team/junit5/issues/1767

marcphilipp: > Also, if you have some time, could you briefly explain how the discussion went on the issue below? 🙂

I just posted an update: https://github.com/junit-team/junit5/issues/1767#issuecomment-2778525332

PR #4466, opened 2025-04-09
Fix handling of `CleanupMode.ON_SUCCESS`

### LINKED ISSUE #4464: `@TempDir` directories of successful tests are not cleaned with `ON_SUCCESS` mode
In our test suite we’ve noticed that if any test fails (for whatever reason), then a lot of subsequent tests fail due to out-of-disk problems. This disk space is taken by temporary JUnit directories ([`org.junit.jupiter.api.io.TempDir`](https://github.com/junit-team/junit5/blob/main/junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDir.java)). We use `junit.jupiter.tempdir.cleanup.mode.default=ON_SUCCESS` property, so theoretically only directories for failed tests should be preserved, but we also see many directories for successful ones.

### Expected behavior

All temporary directories of successful tests should be deleted. It worked this way in `5.10.0`.

### Actual behavior

If one test fails, then temporary directories of subsequent successful tests are preserved. First observed in `5.10.1`.

I think it was introduced in https://github.com/junit-team/junit5/issues/3510 . Looks like [`selfOrChildFailed`](https://github.com/junit-team/junit5/blob/c83bfbc7e2739baae273051de79473b2e3ed640c/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java#L330-L335) starts treating all subsequent tests as failed.

## Steps to reproduce

I’ve put a minimal reproducible example in this repository: https://github.com/msulima/junit-tmpdir-cleanup

## Context

 - JUnit: 5.10.1, 5.11.0, 5.12.1
 - Maven 3.9.9 + surefire 3.3.0

marcphilipp:
## Overview

- **Add 5.12.2 release notes from template**
- **Fix handling of `CleanupMode.ON_SUCCESS`**

Fixes #4464.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
@@ -129,12 +129,17 @@ class TempDirectory implements BeforeAllCallback, BeforeEachCallback, ParameterR
 	}
 
 	private static void installFailureTracker(ExtensionContext context) {
-		context.getStore(NAMESPACE).put(FAILURE_TRACKER, (CloseableResource) () -> context.getParent() //
-				.ifPresent(parentContext -> {
-					if (selfOrChildFailed(context)) {
-						parentContext.getStore(NAMESPACE).put(CHILD_FAILED, true);
-					}
-				}));
+		context.getParent() //
+				.filter(parentContext -> !context.getRoot().equals(parentContext)) //
+				.ifPresent(parentContext -> installFailureTracker(context, parentContext));
+	}
+
+	private static void installFailureTracker(ExtensionContext context, ExtensionContext parentContext) {
+		context.getStore(NAMESPACE).put(FAILURE_TRACKER, (CloseableResource) () -> {
+			if (selfOrChildFailed(context)) {
+				getContextSpecificStore(parentContext).put(CHILD_FAILED, true);
+			}
+		});
 	}
 
 	private void injectStaticFields(ExtensionContext context, Class<?> testClass) {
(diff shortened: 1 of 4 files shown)


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -286,7 +291,11 @@ static CloseablePath createTempDir(TempDirFactory factory, CleanupMode cleanupMo
 
 	private static boolean selfOrChildFailed(ExtensionContext context) {
 		return context.getExecutionException().isPresent() //
-				|| context.getStore(NAMESPACE).getOrDefault(CHILD_FAILED, Boolean.class, false);
+				|| getContextSpecificStore(context).getOrDefault(CHILD_FAILED, Boolean.class, false);
+	}
+
+	private static ExtensionContext.Store getContextSpecificStore(ExtensionContext context) {
+		return context.getStore(NAMESPACE.append(context));
```
marcphilipp, line 298: :speech_balloon: This is the important bit: by including the context in the namespace, we prevent it from looking up the key on the parent

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
```diff
@@ -129,12 +129,17 @@ public void beforeEach(ExtensionContext context) {
...
 
 	private static void installFailureTracker(ExtensionContext context) {
-		context.getStore(NAMESPACE).put(FAILURE_TRACKER, (CloseableResource) () -> context.getParent() //
-				.ifPresent(parentContext -> {
-					if (selfOrChildFailed(context)) {
-						parentContext.getStore(NAMESPACE).put(CHILD_FAILED, true);
-					}
-				}));
+		context.getParent() //
+				.filter(parentContext -> !context.getRoot().equals(parentContext)) //
```
marcphilipp, line 133: :speech_balloon: This line is just an optimization because there never are temp dirs on the engine level so we don't need to store the information there because it will never be read.

sormuras: Unrelated and potentially breaking - but should be okay.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/TempDirectory.java
sormuras: Sounds reasonable.

sormuras (approved):
Implementation and new test cases look good.

PR #4481, opened 2025-04-14
Relax Kotlin contracts for `Executable` parameters

marcphilipp:
- **Relax Kotlin contracts for `Executable` parameters**
- **Treat Kotlin compiler warnings as errors again**

## Overview

---

---

### Definition of Done


jupiter-tests/src/test/kotlin/org/junit/jupiter/api/KotlinAssertTimeoutAssertionsTests.kt
```diff
@@ -156,26 +156,26 @@ internal class KotlinAssertTimeoutAssertionsTests {
     }
 
     @Test
-    fun `assertTimeout with value initialization in lambda`() {
-        val value: Int
+    fun `assertTimeout with value assignment in lambda`() {
```
awelless, line 159: The problem is that this test will pass even when no contract is specified. The `AT_LEAST_ONE` contract allows us to assign `val`s in lamda body. A test like this would be a better candidate, as it will pass only with `AT_LEAST_ONCE` or `EXACTLY_ONCE` contracts. This applies to other changes in this file too.
```kt
       @Test
    fun `assertTimeout with value assignment in lambda`() {
        val value: Int

        assertTimeout(ofMillis(500)) {
            value = 10 // Val can be assigned in the message supplier lambda.
            assertEquals(10, value)
        }
    }
```

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -411,7 +411,7 @@ inline fun <R> assertDoesNotThrow(
 @PublishedApi
 internal inline fun <R> evaluateAndWrap(executable: () -> R): ThrowingSupplier<R> {
     contract {
-        callsInPlace(executable, EXACTLY_ONCE)
+        callsInPlace(executable, AT_MOST_ONCE)
```
awelless, line 414: Although this makes the code compile, I'm still not entirely happy that we must mark this as `AT_MOST_ONCE` only to make kotlin `assertDoesNotThrow` compatible with its java counterpart.

If we were able to just run lambda in kotlin, we'd be able to mark all `assertDoesNotThrow` methods as `EXACTLY_ONCE`. That being set, this will cause duplication of `AssertDoesNotThrow` logic which doesn't sound optimal either. 
```kt
internal inline fun <R> evaluateAndWrap(executable: () -> R): ThrowingSupplier<R> {
    contract {
        callsInPlace(executable, EXACTLY_ONCE)
    }

    return try {
        executable()
    } catch (throwable: Throwable) {
        // logic from java's AssertDoesNotThrow
        UnrecoverableExceptions.rethrowIfUnrecoverable(t)
        throw AssertDoesNotThrow.createAssertionFailedError(messageOrSupplier, t)
    }
}
```

awelless (commented):
The tests prevent me from approval.
The decision about `assertDoesNotThrow` I leave to you

jupiter-tests/src/test/kotlin/org/junit/jupiter/api/KotlinAssertTimeoutAssertionsTests.kt
marcphilipp: Thanks for the suggestion. Applied in 5cac4e3a902a58982d1aca7b063a06c83b3340df

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
marcphilipp: Done in 0646ebae7f1ddd637a47db1a57acf4a4098e0933

marcphilipp:
@awelless Thanks for the review! Do you have time to take another look at my latest changes?

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -374,7 +379,7 @@ inline fun <R> assertDoesNotThrow(
     executable: () -> R
 ): R {
     contract {
-        callsInPlace(executable, EXACTLY_ONCE)
```
awelless, line 377: This executable should be called `EXACTLY_ONCE`, as other `assertDoesNotThrow` methods.

awelless (approved):
The changes look good, besides the contract in one `assertDoesNotThrow`.
Approved, assuming the comment will be addressed

marcphilipp:
Superseded by #4551

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
marcphilipp: Good catch! Addressed in [`ef78d82` (#4551)](https://github.com/junit-team/junit5/pull/4551/commits/ef78d827a708d5a14b7af2a0470bbfbb9e189d42)

PR #4515, opened 2025-05-05
Spike support for Kotlin suspending functions

marcphilipp:
## Overview

Issue: #1914

Given the function definition of `suspend fun suspendingTest(message: String): Unit`, the Kotlin compiler generates byte code that results in Java's reflection API seeing:

```
public final java.lang.Object suspendingTest(
  java.lang.String,kotlin.coroutines.Continuation<? super kotlin.Unit>)
```

The return type is `java.lang.Object` and an extra parameter of type `kotlin.coroutines.Continuation<? super T>` where `T` is the declared return type is added. Since `Continuation<...>` can be used as a method parameter for a regular function, its presence is insufficient to determine whether a method is a suspending function. Thus, a dependency on `kotlin-reflect` is needed so `KFunction.isSuspend` can be called.

If a suspending function is detected, it can be called using `KCallables.callSuspend` (also from `kotlin-reflect`) while wrapping it in `runBlocking` from `kotlinx-coroutines`.

To avoid the extra dependencies from being added to `junit-jupiter-engine`, the prototype in this PR adds a new `junit-jupiter-kotlin` artifact. The former defines a new `MethodAdapterFactory` interface that the latter implements via `ServiceLoader`. The returned `MethodAdapter` is used for validation (return type), parameter resolution (ignore the `Continuation` parameter), and invocation (via `KCallables.callSuspend` rather than `ReflectionUtils`).

The PR adds support for using `suspend` on `@Test`, `@TestTemplate` (including but not limited to `@ParameterizedTest` and `@RepeatedTest`), `@TestFactory`, and all lifecycle methods. Moreover, it also adds support for calling suspend methods via `ExecutableInvoker` which is used to support `@After`/`@BeforeClassTemplateInvocationCallback` lifecycle methods. See the `KotlinSuspendFunctionsTests` test class for a demo.

### Open Questions

* Can the extra `junit-jupiter-kotlin` artifact be avoided somehow without forcing Kotlin dependencies on plain-Java users and without forcing Kotlin users to declare extra dependencies on `kotlin-reflect` and `kotlinx-coroutines`?

---

---

### Definition of Done


junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -195,7 +196,15 @@ default String generateDisplayNameForMethod(List<Class<?>> enclosingInstanceType
...
 		Preconditions.notNull(method, "Method must not be null");
-		return '(' + ClassUtils.nullSafeToString(Class::getSimpleName, method.getParameterTypes()) + ')';
+		if (method.getParameterCount() == 0) {
+			return "()";
+		}
+		var parameterTypes = method.getParameterTypes();
+		if (parameterTypes[parameterTypes.length - 1].getName().equals("kotlin.coroutines.Continuation")) {
+			// Kotlin suspend functions have a Continuation parameter at the end that should not be included
+			parameterTypes = Arrays.copyOf(parameterTypes, parameterTypes.length - 1);
+		}
```
marcphilipp, line 206: :speech_balloon: This removes the `Continuation` parameter from display names because it's generated by the Kotlin compiler.

junit-jupiter-engine/junit-jupiter-engine.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
-	id("junitbuild.kotlin-library-conventions")
+	id("junitbuild.java-library-conventions")
```
marcphilipp, line 2: :speech_balloon: We were not actually using Kotlin in this project.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedClassContext.java
```diff
@@ -155,9 +153,6 @@ private static <A extends Annotation> List<ArgumentSetLifecycleMethod> findLifec
 		List<Method> methods = findAnnotatedMethods(testClass, annotationType, traversalMode);
 
 		return methods.stream() //
-				.filter(ModifierSupport::isNotPrivate) //
-				.filter(testInstanceLifecycle == PER_METHOD ? ModifierSupport::isStatic : __ -> true) //
-				.filter(ReflectionUtils::returnsPrimitiveVoid) //
```
marcphilipp, line 160: :speech_balloon: This is already validated in `junit-jupiter-engine` so there's no need to do so again.

junit-jupiter-kotlin/junit-jupiter-kotlin.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+	//
+	//Note: The summary above may be incomplete. Please check the full output below for more hints.
+	//Resolution failed. Capabilities satisfying the following requirements could not be found:
+	//
+	//    [<<INITIAL>>]
+	//      ⇒ osgi.identity: (osgi.identity=junit-jupiter-kotlin)
+	//          ⇒ [junit-jupiter-kotlin version=6.0.0.SNAPSHOT]
+	//              ⇒ osgi.wiring.package: (osgi.wiring.package=kotlinx.coroutines)
+	enabled = false
+}
```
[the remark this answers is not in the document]
marcphilipp: Thanks for the quick reply! I tried that but it doesn't contain `kotlinx.coroutines`.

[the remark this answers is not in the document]
bjhargrave: I don't think you want to maintain a modified version of that jar to add OSGi metadata to it.

Perhaps you should try an optional resolution for `kotlinx.*` similar to:

https://github.com/junit-team/junit5/blob/c27907daeef1d7bbaee456e7330de849e0a497a5/gradle/plugins/common/src/main/kotlin/junitbuild.osgi-conventions.gradle.kts#L41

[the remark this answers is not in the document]
marcphilipp: I went that route for now in 34b9700e8f53e57fe32e733ee5ee7b9d9d743764

jlink:
> @sbrannen Thanks for the pointers!
> 
> I think we have two options:
> 
>     1. Make the Kotlin dependencies optional (like Spring does)
> 
>     2. Create an extra artifact/variant that carries the dependencies
> 
> 
> As far as manual effort is concerned, (1) requires adding at least two dependencies (`kotlin-reflect` and `kotlinx-coroutines`) while (2) requires adding only one (`junit-jupiter-kotlin` or whichever name we decide for). If we add support for some other Kotlin feature, it might require additional dependencies which we could add to our artifact/variant in case of (2) without users having to modify their builds again. Moreover, I think (2) is cleaner and carries a smaller risk of running into `ClassNotFoundErrors` etc. at runtime.
> 
> Are there any other pros/cons?

IMO having an explicit artifact for Kotlin support is sending a clear message: JUnit Jupiter wants to support Kotlin. That's a win in my book. And it opens the door for further support, e.g. providing Kotlin-style API for some constructs.

junit-jupiter-kotlin/src/main/kotlin/org/junit/jupiter/kotlin/coroutines/KotlinSuspendFunctionAdapterFactory.kt
```diff
@@ -0,0 +1,72 @@
...
+                }
+            return object : MethodAdapter {
+                override fun getMethod() = method
+
+                override fun getName() = kotlinFunction.name
+
+                override fun getReturnType() =
+                    with(kotlinFunction.returnType.jvmErasure) {
+                        if (this == Unit::class) {
+                            Void.TYPE
```
marcphilipp, line 46:
```suggestion
                            void.class
```

marcphilipp:
Superseded by #4530.

PR #4557, opened 2025-05-20
Add nullability annotations and compile-time checks

marcphilipp:
## Overview

This PR uses JSpecify's annotations to annotate packages as `@NullMarked` which implies fields, parameters, return types etc. in those package are not nullable by default. If `null` is allowed, the `@Nullable` annotation is used to indicate that.

At compile-time Error Prone with the NullAway plugin is used to verify consistency of the annotations causing the build to fail, for example, if `null` is passed as a parameter where it isn't allowed.

---

---

### Definition of Done


sormuras:
> This PR uses JSpecify's annotations to annotate packages as `@NullMarked` which implies fields, parameters, return types etc. in those package are not nullable by default. If null is allowed, the `@Nullable` annotation is used to indicate that.

Would it make sense to annotate modules instead of packages with `@NullMarked` instead?

marcphilipp:
> Would it make sense to annotate modules instead of packages with `@NullMarked` instead?

Yes, I think that's a good idea! I went package by package but when a module is done moving the annotation to the module declaration makes sense. :+1:

I'll give that a shot.

marcphilipp:
Unfortunately, NullAway does not support that:

- https://github.com/uber/NullAway/issues/1083

marcphilipp:
I think a found a good "interim" solution:

* For NullAway, the config is now in the build setup because it doesn't yet support module descriptors
* For IntelliJ, the config is now in the module descriptors

junit-platform-commons/src/main/java/module-info.java
```diff
@@ -8,15 +8,20 @@
...
  *
  * @since 1.0
  */
+@NullMarked
 module org.junit.platform.commons {
 	requires java.logging;
 	requires java.management; // needed by RuntimeUtils to determine input arguments
+
 	requires static transitive org.apiguardian.api;
+	requires static transitive org.jspecify;
```
marcphilipp, line 24: We should discuss whether this is the right "scope". It requires downstream projects  to have the JSpecify jar on their module path (should they be compiled using the module path).

marcphilipp: `requires static` would make it optional, but we'd have to suppress [warnings like these](https://ge.junit.org/s/bcfbrjy3gr5yk/console-log/task/:junit-platform-commons:compileJava?anchor=45&page=1).

sormuras: Mh, yeah. It ships with a compiled module descriptor, though: [org.jspecify](https://github.com/sormuras/modules/blob/d2d9fc919c4ab285c40ef75228887d88ff76c40a/com.github.sormuras.modules/com/github/sormuras/modules/modules.properties#L6374C1-L6374C97)

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+	errorprone(dependencyFromLibs("nullaway"))
+}
+
+nullaway {
+	annotatedPackages.add(javaModuleName)
+}
+
+tasks.withType<JavaCompile>().configureEach {
+	options.errorprone {
+		disableAllChecks = true
```
marcphilipp, line 23: We can later (after this PR) enable additional checks provided by Error Prone but I didn't want to mix those changes into this already large set of changes.

platform-tooling-support-tests/src/archUnit/java/platform/tooling/support/tests/ArchUnitTests.java
```diff
@@ -76,6 +80,32 @@ class ArchUnitTests {
...
+				.filter(pkg -> !pkg.getClasses().isEmpty()) //
+				.toList();
+		assertThat(subpackages).isNotEmpty();
+
+		var violations = subpackages.stream() //
+				.filter(pkg -> !pkg.isAnnotatedWith(NullMarked.class)) //
+				.map(JavaPackage::getName) //
+				.sorted();
+		assertThat(violations).describedAs("The following packages are missing the @NullMarked annotation").isEmpty();
+	}
```
marcphilipp, line 107: This check will help us to avoid forgetting to add the annotation on new packages.

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -565,7 +566,7 @@ static boolean isWideningConversion(Class<?> sourceType, Class<?> targetType) {
 	 * @return the corresponding wrapper type or {@code null} if the
 	 * supplied type is {@code null} or not a primitive type
 	 */
-	public static Class<?> getWrapperType(Class<?> type) {
+	public static @Nullable Class<?> getWrapperType(Class<?> type) {
```
scordio, line 569: @marcphilipp given that the Javadoc mentions:

```
* @return the corresponding wrapper type or {@code null} if the
* supplied type is {@code null} or not a primitive type
```

should the parameter also have `@Nullable`?

```java
public static @Nullable Class<?> getWrapperType(@Nullable Class<?> type)
```

(I spotted it during the rebase of #4219)

After a quick look, I think there are no unit tests for `getWrapperType()` in `ReflectionUtilsTests`.

Happy to submit a PR to improve both topics, if you agree 🙂

scordio: From my branch:

https://github.com/junit-team/junit5/blob/791778e4189e9e19a7908500b49113d21d20014e/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/TypeDescriptor.java#L61-L64

triggered this, which is somehow related to the following on `main`:

https://github.com/junit-team/junit5/blob/261c2f5a1576ea29add753b92499e7cf29342f23/junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/ConversionSupport.java#L117

I try to figure out why the warning doesn't happen on `main` 🤔

CONTRIBUTING.md
```diff
@@ -120,6 +120,17 @@ See [`ExtensionContext`](junit-jupiter-api/src/main/java/org/junit/jupiter/api/e
...
 
 
+### Nullability
+
+This project uses JSpecify's annotation to indicate nullability. In general, the approach
+is as follows:
+
+- The Gradle build is set up to treat all code as being `@NullMarked`
+- The descriptor of each module is annotated with `@NullMarked` for IDEs such as IntelliJ
+  IDEA to treat code correctly.
```
scordio, line 130: @marcphilipp IIUC, this eventually didn't happen due to uber/NullAway#1083, and you instead went for annotating the packages, right?

marcphilipp: Good catch, thank you! Updated in df24efbe7759018c2fa69a1c0b418299ab4eed25

PR #4574, opened 2025-05-24
Add List-based support to Arguments API

### LINKED ISSUE #4535: Introduce `List` support in the `Arguments` API for parameterized tests
## Overview

Please excuse the somewhat long context, which I feel is necessary to establish the need for this ticket.

I run golden tests by reading from a zip file, that has pairs of input and output files. Please note that the following problem will exist even if these in/out files were read from a directory instead of a zip file.

Each individual test is fed by a `@MethodSource`, that calls the zip utility with two functions, that are invoked for each in/out pair, respectively. Since the order of the in/out files is nondeterministic, the results of these function invocations are saved in a map. The code in question can be seen [here](https://github.com/asarkar/cses-java/blob/main/sorting%2Bsearching/src/test/java/util/ArgumentsProvider.java).

Here’s where the `Arguments` API is lacking for non-trivial use cases. 
1. There’s no way to concat two `Arguments` or add an item to an existing `Arguments`. 
2. An `Arguments` can only be created from an array, not a collection. 
3. In the above use case, the return type of the in/out functions could be `Arguments`, and we would concat the two `Arguments`, either by append or prepend, based on whether in or out is encountered first. We could also create the `Arguments` from collections, not arrays. Currently, all of these have to done by first pulling the array out of an `Arguments`, creating a new one, and copying stuff over.

## Original Proposal

- Add a constructor that accepts a `Collection`.
- Add a method `append/add/addLast` that appends an item to an existing `Arguments`, and returns a new one.
- Add a method `prepend`/`addFirst`
- Add a method `concat` that appends all items of the `Arguments` parameter to the `Arguments` it’s called on.
- Add a method `size`.

## Current Deliverables

- [ ] Add `static Arguments Arguments.of(List<@Nullable Object>)`
- [ ] Add `static Arguments Arguments.arguments(List<@Nullable Object>)`
- [ ] Add `static ArgumentSet Arguments.argumentSet(String, List<@Nullable Object>)`
- [ ] Add `default List<@Nullable Object> Arguments.toList()`, implemented as `new ArrayList<>(List.of(get()))` (or similar) so that the returned `List` is mutable.

mariokhoury4:
## Overview

This PR adds List-based support to the `Arguments` API for parameterized tests.

### New additions:
- `Arguments.of(List<@Nullable Object>)`
- `Arguments.arguments(List<@Nullable Object>)` (alias)
- `Arguments.argumentSet(String, List<@Nullable Object>)`
- `Arguments.toList()` — returns a mutable `List<@Nullable Object>`

These additions make it easier to dynamically build arguments from collections when using `@ParameterizedTest`.

All additions are tested in `ArgumentsTests.java`.

Fixes #4535

---

---

### Definition of Done


junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptyArgumentsProvider.java
```diff
@@ -57,7 +57,7 @@ public Stream<? extends Arguments> provideArguments(ParameterDeclarations parame
 			return Stream.of(arguments(Collections.emptySet()));
 		}
 		if (List.class.equals(parameterType)) {
-			return Stream.of(arguments(Collections.emptyList()));
+			return Stream.of(Arguments.of((Object) Collections.emptyList()));
```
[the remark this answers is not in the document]
sbrannen: Well, we unfortunately overlooked the fact that a single `List` is also a single `Object`, so we definitely cannot go with this as-is.

Existing code must still compile without adding casts.

Though, our in defense, it is easy to overlook something like that when "coding in the browser". 😉 

> Maybe `copyOf`, `argumentsOf`, and `argumentSetOf()`?

I would rather go with something consistent, instead of having "copy" in only one of the names.

How about something like `fromList()`, `argumentsFromList()`, and `argumentSetFromList()`?

We could also drop the `List` part of the names, if we are somehow _sure_ we will never introduce yet-another-set-of-overloads. So, something like `from()`, `argumentsFrom()`, and `argumentSetFrom()`.

Thoughts?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -172,4 +176,72 @@ public String toString() {
 
 	}
 
+	/**
```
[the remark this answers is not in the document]
mariokhoury4: Thanks for the suggestion! I'll move the method next to the other static factory methods to keep things consistent.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptyArgumentsProvider.java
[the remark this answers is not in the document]
marcphilipp: +1 on the shorter variants. We might even get away with adding overloads to that in the future because "conflicts" with `List` seem less likely than with `Object...`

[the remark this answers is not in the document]
mariokhoury4: I have updated the code with the suggestions!

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -125,6 +129,80 @@ static ArgumentSet argumentSet(String name, @Nullable Object... arguments) {
...
+	 * Factory method for creating an instance of {@code Arguments} based on
+	 * the supplied {@link List} of {@code arguments}.
+	 *
+	 * @param arguments the arguments to be used for an invocation of the test
+	 * method; must not be {@code null} but may contain {@code null}.
+	 * @return an instance of {@code Arguments}; never {@code null}.
+	 * @since 6.0
```
sbrannen, line 139:
```suggestion
	 * @param arguments the arguments to be used for an invocation of the test
	 * method; must not be {@code null} but may contain {@code null}
	 * @return an instance of {@code Arguments}; never {@code null}
	 * @since 6.0
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -125,6 +129,80 @@ static ArgumentSet argumentSet(String name, @Nullable Object... arguments) {
...
+	 * Factory method for creating an instance of {@code Arguments} based on
+	 * the supplied {@link List} of {@code arguments}.
+	 *
+	 * <p>This method is an <em>alias</em> for {@link Arguments#from} and is
+	 * intended to be used when statically imported &mdash; for example, via:
+	 * {@code import static org.junit.jupiter.params.provider.Arguments.argumentsFrom;}
+	 *
+	 * @param arguments the arguments to be used for an invocation of the test
+	 * method; must not be {@code null} but may contain {@code null}.
+	 * @return an instance of {@code Arguments}; never {@code null}.
```
sbrannen, line 158:
```suggestion
	 * @param arguments the arguments to be used for an invocation of the test
	 * method; must not be {@code null} but may contain {@code null}
	 * @return an instance of {@code Arguments}; never {@code null}
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -125,6 +129,80 @@ static ArgumentSet argumentSet(String name, @Nullable Object... arguments) {
...
+	 *
+	 * <p>This method is a convenient alternative to
+	 * {@link #argumentSet(String, Object...)} when working with {@link List}
+	 * based inputs.
+	 *
+	 * @param name the name of the argument set; must not be {@code null}
+	 * or blank.
+	 * @param arguments the arguments to be used for an invocation of the test
+	 * method; must not be {@code null} but may contain {@code null}.
+	 * @return an {@code ArgumentSet}; never {@code null}.
```
sbrannen, line 179:
```suggestion
	 * @param name the name of the argument set; must not be {@code null}
	 * or blank
	 * @param arguments the arguments to be used for an invocation of the test
	 * method; must not be {@code null} but may contain {@code null}
	 * @return an {@code ArgumentSet}; never {@code null}
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -125,6 +129,80 @@ static ArgumentSet argumentSet(String name, @Nullable Object... arguments) {
...
+
+	/**
+	 * Convert the arguments to a new mutable {@link List} containing the same
+	 * elements as {@link #get()}.
+	 *
+	 * <p>This is useful for test logic that benefits from {@code List}
+	 * operations such as filtering, transformation, or assertions.
+	 *
+	 * @return a mutable List of arguments; never {@code null} but may contain
+	 * {@code null}.
```
sbrannen, line 198:
```suggestion
	 * @return a mutable List of arguments; never {@code null} but may contain
	 * {@code null}
```

sbrannen (changes requested):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -125,6 +129,80 @@ static ArgumentSet argumentSet(String name, @Nullable Object... arguments) {
...
+	 * the supplied {@link List} of {@code arguments}.
+	 *
+	 * @param arguments the arguments to be used for an invocation of the test
+	 * method; must not be {@code null} but may contain {@code null}.
+	 * @return an instance of {@code Arguments}; never {@code null}.
+	 * @since 6.0
+	 * @see #argumentsFrom(List)
+	 */
+	@API(status = EXPERIMENTAL, since = "6.0")
+	static Arguments from(List<@Nullable Object> arguments) {
```
vlsi: Should this use generic variance?
Otherwise users would not be able to pass a list with non-nullable values.

```suggestion
	static Arguments from(List<?> arguments) {
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -125,6 +129,80 @@ static ArgumentSet argumentSet(String name, @Nullable Object... arguments) {
...
+	 * {@code import static org.junit.jupiter.params.provider.Arguments.argumentsFrom;}
+	 *
+	 * @param arguments the arguments to be used for an invocation of the test
+	 * method; must not be {@code null} but may contain {@code null}.
+	 * @return an instance of {@code Arguments}; never {@code null}.
+	 * @since 6.0
+	 * @see #argumentSet(String, Object...)
+	 */
+	@API(status = EXPERIMENTAL, since = "6.0")
+	static Arguments argumentsFrom(List<@Nullable Object> arguments) {
```
vlsi:
```suggestion
	static Arguments argumentsFrom(List<?> arguments) {
```

vlsi (commented):
Just wondering: have you considered `Iterable<? extends @Nullable Object>` rather than using a plain `List`?

---

Based on my experience with checkerframework's `@Nullable`, I would suggest `List<? extends @Nullable Object>` rather than `List<Object>`.

I've asked jspecify to clarify the documentation anyways: https://github.com/jspecify/jspecify/issues/748

At the same time, it would be nice to add tests that pass something like `List<String>` to the newly added methods.

For instance, `from(Files.readAllLines("path"))`. I guess the code won't compile if the method uses `List<Object>` signatures.

vlsi:
One more thing: `List<?>` is a good enough replacement for `List<? extends @Nullable Object>`. In this case, the method should probably have a signature like `from(Iterable<?>)`

marcphilipp:
> One more thing: `List<?>` is a good enough replacement for `List<? extends @Nullable Object>`

I didn't know that but sounds like you're right, according to the [spec](https://jspecify.dev/docs/spec/#unbounded-wildcard):

> If an unbounded wildcard appears in a [null-marked scope](https://jspecify.dev/docs/spec/#null-marked-scope), then it has a single upper bound whose [base type](https://jspecify.dev/docs/spec/#base-type) is Object and whose [nullness operator](https://jspecify.dev/docs/spec/#nullness-operator) is UNION_NULL.

Therefore, I think using `List<?>` as method parameter and return type of `toList()` makes sense.

> Just wondering: have you considered `Iterable` rather than using a plain `List`?

The use case we're addressing here is constructing new `Arguments` instances from existing ones so I don't think that flexibility is needed. Moreover, we could introduce it later, if necessary.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -125,6 +129,80 @@ static ArgumentSet argumentSet(String name, @Nullable Object... arguments) {
 		return new ArgumentSet(name, arguments);
 	}
 
+	/**
+	 * Factory method for creating an instance of {@code Arguments} based on
+	 * the supplied {@link List} of {@code arguments}.
+	 *
+	 * @param arguments the arguments to be used for an invocation of the test
+	 * method; must not be {@code null} but may contain {@code null}.
+	 * @return an instance of {@code Arguments}; never {@code null}.
```
vlsi: Is It really worth documenting `never {@code null}.`? I'm afraid it adds excessive verbosity.

I think the best practice is `non-null` by default, so WDYT of documenting only the places where nulls can appear rather than trying to document every possible case?

```suggestion
	 * @param arguments the arguments to be used for an invocation of the test
	 * method; it may contain {@code null} elements.
	 * @return an instance of {@code Arguments}
```

vlsi:
@marcphilipp , sorry for asking the same thing twice, however, could you please clarify why do you mean `Arguments.from(List)`, `Arguments argumentsFrom(List)`, and `ArgumentSet argumentSetFrom(String name, List)` relate to "from existing ones"?

To me the second case of #4535 sounds like a natural case for using `argumentsFrom(Iterable<?>)` rather than a pure List.

> 2. An Arguments can only be created from an array, not a collection.

I've no issues if `Arguments.toList` returns `List`. However, I believe on the consumption side, `Arguments.from` should accept `Iterable` so the users could supply various types of collections they have.

Of course, it would be great if Java had something like `IterableWithPredefinedIterationOrder`, so users don't accidentally pass `HashSet` and get their arguments shuffled.

----

If you mean you force `List` so the users do not supply collections like `Stream...parallel()...` and somehow complain "the order of arguments is wrong sometimes"? If that is the case, it is probably worth documenting as "this `List<?>` could have been `Iterable<?>`" is a typical code review case.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
marcphilipp: It may be excessive but clear. Moreover, we'd have to change that throughout the entire code base.

marcphilipp:
> sorry for asking the same thing twice, however, could you please clarify why do you mean `Arguments.from(List)`, `Arguments argumentsFrom(List)`, and `ArgumentSet argumentSetFrom(String name, List)` relate to "from existing ones"?

The use case we're trying to address here is to go from an existing `Arguments` instance to a new one without re-inventing a new collection API.

> I've no issues if `Arguments.toList` returns `List`. However, I believe on the consumption side, `Arguments.from` should accept `Iterable` so the users could supply various types of collections they have.

We'll discuss that in one of our next team calls.

marcphilipp:
**Team decision:** Use `Iterable<?>` as parameter types and documentation that the iteration order should be deterministic since it will be converted to an array.

mpkorstanje:
@mariokhoury4 thanks for working on this! Unfortunately we took a long time before we could discuss this and imagine you'll have moved on. So I'll pick up the remaining changes and do some polishing to wrap this up in the next few weeks.

mpkorstanje:
> Therefore, I think using List<?> as method parameter and return type of toList() makes sense.

Using wildcards in the method parameter is fine, but for the return type of `toList()`, `List<@Nullable Object>` is preferable.  `List<?>`  can't be passed to function that expect a `List<? extends T>`.

mpkorstanje:
Note: Manually set DTO check to pass. PR was offered before DTO bot was enabled.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/Arguments.java
```diff
@@ -94,45 +149,118 @@ static Arguments of(@Nullable Object... arguments) {
...
+	static ArgumentSet argumentSetFrom(String name, Iterable<?> arguments) {
+		Preconditions.notBlank(name, "name must not be null or blank");
+		Preconditions.notNull(arguments, "arguments list must not be null");
+
+		if (arguments instanceof Collection<?> collection) {
+			return argumentSet(name, collection.toArray());
+		}
+
+		var collection = new ArrayList<>();
+		arguments.forEach(collection::add);
```
marcphilipp, line 246: nit: we could introduce a `private @Nullable Object[] toArray(Iterable<?>)` method to avoid the duplication with `from`.

PR #4598, opened 2025-05-29
Remove short-circuiting to ensure all classes are checked for cycles

### LINKED ISSUE #4597: `TestClassPredicates.looksLikeIntendedTestClass()` does not consistently detect nested class hierarchy cycles
## Overview

`TestClassPredicatesTests.StandaloneTestClasses.recursiveHierarchies()` currently passes when run with the Gradle build or in IntelliJ IDEA; however, it fails in Eclipse IDE.

Specifically, the assertion on the first line of that test method does not throw an exception in Eclipse.

https://github.com/junit-team/junit5/blob/3c5b457564d559fb0f396d7f303f8effe291daa3/jupiter-tests/src/test/java/org/junit/jupiter/engine/discovery/predicates/TestClassPredicatesTests.java#L221-L224

The reason is twofold:

1. `ReflectionUtils.visitNestedClasses(...)` short-circuits the search algorithm once a `@Nested` test class has been detected.
2. The order in which nested classes are returned from `java.lang.Class.getDeclaredClasses()` apparently differs between the OpenJDK/IntelliJ compilers and the Eclipse compiler.

The result is that `ReflectionUtils.detectInnerClassCycle()` only gets invoked on nested classes until the first `@Nested` class is encountered (or until there are no more nested classes to inspect).

Thus, if a `@Nested` class is detected **before** a nested class with a class hierarchy cycle, the cycle is not detected/reported.

In Eclipse, if you rename `TestClassPredicatesTests.TestCases.OuterClass.InnerClass` to `TestClassPredicatesTests.TestCases.OuterClass.XXXInnerClass`, the test then passes because `TestClassPredicatesTests.TestCases.OuterClass.RecursiveInnerClass` is encountered first, resulting in the expected `JUnitException` stating that it "detected [a] cycle in [the] inner class hierarchy".

## Related Issues

- #242

marcphilipp:
## Overview

Resolves #4597.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
@@ -49,6 +49,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.regex.Matcher;
@@ -1142,17 +1143,9 @@ public final class ReflectionUtils {
 		Preconditions.notNull(clazz, "Class must not be null");
 		Preconditions.notNull(predicate, "Predicate must not be null");
 
-		class Tracker implements Consumer<Class<?>> {
-			boolean foundNestedClass = false;
-			@Override
-			public void accept(Class<?> __) {
-				foundNestedClass = true;
-			}
-		}
-		Tracker tracker = new Tracker();
-
-		visitAllNestedClasses(clazz, predicate, tracker);
-		return tracker.foundNestedClass;
+		AtomicBoolean foundNestedClass = new AtomicBoolean(false);
+		visitAllNestedClasses(clazz, predicate, __ -> foundNestedClass.setPlain(true));
+		return foundNestedClass.getPlain();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
```diff
@@ -1144,8 +1142,17 @@ public static boolean isNestedClassPresent(Class<?> clazz, Predicate<Class<?>> p
...
+			boolean foundNestedClass = false;
+			@Override
+			public void accept(Class<?> __) {
+				foundNestedClass = true;
+			}
+		}
+		Tracker tracker = new Tracker();
+
+		visitAllNestedClasses(clazz, predicate, tracker);
+		return tracker.foundNestedClass;
```
sbrannen, line 1155:
```suggestion
		AtomicBoolean found = new AtomicBoolean();
		visitAllNestedClasses(clazz, predicate, __ -> found.set(true));
		return found.get();
```

I think I would prefer that we use an `AtomicBoolean` here for simplicity.

sbrannen (changes requested):
Thanks for putting together this PR at the last minute!

I only requested one minor change.

sbrannen (approved):

junit-platform-commons/src/main/java/org/junit/platform/commons/util/ReflectionUtils.java
sbrannen: Ahhh... looks like our comments crossed paths in the ether. 😉

PR #4603, opened 2025-05-31
Ensure `@Nested` classes are executed after sibling test methods

### LINKED ISSUE #4600: `@Nested` tests are executed in inverted order when selected via classpath scanning
## Overview

In the test suite for the Spring Framework, we have a [`TestExecutionListenersNestedTests`](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/test/java/org/springframework/test/context/junit/jupiter/nested/TestExecutionListenersNestedTests.java) test class that is only intended to be run as a top-level test class. Specifically, if you run a `@Test` in a `@Nested` test class directly, that test will fail, because the `@Test` methods in the enclosing test classes have not been run. Although that is perhaps questionable behavior, it is actually by design: this particular test fixture tests the interactions between JUnit Jupiter and the Spring TestContext Framework.

We also have [`SpringJUnitJupiterTestSuite`](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/test/java/org/springframework/test/context/junit/SpringJUnitJupiterTestSuite.java), which selects `TestExecutionListenersNestedTests` via `@SelectPackages("org.springframework.test.context.junit.jupiter")` and `@IncludeClassNamePatterns(".*Tests$")`.

Historically `SpringJUnitJupiterTestSuite` succeeded (with JUnit Jupiter versions prior to 5.13), but it now fails.

Since that setup is a bit complex, I have reduced this scenario to just use the following JUnit Jupiter extensions.

```java
abstract class BaseExtension implements BeforeAllCallback {

	@Override
	public void beforeAll(ExtensionContext context) throws Exception {
		ExecutionOrderForNestedTests.extensions.add(getClass().getSimpleName());
	}
}

class Foo extends BaseExtension {
}

class Bar extends BaseExtension {
}

class Baz extends BaseExtension {
}

class Qux extends BaseExtension {
}
```

And the simplified test class looks like this.

```java
package example;

// imports...

@ExtendWith(Foo.class)
class ExecutionOrderForNestedTests {

	static final String FOO = "Foo";
	static final String BAR = "Bar";
	static final String BAZ = "Baz";
	static final String QUX = "Qux";

	static final List<String> extensions = new ArrayList<>();

	@AfterEach
	void resetExtensions() {
		extensions.clear();
	}

	@Test
	void test() {
		assertThat(extensions).containsExactly(FOO);
	}

	@Nested
	class InheritedExtensionsTests {

		@Test
		void test() {
			assertThat(extensions).containsExactly(FOO);
		}
	}

	@Nested
	@ExtendWith(Bar.class)
	class InheritedAndLocalExtensionsTests {

		@Test
		void test() {
			assertThat(extensions).containsExactly(FOO, BAR);
		}

		@Nested
		@ExtendWith(Baz.class)
		class DoubleNestedWithInheritedAndLocalExtensionsTests {

			@Test
			void test() {
				assertThat(extensions).containsExactly(FOO, BAR, BAZ);
			}

			@Nested
			@ExtendWith(Qux.class)
			class TripleNestedWithInheritedAndLocalExtensionsTests {

				@Test
				void test() {
					assertThat(extensions).containsExactly(FOO, BAR, BAZ, QUX);
				}
			}
		}
	}
}
```

If you run `ExecutionOrderForNestedTests` by itself, all of the tests (including the tests in the `@Nested` test classes) pass.

As I mentioned above, if you run an individual `@Test` in a `@Nested` test class directly, that will fail, but this is by design. So, that is not a regression.

## The Regression

The regression is apparent only when running `ExecutionOrderForNestedTests` via a `@Suite`.

✅ This suite passes:

```java
@Suite
@SelectPackages("example")
@IncludeClassNamePatterns(".+ExecutionOrderForNestedTests$")
class TestSuite {
}
```

✅ This suite also passes:

```java
@Suite
@SelectClasses(ExecutionOrderForNestedTests.class)
class TestSuite {
}
```

❌ This suite fails:

```java
@Suite
@SelectPackages("example")
class TestSuite {
}
```

❌ This suite also fails:

```java
@Suite
@SelectPackages("example")
@IncludeClassNamePatterns(".+Tests$")
class TestSuite {
}
```

For example, `TripleNestedWithInheritedAndLocalExtensionsTests.test()` fails as follows when run in one of those failing `@Suite` scenarios.

```
org.opentest4j.AssertionFailedError: 
Expecting actual:
  ["Foo", "Bar", "Foo", "Bar", "Baz", "Foo", "Bar", "Baz", "Qux"]
to contain exactly (and in same order):
  ["Foo", "Bar", "Baz", "Qux"]
but some elements were not expected:
  ["Foo", "Bar", "Foo", "Bar", "Baz"]
```

## Analysis

If `ExecutionOrderForNestedTests` is _selected_ explicitly in the IDE or in a `@Suite` (without discovering the individual `@Nested` test classes in `ExecutionOrderForNestedTests`), all tests pass.

If the `@Nested` test classes in `ExecutionOrderForNestedTests` are discovered in a `@Suite` using classpath scanning via `@SelectPackages`, the `@Nested` test methods get executed **before** (or "independently of") the test methods in enclosing classes, which was not previously the case.

marcphilipp:
## Overview

In order to validate them, classpath scanning was changed to find them
in 5.13.0. However, that caused `@Nested` classes to be added to their
parent descriptors before their sibling test methods. This made the
order of execution in classes containing nested test classes dependent
on how they were discovered unless a `MethodOrderer` was configured in
which case `MethodOrderingVisitor` ensured test methods came before
`@Nested` test classes on the same level.

This commit changes `MethodOrderingVisitor` to also ensure this
ordering constraint in case no `MethodOrderer` is configured.

Resolves #4600.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-5.13.1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-5.13.1.adoc
@@ -35,7 +35,13 @@ on GitHub.
 [[release-notes-5.13.1-junit-jupiter-bug-fixes]]
 ==== Bug Fixes
 
-* ❓
+* The 5.13.0 release introduced a regression regarding the execution order in test classes
+  containing both test methods and `@Nested` test classes. When classpath scanning was
+  used during test discovery -- for example, when resolving a package selector for a
+  `@Suite` class -- test methods in `@Nested` classes were executed _before_ test methods
+  in their enclosing class. This undesired change in behavior has now been reverted so
+  that tests in `@Nested` test classes are always executed _after_ tests in enclosing test
+  classes again.
 
 [[release-notes-5.13.1-junit-jupiter-deprecations-and-breaking-changes]]
 ==== Deprecations and Breaking Changes
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodOrderingVisitor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodOrderingVisitor.java
@@ -10,12 +10,14 @@
 
 package org.junit.jupiter.engine.discovery;
 
+import static java.util.Comparator.comparing;
 import static org.junit.platform.commons.support.AnnotationSupport.findAnnotation;
 import static org.junit.platform.commons.support.AnnotationSupport.isAnnotated;
 
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Consumer;
+import java.util.function.UnaryOperator;
 
 import org.junit.jupiter.api.MethodOrderer;
 import org.junit.jupiter.api.Order;
@@ -40,6 +42,9 @@ class MethodOrderingVisitor extends AbstractOrderingVisitor {
 	private final JupiterConfiguration configuration;
 	private final Condition<MethodBasedTestDescriptor> noOrderAnnotation;
 
+	// Not a static field to avoid initialization at build time for GraalVM
+	private final UnaryOperator<List<TestDescriptor>> methodsBeforeNestedClassesOrderer;
+
 	MethodOrderingVisitor(JupiterConfiguration configuration, DiscoveryIssueReporter issueReporter) {
 		super(issueReporter);
 		this.configuration = configuration;
@@ -51,6 +56,7 @@ class MethodOrderingVisitor extends AbstractOrderingVisitor {
 						.source(MethodSource.from(testDescriptor.getTestMethod())) //
 						.build();
 			});
+		this.methodsBeforeNestedClassesOrderer = createMethodsBeforeNestedClassesOrderer();
 	}
 
 	@Override
@@ -80,6 +86,7 @@ class MethodOrderingVisitor extends AbstractOrderingVisitor {
 
 	private void orderContainedMethods(ClassBasedTestDescriptor classBasedTestDescriptor, Class<?> testClass,
 			Optional<MethodOrderer> methodOrderer) {
+
 		DescriptorWrapperOrderer<?, DefaultMethodDescriptor> descriptorWrapperOrderer = createDescriptorWrapperOrderer(
 			testClass, methodOrderer);
 
@@ -88,6 +95,11 @@ class MethodOrderingVisitor extends AbstractOrderingVisitor {
 			toValidationAction(methodOrderer), //
 			DefaultMethodDescriptor::new, //
 			descriptorWrapperOrderer);
+
+		if (methodOrderer.isEmpty()) {
+			// If there is an orderer, this is ensured by the call above
+			classBasedTestDescriptor.orderChildren(methodsBeforeNestedClassesOrderer);
+		}
 
 		// Note: MethodOrderer#getDefaultExecutionMode() is guaranteed
 		// to be invoked after MethodOrderer#orderMethods().
@@ -124,6 +136,14 @@ class MethodOrderingVisitor extends AbstractOrderingVisitor {
 			return Optional.empty();
 		}
 		return Optional.of(noOrderAnnotation::check);
+	}
+
+	private static UnaryOperator<List<TestDescriptor>> createMethodsBeforeNestedClassesOrderer() {
+		var methodsFirst = comparing(MethodBasedTestDescriptor.class::isInstance).reversed();
+		return children -> {
+			children.sort(methodsFirst);
+			return children;
+		};
 	}
 
 }
--- a/jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
+++ b/jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
@@ -71,7 +71,7 @@ class NestedTestClassesTests extends AbstractJupiterTestEngineTests {
 		assertEquals(2, tests.succeeded().count(), "# tests succeeded");
 		assertEquals(1, tests.failed().count(), "# tests failed");
 
-		assertThat(tests.started().map(it -> it.getTestDescriptor().getDisplayName()).toList()) //
+		assertThat(tests.started().map(it -> it.getTestDescriptor().getDisplayName())) //
 				.containsExactlyInAnyOrder("someTest()", "successful()", "failing()") //
 				.containsSubsequence("someTest()", "successful()") //
 				.containsSubsequence("someTest()", "failing()");
--- a/jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
+++ b/jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
@@ -30,12 +30,9 @@ import java.util.regex.Pattern;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.MethodOrderer.OrderAnnotation;
 import org.junit.jupiter.api.Named;
 import org.junit.jupiter.api.Nested;
-import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestMethodOrder;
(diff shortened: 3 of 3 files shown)


documentation/src/docs/asciidoc/release-notes/release-notes-5.13.1.adoc
```diff
@@ -35,7 +35,13 @@ on GitHub.
...
 ==== Bug Fixes
 
-* ❓
+* The 5.13.0 release introduced a regression regarding the execution order in test classes
+  containing both test methods and `@Nested` test classes. When classpath scanning was
+  used during test discovery -- for example, when resolving a package selector for a
+  `@Suite` class -- test methods in `@Nested` classes were executed _before_ test methods
+  in their enclosing class. This undesired change in behavior has now been reverted so
+  that tests in `@Nested` test classes are always executed _after_ tests in enclosing test
+  classes again.
```
sbrannen, line 44: That very accurately describes the scenario and fix. 👍

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodOrderingVisitor.java
```diff
@@ -40,6 +42,9 @@ class MethodOrderingVisitor extends AbstractOrderingVisitor {
 	private final JupiterConfiguration configuration;
 	private final Condition<MethodBasedTestDescriptor> noOrderAnnotation;
 
+	// Not a constant to avoid initialization at build time for GraalVM
```
sbrannen:
```suggestion
	// Not a static field to avoid initialization at build time for GraalVM
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodOrderingVisitor.java
```diff
@@ -89,6 +96,11 @@ private void orderContainedMethods(ClassBasedTestDescriptor classBasedTestDescri
 			DefaultMethodDescriptor::new, //
 			descriptorWrapperOrderer);
 
+		if (methodOrderer.isEmpty()) {
+			// If there is an orderer, this is ensured by the call above
+			classBasedTestDescriptor.orderChildren(methodsBeforeNestedClassesOrderer);
+		}
```
sbrannen, line 102: I understand how this works; however, it seems to me that this should ideally be an invariant.

Playing the role of _Devil's Advocate_: why do we need to implement it like this? Is the need to do so indicative of some larger issue that should be addressed elsewhere in the discovery or post-discovery algorithms?

I am not familiar with all of the necessary changes you made to support validation and issue reporting, so perhaps this is the best way to achieve this in 5.13, and perhaps we should revisit this in 6.0.

Thoughts?

jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
```diff
@@ -53,20 +61,36 @@ void nestedTestsAreCorrectlyDiscovered() {
 		assertEquals(5, engineDescriptor.getDescendants().size(), "# resolved test descriptors");
 	}
 
-	@Test
-	void nestedTestsAreExecuted() {
-		EngineExecutionResults executionResults = executeTestsForClass(TestCaseWithNesting.class);
-		Events containers = executionResults.containerEvents();
-		Events tests = executionResults.testEvents();
+	@ParameterizedTest(name = "{0}")
+	@MethodSource
```
sbrannen, line 65: Could be a `@FieldSource`. 😉 

(but a `@MethodSource` is also fine)

jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
```diff
@@ -53,20 +61,36 @@ void nestedTestsAreCorrectlyDiscovered() {
 		assertEquals(5, engineDescriptor.getDescendants().size(), "# resolved test descriptors");
 	}
 
-	@Test
-	void nestedTestsAreExecuted() {
-		EngineExecutionResults executionResults = executeTestsForClass(TestCaseWithNesting.class);
-		Events containers = executionResults.containerEvents();
-		Events tests = executionResults.testEvents();
+	@ParameterizedTest(name = "{0}")
```
sbrannen, line 64: Wouldn't `ArgumentSet` be easier?

jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
```diff
@@ -53,20 +61,36 @@ void nestedTestsAreCorrectlyDiscovered() {
...
+				.containsExactlyInAnyOrder("someTest()", "successful()", "failing()") //
+				.containsSubsequence("someTest()", "successful()") //
+				.containsSubsequence("someTest()", "failing()");
+
+		Events containers = executionResults.containerEvents();
 		assertEquals(3, containers.started().count(), "# containers started");
 		assertEquals(3, containers.finished().count(), "# containers finished");
 	}
 
+	static List<Named<Consumer<LauncherDiscoveryRequestBuilder>>> nestedTestsAreExecutedInTheRightOrder() {
```
sbrannen, line 84: `Consumer<LauncherDiscoveryRequestBuilder>` -- clever. 👍

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodOrderingVisitor.java
marcphilipp: Albeit less likely, there are ways to make this happen even in 5.12 and earlier, e.g. using a `ClassSelector` for the nested class and one for its enclosing class (in that order). Yes, this is an artifact of how discovery works but this visitor is part of the "post-discovery" algorithm. The only alternative I thought about would be to override `addChild(TestDescriptor)` in `ClassBasedTestDescriptor` and ensure they don't get added in the wrong way. But, since we're applying orderers here anyway, I think that's the cleaner solution. Happy to discuss on a call, though.

jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
marcphilipp: I wouldn't be able to statically type it then, or am I missing something? While `List<Named<Consumer<LauncherDiscoveryRequestBuilder>>>` is a mouthful, I think it's better than `List<ArgumentSet>` without any compile-time checks. WDYT?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodOrderingVisitor.java
marcphilipp: Change in the latest push

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodOrderingVisitor.java
sbrannen: Yep, I'm aware of the fact that this could surface in other ways.

So, let's go with what you've got for 5.13.x.

And we can discuss future options on a team call.

jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
sbrannen: We don't actually get any compile-time checks between the factory method's return type and the formal parameter list for the parameterized test method.

So, It think it's a toss-up.

I was merely pontificating. 😉 

So, just leave it as-is.

sbrannen (approved):

jupiter-tests/src/test/java/org/junit/jupiter/engine/NestedTestClassesTests.java
marcphilipp: Sure, but we do get compile-time checks that both arguments have the same type within the method providing the arguments.

Issue #4339, opened 2025-02-25
Replace univocity-parsers with FastCSV


marcphilipp:
Unfortunately, univocity-parsers is [no longer maintained](https://github.com/uniVocity/univocity-parsers/issues/534). While there's is a [fork](https://github.com/sonofab1rd/univocity-parsers), it's future is unclear.

[FastCSV](https://fastcsv.org/) has seen a lot of activity in recent months so we should evaluate if we use it as a replacement. There might be subtle differences how edge cases are handled but we should have good coverage for those in our test suite.

## Deliverables

- [ ] Implement `@CsvSource` and `@CsvFileSource` using FastCSV instead of univocity-parsers

vdmitrienko:
@marcphilipp can I take a stab?

marcphilipp:
@vdmitrienko Go for it! 🙂

marcphilipp:
@vdmitrienko Thanks for taking a closer look! Agreed, let's hold off on this until we start working on 6.0 later this year.

osiegmar:
I suggest analyzing if any functionality is missing in FastCSV that JUnit requires. This would give me time to design and implement it, ensuring a smooth integration when JUnit 6 development begins.

vdmitrienko:
> I suggest analyzing if any functionality is missing in FastCSV that JUnit requires. This would give me time to design and implement it, ensuring a smooth integration when JUnit 6 development begins.

@marcphilipp, I could give it a try with the [JDK17](https://github.com/junit-team/junit5/tree/marc/jdk17) branch. WDYT?

sormuras:
Also an option: make it work with FastCSV 2.x and Jupiter 5.x on Java 8 first.

marcphilipp:
> I could give it a try with the [JDK17](https://github.com/junit-team/junit5/tree/marc/jdk17?rgh-link-date=2025-02-27T10%3A51%3A02.000Z) branch. WDYT?

Good idea! But be prepared that we'll force push that one again at some point.

> Also an option: make it work with FastCSV 2.x and Jupiter 5.x on Java 8 first.

I don't think that would be worth the effort but I leave it to @vdmitrienko to decide.

vdmitrienko:
> Also an option: make it work with FastCSV 2.x and Jupiter 5.x on Java 8 first.

[...]

PR #4606, opened 2025-06-01
Replace univocity-parsers with FastCSV

vdmitrienko:
## Overview

https://github.com/junit-team/junit5/issues/4339

---

---

### Definition of Done

- [x] Change is documented in [Release Notes](https://junit.org/junit5/docs/snapshot/user-guide/#release-notes)


gradle/libs.versions.toml
```diff
@@ -71,7 +71,7 @@ slf4j-julBinding = { module = "org.slf4j:slf4j-jdk14", version = "2.0.17" }
 snapshotTests-junit5 = { module = "de.skuzzle.test:snapshot-tests-junit5", version.ref = "snapshotTests" }
 snapshotTests-xml = { module = "de.skuzzle.test:snapshot-tests-xml", version.ref = "snapshotTests" }
 spock1 = { module = "org.spockframework:spock-core", version = "1.3-groovy-2.5" }
-univocity-parsers = { module = "com.sonofab1rd:univocity-parsers", version = "2.10.2" }
+fastcsv = { module = "de.siegmar:fastcsv", version = "4.0.0-SNAPSHOT" }
```
marcphilipp: Please order these alphabetically so it's easier to find a certain dependency

junit-jupiter-params/junit-jupiter-params.gradle.kts
```diff
@@ -20,7 +20,7 @@ dependencies {
 	compileOnlyApi(libs.apiguardian)
 	compileOnly(libs.jspecify)
 
-	shadowed(libs.univocity.parsers)
+	implementation(libs.fastcsv)
```
marcphilipp: We should shadow FastCSV to avoid classpath conflicts in case user code uses a different version of FastCSV.

junit-jupiter-params/junit-jupiter-params.gradle.kts
```diff
@@ -29,7 +29,6 @@ dependencies {
 }
 
 extraJavaModuleInfo {
-	automaticModule(libs.univocity.parsers, "univocity.parsers")
```
marcphilipp, line 32: Looks like we can probably remove the `extraJavaModuleInfo` module plugin from this subproject and `libs.versions.toml`. Could you please give that a try?

junit-jupiter-params/junit-jupiter-params.gradle.kts
```diff
@@ -45,17 +44,10 @@ tasks {
...
-			include("LICENSE-univocity-parsers.md")
-			into("META-INF")
-		}
-	}
 	compileJava {
 		options.compilerArgs.addAll(listOf(
-			"--add-modules", "univocity.parsers",
-			"--add-reads", "${javaModuleName}=univocity.parsers"
+			"--add-modules", "de.siegmar.fastcsv",
+			"--add-reads", "${javaModuleName}=de.siegmar.fastcsv"
```
marcphilipp, line 57: This should only be necessary if we do shadow it (as requested above).

junit-jupiter-params/junit-jupiter-params.gradle.kts
```diff
@@ -45,17 +44,10 @@ tasks {
 			""")
 		}
 	}
-	shadowJar {
-		relocate("com.univocity", "org.junit.jupiter.params.shadow.com.univocity")
-		from(projectDir) {
-			include("LICENSE-univocity-parsers.md")
```
marcphilipp, line 51: We need to include FastCSV's license instead.

junit-platform-console-standalone/junit-platform-console-standalone.gradle.kts
```diff
@@ -44,10 +44,6 @@ tasks {
 			include("LICENSE-picocli.md")
 			into("META-INF")
 		}
-		from(dependencyProject(project.projects.junitJupiterParams).projectDir) {
-			include("LICENSE-univocity-parsers.md")
```
marcphilipp, line 48: Ditto

jupiter-tests/src/test/java/org/junit/jupiter/params/provider/CsvArgumentsProviderTests.java
```diff
@@ -275,7 +275,7 @@ void throwsExceptionIfSourceExceedsMaxCharsPerColumnConfig() {
 		assertThatExceptionOfType(CsvParsingException.class)//
 				.isThrownBy(() -> provideArguments(annotation).findAny())//
 				.withMessageStartingWith("Failed to parse CSV input configured via Mock for CsvSource")//
-				.withRootCauseInstanceOf(ArrayIndexOutOfBoundsException.class);
+				.withRootCauseInstanceOf(de.siegmar.fastcsv.reader.CsvParseException.class);
```
marcphilipp: Please import the class

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
```diff
@@ -59,6 +59,8 @@ class ModularUserGuideTests {
 			  // Byte Buddy is used by AssertJ's soft assertions which are used by the Engine Test Kit
 			  requires net.bytebuddy;
 
+			  requires de.siegmar.fastcsv;
+
```
marcphilipp: This shouldn't be necessary anymore once it's shadowed.

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
```diff
@@ -94,6 +96,7 @@ private static List<String> compile(Path temp, Writer out, Writer err) throws Ex
 		ThirdPartyJars.copy(lib, "org.opentest4j.reporting", "open-test-reporting-tooling-spi");
 		ThirdPartyJars.copy(lib, "com.google.jimfs", "jimfs");
 		ThirdPartyJars.copy(lib, "com.google.guava", "guava");
+		ThirdPartyJars.copy(lib, "siegmar", "fastcsv");
```
marcphilipp: Ditto

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
```diff
@@ -135,6 +138,7 @@ private static void junit(Path temp, OutputFiles outputFiles) throws Exception {
 					temp.resolve("lib").toString() //
 				)) //
 				.addArguments("--add-modules", "documentation") //
+				.addArguments("--add-reads", "org.junit.jupiter.params=de.siegmar.fastcsv") //
```
marcphilipp: Ditto

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
```diff
@@ -170,6 +174,7 @@ void runTestsFromUserGuideWithinModularBoundaries(@TempDir Path temp,
 			"lib/apiguardian-api-.+\\.jar", //
 			"lib/assertj-core-.+\\.jar", //
 			"lib/byte-buddy-.+", //
+			"lib/fastcsv-.+\\.jar", //
```
marcphilipp: Ditto

platform-tooling-support-tests/platform-tooling-support-tests.gradle.kts
```diff
@@ -64,6 +64,7 @@ dependencies {
 	thirdPartyJars(libs.opentest4j)
 	thirdPartyJars(libs.openTestReporting.tooling.spi)
 	thirdPartyJars(libs.jimfs)
+	thirdPartyJars(libs.fastcsv)
```
marcphilipp: This shouldn't be necessary anymore once it's shadowed.

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
```diff
@@ -110,13 +110,25 @@ repository on GitHub.
 * The contracts for the `Executable` parameters of Kotlin-specific `assertTimeout`
   functions were changed from `callsInPlace(executable, EXACTLY_ONCE)` to
   `callsInPlace(executable, AT_MOST_ONCE)` which might result in compilation errors.
+* The `CsvFileSource.lineSeparator()` parameter is deprecated because line separators
+  are now detected automatically during CSV parsing. This setting is no longer required
+  and will be ignored.
```
marcphilipp: Does auto-detection work in all cases? What happens if `\n` is used in a cell like in the following example with 4 columns?

```
a;b;\n
c;d\r\n
e;f;g;h\r\n
```
(assuming `\n` and `\r` are replaced with the corresponding character)

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -37,108 +32,67 @@
...
+			for (int i = 0; i < csvSource.value().length; i++) {
+				if (csvSource.value()[i].isEmpty()) {
+					String message = String.format("CSV record at index %d is empty", i + 1);
+					throw new PreconditionViolationException(message);
 				}
-				Preconditions.notNull(csvRecord,
-					() -> "Record at index " + index + " contains invalid CSV: \"" + input + "\"");
-				argumentsList.add(processCsvRecord(csvRecord, nullValues, useHeadersInDisplayName, headers));
 			}
+			return String.join("\n", csvSource.value());
```
marcphilipp: Does FastCSV provide an API for line-by-line reading so we don't have to create a string first? It's probably not a big deal since it comes from literals in an annotation.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -37,108 +32,67 @@
...
+
+		if (!csvSource.textBlock().isEmpty()) {
+			return csvSource.textBlock();
+		}
+		else {
+			for (int i = 0; i < csvSource.value().length; i++) {
+				if (csvSource.value()[i].isEmpty()) {
+					String message = String.format("CSV record at index %d is empty", i + 1);
+					throw new PreconditionViolationException(message);
 				}
```
marcphilipp:
```suggestion
				Preconditions.condition(!csvSource.value()[i].isEmpty(), () -> "CSV record at index %d is empty".formatted(i + 1));
```

marcphilipp (commented):
This looks very promising! :+1:

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
vdmitrienko: > Does auto-detection work in all cases?

The auto-detection treats each of `\r`, `\n`, and `\r\n` as a line separator. For example, given the following input:
```
a;b\r
c;d\n
e;f\r\n
g;h
```
The result is:
```java
[["a", "b"], ["c", "d"], ["e", "f"], ["g", "h"]]
```
In contrast, _univocity-parsers_ (when configured with `\n` as the line separator) produces different results:
```java
["a", "b\rc", "d"], ["e", "f"], ["g", "h"]
```

vdmitrienko: > What happens if \n is used in a cell like in the following example with 4 columns?

In this case, the results from _FastCSV_ and _univocity-parsers_ are mostly similar.

_FastCSV:_
```java
[["a", "b", null], ["c", "d"], ["e", "f", "g", "h"]]
```
_univocity-parsers:_
```java
// .lineSeparator("\n") - same as FastCSV
[["a", "b", null], ["c", "d"], ["e", "f", "g", "h"]]

// .lineSeparator("\r\n") - same as FastCSV
[["a", "b", null], ["c", "d"], ["e", "f", "g", "h"]]

// .lineSeparator("\r") - differs from FastCSV
[["a", "b", "c", "d"], ["e", "f", "g", "h"], [null]]
```

osiegmar: I’m afraid this breaks compatibility if someone uses a character sequence as a line delimiter that is **not** a newline.

vdmitrienko: So, considering 3 possible scenarios, all of them imply a breaking change 😞 

1. User explicitly relies on `\r\n` as the line separator:
`\r` - causes an unexpected line break;
`\n` - causes an unexpected line break;

2. User explicitly relies on `\r` as the line separator:
`\n` - causes an unexpected line break;
`\r\n` - no change, since `\r` is already interpreted as a line break;

3. User explicitly relies on `\n` as the line separator:
`\r` - causes an unexpected line break;
`\r\n` - no change, since `\n` is already interpreted as a line break;

vdmitrienko: @osiegmar, would it be possible to add support for a `lineSeparator()` parameter in FastCSV?

osiegmar: Potentially, yes. Of course, this wouldn’t be a valid CSV file at all. Is this really a desired feature or just lack of specification/documentation and a good chance to change that with the new major version of JUnit?

Is there a (good) reason, someone separates text records by anything that is not a newline sequence?

Is there known usage of this?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
osiegmar: With https://github.com/osiegmar/FastCSV/commit/1077389279133ad3e63e2c5d0cd1ffa03bca74d5 there is one now. @vdmitrienko You may want to give it a try if it simplify things for you.

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
marcphilipp: I think dropping this in the new major version makes sense. I'm not aware of any cases other than using the same line separator on different operating systems. IIRC I initially introduced it because univocity-parsers would use the system line separator (and only that) be default.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
vdmitrienko: Thanks, @osiegmar. This works well with individual strings, but it doesn't support headers. I think adding an overload that accepts an array (or varargs) of strings could simplify this use case:
```java
build(final CsvCallbackHandler<T> callbackHandler, final String... data)
```

vdmitrienko: Regarding the validation of empty records, having a setting for that could be quite handy. It would also be great if the exception message included the index of the empty record. That said, we could also handle this on our side 🙂

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
vdmitrienko: > I think dropping this in the new major version makes sense.

Great 👍 
I'll make sure to clarify that in the release notes. Adding a few tests wouldn't hurt either.

gradle/libs.versions.toml
vdmitrienko: Done

junit-jupiter-params/junit-jupiter-params.gradle.kts
vdmitrienko: Done

junit-jupiter-params/junit-jupiter-params.gradle.kts
vdmitrienko: Indeed. Removed

junit-jupiter-params/junit-jupiter-params.gradle.kts
vdmitrienko: Left as is, since FastCSV is now shadowed.

junit-jupiter-params/junit-jupiter-params.gradle.kts
vdmitrienko: Included to the shadowed JAR

junit-platform-console-standalone/junit-platform-console-standalone.gradle.kts
vdmitrienko: Done

jupiter-tests/src/test/java/org/junit/jupiter/params/provider/CsvArgumentsProviderTests.java
vdmitrienko: Done

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
vdmitrienko: Removed

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
vdmitrienko: Removed

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
vdmitrienko: Removed

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
vdmitrienko: Removed

platform-tooling-support-tests/platform-tooling-support-tests.gradle.kts
vdmitrienko: Removed

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
vdmitrienko: Applied

jupiter-tests/jupiter-tests.gradle.kts
```diff
@@ -24,6 +24,7 @@ dependencies {
 	testImplementation(libs.kotlinx.coroutines)
 	testImplementation(libs.groovy4)
 	testImplementation(libs.memoryfilesystem)
+	testImplementation(libs.fastcsv)
```
osiegmar: still necessary?

junit-jupiter-params/LICENSE-fastcsv.md
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Oliver Siegmar
```
osiegmar: Just noticed that this hasn't been updated for a while. Now, I did.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderFactory.java
```diff
@@ -0,0 +1,168 @@
...
+
+	private static void validateDelimiter(char delimiter, String delimiterString, Annotation annotation) {
+		Preconditions.condition(delimiter == EMPTY_CHAR || delimiterString.isEmpty(),
+			() -> "The delimiter and delimiterString attributes cannot be set simultaneously in " + annotation);
+	}
+
+	static CsvReader<? extends CsvRecord> createReaderFor(CsvSource csvSource, String data) {
+		String delimiter = selectDelimiter(csvSource.delimiter(), csvSource.delimiterString());
+		// @formatter:off
+		CsvReader.CsvReaderBuilder builder = CsvReader.builder()
```
osiegmar: As you rely on `skipEmptyLines(true)` (current default), you may want to set that explicitly.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderFactory.java
```diff
@@ -0,0 +1,168 @@
...
+		// @formatter:on
+		return builder.build(callbackHandler, data);
+	}
+
+	static CsvReader<? extends CsvRecord> createReaderFor(CsvFileSource csvFileSource, InputStream inputStream,
+			Charset charset) {
+
+		String delimiter = selectDelimiter(csvFileSource.delimiter(), csvFileSource.delimiterString());
+		// @formatter:off
+		CsvReader.CsvReaderBuilder builder = CsvReader.builder()
```
osiegmar: As you rely on `skipEmptyLines(true)` (current default), you may want to set that explicitly.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderFactory.java
```diff
@@ -0,0 +1,168 @@
...
+	private static CsvCallbackHandler<? extends CsvRecord> createCallbackHandler(String emptyValue,
+			Set<String> nullValues, boolean ignoreLeadingAndTrailingWhitespaces, int maxCharsPerColumn,
+			boolean useHeadersInDisplayName) {
+
+		int maxFieldSize = maxCharsPerColumn == -1 ? Integer.MAX_VALUE : maxCharsPerColumn;
+		FieldModifier modifier = new DefaultFieldModifier(emptyValue, nullValues, ignoreLeadingAndTrailingWhitespaces);
+
+		// @formatter:off
+		if (useHeadersInDisplayName) {
+			return NamedCsvRecordHandler.builder()
```
osiegmar, line 133: Current implementation allows duplicate header names. For compatibility, add `allowDuplicateHeaderFields(true)`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -37,108 +32,69 @@
...
-		Object[] arguments = new Object[csvRecord.length];
-		for (int i = 0; i < csvRecord.length; i++) {
-			Object column = csvRecord[i];
-			if (column != null && nullValues.contains(column)) {
-				column = null;
-			}
+		Object[] arguments = new Object[record.getFields().size()];
+
+		for (int i = 0; i < record.getFields().size(); i++) {
+			String field = record.getFields().get(i);
```
osiegmar: You should either call `getFields()` only once (as it constructs new objects internally) or (preferably) use the methods `getFieldCount()` and `getField(int)`.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -147,6 +103,10 @@ static Arguments processCsvRecord(@Nullable String[] csvRecord, Set<String> null
 		return Named.of(name, column);
 	}
 
+	private static List<String> getHeaders(CsvRecord record) {
+		return ((NamedCsvRecord) record).getHeader();
```
osiegmar, line 103: Previously, the header field names were trimmed.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
osiegmar: Got it. In that case, I'd rather stick to the current `String.join()` implementation. The effort for this extra build-method (bigger than initially thought) doesn't seem to match this edge-case use. I also highly doubt that it would improve performance.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
osiegmar: I also just noticed, that empty header fields (like in `foo,,bar`) are causing NullPointerExceptions with the univocity implementation. I also noticed that while the emptyValue is applied to header fields, nullValues are not.

The new implementation currently uses the fieldModifier also for the header record and would produce unexpected results (NULL_MARKER, nullValues, emptyValue).

junit-jupiter-params/LICENSE-fastcsv.md
marcphilipp: Is this part of FastCSV's jar file? If so, we should probably extract it from there to avoid it becoming outdated.

osiegmar: Just added `META-INF/LICENSE`

marcphilipp: @vdmitrienko Please let me know if you need help with Gradle to achieve that! :slightly_smiling_face:

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
vdmitrienko: Thanks! Updated

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderFactory.java
vdmitrienko: Good point. Updated

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderFactory.java
vdmitrienko: Updated

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderFactory.java
vdmitrienko: Updated

jupiter-tests/jupiter-tests.gradle.kts
vdmitrienko: Since the library is now shadowed, this is no longer necessary 👍

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
vdmitrienko: > Previously, the header field names were trimmed.

Currently, they are still trimmed because header fields are treated as regular fields, meaning that `ignoreLeadingAndTrailingWhitespace()` (default `true`) applies to them as well.

> The new implementation currently uses the fieldModifier also for the header record and would produce unexpected results (NULL_MARKER, nullValues, emptyValue).

Thanks for the observation! Now it applies to headers consistently as well.

I believe treating headers as regular fields would be the simplest and most reliable solution, this way, attributes like `ignoreLeadingAndTrailingWhitespace()`, `nullValues()`, etc. apply uniformly to both headers and records.

@marcphilipp @osiegmar WDYT?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -37,120 +32,85 @@
...
+				if (csvSource.value()[i].isEmpty()) {
+					int finalI = i;
+					Preconditions.condition(!csvSource.value()[i].isEmpty(), //
+						() -> "CSV record at index %d is empty".formatted(finalI + 1) //
+					);
 				}
-				Preconditions.notNull(csvRecord,
-					() -> "Record at index " + index + " contains invalid CSV: \"" + input + "\"");
-				argumentsList.add(processCsvRecord(csvRecord, nullValues, useHeadersInDisplayName, headers));
 			}
```
osiegmar: I'm curious: why are empty strings in the value array handled as an error? In text blocks empty lines are simply skipped. I also couldn't find something in the javadoc about it. Seems like an unnecessary difference.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
osiegmar: > > Previously, the header field names were trimmed.
> 
> Currently, they are still trimmed because header fields are treated as regular fields, meaning that `ignoreLeadingAndTrailingWhitespace()` (default `true`) applies to them as well.

Regardless of the setting `ignoreLeadingAndTrailingWhitespace` and regardless of whether quoted or not, headers were always trimmed:

https://github.com/junit-team/junit5/blob/b580aa90923f44aaa412ae32ba0915b06f96c697/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java#L107-L112

> I believe treating headers as regular fields would be the simplest and most reliable solution, this way, attributes like `ignoreLeadingAndTrailingWhitespace()`, `nullValues()`, etc. apply uniformly to both headers and records.
> 
> @marcphilipp @osiegmar WDYT?

I believe the same, just wanted to point out that difference.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
```diff
@@ -37,120 +32,85 @@
...
-				column = null;
-			}
+		Object[] arguments = new Object[record.getFieldCount()];
+
+		for (int i = 0; i < record.getFieldCount(); i++) {
+			String field = record.getField(i);
+			Object argument = resolveNullMarker(field);
 			if (useHeadersInDisplayName) {
-				column = asNamed(requireNonNull(headers)[i] + " = " + column, column);
+				String header = resolveNullMarker(getHeaders(record).get(i));
```
osiegmar: Didn't see this before: `getHeaders` (which calls `getHeader` on `NamedCsvRecord`) should also better not called in a loop - it creates objects.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
vdmitrienko: This behavior originates from [univocity parseLine() implementation](https://github.com/uniVocity/univocity-parsers/blob/master/src/main/java/com/univocity/parsers/common/AbstractParser.java#L666), which we use to parse individual lines: it returns `null` in case a single line is empty. For text blocks, however, we use a different method: [parseAll()](https://github.com/uniVocity/univocity-parsers/blob/master/src/main/java/com/univocity/parsers/common/AbstractParser.java#L533), which simply discards empty lines.
I preserved this behavior for the compatibility sake, but seems like we could actually drop it and unify the handling of value arrays and text blocks.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
vdmitrienko: Missed that. Updated

junit-jupiter-params/LICENSE-fastcsv.md
vdmitrienko: @marcphilipp, please, take a look at https://github.com/junit-team/junit5/pull/4606/commits/d5139e2129d8df0d1dff72f8d3ea3cba83dd36fc

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderFactory.java
```diff
@@ -0,0 +1,177 @@
...
+
+	static void validate(CsvSource csvSource) {
+		validateMaxCharsPerColumn(csvSource.maxCharsPerColumn());
+		validateDelimiter(csvSource.delimiter(), csvSource.delimiterString(), csvSource);
+	}
+
+	static void validate(CsvFileSource csvFileSource) {
+		validateMaxCharsPerColumn(csvFileSource.maxCharsPerColumn());
+		validateDelimiter(csvFileSource.delimiter(), csvFileSource.delimiterString(), csvFileSource);
+	}
```
osiegmar, line 54: A validation should be added to ensure that the (now deprecated) lineSeparator contains only \r\n, \r, or \n.

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
vdmitrienko: Just to clarify, do we intend to remove `lineSeparator()` entirely, or simply deprecate it? I initially considered deprecation, but now I’m not sure it’s appropriate here, since deprecation generally implies the feature will continue to function as before, which is not the case.

osiegmar: @marcphilipp, I think this question is for you. I also favor removing the `lineSeparator` argument.

marcphilipp: I'll discuss this with the team and will get back to you.

sormuras: Skimmed https://www.rfc-editor.org/rfc/rfc4180 and it seems that only `CRLF`, the "Internet standard newline", should be supported. Thus, no auto-detection is required at all, no? 😉 

Having an explicit configuration option to enforce a specific line separator being used in input files still makes some sense to me. At leasst, as a fallback for when auto-detection fails for what-ever reason.

Supporting uncommon separators, like `|`, might be ... uncommon, but still be useful for some folk. For example as in:
```
a,b|c,d|e,f|h,g
```
Not sure, if we should drop support for such use-cases.

sbrannen: > Having an explicit configuration option to enforce a specific line separator being used in input files still makes some sense to me. At least, as a fallback for when auto-detection fails for whatever reason.

I agree with what @sormuras said here.

Just update the documentation for `lineSeparator` to explain the auto-detection algorithm and point out that an explicit line separator can be specified to disable auto-detection.

marcphilipp: IIUC FastCSV does not support an explicit line separator: https://github.com/junit-team/junit5/pull/4606/files#discussion_r2122143444

sbrannen: Well, if FastCSV doesn't support it, we have two options.

1. Get FastCSV to support it.
2. Deprecate `lineSeparator` and document that it has been replaced by auto-detection.

If we go for the latter, we could simply "wait and see" to find out if any JUnit users are actually negatively impacted and then reassess our options at that point in time.

osiegmar: Please note that the `lineSeparator` argument in JUnit was never intended—or documented—to be used with any character sequence other than `\r`, `\r\n`, or `\n`, if I understood Marc correctly. However, it is very likely that a few users have misused the flexibility of the missing argument validation to process data that is not actually CSV.

I am hesitant to add support for this in FastCSV, as it could lead to many other non-standard features—not just for CSV, where FastCSV is already quite flexible, but also for entirely non-CSV use cases. This raises many questions about escaping, quoting, handling `\0` characters, and more. Since there is no standard for these cases, different interpretations and expectations could arise, making it difficult to find common ground.

While working on [RFC 4180-bis](https://datatracker.ietf.org/doc/html/draft-shafranovich-rfc4180-bis-07) (the intended successor to RFC 4180, which unfortunately was not published by the IETF), I learned a great deal about the problems caused by incorrect CSV interpretations and implementations. I have summarized my findings here: https://osiegmar.github.io/JavaCsvComparison/

In my view, if someone needs to process data that is not CSV, they should avoid using `@CsvFileSource` and instead implement a custom method source that reads the data in a way appropriate for its format.

vdmitrienko: I agree with this point.
> Please note that the lineSeparator argument in JUnit was never intended—or documented—to be used with any character sequence other than \r, \r\n, or \n

@osiegmar, @marcphilipp, @sormuras, @sbrannen,  what do you think about limiting `lineSeparator` support to just three options: `\r`, `\r\n`, and `\n`?

This should cover the vast majority of current `CsvFileSource` use cases (likely 99%) and I believe it wouldn't introduce any issues on the FastCSV side. Does that sound reasonable to you?

vdmitrienko: To be more precise, FastCSV currently treats `\r`, `\r\n`, and `\n` as valid line separators. Providing the ability to select an exact one could help preserve most of the current JUnit behavior, I believe.

marcphilipp: @vdmitrienko How would you implement that without support in FastCSV?

vdmitrienko: @marcphilipp, I meant asking @osiegmar about supporting this on the FastCSV side. As I understand it, the main concern is that allowing non-standard line separators (anything other than `\r`, `\n`, or `\r\n`) could cause issues for FastCSV in the future. What I'm proposing is to keep support limited to these three standard separators, but allow users to explicitly select which one to use. 

This approach largely maintains compatibility on the JUnit side, since:
1. `lineSeparator()` was originally intended to support only `\r`, `\n`, or `\r\n`.
2. Only one of these separators was used at a time

At the same time, it avoids asking @osiegmar to support anything that could potentially cause issues on the FastCSV side.

sbrannen: @vdmitrienko, your proposal sounds reasonable to me, if @osiegmar is willing to provide support for setting a specific line separator (which only permits `\r`, `\n`, and `\r\n`).

osiegmar: @vdmitrienko As you pointed out earlier, the current implementation is quite inconsistent:

**Regardless** of whether you configure `\n` or `\r\n` as the `lineSeparator`, files with either `\n` or `\r\n` line endings are **parsed identically (!)**, while a standalone `\r` is parsed as a field value. If you configure `\r` as the line separator, it will also parse data with `\r\n` line endings, but will add an additional `null` record at the end.

To me, this is very unexpected behavior. If I were to add fixed line separator handling to FastCSV (which I am very skeptical about), it would behave differently.

Anyone using proper CSV files wouldn’t notice any difference. Anyone using non-compliant CSV files would either have to fix their files (control characters must always be quoted if part of the data) or rely on the same (broken?) implementation we currently have.

vdmitrienko: Then I think we should go with the 2-nd option that @sbrannen suggested:

> Well, if FastCSV doesn't support it, we have two options.
> 
> 1. Get FastCSV to support it.
> 2. Deprecate `lineSeparator` and document that it has been replaced by auto-detection.
> 
> If we go for the latter, we could simply "wait and see" to find out if any JUnit users are actually negatively impacted and then reassess our options at that point in time.

sbrannen: Yep, I'm leaning toward option # 2 at this point as well.

Alternatively, since 6.0 is a "major", we could just remove it altogether. Upgrading to a major sometimes requires adjustments to code, and for this use case users would simply have to delete the `lineSeparator` declaration.

sormuras: Option 2, "Deprecate lineSeparator and document that it has been replaced by auto-detection.", sounds like most reasonable. That is the way.

vdmitrienko: > Alternatively, since 6.0 is a "major", we could just remove it altogether. Upgrading to a major sometimes requires adjustments to code, and for this use case users would simply have to delete the `lineSeparator` declaration.

I think this is the most reasonable approach, because a deprecated feature is expected to continue working as before, but that's not the case here. The `lineSeparator` is completely ignored, and auto-detection is used instead.

sbrannen: OK. Let's just drop it in 6.0.

@marcphilipp and @sormuras, WDYT?

marcphilipp: SGTM

sbrannen: Then we're in agreement: drop the `lineSeparator` in 6.0.

osiegmar:
FastCSV 4.0.0 has just been released: https://github.com/osiegmar/FastCSV/releases/tag/v4.0.0

marcphilipp:
@osiegmar Thank you!

@vdmitrienko I'd like to merge this PR before releasing 6.0.0-M1. Do you have time to continue working on it in the next few days? Please let me know if you're waiting on a decision or anything else.

vdmitrienko:
> @osiegmar Thank you!
> 
> @vdmitrienko I'd like to merge this PR before releasing 6.0.0-M1. Do you have time to continue working on it in the next few days? Please let me know if you're waiting on a decision or anything else.

@marcphilipp, yep I'm working on it right now and plan to finalize it in the next couple of hours 🙂

junit-jupiter-params/LICENSE-fastcsv.md
marcphilipp: Thanks! I pushed a commit to polish that a bit.

marcphilipp: And another one. I'll keep my hands off your branch now. :sweat_smile:

vdmitrienko:
Well, I think everything is ready to hand over this branch to you, @marcphilipp 🙂

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
```diff
@@ -115,6 +115,21 @@ repository on GitHub.
...
   `callsInPlace(executable, AT_MOST_ONCE)` which might result in compilation errors.
+* The `CsvFileSource.lineSeparator()` parameter has been removed. The line separator
+  is now automatically detected, meaning that any of `\r`, `\n`, or `\r\n` is
+  treated as a line separator.
+* As a result of migrating from
+  https://github.com/uniVocity/univocity-parsers[univocity-parsers] to
+  https://fastcsv.org/[FastCSV] for CSV input handling, the root causes and messages
+  of exceptions thrown when parsing malformed CSV input may differ in rare cases.
+  While overall parsing behavior remains consistent, this may affect
+  custom error handling that relies on specific exception types or messages.
```
marcphilipp: I wonder if we should catch FastCSV's `CsvParseException` and convert or wrap it so we users can rely on a specific exception class. WDYT?

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
```diff
@@ -115,6 +115,21 @@ repository on GitHub.
...
+* Parameters such as `ignoreLeadingAndTrailingWhitespace()`, `nullValues()`,
+  and others in `CsvSource` and `CsvFileSource` now apply to header fields as well
+  as to regular fields.
+* Empty lines in `CsvSource.value()` no longer cause a
+  `PreconditionViolationException`; such lines are now ignored, consistent with
+  the behavior of `CsvSource.textBlock()`.
```
marcphilipp: Did we discuss this? I think blank lines in a text block may be used to increase readability while passing blank strings to `@CsvSource` looks fishy. :thinking:

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
marcphilipp: I retract my thumbs-up reaction. :sweat_smile:

I think blank lines in a text block may be used to increase readability while passing blank strings to `@CsvSource` looks fishy. WDYT?

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
marcphilipp: Never mind, found the thread. Let's continue the discussion there: https://github.com/junit-team/junit5/pull/4606#discussion_r2161041819

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
vdmitrienko: > while passing blank strings to @CsvSource looks fishy.

That sounds reasonable. Honestly, it’s hard for me to think of a situation where passing blank strings would actually be necessary.

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
vdmitrienko: We already catch it; only the root causes should have changed. You can check see it in the `CsvArgumentsProviderTests` and `CsvFileArgumentsProviderTests`.

> the root causes and messages of exceptions thrown

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
vdmitrienko: I can take care of it within the next couple of days, or, if you'd like to merge it sooner, you can just revert https://github.com/junit-team/junit5/pull/4606/commits/0872858f29dd0eb5cc85abf9f1c5185ad82b3c31 and mention this behavior in JavaDoc. As @osiegmar noted, it isn't documented at the moment.

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
marcphilipp: You're right, of course. I didn't look closely enough.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvArgumentsProvider.java
marcphilipp: I'll take care of it. :+1:

marcphilipp (approved):
@vdmitrienko Thanks for all the hard work! :+1:

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-M1.adoc
```diff
@@ -115,6 +115,18 @@ repository on GitHub.
 * The contracts for the `Executable` parameters of Kotlin-specific `assertTimeout`
   functions were changed from `callsInPlace(executable, EXACTLY_ONCE)` to
   `callsInPlace(executable, AT_MOST_ONCE)` which might result in compilation errors.
+* As a result of migrating from
+  https://github.com/uniVocity/univocity-parsers[univocity-parsers] to
+  https://fastcsv.org/[FastCSV] for `@CsvSource` and `@CsvFileSource`, root causes and
+  messages  of exceptions thrown for malformed CSV input may differ in some cases. While
```
sormuras:
```suggestion
  messages of exceptions thrown for malformed CSV input may differ in some cases. While
```

gradle/libs.versions.toml
```diff
@@ -99,7 +99,6 @@ buildParameters = { id = "org.gradlex.build-parameters", version = "1.4.4" }
 commonCustomUserData = { id = "com.gradle.common-custom-user-data-gradle-plugin", version = "2.3" }
 develocity = { id = "com.gradle.develocity", version = "4.0.2" }
 errorProne = { id = "net.ltgt.errorprone", version = "4.2.0" }
-extraJavaModuleInfo = { id = "org.gradlex.extra-java-module-info", version = "1.12" }
```
sormuras, line 102: No longer needed because it was only required for `univocity-parsers`?

junit-jupiter-params/junit-jupiter-params.gradle.kts
sormuras: Okay - that answers my question from above.

gradle/libs.versions.toml
marcphilipp: Yes, since FastCSV includes a module descriptor.

sormuras: This is the way.

PR #4618, opened 2025-06-04
Enable NullAway's JSpecify mode and fix resulting issues

marcphilipp:
## Overview

In order to check nullability annotations on generics in the build.

---

---

### Definition of Done


junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/InvocationInterceptor.java
```diff
@@ -218,14 +218,15 @@ default void interceptAfterAllMethod(Invocation<@Nullable Void> invocation,
...
+	interface Invocation<T> {
 
 		/**
 		 * Proceed with this invocation.
 		 *
 		 * @return the result of this invocation; potentially {@code null}.
 		 * @throws Throwable in case the invocation failed
 		 */
+		@Nullable
 		T proceed() throws Throwable;
```
vlsi, line 230: Does it really make sense to force all `Invocation` implementations to return `@Nullable` value?
Would it be better to keep an older `Invocation<T extends @Nullable Object>` approach so every implementation declares if it really wants to produce nullable value or not?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/LoggingListener.java
```diff
@@ -79,14 +79,15 @@ public static LoggingListener forJavaUtilLogging(Level logLevel) {
 	 * @see #forJavaUtilLogging()
 	 * @see #forJavaUtilLogging(Level)
 	 */
-	public static LoggingListener forBiConsumer(BiConsumer<Throwable, Supplier<String>> logger) {
+	public static LoggingListener forBiConsumer(BiConsumer<@Nullable Throwable, Supplier<String>> logger) {
```
vlsi, line 82: In most of the cases, the proper signature for `Function`-receiving methods is as follows:

```suggestion
	public static LoggingListener forBiConsumer(BiConsumer<? super @Nullable Throwable, ? super Supplier<String>> logger) {
```

That would allow users to pass even something that `implements BiConsumer<@Nullable Object, @Nullable Object>`.

Unfortunately, [Java uses use-site variance](https://gist.github.com/vlsi/3e7c6a0aa3f13ec0cfc94c0f7447ceb7#issue-1-filter-does-not-accept-a-more-generic-predicate), so effectively it requires users to add `super` and `extends` for almost every method that uses generic parameters.

jupiter-tests/src/test/java/org/junit/jupiter/engine/execution/AbstractExecutableInvokerTests.java
```diff
@@ -98,6 +98,6 @@ private void register(ParameterResolver... resolvers) {
 
 	abstract void invokeMethod();
 
-	abstract <T> T invokeConstructor(Constructor<T> constructor, @Nullable Object outerInstance);
+	abstract <T> @Nullable T invokeConstructor(Constructor<T> constructor, @Nullable Object outerInstance);
```
vlsi, line 101: It does not sound right for `invokeConstructor` to return null. Does it mean the null signals the failure?

marcphilipp:
@vlsi Thanks for your feedback! I have since talked to @sdeleuze who told me that NullAway's JSpecify mode currently only works properly with JDK 24 because some compiler fixes haven't yet been backported to JDK 21.

marcphilipp:
Superseded by #4628

PR #4619, opened 2025-06-04
Store `annotationType.getSimpleName()` in a field in `BooleanExecutionCondition`

Adagedo:
## Overview

My PR optimizes the performance and readability of the BooleanExecutionCondition class by memoizing the annotationType.SimpleName method.
I did not  alter any logic though. I only introduced  a lightweight internal optimization .

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/BooleanExecutionCondition.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/BooleanExecutionCondition.java
@@ -52,4 +52,5 @@ abstract class BooleanExecutionCondition<A extends Annotation> implements Execut
 		String reason = "@%s is not present".formatted(this.annotationName);
 		return enabled(reason);
 	}
+	
 }


Adagedo:
this is the main PR name. "memoized annotationType.getsimple method to improve performance", the "name their is a typo error"

sbrannen:
> this is the main PR name. "memoized annotationType.getsimple method to improve performance", the "name their is a typo error"

Please note that you can actually edit the title, which I have done for you.

sbrannen (changes requested):
Congratulations on submitting your first PR for JUnit! 👍 

> My PR optimizes the performance and readability of the BooleanExecutionCondition class by memoizing the annotationType.SimpleName method.

There is no performance gain with that change since `annotationType.getSimpleName()` is only invoked once.

I would also argue that it does not really improve the readability of the code.

However, the two subclasses of `BooleanExecutionCondition` both declare `private final String annotationName` fields.

Thus, if you change the `annotationName` field you have introduced to `protected` and modify `AbstractJreCondition` and `AbstractJreRangeCondition` so that they no longer declare or set the `annotationName` field, I would reconsider accepting this PR.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/BooleanExecutionCondition.java
```diff
@@ -43,13 +45,11 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
 				.map(annotation -> isEnabled(annotation) ? enabled(this.enabledReason)
 						: disabled(this.disabledReason, this.customDisabledReason.apply(annotation))) //
 				.orElseGet(this::enabledByDefault);
-	}
-
+
```
sbrannen, line 48: Please do not modify lines of code unrelated to your change, and please do not add or remove blank lines.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/BooleanExecutionCondition.java
```diff
@@ -43,13 +45,11 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
...
-
+	
 	abstract boolean isEnabled(A annotation);
 
 	private ConditionEvaluationResult enabledByDefault() {
-		String reason = "@%s is not present".formatted(this.annotationType.getSimpleName());
+		String reason = "@%s is not present".formatted(this.annotationName);
 		return enabled(reason);
 	}
-
```
sbrannen, line 54: Please restore this blank line.

sormuras:
Micro "optimization" shifting the work from the method (maybe not called) to the constructor (always called). Let's not do it - keep it as-is.

sbrannen:
> Micro "optimization" shifting the work from the method (maybe not called) to the constructor (always called). Let's not do it - keep it as-is.

@sormuras, I think what I proposed in https://github.com/junit-team/junit5/pull/4619#pullrequestreview-2899101120 is worth doing.

sormuras:
I'd rather convert those fields into variables. Why keep/introduce an extra state for something that's only used locally in a single method?

sbrannen:
> Why keep/introduce an extra state for something that's only used locally in a single method?

If we create a private field in in `BooleanExecutionCondition` like in the original proposal in this PR, then I'd agree with you.

However, `AbstractJreCondition` and `AbstractJreRangeCondition` currently both store the simple name in private fields.

So, if we want to continue to store the simple name in fields, we should just store the simple name once in a protected field in the common superclass.

> I'd rather convert those fields into variables. 

If we don't want to continue storing the simple name in fields in any of the affected classes in that hierarchy, yes, we could calculate the simple name based on `super.annotationType.getSimpleName()` in `AbstractJreCondition.validatedVersions()` and `AbstractJreRangeCondition.isCurrentVersionWithinRange()` and leave `BooleanExecutionCondition.enabledByDefault()` as is.

I'm also OK with going with that approach.

@marcphilipp and @sormuras, WDYT?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/BooleanExecutionCondition.java
Adagedo: alright thank you, my apologies, it was unoticed

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/BooleanExecutionCondition.java
Adagedo: Ok I will do just that

marcphilipp:
> > I'd rather convert those fields into variables.
> 
> If we don't want to continue storing the simple name in fields in any of the affected classes in that hierarchy, yes, we could calculate the simple name based on `super.annotationType.getSimpleName()` in `AbstractJreCondition.validatedVersions()` and `AbstractJreRangeCondition.isCurrentVersionWithinRange()` and leave `BooleanExecutionCondition.enabledByDefault()` as is.
> 
> I'm also OK with going with that approach.

I'd prefer that approach. :+1:

junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/BooleanExecutionCondition.java
vlsi: @sbrannen , it would be great to add that into `.editorconfig` so IDEs trim the whitespace automatically.

I've filed a PR for that: https://github.com/junit-team/junit5/pull/4623

marcphilipp:
**Team decision:** @sbrannen himself will change the code according to his proposal in https://github.com/junit-team/junit5/pull/4619#issuecomment-2943561274.

sbrannen:
@Adagedo, thanks again for the PR, even though we decided to address this differently.

> **Team decision:** @sbrannen himself will change the code according to his proposal in [#4619 (comment)](https://github.com/junit-team/junit5/pull/4619#issuecomment-2943561274).

Done in 483fcc90c3cb1852c4c475ccdefeb298a35f5161.

marcphilipp:
> @Adagedo, thanks again for the PR, even though we decided to address this differently.

+1 :+1:

Adagedo:
@sbrannen My pleasure

PR #4626, opened 2025-06-05
chore: replace List<@Nullable Object> with List<?>

vlsi:
## Overview

The latter is easier to read, and it forbids list.add(...) at the compile-time.

See https://jspecify.dev/docs/user-guide/#wildcard-bounds

---

---

### Definition of Done

### DIFF
--- a/junit-platform-commons/src/main/java/org/junit/platform/commons/support/AnnotationSupport.java
+++ b/junit-platform-commons/src/main/java/org/junit/platform/commons/support/AnnotationSupport.java
@@ -450,7 +450,9 @@ public final class AnnotationSupport {
 		List<Field> fields = findAnnotatedFields(instance.getClass(), annotationType, ModifierSupport::isNotStatic,
 			HierarchyTraversalMode.TOP_DOWN);
 
-		return ReflectionUtils.readFieldValues(fields, instance);
+		@SuppressWarnings("unchecked")
+		List<@Nullable Object> result = (List<@Nullable Object>) ReflectionUtils.readFieldValues(fields, instance);
+		return result;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


vlsi:
It looks like AssertJ can't assert on values like `List<?>` and `List<? extends Object>`:

```
List<?> values = List.of("1", 2)

assertThat(values).containsExactly("1", 2); // this does not compile
Assertions.<Object>assertThat(values).containsExactly("1", 2); // this compiles
```

Currently, the build fails because some of the JUnit tests attempt things like the following:

```java
var values = readFieldValues(fields, new ClassWithFields());
assertThat(values).containsExactly("enigma", 3.14, "text", 2.5, null, 42, "constant", 99);
```

I'm not sure if AssertJ's `assertThat` is the reason to refrain from `List<?>`.

vlsi:
A funny case is that an illegal `.set` was there in the test code:

```java
// Modify local copy:
assertThrows(UnsupportedOperationException.class, () -> copy.set(0, "Boom!"));
```

The code fails to compile with `List<?>`, so it is a good case: it moves a runtime failure to a compile-time failure.

marcphilipp (changes requested):
Since this is a breaking change, I'm not convinced we should do it for public APIs. Could you please revert all changes except those in `ReflectionUtils` and its test class?

junit-platform-commons/src/main/java/org/junit/platform/commons/support/AnnotationSupport.java
```diff
@@ -450,7 +450,9 @@ public static List<Field> findAnnotatedFields(Class<?> clazz, Class<? extends An
 		List<Field> fields = findAnnotatedFields(instance.getClass(), annotationType, ModifierSupport::isNotStatic,
 			HierarchyTraversalMode.TOP_DOWN);
 
-		return ReflectionUtils.readFieldValues(fields, instance);
+		@SuppressWarnings("unchecked")
+		List<Object> result = (List<Object>) ReflectionUtils.readFieldValues(fields, instance);
+		return result;
```
vlsi: I'm not fond of this cast, however, it seems to be the least evil as the added suppression holds for a single statement only rather than a method.

marcphilipp:
Thanks, @vlsi! :+1:

PR #4628, opened 2025-06-05
Use JDK 24 and enable NullAway's JSpecify mode

marcphilipp:
## Overview

- **Use JDK 24 toolchain to run Gradle and Java-related tasks**
- **Avoid ByteBuddy from trying to use `Unsafe`**
- **Enable NullAway's JSpecify mode**
- **Resolve nullability issues surfaced by NullAway's JSpecify mode**

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
@@ -32,22 +32,17 @@ import org.opentest4j.AssertionFailedError;
 @API(status = STABLE, since = "5.9")
 public class AssertionFailureBuilder {
 
-	@Nullable
-	private Object message;
+	private @Nullable Object message;
 
-	@Nullable
-	private Throwable cause;
+	private @Nullable Throwable cause;
 
 	private boolean mismatch;
 
-	@Nullable
-	private Object expected;
+	private @Nullable Object expected;
 
-	@Nullable
-	private Object actual;
+	private @Nullable Object actual;
 
-	@Nullable
-	private String reason;
+	private @Nullable String reason;
 
 	private boolean includeValuesInMessage = true;
 
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
@@ -34,8 +34,7 @@ public abstract class DynamicNode {
 	private final String displayName;
 
 	/** Custom test source {@link URI} associated with this node; potentially {@code null}. */
-	@Nullable
-	private final URI testSourceUri;
+	private final @Nullable URI testSourceUri;
 
 	DynamicNode(String displayName, @Nullable URI testSourceUri) {
 		this.displayName = Preconditions.notBlank(displayName, "displayName must not be null or blank");
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/OS.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/OS.java
@@ -95,8 +95,7 @@ public enum OS {
 
 	private static final Logger logger = LoggerFactory.getLogger(OS.class);
 
-	@Nullable
-	private static final OS CURRENT_OS = determineCurrentOs();
+	private static final @Nullable OS CURRENT_OS = determineCurrentOs();
 
 	/**
 	 * {@return the current operating system, if known; otherwise, {@code null}}
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/AbstractExtensionContext.java
@@ -55,8 +55,7 @@ abstract class AbstractExtensionContext<T extends TestDescriptor> implements Ext
 
 	private static final Logger LOGGER = LoggerFactory.getLogger(AbstractExtensionContext.class);
 
-	@Nullable
-	private final ExtensionContext parent;
+	private final @Nullable ExtensionContext parent;
 	private final EngineExecutionListener engineExecutionListener;
 	private final T testDescriptor;
 	private final Set<String> tags;
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
@@ -93,11 +93,9 @@ public abstract class ClassBasedTestDescriptor extends JupiterTestDescriptor
 
 	protected final ClassInfo classInfo;
 
-	@Nullable
-	private LifecycleMethods lifecycleMethods;
+	private @Nullable LifecycleMethods lifecycleMethods;
 
-	@Nullable
-	private TestInstanceFactory testInstanceFactory;
+	private @Nullable TestInstanceFactory testInstanceFactory;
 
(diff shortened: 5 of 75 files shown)


gradle/base/dsl-extensions/build.gradle.kts
```diff
@@ -1,3 +1,14 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_21
+
 plugins {
     `kotlin-dsl`
 }
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_21
```
vlsi: please do not use compatibility properties, use release instead.

See https://github.com/apple/pkl/pull/1080

gradle/base/dsl-extensions/build.gradle.kts
```diff
@@ -1,3 +1,14 @@
...
+    targetCompatibility = JavaVersion.VERSION_21
+}
+
+kotlin {
+    compilerOptions.jvmTarget = JVM_21
```
vlsi: Add `-Xjdk-release` as well

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/NamespaceAwareStore.java
```diff
@@ -41,49 +41,58 @@ public NamespaceAwareStore(NamespacedHierarchicalStore<Namespace> valuesStore, N
...
 		Preconditions.notNull(defaultCreator, "defaultCreator function must not be null");
-		return accessStore(() -> this.valuesStore.getOrComputeIfAbsent(this.namespace, key, defaultCreator));
+		Supplier<@Nullable Object> action = () -> this.valuesStore.getOrComputeIfAbsent(this.namespace, key,
+			defaultCreator);
+		return accessStore(action);
 	}
 
 	@Override
-	public <K, V> @Nullable V getOrComputeIfAbsent(K key, Function<K, V> defaultCreator, Class<V> requiredType) {
+	public <K, V extends @Nullable Object> @Nullable V getOrComputeIfAbsent(K key, Function<K, V> defaultCreator,
```
vlsi: Should this be `Function<? super K, ? extends V>` ?

vlsi: Why the return value is "forced nullable"? I guess the nullability should be the same as V's one

vlsi (commented):
It would be great to stop using sourceCompatibility/targetCompatibility properties

sdeleuze:
Setting the "release" is indeed better, possible via Gradle JavaCompile options.

marcphilipp:
> It would be great to stop using sourceCompatibility/targetCompatibility properties

> Setting the "release" is indeed better, possible via Gradle JavaCompile options.

I've only used them for build logic projects, not production code. I'll change it to use `--release` instead.

sormuras:
> I'll change it to use `--release` instead.

Didn't the Eclipse project export feature had issues with this? 🤔

vlsi:
@marcphilipp , @sdeleuze , @sormuras , it was a pleasure to listen to your talks and comments at IDEA Conf. Many thanks for pushing all this forward.

marcphilipp:
> > I'll change it to use `--release` instead.
> 
> Didn't the Eclipse project export feature had issues with this? 🤔

@sormuras Are you referring to the separate setting targeting Eclipse that @sbrannen added in ac0b9c1551a828d741aa9e0c1904724c83bea7ad?

gradle/base/dsl-extensions/build.gradle.kts
marcphilipp: Thanks for the pointer! I didn't know about this flag and have added it now.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/NamespaceAwareStore.java
marcphilipp: > Should this be `Function<? super K, ? extends V>` ?

Yes, they should probably have been since the start. Since we're working towards a major release, now would be a good time to change them. I'll take care of that.

> Why the return value is "forced nullable"? I guess the nullability should be the same as V's one

If a previous call has stored `null` using `key` (e.g. as `put(key, null)`), then the return value will be `null` regardless of whether `V` is nullable.

vlsi: Javadoc says the key must not be `null`: https://github.com/junit-team/junit5/blob/7d02a6a2521617ab8502f8614794b7bb0a568603/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java#L672

ALUMINIS650:
Anffything basic i can help with?

ALUMINIS650:
How do we enable the mode?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/NamespaceAwareStore.java
marcphilipp: I'm referring to the value, not the key.

vlsi: Got it. So the issue boils down to the fact that neither the `Store` nor the `key` are parameterized.

Is it worth filing an issue to streamline the signatures? Frankly, I find it sad that currently every `store.get/put` call site has to to carefully use the same types. It is prone to errors.
I would like to have `Store<Key, Value>` or have something like `Key<KeyType, ValueType>` + `ValueType Store.get(Key<KeyType, ValueType> key)` so `put("key", 1); put("key", 2);` would be impossible

marcphilipp: `Store` is used across all extensions to store all kinds of objects so I don't think adding class-level type parameters is feasible.

vlsi: If class-level "key-value" parameter is not feasible, then it should be very well feasible to add type parameter to the key:

```java
interface Store {
  <KeyType, ValueType> ValueType get(StoreKey<KeyType, ValueType> key);
  <KeyType, ValueType> ValueType put(StoreKey<KeyType, ValueType> key, ValueType value);
}

interface StoreKey<KeyType, ValueType> {
}
```

Then the compiler would be able to type-check the values associated with every key

marcphilipp: That would be a big API change and break many existing extensions. Therefore, I'm not convinced it's worth doing.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/ResolutionCache.java
```diff
@@ -36,7 +36,7 @@ class Concurrent implements ResolutionCache {
 		private final Map<ParameterDeclaration, Object> cache = new ConcurrentHashMap<>();
 
 		@Override
-		public Object resolve(ParameterDeclaration declaration, Supplier<Object> resolver) {
+		public @Nullable Object resolve(ParameterDeclaration declaration, Supplier<@Nullable Object> resolver) {
```
vlsi: Should it be `Supplier<? extends @Nullable Object>`?

I am afraid the current type signature does not accept `Supplier<@NonNull Object>` suppliers.

marcphilipp: It's only used internally so the impact is minimal but I simplified it to `Supplier<?>` in the most recent push.

vlsi: Assertions have a lot of `Supplier<@Nullable String>` code which would be incompatible with `Supplier<String>` values :-/

PR #4629, opened 2025-06-06
Add `withPosition(FilePosition)` method to `FileSource`

lslonina:
- Refactor creation to avoid redundant canonicalization calls
- Introduced internal factory methods for clearer construction paths

## Overview

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
@@ -112,8 +112,13 @@ public final class FileSource implements FileSystemSource {
 	}
 
 	/**
-	 * {@return a new {@code FileSource} based on this instance but with the
+	 * {@return a {@code FileSource} based on this instance but with the
 	 * supplied {@link FilePosition}}
+	 *
+	 * <p>If the supplied {@code FilePosition}
+	 * {@linkplain Objects#equals(Object, Object) equals} the existing one, this
+	 * method returns {@code this}. Otherwise, a new instance is created and
+	 * returned.
 	 *
 	 * <p>Calling this method rather than creating a new {@code FileSource} via
 	 * {@link #from(File, FilePosition)} avoids the overhead of redundant
@@ -124,6 +129,9 @@ public final class FileSource implements FileSystemSource {
 	 */
 	@API(status = EXPERIMENTAL, since = "6.0")
 	public FileSource withPosition(@Nullable FilePosition filePosition) {
+		if (Objects.equals(this.filePosition, filePosition)) {
+			return this;
+		}
 		return new FileSource(this, filePosition);
 	}
 
--- a/platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/FileSystemSourceTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/engine/support/descriptor/FileSystemSourceTests.java
@@ -90,6 +90,11 @@ class FileSystemSourceTests extends AbstractTestSourceTests {
 		assertThat(source).isNotSameAs(sourceWithPosition);
 		assertThat(source.getFile()).isSameAs(sourceWithPosition.getFile());
 		assertThat(sourceWithPosition.getPosition()).hasValue(position);
+
+		assertThat(sourceWithPosition.withPosition(null).getPosition()).isEmpty();
+
+		assertThat(source.withPosition(null)).isSameAs(source);
+		assertThat(sourceWithPosition.withPosition(position)).isSameAs(sourceWithPosition);
 	}
 
 	@Test


junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
```diff
@@ -55,27 +55,45 @@ public static FileSource from(File file) {
...
-	private FileSource(File file) {
-		this(file, null);
-	}
-
 	private FileSource(File file, @Nullable FilePosition filePosition) {
+		this.file = file;
+		this.filePosition = filePosition;
+	}
+	
+	private static FileSource canonicalized(File file, @Nullable FilePosition filePosition) {
```
mpkorstanje: This method can be much smaller:

```suggestion
	private static File canonicalized(File file) {
```

And then used as `new FileSource(canonicalized(file), filePosition)`, and in turn the precondition can be lifted up to the factory methods.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
```diff
@@ -55,27 +55,45 @@ public static FileSource from(File file) {
...
+	 * Create a new {@code FileSource} from an existing instance but with a different
+	 * {@link FilePosition}. This avoids redundant canonical path resolution.
+	 *
+	 * @param source the existing {@code FileSource}; must not be {@code null}
+	 * @param filePosition the new {@code FilePosition}; may be {@code null}
+	 * @return a new {@code FileSource} with same file and updated position
+	 */
+	public static FileSource withPosition(FileSource source, @Nullable FilePosition filePosition) {
+		Preconditions.notNull(source, "source must not be null");
+		return direct(source.file, filePosition);
```
mpkorstanje: invoking `direct` here is redundant. There are now precondition checks for essentially the same thing.

```suggestion
		return new FileSource(source.file, filePosition);
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
```diff
@@ -55,27 +55,45 @@ public static FileSource from(File file) {
...
+
+	/**
+	 * Create a new {@code FileSource} from an existing instance but with a different
+	 * {@link FilePosition}. This avoids redundant canonical path resolution.
+	 *
+	 * @param source the existing {@code FileSource}; must not be {@code null}
+	 * @param filePosition the new {@code FilePosition}; may be {@code null}
+	 * @return a new {@code FileSource} with same file and updated position
+	 */
+	public static FileSource withPosition(FileSource source, @Nullable FilePosition filePosition) {
```
mpkorstanje: This needs an `@API` annotation to mark the method as experimental with a `since = "1.14"` field.

mpkorstanje (commented):
Thanks for starting this! 

I've left some remarks. 

But keep in mind the way JUnit team works. They'll have a brief team discussion before making a decision. So they may request additional changes or propose an entirely different solution direction al together. 

So bear that in mind before putting in too much work. 

You'll also want to polish up the pull request description, or if you're still working on it, mark it as a draft.  Currently you're missing an reference to the issue that is being addressed:

1. 
https://github.com/junit-team/junit5/blob/main/CONTRIBUTING.md#commit-messages

mpkorstanje:
I'm not sure why my review got posted twice. I've hidden one, but now they're both hidden. :confused:

ALUMINIS650:
how to add?

lslonina:
@mpkorstanje Thanks for the feedback — the code can definitely be simplified, especially with the instance method approach. Let’s see how the discussion on this topic evolves; I’ll keep this marked as a draft for now.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
marcphilipp: It should actually be `since = "6.0"` since we're working towards that version (even in Platform projects).

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
```diff
@@ -55,26 +56,23 @@ public static FileSource from(File file) {
...
+		}
+		catch (IOException ex) {
+			throw new JUnitException("Failed to retrieve canonical path for file: " + file, ex);
+		}
 	}
 
 	private final File file;
 
 	@Nullable
 	private final FilePosition filePosition;
```
vlsi: It should probably be `private final @Nullable FilePosition filePosition` since the nullability belongs to a type rather than to a field

marcphilipp: Yep, you're right. I initially did it wrong and tried to replace them all using a pattern. Looks like I missed a few. I'll do another pass.

vlsi: There was an OpenRewrite issue for rewriting the type annotations close to the type, however, I'm not sure if it has been implemented or not: https://github.com/openrewrite/rewrite/issues/3880

marcphilipp: Added to #4628

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
```diff
@@ -104,6 +102,20 @@ public final Optional<FilePosition> getPosition() {
...
+	* Return a new {@code FileSource} based on this instance but with a different
+	* {@link FilePosition}. This avoids redundant canonical path resolution
+	* by reusing the already-canonical file.
+	*
+	* @param filePosition the new {@code FilePosition}; must not be {@code null}
+	* @return a new {@code FileSource} with the same file and updated position
+	*/
+	@API(status = EXPERIMENTAL, since = "6.0")
+	public FileSource withPosition(FilePosition filePosition) {
+		Preconditions.notNull(filePosition, "position must not be null");
```
mpkorstanje:
```suggestion
		Preconditions.notNull(filePosition, "filePosition must not be null");
```

mpkorstanje:
> @mpkorstanje Thanks for the feedback — the code can definitely be simplified, especially with the instance method approach. Let’s see how the discussion on this topic evolves; I’ll keep this marked as a draft for now.

@lslonina the issue has moved off team discussion. Do you have time available to patch up your PR or would you prefer to hand it over?

lslonina:
Thanks for picking this up. I’ve added `@API(status = EXPERIMENTAL, since = "6.0")`, but if this is a backport candidate, should we change it to `since = "5.13.5"` instead?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/descriptor/FileSource.java
```diff
@@ -104,6 +102,20 @@ public Optional<FilePosition> getPosition() {
...
+	* {@link FilePosition}. This avoids redundant canonical path resolution
+	* by reusing the already-canonical file.
+	*
+	* @param filePosition the new {@code FilePosition}; must not be {@code null}
+	* @return a new {@code FileSource} with the same file and updated position
+	*/
+	@API(status = EXPERIMENTAL, since = "6.0")
+	public FileSource withPosition(FilePosition filePosition) {
+		Preconditions.notNull(filePosition, "filePosition must not be null");
+		return new FileSource(this.file, filePosition);
```
marcphilipp: Please revert the changes to the constructors and factory methods above and introduce a copy constructor as proposed in https://github.com/junit-team/junit-framework/issues/4630#issuecomment-3279151335 instead.

```suggestion
		return new FileSource(this, filePosition);
```

marcphilipp:
> Thanks for picking this up. I’ve added `@API(status = EXPERIMENTAL, since = "6.0")`, but if this is a backport candidate, should we change it to `since = "5.13.5"` instead?

Good question! Please keep using `6.0`. We'll adjust it when backporting.

marcphilipp:
@lslonina Are you going to finish this PR or is @mpkorstanje taking over?

marcphilipp (approved):

marcphilipp:
Thanks, @lslonina! :+1:

Issue #4698, opened 2025-06-30
`ConditionEvaluationResult` APIs should accept `@Nullable` reasons


sbrannen:
## Overview

In #4550 we added support for JSpecify nullability annotations; however, the `reason` and `customReason` parameters in various `ConditionEvaluationResult` APIs do not signal that those values are permitted to be `null`.

For example, when migrating to JUnit 6.0 M1 I had to add the following suppression in Spring Framework's `DisabledInAotModeCondition`.

```java
@SuppressWarnings("NullAway") // ConditionEvaluationResult.disabled(): customReason should be @Nullable
```

## Related Issues

- #4550 
- #4715 
- https://github.com/spring-projects/spring-framework/issues/35124

PR #4699, opened 2025-06-30
Support `@Nullable` reasons in `ConditionEvaluationResult` APIs

sbrannen:
## Overview

This PR addresses the issues raised in #4698 by annotating all "reason" parameters with `@Nullable`.

## Related Issues

- #4698
- #4715

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ConditionEvaluationResult.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ConditionEvaluationResult.java
@@ -15,6 +15,7 @@ import static org.apiguardian.api.API.Status.STABLE;
 import java.util.Optional;
 
 import org.apiguardian.api.API;
+import org.jspecify.annotations.Nullable;
 import org.junit.platform.commons.util.StringUtils;
 import org.junit.platform.commons.util.ToStringBuilder;
 
@@ -35,7 +36,7 @@ public class ConditionEvaluationResult {
 	 * or an <em>empty</em> reason if the reason is unknown
 	 * @see StringUtils#isBlank(String)
 	 */
-	public static ConditionEvaluationResult enabled(String reason) {
+	public static ConditionEvaluationResult enabled(@Nullable String reason) {
 		return new ConditionEvaluationResult(true, reason);
 	}
 
@@ -48,7 +49,7 @@ public class ConditionEvaluationResult {
 	 * or an <em>empty</em> reason if the reason is unknown
 	 * @see StringUtils#isBlank(String)
 	 */
-	public static ConditionEvaluationResult disabled(String reason) {
+	public static ConditionEvaluationResult disabled(@Nullable String reason) {
 		return new ConditionEvaluationResult(false, reason);
 	}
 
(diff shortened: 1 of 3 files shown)


sbrannen:
> However, while working on this issue I realized two things.
> 
> 1. We always allowed `null` for any of the reasons, but the Javadoc did not reflect that.
> 
> 2. When a custom reason was supplied but a `null` was supplied for the default reason to `ConditionEvaluationResult.disabled(String, String)`, the generated reason was `"null ==> custom"`; whereas, ideally that should just be `"custom"`.
> 3. We allowed a _blank_ reason (such as an empty string or a string containing only whitespace).

We should decide if we want to backport any of the following to 5.13.x.


- the "null is allowed" Javadoc changes
- the `"null ==> custom"` _bug_
- the conversion of an empty/blank reason string to an empty `Optional`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ConditionEvaluationResult.java
```diff
@@ -29,47 +30,57 @@ public class ConditionEvaluationResult {
...
+	 * or an <em>empty</em> reason
 	 * @since 5.7
 	 */
 	@API(status = STABLE, since = "5.7")
-	public static ConditionEvaluationResult disabled(String reason, String customReason) {
+	public static ConditionEvaluationResult disabled(@Nullable String reason, @Nullable String customReason) {
 		if (StringUtils.isBlank(customReason)) {
 			return disabled(reason);
 		}
+		if (StringUtils.isBlank(reason)) {
```
marcphilipp, line 75: I think it would be more conventional to check the first argument first.

marcphilipp (approved):

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ConditionEvaluationResult.java
```diff
@@ -29,47 +30,57 @@ public class ConditionEvaluationResult {
...
 
 	private final boolean enabled;
 
 	private final Optional<String> reason;
 
-	private ConditionEvaluationResult(boolean enabled, String reason) {
+	private ConditionEvaluationResult(boolean enabled, @Nullable String reason) {
 		this.enabled = enabled;
-		this.reason = Optional.ofNullable(reason);
+		this.reason = Optional.ofNullable(StringUtils.isBlank(reason) ? null : reason.strip());
```
marcphilipp:
```suggestion
		this.reason = StringUtils.isBlank(reason) ? Optional.empty() : Optional.of(reason.strip());
```

marcphilipp:
> We should decide if we want to backport any of the following to 5.13.x.
> 
> * the "null is allowed" Javadoc changes
> * the `"null ==> custom"` _bug_
> * the conversion of an empty/blank reason string to an empty `Optional`

I think we should backport all of them.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ConditionEvaluationResult.java
sbrannen: Indeed. I just pushed a "Polishing" commit that includes that change.

sbrannen:
> I think we should backport all of them.

In that case, I will create a separate GitHub issue to track all of that, since they are actually unrelated to the `@Nullable` changes necessary on `main` for 6.0.

PR #4704, opened 2025-07-01
activate `removeWildcardImports()`

Pankraz76:
activation of:

- https://github.com/diffplug/spotless/pull/2517

Might be forbidden to use wildcard imports (e.g., import static org.junit.jupiter.api.Assertions.*;) in Java code.

But build seems not to care about. As my BUILD SUCCESSFUL.

---

---

### Definition of Done


sbrannen:
When I run the build with that wildcard import, it fails:

```
> Task :junit-jupiter-api:checkstyleMain
[ant:checkstyle] [ERROR] /Users/sbrannen/source/junit5/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/AbstractJreCondition.java:17:26: Using the '.*' form of import should be avoided - java.util.function.*. [AvoidStarImport]

> Task :junit-jupiter-api:checkstyleMain FAILED
```

In other words, this is already covered by Checkstyle.

https://github.com/junit-team/junit-framework/blob/7dc279167e80b20aab3432a393cf98cb7e42e2f4/gradle/config/checkstyle/checkstyleMain.xml#L19

sbrannen:
With regard to the aforementioned [`removeWildcardImports()` feature](https://github.com/diffplug/spotless/pull/2517) in Spotless, that does not yet appear to have been released in the Gradle plugin.

In any case, removing wildcard imports without replacing them with actual valid imports does not seem very helpful in my opinion.

Pankraz76:
`gradle build -x test` did not fail for me.

Pankraz76:
while having the same feat. spotless goes beyond annoying but fixing the issue. Actually there is no need to have checkstyle anymore if spot is configured, its ready to take over.

sbrannen:
> `gradle build -x test` did not fail for me.

Before submitting a PR (or before the core team pushes anything to GitHub), a full build with checks and tests is required.

`./gradlew check` will fail, as will `./gradlew build`.

As for whether `-x test` should omit the Checkstyle checks, I suppose that is debatable.

Pankraz76:
right, sorry for that.



> should omit the Checkstyle checks

Yes in maven it would be a verify phase, gradle has something similar for sure.

sbrannen:
> while having the same feat. spotless goes beyond annoying but fixing the issue. Actually there is no need to have checkstyle anymore if spot is configured, its ready to take over.

From what I can tell, the new Spotless feature does not "fix" anything.

Rather, it simply removes all wildcard imports, which can result in the code no longer compiling.

Thus, rather than "fixing" the code, I'd say it actually "breaks" the code.

Pankraz76:
Yes in this case its actually true, wanted to mention this detail as well. But the this is the exception. Normally spot will replace with equivalent fix. We will work on the full replacement that IDEA offers.

If IDEA is correctly configured, avoiding wildcards, this could be leveraged for skilled IDE users, as the non-compiling file will be open in an ease, and then imports are added automatically sometimes.

Or just replace the diff in IDE and apply the wildcard replacement. Yes thats complicated as requires one extra step.

While check just breaks, assuming it offers no navigation.



> "breaks" the code

"Breaks" the build, assuming and hoping that you cannot deliver non-compiling code.  

So that's ment to be a feature, but it still imposes a small risk on a build, if its not checking for compiling. This will cause problems with this solution and should be reconsidered.

For real security and stability, we should just fail the build on the spot and not change it, in my opinion.

Just changing the goal from `apply` to `check`, as a failing build, with manual interaction required, is the desired outcome.

This is a very small risk, but a real one.  
So the step should just fail the build during apply, as a real apply is not currently possible—assuming other steps deliver intentionally breaking code.  

Whats you opinion ?
@iddeepak  
@Goooler

sbrannen:
> For real security and stability, we should just fail the build on the spot and not change it, in my opinion.

That is the exact behavior we currently have with the Checkstyle check. See https://github.com/junit-team/junit-framework/pull/4704#issuecomment-3023341806.

> If IDEA is correctly configured, avoiding wildcards, 

The same is true for any IDE that one uses to work on JUnit code, and the core JUnit team developers rely on that, as I assume most developers do.

PR #4760, opened 2025-07-12
Add `computeIfAbsent` variants for non-nullable types to stores

marcphilipp:
Since both `ExtensionContext.Store` and its backing
`NamespaceAwareStore` implementation allow storing `null` values, the
`getOrComputeIfAbsent` methods potentially returned `null`, even if
documented otherwise (in case of `getOrComputeIfAbsent(Class)` in
`ExtensionContext.Store`). Now that the API uses JSpecify's nullability
annotations, this became more apparent and also caused
Jupiter `Extension` implementations to have to deal with values being
potentially `null` despite their best intentions to only use
`getOrComputeIfAbsent` to access a certain `key` with a function that
never returned `null`. Therefore, this commit introduces null-safe
variants called `computeIfAbsent` that treat `null` as absent in
addition to the key not being present in the backing `Map`.

---

---

### Definition of Done


marcphilipp:
Sam and I went through the code in today's team call without finding any issues.

Issue #4716, opened 2025-07-02
Introduce `quoteText` support in `@ParameterizedClass` and `@ParameterizedTest`


sbrannen:
## Overview

While discussing #4714 with @sormuras, he inquired about the `'null'` display name, which sparked an idea...

Namely, it would be quite nice to have an auto-quote feature for argument values in the display name for a `@ParameterizedClass` or `@ParameterizedTest`.

I therefore propose that we introduce a `boolean quoteText() default true` attribute (or similar) in `@ParameterizedClass` or `@ParameterizedTest` which signals to the `ParameterizedInvocationNameFormatter` that it should wrap `CharSequence` values (naturally including `String`) in double quotes (`""`) and wrap `char`/`Character` values in single quotes (`''`).

Defaulting to `true` would make this an opt-out feature, but we could of course default to `false` to make it an opt-in feature.

Though, since we are introducing this in 6.0, perhaps it's better to make it an opt-out feature.

However, a `null` `CharSequence`/`Character` value would continue to be displayed as `null`, not `"null"` or `'null'`.

## Related Issues

- #4714

sbrannen:
As discussed in https://github.com/junit-team/junit-framework/issues/4714#issuecomment-3031581856, the `quoteText` feature would result in automatic escaping of special characters within `""` and `''` quoted values.

sbrannen:
#### Update

Current work on this can be viewed in the following feature branch.

https://github.com/junit-team/junit-framework/compare/main...sbrannen:junit-framework:issues/4716-quote-text-parameterized-tests

Issue #1713, opened 2018-12-28
New line characters in (dynamic) display names


sormuras:
New line characters may disrupt layout of the tree-based test plan rendering:

```
[INFO] Launching JUnit Platform 1.4.0-M1...
[INFO] .
[INFO] '-- JUnit Jupiter [OK]
[INFO]   +-- LayoutTests [OK]
[INFO]   | +-- readModuleName() [OK]
[INFO]   | | +-- module a { [OK]
[INFO]   | | +-- open module a{ [OK]
[INFO]   | | '-- what ever 
[INFO]  prefix module    a   
[INFO] { [OK]
[INFO]   | '-- checkMavenProjects() [OK]
...
```

Other [ASCII control characters](https://en.wikipedia.org/wiki/Control_character) may even produce side-effects or more severe issues.

Should the tree renderer replace those control characters before emitting them?

What about other display name emitters, like reporting?

sbrannen:
Tentatively slated for 5.4 M2 for _team discussion_.

sbrannen:
**Team Decision**:

- _Sanitize_ display names before they are output via the `ConsoleLauncher`.
  - Consider introducing a sanitizing convenience method somewhere in `junit-platform-commons`.
- Do not do anything to display names otherwise. In other words, due to backward compatibility constraints (e.g., `TestDescriptor#getDisplayName()`), we cannot enforce any rules other than "not null and not blank". Build tools and IDEs are responsible for _consuming_ display names and deciding how to display them.

sbrannen:
Here's a new example in our own test suite.....

https://travis-ci.org/junit-team/junit5/jobs/475989143#L819

😉

sbrannen:
FYI: that comes from the new examples for `@NullSource` and `@EmptySource` introduced in #1637.

sbrannen:
It actually occurs twice: once in the above link, and [here](https://travis-ci.org/junit-team/junit5/jobs/475989143#L828) as well.

sbrannen:
FWIW, I noticed that Eclipse IDE simply displays all of those display names containing only whitespace, newlines, etc. as an _empty_ string in the GUI.

sormuras:
_in progress..._

sormuras:
> Here's a new example in our own test suite.....
> https://travis-ci.org/junit-team/junit5/jobs/475989143#L819

```
│  │  ├─ NullAndEmptySource_2 ✔
[...]

PR #4813, opened 2025-08-06
Replace ISO control characters in display names

### LINKED ISSUE #4714: Replace ISO control characters in display names
## Background

While working on #4698, I created tests using _blank_ strings like the following.

```java
@ParameterizedTest
@NullSource
@ValueSource(strings = { "", " ", "   ", "\t", "\r", "\r\n", "\n" })
@interface EmptyReasonsTest {
}
```

That produces display names similar to the following, which are not very useful since the last 7 are identical.

Image

If I wrap the values in single quotes as follows...

```java
@ParameterizedTest(name = "[{index}] reason=''{0}''")
```

... the results are better.

Image

## Proposal

But... I think we can do even better.

Namely if we replace non-printable characters, then configuration like this...

```java
@ParameterizedTest(name = "[{index}] reason=''{0}''")
@NullSource
@ValueSource(strings = { "", " ", "   ", "\t", "\r", "\r\n", "\n", "\u200B" })
```

... could have display names like this:

Image

## Proof of Concept

Invoking the following from `ParameterizedInvocationNameFormatter.MessageFormatPartialFormatter`'s `makeReadable()` method resulted in the last screenshot in the _Proposal_ section.

```java
private String replaceNonPrintableCharacters(String string) {
	return string
			.replace("\n", "\\n")
			.replace("\r", "\\r")
			.replace("\t", "\\t")
			.replaceAll("[\\p{Cc}\\p{Cf}\\p{Co}\\p{Cn}]", "?");
}
```

## Related Issues

- #1713
- #4716 
- #4783 
- #4786

sbrannen:
Prior to this commit, display names were "sanitized" before they were printed via the `ConsoleLauncher` (#1713), and text-based arguments in display names for parameterized tests were quoted and escaped (#4716). However, there was still the chance that display names could contain control characters such as CR or LF.

To address that, this commit introduces support for automatically replacing non-printable characters in any display name passed to one of the constructors for `AbstractTestDescriptor`. Doing so automatically covers all display names in Jupiter, `@SuiteDisplayName`, and any other test engines that subclass `AbstractTestDescriptor`, which should cover most common use cases.

Specifically, the following replacements are performed.

- `\r` -> `<CR>`
- `\n` -> `<LF>`
- ISO control character -> � (Unicode replacement character)

This commit also removes the special handling of ISO control characters from `QuoteUtils`, since this is now handled within `AbstractTestDescriptor`.

- See #1713
- See #4716
- Closes #4714

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-RC1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-RC1.adoc
@@ -32,6 +32,11 @@ repository on GitHub.
 * Convention-based conversion in `ConversionSupport` now supports factory methods and
   factory constructors that accept a single `CharSequence` argument in addition to the
   existing support for factories that accept a single `String` argument.
+* Non-printable control characters in display names are now replaced with alternative
+  representations. For example, `\n` is replaced with `<LF>`. This applies to all display
+  names in JUnit Jupiter, `@SuiteDisplayName`, and any other test engines that subclass
+  `AbstractTestDescriptor`. Please refer to the
+  <<../user-guide/index.adoc#writing-tests-display-names, User Guide>> for details.
 
 
 [[release-notes-6.0.0-RC1-junit-jupiter]]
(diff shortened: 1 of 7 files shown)


sbrannen:
@sormuras, I already used Japanese text in the test, but I just added (and force pushed) Thai text in the test.

https://github.com/junit-team/junit-framework/pull/4813/files#diff-56cff8e2d4043d4b5b09a824f4b3073c6fe7ed66ecbd7af98a5025f08489bbfdR172-R194

And I had already used an emoji.

So I think we are adequately testing support for multi-byte characters.

sbrannen:
Running `AbstractTestDescriptorTests.specialCharactersInDisplayNamesAreEscaped(String, String)` in Eclipse IDE now results in the following output in the UI.

AbstractTestDescriptorTests

sormuras (approved):
Escape implementation based on `char` and associated tests look good.

Issue #4810, opened 2025-08-04
Reject blank value for `@SuiteDisplayName`


sbrannen:
## Overview

While investigating #4714, I noticed that we currently ignore a blank value for `@SuiteDisplayName`.

https://github.com/junit-team/junit-framework/blob/99380e033e5afcb4dd5fab67b87184055ab2b662/junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java#L143-L150

However, the Javadoc for `org.junit.platform.suite.api.SuiteDisplayName.value()` clearly states that the value is:

> never blank or consisting solely of whitespace

We should therefore reject blank values for `@SuiteDisplayName` in 6.0, along the lines of similar changes in 6.0.

## Related Issues

- #743
- #242

raccoonback:
@sbrannen 
Hello!
I'm interested in working on this issue.
Would it be possible for me to take it on and start working on it?
Additionally, if there are any guidelines or specific points I should keep in mind while addressing this issue, I'd appreciate it if you could share them.

sbrannen:
Hi @raccoonback,

I see that you've already submitted a PR to address this.

Thus, I have assigned this issue to you.

PR #4822, opened 2025-08-08
Report warning when @SuiteDisplayName is blank or whitespace-only

raccoonback:
### Purpose
This PR adds a discovery-time validation for `@SuiteDisplayName` in the Suite Engine.
If the annotation value is blank or whitespace-only, the engine now reports a `DiscoveryIssue` with WARNING severity.

### Related issue
- https://github.com/junit-team/junit-framework/issues/4810

### Key Changes
- `SuiteTestDescriptor`: invoke `inspectSuiteDisplayName(...)` during construction and report a warning when needed.
  (referred to [DisplayNameUtils.validateAnnotation()](https://github.com/junit-team/junit-framework/blob/24b3c407e9a3640cf3003b7241e1e1378c7c6e89/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DisplayNameUtils.java#L83-L94).)
- I initially considered using [`Preconditions.notBlank()`](https://github.com/junit-team/junit-framework/blob/24b3c407e9a3640cf3003b7241e1e1378c7c6e89/junit-platform-commons/src/main/java/org/junit/platform/commons/util/Preconditions.java#L285-L289) to throw a `PreconditionViolationException` when the value contains only whitespace.
However, to maintain backward compatibility, I decided that reporting a **warning** would be more appropriate than throwing an exception.

### Definition of Done

### DIFF
--- a/junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
+++ b/junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
@@ -87,6 +87,7 @@ final class SuiteTestDescriptor extends AbstractTestDescriptor {
 		this.suiteClass = suiteClass;
 		this.lifecycleMethods = new LifecycleMethods(suiteClass, issueReporter);
 		this.discoveryRequestBuilder.listener(DiscoveryIssueForwardingListener.create(id, discoveryListener));
+		inspectSuiteDisplayName(suiteClass, issueReporter);
 	}
 
 	private static Boolean getFailIfNoTests(Class<?> suiteClass) {
@@ -147,6 +148,16 @@ final class SuiteTestDescriptor extends AbstractTestDescriptor {
 				.filter(StringUtils::isNotBlank)
 				.orElse(testClass.getSimpleName());
 		// @formatter:on
+	}
+
+	private static void inspectSuiteDisplayName(Class<?> suiteClass, DiscoveryIssueReporter issueReporter) {
+		findAnnotation(suiteClass, SuiteDisplayName.class).map(SuiteDisplayName::value).filter(
+			StringUtils::isBlank).ifPresent(__ -> {
+				String message = "@SuiteDisplayName on %s must be declared with a non-blank value.".formatted(
+					suiteClass.getName());
+				issueReporter.reportIssue(DiscoveryIssue.builder(DiscoveryIssue.Severity.WARNING, message).source(
+					ClassSource.from(suiteClass)).build());
+			});
 	}
 
 	void execute(EngineExecutionListener executionListener, NamespacedHierarchicalStore<Namespace> requestLevelStore,
--- a/platform-tests/src/test/java/org/junit/platform/suite/engine/SuiteEngineTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/suite/engine/SuiteEngineTests.java
@@ -68,6 +68,7 @@ import org.junit.platform.suite.engine.testcases.MultipleTestsTestCase;
 import org.junit.platform.suite.engine.testcases.SingleTestTestCase;
 import org.junit.platform.suite.engine.testcases.TaggedTestTestCase;
 import org.junit.platform.suite.engine.testsuites.AbstractSuite;
+import org.junit.platform.suite.engine.testsuites.BlankSuiteDisplayNameSuite;
 import org.junit.platform.suite.engine.testsuites.ConfigurationSuite;
 import org.junit.platform.suite.engine.testsuites.CyclicSuite;
 import org.junit.platform.suite.engine.testsuites.DynamicSuite;
@@ -88,6 +89,7 @@ import org.junit.platform.suite.engine.testsuites.SuiteDisplayNameSuite;
 import org.junit.platform.suite.engine.testsuites.SuiteSuite;
 import org.junit.platform.suite.engine.testsuites.SuiteWithErroneousTestSuite;
 import org.junit.platform.suite.engine.testsuites.ThreePartCyclicSuite;
+import org.junit.platform.suite.engine.testsuites.WhitespaceSuiteDisplayNameSuite;
 import org.junit.platform.testkit.engine.EngineTestKit;
 
 /**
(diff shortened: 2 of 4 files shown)


sbrannen:
> I initially considered using [`Preconditions.notBlank()`](https://github.com/junit-team/junit-framework/blob/24b3c407e9a3640cf3003b7241e1e1378c7c6e89/junit-platform-commons/src/main/java/org/junit/platform/commons/util/Preconditions.java#L285-L289) to throw a `PreconditionViolationException` when the value contains only whitespace.
>       However, to maintain backward compatibility, I decided that reporting a **warning** would be more appropriate than throwing an exception.

Indeed, we do not want to use `Preconditions` for this use case.

Even though I did not explicitly state it in #4810, the intention was to make use of the discovery issue mechanism to report a warning like with `@DisplayName`.

Though, for 6.0 I personally think it should be an error instead of a warning.

@marcphilipp, thoughts?

platform-tests/src/test/java/org/junit/platform/suite/engine/testsuites/BlankSuiteDisplayNameSuite.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.junit.platform.suite.api.SelectClasses;
+import org.junit.platform.suite.api.Suite;
+import org.junit.platform.suite.api.SuiteDisplayName;
+import org.junit.platform.suite.engine.testcases.SingleTestTestCase;
+
+/**
+ * Test suite with blank @SuiteDisplayName to verify validation.
+ *
+ * @since 6.0.0-RC1
```
sbrannen:
```suggestion
 * @since 6.0
```

We do not use values such as `6.0.0-RC1` for `@since` tags.

platform-tests/src/test/java/org/junit/platform/suite/engine/testsuites/WhitespaceSuiteDisplayNameSuite.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.junit.platform.suite.api.SelectClasses;
+import org.junit.platform.suite.api.Suite;
+import org.junit.platform.suite.api.SuiteDisplayName;
+import org.junit.platform.suite.engine.testcases.SingleTestTestCase;
+
+/**
+ * Test suite with whitespace-only @SuiteDisplayName to verify validation.
+ *
+ * @since 6.0.0-RC1
```
sbrannen:
```suggestion
 * @since 6.0
```

sbrannen (approved):
I took a quick glance and think this PR looks pretty good; however, I will probably let @marcphilipp review it as well.

In addition, I requested two very minor changes to Javadoc.

raccoonback:
@sbrannen 
Thank you for reviewing!
I’ve applied the changes based on your feedback.

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
```diff
@@ -149,6 +150,16 @@ private static String getSuiteDisplayName(Class<?> testClass) {
 		// @formatter:on
 	}
 
+	private static void inspectSuiteDisplayName(Class<?> suiteClass, DiscoveryIssueReporter issueReporter) {
```
marcphilipp: I think it would be better to address this in one place by changing `getSuiteDisplayName` above to the following:

```java
private static String getSuiteDisplayName(Class<?> suiteClass, DiscoveryIssueReporter issueReporter) {
	var nonBlank = issueReporter.createReportingCondition(StringUtils::isNotBlank, __ -> {
		String message = "@SuiteDisplayName on %s must be declared with a non-blank value.".formatted(
			suiteClass.getName());
		return DiscoveryIssue.builder(Severity.WARNING, message) //
				.source(ClassSource.from(suiteClass)) //
				.build();
	}).toPredicate();
	return findAnnotation(suiteClass, SuiteDisplayName.class) //
			.map(SuiteDisplayName::value) //
			.filter(nonBlank) //
			.orElse(suiteClass.getSimpleName());
}
```

raccoonback: @marcphilipp 
I think that’s a good suggestion.
I will reflect it!

raccoonback:
@marcphilipp 
I have applied the changes based on your review.
There is one point I would like to confirm with you.
Assuming that an implementation of the `DiscoveryIssueReporter` interface may be added in the future,
do you think we should also guard against cases where the `createReportingCondition(Predicate<T> predicate, Function<T, DiscoveryIssue> issueCreator)` default method is overridden to return null?
(If I’m overthinking this, please let me know...)

```java
class SomethingDiscoveryIssueReporter implements DiscoveryIssueReporter {

 // ...

 @Override
 public <T> Condition<T> createReportingCondition(Predicate<T> predicate, Function<T, DiscoveryIssue> issueCreator) {
	return null;
 }

 // ....

}
```

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
```diff
@@ -140,12 +140,20 @@ public Type getType() {
 		return Type.CONTAINER;
 	}
 
-	private static String getSuiteDisplayName(Class<?> testClass) {
+	private static String getSuiteDisplayName(Class<?> suitClass, DiscoveryIssueReporter issueReporter) {
```
marcphilipp:
```suggestion
	private static String getSuiteDisplayName(Class<?> suiteClass, DiscoveryIssueReporter issueReporter) {
```

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
```diff
@@ -140,12 +140,20 @@ public Type getType() {
...
 		// @formatter:off
-		return findAnnotation(testClass, SuiteDisplayName.class)
+		var nonBlank = issueReporter.createReportingCondition(StringUtils::isNotBlank, __ -> {
+			String message = "@SuiteDisplayName on %s must be declared with a non-blank value.".formatted(
+					suitClass.getName());
+			return DiscoveryIssue.builder(DiscoveryIssue.Severity.WARNING, message)
+					.source(ClassSource.from(suitClass))
```
marcphilipp:
```suggestion
					suiteClass.getName());
			return DiscoveryIssue.builder(DiscoveryIssue.Severity.WARNING, message)
					.source(ClassSource.from(suiteClass))
```

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
```diff
@@ -140,12 +140,20 @@ public Type getType() {
...
+					.build();
+		}).toPredicate();
+
+		return findAnnotation(suitClass, SuiteDisplayName.class)
```
marcphilipp:
```suggestion
		return findAnnotation(suiteClass, SuiteDisplayName.class)
```

junit-platform-suite-engine/src/main/java/org/junit/platform/suite/engine/SuiteTestDescriptor.java
```diff
@@ -140,12 +140,20 @@ public Type getType() {
...
 				.map(SuiteDisplayName::value)
-				.filter(StringUtils::isNotBlank)
-				.orElse(testClass.getSimpleName());
+				.filter(nonBlank)
+				.orElse(suitClass.getSimpleName());
```
marcphilipp:
```suggestion
				.orElse(suiteClass.getSimpleName());
```

platform-tests/src/test/java/org/junit/platform/suite/engine/SuiteEngineTests.java
```diff
@@ -705,6 +707,38 @@ void reportsChildrenOfEnginesInSuiteAsSkippedWhenCancelledDuringExecution() {
...
+
+		assertThat(testKit.discover().getDiscoveryIssues()).contains(expectedIssue);
+	}
+
+	@Test
+	void validSuiteDisplayNameGeneratesNoWarning() {
+		var testKit = EngineTestKit.engine(ENGINE_ID).selectors(selectClass(SuiteDisplayNameSuite.class));
+
+		assertThat(testKit.discover().getDiscoveryIssues()).noneMatch(
+			issue -> issue.message().contains("@SuiteDisplayName"));
```
marcphilipp:
```suggestion
		assertThat(testKit.discover().getDiscoveryIssues()) //
				.noneMatch(issue -> issue.message().contains("@SuiteDisplayName"));
```

marcphilipp (approved):

marcphilipp:
> Assuming that an implementation of the `DiscoveryIssueReporter` interface may be added in the future,
> do you think we should also guard against cases where the `createReportingCondition(Predicate<T> predicate, Function<T, DiscoveryIssue> issueCreator)` default method is overridden to return null?

No, since that method is documented to never return `null`, we don't have to guard against such cases. But thanks for being diligent!

marcphilipp:
Thank you for your contribution, @raccoonback! :+1:

injae-kim:
Nice work @raccoonback 🎉🎉🎉

PR #4823, opened 2025-08-09
feat: log invalid search path roots (#4772)

### LINKED ISSUE #4772: Add logging to have console launcher inform about bad/missing searchpath roots
as discussed in https://github.com/junit-team/junit-framework/discussions/4764#discussioncomment-13772600 it would useful if when running console cli to get info/debug log about directories and/or jars that are not available.

Today they are silently ignored and it could easily be missed that tests are not being picked up as expected.

currenjin:
## Summary

Detect and log non-existent, unreadable, or non-dir/jar classpath roots in
`ConsoleLauncher` before discovery.

This change improves diagnostics when using `--scan-class-path` with invalid
entries, addressing issue #4772.

## Motivation

Currently, invalid search path roots are silently ignored, which may cause
users to miss that some tests are not being discovered.  
This PR makes such issues visible by logging them at INFO level, helping users
quickly identify and fix misconfigured paths.

## Changes

- Added `validateAndLogInvalidRoots()` to `DiscoveryRequestCreator`
- Logs invalid roots with the format:  
  `"Ignoring invalid search path root: %s (exists=%s, readable=%s, dirOrJar=%s)"`
- Retains the existing behaviour of skipping invalid roots
- Added a test verifying that an invalid root triggers the log entry
- Updated User Guide to mention this logging behaviour

## Before / After Example

**Before** (invalid path silently ignored):
``` shell
0 tests found
```

**After**:
```shell
Ignoring invalid search path root: /does/not/exist (exists=false, readable=false, dirOrJar=false)
0 tests found
```

## Related Issue
Closes #4772

---

---

### Definition of Done

### DIFF
--- a/junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
@@ -24,7 +24,9 @@ import static org.junit.platform.launcher.TagFilter.excludeTags;
 import static org.junit.platform.launcher.TagFilter.includeTags;
 import static org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder.request;
 
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
@@ -32,6 +34,8 @@ import java.util.Set;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
+import org.junit.platform.commons.logging.Logger;
+import org.junit.platform.commons.logging.LoggerFactory;
 import org.junit.platform.commons.util.ModuleUtils;
 import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.commons.util.ReflectionUtils;
@@ -48,6 +52,8 @@ import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
  * @since 1.0
  */
 class DiscoveryRequestCreator {
+
+	private static final Logger logger = LoggerFactory.getLogger(DiscoveryRequestCreator.class);
 
 	static LauncherDiscoveryRequestBuilder toDiscoveryRequestBuilder(TestDiscoveryOptions options) {
 		LauncherDiscoveryRequestBuilder requestBuilder = request();
@@ -87,9 +93,37 @@ class DiscoveryRequestCreator {
 		if (selectedClasspathEntries.isEmpty()) {
 			Set<Path> rootDirs = new LinkedHashSet<>(ReflectionUtils.getAllClasspathRootDirectories());
 			rootDirs.addAll(options.getExistingAdditionalClasspathEntries());
-			return rootDirs;
+			return validateAndLogInvalidRoots(rootDirs);
 		}
-		return new LinkedHashSet<>(selectedClasspathEntries);
+		return validateAndLogInvalidRoots(new LinkedHashSet<>(selectedClasspathEntries));
+	}
+
+	private static Set<Path> validateAndLogInvalidRoots(Set<Path> roots) {
+		LinkedHashSet<Path> valid = new LinkedHashSet<>();
+		HashSet<Path> seen = new HashSet<>();
+
+		for (Path root : roots) {
+			if (!seen.add(root))
+				continue;
+
+			boolean exists = Files.exists(root);
+			boolean readable = Files.isReadable(root);
+			boolean dirOrJar = isDirOrJar(root);
+
+			if (!exists || !readable || !dirOrJar) {
+				logger.info(
+					() -> "Ignoring invalid search path root: %s (exists=%s, readable=%s, dirOrJar=%s)".formatted(root,
+						exists, readable, dirOrJar));
+				continue;
+			}
+			valid.add(root);
+		}
+
+		return valid;
+	}
+
+	private static boolean isDirOrJar(Path root) {
+		return Files.isDirectory(root) || root.toString().endsWith(".jar");
 	}
 
 	private static void addFilters(LauncherDiscoveryRequestBuilder requestBuilder, TestDiscoveryOptions options,
--- a/platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
@@ -28,6 +28,7 @@ import static org.junit.platform.engine.discovery.DiscoverySelectors.selectUri;
 import java.io.File;
 import java.net.URI;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -35,6 +36,8 @@ import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Test;
 import org.junit.platform.commons.PreconditionViolationException;
+import org.junit.platform.commons.logging.LogRecordListener;
+import org.junit.platform.commons.logging.LoggerFactory;
 import org.junit.platform.console.options.TestDiscoveryOptions;
 import org.junit.platform.engine.Filter;
 import org.junit.platform.engine.UniqueId;
@@ -370,6 +373,27 @@ class DiscoveryRequestCreatorTests {
 		assertThat(configurationParameters.get("foo")).contains("bar");
 		assertThat(configurationParameters.get("com.example.prop.first")).contains("baz");
 		assertThat(configurationParameters.get("com.example.prop.second")).contains("second value");
+	}
+
+	@Test
+	void logs_when_invalid_search_path_present() {
+		LogRecordListener listener = new LogRecordListener();
+		LoggerFactory.addListener(listener);
+		try {
+			var opts = new TestDiscoveryOptions();
+			opts.setScanClasspath(true);
+			opts.setSelectedClasspathEntries(List.of(Paths.get("/does/not/exist")));
+
+			DiscoveryRequestCreator.toDiscoveryRequestBuilder(opts);
+
+			boolean saw = listener.stream(DiscoveryRequestCreator.class).anyMatch(
+				r -> String.valueOf(r.getMessage()).contains("/does/not/exist"));
+
+			assertThat(saw).as("should log about invalid search path root").isTrue();
+		}
+		finally {
+			LoggerFactory.removeListener(listener);
+		}
 	}
 
 	private LauncherDiscoveryRequest convert() {


sormuras:
I note that `javac` emits the following messages for invalid entries in search paths:
```
warning: [path] bad path element "${PATH}": no such file or directory
```

currenjin:
@sormuras 
Yes, exactly. This change makes the ConsoleLauncher behave more like `javac` in that it logs invalid classpath roots instead of silently ignoring them. While `javac` emits a warning to stderr, JUnit logs them at INFO level so that users can see misconfigurations during discovery.

maxandersen:
Love it. 

But should this not be WARN's ? Or as pointed out for javac - stderr message ?

currenjin:
@maxandersen 

Good point. Updated the log level from INFO to WARN so that users can better notice invalid search path roots.

currenjin:
@maxandersen 

Thanks for the feedback! I switched the level to `WARN`, but that surfaced an issue in `platform-tooling-support-tests` (`StandaloneTests.execute*`) where console output expectations are strict.

To keep the console output stable while still alerting users, would you prefer:

1. WARN only for explicitly provided roots (e.g., via `--scan-class-path <paths>` or `--classpath`), while system-derived roots are filtered silently (or at `DEBUG`), or

2. WARN for all invalid roots, but accept extra warnings on stderr and update the expected output accordingly?

I can implement (1) right away; it avoids changing the console output in common scenarios and keeps the tests green. If you’d rather go with (2), I’ll update the tests and docs.

Also, do we want to mention this behavior in the User Guide (ConsoleLauncher section)?

currenjin:
@maxandersen 

Found the root cause of the test failure.

I was running the build in Korea (ko_KR locale), and the execute-related tests did not explicitly set the language.
This caused JUL log levels in the output to be localized (INFO → 정보), so the expected-err.txt assertions failed.

I fixed this by forcing -Duser.language=en and -Duser.country=US in the relevant tests.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
```diff
@@ -87,9 +93,37 @@ private static Set<Path> determineClasspathRoots(TestDiscoveryOptions options) {
 		if (selectedClasspathEntries.isEmpty()) {
 			Set<Path> rootDirs = new LinkedHashSet<>(ReflectionUtils.getAllClasspathRootDirectories());
 			rootDirs.addAll(options.getExistingAdditionalClasspathEntries());
```
marcphilipp: This method is already removing non-existing paths, before they are validated, isn't it?

platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
```diff
@@ -372,6 +375,27 @@ void convertsConfigurationParametersResources() {
...
 	}
 
+	@Test
+	void logs_when_invalid_search_path_present() {
+		LogRecordListener listener = new LogRecordListener();
+		LoggerFactory.addListener(listener);
+		try {
+			var opts = new TestDiscoveryOptions();
+			opts.setScanClasspath(true);
+			opts.setSelectedClasspathEntries(List.of(Paths.get("/does/not/exist")));
```
marcphilipp: We need another test that calls `setAdditionalClasspathEntries` instead.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
```diff
@@ -87,9 +93,37 @@ private static Set<Path> determineClasspathRoots(TestDiscoveryOptions options) {
...
+
+		for (Path root : roots) {
+			if (!seen.add(root))
+				continue;
+
+			boolean exists = Files.exists(root);
+			boolean readable = Files.isReadable(root);
+			boolean dirOrJar = isDirOrJar(root);
+
+			if (!exists || !readable || !dirOrJar) {
```
marcphilipp: Why have you added checks for being readable and being a dir/jar?

platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
```diff
@@ -372,6 +375,27 @@ void convertsConfigurationParametersResources() {
...
+			DiscoveryRequestCreator.toDiscoveryRequestBuilder(opts);
+
+			boolean saw = listener.stream(DiscoveryRequestCreator.class).anyMatch(
+				r -> String.valueOf(r.getMessage()).contains("/does/not/exist"));
+
+			assertThat(saw).as("should log about invalid search path root").isTrue();
+		}
+		finally {
+			LoggerFactory.removeListener(listener);
+		}
```
marcphilipp:
```suggestion
	void logsInvalidSearchPathRoots(@TrackLogRecords LogRecordListener listener) {
		var opts = new TestDiscoveryOptions();
		opts.setScanClasspath(true);
		opts.setSelectedClasspathEntries(List.of(Paths.get("/does/not/exist")));

		DiscoveryRequestCreator.toDiscoveryRequestBuilder(opts);

		assertThat(listener.stream(DiscoveryRequestCreator.class)) //
				.map(LogRecord::getMessage) //
				.filteredOn(message -> message.contains("/does/not/exist")) //
				.hasSize(1);
```

marcphilipp (changes requested):

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
currenjin: You’re right — for `selectedClasspathEntries`, non-existing paths are already filtered out earlier, so the additional checks I added here are redundant. I’ll drop those and keep validation focused on what isn’t handled upstream. I’ll also add a separate test to cover `setAdditionalClasspathEntries(...)`, which wasn’t covered before.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
currenjin: That was an overreach on my side. Since non-existing entries are already removed and the downstream code only needs valid roots, those extra checks are unnecessary here. I’ll remove them to avoid duplication and keep the method focused.

platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
currenjin: Thanks! I’ve refactored the test to use `@TrackLogRecords`, asserted on the `LogRecord` stream directly, and cleaned up the manual listener management.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
marcphilipp: I think we should instead delete the `getExistingAdditionalClasspathEntries()` method and call `getAdditionalClasspathEntries()` which would return the unfiltered list so the checks you added become relevant.

platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
```diff
@@ -372,6 +375,27 @@ void convertsConfigurationParametersResources() {
 		assertThat(configurationParameters.get("com.example.prop.second")).contains("second value");
 	}
 
+	@Test
+	void logs_when_invalid_search_path_present() {
```
marcphilipp: Please use camel case for test methods names.

currenjin: Thanks, I’ve renamed the test methods to follow camel case naming.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
currenjin: Applied your suggestion — switched to `getAdditionalClasspathEntries()` so that 
the existence checks are now relevant. Tests updated accordingly.

currenjin:
@marcphilipp Thanks! All suggested changes have been applied — please take a look when you can.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
```diff
@@ -86,10 +92,30 @@ private static Set<Path> determineClasspathRoots(TestDiscoveryOptions options) {
...
+			if (!seen.add(root))
+				continue;
+
+			boolean exists = Files.exists(root);
+
+			if (!exists) {
+				logger.warn(() -> "Ignoring invalid search path root: %s (exists=%s)".formatted(root, exists));
+				continue;
+			}
+			valid.add(root);
```
marcphilipp:
```suggestion
			if (!seen.add(root)) {
				continue;
			}

			if (Files.exists(root)) {
				valid.add(root);
			} else {
				logger.warn(() -> "Ignoring non-existing classpath root: %s".formatted(root));
			}
```

platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
```diff
@@ -372,6 +376,32 @@ void convertsConfigurationParametersResources() {
...
+	@Test
+	void logsInvalidAdditionalClasspathRoots(@TrackLogRecords LogRecordListener listener) {
+		var opts = new TestDiscoveryOptions();
+		opts.setScanClasspath(true);
+		opts.setAdditionalClasspathEntries(List.of(Paths.get("/also/does/not/exist")));
+
+		DiscoveryRequestCreator.toDiscoveryRequestBuilder(opts);
+
+		assertThat(listener.stream(DiscoveryRequestCreator.class)).map(LogRecord::getMessage).filteredOn(
+			msg -> msg.contains("/also/does/not/exist")).hasSize(1);
```
marcphilipp:
```suggestion
		assertThat(listener.stream(DiscoveryRequestCreator.class)) //
				.map(LogRecord::getMessage) //
				.filteredOn(message -> message.contains("/also/does/not/exist")) //
				.hasSize(1);
```

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
```diff
@@ -86,10 +92,30 @@ private static Set<Path> determineClasspathRoots(TestDiscoveryOptions options) {
 			() -> "No classpath entries selected");
 		if (selectedClasspathEntries.isEmpty()) {
 			Set<Path> rootDirs = new LinkedHashSet<>(ReflectionUtils.getAllClasspathRootDirectories());
-			rootDirs.addAll(options.getExistingAdditionalClasspathEntries());
-			return rootDirs;
+			rootDirs.addAll(options.getAdditionalClasspathEntries());
+			return validateAndLogInvalidRoots(rootDirs);
+		}
+		return validateAndLogInvalidRoots(new LinkedHashSet<>(selectedClasspathEntries));
```
marcphilipp: I think it would be slightly better to call `validateAndLogInvalidRoots` from `createClasspathRootSelectors` so it only needs to be called once and it's clear that it's called in all cases.

platform-tests/src/test/java/org/junit/platform/console/tasks/DiscoveryRequestCreatorTests.java
currenjin: Updated the assertion formatting in `DiscoveryRequestCreatorTests` to match the suggested style for readability.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
currenjin: Applied your suggestion — merged the duplicate `seen.add(root)` checks and simplified the existence validation as you proposed. This removes redundancy while keeping the behavior unchanged.

junit-platform-console/src/main/java/org/junit/platform/console/tasks/DiscoveryRequestCreator.java
currenjin: Good point. I moved the validation to `createClasspathRootSelectors(...)` so it’s applied in a single place for both branches. `determineClasspathRoots(...)` now only collects roots without side effects. Tests still pass, and the logging behavior remains unchanged. Please take another look.

currenjin:
Added a new entry in `release-notes-6.0.0-RC1.adoc` under JUnit Platform > New Features and Improvements documenting the improved validation of additional classpath roots.

marcphilipp:
Thank you for you contribution, @currenjin! :bow:

PR #4838, opened 2025-08-15
Fix potential security issues in GitHub actions

juliette-derancourt:
I ran [zizmor](https://docs.zizmor.sh/) static analysis tool as suggested in https://github.com/junit-team/junit-framework/issues/4558 and found several warnings/errors.

Some of them are addressed in this PR:

- https://docs.zizmor.sh/audits/#artipacked
- https://docs.zizmor.sh/audits/#template-injection

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -124,41 +124,8 @@ jobs:
         with:
           result-encoding: string
           script: |
-            const releaseVersion = process.env.RELEASE_VERSION;
-            const query = `
-                query ($owner: String!, $repo: String!, $title: String!) {
-                    repository(owner: $owner, name: $repo) {
-                        milestones(first: 100, query: $title) {
-                            nodes {
-                                title
-                                number
-                                openIssueCount
-                            }
-                        }
-                    }
-                }
-            `;
-            const {repository} = await github.graphql(query, {
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                title: releaseVersion
-            });
-            const [milestone] = repository.milestones.nodes.filter(it => it.title === releaseVersion)
-            if (!milestone) {
-                throw new Error(`Milestone "${releaseVersion}" not found`);
-            }
-            if (milestone.openIssueCount > 0) {
-                throw new Error(`Milestone "${releaseVersion}" has ${milestone.openIssueCount} open issue(s)`);
-            }
-            const requestBody = {
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                milestone_number: milestone.number,
-                state: 'closed',
-                due_on: new Date().toISOString()
-            };
-            console.log(requestBody);
-            await github.rest.issues.updateMilestone(requestBody);
+            const closeGithubMilestone = require('./.github/scripts/close-github-milestone.js');
+            closeGithubMilestone({ github, context });
 
   publish_deployment:
     name: Publish to Maven Central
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -187,6 +193,7 @@ jobs:
         with:
           fetch-depth: 1
           ref: "refs/tags/${{ env.RELEASE_TAG }}"
+          persist-credentials: true
```
marcphilipp: Why are we persisting them here?

marcphilipp (commented):
Thanks for taking care of this! :+1: 

Is there a way of running this check continuously, e.g. as a GitHub Actions workflow?

juliette-derancourt:
[This usage](https://github.com/junit-team/junit-framework/blob/main/.github/actions/run-gradle/action.yml#L33) is also found as an error, but given the nature of the input (an optional string with `\` characters), it does not work with this "naive" replacement:

```yml
      env:
        ARGUMENTS: ${{ inputs.arguments }}
      run: |
        ./gradlew \
        ... \
        ${ARGUMENTS}
```

Since the input value comes from a workflow, I think we can leave it as it is?

.github/workflows/release.yml
juliette-derancourt: I left it because the job was using some git commands later, but I was actually wondering if it's necessary. You think it can be set to false as well?

juliette-derancourt:
> Is there a way of running this check continuously, e.g. as a GitHub Actions workflow?

@marcphilipp I was wondering about that as well, but I guess we would need some mechanism to say "this is ignored intentionally" for that to be possible (I haven't checked that yet, will do)

.github/workflows/main.yml
```diff
@@ -108,6 +112,7 @@ jobs:
       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       with:
         fetch-depth: 1
+        persist-credentials: true
```
marcphilipp: We shouldn't need them here because we inject Git credentials to the Gradle build below.

```suggestion
        persist-credentials: false
```

.github/workflows/release.yml
```diff
@@ -244,33 +253,34 @@ jobs:
           token: ${{ secrets.JUNIT_BUILDS_GITHUB_TOKEN_EXAMPLES_REPO }}
           fetch-depth: 1
           ref: develop/6.x
+          persist-credentials: true
```
marcphilipp, line 223: Looks like we do need them here because we're calling `git push` below without injecting any credentials.

.github/workflows/release.yml
marcphilipp: Since we're injecting Git credentials to the build step that pushes I don't think we need them here.

```suggestion
          persist-credentials: false
```

juliette-derancourt: I removed it 👍
Do you think we could/should trigger a dry-mode release workflow run to verify everything still works as expected? (I'd like to double check the changes in the JS scripts, even though it should work according to the docs)
Speaking of this, would you mind if I moved those scripts in real javascript files, to get IDE support when editing them?

juliette-derancourt:
> I guess we would need some mechanism to say "this is ignored intentionally" for that to be possible

Seems like [there is a way](https://docs.zizmor.sh/configuration/), and a github action as well. I'll add that to https://github.com/junit-team/junit-framework/issues/4558 deliverables and set that up in another PR.

.github/workflows/release.yml
marcphilipp: > Do you think we could/should trigger a dry-mode release workflow run to verify everything still works as expected? (I'd like to double check the changes in the JS scripts, even though it should work according to the docs)

I think it's probably not worth it. If it fails during the 6.0.0-RC1 release, we'll just have to fix it on the fly.

> Speaking of this, would you mind if I moved those scripts in real javascript files, to get IDE support when editing them?

I think that's a good idea! :+1:

juliette-derancourt: Done in [`8216b93` (#4838)](https://github.com/junit-team/junit-framework/pull/4838/commits/8216b93120e267417efe6a5fadacdd39468065b7)

PR #4841, opened 2025-08-16
Cancel in-progress GitHub Actions runs when PRs are updated

juliette-derancourt:
See examples in GitHub documentation [here](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-using-a-fallback-value) and [here](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -13,6 +13,11 @@ on:
   schedule:
     - cron: '0 19 * * 3'
 
+concurrency:
+  # Cancels in-progress runs only for pull requests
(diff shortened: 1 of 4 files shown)


juliette-derancourt:
Seems to work as intended (see [this run](https://github.com/junit-team/junit-framework/actions/runs/17009810456?pr=4841) that was cancelled)

marcphilipp (approved):

Issue #4873, opened 2025-08-26
Introduce `selectClasspathResources(Set)` in `DiscoverySelectors` and deprecate `selectClasspathResource(Set)`


sbrannen:
## Overview

While working on #4868, I realized that we accidentally left of the "s" in the name of `selectClasspathResource(Set)` in `DiscoverySelectors`.

It should rather be called `selectClasspathResources(Set)` (plural), analogous to `selectClasspathRoots()` and `selectModules()`.

Since `selectClasspathResource(Set)` was introduced in JUnit Platform 1.12, we need to leave it in place for the time being but in a `@Deprecated` state.

## Related Issues

- #4868

sbrannen:
I looked into this in greater detail and noticed that the semantics of `selectClasspathResource(Set)` are rather "special".

`selectClasspathResource(Set)` selects from a `Set` but returns a single `ClasspathResourceSelector`.

Whereas, other "plural" selector methods in `DiscoverySelectors` return a `List` of selectors.

`ClasspathResourceSelector` was recently retrofitted to provide a `Set<Resource> getClasspathResources()` method. So, you still end up with potentially multiple resources, but that is in strict contrast to how other `DiscoverySelector` implementations work.

In other words, `selectClasspathResource(Set)` selects from multiple things (a set) and returns a single thing (a `ClasspathResourceSelector`) that can in turn be queried for multiple things (`getClasspathResources()`).

Compared to how selection typically/historically works, I would have rather expected something like the following in `DicoverySelectors`.

```java
@API(status = EXPERIMENTAL, since = "6.0")
public static List<ClasspathResourceSelector> selectClasspathResources(Resource... classpathResources) {
	Preconditions.notEmpty(classpathResources, "classpath resources must not be null or empty");
	return selectClasspathResources(List.of(classpathResources));
}

@API(status = EXPERIMENTAL, since = "6.0")
public static List<ClasspathResourceSelector> selectClasspathResources(List<Resource> classpathResources) {
	Preconditions.notEmpty(classpathResources, "classpath resources must not be null or empty");
	Preconditions.containsNoNullElements(classpathResources, "individual classpath resources must not be null");

	// @formatter:off
	return classpathResources.stream()
			.distinct()
			.peek(resource -> Preconditions.notBlank(resource.getName(),
				"classpath resource names must not be null or blank"))
			.map(ClasspathResourceSelector::new)
			.toList();
	// @formatter:on
}
```

[...]

PR #4891, opened 2025-08-29
Relocate `Resource` to `org.junit.platform.commons.io` package

### LINKED ISSUE #4885: Relocate `Resource` to `commons.io` package and deprecate existing interface
## Overview

As I mentioned in #4876, `org.junit.platform.commons.support.Resource` causes cycles between the `support`, `scanning`, and `util` packages in `junit-platform-commons`.

Since it's rather likely that we will have the need to support additional I/O types at the platform level in the future, it's probably best that we relocate `Resource` to a new `org.junit.platform.commons.io` package and deprecate the existing interface.

Specifically, I am proposing the following.

1. Introduce `org.junit.platform.commons.io.Resource` as a copy of `org.junit.platform.commons.support.Resource`.
2. Have `org.junit.platform.commons.support.Resource` extend `org.junit.platform.commons.io.Resource`.
3. Deprecate `org.junit.platform.commons.support.Resource` for removal.
4. Update all APIs that reference `org.junit.platform.commons.support.Resource` to reference `org.junit.platform.commons.io.Resource` instead.

Since `org.junit.platform.commons.support.Resource` is a _relatively_ new API, we assume that it is not actively used by many projects yet. In light of that, we are considering making these breaking changes for JUnit 6.0.

However, we hope that affected users will only need to change their imports from `org.junit.platform.commons.support.Resource` to `org.junit.platform.commons.io.Resource` to address any issues.

## Related Issues

- #4876

marcphilipp:
This PR introduces a new `Resource` interface in the new `org.junit.platform.commons.io` package.

The old interface extends the new one so it can be used in places that now consume the new interface.

The resource finding/scanning methods in `ReflectionSupport` are now deprecated in favor of the new ones in the `ResourceSupport` class.

Since `Predicate<...io.Resource>` and `Predicate<...support.Resource>` have the same erasure, this PR introduces `ResourceFilter` (mimicking the existing `ClassFilter`) in order to be able to overload methods such as `EngineDiscoveryRequestResolver.Builder.addResourceContainerSelectorResolver`. 

Resolves #4885.

---

---

### Definition of Done

### DIFF
--- a/platform-tooling-support-tests/src/archUnit/java/platform/tooling/support/tests/ArchUnitTests.java
+++ b/platform-tooling-support-tests/src/archUnit/java/platform/tooling/support/tests/ArchUnitTests.java
@@ -151,8 +151,8 @@ class ArchUnitTests {
 		slices().matching("org.junit.(*)..").should().beFreeOfCycles().check(classes);
 	}
 
-	@SuppressWarnings("removal")
 	@ArchTest
+	@SuppressWarnings("removal")
 	void freeOfPackageCycles(JavaClasses classes) throws Exception {
 		slices().matching("org.junit.(**)").should().beFreeOfCycles() //
 


junit-platform-commons/src/main/java/org/junit/platform/commons/support/scanning/ClasspathScanner.java
marcphilipp, line 1: The changes to this interface are a deliberate breaking change. So far, we are not aware of concrete implementations that would justify a gentler migration path.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -362,9 +363,48 @@ public static ClasspathResourceSelector selectClasspathResource(String classpath
...
+	 * {@code null} or empty. All resources must have the same name, may not
+	 * be {@code null} or blank.
+	 * @since 6.0
+	 * @see #selectClasspathResource(String, FilePosition)
+	 * @see #selectClasspathResource(String)
+	 * @see ClasspathResourceSelector
+	 * @see org.junit.platform.commons.support.ResourceSupport#tryToGetResources(String)
+	 */
+	@API(status = MAINTAINED, since = "6.0")
+	public static ClasspathResourceSelector selectClasspathResourceByName(Set<? extends Resource> classpathResources) {
```
marcphilipp, line 407: I'm not really happy with this name. Maybe we can remove this method in the context of #4873 and add `selectClasspathResources(List<Resource>)` instead that would group by name, create a `ClasspathResourceSelector` for each group, and return a `List<ClasspathResourceSelector>`.

marcphilipp: Draft PR: #4915

junit-platform-commons/src/main/java/org/junit/platform/commons/support/scanning/ClasspathScanner.java
marcphilipp: I think we should delete the old methods in a separate PR so this can be backported to 5.x without breaking compatibility.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/scanning/ClasspathScanner.java
```diff
@@ -62,6 +64,55 @@ public interface ClasspathScanner {
...
+	 * {@link #scanForResourcesInPackage(String, ResourceFilter)} instead
+	 */
+	@API(status = DEPRECATED, since = "6.0")
+	@Deprecated(since = "6.0", forRemoval = true)
+	@SuppressWarnings({ "removal", "unused" })
+	default List<org.junit.platform.commons.support.Resource> scanForResourcesInPackage(String basePackageName,
+			Predicate<org.junit.platform.commons.support.Resource> resourceFilter) {
+		throw new UnsupportedOperationException("Implement scanForResourcesInPackage(String, ResourceFilter) instead");
+	}
+
```
marcphilipp, line 90: I've kept these methods for backwards compatibility with existing `ClasspathScanner` implementations. They will be removed in a separate PR: #4918.

sormuras (approved):
Looks okay - and it's always good to see "ignore" directives vanish from ArchUnit tests.

junit-platform-commons/src/main/java/org/junit/platform/commons/io/DefaultResource.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jspecify.annotations.Nullable;
+import org.junit.platform.commons.PreconditionViolationException;
+import org.junit.platform.commons.annotation.Contract;
+
+/**
+ * Default implementation of {@link Resource}.
+ *
+ * @since 6.0
+ */
+record DefaultResource(String name, URI uri) implements Resource {
```
marcphilipp, line 25: Implement `equals()` and `hashCode()` so both `Resource` implementations are logically equal if they have the same `name` and `uri`.

junit-platform-commons/src/main/java/org/junit/platform/commons/support/scanning/ClasspathScanner.java
```diff
@@ -90,7 +147,19 @@ public interface ClasspathScanner {
...
+	@SuppressWarnings("removal")
+	default List<Resource> scanForResourcesInClasspathRoot(URI root, ResourceFilter resourceFilter) {
+		return scanForResourcesInClasspathRoot(root, r -> resourceFilter.match(oldToNew(r))).stream() //
+				.map(ClasspathScanner::oldToNew) //
+				.toList();
+	}
+
+	@SuppressWarnings("removal")
+	private static Resource oldToNew(org.junit.platform.commons.support.Resource oldResource) {
+		return Resource.of(oldResource.getName(), oldResource.getUri());
```
marcphilipp:
```suggestion
		return oldResource;
```

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
marcphilipp: For now, we decided against #4915 and to go with this name.

junit-platform-commons/src/main/java/org/junit/platform/commons/io/DefaultResource.java
marcphilipp: Done in ebc6a85854483e2250b1a37aea631be37c45c003

junit-platform-commons/src/main/java/org/junit/platform/commons/io/DefaultResource.java
```diff
@@ -0,0 +1,67 @@
...
+	public boolean equals(Object obj) {
+		if (obj == this) {
+			return true;
+		}
+		if (obj instanceof org.junit.platform.commons.io.Resource that) {
+			return name.equals(that.getName()) //
+					&& uri.equals(that.getUri());
+		}
+		return false;
+	}
```
sbrannen:
```suggestion
	public boolean equals(Object obj) {
		return obj == this || (obj instanceof org.junit.platform.commons.io.Resource that
				&& this.name.equals(that.getName()) && this.uri.equals(that.getUri()));
	}
```

I tend to prefer the above "this and that" pattern when using `instanceof` pattern matching.

documentation/src/docs/asciidoc/release-notes/release-notes-6.0.0-RC3.adoc
```diff
@@ -32,12 +32,36 @@ JUnit repository on GitHub.
...
+  - `scanForResourcesInPackage(String, Predicate)`
+  - `scanForResourcesInClasspathRoot(URI, Predicate)`
 
 [[release-notes-6.0.0-RC3-junit-platform-new-features-and-improvements]]
 ==== New Features and Improvements
 
-* New `Resource.from(String, URI)` static factory method for creating an
-  `org.junit.platform.commons.support.Resource`.
+* New classpath resource abstraction in `org.junit.platform.commons.io.Resource` with
+  support class for loading resources or finding them on the classpath via static utility
```
sbrannen:
```suggestion
  support for loading resources or finding them on the classpath via static utility
```

junit-platform-commons/src/main/java/org/junit/platform/commons/io/DefaultResource.java
marcphilipp: I also prefer making `this` explicit but somehow forgot to do so here. Added now in 839b8eeebe700d4da013dc82261af3f3530438ea. I think the longer `if` version is more readable so I kept that.

sbrannen: Fair enough, and thanks for adding `this.`.

As for the `equals()` implementation _style_, that's of course a matter of taste, which we can debate offline. 😉

Issue #4882, opened 2025-08-28
Consider removing `TestIdentifier.SerializedForm`


sbrannen:
## Overview

With the release of JUnit 6.0, we may be able to safely remove `org.junit.platform.launcher.TestIdentifier.SerializedForm`.

## Related Issues

- #2460
- #3819

mehulimukherjee:
Hi @sbrannen , I am really interested in working on this issue, could you please assign this to me?
Thank you,
Mehuli

marcphilipp:
The idea is to remove `TestIdentifier.SerializedForm` along with the custom `writeObject` and `readObject` methods in `TestIdentifier` and increment `TestIdentifier.serialVersionUID`.

@mehulimukherjee I've assigned the issue to you now. Sorry for the delay! Since we want to relase 6.0.0-RC3 in the next few days and want to include this change, it's rather time-critical. If you don't have time to work on this in the next few days, I can take it on and we can find a different issue for you to work on.

mehulimukherjee:
Thanks @marcphilipp , for assigning it to me, I will work on it, and create the PR by Monday eod.

PR #4913, opened 2025-09-07
#4882 : Removed the TestIdentifier.SerializedForm

mehulimukherjee:
**This PR removes the old, custom Java serialization support for TestIdentifier. The SerializedForm inner class has been deleted, and TestIdentifier no longer implements Serializable.**

**What changed**

1. Deleted the SerializedForm inner class from TestIdentifier.
2. Removed implements Serializable and related code.
3. Updated tests:
   - Removed old round-trip serialization tests.
   - Added behavior-based tests to check IDs, names, tags, type, and parent linkage.
   - Added a negative test to confirm that trying to serialize a TestIdentifier now fails with NotSerializableException.
   
TestIdentifier is no longer serializable. If you were persisting or sending these objects via Java serialization, you’ll need to switch to using stable identifiers (getUniqueId()) or create your own small DTO with just the fields you need.
Instead of writing out the whole TestIdentifier, do something like this:

```
// Before (JUnit 5.x) - not supported anymore in JUnit 6
// oos.writeObject(testIdentifier);

// After - recommended
String id = testIdentifier.getUniqueId();
// send/store this string instead
```

**Open questions:**
1. Should I add a short note in the Javadoc for TestIdentifier saying “not serializable”?

---

---

### Definition of Done

### DIFF
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestIdentifier.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestIdentifier.java
@@ -17,13 +17,6 @@ import static org.apiguardian.api.API.Status.INTERNAL;
 import static org.apiguardian.api.API.Status.STABLE;
 import static org.junit.platform.commons.util.CollectionUtils.getOnlyElement;
 
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
-import java.io.ObjectStreamClass;
-import java.io.ObjectStreamField;
-import java.io.Serial;
-import java.io.Serializable;
 import java.util.LinkedHashSet;
 import java.util.Objects;
 import java.util.Optional;
@@ -47,14 +40,7 @@ import org.junit.platform.engine.UniqueId;
  * @see TestPlan
  */
 @API(status = STABLE, since = "1.0")
-public final class TestIdentifier implements Serializable {
-
-	@Serial
-	private static final long serialVersionUID = 1L;
-	@Serial
-	@SuppressWarnings("UnusedVariable")
-	private static final ObjectStreamField[] serialPersistentFields = ObjectStreamClass.lookup(
-		SerializedForm.class).getFields();
+public final class TestIdentifier {
 
 	// These are effectively final but not technically due to late initialization when deserializing
 	private /* final */ UniqueId uniqueId;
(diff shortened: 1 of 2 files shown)


junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestIdentifier.java
```diff
@@ -47,14 +40,7 @@
...
 @API(status = STABLE, since = "1.0")
-public final class TestIdentifier implements Serializable {
-
-	@Serial
-	private static final long serialVersionUID = 1L;
-	@Serial
-	@SuppressWarnings("UnusedVariable")
-	private static final ObjectStreamField[] serialPersistentFields = ObjectStreamClass.lookup(
-		SerializedForm.class).getFields();
+public final class TestIdentifier {
```
marcphilipp:
```suggestion
public final class TestIdentifier implements Serializable {

	@Serial
	private static final long serialVersionUID = 2L;
```

platform-tests/src/test/java/org/junit/platform/launcher/TestIdentifierTests.java
```diff
@@ -57,34 +60,102 @@ void inheritsTypeFromDescriptor() {
...
+		var engineDescriptor = new EngineDescriptor(UniqueId.forEngine("engine"), "Engine");
+		engineDescriptor.addChild(recreatedDescriptor);
+
+		var rebuiltIdentifier = TestIdentifier.from(recreatedDescriptor);
+		assertDeepEquals(originalIdentifier, rebuiltIdentifier);
 	}
 
 	@Test
-	void initialVersionCanBeDeserialized() throws Exception {
-		try (var inputStream = getClass().getResourceAsStream("/serialized-test-identifier")) {
```
marcphilipp, line 68: I think we can remove this test and the corresponding file (`platform-tests/src/test/resources/serialized-test-identifier`).

marcphilipp (changes requested):

mehulimukherjee:
Hi @marcphilipp ,
Making the changes as you suggested.

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestIdentifier.java
```diff
@@ -50,10 +42,6 @@ public final class TestIdentifier implements Serializable {
 
 	@Serial
 	private static final long serialVersionUID = 1L;
```
marcphilipp: This should be incremented to indicate to Java that the pre-6.0 version of this class and this version are not compatible regarding serialization.

```suggestion
	private static final long serialVersionUID = 2L;
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestIdentifier.java
```diff
@@ -50,10 +42,6 @@ public final class TestIdentifier implements Serializable {
...
-	@Serial
-	@SuppressWarnings("UnusedVariable")
-	private static final ObjectStreamField[] serialPersistentFields = ObjectStreamClass.lookup(
-		SerializedForm.class).getFields();
 
 	// These are effectively final but not technically due to late initialization when deserializing
 	private /* final */ UniqueId uniqueId;
```
marcphilipp: This is no longer the case. Please remove this comment and convert `/* final */` to `final` on all fields.

```suggestion
	private final UniqueId uniqueId;
```

junit-platform-launcher/src/main/java/org/junit/platform/launcher/TestIdentifier.java
```diff
@@ -90,19 +78,11 @@ private TestIdentifier(UniqueId uniqueId, String displayName, @Nullable TestSour
 		this.parentId = parentId;
 		this.displayName = displayName;
 		this.source = source;
-		this.tags = copyOf(tags);
+		this.tags = new LinkedHashSet<>(tags);
```
marcphilipp: Please revert this particular change and restore the `copyOf()` method. Since most descriptors don't have tags it's a non-negligible optimization.

platform-tests/src/test/java/org/junit/platform/launcher/TestIdentifierTests.java
```diff
@@ -63,15 +63,6 @@ void currentVersionCanBeSerializedAndDeserialized() throws Exception {
 		assertDeepEquals(originalIdentifier, deserializedIdentifier);
 	}
```
marcphilipp, line 101: I think we should parameterize this test with the number of tags so serialization is tested with 0, 1, and 2 tags to ensure `copyOf()` actually always returns `Serializable` implementations of `Set`. I can do it as a follow-up, though.

marcphilipp (changes requested):

marcphilipp:
Thank you for your contribution, @mehulimukherjee! :+1:

PR #4915, opened 2025-09-07
Introduce `DiscoverySelectors.selectClasspathResources(List)`

### LINKED ISSUE #4873: Introduce `selectClasspathResources(Set)` in `DiscoverySelectors` and deprecate `selectClasspathResource(Set)`

marcphilipp:
Resolves #4873.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
@@ -10,7 +10,9 @@
 
 package org.junit.platform.engine.discovery;
 
+import static java.util.stream.Collectors.collectingAndThen;
 import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.toCollection;
 import static org.apiguardian.api.API.Status.DEPRECATED;
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.MAINTAINED;
@@ -24,6 +26,7 @@ import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
@@ -376,14 +379,11 @@ public final class DiscoverySelectors {
 
 		Preconditions.notEmpty(classpathResources, "classpath resources must not be null or empty");
 		Preconditions.containsNoNullElements(classpathResources, "individual classpath resources must not be null");
-		List<String> resourceNames = classpathResources.stream().map(Resource::getName).distinct().toList();
-		Preconditions.condition(resourceNames.size() == 1, "all classpath resources must have the same name");
-		Preconditions.notBlank(resourceNames.get(0), "classpath resource names must not be null or blank");
-		return new ClasspathResourceSelector(classpathResources);
+		return new ClasspathResourceSelector(new LinkedHashSet<>(classpathResources));
 	}
 
 	/**
-	 * Create a {@code ClasspathResourceSelector} for the supplied classpath
+	 * Create {@code ClasspathResourceSelectors} for the supplied classpath
 	 * resources.
 	 *
 	 * <p>Since {@linkplain org.junit.platform.engine.TestEngine engines} are not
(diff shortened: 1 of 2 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -404,13 +413,19 @@ public static ClasspathResourceSelector selectClasspathResource(
...
+		Preconditions.notNull(classpathResources, "classpath resources must not be null");
 		Preconditions.containsNoNullElements(classpathResources, "individual classpath resources must not be null");
-		List<String> resourceNames = classpathResources.stream().map(Resource::getName).distinct().toList();
-		Preconditions.condition(resourceNames.size() == 1, "all classpath resources must have the same name");
-		Preconditions.notBlank(resourceNames.get(0), "classpath resource names must not be null or blank");
-		return new ClasspathResourceSelector(classpathResources);
+		Preconditions.condition(classpathResources.stream().map(Resource::getName).allMatch(StringUtils::isNotBlank),
+			"classpath resource names must not be null or blank");
+		return classpathResources.stream() //
+				.collect(groupingBy(Resource::getName)) //
```
mpkorstanje: The map produced by `groupingBy` does not have a guaranteed order, so the returned list won't have a guaranteed order.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -404,13 +413,19 @@ public static ClasspathResourceSelector selectClasspathResource(
...
+				.values() //
+				.stream() //
+				.map(LinkedHashSet::new) //
```
mpkorstanje: This will quietly discard duplicate resources. This was not a problem in the original context as there could be no duplicates, but that assumption does not hold here. And so discarding is probably not intended here either.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
```diff
@@ -404,13 +413,19 @@ public static ClasspathResourceSelector selectClasspathResource(
 	 * @see org.junit.platform.commons.support.ResourceSupport#tryToGetResources(String)
 	 */
 	@API(status = MAINTAINED, since = "6.0")
-	public static ClasspathResourceSelector selectClasspathResourceByName(Set<? extends Resource> classpathResources) {
-		Preconditions.notEmpty(classpathResources, "classpath resources must not be null or empty");
+	public static List<ClasspathResourceSelector> selectClasspathResources(
```
mpkorstanje, line 417: Java doc mentions Set

marcphilipp: Thanks, I've updated that now.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
marcphilipp: I've documented that duplicates are allowed in the Javadoc of the method now. I don't see a problem with discarding duplicates. Do you?

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
marcphilipp: Good catch! Fixed in the latest push.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/DiscoverySelectors.java
mpkorstanje: I think the documentation will do.

Though I'd consider it a programming error. Resources are presumably unique, so if caller manages to discover the same resource multiple times something is wrong. 

But I also see that the semantics around selectors appear to be to ignore later selectors. So if this is okay:

```java
LauncherDiscoveryRequestBuilder.request()
    .selectors(
            selectClass(Test.class),
            selectClass(Test.class)
    );

LauncherDiscoveryRequestBuilder.request()
    .selectors(
            selectClasses(List.of(
                Test.class,
                Test.class
            ))
    );
```

Then this too should be okay.

```java
LauncherDiscoveryRequestBuilder.request()
    .selectors(
            selectClasspathResources(List.of(
                new MyResource("a"),
                new MyResource("a")
            ))
    );
```    

Even though the effect under the hood is a little bit different.

Issue #4886, opened 2025-08-28
Relocate `MediaType` to `jupiter.api` package and deprecate existing class


sbrannen:
## Overview

As can be seen in https://github.com/junit-team/junit-framework/issues/4876#issuecomment-3233342829, `org.junit.jupiter.api.extension.MediaType` causes cycles between the `api` and `api.extension` packages in `junit-jupiter-api`.

Since the `junit-jupiter-api` artifact constitutes the public API for JUnit Jupiter, we have traditionally not allowed package cycles within that artifact.

Thus, in order to reduce package cycles and improve the quality of the code base, we should _relocate_ `MediaType` to the `org.junit.jupiter.api` package and deprecate the existing class.

Specifically, I am proposing the following.

1. Introduce `org.junit.jupiter.api.MediaType` as a copy of `org.junit.jupiter.api.extension.MediaType`.
2. Declare `org.junit.jupiter.api.MediaType` as a sealed type permitting only `org.junit.jupiter.api.extension.MediaType` as a subtype.
3. Have `org.junit.jupiter.api.extension.MediaType` extend `org.junit.jupiter.api.MediaType`.
4. Deprecate `org.junit.jupiter.api.extension.MediaType` for removal.
5. Update all APIs that reference `org.junit.jupiter.api.extension.MediaType` to reference `org.junit.jupiter.api.MediaType` instead.

Since `org.junit.jupiter.api.MediaType` is a _relatively_ new API, we assume that it is not actively used by many projects yet. In light of that, we are considering deprecating the current APIs for JUnit 6.0.

However, we hope that affected users will only need to change their imports from `org.junit.jupiter.api.extension.MediaType` to `org.junit.jupiter.api.MediaType` to address any issues.

## Related Issues

- #4876
- #4885

sbrannen:
### Update

Current work on this can be viewed in the following feature branch.

https://github.com/junit-team/junit-framework/compare/main...sbrannen:junit-framework:issues/4886-jupiter.api.MediaType

Issue #4876, opened 2025-08-27
Detect and remove package cycles within the code base


sbrannen:
## Overview

`Resource` and `DefaultResource` currently reside in the `org.junit.platform.commons.support` package.

However, they are used in `org.junit.platform.commons.util.ReflectionUtils` which is used by `org.junit.platform.commons.support.ReflectionSupport`, resulting in a `support`/`util` package cycle.

They are also used in `org.junit.platform.commons.support.scanning.DefaultClasspathScanner` which in turn uses types in `org.junit.platform.commons.util`, resulting in a `scanning`/`support`/`util` package cycle.

I suppose the `Resource` and `DefaultResource` types should ideally reside in a separate `org.junit.platform.commons.io` package.

In any case, we should decide if and how we would like to address this.

As a side note, I don't think `DefaultResource` needs to be a `public` type. It could rather be package-private and created via a public static factory method.

## Related Issues

- #4879 
- #4885 
- #4886

sbrannen:
I thought we had an ArchUnit test in place to detect cycles, and I was right.

https://github.com/junit-team/junit-framework/blob/3ba3c8bd3360eda9af3f582203a03d43e40e31c2/platform-tooling-support-tests/src/archUnit/java/platform/tooling/support/tests/ArchUnitTests.java#L121-L124

However, it turns out that we configured that incorrectly (see https://github.com/TNG/ArchUnit/issues/106#issuecomment-417868120).

Changing the test to the following...

```java
@ArchTest
void freeOfCycles(JavaClasses classes) {
	slices().matching("org.junit.(**)").should().beFreeOfCycles().check(classes);
}
```

... results in several violations.

<details>

<summary>Architecture Violations</summary>

```
java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'slices matching 'org.junit.(**)' should be free of cycles' was violated (5 times):
Cycle detected: Slice jupiter.api -> 
                Slice jupiter.api.extension -> 
                Slice jupiter.api
  1. Dependencies of Slice jupiter.api
    - Method <org.junit.jupiter.api.TestReporter.publishFile(java.lang.String, org.junit.jupiter.api.extension.MediaType, org.junit.jupiter.api.function.ThrowingConsumer)> has parameter of type <org.junit.jupiter.api.extension.MediaType> in (TestReporter.java:0)
[...]

PR #4916, opened 2025-09-07
Relocate `MediaType` to `jupiter.api` package and deprecate existing class

sbrannen:
## Overview

`org.junit.jupiter.api.extension.MediaType` currently causes cycles between the `api` and `api.extension` packages in `junit-jupiter-api`.

Since the `junit-jupiter-api` artifact constitutes the public API for JUnit Jupiter, we have traditionally not allowed package cycles within that artifact.

Thus, in order to reduce package cycles and improve the quality of the code base, this PR introduces a new `MediaType` in the `org.junit.jupiter.api` package. Note that the existing `MediaType` now extends the new `MediaType`.

To encourage users to migrate to the new supported `MediaType` this commit also introduces the following deprecations.

* `org.junit.jupiter.api.extension.MediaType` is now deprecated in favor of the new `org.junit.jupiter.api.MediaType`.
* The `publishFile(...)` methods in `TestReporter` which accept an `org.junit.jupiter.api.extension.MediaType` are now deprecated in favor of new variants which accept an `org.junit.jupiter.api.MediaType`.
* The `publishFile(...)` method in `ExtensionContext` which accepts an `org.junit.jupiter.api.extension.MediaType` is now deprecated in favor of a new variant which accepts an `org.junit.jupiter.api.MediaType`.

Furthermore, we plan to remove the deprecated `MediaType` in JUnit 6.2.

## Related Issues

- #4876
- #4886


junit-jupiter-api/src/main/java/org/junit/jupiter/api/MediaType.java
```diff
@@ -0,0 +1,161 @@
...
+
+	protected MediaType(String type, String subtype, @Nullable Charset charset) {
+		this("%s/%s%s".formatted(//
+			Preconditions.notBlank(type, "type must not be null or blank").strip(),
+			Preconditions.notBlank(subtype, "subtype must not be null or blank").strip(),
+			(charset != null ? ("; charset=" + charset.name()) : "")));
+	}
+
+	protected MediaType(String value) {
+		// Mimic sealed types, permitting only this class and api.extension.MediaType.
```
marcphilipp, line 128: Why not use an actual `sealed` type? So it can be backported to 5.x?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
```diff
@@ -409,13 +409,42 @@ default void publishReportEntry(String value) {
...
+	 * @param name the name of the file to be attached; never {@code null} or
+	 * blank and must not contain any path separators
+	 * @param mediaType the media type of the file; never {@code null}; use
+	 * {@link org.junit.jupiter.api.MediaType#APPLICATION_OCTET_STREAM} if unknown
+	 * @param action the action to be executed to write the file; never {@code null}
+	 * @since 6.0
+	 * @see org.junit.platform.engine.EngineExecutionListener#fileEntryPublished
+	 */
+	@API(status = MAINTAINED, since = "6.0")
+	void publishFile(String name, org.junit.jupiter.api.MediaType mediaType, ThrowingConsumer<Path> action);
```
marcphilipp: I think we should import `org.junit.jupiter.api.MediaType` to minimize the change set when we delete the old type.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/MediaType.java
marcphilipp: This was due to #4939

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java
sbrannen: I made that change and force pushed.

When I initially introduced that, my IDE complained about the `api.MediaType` import, but I've updated my IDE, and it doesn't complain any more... for whatever that's worth. 🤔

So, thanks for the suggestion. 👍

marcphilipp (approved):
Looks good except for the missing release notes entry.

sbrannen:
> Looks good except for the missing release notes entry.

I just added that and also made a final pass over the change set, polishing a bit along the way.

So, feel free to take another look...

marcphilipp (approved):

PR #4987, opened 2025-09-23
Reintroduce Leyden preview build

marcphilipp:


.github/workflows/cross-version.yml
```diff
@@ -32,6 +32,9 @@ jobs:
           distribution: oracle
         - version: 26
           type: ea
+        - version: 26
+          type: ea
+          release: leyden
```
sormuras:
```suggestion
          release: leydenpremain
```

sormuras: The EA binaries use a different name schema...

PR #5006, opened 2025-09-29
Add `error-prone.picnic.tech` featuring `RedundantStringConversion`

Pankraz76:
---

---

### Definition of Done

Links:

- https://github.com/junit-team/junit-framework/issues/5002#issuecomment-3343498159
- https://github.com/gradle/gradle/blob/dea495b3b841ceaea73c5832ae5bb6b04333b0fc/build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.code-quality.gradle.kts#L148
- https://medium.com/@lfgcampos/error-prone-error-pruning-8f1eb51da5ca
- https://sonarsource.github.io/rspec/#/rspec/S1488

### DIFF
--- a/gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
+++ b/gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
@@ -28,11 +28,11 @@ tasks.withType<JavaCompile>().configureEach {
 		val shouldDisableErrorProne = java.toolchain.implementation.orNull == JvmImplementation.J9
 		if (name == "compileJava" && !shouldDisableErrorProne) {
 			disable(
-				"AnnotateFormatMethod", // We don`t want to use ErrorProne`s annotations.
+				"AnnotateFormatMethod", // We don`t want to use ErrorProne's annotations.
 				"BadImport", // This check is opinionated wrt. which method names it considers unsuitable for import which includes a few of our own methods in `ReflectionUtils` etc.
-				"DoNotCallSuggester",
-				"ImmutableEnumChecker",
-				"InlineMeSuggester",
+				"DoNotCallSuggester", // We don`t want to use ErrorProne's annotations.
+				"ImmutableEnumChecker", // We don`t want to use ErrorProne's annotations.
+				"InlineMeSuggester", // We don`t want to use ErrorProne's annotations.
 				"MissingSummary", // Produces a lot of findings that we consider to be false positives, for example for package-private classes and methods.
 				"StringSplitter", // We don`t want to use Guava.
 				"UnnecessaryLambda", // The findings of this check are subjective because a named constant can be more readable in many cases.


gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,6 +30,13 @@ nullaway {
 
 tasks.withType<JavaCompile>().configureEach {
 	options.errorprone {
+		disableWarningsInGeneratedCode = true
+		allErrorsAsWarnings = true
+		if(getenv("CI").isNullOrEmpty()) {
+			errorproneArgs.addAll(
+				"-XepPatchChecks:UnnecessaryLambda,MissingOverride,DefaultCharset,DeadException",
+				"-XepPatchLocation:IN_PLACE")
```
Pankraz76: Somehow it’s not working — its not replacing or failing the build.

I was wondering how to configure all the rules assuming they’re active by default and the plugin is opinionated.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,6 +29,65 @@ nullaway {
 
 tasks.withType<JavaCompile>().configureEach {
 	options.errorprone {
+		disableWarningsInGeneratedCode = true
+		allErrorsAsWarnings = true
```
marcphilipp: Why did you set these options?

documentation/src/main/java/example/domain/Person.java
```diff
@@ -55,8 +55,8 @@ public int hashCode() {
 		final int prime = 31;
 		int result = 1;
 		result = prime * result + ((firstName == null) ? 0 : firstName.hashCode());
-		result = prime * result + ((lastName == null) ? 0 : lastName.hashCode());
-		return result;
+		return prime * result + ((lastName == null) ? 0 : lastName.hashCode());
```
marcphilipp: TBH the old code looks more readable to me. It's probably a matter of taste, though. Nevertheless, please disable this rule.

Pankraz76: every sonar rule is an individual decision. checkstyle, spotless and PMD apply it.

- https://github.com/checkstyle/checkstyle/pull/14250

Will disable like requested.

Thanks.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: from gradle and from tutorial it seems best practise. 

- https://medium.com/@lfgcampos/error-prone-error-pruning-8f1eb51da5ca
- https://github.com/gradle/gradle/blob/360f9eab2f6f1595025f746a03ee5895659b0b8c/build-logic-commons/gradle-plugin/src/main/kotlin/gradlebuild.code-quality.gradle.kts#L148

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,29 +29,116 @@ nullaway {
...
 			disable(
-
-				// This check is opinionated wrt. which method names it considers unsuitable for import which includes
-				// a few of our own methods in `ReflectionUtils` etc.
-				"BadImport",
-
-				// The findings of this check are subjective because a named constant can be more readable in many cases
-				"UnnecessaryLambda",
-
-				// Resolving findings for these checks requires ErrorProne's annotations which we don't want to use
```
Pankraz76, line 41: Assumed this is for the whole block as the block is some kind of pattern here but its a nice trap, assuming the comment only regards to `AnnotateFormatMethod`.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,29 +29,116 @@ nullaway {
...
-
-				// Resolving findings for this check requires using Guava which we don't want to use
-				"StringSplitter",
-
-				// Produces a lot of findings that we consider to be false positives, for example for package-private
-				// classes and methods
-				"MissingSummary",
+				"BadImport", // This check is opinionated wrt. which method names it considers unsuitable for import which includes a few of our own methods in `ReflectionUtils` etc.
+				"UnnecessaryLambda", // The findings of this check are subjective because a named constant can be more readable in many cases.
+				"AnnotateFormatMethod", // We don't want to use ErrorProne's annotations.
```
Pankraz76: inline this is an clear intent and not an obsolete question.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/PreInterruptThreadDumpPrinter.java
```diff
@@ -49,7 +49,7 @@ public void beforeThreadInterrupt(PreInterruptContext preInterruptContext, Exten
 					sb.append(NL);
 					// Use the same prefix as java.lang.Throwable.printStackTrace(PrintStreamOrWriter)
 					sb.append("\tat ");
-					sb.append(stackTraceElement.toString());
```
Pankraz76, line 52: assuming `RedundantStringEscape` is the fix.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,29 +30,117 @@ nullaway {
 
 tasks.withType<JavaCompile>().configureEach {
 	options.errorprone {
+		allErrorsAsWarnings = true
+		disableWarningsInGeneratedCode = true
+		errorproneArgs.add("-XepOpt:Refaster:NamePattern=^(?!.*Rules\\$).*") // might consider.
+		if (getenv("IN_PLACE").toBoolean()) {
+			errorproneArgs.addAll(
+				"-XepPatchLocation:IN_PLACE", // why only certain picnic rules apply?
```
marcphilipp: How does it know where to find the source files to patch? Does patching actually work? This comment suggests otherwise: https://github.com/tbroyer/gradle-errorprone-plugin/issues/38#issuecomment-1428594046

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,29 +30,117 @@ nullaway {
 
 tasks.withType<JavaCompile>().configureEach {
 	options.errorprone {
+		allErrorsAsWarnings = true
+		disableWarningsInGeneratedCode = true
+		errorproneArgs.add("-XepOpt:Refaster:NamePattern=^(?!.*Rules\\$).*") // might consider.
```
marcphilipp: What does this do?

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,29 +30,117 @@ nullaway {
 
 tasks.withType<JavaCompile>().configureEach {
 	options.errorprone {
+		allErrorsAsWarnings = true
```
marcphilipp: Doesn't this downgrade errors to warnings? If so, why do we need it?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/PreInterruptThreadDumpPrinter.java
marcphilipp: Did you change this class manually or by applying a patch?

Pankraz76: This patch is error-prone — it depends on the library’s behavior instead of individual, opinionated interpretation.
It would probably still pass the build even if reverted, which is exactly what we want to avoid. The setup isn’t really complete yet, and we’re still aiming for the proper rewrite equivalent.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: yes sound so. will undo.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: It compiles, which means it’s quite involved and includes all the necessary information, producing output without any errors. Some of the changes might be critical, so it’s recommending to review the automatically generated modifications — just as rewrite advises — since these are machine suggestions and not guaranteed to be perfect. Double-checking is always necessary.

That’s why this feature is **opt-in**. It’s designed to catch issues, highlight them, and allow to fix the build with a single simple command. Just like check and spot have evolved to.


- https://errorprone.info/docs/patching
- https://github.com/diffplug/spotless/pull/2663
- https://github.com/diffplug/spotless/pull/2656

Pankraz76 (commented):
thanks for the effort spend.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: Ignore all Refaster Rules for now, as we fail many of them and maintaining a whitelist is too costly.

This means we’ll ignore all rules by default and selectively activate specific ones when needed. Since Refaster rules are not applied automatically by the autofix process, it’s not practical to enforce them at scale.

While `rewrite` can apply them as shown, `error-prone` should ideally be able to do the same — so this might just be a configuration or bug issue.

We could fix issues locally using `rewrite` and then activate Picnic, though that’s quite a hack. You mentioned it’s not allowed to be part of the main pipeline, but we could still integrate it without enabling it. This way, I get my `rewrite` integration, and you still maintain the rejection of it since it’s not actively applied in the change.  

In a sense, this creates a kind of technical "detour," but one could argue that `rewrite` acts as a refactoring tool that goes beyond even what `error-prone` can do. At the end, we would benefit from both: `rewrite` can cover Refaster rules from Picnic (but not all bug patterns), while `error-prone` focuses on bug patterns. Each tool has its own benefits. One could also see `rewrite` as a "big brother" running in the cloud, and `error-prone` as the fast, local feedback tool.

This approach could help address `JavaUtilAPIs` changes and Java migrations that `error-prone` doesn’t cover, while still improving the overall codebase and developer experience.

- https://github.com/diffplug/spotless/pull/2660

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/PreInterruptThreadDumpPrinter.java
Pankraz76: works as expected, failing the build on corruption:

meaning on cloud, when if local fixes are active, but not pushed, it will fail in the quality gate. 

still this is kind of wack having to know the trick about the custom env, but okey.

Rewrite directly outputs the command how to apply the suggested/enforced change upfront.

```
11:18:41: Executing 'assemble clean'…

Starting Gradle Daemon...
Gradle Daemon started in 1 s 213 ms
> Task :junit-jupiter-api:javadoc FROM-CACHE

> Task :junit-jupiter-engine:compileJava
/Users/vincent.potucek/IdeaProjects/junit-framework/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/PreInterruptThreadDumpPrinter.java:52: warning: [RedundantStringConversion] Avoid redundant string conversions when possible
					sb.append(stackTraceElement.toString());
					         ^
    (see https://error-prone.picnic.tech/bugpatterns/RedundantStringConversion)
  Did you mean 'sb.append(stackTraceElement);'?

> Task :junit-platform-console:compileJava
> Task :junit-platform-console:classes
> Task :junit-jupiter-api:javadocJar
> Task :junit-jupiter-api:assemble

> Task :junit-jupiter-engine:compileJava FAILED
error: warnings found and -Werror specified
1 error
1 warning

> Task :junit-platform-console:shadowJar

[Incubating] Problems report is available at: file:///Users/vincent.potucek/IdeaProjects/junit-framework/build/reports/problems/problems-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/9.1.0/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
135 actionable tasks: 77 executed, 11 from cache, 47 up-to-date

* What went wrong:
Execution failed for task ':junit-jupiter-engine:compileJava'.
> Compilation failed; see the compiler output below.
  /Users/vincent.potucek/IdeaProjects/junit-framework/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/PreInterruptThreadDumpPrinter.java:52: warning: [RedundantStringConversion] Avoid redundant string conversions when possible
  					sb.append(stackTraceElement.toString());
  					         ^
      (see https://error-prone.picnic.tech/bugpatterns/RedundantStringConversion)
    Did you mean 'sb.append(stackTraceElement);'?
  1 error
  1 warning

BUILD FAILED in 59s
11:19:40: Execution finished 'assemble clean'.
```

junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java
```diff
@@ -155,22 +155,18 @@ public static class FilterOptions {
...
 		private final List<String> excludeClassNamePatterns = new ArrayList<>();
 
-		@Option(names = {
-				"--include-package" }, paramLabel = "PKG", arity = "1", description = "Provide a package to be included in the test run. This option can be repeated.")
+		@Option(names = "--include-package", paramLabel = "PKG", arity = "1", description = "Provide a package to be included in the test run. This option can be repeated.")
 		private final List<String> includePackages = new ArrayList<>();
 
-		@Option(names = {
-				"--exclude-package" }, paramLabel = "PKG", arity = "1", description = "Provide a package to be excluded from the test run. This option can be repeated.")
+		@Option(names = "--exclude-package", paramLabel = "PKG", arity = "1", description = "Provide a package to be excluded from the test run. This option can be repeated.")
```
Pankraz76:
```
Refactoring changes were successfully applied to file:///Users/vincent.potucek/IdeaProjects/junit-framework/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java, please check the refactored code and recompile.
```

```
> Task :junit-platform-console-standalone:classes UP-TO-DATE
/Users/vincent.potucek/IdeaProjects/junit-framework/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java:158: Note: [CanonicalAnnotationSyntax] Omit redundant syntax from annotation declarations
		@Option(names = {
		^
    (see https://error-prone.picnic.tech/bugpatterns/CanonicalAnnotationSyntax)
  Did you mean '@Option(names = "--include-package", paramLabel = "PKG", arity = "1", description = "Provide a package to be included in the test run. This option can be repeated.")'?
/Users/vincent.potucek/IdeaProjects/junit-framework/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java:162: Note: [CanonicalAnnotationSyntax] Omit redundant syntax from annotation declarations
		@Option(names = {
		^
    (see https://error-prone.picnic.tech/bugpatterns/CanonicalAnnotationSyntax)
  Did you mean '@Option(names = "--exclude-package", paramLabel = "PKG", arity = "1", description = "Provide a package to be excluded from the test run. This option can be repeated.")'?
/Users/vincent.potucek/IdeaProjects/junit-framework/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java:166: Note: [CanonicalAnnotationSyntax] Omit redundant syntax from annotation declarations
		@Option(names = {
		^
    (see https://error-prone.picnic.tech/bugpatterns/CanonicalAnnotationSyntax)
  Did you mean '@Option(names = "--include-methodname", paramLabel = "PATTERN", arity = "1", description = "Provide a regular expression to include only methods whose fully qualified names without parameters match. "'?
/Users/vincent.potucek/IdeaProjects/junit-framework/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java:171: Note: [CanonicalAnnotationSyntax] Omit redundant syntax from annotation declarations
		@Option(names = {
		^
    (see https://error-prone.picnic.tech/bugpatterns/CanonicalAnnotationSyntax)
  Did you mean '@Option(names = "--exclude-methodname", paramLabel = "PATTERN", arity = "1", description = "Provide a regular expression to exclude those methods whose fully qualified names without parameters match. "'?
/Users/vincent.potucek/IdeaProjects/junit-framework/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java:220: Note: [CanonicalAnnotationSyntax] Omit redundant syntax from annotation declarations
		@Option(names = {
		^
    (see https://error-prone.picnic.tech/bugpatterns/CanonicalAnnotationSyntax)
  Did you mean '@Option(names = "--config-resource", paramLabel = "PATH", arity = "1", description = "Set configuration parameters for test discovery and execution via a classpath resource. This option can be repeated.")'?
Refactoring changes were successfully applied to file:///Users/vincent.potucek/IdeaProjects/junit-framework/junit-platform-console/src/main/java/org/junit/platform/console/options/TestDiscoveryOptionsMixin.java, please check the refactored code and recompile.
> Task :junit-platform-console:shadowJar
> Task :junit-platform-console:jar SKIPPED
> Task :junit-platform-console-standalone:extractThirdPartyLicenses
> Task :junit-platform-console:javadoc FROM-CACHE
> Task :junit-platform-console:javadocJar
> Task :junit-platform-console:assemble
> Task :junit-platform-console-standalone:shadowJar
> Task :junit-platform-console-standalone:jar SKIPPED
> Task :junit-platform-console-standalone:javadoc NO-SOURCE
> Task :junit-platform-console-standalone:javadocJar
> Task :junit-platform-console-standalone:assemble

[Incubating] Problems report is available at: file:///Users/vincent.potucek/IdeaProjects/junit-framework/build/reports/problems/problems-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/9.1.0/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD SUCCESSFUL in 7s
168 actionable tasks: 81 executed, 33 from cache, 54 up-to-date
Consider enabling configuration cache to speed up this build: https://docs.gradle.org/9.1.0/userguide/configuration_cache_enabling.html
12:32:08: Execution finished 'clean assemble'.
```

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,29 +30,270 @@ nullaway {
...
 tasks.withType<JavaCompile>().configureEach {
 	options.errorprone {
+		disableWarningsInGeneratedCode = true
+		errorproneArgs.add("-XepOpt:Refaster:NamePattern=^(?!.*Rules\\$).*") // might consider.
+		if (getenv("IN_PLACE").toBoolean()) {
+			errorproneArgs.addAll(
+				"-XepPatchLocation:IN_PLACE", // why only certain picnic rules apply?
+				"-XepPatchChecks:" +
+						"AddNullMarkedToPackageInfo," +
+						"AlwaysThrows," +
```
Pankraz76: For sure, we don’t need all of these, but technically or ideally, some definitely apply.

Continuing to do so is the natural outcome of the plugin being consistent—failing the build if something goes wrong—thus automating the process and reinforcing automation.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,29 +30,270 @@ nullaway {
...
+				"FormatStringConcatenation",
+				"IdentityConversion",
+				"LexicographicalAnnotationAttributeListing",
+				"LexicographicalAnnotationListing",
+				"NestedOptionals",
+				"NonStaticImport",
+				"OptionalOrElseGet",
+				"PrimitiveComparison",
+				"StaticImport",
+				"TimeZoneUsage",
```
Pankraz76, line 52: These are failing and cannot be fixed through patching, which is somewhat of a mystery.

Pankraz76 (commented):
ready to go.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
marcphilipp: @Pankraz76 https://github.com/diffplug/spotless/pull/2664#discussion_r2406542994 is in reference to my question, right?

Why does it work in the JUnit build but not in build of Spotless?

Pankraz76: assuming because JUnit codebase is complain, to the default ruleset, while spotless is not (yet).

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: as much as i might like, this feels wrong just by looking at it, undo.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -27,43 +22,49 @@ nullaway {
...
+				"OptionalOrElseGet",
+				"PrimitiveComparison",
+				"TimeZoneUsage",
 			)
-			error("PackageLocation")
-		} else {
-			disableAllChecks = true
+			error(
+				"PackageLocation",
+				"StaticImport"
```
Pankraz76: this is just `suggestion`, but still demand to disable or obey. So this behavior of the plugin is surpricing.

- https://error-prone.picnic.tech/bugpatterns/StaticImport

Pankraz76:
using the `in_place` mode only works with whitelisting which is is not the desired case. It would still fail on ci demanding fixes but thats not a good behavior.

Prone demands the manual fixes, not being an rewrite equivlent therefore a new mode might be needed.

marcphilipp:
@Pankraz76 Could you please summarize the state of this PR? Is it ready to be reviewed? What do the changes mean for local builds and on CI?

Pankraz76:
## Explanation

Sorry for the confusion, and thanks for considering.

This is just a new **error-prone check**, originating from **Picnic**, and it behaves the same as the other rules for all builds.
If the check finds an issue, it will **fail the build** and **show a suggested fix**, just like its currently the case with others.

We can’t use the **optional patching mode (`IN_PLACE`)** because it overrides the rule settings when activated.
While this could work locally, it would require significant overhead — such as maintaining an explicit whitelist to keep the ruleset in sync.
Otherwise, a locally green build might still fail in CI, as some checks aren’t included, imposing a rule delta between local and CI builds, which is not the ideal path.
This would lead to manual intervention, which patching is meant to avoid (similar to what’s done with rewrite).

That means only the specified checks would run — leading to a **successful local build** but a **failed CI build**, since CI does not patch and still enforces all rules for quality control.

In order to have patching, we would either need to 
- adapt **Error Prone** to support this as a new feature or
-  create a whitelist approach similar to **Spotless**.

Both options are **unrealistic**, so we should move forward with the current setup and leave the patching topic aside for now — to stop staring and start finishing.
This can be revisited later, along with considerations like `StaticImport`.

Currently, **Spotless** uses a **fully configured ruleset**, which checks only the configured rules instead of following the convention.
While this allows **auto-fix capability**, it’s not ideal in terms of maintaining consistency with conventions.

Given that the required changes here are minimal, keeping our current setup is the better approach.

related:

- https://github.com/diffplug/spotless/pull/2664/files#r2416889516
- https://github.com/gradle/gradle/pull/35107

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,61 +10,65 @@ plugins {
...
 			)
-			error("PackageLocation")
-		} else {
-			disableAllChecks = true
 		}
 		nullaway {
-			if (shouldDisableErrorProne) {
-				disable()
-			} else {
+			if (J9 == java.toolchain.implementation.orNull) {
```
Pankraz76: imho, now its clear about J9. 

Its better to have a DRYness on this field then coupling on some custom one.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,61 +10,65 @@ plugins {
...
 	}
 }
 
-nullaway {
-	onlyNullMarked = true
-}
-
 tasks.withType<JavaCompile>().configureEach {
 	options.errorprone {
-		val shouldDisableErrorProne = java.toolchain.implementation.orNull == JvmImplementation.J9
```
Pankraz76, line 30: Imho, Its better to have a DRYness on `J9` field then coupling on some custom one.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,61 +10,65 @@ plugins {
...
 				require("33.4.8-jre")
 			}
 			because("Older versions use deprecated methods in sun.misc.Unsafe")
+			// https://github.com/junit-team/junit-framework/pull/5039#discussion_r2414490581
 		}
 	}
 }
 
-nullaway {
-	onlyNullMarked = true
```
Pankraz76, line 25: merged into other declaration assuming its equivalent.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,61 +10,65 @@ plugins {
...
+				"PackageLocation",
+				"RedundantStringConversion",
+				"RedundantStringEscape",
 			)
-			error("PackageLocation")
-		} else {
-			disableAllChecks = true
 		}
 		nullaway {
-			if (shouldDisableErrorProne) {
```
Pankraz76, line 59: happy path makes everybody happy

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
rickie: > Ignore all Refaster Rules for now, as we fail many of them and maintaining a whitelist is too costly.

In that case, would say disabling Refaster is easier, right? 

> This means we’ll ignore all rules by default and selectively activate specific ones when needed. Since Refaster rules are not applied automatically by the autofix process, it’s not practical to enforce them at scale.

Refaster is definitely meant for auto-applying. Refaster is originally from Error Prone, and simply runs as part of Error Prone. (Not important but to be more accurate even: Refaster is actually implemented as an Error Prone check).

I'm not a big expert of Gradle, but my guess is that this is an configuration issue or bug indeed. We use Refaster internally and in Error Prone Support a lot, and it is capable of doing this (we do use Maven, but that shouldn't matter). 

---

Can we get to a minimal reproducible example of where the patching doesn't work? With all the various states the PRs are in, it's hard to identify what precisely doesn't work. 
Perhaps we can get some ideas from examples, e.g. [EPS](https://github.com/PicnicSupermarket/error-prone-support/blob/master/pom.xml#L1941) or this command from Picnic internally: `-Xep:Refaster:ERROR -XepPatchLocation:IN_PLACE -XepOpt:Refaster:NamePattern="<something>" `.

---

If we foresee issues with how we configure which Refaster rules to run and not run, we might be able to consider some improvements on the side of Error Prone Support. We know that part is not great. Though, for now we could have a regular expression, where we can simply keep adding names of Refaster rule collections to the list of the ones we enable.

Pankraz76: this should reproduce the issue:

- https://github.com/Pankraz76/junit5/pull/10

documentation/src/main/java/example/domain/Person.java
rickie: Note the following from [Error Prone Support's README:](https://github.com/PicnicSupermarket/error-prone-support#error-prone-support)

> Error Prone Support is a [Picnic](https://blog.picnic.nl/)-opinionated extension of Google's [Error Prone](https://github.com/google/error-prone). It aims to improve code quality, focussing on maintainability, consistency and avoidance of common pitfalls.

Not every check is a clear improvement for everyone, therefore everyone is able t configure what they prefer :). 
Just want to point that a coding standard / guideline for one company is a clear improvement, and for another it's not 😄.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/PreInterruptThreadDumpPrinter.java
rickie: > Rewrite directly outputs the command how to apply the suggested/enforced change upfront.

This we can also set up in almost a similar way. Instead of doing patch mode, you can use Error Prone to create a `.patch` file ([docs](https://errorprone.info/docs/patching)) and use that to also generate suggestions as PR comments, just like OpenRewrite can do.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
rickie: There are a few checks that simply do not have a suggested fix, and therefore cannot be automatically fixed. 

E.g. `TimeZoneUsage` is a check that only _flags_ violations, but cannot rewrite one.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,61 +10,65 @@ plugins {
 }
 
 dependencies {
-	errorprone(dependencyFromLibs("errorProne-core"))
+	errorprone(dependencyFromLibs("error-prone-contrib"))
+	errorprone(dependencyFromLibs("error-prone-core"))
 	errorprone(dependencyFromLibs("nullaway"))
+	errorprone(dependencyFromLibs("refaster-runner"))
```
marcphilipp: This dependency is only needed for refaster rules which are disabled below, aren't they?

```suggestion
```

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
marcphilipp: NullAway should be _disabled_, not enabled on OpenJ9.

IMHO it was clearer before. The reason for disabling NullAway isn't OpenJ9 but that Error Prone needs doesn't run on OpenJ9. Please change it back to the way it was before.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
marcphilipp: I double-checked that it's equivalent.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
marcphilipp: How is it DRY as it's now repeated, i.e. the opposite of DRY?

gradle/libs.versions.toml
```diff
@@ -32,7 +32,9 @@ bndlib = { module = "biz.aQute.bnd:biz.aQute.bndlib", version.ref = "bnd" }
 checkstyle = { module = "com.puppycrawl.tools:checkstyle", version.ref = "checkstyle" }
 classgraph = { module = "io.github.classgraph:classgraph", version = "4.8.181" }
 commons-io = { module = "commons-io:commons-io", version = "2.20.0" }
-errorProne-core = { module = "com.google.errorprone:error_prone_core", version = "2.42.0" }
+error-prone-core = { module = "com.google.errorprone:error_prone_core", version = "2.42.0" }
+error-prone-contrib = { module = "tech.picnic.error-prone-support:error-prone-contrib", version = "0.25.0" }
+refaster-runner = { module = "tech.picnic.error-prone-support:refaster-runner", version = "0.25.0" }
```
marcphilipp:
```suggestion
```

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,61 +10,65 @@ plugins {
...
-	onlyNullMarked = true
-}
-
 tasks.withType<JavaCompile>().configureEach {
 	options.errorprone {
-		val shouldDisableErrorProne = java.toolchain.implementation.orNull == JvmImplementation.J9
-		if (name == "compileJava" && !shouldDisableErrorProne) {
+		disableAllChecks = !(J9 != java.toolchain.implementation.orNull && "compileJava" == name)
+		if (!disableAllChecks.get()) {
+			errorproneArgs.add("-XepOpt:Refaster:NamePattern=^(?!.*Rules\\$).*") // currently failing Refaster; might consider whitelist.
```
marcphilipp:
```suggestion
```

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: yes im loosing focus here, im sorry.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: undo like suggested, considered off-topic, sorry for bothering.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: thanks.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
Pankraz76: thank you

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,49 +9,53 @@ plugins {
...
-				// This check is opinionated wrt. which method names it considers unsuitable for import which includes
-				// a few of our own methods in `ReflectionUtils` etc.
-				"BadImport",
-
-				// The findings of this check are subjective because a named constant can be more readable in many cases
-				"UnnecessaryLambda",
-
-				// Resolving findings for these checks requires ErrorProne's annotations which we don't want to use
-				"AnnotateFormatMethod",
+				"AnnotateFormatMethod", // We don`t want to use ErrorProne`s annotations.
```
marcphilipp:
```suggestion
				"AnnotateFormatMethod", // We don`t want to use ErrorProne's annotations.
```

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,49 +9,53 @@ plugins {
...
-				"InlineMeSuggester",
 				"ImmutableEnumChecker",
-
-				// Resolving findings for this check requires using Guava which we don't want to use
-				"StringSplitter",
-
-				// Produces a lot of findings that we consider to be false positives, for example for package-private
-				// classes and methods
-				"MissingSummary",
+				"InlineMeSuggester",
```
marcphilipp:
```suggestion
				"InlineMeSuggester", // We don`t want to use ErrorProne's annotations.
```

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,49 +9,53 @@ plugins {
...
-				"BadImport",
-
-				// The findings of this check are subjective because a named constant can be more readable in many cases
-				"UnnecessaryLambda",
-
-				// Resolving findings for these checks requires ErrorProne's annotations which we don't want to use
-				"AnnotateFormatMethod",
+				"AnnotateFormatMethod", // We don`t want to use ErrorProne`s annotations.
+				"BadImport", // This check is opinionated wrt. which method names it considers unsuitable for import which includes a few of our own methods in `ReflectionUtils` etc.
 				"DoNotCallSuggester",
```
marcphilipp:
```suggestion
				"DoNotCallSuggester", // We don`t want to use ErrorProne's annotations.
```

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,49 +9,53 @@ plugins {
...
-				"InlineMeSuggester",
 				"ImmutableEnumChecker",
```
marcphilipp:
```suggestion
				"ImmutableEnumChecker", // We don`t want to use ErrorProne's annotations.
```

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -9,49 +9,53 @@ plugins {
...
+				"IdentityConversion",
+				"LexicographicalAnnotationAttributeListing",
+				"LexicographicalAnnotationListing",
+				"MissingTestCall",
+				"NestedOptionals",
+				"NonStaticImport",
+				"OptionalOrElseGet",
+				"PrimitiveComparison",
+				"StaticImport",
+				"TimeZoneUsage",
```
marcphilipp, line 52: Are you going to address these (except `DirectReturn`) in separate PRs?

Pankraz76: Yes, Lexa is something easy to consider, if this meets the taste of the community.

- https://github.com/checkstyle/checkstyle/pull/17892

Static import is something to suggest to avoid lots of redundant context. `Collection.` stuff.

`ConstantNaming` is also kind of a no-brainer, as it's the quintessence of JCC, but here it's not fixable.
We'll handle this successfully in Spotless, so this is something to investigate further.

Everything else is then up to the community, kind of, as we definitely need all of these, as we are failing them even though it should not be necessary! (since they are not all error severity, but even low-level like suggestion). Isn't it? @rickie

This itself indicates some misconfigured or problematic tooling.

Thank you both for the great learning experience - both technical and foremost in social regards, which is kind of a big challenge for me personally.

Please excuse me for being inconsistent sometimes.

gradle/plugins/common/src/main/kotlin/junitbuild.java-nullability-conventions.gradle.kts
```diff
@@ -28,11 +28,11 @@ tasks.withType<JavaCompile>().configureEach {
 		val shouldDisableErrorProne = java.toolchain.implementation.orNull == JvmImplementation.J9
 		if (name == "compileJava" && !shouldDisableErrorProne) {
 			disable(
-				"AnnotateFormatMethod", // We don`t want to use ErrorProne`s annotations.
+				"AnnotateFormatMethod", // We don`t want to use ErrorProne's annotations.
 				"BadImport", // This check is opinionated wrt. which method names it considers unsuitable for import which includes a few of our own methods in `ReflectionUtils` etc.
-				"DoNotCallSuggester",
-				"ImmutableEnumChecker",
-				"InlineMeSuggester",
+				"DoNotCallSuggester", // We don`t want to use ErrorProne's annotations.
```
Pankraz76, line 33: now its very clear, thanks.

PR #5027, opened 2025-10-06
Introduce `ParallelExecutionInterceptor` and sample implementation

### LINKED ISSUE #3108: JUnit parallel executor running too many tests with a fixed policy.

marcphilipp:
The new `junit.jupiter.execution.parallel.config.interceptor.class`
configuration parameter allows configuring the fully qualified class
name of a `ParallelExecutionInterceptor` implementation that gets called
for each executed `TestTask`. The `FixedThreadPoolForTests` sample
implementation uses a separate fixed thread pool to run all tasks of
type `TEST` with execution mode `CONCURRENT` in.

Resolves #3108.

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
@@ -34,6 +34,7 @@ import org.junit.platform.commons.JUnitException;
 import org.junit.platform.commons.PreconditionViolationException;
 import org.junit.platform.commons.logging.LoggerFactory;
 import org.junit.platform.commons.util.ExceptionUtils;
+import org.junit.platform.commons.util.Preconditions;
 import org.junit.platform.commons.util.ReflectionUtils;
 import org.junit.platform.engine.ConfigurationParameters;
 
@@ -158,10 +159,17 @@ public class ForkJoinPoolHierarchicalTestExecutorService implements Hierarchical
 
 	@Override
 	public void invokeAll(List<? extends TestTask> tasks) {
+
 		if (tasks.size() == 1) {
 			new ExclusiveTask(tasks.get(0)).execSync();
 			return;
 		}
+
+		// If this method is called from outside the used ForkJoinPool,
+		// calls to fork() will schedule tasks in the commonPool
+		Preconditions.condition(isAlreadyRunningInForkJoinPool(),
+			"invokeAll() must be called from a thread in the ForkJoinPool");
+
 		Deque<ExclusiveTask> isolatedTasks = new ArrayDeque<>();
 		Deque<ExclusiveTask> sameThreadTasks = new ArrayDeque<>();
 		Deque<ExclusiveTask> concurrentTasksInReverseOrder = new ArrayDeque<>();


mpkorstanje:
> The FixedThreadPoolForTests sample implementation uses a separate fixed thread pool to run all tasks of type TEST with execution mode CONCURRENT in.

While this solves the specific problem, the implementation feels incomplete. Is that intentional?

When using a fixed thread pool, for each subtrees that must be executed on the same thread, I would expect that a thread from the thread pool is used for that entire subtree. With tasks of the type tests as the trivial case.

Edit: wording for clarity.

marcphilipp:
> When using a fixed thread pool, for each subtrees that must be executed on the same thread, I would expect that a thread from the thread pool is used for that entire subtree. With tasks of the type tests as the trivial case.

Since tests using `ExecutionMode.CONCURRENT` can be registered dynamically at runtime, we could only compute this value reliably if all `TestDescriptors` in the subtree returned `false` from their `mayRegisterTests()` methods. I guess we would restrict dynamic children to use `SAME_THREAD` regardless of what they return but we can't do so in general because that would be a regression. Thoughts?

mpkorstanje:
I was under the impression that the `NodeExecutionAdvisor` would force the execution mode of all descendants to `SAME_THREAD`. So all dynamic children are already have the `SAME_THREAD` execution mode if any of their ancestors have `SAME_THREAD` regardless of their own execution mode.

marcphilipp:
> I was under the impression that the `NodeExecutionAdvisor` would force the execution mode of all descendants to `SAME_THREAD`.

It only does so if exclusive resource / resource locks are used.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
```diff
@@ -129,10 +159,17 @@ private boolean isAlreadyRunningInForkJoinPool() {
...
+
 		if (tasks.size() == 1) {
 			new ExclusiveTask(tasks.get(0)).execSync();
 			return;
 		}
+
+		// If this method is called from outside the used ForkJoinPool,
+		// calls to fork() will schedule tasks in the commonPool
+		Preconditions.condition(isAlreadyRunningInForkJoinPool(),
+			"invokeAll() must be called from a thread in the ForkJoinPool");
```
mpkorstanje, line 171: Would it be possible to create a dummy task that calls `invokeAll` and `submit` that task instead?

mpkorstanje: Ah. No then `SAME_THREAD` won't work.

mpkorstanje: But what if only the call to `joinConcurrentTasksInReverseOrderToEnableWorkStealing(concurrentTasksInReverseOrder);` is submitted?

mpkorstanje:
>  I guess we would restrict dynamic children to use SAME_THREAD regardless of what they return

What if, upon encountering `DYNAMIC` children, execution goes back to the `ForkJoinPoolHierarchicalTestExecutorService` again, and then invokes the `ParallelExecutionInterceptor` for each task as needed?

mpkorstanje:
To clarify what I meant by incompleteness. When using `junit.jupiter.execution.parallel.config.interceptor.class` I would expect that it is possible to invoke all methods without a fork join pool. Not just the concurrent test methods methods but also the before all and same thread annotated methods. 

Expressed as a test case:

```java
@Execution(CONCURRENT)
static class ExampleTest {
	
	final static ThreadLocal<Integer> threadLocal = new ThreadLocal<>();

	@BeforeAll
	static void initialize() {
		assertThat(Thread.currentThread().getName()).doesNotStartWith("ForkJoinPool");
		threadLocal.set(42);
	}

	@Test
	@Execution(SAME_THREAD)
	void firstTest() throws Exception {
		assertThat(Thread.currentThread().getName()).doesNotStartWith("ForkJoinPool");
		assertEquals(42, threadLocal.get());
	}

	@Test
	void secondTest() throws Exception {
		assertThat(Thread.currentThread().getName()).doesNotStartWith("ForkJoinPool");
		assertNull(threadLocal.get());
	}
}
```

But I haven't quite got the time to try out my suggestions in code and see if they would actually work as expected..  Apologies for that.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ForkJoinPoolHierarchicalTestExecutorService.java
mpkorstanje: Sketched out a way to make this work in https://github.com/junit-team/junit-framework/pull/5038

marcphilipp: Thanks! I still can't shake the feeling that this entire PR is a workaround and we should try to fix the problem at its source, i.e. our usage of `ForkJoinPool`. It might be feasible to re-implement with a regular thread pool and a queue. I'll ponder it.

marcphilipp: I just opened a draft PR: #5060

marcphilipp:
Closing in favor of #5060

PR #5060, opened 2025-10-13
Implement parallel executor service without `ForkJoinPool`

### LINKED ISSUE #3108: JUnit parallel executor running too many tests with a fixed policy.

marcphilipp:
This PR introduces a new implementation of `HierarchicalTestExecutorService` that runs rests in parallel and has limited work stealing capabilities but is not based on `ForkJoinPool`. It avoids its pitfalls (such as #3945) and #3108 but may require additional threads because its work stealing is limited to direct children. Contrary to the `ForkJoinPool` implementation, the new executor service guarantees that no more than `parallelism` test nodes are executed in parallel.

My intention is to initially ship this implementation as an opt-in feature (via the new `junit.jupiter.execution.parallel.executor` configuration parameter) in 6.1, make it an opt-out feature in 6.2, and drop support for the `ForkJoinPool`-based implementation in a later to-be-determined release.

The PR is not yet finished but feedback is already welcome! If you use parallel test execution in your projects (or other test engines), it would be great if you could try out the new implementation and report your observations.

Resolves #3108.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/test/kotlin/example/KotlinCoroutinesDemo.kt
+++ b/documentation/src/test/kotlin/example/KotlinCoroutinesDemo.kt
@@ -14,9 +14,13 @@ import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 
 // end::user_guide[]
+// @formatter:off
 @Suppress("JUnitMalformedDeclaration")
 // tag::user_guide[]
 class KotlinCoroutinesDemo {
+    // end::user_guide[]
+    // @formatter:on
+    // tag::user_guide[]
     @BeforeEach
     fun regularSetUp() {
     }
--- a/gradle/plugins/common/src/main/kotlin/junitbuild.spotless-conventions.gradle.kts
+++ b/gradle/plugins/common/src/main/kotlin/junitbuild.spotless-conventions.gradle.kts
@@ -58,6 +58,7 @@ spotless {
 			licenseHeaderFile(license.headerFile)
 			trimTrailingWhitespace()
 			endWithNewline()
+			toggleOffOn("formatter:off", "formatter:on")
 		}
 	}
 


junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,549 @@
...
+		return new BlockingAwareFuture<@Nullable Void>(enqueue(testTask).future(), WorkerThread.BlockHandler.INSTANCE);
+	}
+
+	@Override
+	public void invokeAll(List<? extends TestTask> testTasks) {
+		LOGGER.trace(() -> "invokeAll: " + testTasks);
+
+		var workerThread = WorkerThread.get();
+		Preconditions.condition(workerThread != null && workerThread.executor() == this,
+			"invokeAll() must not be called from a thread that is not part of this executor");
```
mpkorstanje: double negative, maybe reword for clarity.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,549 @@
...
+			var workerThread = get();
+			if (workerThread == null) {
+				throw new IllegalStateException("Not on a worker thread");
+			}
+			return workerThread;
+		}
+
+		void processQueueEntries() {
+			while (!threadPool.isShutdown()) {
+				workerLease = workerLeaseManager.tryAcquire();
```
mpkorstanje: nitpick: Making the worker lease auto-closable and using it with a try-with-resources, can remove the  `.release()` call.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,549 @@
...
+		boolean remove(Entry entry) {
+			return queue.remove(entry);
+		}
+
+		boolean isEmpty() {
+			return queue.isEmpty();
+		}
+
+		private record Entry(TestTask task, CompletableFuture<@Nullable Void> future, int level, int attempts)
+				implements Comparable<Entry> {
```
mpkorstanje: I think it would make sense to document that low level items are prioritized to maximize the potential for stealing. It took me a while to puzzle that out.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,549 @@
...
+		if (testTasks.size() == 1) {
+			executeTask(testTasks.get(0));
+			return;
+		}
+
+		List<TestTask> isolatedTasks = new ArrayList<>(testTasks.size());
+		List<TestTask> sameThreadTasks = new ArrayList<>(testTasks.size());
+		var queueEntries = forkConcurrentChildren(testTasks, isolatedTasks::add, sameThreadTasks::add);
+		executeAll(sameThreadTasks);
+		var remainingForkedChildren = stealWork(queueEntries);
```
mpkorstanje: If there are no `sameThreadTasks`, I think it would make sense to reserve one of the children for this thread. Preferably a container to make sure this thread "runs away" from any work stealers to make sure it doesn't have all its work stolen.

mpkorstanje: For the same reason I think it would also improve performance if test tasks are queued before the container tasks.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: Done in 0d201d1ce9cc45a174addb0e2679e3e7673604cf

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: Done in 08568e4065f9c266ccc1487525d0d9d873ad23ec.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: > For the same reason I think it would also improve performance if test tasks are queued before the container tasks.

Good idea! Done in 1055dbb92802cf0b88f88ad3335dacb7a7677bbd.

> If there are no `sameThreadTasks`, I think it would make sense to reserve one of the children for this thread.

Done in 1196b178a76fc9c1512615df7b39e3ad63a997b9.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: Yes, I need to do a bunch of documenting. :slightly_smiling_face:

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,625 @@
...
+				executeTask(testTasks.get(0));
+				return;
+			}
+
+			List<TestTask> isolatedTasks = new ArrayList<>(testTasks.size());
+			List<TestTask> sameThreadTasks = new ArrayList<>(testTasks.size());
+			var allForkedChildren = forkConcurrentChildren(testTasks, isolatedTasks::add, sameThreadTasks);
+			executeAll(sameThreadTasks);
+			var remainingForkedChildren = stealWork(allForkedChildren);
+			waitFor(remainingForkedChildren);
```
mpkorstanje: The name `remainingForkedChildren` isn't quite right. These can be children that are picked up by other threads or children that are currently blocked by a locked resource.

So before waiting this thread could execute any remaining unexecuted children in a blocking fashion.

marcphilipp: Do you have a proposal? `concurrentChildren`?

mpkorstanje: Not immediately. I'd rework the logic first then see what falls out.

The logic would have to make a distinction between children that executed by another thread and children that were skipped over to due to resource locks. The latter can be executed blockingly before calling `waitFor`. While looping through the `resourceLockedChildren`(?)  the `remainingForkedChildren`  would also have to be updated as other threads pick them up.

So maybe name `remainingForkedChildren` to `childrenExecutedOnAnotherThread`, but that is unwieldy. 😅

mpkorstanje: And I should have prefaced this by noting that the current thread starts to wait to early, there is potentially still work left to be done. But the naming is hiding that.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,625 @@
...
+			maybeStartWorker();
+		}
+	}
+
+	private void maybeStartWorker() {
+		if (threadPool.isShutdown() || !workerLeaseManager.isLeaseAvailable() || workQueue.isEmpty()) {
+			return;
+		}
+		try {
+			threadPool.execute(() -> {
```
mpkorstanje: Tested with Cucumber, no easy reproducer, sorry.

```
cucumber.execution.parallel.enabled=true
cucumber.execution.parallel.config.strategy=fixed
cucumber.execution.parallel.config.fixed.parallelism=4
cucumber.execution.parallel.config.fixed.max-pool-size=4
```

Got:

```
Exception in thread "junit-2-worker-6" java.util.concurrent.RejectedExecutionException: Task org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService$$Lambda$507/0x0000767a601808d0@48e4bb11 rejected from java.util.concurrent.ThreadPoolExecutor@12061f86[Running, pool size = 4, active threads = 4, queued tasks = 0, completed tasks = 2]
	at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2065)
	at java.base/java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:833)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1365)
	at org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService.maybeStartWorker(ConcurrentHierarchicalTestExecutorService.java:135)
	at org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService$WorkerLeaseManager.release(ConcurrentHierarchicalTestExecutorService.java:578)
	at org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService$WorkerLease.release(ConcurrentHierarchicalTestExecutorService.java:615)
	at org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService$WorkerLease.close(ConcurrentHierarchicalTestExecutorService.java:610)
	at org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService$WorkerThread.processQueueEntries(ConcurrentHierarchicalTestExecutorService.java:217)
	at org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService.lambda$maybeStartWorker$0(ConcurrentHierarchicalTestExecutorService.java:138)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1365)
```	

So I don't think the check against `isLeaseAvailable` is sufficient here to prevent races on `execute`.

marcphilipp: It's expected that it needs more threads in some cases, though. Could you please try without the max-pool-size setting?

mpkorstanje: This shouldn't be one of those cases. There are no resource locks. I think you can reproduce this with JUnit with 2 classes, each on separate workers that both race starting more workers.

I'll have a look at making a test case with JUnit.

marcphilipp: I've pushed a commit that avoids that race. A test case would be welcome, though!

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,625 @@
...
+		private boolean tryToStealWork(WorkQueue.Entry entry) {
+			if (entry.future.isDone()) {
+				return false;
+			}
+			var claimed = workQueue.remove(entry);
+			if (claimed) {
+				LOGGER.trace(() -> "stole work: " + entry);
+				var executed = tryExecute(entry);
+				if (!executed) {
+					workQueue.reAdd(entry);
```
mpkorstanje: If https://github.com/junit-team/junit-framework/pull/5060/files#r2429023943 is resolved, there is another potential loss of efficiency hiding here.

In `stealWork` the current thread considers each of the `forkedChildren` exactly once. If that were to be improved to consider all unexecuted children, by re-adding we'd lose track of this `entry` as eventually `incrementAttempts` creates a new entry and this thread would assume that because it is no longer in the queue, it has been picked up by another thread. Even if it was then locked, re-added and at present available for workstealing by the current thread.

Not sure how to fix this. Making entries mutable would be one solution.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,625 @@
...
+				return new Entry(task(), future, level, attempts + 1);
+			}
+
+			@Override
+			public int compareTo(Entry that) {
+				var result = Integer.compare(that.level, this.level);
+				if (result == 0) {
+					result = Boolean.compare(this.isContainer(), that.isContainer());
+					if (result == 0) {
+						result = Integer.compare(that.attempts, this.attempts);
```
mpkorstanje: I think `attempt` should be sorted on first.

```mermaid
---
config:
    layout: elk
---
erDiagram
    A ||--o{ AA: ""
    A ||--o{ AB: ""
    A ||--o{ AC: ""
    AA ||--o{ AAA: ""
    AA ||--o{ AAB: ""
    AB ||--o{ ABA: ""
    AB ||--o{ ABB: ""
```

Given this tree. Suppose 
 * thread 1 is working on AA, claiming resource A, 
 * thread 2 is working on AB and ABA, no resources needed. 
 * thread 3 wants to work on AC, and would need resource A.

Item ABB would be available for work stealing, but isn't picked up by thread 3  because every time item AC is considered it readded to the front of the queue with an incremented attempt counter.

mpkorstanje: Ah. No it won't because the work stealer will use  `executeTask` which blocks. So it hangs on `ABB`. This doesn't seem efficient.

I need to think about this a bit. It feels like there should be a more efficient way.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: Good catch! I guess we could track the attempts in a separate data structure. I'll first look into https://github.com/junit-team/junit-framework/pull/5060#discussion_r2429023943 and then this one.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: We could work steal across subtrees but then would have to make sure the resource locks are compatible. My intention was to avoid that complication at the expense of needing an extra thread in these cases. I think that tradeoff is acceptable as long as we guarantee that no more than `parallelism` tests are running concurrently which is now accomplished via worker leases. Happy to discuss ideas, though, of course!

mpkorstanje: I wasn't thinking about stealing work across subtrees. That's probably not necessary (yet?). Rather, I'm looking to keep the maximum number of created threads low and reuse existing once as much as possible - or when the maximum pool size is limited, optimize utilization. 

This is a useful property, for example when using a web driver, it is reused between tests by using a thread locals. So fewer threads means fewer drivers are created.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: Done: https://github.com/junit-team/junit-framework/pull/5068

mpkorstanje: Failures on [d111712](https://github.com/junit-team/junit-framework/pull/5060/commits/d11171283d022fa9bb8c8538705151d1587aef1b). Looks like [ba98383](https://github.com/junit-team/junit-framework/pull/5060/commits/ba9838334de5b12819c93cf4eec821b473b7c565) didn't quite fix things.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,625 @@
...
+				}
+				else {
+					queueEntries.add(WorkQueue.Entry.create(child));
+				}
+			}
+
+			if (!queueEntries.isEmpty()) {
+				if (sameThreadTasks.isEmpty()) {
+					// hold back one task for this thread
+					sameThreadTasks.add(queueEntries.poll().task);
```
mpkorstanje: I'm observing that invoking `poll()` does not preserve `queueEntries` insertion order. This means that we put children, a, b and c in the queue they'll come out a, c, b. 

To ensure our tests are stable and understandable  is probably not what we want to happen. And while there is no guaranteed ordering during test execution, when users do order their tests, I would expect that we pick them up in that order if possible.

Perhaps each entry needs an index in addition to its level?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,625 @@
...
+		}
+
+		private boolean tryToStealWork(WorkQueue.Entry entry) {
+			if (entry.future.isDone()) {
+				return false;
+			}
+			var claimed = workQueue.remove(entry);
+			if (claimed) {
+				LOGGER.trace(() -> "stole work: " + entry);
+				var executed = tryExecute(entry);
```
mpkorstanje: I have an intuition that the current thread should await any locks while the queue processing threads should skip over locked tasks. I.e. the reverse of the current setup.

If implemented correctly, the invariant of the current thread is it will finish all nodes of the sub-tree it working on unless they're stolen away. Because the current thread can't steal any work, it is not wasted if it has to await a lock and there is nothing else to do. The queue processing nodes however are wasted, they could be stealing something else.

We can then refine that further by making the current thread also skip over locked tasks and come back to them later.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: Also, now seeing with parallelism 4 and max-pool-size 4 

```
java.lang.StackOverflowError
	at java.base/java.lang.Exception.<init>(Exception.java:67)
	at java.base/java.lang.RuntimeException.<init>(RuntimeException.java:63)
	at java.base/java.util.concurrent.RejectedExecutionException.<init>(RejectedExecutionException.java:64)
	at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2065)
	at java.base/java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:833)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1365)
```

And with parallelism 4 and max-pool-size 6

```
java.util.concurrent.RejectedExecutionException: Task org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService$$Lambda$510/0x00007d95641813c0@5ef506a2 rejected from java.util.concurrent.ThreadPoolExecutor@786bc1c7[Running, pool size = 6, active threads = 6, queued tasks = 0, completed tasks = 0]

	at java.base/java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2065)
	at java.base/java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:833)
	at java.base/java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1365)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

I'll have a look at turning these into tests.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: Would you like to give that a try on another branch?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: Overall I think the problem is that worker lease manager keeps track of active threads, not the active + blocked threads. Could another manager be the solution?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: I'll make a test for this too. There are at least a few (2+) entries needed to make the flip happen.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: Yes, though I don't know how much work you've already got in progress.

marcphilipp: I've now pushed everything.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: I've added a commit that addresses this now.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: How would knowing the number of blocked threads help?

mpkorstanje: When `runBlocking` is invoked, the worker lease is released. So total threads in the pool = running threads + blocked threads.

mpkorstanje: Though, it looks like we can't really do that anyway. Only the thread pool knows what the thread pool is doing internally. What about using a `RejectedExecutionHandler` that discards the runnable?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
marcphilipp: Sorry, pushed a few more commits. I hope they are not throwing a wrench in your plans.

mpkorstanje: No problem.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: After [c8ddb46](https://github.com/junit-team/junit-framework/pull/5060/commits/c8ddb46e4991cf3b976d48e3a51750f83ee27739) I'm seeing:

```
> Task :platform-tests:test
Predictive Test Selection: 26 of 162 test classes selected with profile 'Standard' (saving 34.271s serial time)

ConcurrentHierarchicalTestExecutorServiceTests > limitsWorkerThreadsToMaxPoolSize() > repetition 77 of 100 FAILED
    java.util.concurrent.ExecutionException: java.util.concurrent.RejectedExecutionException: Task org.junit.platform.engine.support.hierarchical.ConcurrentHierarchicalTestExecutorService$$Lambda/0x0000000057398460@70fc0dae rejected from java.util.concurrent.ThreadPoolExecutor@39ad1fdb[Running, pool size = 3, active threads = 3, queued tasks = 0, completed tasks = 0]
```

But only with `./gradlew test`.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,730 @@
...
+			}
+
+			return queueEntries;
+		}
+
+		private Map<WorkStealResult, List<WorkQueue.Entry>> tryToStealWorkWithoutBlocking(
+				List<WorkQueue.Entry> forkedChildren) {
+
+			Map<WorkStealResult, List<WorkQueue.Entry>> queueEntriesByResult = new HashMap<>(
+				WorkStealResult.values().length);
```
mpkorstanje: A rare use case for an `EnumMap`?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: Can't quite reproduce it reliably.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: https://github.com/junit-team/junit-framework/pull/5076

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,729 @@
...
+			threadPool.execute(() -> {
+				LOGGER.trace(() -> "starting worker");
+				try {
+					WorkerThread.getOrThrow().processQueueEntries(workerLease, doneCondition);
+				}
+				finally {
+					workerLease.release(false);
+					LOGGER.trace(() -> "stopping worker");
+				}
+				maybeStartWorker(doneCondition);
```
mpkorstanje: Note to self: I still can't quite get my head around this line.

If there is a race condition where `processQueueEntries` may run out of tasks, and `forkAll` adding tasks and starting new workers, but failing because this worker is still active.

Then I would expect `processQueueEntries` to read from the work queue in a blocking manner.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: This delays blocking until there is nothing else to do https://github.com/junit-team/junit-framework/pull/5078

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: A good suggestion fell out in https://github.com/junit-team/junit-framework/pull/5078/commits/1ffca7b4af1ef39ef2e008f65eef3c7d47903bb8, not included on this branch yet.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
mpkorstanje: Answer: Because we're managing the degree of parallelism ourselves, blocking here will occupy a lease.

marcphilipp:
I changed how to opt-in to the feature and have updated the docs.

@mpkorstanje I'd appreciate your feedback! :slightly_smiling_face:

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -3429,6 +3429,33 @@ used instead.
...
+concurrently. You can configure which implementation of `HierarchicalTestExecutorService`
+is used be setting the `junit.jupiter.execution.parallel.config.executor-service`
+configuration parameter to one of the following options:
+
+`fork_join_pool` (default)::
+Use an executor service that is backed by a `ForkJoinPool` from the JDK. This will cause
+tests to be executed in a `ForkJoinWorkerThread` and, in some cases, cause usages of
+`ForkJoinPool` in test or production code or calls to blocking JDK APIs to increase the
+number of concurrently executing tests. To avoid this limitation, please use
+`worker_thread_pool`.
```
mpkorstanje, line 3449: Long sentence with same fragmentation. Cant' get multi-line suggestions to work.

```
`fork_join_pool` (default)::
Use an executor service that is backed by a `ForkJoinPool` from the JDK. This will cause
tests to be executed in a `ForkJoinWorkerThread`. In some cases usages of
`ForkJoinPool` in test or production code or calls to blocking JDK APIs may cause the
number of concurrently executing tests  to increase. To avoid this situation, please use
`worker_thread_pool`.
```

marcphilipp: Applied in 7a250e00101c442955ab555879a542143ffb5dc2

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/DefaultJupiterConfiguration.java
```diff
@@ -100,6 +102,17 @@ private void validateConfigurationParameters(DiscoveryIssueReporter issueReporte
...
 				}));
+		if (isParallelExecutionEnabled()
+				&& configurationParameters.get(PARALLEL_CONFIG_EXECUTOR_SERVICE_PROPERTY_NAME).isEmpty()) {
+			var info = DiscoveryIssue.create(Severity.INFO,
+				"Parallel test execution is enabled but the default ForkJoinPool-based executor service will be used. "
+						+ "Please give the new implementation based on a regular thread pool a try by setting the '"
+						+ PARALLEL_CONFIG_EXECUTOR_SERVICE_PROPERTY_NAME + "' configuration parameter to '"
+						+ WORKER_THREAD_POOL + "' and report any issues to the JUnit team. "
+						+ "Alternatively, set the configuration parameter to '" + FORK_JOIN_POOL
+						+ "' to hide this message.");
```
marcphilipp:
```suggestion
						+ "' to hide this message and use the original implementation.");
```

documentation/src/docs/asciidoc/release-notes/release-notes-6.1.0-M1.adoc
```diff
@@ -21,13 +21,20 @@ repository on GitHub.
...
+* Deprecate constructors for `ForkJoinPoolHierarchicalTestExecutorService` in favor of the
+  new `ParallelHierarchicalTestExecutorServiceFactory` that also supports
+  `WorkerThreadPoolHierarchicalTestExecutorService`.
 
 [[release-notes-6.1.0-M1-junit-platform-new-features-and-improvements]]
 ==== New Features and Improvements
 
 * Support for creating a `ModuleSelector` from a `java.lang.Module` and using
   its classloader for test discovery.
+* New `WorkerThreadPoolHierarchicalTestExecutorService` implementation of parallel test
```
mpkorstanje:
```suggestion
* New `WorkerThreadPoolHierarchicalTestExecutorService` implementation used for parallel test
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/WorkerThreadPoolHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,830 @@
...
+		`parallelism` tests are running concurrently, regardless whether the user code performs any	blocking operations.
+		If a worker lease was acquired, a worker thread is started. Each worker thread polls the shared work queue for
+		tasks to run. Since the tasks represent hierarchically structured tests, container tasks will call
+		`submit(TestTask)` or `invokeAll(List<TestTask>)` for their children, recursively. Child tasks with execution
+		mode `CONCURRENT` are submitted to the shared queue be prior to executing those with execution mode
+		`SAME_THREAD` directly. Each worker thread attempts to "steal" queue entries for its children and execute them
+		itself prior to waiting for its children to finish. In case it does need to block, it temporarily gives	up its
+		worker lease and starts another worker thread to compensate for the reduced `parallelism`. If the max pool size
+		does not permit starting another thread, that is ignored in case there are still other active worker threads.
+		The same happens in case a resource lock needs to be acquired.
```
mpkorstanje, line 91: This reads like a flow of consciousness.

I'd suggest rewriting it to `<objective>  <action>` format.

```
		This implementation is based on a regular thread pool and a work queue shared among all worker threads.

		Each worker thread scans the shared work queue for tasks to run. Since the tasks represent hierarchically
		structured tests, container tasks will call `submit(TestTask)` or `invokeAll(List<TestTask>)` for their
		children, recursively.
		
		To maintain the desired parallelism -- regardless whether the user code performs any blocking operations --
		a fixed number of worker leases is configured. Whenever a task is submitted to the work queue to be executed
		concurrently, an attempt is made to acquire a worker lease. If a worker lease was acquired, a worker thread is
		started. Each worker thread attempts to "steal" queue entries for its children and execute them itself prior to waiting
		for its children to finish.
		 
		To optimize CPU utilization, whenever a worker thread does need to block, it temporarily gives up its worker
		lease and attempts to start another worker thread to compensate for the reduced `parallelism`. If the max pool size does 
		not permit starting another thread, the attempt is ignored in case there are still other active worker threads.
		
		The same happens in case a resource lock needs to be acquired.
		
		To minimize the number of idle workers, worker threads will prefer to steal top level tasks, while working
		through their own task hierarchy in a depth first fashion. Furthermore, child tasks with execution mode
		`CONCURRENT` are submitted to the shared queue be prior to executing those with execution mode `SAME_THREAD`
		directly.
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/WorkerThreadPoolHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,830 @@
...
+		}
+
+	}
+
+	private enum BlockingMode {
+		NON_BLOCKING, BLOCKING
+	}
+
+	private static class WorkQueue implements Iterable<WorkQueue.Entry> {
+		private final AtomicLong index = new AtomicLong();
```
mpkorstanje: I'm having some second thoughts about this. This is an exhaustible resource, especially compared to `ConcurrentSkipListSet`. Not that I think any one could reasonably exhaust it, but those are famous last words.

mpkorstanje: I've had a look at replacing the `ConcurrentSkipListSet` with a `PriorityBlockingQueue`. Because we depend on both queue and iteration order, the `PriorityBlockingQueue` with its "messy"  iteration order can't be used because it is not guaranteed that the executing thread and work stealer will approach each other from opposite directions..

Resolved the potential exhaustion in  1f9ffde7811e814c98437a1742c5b4c70cc9ee60 by using the `UniqueId` to create an arbitrary ordering. This however meant that invokeAll order wasn't used any more. 592a05e8bb51a0d8b57b10b0e7c767691c43c5ae solves that.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/config/DefaultJupiterConfiguration.java
marcphilipp: Improved in ea28eee5476df446c66e29928ef6a718523a1deb.

documentation/src/docs/asciidoc/release-notes/release-notes-6.1.0-M1.adoc
marcphilipp: Done in 98d7ce88c91e2afd18d203b2eebec799ea7ce01c

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/WorkerThreadPoolHierarchicalTestExecutorService.java
marcphilipp: Thanks! Done in 8495e8b350bb3edcabc7585a69060d9be92bb826

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/WorkerThreadPoolHierarchicalTestExecutorService.java
```diff
@@ -0,0 +1,922 @@
...
+					// Process the queue from the beginning
+					return;
+				}
+				if (result == WorkStealResult.RESOURCE_LOCK_UNAVAILABLE) {
+					entriesRequiringResourceLocks.add(entry);
+				}
+			}
+			if (!entriesRequiringResourceLocks.isEmpty()) {
+				// One entry at a time to avoid blocking too much
+				tryToStealWork(entriesRequiringResourceLocks.get(0), BlockingMode.BLOCKING);
```
mpkorstanje: Entries can be stolen by other workers so, this restarts scanning the queue from the beginning too early.

mpkorstanje: Fixed e7791ae5f9f17eb5d333e81a282edfd5cc4fb41a

mpkorstanje:
CI failing on https://github.com/junit-team/junit-framework/commit/85ccf17aacaa2d251f0eb50ee38a174c48de975e from `main`.

PR #5078, opened 2025-10-19
Skip over unavailable resources in ConcurrentHierarchicalTestExecutorService

mpkorstanje:
Depends on https://github.com/junit-team/junit-framework/pull/5076.

---

---

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
@@ -351,7 +351,7 @@ public class ConcurrentHierarchicalTestExecutorService implements HierarchicalTe
 			}
 			var claimed = workQueue.remove(entry);
 			if (claimed) {
-				LOGGER.trace(() -> "stole work: " + entry);
+				LOGGER.trace(() -> "stole work: " + entry.task);
 				var executed = executeStolenWork(entry, blockingMode);
 				if (executed) {
 					return WorkStealResult.EXECUTED_BY_THIS_WORKER;


junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -528,24 +537,27 @@ private enum BlockingMode {
 		NON_BLOCKING, BLOCKING
 	}
 
-	private static class WorkQueue {
-
-		private final EntryOrdering ordering = new EntryOrdering();
-		private final Queue<Entry> queue = new PriorityBlockingQueue<>();
```
marcphilipp, line 534: Looks like I forgot to use the custom comparator... 😳

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/ConcurrentHierarchicalTestExecutorService.java
```diff
@@ -226,13 +227,25 @@ void processQueueEntries(WorkerLease workerLease, BooleanSupplier doneCondition)
...
+			var queueEntriesByResult = tryToStealWorkWithoutBlocking(workQueue);
+			var queueModified = queueEntriesByResult.containsKey(WorkStealResult.EXECUTED_BY_THIS_WORKER) //
+					|| queueEntriesByResult.containsKey(WorkStealResult.EXECUTED_BY_DIFFERENT_WORKER);
+			if (queueModified) {
+				return;
+			}
+			var entriesRequiringResourceLocks = queueEntriesByResult.get(WorkStealResult.RESOURCE_LOCK_UNAVAILABLE);
+			if (entriesRequiringResourceLocks != null) {
+				// One entry at a time to avoid blocking too much
+				tryToStealWork(entriesRequiringResourceLocks.get(0), BlockingMode.BLOCKING);
```
marcphilipp, line 254: Could you please elaborate on this? Is there evidence this is better than doing all of them?

mpkorstanje: After blocking and executing one task, the queue state may have changed and other tasks that don't require blocking may have been added. At the same time the tasks left  were known to be blocked. So it makes sense to start the queue processing from the beginning. 

Ideally we'd avoid blocking execution here all together. Some other thread owns the resource that blocks these tasks. So they'll get around to the other blocked tasks eventually making the extra workers redundant.

Currently that puts the `processQueueEntries` into a busy waiting loop and if we do solve that, then we have to find a way to maintain parallelism when we have one big task that claims resources A,B, and C and three smaller tasks that require A, B, C respectively.

PR #5092, opened 2025-10-25
Make jdk.jfr import optional in OSGi manifest

laeubi:
With the recent adoption of JUnit 6 we noticed that JUnit no longer can run on stripped down JVMs that are missing the flight recorder module because it has a mandatory import in the OSGi manifest.

As the flightrecorder integration seems optional, one might want to make the import of the package optional as well

---

---

The content was generated [by copilote here](https://github.com/laeubi/junit-framework/pull/1) I squashed the commits, reviewed the result and found it being sane compared to how it was done for other optional imports. Please let me know if any adjustments are necessary as I'm not familiar with gradle.

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-6.0.1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-6.0.1.adoc
@@ -16,6 +16,8 @@ repository on GitHub.
 [[release-notes-6.0.1-junit-platform-bug-fixes]]
 ==== Bug Fixes
 
+* The `jdk.jfr` package is now an optional import when using the `junit-platform-launcher`
+  as an OSGi bundle.
 * Remove remnants of Java 8 compatibility from User Guide.
 
 [[release-notes-6.0.1-junit-platform-deprecations-and-breaking-changes]]


gradle/plugins/common/src/main/kotlin/junitbuild.osgi-conventions.gradle.kts
```diff
@@ -38,6 +39,7 @@ tasks.withType<Jar>().named {
 				Import-Package: \
 					${importAPIGuardian},\
 					${importJSpecify},\
+					${importJdkJfr},\
```
marcphilipp: Please move these changes to the build script of junit-platform-launcher as they're specific to that module/bundle (see junit-platform-commons for an example).

laeubi: @marcphilipp I'm sadly not very familiar with gradle build scripts and looking at that build-script I'm not completely sure how to archive this, as that one adds an *additional* header and here we want to modify an existing definition. I could repeat the definition there but it seems a bit odd to maintain then.

Any suggestion?

marcphilipp: No worries! I pushed a few commits. I agree that modifying an existing definition is not as DRY as one would wish.

marcphilipp:
Cherry-picked to `releases/6.0.x` in fedda882590c121e66a26de6a78ba5b532ec8aec

PR #5099, opened 2025-10-27
Allow specifying which `Nodes` require the global read lock

### LINKED ISSUE #2982: Execution mode is needlessly c
## Step to reproduce

Originally reported at: https://github.com/cucumber/cucumber-jvm/issues/2589. 

Given a tree of test descriptors where one test declares a global resource lock:

```
|- Container A
| |- Test 1 ( .getExclusiveResources() => GLOBAL_READ_WRITE )
| |- Test 2
| \- Test 3
|- Container B
  |- Test 4
  |- Test 5
  \- Test 6
```

When executed in parallel `Container A` is executed with `ExecutionMode.SAME_THREAD`. This reduces the efficiency of parallel execution somewhat as the serial part of the execution now also includes tests 2 and 3.

## Analysis 

The same thread execution mode is applied to a container if any of its children has a `GLOBAL_READ_WRITE` lock:

https://github.com/junit-team/junit5/blob/5fdb138924f59703a6a4fcd0cdcde6d1a224a950/junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/NodeTreeWalker.java#L73-L76

The rationale for this behaviour is given in https://github.com/junit-team/junit5/issues/2038#issuecomment-541171707:
> The (pessimistic) assumption behind forcing children of containers with resource locks to use execution mode SAME_THREAD is that the resource might not be safe to access in e.g. `@BeforeAll` methods and in different threads from `@Test` methods. Thus it's not the same as having `@ResourceLock` on every `@Test` method. IIRC we had a similar issue for `@Execution` only applying to methods while being declared on the class level.

However while this is true for JUnit Jupiter, this is not the case for Cucumber. In Cucumber containers do not execute any hooks. So providing containers with a way to signal that the execution mode does not need to be coarsened would allow the execution mode to be applied more selectively.

marcphilipp:
Implementations of `HierarchicalTestEngine` can now specify which nodes
require the global read lock by overriding the new
`Node.isGlobalReadLockRequired()` method to return `false`.

This allows for more fine-grained locks when intermediate container do
not have any behavior of their own and thus do not require even the
global read lock.

Resolves #2982.

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/release-notes/release-notes-6.0.1.adoc
+++ b/documentation/src/docs/asciidoc/release-notes/release-notes-6.0.1.adoc
@@ -45,6 +45,8 @@ repository on GitHub.
   to avoid conflicts with other control characters.
 * Fix `IllegalAccessError` thrown when using the Kotlin-specific `assertDoesNotThrow`
   assertion.
+* Stop reporting discovery issues for synthetic methods, particularly in conjunction with
+  Kotlin suspend functions.
 
 [[release-notes-6.0.1-junit-jupiter-deprecations-and-breaking-changes]]
 ==== Deprecations and Breaking Changes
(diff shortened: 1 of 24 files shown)


marcphilipp:
@leonard84 I'd also appreciate your opinion, if you have time to take a look.

PR #5100, opened 2025-10-27
Fix support for package-private test methods

### LINKED ISSUE #5098: Test method is silently skipped when parent class has same-named package-private test method
JUnit supports public, protected and package-private test methods.
JUnit supports test class inheritance.
Public and protected methods can be overridden in JLS sense. JUnit recognizes this situation and runs the test only once. In test source code this is indicated with `@Override` (in any reasonably managed project).
Package-private methods do not override each other in JLS sense. Both parent and child classes can declare `@Test void testFoo` method without any warning and without need for `@Override`.  It is expected that such test methods are considered unrelated by JUnit and both are run. Observed behavior is different: only the parent class test is run. The child test is silently ignored.

Using JUnit 6.0.0

## Steps to reproduce

```java
package a;

public abstract class TestParent {
    @Test
    void testParent()
    {
        System.out.println("TestParent.testParent");
    }

    @Test
    public void testPublicNameCollision() {
        System.out.println("TestParent.testPublicNameCollision");
    }

    @Test
    void testPackagePrivateNameCollision() {
        System.out.println("TestParent.testPackagePrivateNameCollision");
    }
}
```

```java
package b;

public class TestChild extends a.TestParent {
    @Test
    void testChild() {
        System.out.println("TestChild.testChild");
    }

    @Test
    @Override // clear source-level indication of test overriding. As expected, only the child test code is run.
    public void testPublicNameCollision() {
        System.out.println("TestChild.testPublicNameCollision");
    }

    @Test
    // no source-level indication of test overriding. This test code does not run, i.e. is silently skipped.
    void testPackagePrivateNameCollision() {
        System.out.println("TestChild.testPackagePrivateNameCollision");
    }
}
```

<details>
<summary>maven setup</summary>

```xml
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>org.junit</groupId>
                <artifactId>junit-bom</artifactId>
                <version>6.0.0</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
        <dependency>
            <groupId>org.junit.jupiter</groupId>
            <artifactId>junit-jupiter</artifactId>
            <scope>test</scope>
        </dependency>
        <dependency>
            <groupId>org.junit.jupiter</groupId>
            <artifactId>junit-jupiter-api</artifactId>
            <scope>test</scope>
        </dependency>
    </dependencies>
```

</details>

`mvn test` prints

```
TestParent.testPackagePrivateNameCollision
TestParent.testParent
TestChild.testChild
TestChild.testPublicNameCollision
```

Note that there is no `TestChild.testPackagePrivateNameCollision` in the output.

## Context

 - Used versions (Jupiter/Vintage/Platform): junit-bom `6.0.0`

marcphilipp:
Prior to this commit, when the test super class wss declared in another
package and declared a package-private test method with the same
signature as the subclass, the method from the subclass was silently
ignored. With the changes in this commit, both test methods are
executed, albeit with the same display name.

Fixes #5098.

---

---

### Definition of Done


junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodSelectorResolver.java
```diff
@@ -223,15 +224,22 @@ Optional<TestDescriptor> resolveUniqueIdIntoTestDescriptor(UniqueId uniqueId, Co
...
-			return parent.getUniqueId().append(segmentType, methodId);
+		private UniqueId createUniqueId(Method method, TestDescriptor parent, Class<?> testClass) {
+			return parent.getUniqueId().append(segmentType, computeMethodId(method, testClass));
+		}
+
+		private static String computeMethodId(Method method, Class<?> testClass) {
+			var parameterTypes = ClassUtils.nullSafeToString(method.getParameterTypes());
+			if (isPackagePrivate(method)
+					&& !method.getDeclaringClass().getPackageName().equals(testClass.getPackageName())) {
+				return "%s#%s(%s)".formatted(method.getDeclaringClass().getName(), method.getName(), parameterTypes);
```
marcphilipp: The declaring class is added to the unique ID segment of the test method, but only in this case, not in general, to avoid other unique IDs from becoming longer unnecessarily.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/MethodSelector.java
```diff
@@ -275,9 +275,17 @@ public boolean equals(Object o) {
...
 		}
 		MethodSelector that = (MethodSelector) o;
-		return Objects.equals(this.className, that.className)//
+		var equal = Objects.equals(this.className, that.className)//
 				&& Objects.equals(this.methodName, that.methodName)//
 				&& Objects.equals(this.parameterTypeNames, that.parameterTypeNames);
+		if (equal) {
+			var thisJavaMethod = this.javaMethod;
+			var thatJavaMethod = that.javaMethod;
+			if (thisJavaMethod != null && thatJavaMethod != null) {
```
mpkorstanje: Depending on wether or not `lazyLoadJavaMethod` was called, the equality changes. Is that intentional?

mpkorstanje: I suppose a `String declaringClassName` field would have to be added to the  `MethodSelector` to close that gap.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodSelectorResolver.java
mpkorstanje: This would be good to have as a method level comment.

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/MethodSelector.java
mpkorstanje: Concrete example:

```java
@Test // Fails but shouldn't
void distinguishesDeclaringClass2() throws Exception {
	var parentMethodSelector = new MethodSelector(TestCase.class,
		ParentTestCase.class.getDeclaredMethod("method", int.class, boolean.class));
	var childMethodSelector = new MethodSelector(TestCase.class, "method", int.class, boolean.class);

	assertThat(parentMethodSelector).isNotEqualTo(childMethodSelector);
	assertThat(childMethodSelector).isNotEqualTo(parentMethodSelector);
}


@Test // passes, but only because the childMethodSelector was mutated.
void distinguishesDeclaringClass3() throws Exception {
	var parentMethodSelector = new MethodSelector(TestCase.class,
		ParentTestCase.class.getDeclaredMethod("method", int.class, boolean.class));
	var childMethodSelector = new MethodSelector(TestCase.class, "method", int.class, boolean.class);

	childMethodSelector.getJavaMethod();
	
	assertThat(parentMethodSelector).isNotEqualTo(childMethodSelector);
	assertThat(childMethodSelector).isNotEqualTo(parentMethodSelector);
}
```

marcphilipp: Good catch! Changed in 7b67a8c9ffc348f39cfbd2e3ec8b8fc79c789e4d

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/discovery/MethodSelectorResolver.java
marcphilipp: Done in aa4a84aa4a64d66f97fdbc1225da6b31cbd9d83c

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/MethodSelector.java
mpkorstanje, line 293: We could consider including `declaringClassName` here. 

Though I do think it is an implementation detail, it would make debugging the test easier..

junit-platform-engine/src/main/java/org/junit/platform/engine/discovery/MethodSelector.java
```diff
@@ -75,51 +76,51 @@ public final class MethodSelector implements DiscoverySelector {
...
+		this.parameterTypes = parameterTypes;
+	}
+
+	private MethodSelector(@Nullable ClassLoader classLoader, String className, String methodName,
+			String parameterTypeNames, @Nullable String declaringClassName) {
+		this.classLoader = classLoader;
+		this.className = className;
+		this.methodName = methodName;
+		this.parameterTypeNames = parameterTypeNames;
+		this.declaringClassName = className.equals(declaringClassName) ? null : declaringClassName;
```
mpkorstanje: This echo's the considerations of `formatMethodSpecPart` but here we check for equality on class names rather than package names. This doesn't seem quite right (yet). 

Though the `MethodSelectorTests` uses a private method in the `ParentTestCase` rather than putting the `ParentTestCase` in a different package. So I can't work out if that was intentional.

mpkorstanje (changes requested):
I've pushed 31128d6 to make the build green. But I've got some reservations about the way I've done so.  For a lack of a better word, I would expect a symmetry of concerns between the unique id segment of a method and the method selector. Currently that symmetry is broken.

marcphilipp:
> I've pushed [31128d6](https://github.com/junit-team/junit-framework/commit/31128d6a991afad7a30d62b803ad0344429c0f4c) to make the build green. But I've got some reservations about the way I've done so. For a lack of a better word, I would expect a symmetry of concerns between the unique id segment of a method and the method selector. Currently that symmetry is broken.

Thanks! I have similar reservations. Therefore, I've pushed two commits that solve the issue by introducing a Jupiter-specific, internal `DeclaredMethodSelector`.

@mpkorstanje WDYT?

marcphilipp:
> I would expect that while using the `EngineDiscoveryRequestResolver` I should be able to produce the same test plan by using either one class selector, or a collection of method selectors. Because the `MethodSelector` `equals` implementation lacks fidelity that isn't possible yet.

I agree but the scenario here is truly an edge case and I think we can postpone tackling the adjacent issue. When we do, we also need to think about `NestedMethodSelector`.

mpkorstanje:
> I agree but the scenario here is truly an edge case and I think we can postpone tackling the adjacent issue. When we do, we also need to think about NestedMethodSelector.

And `MethodSource` apparently :smile:

PR #5142, opened 2025-11-07
Contribute DefaultLocaleExtension and DefaultTimeZoneExtension

### LINKED ISSUE #4727: Support set / reset time zone and locale during test execution
Code under test is oftentimes influenced by current locale and time zone. This usually leads to boilerplate code such as

```java
private Locale locale;
private TimeZone timeZone;


@BeforeEach
void setUp() {
  locale = Locale.getDefault();
  Locale.setDefault(Locale.US);

  timeZone = TimeZone.getDefault();
  TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
}

@AfterEach
void tearDown() {
  Locale.setDefault(locale);
  TimeZone.setDefault(timeZone);
}
```

An annotation-based solution would remove the need for such boilerplate code and allow a cleaner code style and thus improve readability. The annotation should be supported on both, class and method level. Further, in order to address concurrency issues `ResourceLock` should be utilized to allow safe manipulation of locale and time zones when tests run in parallel.

There is an existing solution in `junit-pioneer`. https://junit-pioneer.org/docs/default-locale-timezone/ contains more examples for potential use cases.
That being said, while there is a 3rd-party library, I still think that this is a feature that should be included in JUnit already.

I'd be happy to prepare a pull request for this in case there is positve feedback for this feature.

Bukama:
JUnit Pioneer is happy to contribute its DefaultLocaleExtension and DefaultTimeZoneExtension to JUnit Jupiter.

closes #4727

---

---

### Definition of Done

### DIFF
--- a/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
+++ b/documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
@@ -3835,6 +3835,7 @@ work on the test class level and on the test method level, and are inherited fro
 higher-level containers. After the annotated element has been executed, the initial
 default value is restored.
 
+[[writing-tests-built-in-extensions-DefaultLocale]]
 ===== @DefaultLocale
 
 The default `Locale` can be specified using an
@@ -3885,6 +3886,7 @@ include::{testDir}/example/DefaultLocaleTimezoneExtensionDemo.java[tag=default_l
 
 NOTE: The provider implementation must have a no-args (or the default) constructor.
 
+[[writing-tests-built-in-extensions-DefaultTimeZone]]
 ===== @DefaultTimeZone
 
 The default `TimeZone` is specified according to the
--- a/junit-jupiter-api/src/main/java/module-info.java
+++ b/junit-jupiter-api/src/main/java/module-info.java
@@ -34,4 +34,5 @@ module org.junit.jupiter.api {
 	exports org.junit.jupiter.api.util;
 
 	opens org.junit.jupiter.api.condition to org.junit.platform.commons;
+	opens org.junit.jupiter.api.util to org.junit.platform.commons;
 }
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/DefaultLocale.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/DefaultLocale.java
@@ -17,20 +17,21 @@ import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.util.Locale;
 
 import org.apiguardian.api.API;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 /**
- * {@code @DefaultLocale} is a JUnit Jupiter extension to change the value
- * returned by {@link java.util.Locale#getDefault()} for a test execution.
+ * {@code @DefaultLocale} is a JUnit Jupiter extension for changing the value
+ * returned by {@link Locale#getDefault()} for a test execution.
  *
- * <p>The {@link java.util.Locale} to set as the default locale can be
- * configured in several ways:</p>
+ * <p>The {@link Locale} to set as the default locale can be
+ * configured in several ways:
  *
  * <ul>
- *     <li>using a {@link java.util.Locale#forLanguageTag(String) language tag}</li>
- *     <li>using a {@link java.util.Locale.Builder Locale.Builder} together with
+ *     <li>using a {@link Locale#forLanguageTag(String) language tag}</li>
+ *     <li>using a {@link Locale.Builder Locale.Builder} together with
  *         <ul>
  *            <li>a language</li>
  *            <li>a language and a county</li>
@@ -39,8 +40,9 @@ import org.junit.jupiter.api.extension.ExtendWith;
  *     </li>
  * </ul>
  *
- * <p>Please keep in mind that the {@code Locale.Builder} does a syntax check, if you use a variant!
- * The given string must match the BCP 47 (or more detailed <a href="https://www.rfc-editor.org/rfc/rfc5646.html">RFC 5646</a>) syntax.</p>
+ * <p>Please keep in mind that the {@code Locale.Builder} does a syntax check,
+ * if you use a variant. The given string must match the BCP 47 (or more
+ * detailed <a href="https://www.rfc-editor.org/rfc/rfc5646.html">RFC 5646</a>) syntax.
  *
  * <p>If a language tag is set, none of the other fields must be set. Otherwise, an
  * {@link org.junit.jupiter.api.extension.ExtensionConfigurationException} will
@@ -48,24 +50,28 @@ import org.junit.jupiter.api.extension.ExtendWith;
  * {@link #variant()} but no {@link #country()} and {@link #language()} will
  * also cause an {@code ExtensionConfigurationException}. After the annotated
  * element has been executed, the default {@code Locale} will be restored to
- * its original value.</p>
+ * its original value.
  *
  * <p>{@code @DefaultLocale} can be used on the method and on the class level. It
  * is inherited from higher-level containers, but can only be used once per method
  * or class. If a class is annotated, the configured {@code Locale} will be the
  * default {@code Locale} for all tests inside that class. Any method level
- * configurations will override the class level default {@code Locale}.</p>
+ * configurations will override the class level default {@code Locale}.
  *
  * <p>During
- * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
- * all tests annotated with {@link DefaultLocale}, {@link ReadsDefaultLocale}, and {@link WritesDefaultLocale}
- * are scheduled in a way that guarantees correctness under mutation of shared global state.</p>
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution">parallel test execution</a>,
+ * all tests annotated with {@link DefaultLocale @DefaultLocale},
+ * {@link ReadsDefaultLocale @ReadsDefaultLocale}, and
+ * {@link WritesDefaultLocale} are scheduled in a way that guarantees
+ * correctness under mutation of shared global state.
  *
- * <p>For more details and examples, see
- * <a href="https://docs.junit.org/current/user-guide/#writing-tests-built-in-extensions-DefaultLocaleAndTimeZone" target="_top">the documentation on <code>@DefaultLocale</code> and <code>@DefaultTimeZone</code></a>.</p>
+ * <p>For more details and examples, see the
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-built-in-extensions-DefaultLocale">User Guide</a>.
  *
  * @since 6.1
- * @see java.util.Locale#getDefault()
+ * @see Locale#getDefault()
+ * @see ReadsDefaultLocale
+ * @see WritesDefaultLocale
  * @see DefaultTimeZone
  */
 @Retention(RetentionPolicy.RUNTIME)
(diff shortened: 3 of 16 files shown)


documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -3826,3 +3826,101 @@ include::{testDir}/example/AutoCloseDemo.java[tags=user_guide_example]
 <1> Annotate an instance field with `@AutoClose`.
 <2> `WebClient` implements `java.lang.AutoCloseable` which defines a `close()` method that
     will be invoked after each `@Test` method.
+
+[[writing-tests-built-in-extensions-DefaultLocalAndTimezone]]
+==== The @DefaultLocal and @DefaultTimezone Extensions
```
marcphilipp:
```suggestion
[[writing-tests-built-in-extensions-DefaultLocaleAndTimeZone]]
==== The @DefaultLocale and @DefaultTimeZone Extensions
```

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -3826,3 +3826,101 @@ include::{testDir}/example/AutoCloseDemo.java[tags=user_guide_example]
...
+include::{testDir}/example/DefaultLocaleTimezoneExtensionDemo.java[tags=default_locale_language]
+----
+
+The `@DefaultLocale` and `@DefaultTimeZone` annotations can be used to change the values returned from `Locale.getDefault()` and `TimeZone.getDefault()`, respectively, which are often used implicitly when no specific locale or time zone is chosen.
+Both annotations work on the test class level and on the test method level, and are inherited from higher-level containers.
+After the annotated element has been executed, the initial default value is restored.
+
+===== `@DefaultLocale`
+
+The default `Locale` can be specified using an https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html#forLanguageTag-java.lang.String-[IETF BCP 47 language tag string]
```
marcphilipp:
```suggestion
The default `Locale` can be specified using an {jdk-javadoc-base-url}/java.base/java/util/Locale.html#forLanguageTag-java.lang.String-[IETF BCP 47 language tag string]
```

documentation/src/docs/asciidoc/user-guide/writing-tests.adoc
```diff
@@ -3826,3 +3826,101 @@ include::{testDir}/example/AutoCloseDemo.java[tags=user_guide_example]
...
+
+[source,java,indent=0]
+----
+include::{testDir}/example/DefaultLocaleTimezoneExtensionDemo.java[tags=default_locale_language]
+----
+
+Alternatively the default `Locale` can be created using the following attributes of which a https://docs.oracle.com/javase/8/docs/api/java/util/Locale.Builder.html[Locale Builder] can create an instance with:
```
marcphilipp:
```suggestion
Alternatively the default `Locale` can be created using the following attributes of which a {jdk-javadoc-base-url}/java.base/java/util/Locale.Builder.html[Locale Builder] can create an instance with:
```

jupiter-tests/src/test/java/org/junit/jupiter/engine/extension/DefaultTimeZoneTests.java
```diff
@@ -0,0 +1,378 @@
...
+		@Test
+		@ReadsDefaultTimeZone
+		@DisplayName("on method level, throws exception")
+		void throwsWhenConfigurationIsBad() {
+			ExecutionResults results = executeTestMethod(BadMethodLevelConfigurationTestCases.class,
+				"badConfiguration");
+
+			assertThat(results).hasSingleFailedTest().withExceptionInstanceOf(
+				ExtensionConfigurationException.class).hasMessageNotContaining(
+					"should never execute").hasMessageContaining("@DefaultTimeZone not configured correctly.");
```
marcphilipp: Please remove the classes in `org.junit.jupiter.testkit.assertion.**` and use "regular" `EngineTestKit` assertions:

```suggestion
			results.testEvents().assertThatEvents() //
					.haveExactly(1, finishedWithFailure(instanceOf(ExtensionConfigurationException.class),
						message(it -> it.contains("@DefaultTimeZone not configured correctly."))));
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/DefaultTimeZoneExtension.java
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	private static TimeZone createTimeZone(String timeZoneId) {
+		TimeZone configuredTimeZone = TimeZone.getTimeZone(timeZoneId);
+		// TimeZone::getTimeZone returns with GMT as fallback if the given ID cannot be understood
+		if (configuredTimeZone.equals(TimeZone.getTimeZone("GMT")) && !timeZoneId.equals("GMT")) {
+			throw new ExtensionConfigurationException(String.format(
+				"@DefaultTimeZone not configured correctly. " + "Could not find the specified time zone + '%s'. "
+						+ "Please use correct identifiers, e.g. \"GMT\" for Greenwich Mean Time.",
+				timeZoneId));
```
marcphilipp:
```suggestion
			throw new ExtensionConfigurationException("""
					@DefaultTimeZone not configured correctly. \
					Could not find the specified time zone '%s'. \
					Please use correct identifiers, e.g. "GMT" for Greenwich Mean Time.""".formatted(timeZoneId));
```

jupiter-tests/src/test/java/org/junit/jupiter/api/util/DefaultLocaleTests.java
```diff
@@ -0,0 +1,515 @@
...
+		public Locale get() {
+			return Locale.FRENCH;
+		}
+
+	}
+
+	static class ReturnsNullLocaleProvider implements LocaleProvider {
+
+		@Override
+		public Locale get() {
```
marcphilipp, line 498:
```suggestion
		@SuppressWarnings("NullAway")
		public Locale get() {
```

jupiter-tests/src/test/java/org/junit/jupiter/api/util/DefaultTimeZoneTests.java
```diff
@@ -0,0 +1,378 @@
...
+		public TimeZone get() {
+			return TimeZone.getTimeZone("Europe/Prague");
+		}
+
+	}
+
+	static class NullProvider implements TimeZoneProvider {
+
+		@Override
+		public TimeZone get() {
```
marcphilipp, line 358:
```suggestion
		@SuppressWarnings("NullAway")
		public TimeZone get() {
```

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/DefaultTimeZoneExtension.java
```diff
@@ -0,0 +1,106 @@
...
+		if (noValue == noProvider)
+			throw new ExtensionConfigurationException(
+				"Either a valid time zone id or a TimeZoneProvider must be provided to "
+						+ DefaultTimeZone.class.getSimpleName());
+	}
+
+	private static TimeZone createTimeZone(String timeZoneId) {
+		TimeZone configuredTimeZone = TimeZone.getTimeZone(timeZoneId);
+		// TimeZone::getTimeZone returns with GMT as fallback if the given ID cannot be understood
+		if (configuredTimeZone.equals(TimeZone.getTimeZone("GMT")) && !timeZoneId.equals("GMT")) {
```
marcphilipp:
```suggestion
		if (configuredTimeZone.equals(TimeZone.getTimeZone("GMT")) && !"GMT".equals(timeZoneId)) {
```

jupiter-tests/src/test/java/org/junit/jupiter/engine/extension/DefaultTimeZoneTests.java
```diff
@@ -0,0 +1,378 @@
...
+			return null;
+		}
+
+	}
+
+	static class ComplicatedProvider implements TimeZoneProvider {
+
+		private final String timeZoneString;
+
+		public ComplicatedProvider(String timeZoneString) {
```
marcphilipp:
```suggestion
		ComplicatedProvider(String timeZoneString) {
```

marcphilipp (commented):
Thanks for this initial push! :+1: 

> How to use JUnit's internal testkit

We just use `EngineTestKit` directly, without the abstraction that Pioneer has. I commented on one such test.

> Why checkstyle fails just right after `spotless:apply` goal was run

Checkstyle can fail even though Spotless formatted the code. Spotless is purely about formatting, while we use Checkstyle for some more semantic checks.

> That ArchUnit rules are in place

What did you want to check for?

Bukama:
Thank you for your feedback, I think (hope) I applied all of it correctly. I find the testkit methods very confusing and verbose, but I guess that's a thing how used you are to them.

> What did you want to check for?
Nothing, I saw that I have violations against ArchUnit's rules, but was too tired to check which I violate (turned out it was the missing `@API` annotations.

In general I'm done with most parts, I now have to figure out which are the correct `findAnnotation` methods in JUpiter to match those of Pioneer. Especially those with nested and class level ones.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/DefaultLocaleExtension.java
```diff
@@ -0,0 +1,130 @@
...
+class DefaultLocaleExtension implements BeforeEachCallback, AfterEachCallback {
+
+	private static final Namespace NAMESPACE = Namespace.create(DefaultLocaleExtension.class);
+
+	private static final String KEY = "DefaultLocale";
+
+	@Override
+	public void beforeEach(ExtensionContext context) {
+		AnnotatedElement element = context.getElement().orElse(null);
+		AnnotationSupport.findAnnotation(element, DefaultLocale.class).ifPresent(
```
marcphilipp: This won't find annotations on top-level and nested classes (which is why the test fail). You can either search the hierarchy upward (like the Pioneer implementation does) or implement `BeforeAllCallback` and store the value in the class-level store so it can be used as a fallback here.

marcphilipp (changes requested):

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/DefaultLocaleExtension.java
marcphilipp: As discussed via email, I've pushed a commit that uses `BeforeAllCallback` to achieve this. The second commit moves the extension implementation to junit-jupiter-api to avoid its overhead since it should only be active when the `@DefaultLocale` annotation is used.

@Bukama Could you please do the same for `@DefaultTimeZone`?

Bukama: Thank you for your help. Applied it to `@DefaultTimeZone`, but need to figure out, where I have a circular dependency (cause ArchUnit-Report is truncated). 

Had to laugh when reading your move, cause the first thing I did when porting was to split api and extension, because I understood your comment in the issue to do so :D

Another question: As you released 6.1-RC1 today, I guess I should switch the `@since` to 6.2?


edit: Will search again tomorrow, I'm tired after 16h awake

marcphilipp: > Had to laugh when reading your move, cause the first thing I did when porting was to split api and extension, because I understood your comment in the issue to do so :D

Sorry about that!

> Another question: As you released 6.1-RC1 today, I guess I should switch the `@since` to 6.2?

I only released a milestone so some of the features that are already in can be tested sooner. This PR is still targeted for 6.1. Please add a release note entry for 6.1.0-M2 (you'll need to rebase/merge first).

Bukama:
To keep you up to date:
Putting the classes in the `api.util` packages results in a circular dependency of `api.extension` due the TimeoutUtil class in the same package. For testing purpose I moved all of them to a dummy package (will split and rename it later) to make progress on getting the build green. I then spent some time on the module-info, but also got this now, but now test that check the module-info break. But it's far too late today. Will continue tomorrow or Saturday.

marcphilipp:
> Putting the classes in the `api.util` packages results in a circular dependency of `api.extension` due the TimeoutUtil class in the same package.

Hmm, I hadn't thought of that. We should probably move `TimeoutUtil` to a separate package. Let's go with `org.junit.jupiter.api.timeout` for now, ok?

marcphilipp:
> Hmm, I hadn't thought of that. We should probably move `TimeoutUtil` to a separate package. Let's go with `org.junit.jupiter.api.timeout` for now, ok?

Actually, I've just done this on main so it should suffice for you to rebase this PR.

Bukama:
> > Hmm, I hadn't thought of that. We should probably move `TimeoutUtil` to a separate package. Let's go with `org.junit.jupiter.api.timeout` for now, ok?
> 
> Actually, I've just done this on main so it should suffice for you to rebase this PR.

Great, I will work later this day / tomorrow

Bukama:
I give up. After wasting another few hours for getting those tests green which check module system files that do nowhere exists and complain about modules would not exists while the compiler says they do.

platform-tooling-support-tests/src/test/java/platform/tooling/support/tests/ModularUserGuideTests.java
```diff
@@ -49,6 +49,8 @@ class ModularUserGuideTests {
 			  exports example.testkit; // just here to ensure documentation example sources are compiled
 
 			  requires org.junit.jupiter.api;
+			  requires org.junit.jupiter.api.locale;
+			  requires org.junit.jupiter.api.timezone;
```
sormuras: Those two are packages, right?

Here, only modules are expected. Thus, `requires org.junit.jupiter.api;` already takes care of providing the new two packages contained in it.
```suggestion
```

Bukama:
Thanks @sormuras for pointing out the obvious 🤦‍♂️ When you don't see the forest because of trees...

junit-jupiter-api/src/main/java/org/junit/jupiter/api/timezone/DefaultTimeZone.java
```diff
@@ -0,0 +1,75 @@
...
+ * Copyright 2015-2025 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api.timezone;
```
marcphilipp: We discussed the package names in the last team call and would like to stick with a single `org.junit.jupiter.api.util` package for both extensions.

jupiter-tests/src/test/java/org/junit/jupiter/testkit/ExecutionResults.java
```diff
@@ -0,0 +1,152 @@
...
+import org.junit.platform.testkit.engine.EngineExecutionResults;
+import org.junit.platform.testkit.engine.EngineTestKit;
+import org.junit.platform.testkit.engine.Events;
+
+/**
+ * Pioneers' class to handle JUnit Jupiter's {@link org.junit.platform.testkit.engine.EngineExecutionResults}.
+ *
+ * <p>Instantiate with the static factory methods in {@link JUnitJupiterTestKit}.
+ */
+public class ExecutionResults {
```
marcphilipp: I would like to introduce a new mechanism for this. We have a `AbstractJupiterTestEngineTests` base class that should provide a few helper methods that you can use instead.

marcphilipp (changes requested):
Di you have time to make the requested changes? If not, please let me know and I can take over.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/timezone/DefaultTimeZone.java
Bukama: Moved

jupiter-tests/src/test/java/org/junit/jupiter/testkit/ExecutionResults.java
Bukama: Will have a look on Tuesday or maybe Thursday/Weekend. But feel free to take over

Bukama: done

marcphilipp (approved):
Thanks for all the hard work! I polished the documentation a bit and pushed two commits.

PR #5158, opened 2025-11-17
Prune exceptions when JUnit.start is used

mpkorstanje:
When [using `JUnit.start`](https://github.com/junit-team/junit-examples/pull/656) the `pruneStackTrace` algorithm immediately sees the `HelloTest.main` frame and assumes that this is the test method because the test class name matches.

```
org.opentest4j.AssertionFailedError: expected: <11> but was: <12>
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:158)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:139)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:201)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:152)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:147)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:558)
	at com.examp.project/com.example.project.HelloTest.stringLength(HelloTest.java:14)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
...
	at org.junit.platform.launcher@6.1.0-SNAPSHOT/org.junit.platform.launcher.core.SessionPerRequestLauncher.execute(SessionPerRequestLauncher.java:81)
	at org.junit.start@6.1.0-SNAPSHOT/org.junit.start.JUnit.run(JUnit.java:63)
	at org.junit.start@6.1.0-SNAPSHOT/org.junit.start.JUnit.run(JUnit.java:37)
	at com.examp.project/com.example.project.HelloTest.main(HelloTest.java:9)
```

By checking if `org.junit.start` is involved further down the stack we exclude this scenario.

---

[...]

PR #5159, opened 2025-11-17
Trim internal frames from AssertionFailedError

mpkorstanje:
When using `JUnit.start` and creating a failing test, users will be confronted with a large stacktrace with mostly irrelevant information. Even after #5158 is merged, the stacktrace will contain several internal frames:

```
org.opentest4j.AssertionFailedError: expected: <11> but was: <12>
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:158)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:139)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:201)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:152)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:147)
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:558)
	at com.examp.project/com.example.project.HelloTest.stringLength(HelloTest.java:14)
```

By pruning these internal frames, the stacktrace can be reduced to a much more readable:

```
org.opentest4j.AssertionFailedError: expected: <11> but was: <12>
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:558)
	at com.examp.project/com.example.project.HelloTest.stringLength(HelloTest.java:14)
```

The pruning functionality is intentionally added to the `AssertionFailureBuilder` rather than the `ExceptionUtils` to enable other users to also prune the internal frames from their own assertions.

---

---

### Definition of Done

### DIFF
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
@@ -49,8 +49,8 @@ repository on GitHub.
   xref:writing-tests/built-in-extensions.adoc#DefaultLocaleAndTimeZone[User Guide].
 * Trim internal stack frames from `AssertionFailedError` stack traces.
 * Introduce new `trimStacktrace(Class<?>)` and `retainStackTraceElements(int)`
-methods for `AssertionFailureBuilder`. These allow user defined assertions to
-trim their stacktrace.
+  methods for `AssertionFailureBuilder`. These allow user defined assertions to
+  trim their stacktrace.
 
 [[v6.1.0-M2-junit-vintage]]
 === JUnit Vintage


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
```diff
@@ -154,9 +171,37 @@ public AssertionFailedError build() {
...
+
+		var assertionFailedError = mismatch //
 				? new AssertionFailedError(message, expected, actual, cause) //
 				: new AssertionFailedError(message, cause);
+
+		maybeTrimStackTrace(assertionFailedError);
+		return assertionFailedError;
+	}
+
+	private void maybeTrimStackTrace(Throwable throwable) {
```
mpkorstanje, line 211: This needs some boundary testing.

platform-tests/src/test/java/org/junit/platform/StackTracePruningTests.java
```diff
@@ -92,13 +92,12 @@ void shouldAlwaysKeepJupiterAssertionStackTraceElement() {
...
-				>>>>
-				\\Qorg.junit.jupiter.api.Assertions.fail(Assertions.java:\\E.+
+				\\Qorg.junit.platform.StackTracePruningTests$FailingTestTestCase.fail(StackTracePruningTests.java:\\E.+
 				>>>>
 				""");
 	}
 
-	@Test
+	@Test //TODO:
 	void shouldAlwaysKeepJupiterAssumptionStackTraceElement() {
```
mpkorstanje, line 101: Should assumptions also be trimmed?

marcphilipp:
I think we should preserve the bottommost `Assertions`/`Assumptions` stack frame:

```
org.opentest4j.AssertionFailedError: expected: <11> but was: <12>
	at org.junit.jupiter.api@6.1.0-SNAPSHOT/org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:558)
	at com.examp.project/com.example.project.HelloTest.stringLength(HelloTest.java:14)
```

mpkorstanje:
I'm not convinced that this extra frame is useful. Users will never visit it, because the frame above it is their code and the place they'll ultimately have to fix the issue.

```java
public class HelloTest {

    @Test
    void stringLengthJupiter() {
        org.junit.jupiter.api.Assertions.assertEquals(11, "Hello JUnit2".length());
    }
    @Test
    void stringLengthAssertJ() {
        org.assertj.core.api.Assertions.assertThat("Hello JUnit2".length()).isEqualTo(11);
    }
}
```

And looking at other projects: IDEA already filters it out. And AssertJ doesn't include it in the first place. edit: As does Google Truth.

Screenshot From 2025-11-18 20-37-41

On a more technical note, for the exceptions thrown by the `infline fun` in `Assertions.kt` the stack frames must either start at `AssertionFailureBuilder.build` or the test method. In which case the latter seems preferable.

What benefits do you think this extra frame adds?

marcphilipp:
> What benefits do you think this extra frame adds?

To me, the extra frame makes it explicit that a specific assertion method failed. Let's discuss with the rest of the team, though.

platform-tests/src/test/java/org/junit/platform/StackTracePruningTests.java
mpkorstanje: I'm leaving these out of scope for now.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
```diff
@@ -130,6 +137,25 @@ public AssertionFailureBuilder includeValuesInMessage(boolean includeValuesInMes
...
+	 * <p>Removes all but {@code retain - 1} frames before the last frame from
+	 * {@code target}. If {@code retain} is zero, all frames including
+	 * {@code target} are trimmed.
+	 *
+	 * @param target class to trim from the stacktrace
+	 * @param retain depth of trimming, must be non-negative
+	 * @return this builder for method chaining
+	 */
+	@API(status = EXPERIMENTAL, since = "6.1")
+	public AssertionFailureBuilder trimStacktrace(@Nullable Class<?> target, int retain) {
```
sormuras: Would it make sense to overload this with a "pass 1" variant?

```java
public AssertionFailureBuilder trimStacktrace(@Nullable Class<?> target) {
    return trimStacktrace(target, 1);
}
```

mpkorstanje: I've thought about that. But the "natural" default value seems to be 0. We are in most places keeping one extra. So though duplicated quite often (but not always), I think we should keep that.

On reflection I do have my doubts about putting both parameters in a single method though. For a builder, this really should be two separate methods.

mpkorstanje: > On reflection I do have my doubts about putting both parameters in a single method though. For a builder, this really should be two separate methods.

It does force the design towards using a strategy pattern but I don't think we have enough design constraints for that yet. So I've just extracted a new method for the second argument and set the default value to 1.

PR #5186, opened 2025-11-30
Exclude competing `@Deprecated` factory methods in fallback String-to-Object converter

### LINKED ISSUE #4996: Implicit conversion should avoid `@Deprecated` methods
[Implicit conversion fallback](https://docs.junit.org/current/user-guide/#writing-tests-parameterized-tests-argument-conversion-implicit-fallback) allows JUnit to convert a `String` to an arbitrary `Object`. It does this by [searching](https://github.com/junit-team/junit-framework/issues/1231) for a constructor or factory method taking one `String` (with a further fallback to look for `CharSequence` [added later](https://github.com/junit-team/junit-framework/issues/4815)).

This approach recently caused a problem for me. In a piece of shared code there was a factory method named `parseString` which I wanted to rename to the more standard `parse`. I did this by copying the method and marking `parseString` as deprecated (for removal). I didn't expect this to cause any issues, as it is a standard approach to refactoring a change like this from a shared code library. In the downstream code, someone was using the type via JUnit implicit argument conversion, and the test failed because JUnit now saw the class had 2 static factory methods where previously there was 1.

I'd like to propose a tweak to the algorithm.

1. Find all factory methods with a single `String` parameter, if there is only one, then use it
2. Find all factory methods with a single `CharSequence` parameter, if there is only one, then use it
3. (new step) Repeat step 1, removing any method marked as deprecated, if only one method then use it
4. (new step) Repeat step 2, removing any method marked as deprecated, if only one method then use it.

I believe this approach is backwards compatible, and would be a good enhancement. Thanks for your consideration.

jodastephen:
This extends the fallback String to Object converter algorithm to exclude deprecated methods. This occurs after the existing algorithm finished. As such, any code that works today will continue to work. It is possible that this PR will make some code that fails today start to work, but that seems to be an acceptable outcome.

Release notes will need adding.

Fixes #4996

---

---

### Definition of Done


mpkorstanje:
@jodastephen could you satisfy the DCO bot?

Oh and please ignore co-pilot. I'm not sure where that came from, but it is not something from the JUnit team. :thinking:

junit-platform-commons/src/main/java/org/junit/platform/commons/support/conversion/FallbackStringToObjectConverter.java
```diff
@@ -120,9 +125,22 @@ public boolean canConvertTo(Class<?> targetType) {
...
-		List<Method> factoryMethods = findMethods(targetType, new IsFactoryMethod(targetType, parameterType),
-			BOTTOM_UP);
+	private static @Nullable Function<String, @Nullable Object> findFactoryExcludeDeprecated(Class<?> targetType) {
+		Method factoryMethodStr = findFactoryMethod(targetType, String.class, true);
+		if (factoryMethodStr != null) {
+			return source -> invokeMethod(factoryMethodStr, null, source);
+		}
+		Method factoryMethodChSeq = findFactoryMethod(targetType, CharSequence.class, true);
+		if (factoryMethodChSeq != null) {
+			return source -> invokeMethod(factoryMethodChSeq, null, source);
```
mpkorstanje: Because `Book2` has deprecated factory methods for both `String` and `CharSequence` the path involving `Charsequence` is never used.

mpkorstanje (changes requested):

mpkorstanje:
Note: Manually set DCO check to pass. PR was offered before DCO bot was enabled.

jodastephen:
Thanks, especially for the DCO part!

PR #5189, opened 2025-12-01
[prone] fix `UnnecessarilyFullyQualified`

Pankraz76:
- https://errorprone.info/bugpattern/UnnecessarilyFullyQualified

related to:
- https://github.com/gradle/gradle/pull/35882

again this some change to be picked by maintainer if got some spare time to recreate:

image

---

---

### Definition of Done

### DIFF
--- a/gradle/plugins/common/src/main/kotlin/junitbuild.java-errorprone-conventions.gradle.kts
+++ b/gradle/plugins/common/src/main/kotlin/junitbuild.java-errorprone-conventions.gradle.kts
@@ -1,6 +1,7 @@
 import junitbuild.extensions.dependencyFromLibs
 import net.ltgt.gradle.errorprone.errorprone
 import net.ltgt.gradle.nullaway.nullaway
+import org.gradle.jvm.toolchain.JvmImplementation.J9
 
 plugins {
 	`java-library`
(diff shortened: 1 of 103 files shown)


gradle/plugins/common/src/main/kotlin/junitbuild.java-errorprone-conventions.gradle.kts
```diff
@@ -16,17 +17,10 @@ dependencies {
...
-		val shouldDisableErrorProne = java.toolchain.implementation.orNull == JvmImplementation.J9
-		if (name == "compileJava" && !shouldDisableErrorProne) {
-			disable(
-				"AnnotateFormatMethod", // We don`t want to use ErrorProne's annotations.
-				"BadImport", // This check is opinionated wrt. which method names it considers unsuitable for import which includes a few of our own methods in `ReflectionUtils` etc.
-				"DoNotCallSuggester", // We don`t want to use ErrorProne's annotations.
-				"ImmutableEnumChecker", // We don`t want to use ErrorProne's annotations.
-				"InlineMeSuggester", // We don`t want to use ErrorProne's annotations.
-				"MissingSummary", // Produces a lot of findings that we consider to be false positives, for example for package-private classes and methods.
-				"StringSplitter", // We don`t want to use Guava.
```
Pankraz76, line 28: dont need to disable warnings they are already treaded like this for a reason by convention.

gradle/plugins/common/src/main/kotlin/junitbuild.java-errorprone-conventions.gradle.kts
```diff
@@ -16,17 +17,10 @@ dependencies {
...
-				"UnnecessaryLambda", // The findings of this check are subjective because a named constant can be more readable in many cases.
+		val enableErrorProne = java.toolchain.implementation.orNull != J9
+		if (name == "compileJava" && enableErrorProne) {
+			disableAllWarnings = true // considering this immense spam burden, remove this once to fix dedicated flaw. https://github.com/diffplug/spotless/pull/2766
+			disable( // We don`t want to use ErrorProne's annotations.
 				// picnic (https://error-prone.picnic.tech)
 				"ConstantNaming",
```
Pankraz76, line 25: therefore this still seems an bug, as need to disable picnic warnings, as they fail the build if not treated the extra way?

isn´t it? @rickie

Pankraz76:
bare prone errors seem not patchable somehow, only dedicated picnic stuff is like:

- https://github.com/google/error-prone/issues/5277
- https://error-prone.picnic.tech/bugpatterns/NonStaticImport


considering reopening after enabler and local check.

thanks for the effort invested and revealed tweak.

gradle/plugins/common/src/main/kotlin/junitbuild.java-errorprone-conventions.gradle.kts
rickie: Hi @Pankraz76,

Thanks for flagging the issues with the Picnic warnings in these various tickets, such as https://github.com/google/error-prone/issues/5365.

I want to emphasize that I appreciate you taking the time to apply our checks. It is great to see them being applied in the wild.

That said, to make these reports actionable, we really need a **minimal reproduction case** when things break. As noted by @cushon here https://github.com/google/error-prone/issues/5365#issuecomment-3562707386 and here https://github.com/google/error-prone/issues/5277#issuecomment-3495778796, simply knowing that it fails in a repository makes it very hard for us to debug. 

If we can isolate the specific code causing the crash, we can fix the root cause instead of just disabling the check. To be clear, I do believe you could be right and that there is a bug in our code or in Error Prone itself. If so, we definitely want to fix it, but we need those isolated examples to make that happen.

PR #5205, opened 2025-12-05
Remove UniqueId.uniqueIdFormat field

mpkorstanje:
The `UniqueIdFormat` is a singleton instance and not part of the public API, as such is effectively constant. Tracking the format in every `UniqueId` is unnecessary and makes the `UniqueId` object larger than it needs to be.

In 2016 in [0e5ab97](https://github.com/junit-team/junit-framework/commit/0e5ab970ea76cc5dbbf7c8b1115542be94609bf9) the `UniqueId.uniqueIdFormat` was added with the message:
> - UniqueId now tracks the UniqueIdFormat it was created with in order to avoid accidentally switching to the default format when appending new segments to an existing UniqueId.

But at the time UniqueIdFormat was public and not a singleton.

---

---

### Definition of Done

### DIFF
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
@@ -25,7 +25,8 @@ repository on GitHub.
 [[v6.1.0-M2-junit-platform-new-features-and-improvements]]
 ==== New Features and Improvements
 
-* ❓
+* The `UniqueId.uniqueIdFormat` field has been removed, reducing the size of `UniqueId`
+  objects.
 
 [[v6.1.0-M2-junit-jupiter]]
 === JUnit Jupiter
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Open questions for _team discussion_:

* Can we change the serial version of `UniqueId`?

marcphilipp:
> Can we change the serial version of `UniqueId`?

We should _not_ do so in a minor release. We could work around this by introducing custom `writeObject`/`readObject` methods (like we used to have for `TestIdentifier` in 5.x) that write `null` and ignore any read value.

mpkorstanje:
> We should do so in a minor release. We could work around this by introducing custom writeObject/readObject methods

Did you mean that as a workaround for the limitation of releasing this in a minor release and as a condition for a patch release, or as a work around for changing the serial version and as a condition for a minor release rather than a major release? Or can this go into the next minor release without custom methods?

marcphilipp:
Sorry, there was a "not" missing in my first sentence. :facepalm:

PR #5209, opened 2025-12-07
Fix concurrency bug in `NamespacedHierarchicalStore.computeIfAbsent`

### LINKED ISSUE #5171: Concurrency problem in `NamespacedHierarchicalStore#computeIfAbsent`
## Steps to reproduce
1. Clone https://github.com/assertj/assertj locally
2. Setup project according to `CONTRIBUTING.md`
3. In line: https://github.com/assertj/assertj/commit/f90a09a25ba23ee077dbd1719acd643c7fea4da0#diff-5d92da3c93616cd468991f168636c72526bc929869ee121efa0a1478600c950cR45 remove the `@Disabled` annotation
4. Execute `./mvnw clean verify`
## Expected

Tests run successfully, all tests pass.

## Actual

When running all tests, the test in `SoftAssertionsExtension_PER_CLASS_Concurrency_Test.java` fails sometimes, and on some machines. When running it individually using the maven wrapper and through the IDE, it passes consistently, which is why we believe it to be a concurrency issue.

The failure has the following output:
```
[ERROR] Failures: 
[ERROR]   SoftAssertionsExtension_PER_CLASS_Concurrency_Test.concurrent_tests_with_explicit_per_class_annotation_do_not_interfere:116 Test Event Statistics (2 failures)
        org.opentest4j.AssertionFailedError: started ==> expected: <2> but was: <0>                                                                                                                      
        org.opentest4j.AssertionFailedError: failed ==> expected: <2> but was: <0>            
```

We identified that the change that introduced the flakiness is this line: https://github.com/assertj/assertj/commit/bcb4e14275da62fcb26e2d1e5608c3ddd28020b8#diff-1fd14641c6731a83101088806368089b7af173b3a7424a504654008cab1a0dddR411 

Here it was changed from using the deprecated JUnit method `getOrComputeIfAbsent` to `computeIfAbsent` (according to the JUnit Javadoc). Reverting the change in this line only already resolves the flakiness.

The test in `SoftAssertionsExtension_PER_CLASS_Concurrency_Test.java` executes soft assertions concurrently, thus also executing  `getOrComputeIfAbsent`/`computeIfAbsent` concurrently, which suggests that there is a concurrency issue in `computeIfAbsent` that wasn't in `getOrComputeWithAbsent`.

## Analysis

We noticed one of the substantial changes from the `getOrComputeIfAbsent` implementation to the `computeIfAbsent` implementation is that the call on the `ConcurrentMap` was changed from `computeIfAbsent` to `compute`, see:
https://github.com/junit-team/junit-framework/blob/19557e4d5d3a504b02eaee8c5562c31c2d491b10/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java#L247

Our hypothesis is, that if two threads are executing `NamespacedHierarchicalStore#computeIfAbsent` in parallel, then if both threads are executing at the same speed where both end up executing `this.storedValues.compute(...` at the same time, the first thread will execute properly, while the second thread will overwrite the value in the Map. Previously in `getOrComputeIfAbsent`, since `this.storedValues.computeIfAbsent` was used, it would guarantee that the second thread would not overwrite the value, as the `ConcurrentMap` is enforcing synchronization properly.

We were able to verify that temporarily changing the offending line in `NamespacedHierarchicalStore#computeIfAbsent` to use `computeIfAbsent` instead of `compute`, and then publishing it to Maven local, using that build in AssertJ reliably resolves the flakiness even when running the tests in parallel and using `getStore(context).computeIfAbsent` in https://github.com/assertj/assertj/commit/bcb4e14275da62fcb26e2d1e5608c3ddd28020b8#diff-1fd14641c6731a83101088806368089b7af173b3a7424a504654008cab1a0dddR411
You can find this quick and dirty change we made where the issue no longer occurred here: https://github.com/danaebu/junit-framework/commit/9384b94e27ee1ee444b186bb5ff8185f57c71919

See related issue on the flakiness in AssertJ here: https://github.com/assertj/assertj/issues/1996

## Context

 - Used versions (Jupiter/Vintage/Platform): 6.0.1
 - Build Tool/IDE: Maven/IntelliJ (but can be reproduced outside the IDE by using the maven commands on the terminal as well)

As a `ConcurrentMap` is used, we assume the implementation `NamespacedHierarchicalStore#computeIfAbsent` is supposed to be thread-safe.

Was found together with @martinfrancois at [Hack Commit Push](https://switzerland2025.hack-commit-pu.sh/)

## Deliverables

- [ ] `NamespacedHierarchicalStore#computeIfAbsent` is thread-safe
- [ ] Unit test is added to JUnit as well that reproduces this issue reliably to avoid regressions.

martinfrancois:
This PR fixes a concurrency bug in `NamespacedHierarchicalStore.computeIfAbsent` where `defaultCreator` was executed while holding the internal map's bucket lock, causing:
1. Threads accessing different keys in the same hash bucket to block each other during parallel test execution
2. Potential deadlocks when `defaultCreator` accesses other keys that collide in the same hash bucket

## Root Cause

The previous implementation called `defaultCreator.apply(key)` inside `ConcurrentHashMap.compute()`, which holds the bucket lock. Any thread trying to access a *different* key in the same bucket was blocked until `defaultCreator` completed.

## Fix

The fix moves `defaultCreator` execution outside the lock:
- A placeholder (`MemoizingSupplier` or `DeferredSupplier`) is installed via the map operation
- The actual computation runs after the map operation completes
- Failed computations are cleaned up, and `get()` returns `null` (not the exception) during the transient window

**Note:** Two implementation approaches are provided as separate commits. See the comment below for details on each approach.

### Implementation
- Reworked `computeIfAbsent` to execute `defaultCreator` outside the bucket lock
- Ensured `get()` doesn't observe transient exceptions from failed `computeIfAbsent` calls
- Preserved "one initialization per key" semantics

### Tests
- `simulateRaceConditionInComputeIfAbsentWithCollidingKeys`: Verifies correct initialization under contention with colliding keys
- `computeIfAbsentWithCollidingKeysDoesNotBlockConcurrentAccess`: Verifies no blocking between colliding keys
- `computeIfAbsentDoesNotDeadlockWithCollidingKeys`: Verifies no deadlock with colliding keys
- `getOrComputeIfAbsentDoesNotDeadlockWithCollidingKeys`: Same for deprecated method
- `getDoesNotSeeTransientExceptionFromComputeIfAbsent`: Verifies atomicity for transient failures
- `getConcurrentWithFailingComputeIfAbsentDoesNotSeeException`: Stress test for atomicity
- `computeIfAbsentOverridesParentNullValue`: Verifies parent/child store semantics
- Added `CollidingKey` helper to force hash collisions for testing

Fixes #5171

This should also fix the flakiness in AssertJ's `SoftAssertionsExtension_PER_CLASS_Concurrency_Test` (https://github.com/assertj/assertj/issues/1996).

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.FutureTask;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.function.Supplier;
@@ -237,32 +239,82 @@ public final class NamespacedHierarchicalStore<N> implements AutoCloseable {
 	 * closed
 	 * @since 6.0
 	 */
-	@SuppressWarnings("ReferenceEquality")
 	@API(status = MAINTAINED, since = "6.0")
 	public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? extends V> defaultCreator) {
 		Preconditions.notNull(defaultCreator, "defaultCreator must not be null");
 		CompositeKey<N> compositeKey = new CompositeKey<>(namespace, key);
-		StoredValue storedValue = getStoredValue(compositeKey);
-		var result = StoredValue.evaluateIfNotNull(storedValue);
-		if (result == null) {
-			StoredValue newStoredValue = storedValues.compute(compositeKey, (__, oldStoredValue) -> {
-				if (oldStoredValue == null || oldStoredValue == storedValue) {
-					return newStoredValue(new MemoizingSupplier(() -> {
-						rejectIfClosed();
-						return Preconditions.notNull(defaultCreator.apply(key), "defaultCreator must not return null");
-					}, true));
+
+		// CAS-retry loop: retry if another thread concurrently modifies the entry
+		for (;;) {
+			StoredValue localStoredValue = this.storedValues.get(compositeKey);
+			if (localStoredValue != null) {
+				Object localValue = evaluateForComputeIfAbsent(compositeKey, localStoredValue);
+				if (localValue != null) {
+					return localValue;
 				}
-				return oldStoredValue;
-			});
+
+				Object computed = computeAndInstall(compositeKey, localStoredValue, key, defaultCreator);
+				if (computed != null) {
+					return computed;
+				}
+				continue;
+			}
+
+			// No local mapping: consult parent first.
+			if (this.parentStore != null) {
+				StoredValue parentStoredValue = this.parentStore.getStoredValue(compositeKey);
+				Object parentValue = StoredValue.evaluateIfNotNull(parentStoredValue);
+				if (parentValue != null) {
+					return parentValue;
+				}
+			}
+
+			Object computed = computeAndInstall(compositeKey, null, key, defaultCreator);
+			if (computed != null) {
+				return computed;
+			}
+		}
+	}
+
+	private @Nullable Object evaluateForComputeIfAbsent(CompositeKey<N> compositeKey, StoredValue storedValue) {
+		Supplier<@Nullable Object> supplier = storedValue.supplier();
+		if (supplier instanceof DeferredSupplier deferred) {
+			deferred.run();
 			try {
-				return requireNonNull(newStoredValue.evaluate());
+				return deferred.getOrThrow();
 			}
 			catch (Throwable t) {
-				storedValues.remove(compositeKey, newStoredValue);
+				this.storedValues.remove(compositeKey, storedValue);
 				throw t;
 			}
 		}
-		return result;
+		return storedValue.evaluate();
+	}
+
+	private <K, V> @Nullable Object computeAndInstall(CompositeKey<N> compositeKey, @Nullable StoredValue expectedOld,
+			K key, Function<? super K, ? extends V> defaultCreator) {
+
+		var deferred = new DeferredSupplier(() -> {
+			rejectIfClosed();
+			return Preconditions.notNull(defaultCreator.apply(key), "defaultCreator must not return null");
+		});
+		StoredValue newStoredValue = newStoredValue(deferred);
+
+		boolean installed = (expectedOld == null ? this.storedValues.putIfAbsent(compositeKey, newStoredValue) == null
+				: this.storedValues.replace(compositeKey, expectedOld, newStoredValue));
+
+		if (!installed) {
+			return null;
+		}
+
+		deferred.run();
+		try {
+			return requireNonNull(deferred.getOrThrow());
+		}
+		catch (Throwable t) {
+			this.storedValues.remove(compositeKey, newStoredValue);
+			throw t;
+		}
 	}
 
 	/**
@@ -445,15 +497,8 @@ public final class NamespacedHierarchicalStore<N> implements AutoCloseable {
 			return this.supplier.get();
 		}
 
-		private @Nullable Object evaluateOrNullOnFailure() {
-			if (this.supplier instanceof MemoizingSupplier memoizing) {
-				return memoizing.getOrNullOnFailure();
-			}
-			return this.supplier.get();
-		}
-
 		static @Nullable Object evaluateIfNotNull(@Nullable StoredValue value) {
-			return value != null ? value.evaluateOrNullOnFailure() : null;
+			return value != null ? value.evaluate() : null;
 		}
 
 	}
@@ -472,6 +517,66 @@ public final class NamespacedHierarchicalStore<N> implements AutoCloseable {
 	}
 
 	/**
+	 * Deferred computation that can be installed into the store without executing
+	 * user code while holding internal map locks.
+	 *
+	 * <p>For {@link #get(Object, Object)}, failures are treated as logically absent
+	 * (returning {@code null}) so exceptions are not observable via {@code get()}.
+	 *
+	 * <p>For {@link #computeIfAbsent(Object, Object, Function)},
+	 * {@link #getOrThrow()} rethrows the original failure.
+	 */
+	private static final class DeferredSupplier implements Supplier<@Nullable Object> {
+
+		private final FutureTask<@Nullable Object> task;
+
+		private DeferredSupplier(Supplier<@Nullable Object> delegate) {
+			this.task = new FutureTask<>(delegate::get);
+		}
+
+		private void run() {
+			this.task.run();
+		}
+
+		@Override
+		public @Nullable Object get() {
+			try {
+				return this.task.get();
+			}
+			catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+				throw throwAsUncheckedException(e);
+			}
+			catch (ExecutionException e) {
+				Throwable t = e.getCause();
+				if (t == null) {
+					t = e;
+				}
+				UnrecoverableExceptions.rethrowIfUnrecoverable(t);
+				return null;
+			}
+		}
+
+		private @Nullable Object getOrThrow() {
+			try {
+				return this.task.get();
+			}
+			catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+				throw throwAsUncheckedException(e);
+			}
+			catch (ExecutionException e) {
+				Throwable t = e.getCause();
+				if (t == null) {
+					t = e;
+				}
+				UnrecoverableExceptions.rethrowIfUnrecoverable(t);
+				throw throwAsUncheckedException(t);
+			}
+		}
+	}
+
+	/**
 	 * Thread-safe {@link Supplier} that memoizes the result of calling its
 	 * delegate and ensures it is called at most once.
 	 *
@@ -485,18 +590,12 @@ public final class NamespacedHierarchicalStore<N> implements AutoCloseable {
 		private static final Object NO_VALUE_SET = new Object();
 
 		private final Supplier<@Nullable Object> delegate;
-		private final boolean transientFailures;
 
 		@Nullable
 		private volatile Object value = NO_VALUE_SET;
 
 		private MemoizingSupplier(Supplier<@Nullable Object> delegate) {
-			this(delegate, false);
-		}
-
-		private MemoizingSupplier(Supplier<@Nullable Object> delegate, boolean transientFailures) {
 			this.delegate = delegate;
-			this.transientFailures = transientFailures;
 		}
 
 		@Override
@@ -505,19 +604,6 @@ public final class NamespacedHierarchicalStore<N> implements AutoCloseable {
 				computeValue();
 			}
 			if (this.value instanceof Failure failure) {
-				throw throwAsUncheckedException(failure.throwable);
-			}
-			return this.value;
-		}
-
-		private @Nullable Object getOrNullOnFailure() {
-			if (this.value == NO_VALUE_SET) {
-				computeValue();
-			}
-			if (this.value instanceof Failure failure) {
-				if (this.transientFailures) {
-					return null;
-				}
 				throw throwAsUncheckedException(failure.throwable);
 			}
 			return this.value;


junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -237,26 +237,31 @@ public void close() {
...
-				if (StoredValue.evaluateIfNotNull(oldStoredValue) == null) {
-					rejectIfClosed();
-					var computedValue = Preconditions.notNull(defaultCreator.apply(key),
-						"defaultCreator must not return null");
-					return newStoredValue(() -> {
-						rejectIfClosed();
-						return computedValue;
-					});
+				if (oldStoredValue == null || oldStoredValue == storedValue) {
+					return valueFactory.get();
```
Pankraz76:
```suggestion
					return () -> newStoredValue(new MemoizingSupplier(() -> {
						rejectIfClosed();
						return notNull(defaultCreator.apply(key), "defaultCreator must not return null");
					})).get();
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -237,26 +237,31 @@ public void close() {
...
 	 */
+	@SuppressWarnings("ReferenceEquality")
 	@API(status = MAINTAINED, since = "6.0")
 	public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? extends V> defaultCreator) {
 		Preconditions.notNull(defaultCreator, "defaultCreator must not be null");
 		CompositeKey<N> compositeKey = new CompositeKey<>(namespace, key);
 		StoredValue storedValue = getStoredValue(compositeKey);
 		var result = StoredValue.evaluateIfNotNull(storedValue);
 		if (result == null) {
+			Supplier<StoredValue> valueFactory = () -> newStoredValue(new MemoizingSupplier(() -> {
```
Pankraz76: is there a need to share this currently unused reference, or is it technically obsolete, if inlineable?

martinfrancois: No it isn't :)

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
martinfrancois: Thanks for the suggestion, well spotted! I just pushed the change.

martinfrancois:
You're welcome @Pankraz76, thanks as well for the praise and the review, I really appreciate it! :)

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
Pankraz76, line 342:
```suggestion
return castToRequiredType(key, getOrComputeIfAbsent(namespace, key, defaultCreator), requiredType);
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -237,6 +237,7 @@ public void close() {
 	 * closed
 	 * @since 6.0
 	 */
+	@SuppressWarnings("ReferenceEquality")
 	@API(status = MAINTAINED, since = "6.0")
 	public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? extends V> defaultCreator) {
 		Preconditions.notNull(defaultCreator, "defaultCreator must not be null");
```
Pankraz76, line 244:
```suggestion
		notNull(defaultCreator, MANDATORY_DEFAULT_CREATOR);
```

imho, useless coupling, context and noisy burden.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -245,18 +246,21 @@ public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? e
...
-					rejectIfClosed();
-					var computedValue = Preconditions.notNull(defaultCreator.apply(key),
-						"defaultCreator must not return null");
-					return newStoredValue(() -> {
+				if (oldStoredValue == null || oldStoredValue == storedValue) {
+					return newStoredValue(new MemoizingSupplier(() -> {
 						rejectIfClosed();
-						return computedValue;
-					});
+						return Preconditions.notNull(defaultCreator.apply(key), "defaultCreator must not return null");
```
Pankraz76:
```suggestion
						return notNull(defaultCreator.apply(key), MANDATORY_DEFAULT_CREATOR_VALUE);
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
Pankraz76, line 56:
```suggestion
	private static final String MANDATORY_DEFAULT_CREATOR_VALUE = "defaultCreator must not return `null`";
	private static final String MANDATORY_DEFAULT_CREATOR = "defaultCreator must not be `null`";

		private final AtomicInteger insertOrderSequence = new AtomicInteger();
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -245,18 +246,21 @@ public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? e
...
+					}));
 				}
 				return oldStoredValue;
 			});
-			return requireNonNull(newStoredValue.evaluate());
+			try {
+				return requireNonNull(newStoredValue.evaluate());
+			}
+			catch (Throwable t) {
+				this.storedValues.remove(compositeKey, newStoredValue);
```
Pankraz76:
```suggestion
			storedValues.remove(compositeKey, newStoredValue);
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -245,18 +246,21 @@ public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? e
 		var result = StoredValue.evaluateIfNotNull(storedValue);
 		if (result == null) {
 			StoredValue newStoredValue = this.storedValues.compute(compositeKey, (__, oldStoredValue) -> {
```
Pankraz76:
```suggestion
			StoredValue newStoredValue = storedValues.compute(compositeKey, (__, oldStoredValue) -> {
```
the best we can do is to do nothing at all, or do whats close to that meaning delete.

everything we can delete without breaking seems obsolet, mostly imposing unwanted coupling burden.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -245,18 +246,21 @@ public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? e
...
-				if (StoredValue.evaluateIfNotNull(oldStoredValue) == null) {
-					rejectIfClosed();
-					var computedValue = Preconditions.notNull(defaultCreator.apply(key),
-						"defaultCreator must not return null");
-					return newStoredValue(() -> {
+				if (oldStoredValue == null || oldStoredValue == storedValue) {
```
Pankraz76: Condition 'oldStoredValue == null' covered by subsequent condition 'oldStoredValue == storedValue'

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -245,18 +246,21 @@ public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? e
...
+					return newStoredValue(new MemoizingSupplier(() -> {
 						rejectIfClosed();
-						return computedValue;
-					});
+						return Preconditions.notNull(defaultCreator.apply(key), "defaultCreator must not return null");
+					}));
 				}
 				return oldStoredValue;
 			});
```
Pankraz76:
```suggestion
		public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? extends V> defaultCreator) {
		notNull(defaultCreator, CREATOR_MUST_NOT_BE_NULL);
		var compositeKey = new CompositeKey<>(namespace, key);
		return getOrComputeValue(key, defaultCreator, compositeKey, getStoredValue(compositeKey));
	}

	private <K, V> Object getOrComputeValue(K key, Function<? super K, ? extends V> defaultCreator,
			CompositeKey<N> compositeKey, StoredValue value) {
		if (StoredValue.evaluateIfNotNull(value) == null) {
			return computeAndStoreValue(key, defaultCreator, compositeKey, value);
		}
		return value;
	}

	private <K, V> Object computeAndStoreValue(K key, Function<? super K, ? extends V> defaultCreator,
			CompositeKey<N> compositeKey, StoredValue value) {
		var newValue = computeIfUnchanged(key, defaultCreator, compositeKey, value);
		try {
			return requireNonNull(requireNonNull(newValue).evaluate());
		}
		catch (Throwable t) {
			storedValues.remove(compositeKey, newValue);
			throw t;
		}
	}

	private <K, V> StoredValue computeIfUnchanged(K key, Function<? super K, ? extends V> defaultCreator,
			CompositeKey<N> compositeKey, StoredValue value) {
		return requireNonNull(storedValues.compute(compositeKey, (__, oldValue) -> {
			if (oldValue == value) {
				return newValue(new MemoizingSupplier(() -> {
					rejectIfClosed();
					return notNull(defaultCreator.apply(key), CREATOR_MUST_NOT_RETURN_NULL);
				}));
			}
			return oldValue;
		}));
	}
```

Pankraz76 (commented):
2cts.

going fully functional separating the concerns (SoC/SRP).

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
martinfrancois: @Pankraz76 it doesn't unfortunately, I thought so at first too (as IntelliJ suggested it), but that's why I added the test `computeIfAbsentOverridesParentNullValue` which fails if you try to remove `oldStoredValue == null` here

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
martinfrancois: I wanted to do this (`Preconditions.notNull` => `notNull`) too, but since it was like this in the code already I wasn't sure if it was a convention here, as I imagine it could potentially be confused with `requireNonNull` or it could be intentional to make it clear it's a precondition. Could a maintainer please give a second opinion here? I'd be glad to change it.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
martinfrancois: This is a change in `getOrComputeIfAbsent` which is outside of the scope of this PR, but feel free to submit it as a separate PR as it's completely independent of my change.

martinfrancois: As there aren't that many constants in the code for repeating strings like this I'm not sure this is on purpose - I'd agree with that change, can a maintainer please give a second opinion here?

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
martinfrancois: I think that's something I'd move to a separate PR as you suggested, as I feel like it might be too big of a change for this PR.

martinfrancois:
You’re welcome, and thanks again, @Pankraz76, for taking another careful look.
Your suggestions are great, but as you pointed out, they mostly target existing code that could be cleaned up independently. I’d like to keep this PR focused on the bugfix so we do not drag out the review with additional refactoring.
Once this is merged, I am happy to follow up with a separate cleanup PR. I do not want to delay AssertJ being able to update to JUnit 6 again any longer than necessary 🙂

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
mpkorstanje: No, we don't extract constants for pre-condition messages. It creates a layer of indirection that makes the code harder to understand. And I can't give you chapter and verse, but Java will de duplicate constant expressions.

Also note that we don't put backticks around `null`.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
mpkorstanje: We generally use `Preconditions.notNull`.

documentation/modules/ROOT/partials/release-notes/release-notes-6.0.2.adoc
```diff
@@ -17,6 +17,13 @@ repository on GitHub.
 ==== Bug Fixes
 
 * Make `ConsoleLauncher` compatible with JDK 26 by avoiding final field mutations.
+* Fix a concurrency bug in `NamespacedHierarchicalStore.computeIfAbsent` where
+  the `defaultCreator` function was executed while holding the store's internal
+  map lock. Under parallel execution, this could cause threads using the store to
+  block each other and temporarily see a missing or incorrectly initialized state
+  for values created via `computeIfAbsent`. The method now evaluates
+  `defaultCreator` outside the critical section using a memoizing supplier,
+  aligning its behavior with the deprecated `getOrComputeIfAbsent`.
```
mpkorstanje: Could you write this down more concisely? The release notes generally focus on a top-line understanding of what was fixed. You could express this as having solved the symptoms of #5209 rather than its root cause.

You could for clarity also add a second item that describes how `computeIfAbsent` no longer deadlocks.

platform-tests/src/test/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStoreTests.java
```diff
@@ -416,6 +419,115 @@ void simulateRaceConditionInComputeIfAbsent() throws Exception {
...
+				first.join(1000);
+				second.join(1000);
+
+				assertThat(firstThreadTimedOut).as(
+					"getOrComputeIfAbsent should not block subsequent computations on colliding keys").isFalse();
+			}
+		}
+
+		@Test
+		void computeIfAbsentCanDeadlockWithCollidingKeys() throws Exception {
```
mpkorstanje: The naming of this test suggests that `computeIfAbsent` can currently deadlock. But I assume that after your fix this is no longer the case?

mpkorstanje (commented):
I've given this a quick read through and left some comments to resolve open questions, but this is not a full review yet.

It seems that you've found and solved a different problem than described in https://github.com/junit-team/junit-framework/issues/5171. And you claim the solutions overlap.

Unfortunately the original problem is quite tricky and the description for this pull request incredibly verbose. We'll need some time to go through the details. You can help us process this by writing a much more concise PR description.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -237,26 +237,30 @@ public void close() {
...
-			return requireNonNull(newStoredValue.evaluate());
+			try {
+				return requireNonNull(newStoredValue.evaluate());
+			}
+			catch (Throwable t) {
+				storedValues.remove(compositeKey, newStoredValue);
```
mpkorstanje: There is a period of time between `storedValues.compute()` and `storedValues.remove()` where a different thread via `getStoredValue()` can briefly access the `newStoredValue` and encounter its stored exception. As such the stores operations are not atomic. 

And I think this invalidates any approach that tries to avoid execution of the `defaultCreator` outside the `compute` method.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -237,26 +237,30 @@ public void close() {
...
+	@SuppressWarnings("ReferenceEquality")
 	@API(status = MAINTAINED, since = "6.0")
 	public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? extends V> defaultCreator) {
 		Preconditions.notNull(defaultCreator, "defaultCreator must not be null");
 		CompositeKey<N> compositeKey = new CompositeKey<>(namespace, key);
 		StoredValue storedValue = getStoredValue(compositeKey);
 		var result = StoredValue.evaluateIfNotNull(storedValue);
 		if (result == null) {
-			StoredValue newStoredValue = this.storedValues.compute(compositeKey, (__, oldStoredValue) -> {
-				if (StoredValue.evaluateIfNotNull(oldStoredValue) == null) {
```
mpkorstanje, line 248: In your analysis you said:

> In NamespacedHierarchicalStore#computeIfAbsent, the implementation previously relied on ConcurrentMap.computeIfAbsent, which provides the "one logical initialization per key" behavior. After the change to storedValues.compute(…), every call to NamespacedHierarchicalStore.computeIfAbsent for the same key can rerun the initialization logic and replace the existing StoredValue.
>
> That means that even though each compute call is atomic, two threads calling NamespacedHierarchicalStore.computeIfAbsent for the same key can:
>
> 1. Have Thread A initialize the stored value and start tracking statistics.
> 2. Then have Thread B rerun the initialization and replace that value, effectively resetting the statistics.

But looking at the existing implementation, the `defaultCreator` is not applied until after the `oldStoredValue` has been checked. So when `defaultCreator` is applied for a given key a value was either not set at all or that value was set and set not `null`. So on the face of it the `defaultCreator` should be applied at most once and point 2 shouldn't happen.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
mpkorstanje, line 222: Here should be the same problem as https://github.com/junit-team/junit-framework/pull/5209#discussion_r2611282289.  The result of a failing `defaultCreator` can be seen through `get`. This behaviour is guarded against for `computeIfAbsent` and verify with tests. 

Note to self: Write some more tests to cover this.

platform-tests/src/test/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStoreTests.java
martinfrancois: Agreed. After the fix, this scenario should no longer deadlock. I renamed it to `computeIfAbsentDoesNotDeadlockWithCollidingKeys`

martinfrancois:
> I think I've already found one reason this will not work as expected. See comment below.
> 
> > simulateRaceConditionInComputeIfAbsent did not catch this issue because it only exercises contention on a single key and relies on ConcurrentHashMap's per-key atomicity; it does not force different keys into the same bucket or run user code in a way that re-enters the store while the map lock is held, so the problematic interaction never occurs in that test
> 
> With this in mind, I would have expected to see a test like `simulateRaceConditionInComputeIfAbsent` that forces keys into the same bucket.

Added `simulateRaceConditionInComputeIfAbsentWithCollidingKeys` which:
- Uses 20 threads racing on two colliding keys
- Verifies each key's `defaultCreator` is called exactly once
- Verifies all threads see and share the same computed value

This test fails with the old implementation (creator called 10x instead of 1x) and passes with the fix.

I also added `computeIfAbsentWithCollidingKeysDoesNotBlockConcurrentAccess` which specifically tests that `computeIfAbsent` for one key doesn't block concurrent access to a different key in the same hash bucket.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
martinfrancois: > And I think this invalidates any approach that tries to avoid execution of the defaultCreator outside the compute method.

Both implementations now address this concern:

**Approach 1 (MemoizingSupplier):** Added `transientFailures` flag. When `true`, `get()` returns `null` instead of throwing during the transient window.

**Approach 2 (DeferredSupplier):** Uses type-based dispatch. `DeferredSupplier.get()` catches `ExecutionException` and returns `null`, while `getOrThrow()` rethrows for the original `computeIfAbsent` caller.

Both ensure:
- `defaultCreator` runs outside the bucket lock (avoiding blocking/deadlock)
- `get()` returns `null` during the transient window (not the exception)
- The original `computeIfAbsent` caller sees the exception and cleans up
- After cleanup, `get()` continues to return `null` (correct final state)

Tests added:
- `getDoesNotSeeTransientExceptionFromComputeIfAbsent`
- `getConcurrentWithFailingComputeIfAbsentDoesNotSeeException`

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
martinfrancois: You're right that the previous implementation checked `oldStoredValue` before applying `defaultCreator`. The issue wasn't about re-running initialization for the same key, but about **where** `defaultCreator` executes.

The previous implementation ran `defaultCreator.apply(key)` **inside** the `compute()` lambda while holding the bucket lock. Even though the check prevents double-initialization, any thread trying to access a different key in the same bucket is blocked until `defaultCreator` completes.

The fix moves `defaultCreator` execution **outside** the lock, so the map operation is fast and doesn't block other bucket operations.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
martinfrancois: Addressed together with https://github.com/junit-team/junit-framework/pull/5209#discussion_r2617491830. Both implementations ensure that:
- The caller of `computeIfAbsent` sees the exception
- Other callers via `get()` see `null`

This preserves the expected semantics: after `computeIfAbsent` fails and removes the entry, `get()` returns `null`.

martinfrancois:
Thanks for the initial review @mpkorstanje!

I apologize for being a bit verbose on both the release notes and the description of the PR. I now shortened both of them, I hope I didn't shorten them too much - let me know if I should make any changes there.

I also addressed your comments by adding both tests to reproduce the issues you mentioned and also addressed them. I first implemented an approach but didn't like some aspects, so I tried another approach. Both approaches ended up working, and since I'm not sure which one you would prefer I decided to push both as separate commits, so I could later drop the one you don't like.

1. **Commit 1** ([`267e938`](https://github.com/junit-team/junit-framework/pull/5209/changes/267e93854dfea00cc6776693c98ebb7b94b050b1)): Uses `MemoizingSupplier` with a `transientFailures` flag
   - Adds a boolean flag to distinguish between callers that should see exceptions vs. those that shouldn't
   - Uses `storedValues.compute()` with the supplier installed inside the lambda

2. **Commit 2** ([`32faf28`](https://github.com/junit-team/junit-framework/pull/5209/changes/32faf281758f1ebfcc75a360250e1a74445ad8e4)): Uses `DeferredSupplier` with `FutureTask`
   - Uses `DeferredSupplier` for `computeIfAbsent`, `MemoizingSupplier` for `getOrComputeIfAbsent`
   - Uses explicit CAS-retry loop (`for (;;)` with `putIfAbsent`/`replace`) instead of `compute()`
   - Leverages `FutureTask` which already provides "run once" semantics

Both approaches:
- Fix the bucket-lock contention issue (threads no longer block each other on colliding keys)
- Fix the deadlock scenario (first computation waiting for second no longer deadlocks)
- Ensure `get()` doesn't see transient exceptions from failed `computeIfAbsent`
- Pass all existing and new tests

**Trade-offs:**

| Aspect | Approach 1 (MemoizingSupplier) | Approach 2 (DeferredSupplier) |
|--------|-------------------------------|-------------------------------|
| Complexity | Less code, adds flag to existing class | More code, new class |
| Dispatch | Boolean flag (more implicit) | Type-based (cleaner, but more verbose) |
| Pattern | Uses `compute()` | Explicit CAS-retry loop (matches ConcurrentHashMap internals) |
| Dependencies | None | Uses `FutureTask` |

Please let me know which approach you prefer, and I'll drop the other commit.

mpkorstanje (commented):
Overall I think you're going in the right direction.

But I don't think the CAS loop is a good solution. It means that we're not taking advantage of the existing data structure.

Rather, I think it would be important to recognize and make explicit that there are 4 type of stored values. Absent (if nothing was stored yet), Constant (from put), Memoized (from getOrComputeIfAbsent) and Defered (from computeIfAbsent). These are containers that may have the following contents:

| Type         | Value    | Null | Exception |
|--------|--------|--------|--------|
| Absent | No |  No | No |
| Constant | Maybe |  Maybe | No |
| Memoized | Maybe | Maybe | Maybe |
| Defered | Maybe | No | Maybe |


These can be directly stored in the map and evaluated immediately afterwards. Then depending on retrieval function (get, getOrComputeIfAbsent, computeIfAbsent) different things happen.  This means that if the evaluation of a value failed, the value need need not be removed from the map. 


| Method | Type | Containing | Action |
| ----------- | ----------- | ----------- | ----------- |
| get |  Absent | - | return null |
| get |  Constant | value | return value |
| get |  Constant | null | return null |
| get |  Memoized | value | return value |
| get |  Memoized | null | return null |
| get |  Memoized | exception | trow exception |
| get |  Defered | value | return value |
| get |  Defered | exception | return null |
| getOrComputeIfAbsent |  Absent | - | compute new value |
| getOrComputeIfAbsent |  Constant | value | return value |
| getOrComputeIfAbsent |  Constant | null | return null |
| getOrComputeIfAbsent |  Memoized | value | return value |
| getOrComputeIfAbsent |  Memoized | null | return null |
| getOrComputeIfAbsent |  Memoized | exception | trow exception |
| getOrComputeIfAbsent |  Defered | value | return value |
| getOrComputeIfAbsent |  Defered | exception | return null |
| computeIfAbsent |  Absent | - | compute new value |
| computeIfAbsent |  Constant | value | return value |
| computeIfAbsent |  Constant | null | compute new value |
| computeIfAbsent |  Memoized | value | return value |
| computeIfAbsent |  Memoized | null | compute new  value |
| computeIfAbsent |  Memoized | exception | compute new  value |
| computeIfAbsent |  Defered | value | return value |
| computeIfAbsent |  Defered | exception | compute new  value |


You have addressed some this implicitly already but I think it would be good to make this explicit.

For example one of the cases that was overlooked in the current situation happens when a call to `getOrComputeIfAbsent` puts a value in the store but `computeIfAbsent` wins the race to evaluates it. This would allow an exception to be thrown from `computeIfAbsent` that did not originate from its creator function. I think the solution would be to have both `computeIfAbsent` and `getOrComputeIfAbsent` use the `DeferredSupplier`.

Finally, I think the current solution is also very dependent on testing race conditions. This make for rather verbose and potentially flakey tests. I think we can avoid most of that by letting the concurrent hash-map do the heavy lifting and then figure out what to do with it state afterwards. 

Anyway, I'm going to have a look tomorrow and see if I can get all of those concepts into a pull request.

martinfrancois:
Totally agreed @mpkorstanje, let's do that. If that's easier for you - feel free to make pushes to my branch directly, you have write access to it.

mpkorstanje:
Cheers! It was a bit easier to start writing tests on a new branch and copy the relevant bits over. I've given you co-author credits in the git commits!

Superseded by #5231.

PR #5213, opened 2025-12-12
[tryout fix] concurrency bug in `NamespacedHierarchicalStore#computeIfAbsent(Object, Object, Function)` #5171 #5209

Pankraz76:
### [tryout fix] concurrency bug in `NamespacedHierarchicalStore#computeIfAbsent(Object, Object, Function)` #5171 #5209

---

---

### Definition of Done

coverage:

before:

<html>
<body>

100% (7/7) | 97% (41/42) | 97% (132/136) | 87% (35/40)
-- | -- | -- | --

</body>
</html>�

after:

<html>
<body>

100% (7/7) | 97% (40/41) | 97% (131/134) | 90% (40/44)
-- | -- | -- | --

</body>
</html>

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
@@ -133,7 +133,7 @@ public final class NamespacedHierarchicalStore<N> implements AutoCloseable {
 			try {
 				if (this.closeAction != null) {
 					List<Throwable> failures = new ArrayList<>();
-					this.storedValues.entrySet().stream() //
+					storedValues.entrySet().stream() //
 							.map(e -> e.getValue().evaluateSafely(e.getKey())) //
 							.filter(it -> it != null && it.value != null) //
 							.sorted(EvaluatedValue.REVERSE_INSERT_ORDER) //
@@ -213,11 +213,10 @@ public final class NamespacedHierarchicalStore<N> implements AutoCloseable {
 		CompositeKey<N> compositeKey = new CompositeKey<>(namespace, key);
 		StoredValue storedValue = getStoredValue(compositeKey);
 		if (storedValue == null) {
-			storedValue = this.storedValues.computeIfAbsent(compositeKey,
-				__ -> newStoredValue(new MemoizingSupplier(() -> {
-					rejectIfClosed();
-					return defaultCreator.apply(key);
-				})));
+			storedValue = storedValues.computeIfAbsent(compositeKey, __ -> newValue(new MemoizingSupplier(() -> {
+				rejectIfClosed();
+				return defaultCreator.apply(key);
+			})));
 		}
 		return storedValue.evaluate();
 	}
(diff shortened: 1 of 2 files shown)


junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -237,26 +237,34 @@ public void close() {
...
-		StoredValue storedValue = getStoredValue(compositeKey);
+		requireNonNull(defaultCreator);
+		var compositeKey = new CompositeKey<>(namespace, key);
+		var storedValue = getStoredValue(compositeKey);
 		var result = StoredValue.evaluateIfNotNull(storedValue);
 		if (result == null) {
-			StoredValue newStoredValue = this.storedValues.compute(compositeKey, (__, oldStoredValue) -> {
-				if (StoredValue.evaluateIfNotNull(oldStoredValue) == null) {
+			var newStoredValue = this.storedValues.compute(compositeKey, (__, oldStoredValue) -> {
+				if (oldStoredValue == storedValue || oldStoredValue == null) {
```
Pankraz76: flipping this term, seems to solve the NPE issue, in `L:253`.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -237,26 +237,34 @@ public void close() {
 	 * closed
 	 * @since 6.0
 	 */
+	@SuppressWarnings("ReferenceEquality")
 	@API(status = MAINTAINED, since = "6.0")
 	public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? extends V> defaultCreator) {
-		Preconditions.notNull(defaultCreator, "defaultCreator must not be null");
```
Pankraz76, line 242: The precondition itself enforces (for good reason, I admit) a specific string pattern, which is beneficial to avoid being overly vague.

Still, the convention principle seems to always win when using the good old POJO approach.

Considering this kind of implementation detail, and in the sake of not making things too DRY, there is some of the same reasoning as correctly mentioned by @mpkorstanje about not introducing too much coupling and extra, extra.

Of course, this is kind of debatable, non-priority, and off-topic—just wanted to give some reasoning.

In: https://github.com/junit-team/junit-framework/pull/5209#discussion_r2610767321

platform-tests/src/test/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStoreTests.java
```diff
@@ -672,4 +678,569 @@ public String toString() {
...
+
+	/**
+	 * <a href="https://github.com/junit-team/junit-framework/issues/5171">#5171</a>
+	 * <a href="https://github.com/junit-team/junit-framework/pull/5209">#5209</a>
+	 */
+	@Nested
+	class ConcurrencyIssue5171 {
+
+		@Test
+		void computeIfAbsentDoesNotDeadlockWithCollidingKeys() throws Exception {
```
Pankraz76, line 714: actually not sure, if this is any good... this topic too much for my 2ct., cannot afford.

Just curious as it was failing the current impl., trying to fix it right away lead to this PR. 

Maybe you can check it out. Thanks.

platform-tests/src/test/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStoreTests.java
```diff
@@ -672,4 +678,569 @@ public String toString() {
...
+				throw new RuntimeException("First attempt fails");
+			}));
+
+			// Second call should succeed (failed entry was removed)
+			assertDoesNotThrow(() -> {
+				var result = store.computeIfAbsent("ns", "key", __ -> "success");
+				assertEquals("success", result);
+			});
+
+			assertEquals(1, exceptionCount.get()); // First failed, second succeeded why 1?
```
Pankraz76: this a smell. 🧀

Pankraz76:
kindly asking, if this is any good?

Thanks to the community.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
mpkorstanje: In isolation the reasoning makes sense. But `notNull` is but one precondition of many. There are also `notBlank`, `containsNoNullElements`, ect. So we use `Precondition` for consistency.

mpkorstanje (commented):
Thanks for offering!

> kindly asking, if this is any good?

It's not bad. If I was working on this bit of code I would pick up the `var` additions, `this.` removal, and removal of local variables. The tests look like they could be useful too. 

But there are other things to consider as well. 

Changing code just to make it look prettier provides only a small benefit but also comes with a cost. It tends to make the git blame harder to read, makes PRs bigger than they need to be, and overall takes up maintainer time in review. The latter is especially scarce.

Right now for #5171, we haven't quite figured out yet what we want the solution to be, nor do we have a reliable reproducer that we can put into a unit test. So making a PR is rather premature. But I do appreciate the gesture and I'll keep the ideas behind your changes in mind.

PS: If you can reliably reproduce the effects of #5171 as a unit test against the `NamespacedHierarchicalStore`, I think the most appropriate way to provide that would be as a comment on #5171.

PR #5258, opened 2026-01-02
Contribute SystemPropertyExtension

mpkorstanje:
Adopts the [`SystemPropertyExtension`](https://github.com/junit-pioneer/junit-pioneer/blob/062ce51296dab909119fd6c9f090b9d6a2982530/src/main/java/org/junitpioneer/jupiter/SystemPropertyExtension.java) from JUnit Pioneer into JUnit Jupiter. While this is a faithful port some notable changes have been made. With that in mind the API status has been set to experimental.

### `@RestoreSystemProperties` uses `Properties::clone`

When using `@RestoreSystemProperties` JUnit Pioneer would create a copy of the effective properties in the object. This would include any defaults as if they were regular values. I.e:

```java
	static Properties createEffectiveClone(Properties original) {
		Properties clone = new Properties();

		// This implementation is used because:
		// System.getProperties() returns the actual Properties object, not a copy.
		// Clone doesn't include nested defaults, but propertyNames() does.
		original.propertyNames().asIterator().forEachRemaining(k -> {
			String v = original.getProperty(k.toString());

			if (v != null) {
				// v will be null if the actual value was an object
				clone.put(k, original.getProperty(k.toString()));
			}
		});

		return clone;
	}
```

 JUnit instead uses `Properties::clone` and performs a best effort attempt to detect default properties and fail if any were detected. For classes that extend `Properties` it is assumed that `clone` is implemented with sufficient fidelity. I.e:
 
```java
	static Properties cloneWithoutDefaults(ExtensionContext context, Properties properties) {
		// Custom implementations have to implement clone correctly.
		if (properties.getClass() == Properties.class) {
			throwIfHasObservableDefaults(context, properties);
		}
		return (Properties) properties.clone();
	}
```

We do this because:

1. System properties are to the best of our knowledge created without defaults. Nor are we aware of a good usecase for defaults. If you do have a use case for this, please do create  a new issue to explain it.
2. Using `Properties::clone` provides a contract that is easier to document and understand. It also allows subclasses of `Properties` to be provided to tests. For example, a variant that does include defaults when cloned.

### Restore "compromised"  entries

While strongly discouraged non-string values can be added to a properties object.  The Pioneer implementation would not restore these "compromised" values. This has been fixed in the JUnit implementation.

Closes: https://github.com/junit-team/junit-framework/issues/4726

---

---

### Definition of Done

### DIFF
--- a/documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
+++ b/documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
@@ -296,15 +296,21 @@ Tests annotated in this way will never execute in parallel with tests annotated 
 [[SystemProperty]]
 == The System Property Extensions
 
-The System Property extensions consists is a set of annotations that work together to clear, set and restore system properties.
+The System Property extensions consists is a set of annotations that work together to
+clear, set and restore system properties.
 
 [[SystemProperty-clear-system-properties-and-set-system-properties]]
 === @ClearSystemProperty and @SetSystemProperty
 
-The `{ClearSystemProperty}` and `{SetSystemProperty}` extensions can be used to clear and set, respectively, the values of system properties for a test execution.
-Both annotations work on the test method and class level, are repeatable, combinable, and inherited from higher-level containers.
-After the annotated method has been executed, the properties mentioned in the annotation will be restored to their original value or the value of the higher-level container, or will be cleared if they didn't have one before.
-Other system properties that are changed during the test, are *not* restored (unless the `@RestoreSystemProperties` is used).
+The `{ClearSystemProperty}` and `{SetSystemProperty}` extensions can be used to clear and
+set, respectively, the values of system properties for a test execution. Both annotations
+work on the test method and class level, are repeatable, combinable, and inherited from
+higher-level containers. After the annotated method has been executed, the properties
+mentioned in the annotation will be restored to their original value or the value of the
+higher-level container, or will be cleared if they didn't have one before. Other system
+properties that are changed during the test, are _not_ restored (unless
+<<SystemProperty-restore-system-properties, explicitly enabled>> via
+`{RestoreSystemProperties}`).
 
 For example, clearing a system property for a test execution can be done as follows:
 
@@ -337,40 +343,59 @@ include::example$java/example/SystemPropertyExtensionDemo.java[tag=systempropert
 
 [NOTE]
 ====
+Method-level configurations are visible in both `@BeforeEach` setup methods and
+`@AfterEach` teardown methods (see
+xref:extensions/relative-execution-order-of-user-code-and-extensions.adoc#overview[user
+code and extension code execution order]).
 
-Method-level configurations are visible in both `@BeforeEach` setup methods and `@AfterEach` teardown methods (see xref:extensions/relative-execution-order-of-user-code-and-extensions.adoc#overview[user code and extension code execution order]).
-
-A class-level configuration means that the specified system properties are cleared/set before and reset after each individual test in the annotated class.
+A class-level configuration means that the specified system properties are cleared/set
+before and reset after each individual test in the annotated class.
 ====
 
 [[SystemProperty-restore-system-properties]]
 === @RestoreSystemProperties
 
-The `{RestoreSystemProperties}` Extension can be used to restore changes to system properties made directly in code.
-While `@ClearSystemProperty` and `@SetSystemProperty` set or clear specific properties and values, they don't allow property values to be calculated or parameterized, thus there are times you may want to directly set properties in your test code.
-`@RestoreSystemProperties` can be placed on test methods or test classes and will completely restore all system properties to their original state after a test or test class is complete.
+The `{RestoreSystemProperties}` extension can be used to restore changes to system
+properties made directly in code. While `@ClearSystemProperty` and `@SetSystemProperty`
+set or clear specific properties and values, they don't allow property values to be
+calculated or parameterized, thus there are times you may want to directly set properties
+in your test code. `@RestoreSystemProperties` can be placed on test methods or test
+classes and will completely restore all system properties to their original state after a
+test or test class is complete.
 
 [NOTE]
 ====
-During the execution of the annotated scope, the system properties are set to a clone of the original `Properties` object.
-The clone does not include {jdk-javadoc-base-url}/java.base/java/util/Properties.html#defaults[the defaults] from the original.
-The extension will perform a best effor attempt to detect default properties and fail if any were detected.
-For classes that extend `Properties` it is assumed that `clone()` is implemented with sufficient fidelity.
+During the execution of the annotated scope, the system properties are set to a clone of
+the original `Properties` object. The clone does not include
+{jdk-javadoc-base-url}/java.base/java/util/Properties.html#defaults[the defaults] from the
+original. The extension will perform a best effort attempt to detect default properties
+and fail if any were detected. For classes that extend `Properties` it is assumed that
+`clone()` is implemented with sufficient fidelity.
 ====
 
-In this example, `@RestoreSystemProperties` is used on a test method, ensuring any changes made in that method are restored:
+In this example, `@RestoreSystemProperties` is used on a test method, ensuring any changes
+made in that method are restored:
 
 [source,java,indent=0]
 ----
 include::example$java/example/SystemPropertyExtensionDemo.java[tag=systemproperty_restore_test]
 ----
 
-When `@RestoreSystemProperties` is used on a test class, any system properties changes made during the entire lifecycle of the test class, including test methods, `@BeforeAll`, `@BeforeEach` and 'after' methods, are restored after the test class' lifecycle is complete.
-In addition, the annotation is inherited by each test method just as if each one was annotated with `@RestoreSystemProperties`.
+When `@RestoreSystemProperties` is used on a test class, any system properties changes
+made during the entire lifecycle of the test class, including test methods, `@BeforeAll`,
+`@BeforeEach` and 'after' methods, are restored after the  lifecycle of the test class is
+complete. In addition, the annotation is inherited by each test method just as if each one
+was annotated with `@RestoreSystemProperties`.
 
-In the following example, both test methods see the system property changes made in `@BeforeAll` and `@BeforeEach`, however, the test methods are isolated from each other (`isolatedTest2` does not 'see' changes made in `isolatedTest1`).
-As shown in the second example below, the class-level `@RestoreSystemProperties` ensures that system property changes made within the annotated class are completely restored after the class's lifecycle, ensuring that changes are not visible to `SomeOtherTestClass`.
-Note that `SomeOtherTestClass` uses the `@ReadsSystemProperty` annotation: This ensures that JUnit does not schedule the class to run during any test known to modify system properties (see <<SystemProperties-thread-safety>>).
+In the following example, both test methods see the system property changes made in
+`@BeforeAll` and `@BeforeEach`, however, the test methods are isolated from each other
+(`isolatedTest2` does not 'see' changes made in `isolatedTest1`). As shown in the second
+example below, the class-level `@RestoreSystemProperties` ensures that system property
+changes made within the annotated class are completely restored after the class's
+lifecycle, ensuring that changes are not visible to `SomeOtherTestClass`. Note that
+`SomeOtherTestClass` uses the `@ReadsSystemProperty` annotation: This ensures that JUnit
+does not schedule the class to run during any test known to modify system properties (see
+<<SystemProperties-thread-safety>>).
 
 [source,java,indent=0]
 ----
@@ -387,9 +412,10 @@ include::example$java/example/SystemPropertyExtensionDemo.java[tag=systempropert
 [[SystemProperty-using-together]]
 === Using @ClearSystemProperty, @SetSystemProperty, and @RestoreSystemProperties together
 
-All three annotations can be combined, which could be used when some system properties are parameterized (i.e. need to be set in code) and others are not.
-For instance, imagine testing an image generation utility that takes configuration from system properties.
-Basic configuration can be specified using `Set` and `Clear` and the image size parameterized:
+All three annotations can be combined, which could be used when some system properties are
+parameterized (i.e. need to be set in code) and others are not. For instance, imagine
+testing an image generation utility that takes configuration from system properties. Basic
+configuration can be specified using `Set` and `Clear` and the image size parameterized:
 
 [source,java,indent=0]
 ----
@@ -398,21 +424,31 @@ include::example$java/example/SystemPropertyExtensionDemo.java[tag=systempropert
 
 [NOTE]
 ====
-Using `@RestoreSystemProperties` is not necessary to restore system properties modified via `@ClearSystemProperty` or `@SetSystemProperty` - they each automatically restore the referenced properties.
-'Restore', is only needed if system properties are modified in some way _other than_ Clear and Set during a test.
+Using `@RestoreSystemProperties` is not necessary to restore system properties modified
+via `@ClearSystemProperty` or `@SetSystemProperty` - they each automatically restore the
+referenced properties. `Restore` is only needed if system properties are modified in some
+way _other than_ `Clear` and `Set` during a test.
 ====
 
 [[SystemProperties-thread-safety]]
 === Thread Safety
 
-Since system properties are global state, reading and writing them during xref:writing-tests/parallel-execution.adoc[parallel execution] can lead to unpredictable results and flaky tests.
-The system property extension is prepared for that and tests annotated with `@ClearSystemProperty`, `@SetSystemProperty`, or `@RestoreSystemProperties` will never execute in parallel (thanks to xref:writing-tests/parallel-execution.adoc#synchronization[resource locks]) to guarantee correct test results.
+Since system properties are global state, reading and writing them during
+xref:writing-tests/parallel-execution.adoc[parallel execution] can lead to unpredictable
+results and flaky tests. The system property extension is prepared for that and tests
+annotated with `@ClearSystemProperty`, `@SetSystemProperty`, or `@RestoreSystemProperties`
+will never execute in parallel (thanks to
+xref:writing-tests/parallel-execution.adoc#synchronization[resource locks]) to guarantee
+correct test results.
 
-However, this does not cover all possible cases.
-Tested code that reads or writes system properties _independently_ of the extension can still run in parallel to it and may thus behave erratically when, for example, it unexpectedly reads a property set by the extension in another thread.
-Tests that cover code that reads or writes system properties need to be annotated with the respective annotation:
+However, this does not cover all possible cases. Tested code that reads or writes system
+properties _independently_ of the extension can still run in parallel to it and may thus
+behave erratically when, for example, it unexpectedly reads a property set by the
+extension in another thread. Tests that cover code that reads or writes system properties
+need to be annotated with the respective annotation:
 
 * `{ReadsSystemProperty}`
 * `{WritesSystemProperty}` (though consider using `@RestoreSystemProperties` instead)
 
-Tests annotated in this way will never execute in parallel with tests annotated with `@ClearSystemProperty`, `@SetSystemProperty`, or `@RestoreSystemProperties`.
+Tests annotated in this way will never execute in parallel with tests annotated with
+`@ClearSystemProperty`, `@SetSystemProperty`, or `@RestoreSystemProperties`.
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtension.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtension.java
@@ -10,6 +10,7 @@
 
 package org.junit.jupiter.api.util;
 
+import static java.util.stream.Collectors.toSet;
 import static org.junit.platform.commons.support.AnnotationSupport.findRepeatableAnnotations;
 import static org.junit.platform.commons.support.AnnotationSupport.isAnnotated;
 import static org.junit.platform.commons.util.CollectionUtils.forEachInReverseOrder;
@@ -24,7 +25,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.jspecify.annotations.Nullable;
@@ -82,8 +82,8 @@ final class SystemPropertyExtension
 		var allContexts = findAllExtensionContexts(originalContext);
 
 		var restoreAnnotationContext = findFirstRestoreAnnotationContext(allContexts);
-		restoreAnnotationContext.ifPresent(annotatedElement -> {
-			var properties = this.prepareToEnterRestorableContext(annotatedElement);
+		restoreAnnotationContext.ifPresent(annotatedElementContext -> {
+			var properties = this.prepareToEnterRestorableContext(annotatedElementContext);
 			storeCompleteBackup(originalContext, properties);
 		});
 
@@ -94,8 +94,8 @@ final class SystemPropertyExtension
 
 	private Optional<ExtensionContext> findFirstRestoreAnnotationContext(List<ExtensionContext> contexts) {
 		return contexts.stream() //
-				.filter(annotatedElement -> isAnnotated(annotatedElement.getElement(),
-					RestoreSystemProperties.class)).findFirst();
+				.filter(context -> isAnnotated(context.getElement(), RestoreSystemProperties.class)) //
+				.findFirst();
 	}
 
 	private void clearAndSetEntries(ExtensionContext currentContext, ExtensionContext originalContext,
@@ -121,9 +121,10 @@ final class SystemPropertyExtension
 	}
 
 	private Set<String> findEntriesToClear(AnnotatedElement element) {
-		return findRepeatableAnnotations(element, ClearSystemProperty.class).stream().map(ClearSystemProperty::key) //
+		return findRepeatableAnnotations(element, ClearSystemProperty.class).stream() //
+				.map(ClearSystemProperty::key) //
 				// already distinct due to findRepeatableAnnotations
-				.collect(Collectors.toSet());
+				.collect(toSet());
 	}
 
 	private Map<String, String> findEntriesToSet(AnnotatedElement element) {
@@ -145,7 +146,9 @@ final class SystemPropertyExtension
 	private void preventClearAndSetSameEntries(AnnotatedElement element, Set<String> entriesToClear,
 			Set<String> entriesToSet) {
 		requireUniqueEntries(element, //
-			entriesToClear.stream().filter(entriesToSet::contains).collect(Collectors.toSet()));
+			entriesToClear.stream() //
+					.filter(entriesToSet::contains) //
+					.collect(toSet()));
 	}
 
 	private static void requireUniqueEntries(AnnotatedElement annotatedElement, Set<String> duplicatePropertNames) {
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
@@ -51,10 +51,11 @@ repository on GitHub.
 * https://www.junit-pioneer.org/[JUnit Pioneer]'s `DefaultLocaleExtension` and
   `DefaultTimeZoneExtension` are now part of the JUnit Jupiter. Find examples in the
   xref:writing-tests/built-in-extensions.adoc#DefaultLocaleAndTimeZone[User Guide].
-* https://www.junit-pioneer.org/[JUnit Pioneer]'s `SystemPropertyExtension` is now part of the JUnit Jupiter.
-  Find examples in the xref:writing-tests/built-in-extensions.adoc#SystemProperty[User Guide].
-  Note: There are some implementation difference between JUnit Pioneer and Jupiter. See details in https://github.com/junit-team/junit-framework/pull/5258[junit-framework/#5258 - Contribute SystemPropertyExtension].
-
+* https://www.junit-pioneer.org/[JUnit Pioneer]'s `SystemPropertyExtension` is now part of
+  JUnit Jupiter. Find examples in the
+  xref:writing-tests/built-in-extensions.adoc#SystemProperty[User Guide]. Note: There are
+  some implementation differences between JUnit Pioneer and Jupiter. For details, see the
+  https://github.com/junit-team/junit-framework/pull/5258[corresponding pull request].
 * Support the creation of `Arguments` from iterables. These additions make it
 easier to dynamically build arguments from collections when using
 `@ParameterizedTest`:


junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtension.java
```diff
@@ -0,0 +1,117 @@
...
+	 * @param original {@code Properties} to be cloned.
+	 * @return A new {@code Properties} instance containing the same effective entries as the original.
+	 */
+	static Properties createEffectiveClone(Properties original) {
+		Properties clone = new Properties();
+
+		// This implementation is used because:
+		// System.getProperties() returns the actual Properties object, not a copy.
+		// Clone doesn't include nested defaults, but propertyNames() does.
+		original.propertyNames().asIterator().forEachRemaining(k -> {
```
mpkorstanje: Note: This doesn't "faithfully" recreate the distinction between default and non-default values. This could be important.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtension.java
```diff
@@ -0,0 +1,301 @@
...
+		currentContext.getElement().ifPresent(element -> {
+			Set<String> entriesToClear;
+			Map<String, String> entriesToSet;
+
+			try {
+				entriesToClear = findEntriesToClear(element);
+				entriesToSet = findEntriesToSet(element);
+				preventClearAndSetSameEntries(entriesToClear, entriesToSet.keySet());
+			}
+			catch (IllegalStateException ex) {
```
mpkorstanje: Exception used for control flow, sort of:
 * `preventClearAndSetSameEntries`
 * `findEntriesToClear` through `distinctToSet`
 * `findEntriesToSet` through `distinctToSet`

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtension.java
```diff
@@ -0,0 +1,301 @@
...
+		 */
+		List<ExtensionContext> contexts = findAllContexts(originalContext);
+		Collections.reverse(contexts);
+		contexts.forEach(currentContext -> clearAndSetEntries(currentContext, originalContext,
+			!isRestoreAnnotationPresent(originalContext)));
+	}
+
+	private boolean isRestoreAnnotationPresent(ExtensionContext originalContext) {
+		// TODO: Can we do this with more grace?
+		return findAnnotation(originalContext.getElement(), RestoreSystemProperties.class).isPresent()
```
mpkorstanje: How to find the `RestoreSystemProperties` in either method, class, or enclosing class all at once?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtensionUtils.java
```diff
@@ -0,0 +1,62 @@
...
+	 * @return a list of all contexts, "outwards" in the {@link ExtensionContext#getParent() getParent}-order,
+	 *         beginning with the given context; never {@code null} or empty
+	 */
+	static List<ExtensionContext> findAllContexts(ExtensionContext context) {
+		List<ExtensionContext> allContexts = new ArrayList<>();
+		for (var c = context; c != null; c = c.getParent().orElse(null)) {
+			allContexts.add(c);
+		}
+		return allContexts;
+	}
```
jbduncan: This method is used not just in system-properties-related code but `AbstractEntryBasedExtension` too.

Furthermore, it is used in line 230 of `AbstractEntryBasedExtension` which just iterates over the returned contexts. By comparison, other places in the codebase that _use_ a type with a method called `getParent()` (not just `ExtensionContext`) tend to loop over the parents directly rather than gathering them into a list first.

Therefore, I suggest moving this method to another file or inlining it.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtension.java
```diff
@@ -0,0 +1,301 @@
...
+	 * A clone of the String values of the passed {@code Properties}, including defaults.
+	 *
+	 * <p>The clone will have the same effective values, but may not use the same nested
+	 * structure as the original. Object values, which are technically possible,
+	 * are not included in the clone.</p>
+	 *
+	 * @param original {@code Properties} to be cloned.
+	 * @return A new {@code Properties} instance containing the same effective entries as the original.
+	 */
+	static Properties createEffectiveClone(Properties original) {
```
mpkorstanje: Move to `SystemPropertyExtensionUtils` and unit test.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtensionUtils.java
```diff
@@ -0,0 +1,62 @@
...
+			return left;
+		});
+	}
+
+	private static <T> void addButThrowIfDuplicate(Set<T> set, T element) {
+		boolean newElement = set.add(element);
+		if (!newElement) {
+			throw new IllegalStateException("Duplicate element '" + element + "'.");
+		}
+	}
```
jbduncan: I was pleasantly surprised that this wasn't in `CollectionUtils` already.

Furthermore, many years ago I introduced `CollectionUtils::toUnmodifiableList` to make immutable lists easier to write when using streams (though it was rightfully [replaced](https://github.com/junit-team/junit-framework/commit/e0a5504ccd0204ed95c70e17022fb8cc190cc726#diff-b36782efd4b10b3e769e60f0987699d95878887b776fe0c2125f567b9c8466fc) with `.toList()` calls at some point).

So I suggest the following:
- Consider moving `distinctToSet()` to `CollectionUtils` for future reuse.
    - If you decide to do this, wrap the collector in `collectingAndThen(collector, Collections::unmodifiableSet)` before returning it.
- Double check if there is anywhere else in JUnit that throws exceptions on duplicates being added to a collection, and refactor things as appropriate.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtensionUtils.java
jbduncan: Other bits of code in JUnit that operate on a type with a `getParent()` method (not just `ExtensionContext`) tend to loop over the parents directly rather than collecting them in a list first.

Furthermore, this method is called somewhere in `SystemPropertyExtension` where the returned list is reversed with `Collections::reverse`. By comparison, we have `CollectionUtils::forEachInReverseOrder` which makes looping over collections in reverse order easier and reduces the need for mutation.

Therefore, consider inlining this method, or at least changing the aforementioned usage of this method to use `CollectionUtils::forEachInReverseOrder` instead.

jbduncan (commented):
Just a couple of nits for consideration, but otherwise this draft is looking fantastic so far!

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtensionUtils.java
jbduncan: Though I see you have some [thoughts](https://github.com/junit-team/junit-framework/pull/5258#discussion_r2658452613) on `distinctToSet()`'s usages, so take this feedback as you will. :wink:

mpkorstanje: Cheers! Yeah. There is a bit of implicit validation happening in `findEntriesToClear` and `findEntriesToSet`. I think that should be explicit.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtensionUtils.java
mpkorstanje: Cheers, I've adopted `forEachInReverseOrder` at the least. But I don't see how to visit the parents in reverse order while also in-lining the loop.

Though I also needed `findAllContexts` for `isRestoreAnnotationPresent` in 7e2eb7cb3ba135110adc549850f5b57206ec793e. So I think we're missing a capability here.

mpkorstanje: Note to self, candidates for a parent iterator:
 * TestInstancePreDestroyCallback::preDestroyTestInstances (list)

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,119 @@ locale or time zone need to be annotated with the respective annotation:
...
+====
+Using `@RestoreSystemProperties` is not necessary to restore system properties modified via `@ClearSystemProperty` or `@SetSystemProperty` - they each automatically restore the referenced properties.
+'Restore', is only needed if system properties are modified in some way _other than_ Clear and Set during a test.
+====
+
+=== `@RestoreSystemProperties` Limitations
+
+The system `Properties` object is normally just a hashmap of strings, however, it is technically possible to store non-string values and create {jdk-javadoc-base-url}/java.base/java/util/Properties.html#%3Cinit%3E(java.util.Properties)[nested `Properties` with inherited default values].
+`@RestoreSystemProperties` restores the original `Properties` object with all of its potential richness _after_ the annotated scope is complete.
+However, for use during the test _within_ the test scope it provides a cloned `Properties` object with these limitations:
```
mpkorstanje: For team discussion: Are these limitations acceptable?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtensionUtils.java
jbduncan: > Cheers, I've adopted `forEachInReverseOrder` at the least.

Great! 👍

> But I don't see how to visit the parents in reverse order while also in-lining the loop.

You're right, I realised this too when I was typing things but I didn't vocalise it, so sorry if I misled you.

> Though I also needed `findAllContexts` for `isRestoreAnnotationPresent` in [7e2eb7c](https://github.com/junit-team/junit-framework/commit/7e2eb7cb3ba135110adc549850f5b57206ec793e). So I think we're missing a capability here.

Good shout! I'm not sure what this capability would look like, though... 🤔

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtension.java
```diff
@@ -0,0 +1,277 @@
...
+				if (backup == null)
+					this.entriesToClear.add(entry);
+				else
+					this.entriesToSet.put(entry, backup);
+			});
+		}
+
+		void restoreBackup() {
+			entriesToClear.forEach(System::clearProperty);
+			entriesToSet.forEach(System::setProperty);
```
mpkorstanje: Potential bug(?): The backup restores using `setProperty` using a value obtained through `getProperty`. This may have been a default, which case the value should be removed instead.

mpkorstanje:
For discussion:

As originally written the `SystemPropertyExtension` can restore system properties to their original state after testing. It does this by replacing the `Properties` with an effective clone before a test starts, then restoring the original.

This effective clone is limited to the observable entries and defaults of the original properties object. This means that default values that are hidden by an entry are not included. In essence this test will fail:

```java
var defaults = new Properties();
defaults.setProperty("a", "a");

var properties = new Properties(defaults);
properties.put("a", "b");

var clone = createEffectiveClone(properties);

assertThat(clone.getProperty("a")).isEqualTo("b");
clone.remove("a");
assertThat(clone.getProperty("a")).isEqualTo("a"); // fails: actually null.
```		

This can be avoided by changing the cloning process such that the entries from the original properties object are removed to expose the default values, and then restoring the entries after the defaults have been copied. But this requires mutating the original object. Is that acceptable?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtension.java
```diff
@@ -0,0 +1,277 @@
...
+			if (entriesToClear.isEmpty() && entriesToSet.isEmpty())
+				return;
+
+			// Only backup original values if we didn't already do bulk storage of the original state
+			if (doIncrementalBackup) {
+				storeIncrementalBackup(originalContext, entriesToClear, entriesToSet.keySet());
+			}
+
+			entriesToClear.forEach(System::clearProperty);
+			entriesToSet.forEach(System::setProperty);
```
mpkorstanje: Potential bug(?): `System::clearProperty` invokes `Properties::remove` which removes both string and non-string entries. The backup only restores strings.

jbduncan:
> For discussion:
> 
> As originally written the `SystemPropertyExtension` can restore system properties to their original state after testing. It does this by replacing the `Properties` with an effective clone before a test starts, then restoring the original.
> 
> This effective clone is limited to the observable entries and defaults of the original properties object. This means that default values that are hidden by an entry are not included. In essence this test will fail:
> 
> ```java
> var defaults = new Properties();
> defaults.setProperty("a", "a");
> 
> var properties = new Properties(defaults);
> properties.put("a", "b");
> 
> var clone = createEffectiveClone(properties);
> 
> assertThat(clone.getProperty("a")).isEqualTo("b");
> clone.remove("a");
> assertThat(clone.getProperty("a")).isEqualTo("a"); // fails: actually null.
> ```
> 
> This can be avoided by changing the cloning process such that the entries from the original properties object are removed to expose the default values, and then restoring the entries after the defaults have been copied. But this requires mutating the original object. Is that acceptable?

I checked the docs for `Properties` (not a power user of it, myself) and the `store()`/`load()` methods looked like a tempting solution, but they specifically don't serialize/deserialize default values, annoyingly. 😓

I can't believe I'm saying this, but would `clone()` or Java serialization work? Though I'm not holding my breath, given that `Properties` extends `Hashtable`.

The way I see it, there is a small window for tests to flake if this extension mutates the system properties during cloning whilst other tests running in parallel are using them. So whereas this mutation would be acceptable to me because I'm acutely aware of the perils of global state and race conditions and I would avoid calling `System.(get|set)Properties` outside of this extension (or even better use IoC/dependency injection to fake the system properties), I wouldn't count on other engineers knowing this, so I suggest removing the mutation, even if it means things aren't perfect.

mpkorstanje:
> I can't believe I'm saying this, but would clone() or Java serialization work? 

No. Clone doesn't include the defaults. And serialization would duplicate any objects in a corrupted properties object. If they're serializable in the first place...

> The way I see it, there is a small window for tests to flake if this extension mutates the system properties during cloning whilst other tests running in parallel are using them. 

Good point.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SystemPropertyExtensionUtils.java
mpkorstanje: So it turns out that `findRepeatableAnnotations` already de-duplicates making `distinctToSet` redundant.

mpkorstanje:
Note: Manually set DCO check to pass. Originally https://github.com/junit-team/junit-framework/pull/5184 was contributed before the DCO bot was enabled.

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,118 @@ locale or time zone need to be annotated with the respective annotation:
 
 Tests annotated in this way will never execute in parallel with tests annotated with
 `@DefaultLocale` or `@DefaultTimeZone`.
+
+[[writing-tests-built-in-extensions-SystemProperty]]
+==== `@ClearSystemProperty` and `@SetSystemProperty`
+
+The `@ClearSystemProperty` and `@SetSystemProperty` annotations can be used to clear and set, respectively, the values of system properties for a test execution.
```
marcphilipp: Our usual convention is to link the first occurrence of a class in a "section" to its Javadoc page (see above for `{DefaultLocale}`).

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,118 @@ locale or time zone need to be annotated with the respective annotation:
 
 Tests annotated in this way will never execute in parallel with tests annotated with
 `@DefaultLocale` or `@DefaultTimeZone`.
+
+[[writing-tests-built-in-extensions-SystemProperty]]
+==== `@ClearSystemProperty` and `@SetSystemProperty`
```
marcphilipp: Code snippets in headers look weird in Antora's theme so we should avoid them.

```suggestion
== @ClearSystemProperty and @SetSystemProperty
```

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,118 @@ locale or time zone need to be annotated with the respective annotation:
...
+----
+
+[NOTE]
+====
+Method-level configurations are visible in both `@BeforeEach` setup methods and `@AfterEach` teardown methods (see <<extensions-execution-order-overview, user code and extension code execution order>>).
+
+A class-level configuration means that the specified system properties are cleared/set before and reset after each individual test in the annotated class.
+====
+
+== `@RestoreSystemProperties`
```
marcphilipp: Ditto

```suggestion
[[RestoreSystemProperties]]
== @RestoreSystemProperties
```

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,118 @@ locale or time zone need to be annotated with the respective annotation:
...
+include::example$java/example/SystemPropertyExtensionDemo.java[tag=systemproperty_method_combine_all_test]
+----
+
+[NOTE]
+====
+Using `@RestoreSystemProperties` is not necessary to restore system properties modified via `@ClearSystemProperty` or `@SetSystemProperty` - they each automatically restore the referenced properties.
+'Restore', is only needed if system properties are modified in some way _other than_ Clear and Set during a test.
+====
+
+=== `@RestoreSystemProperties` Limitations
```
marcphilipp: Should be enough since it's below `@RestoreSystemProperties`, right?

```suggestion
[[RestoreSystemProperties-limitations]]
=== Limitations
```

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,118 @@ locale or time zone need to be annotated with the respective annotation:
...
+However _during_ the execution of the annotated scope, the system properties are set to a cloned `Properties` object with these limitations:
+
+- Properties with non-string values are removed.
+- Nested `Properties` are flattened into a non-nested instance that has the same effective values.
+
+== Thread-Safety
+
+Since system properties are global state, reading and writing them during <<writing-tests-parallel-execution, parallel test execution>>  can lead to unpredictable results and flaky tests.
+The system property extension is prepared for that and tests annotated with `@ClearSystemProperty`, `@SetSystemProperty`, or `@RestoreSystemProperties` will never execute in parallel (thanks to https://docs.junit.org/current/api[resource locks]) to guarantee correct test results.
```
marcphilipp:
```suggestion
The system property extension is prepared for that and tests annotated with `@ClearSystemProperty`, `@SetSystemProperty`, or `@RestoreSystemProperties` will never execute in parallel (thanks to xref:writing-tests/parallel-execution.adoc#synchronization[resource locks]) to guarantee correct test results.
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/ClearSystemProperty.java
```diff
@@ -0,0 +1,77 @@
...
+ * After the annotated element has been executed, the original value or the value of the
+ * higher-level container is restored.</p>
+ *
+ * <p>{@code ClearSystemProperty} can be used on the method and on the class level.
+ * It is repeatable and inherited from higher-level containers. If a class is
+ * annotated, the configured property will be cleared before every test inside that
+ * class.</p>
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/parallel-execution.html" target="_top">parallel test execution</a>,
```

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,118 @@ locale or time zone need to be annotated with the respective annotation:
 
 Tests annotated in this way will never execute in parallel with tests annotated with
 `@DefaultLocale` or `@DefaultTimeZone`.
+
+[[writing-tests-built-in-extensions-SystemProperty]]
```
marcphilipp:
```suggestion
[[SystemProperty]]
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/ClearSystemProperty.java
```diff
@@ -0,0 +1,77 @@
...
+ * all tests annotated with {@link ClearSystemProperty}, {@link SetSystemProperty}, {@link ReadsSystemProperty}, and {@link WritesSystemProperty}
+ * are scheduled in a way that guarantees correctness under mutation of shared global state.</p>
+ *
+ * <p>For more details and examples, see
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-built-in-extensions-SystemProperty" target="_top">the documentation on <code>@ClearSystemProperty and @SetSystemProperty</code></a>.</p>
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/built-in-extensions.html#SystemProperty" target="_top">the documentation on <code>@ClearSystemProperty and @SetSystemProperty</code></a>.</p>
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/ReadsSystemProperty.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Marks tests that read system properties but don't use the system property extension themselves.
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
+ * all tests annotated with {@link ClearSystemProperty}, {@link SetSystemProperty}, {@link ReadsSystemProperty}, and {@link WritesSystemProperty}
+ * are scheduled in a way that guarantees correctness under mutation of shared global state.</p>
+ *
+ * <p>For more details and examples, see
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-built-in-extensions-SystemProperty" target="_top">the documentation on <code>@ClearSystemProperty</code> and <code>@SetSystemProperty</code></a>.</p>
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/built-in-extensions.html#SystemProperty" target="_top">the documentation on <code>@ClearSystemProperty</code> and <code>@SetSystemProperty</code></a>.</p>
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/RestoreSystemProperties.java
```diff
@@ -0,0 +1,78 @@
...
+ * annotated with {@code RestoreSystemProperties}.
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
+ * all tests annotated with {@link RestoreSystemProperties}, {@link SetSystemProperty},
+ * {@link ReadsSystemProperty}, and {@link WritesSystemProperty}
+ * are scheduled in a way that guarantees correctness under mutation of shared global state.</p>
+ *
+ * <p>For more details and examples, see
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-built-in-extensions-SystemProperty" target="_top">the documentation on
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/built-in-extensions.html#SystemProperty" target="_top">the documentation on
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SetSystemProperty.java
```diff
@@ -0,0 +1,85 @@
...
+ * configurations.</p>
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
+ * all tests annotated with {@link ClearSystemProperty}, {@link SetSystemProperty}, {@link ReadsSystemProperty},
+ * and {@link WritesSystemProperty} are scheduled in a way that guarantees correctness under mutation of shared global
+ * state.</p>
+ *
+ * <p>For more details and examples, see
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-built-in-extensions-SystemProperty" target="_top">the documentation on <code>@ClearSystemProperty</code> and <code>@SetSystemProperty</code></a>.</p>
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/built-in-extensions.html#SystemProperty" target="_top">the documentation on <code>@ClearSystemProperty</code> and <code>@SetSystemProperty</code></a>.</p>
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/WritesSystemProperty.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Marks tests that write system properties but don't use the system property extension themselves.
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
+ * all tests annotated with {@link ClearSystemProperty}, {@link SetSystemProperty}, {@link ReadsSystemProperty}, and {@link WritesSystemProperty}
+ * are scheduled in a way that guarantees correctness under mutation of shared global state.</p>
+ *
+ * <p>For more details and examples, see
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-built-in-extensions-SystemProperty" target="_top">the documentation on <code>@ClearSystemProperty</code> and <code>@SetSystemProperty</code></a>.</p>
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/built-in-extensions.html#SystemProperty" target="_top">the documentation on <code>@ClearSystemProperty</code> and <code>@SetSystemProperty</code></a>.</p>
```

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,118 @@ locale or time zone need to be annotated with the respective annotation:
...
+
+=== `@RestoreSystemProperties` Limitations
+
+The system `Properties` object normally acts like a map of strings. While strongly discouraged, it is possible to use non-string keys and values. It is also possible to create {jdk-javadoc-base-url}/java.base/java/util/Properties.html#%3Cinit%3E(java.util.Properties)[nested `Properties` with inherited default values]. `@RestoreSystemProperties` restores the original `Properties` object with all of its potential richness _after_ the annotated scope is complete.
+However _during_ the execution of the annotated scope, the system properties are set to a cloned `Properties` object with these limitations:
+
+- Properties with non-string values are removed.
+- Nested `Properties` are flattened into a non-nested instance that has the same effective values.
+
+== Thread-Safety
```
marcphilipp:
```suggestion
[[SystemProperties-thread-safety]]
=== Thread Safety
```

jupiter-tests/src/test/java/org/junit/jupiter/engine/extension/ExtensionRegistryTests.java
```diff
@@ -42,7 +42,7 @@
 class ExtensionRegistryTests {
 
 	private static final int NUM_CORE_EXTENSIONS = 7;
-	private static final int NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT = 1; // OpenTestReportGenerationSystemPropertyOverride
+	private static final int NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT = 0;
 	private static final int NUM_DEFAULT_EXTENSIONS = NUM_CORE_EXTENSIONS
 			+ NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT;
```
marcphilipp: This can be simplified, right?

jupiter-tests/src/test/java/org/junit/jupiter/engine/extension/ExtensionRegistryTests.java
```diff
@@ -83,7 +83,7 @@ void registryIncludesAndExcludesSpecificAutoDetectedExtensions() {
 
 		List<Extension> extensions = registry.getExtensions(Extension.class);
 
-		assertEquals(NUM_DEFAULT_EXTENSIONS, extensions.size());
+		assertEquals(NUM_DEFAULT_EXTENSIONS + 1, extensions.size());
```
marcphilipp: Why +1?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/ReadsSystemProperty.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apiguardian.api.API;
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
+import org.junit.jupiter.api.parallel.Resources;
+
+/**
+ * Marks tests that read system properties but don't use the system property extension themselves.
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/parallel-execution.html" target="_top">parallel test execution</a>,
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/RestoreSystemProperties.java
```diff
@@ -0,0 +1,78 @@
...
+ * <p>When placed on a test class, a snapshot of system properties is stored prior to any
+ * {@code @BeforeAll} blocks in scope and before any {@link SetSystemProperty @SetSystemProperty}
+ * or {@link ClearSystemProperty @ClearSystemProperty} annotations on that class.
+ * After the test class completes, system properties are restored from the snapshot after any
+ * {@code @AfterAll} blocks have completed.
+ * In addition, a class level annotation is inherited by each test method just as if each one was
+ * annotated with {@code RestoreSystemProperties}.
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/parallel-execution.html" target="_top">parallel test execution</a>,
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/SetSystemProperty.java
```diff
@@ -0,0 +1,85 @@
...
+ * restored.</p>
+ *
+ * <p>{@code SetSystemProperty} can be used on the method and on the class level.
+ * It is repeatable and inherited from higher-level containers. If a class is
+ * annotated, the configured property will be set before every test inside that
+ * class. Any method-level configurations will override the class-level
+ * configurations.</p>
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/parallel-execution.html" target="_top">parallel test execution</a>,
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/WritesSystemProperty.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apiguardian.api.API;
+import org.junit.jupiter.api.parallel.ResourceAccessMode;
+import org.junit.jupiter.api.parallel.ResourceLock;
+import org.junit.jupiter.api.parallel.Resources;
+
+/**
+ * Marks tests that write system properties but don't use the system property extension themselves.
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/user-guide/#writing-tests-parallel-execution" target="_top">parallel test execution</a>,
```
marcphilipp:
```suggestion
 * <a href="https://docs.junit.org/current/writing-tests/parallel-execution.html" target="_top">parallel test execution</a>,
```

marcphilipp:
> > The way I see it, there is a small window for tests to flake if this extension mutates the system properties during cloning whilst other tests running in parallel are using them.
> 
> Good point.

I also think we should avoid mutating it

marcphilipp:
**Team decision:** Use `Properties.clone()` to create the copy and add best-effort check for `defaults` being not empty by enumerating the property keys and comparing them with the `map` keys.

jupiter-tests/src/test/java/org/junit/jupiter/engine/extension/ExtensionRegistryTests.java
mpkorstanje: Previously these were the available extensions:

```
0 = {DisabledCondition@5781} 
1 = {AutoCloseExtension@5782} 
2 = {TimeoutExtension@5783} 
3 = {RepeatedTestExtension@5784} 
4 = {TestInfoParameterResolver@5785} 
5 = {TestReporterParameterResolver@5786} 
6 = {TempDirectory@5787} 
7 = {ServiceLoaderExtension@5788} 
8 = {ConfigLoaderExtension@5789} 
9 = {OpenTestReportGenerationSystemPropertyOverride@5790} 
```

This results in equation:
 
 .1. `NUM_CORE_EXTENSIONS(7) + NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT(1) + DEMO_EXTENSIONS(2) = 10`.

This test explicitly excluded the `ConfigLoaderExtension`, explicitly included the `ServiceLoaderExtension` and because of that implicitly also excluded the `OpenTestReportGenerationSystemPropertyOverride` extension.

With two excluded extensions this result in equation:

.2. `NUM_CORE_EXTENSIONS(7) + NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT(1) + DEMO_EXTENSIONS(2) - EXCLUDED(2) = 8`.

This was inaccurately simplified to:

.3. `NUM_CORE_EXTENSIONS(7) + NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT(1) = 8`.

Deleting the `OpenTestReportGenerationSystemPropertyOverride` means that we have to reduce
`NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT` by 1, reduce EXCLUDED by 1. The right hand side does not change because the deleted extension was excluded. 

Applying that to equation .2. yields:

.4. `NUM_CORE_EXTENSIONS(7) + NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT(0) + DEMO_EXTENSIONS(2) - EXCLUDED(1) = 8`.

Which simplifies (with inaccuracy) to:

.5. `NUM_CORE_EXTENSIONS(7) + NUM_AUTO_REGISTERED_EXTENSIONS_IN_THIS_PROJECT(0) + 1 = 8`.

The confusing part is that we jumped from inaccurate equation .3. to inaccurate .5. which makes the + 1 appear out of nowhwere.

documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
```diff
@@ -292,3 +292,128 @@ locale or time zone need to be annotated with the respective annotation:
...
+Using `@RestoreSystemProperties` is not necessary to restore system properties modified via `@ClearSystemProperty` or `@SetSystemProperty` - they each automatically restore the referenced properties.
+'Restore', is only needed if system properties are modified in some way _other than_ Clear and Set during a test.
+====
+
+[[SystemProperties-thread-safety]]
+=== Thread Safety
+
+TODO: Fix links here:
+
+Since system properties are global state, reading and writing them during <<writing-tests-parallel-execution, parallel test execution>>  can lead to unpredictable results and flaky tests.
```
mpkorstanje: Fix link

documentation/antora.yml
```diff
@@ -165,6 +165,10 @@ asciidoc:
     ReadsDefaultTimeZone: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/util/ReadsDefaultTimeZone.html[@ReadsDefaultTimeZone]'
     WritesDefaultLocale: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/util/WritesDefaultLocale.html[@WritesDefaultLocale]'
     WritesDefaultTimeZone: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/util/WritesDefaultTimeZone.html[@WritesDefaultTimeZone]'
+    # System Properties Extension
+    ClearSystemProperty: TODO
+    SetSystemProperty: TODO
+    Restoresystemproperties: TODO
```
mpkorstanje: Finish these

junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/ClearSystemProperty.java
```diff
@@ -0,0 +1,77 @@
...
+ * After the annotated element has been executed, the original value or the value of the
+ * higher-level container is restored.</p>
+ *
+ * <p>{@code ClearSystemProperty} can be used on the method and on the class level.
+ * It is repeatable and inherited from higher-level containers. If a class is
+ * annotated, the configured property will be cleared before every test inside that
+ * class.</p>
+ *
+ * <p>During
+ * <a href="https://docs.junit.org/current/writing-tests/parallel-execution.html" target="_top">parallel test execution</a>,
```
mpkorstanje: Double check all links in javadoc

jupiter-tests/src/test/java/org/junit/jupiter/engine/extension/ExtensionRegistryTests.java
mpkorstanje: Clarified the whole thing in 15e79cfcd1c77a5e01a9205ecc7c8148ee4c921d.

marcphilipp (approved):

PR #5316, opened 2026-01-26
Add Kotlin code examples with tabs to User Guide

yalishevant:
- Add @asciidoctor/tabs extension for tabbed code blocks
- Add Kotlin examples for Writing Tests intro
- Add Kotlin examples for Dependency Injection documentation
- Add UI supplemental files (tabs.css, tabs.js) for tab styling

This PR adds the first tabbed Java/Kotlin code examples to the documentation, allowing users to switch between languages.

---

---

### Definition of Done

- [x] ~~Method preconditions are checked and documented~~ — N/A (no new API)
- [x] Change is covered by automated tests (Kotlin examples compile and tests pass)
- [x] ~~Public API has Javadoc and @API annotations~~ — N/A (documentation only)
- [x] Change is documented in the User Guide (this is the change itself)
- [x] ~~Release Notes~~ — N/A (documentation enhancement, no functional changes)

### DIFF
--- a/.github/actions/run-gradle/action.yml
+++ b/.github/actions/run-gradle/action.yml
@@ -17,7 +17,7 @@ runs:
         distribution: temurin
         java-version: 25
         check-latest: true
-    - uses: gradle/actions/setup-gradle@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # v5.0.0
+    - uses: gradle/actions/setup-gradle@f29f5a9d7b09a7c6b29859002d29d24e1674c884 # v5.0.1
       with:
         cache-encryption-key: ${{ inputs.encryptionKey }}
     - uses: testlens-app/setup-testlens@cbf46f117aa1b73f2f101616333e9625b46d4b95 # v1.6.1
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -45,7 +45,7 @@ jobs:
       with:
         persist-credentials: false
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
+      uses: github/codeql-action/init@45cbd0c69e560cd9e7cd7f8c32362050c9b7ded2 # v4.32.2
       with:
         languages: ${{ matrix.language }}
         build-mode: ${{ matrix.build-mode }}
@@ -60,6 +60,6 @@ jobs:
           -Dscan.tag.CodeQL \
           classes
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
+      uses: github/codeql-action/analyze@45cbd0c69e560cd9e7cd7f8c32362050c9b7ded2 # v4.32.2
       with:
         category: "/language:${{matrix.language}}"
--- a/.github/workflows/gradle-dependency-submission.yml
+++ b/.github/workflows/gradle-dependency-submission.yml
@@ -29,4 +29,4 @@ jobs:
         java-version: 25
         check-latest: true
     - name: Generate and submit dependency graph
-      uses: gradle/actions/dependency-submission@4d9f0ba0025fe599b4ebab900eb7f3a1d93ef4c2 # v5.0.0
+      uses: gradle/actions/dependency-submission@f29f5a9d7b09a7c6b29859002d29d24e1674c884 # v5.0.1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -101,7 +101,7 @@ jobs:
           publishAllPublicationsToMavenCentralSnapshotsRepository -x check \
           prepareGitHubAttestation
     - name: Generate build provenance attestations
-      uses: actions/attest-build-provenance@00014ed6ed5efc5b1ab7f7f34a39eb55d41aa4f8 # v3.1.0
+      uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
       with:
         subject-path: documentation/build/attestation/*.jar
 
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -57,6 +57,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
+        uses: github/codeql-action/upload-sarif@45cbd0c69e560cd9e7cd7f8c32362050c9b7ded2 # v4.32.2
         with:
           sarif_file: results.sarif
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,7 +65,7 @@ jobs:
             --remote-repo-url=${{ env.STAGING_REPO_URL }}
       - name: Generate build provenance attestations
         if: ${{ inputs.dryRun == false }}
-        uses: actions/attest-build-provenance@00014ed6ed5efc5b1ab7f7f34a39eb55d41aa4f8 # v3.1.0
+        uses: actions/attest-build-provenance@96278af6caaf10aea03fd8d33a09a777ca52d62f # v3.2.0
         with:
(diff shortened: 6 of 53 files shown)


documentation/modules/ROOT/pages/writing-tests/dependency-injection-for-constructors-and-methods.adoc
```diff
@@ -80,6 +117,29 @@ class MyRandomParametersTest {
...
+    @Test
+    fun injectsInteger(@Random i: Int, @Random j: Int) {
+        assertNotEquals(i, j)
+    }
+
+    @Test
+    fun injectsDouble(@Random d: Double) {
+        assertEquals(0.0, d, 1.0)
+    }
+}
```
marcphilipp: Please move this snippet into a `.kt` file in `src/test/java` so it gets compiled and executed.

documentation/src/test/kotlin/example/kotlin/TestReporterDemo.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestReporter
+import org.junit.jupiter.api.io.TempDir
+import org.junit.jupiter.api.parallel.Execution
+import org.junit.jupiter.api.parallel.ExecutionMode
+import java.nio.file.Files
+import java.nio.file.Path
+
+// tag::user_guide[]
+@Execution(ExecutionMode.SAME_THREAD)
```
marcphilipp:
```suggestion
@Execution(ExecutionMode.SAME_THREAD)
// tag::user_guide[]
```

documentation/modules/ROOT/pages/writing-tests/dependency-injection-for-constructors-and-methods.adoc
```diff
@@ -25,10 +25,26 @@ There are currently three built-in resolvers that are registered automatically.
...
 include::example$java/example/TestInfoDemo.java[tags=user_guide]
 ----
+--
+
+Kotlin::
++
+--
+[source,kotlin,indent=0]
+----
+include::example$kotlin/example/KotlinTestInfoDemo.kt[tags=user_guide]
```
marcphilipp: Please remove the Kotlin prefixes from the class names here and in the other places. You've probably did this for consistency with samples like `KotlinAssertionsDemo` but show special feature for Kotlin to justify the prefix.

```suggestion
include::example$kotlin/example/TestInfoDemo.kt[tags=user_guide]
```

documentation/modules/ROOT/pages/writing-tests/dependency-injection-for-constructors-and-methods.adoc
yalishevant: Let's then maybe also extract the Java version of MyRandomParametersTest to a file in src/test/java? Currently it's inline in the .adoc file, and extracting only the Kotlin version would be inconsistent. I can do both.

documentation/modules/ROOT/pages/writing-tests/dependency-injection-for-constructors-and-methods.adoc
yalishevant: I actually added the Kotlin cause without it I saw red code in IDE - I had duplicate FQNs in the same classpath:
example.MyFirstJUnitJupiterTests (Java)
example.MyFirstJUnitJupiterTests (Kotlin)

That's why I applied the same policy as for KotlinAssertionsDemo. So will it be fine to keep it, or would you prefer a different naming approach?

marcphilipp: Hmm, that's a good point. How about using a different package?

yalishevant: I'm fine with that. It could be example.kotlin.TestInfoDemo. 
Inside Kotlin team we rather get used to add Kt suffix, like TestInfoDemoKt. But as far as I see - it will break checks for "Demo" suffix. So yes, let's use example.kotlin.TestInfoDemo then.

marcphilipp: Alternatively, we could move the `Kotlin` test classes to a separate source set. Let's see if the package idea is feasible first, though.

yalishevant: Ok, done

sbrannen: FWIW, we use separate source sets in Spring Framework.

https://github.com/spring-projects/spring-framework/tree/main/framework-docs/src/main

marcphilipp: That's the same source set (`main`) but there's a config that avoids the redeclaration error: https://github.com/spring-projects/spring-framework/blob/516bf1c60627e053323ed6e369700cc0b3844bc7/framework-docs/framework-docs.gradle#L49

marcphilipp: I tried that locally and it complicates other things (e.g. using the `RandomNumberExtension` from both places) so let's stick with a separate package for now.

documentation/modules/ROOT/pages/writing-tests/dependency-injection-for-constructors-and-methods.adoc
marcphilipp: I missed that. Yes, we should do that as well!

yalishevant: It seems I know why it was inlined. The Java snippet uses RandomParametersExtension which is in the junit-examples repository, not here. If extracted to a file, it won't compile. The local RandomNumberExtension also only supports int, not double, so injectsDouble wouldn't work either. 
Options: 
- Keep Java inline (as is). And probably it will be right to inline Kotlin as well, with using both injectsInteger + injectsDouble).
- Extract Java but adapt it to use local @Random annotation and remove injectsDouble.

What do you prefer?

sbrannen: TBH, I don't see any point in referencing two different "random number" extensions within our User Guide.

It's of course OK for the `RandomParametersExtension` to remain in the examples repository, but I think it's fine to stick with the local `RandomNumberExtension` in the User Guide, both in [Dependency Injection for Constructors and Methods](https://docs.junit.org/6.1.0/writing-tests/dependency-injection-for-constructors-and-methods.html) as well as in [Registering Extensions](https://docs.junit.org/6.1.0/extensions/registering-extensions.html).

So, I'm in favor of revising the current Java-based example to use `RandomNumberExtension`, and the Kotlin-based example would copy that. Perhaps demonstrating parameter resolution for the constructor and a test method would suffice -- basically like `RandomNumberDemo` but without the field injection.

yalishevant: Ok! Then I've extracted both Java and Kotlin `MyRandomParametersTest` to separate files using the local `RandomNumberExtension`. The examples now demonstrate constructor and method parameter injection.

documentation/modules/ROOT/pages/writing-tests/dependency-injection-for-constructors-and-methods.adoc
```diff
@@ -63,23 +95,26 @@ simplicity and expressiveness of both the extension model and the parameter reso
 process. `MyRandomParametersTest` demonstrates how to inject random values into `@Test`
 methods.
 
+[tabs]
```
sbrannen, line 98: Please update the preceding paragraph to reference the `RandomNumberExtension` instead of the `RandomParametersExtension`.

yalishevant: Sure - it now references RandomNumberExtension

documentation/src/test/kotlin/example/kotlin/TestReporterDemo.kt
sbrannen: Since we don't want `@Execution(ExecutionMode.SAME_THREAD)` to show up in the code listing in the User Guide, please apply the suggested change.

sbrannen:
@yalishevant, thanks for making all of the requested changes. 👍 

@marcphilipp, since your review is still pending, do you have any final input before we merge this PR?

marcphilipp:
Thank you, @yalishevant! :+1:

Issue #5327, opened 2026-01-28
Failures to resolve selectors are not forwarded by Suite Engine


mpkorstanje:
Discovered as part of https://github.com/cucumber/cucumber-jvm/issues/3157.

## Steps to reproduce

1. Checkout https://github.com/cucumber/cucumber-jvm-starter-gradle-java/commit/0a8c759edec0a35e71fd0ef56d2966270bc8e168
2. Use `"io.cucumber:cucumber-bom:7.34.0"` in `build.gradle.kts`
3. Put a break point in the catch block of `EngineDiscoveryRequestResolution.resolveCompletely`
4. Run the tests.
5. Observe that a discovery issue is reported but not printed.

## Context

 - Used versions (Jupiter/Vintage/Platform): 6.0.2
 - Build Tool/IDE: Gradle

## Deliverables

- [x] Investigate why the discovery issue is not reported or forwarded.
- [x] Forward the discovery issues from the Suite Engine.

mpkorstanje:
Discovery issues are not forwarded because the `SuiteLauncher` invokes `EngineDiscoveryOrchestrator.discover` without a phase. Thus causing all issues to go unforwarded. Adding configuration parameter `junit.platform.discovery.issue.failure.phase=discovery` to the suite will result in the issues being forwarded.

Proposed solution: the `SuiteLauncher` is a `Launcher` and should provide it's launcher phase. This does require making `LauncherPhase` public.

mpkorstanje:
Interestingly enough we test against this in `discoveryIssueOfNestedTestEnginesAreReported` but in our case the test engine failed to discovery any test and is pruned.

mpkorstanje:
Additionally while the `DiscoveryIssueCollector` collects issues through `selectorProcessed`, the `DiscoveryIssueForwardingListener` does not forward `selectorProcessed` calls.

mpkorstanje:
Got it.

The problem occurs specifically when `EngineDiscoveryRequestResolution.resolveCompletely` encounters an exception while resolving a selector. This exception is turned into a `selectorProcessed` event which the `DiscoveryIssueCollector` then observes and turns into as a discovery issue. 

Under normal circumstances, this issue would then be reported by `EngineDiscoveryOrchestrator.reportDiscoveryIssues`.

But the Suite Engine doesn't use a discovery phase and instead the discovery request created by the suite engine forwards all discovery issues to its parent request. 

[...]

PR #5334, opened 2026-01-29
Separate discovery issue collection and creation

mpkorstanje:
During discovery the `EngineDiscoveryOrchestrator` uses the `DiscoveryIssueCollector` to collect `issueEncountered` and `selectorProcessed` events. The latter are then turned into issues and stored along side the other issues emitted by a test engine during discovery. Under normal circumstances this is not a problem as all issues would then be reported by `EngineDiscoveryOrchestrator.reportDiscoveryIssues`. 

But the Suite Engine doesn't use a discovery phase and instead the discovery request created by the suite engine forwards all discovery issues to its parent request. Because the issues created in response to `selectorProcessed` are not emitted by the `DiscoveryIssueCollector`, they are also not forwarded.

Separating the creation and collection of issues resolves this problem.

Closes: #5327

---

---

### Definition of Done

### DIFF
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
@@ -19,6 +19,7 @@ repository on GitHub.
 * Recheck done condition before rejecting extraneous worker in
   `WorkerThreadPoolHierarchicalTestExecutorService`
 * Add missing precondition checks to `Launcher` implementations
+* Failures to resolve selectors are now forwarded by Suite Engine
 
 [[v6.1.0-M2-junit-platform-deprecations-and-breaking-changes]]
 ==== Deprecations and Breaking Changes


junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/EngineDiscoveryRequestResolution.java
```diff
@@ -77,6 +77,7 @@ class EngineDiscoveryRequestResolution {
 		this.visitors = visitors;
 		this.defaultContext = new DefaultContext(null);
 		this.resolvedUniqueIds.put(engineDescriptor.getUniqueId(), Match.exact(engineDescriptor));
+		this.discoveryListener = new IssueReportingEngineDiscoveryListener(request.getDiscoveryListener());
```
mpkorstanje: @marcphilipp I'm not sure if this is the right scope. 

The `DiscoveryIssueCollector` is created in the `EngineDiscoveryOrchestrator` which would create issues for any test engine, not just those that use the `EngineDiscoveryRequestResolver`. Though  before discovery issues were introduced, the `EngineDiscoveryRequestResolution` was logging these issues instead, and I can't think of a good reason that scope might have expanded.

platform-tests/src/test/java/org/junit/platform/launcher/core/DiscoveryIssueCollectorTests.java
```diff
@@ -11,78 +11,21 @@
...
-import org.junit.platform.engine.support.descriptor.PackageSource;
-import org.junit.platform.engine.support.descriptor.UriSource;
 
 class DiscoveryIssueCollectorTests {
 
-	@ParameterizedTest(name = "{0}")
-	@MethodSource("pairs")
-	void reportsFailedResolutionResultAsDiscoveryIssue(DiscoverySelector selector, TestSource source) {
+	@Test
+	void reportsFailedResolutionResultAsDiscoveryIssue() {
```
marcphilipp:
```suggestion
	void reportsCollectedDiscoveryIssues() {
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/EngineDiscoveryRequestResolution.java
marcphilipp: This works for all engines using `EngineDiscoveryRequestResolver`. Since that's not mandatory, I changed it to be done in `EngineDiscoveryOrchestrator`.

marcphilipp (approved):

junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/EngineDiscoveryRequestResolution.java
mpkorstanje: Then we should probably document that engines should not create issues for failing to resolve their selectors.

marcphilipp: If they did, they would already be reported multiple times (unless the Suite engine is used). We can document it but I doubt it makes a difference.

PR #5337, opened 2026-01-30
Implement meaningful tests in `ColorPaletteTests`

rogozhinp:
This PR addresses the TODO in `DemonstratePalettesTests` that requested actual assertions instead of printing to SYSOUT.

Changes:

- Replace `System.out` with `StringWriter` to capture output
- Add assertions verifying ANSI color codes are present

----

### DIFF
--- a/platform-tests/src/test/java/org/junit/platform/console/output/ColorPaletteTests.java
+++ b/platform-tests/src/test/java/org/junit/platform/console/output/ColorPaletteTests.java
@@ -120,18 +120,8 @@ class ColorPaletteTests {
 	@Nested
 	class DemonstratePalettesTests {
 
-		private static final String ANSI_ESCAPE = "\u001B[";
-		// Default palette ANSI codes
-		private static final String SUCCESSFUL_GREEN = ANSI_ESCAPE + "32m";
-		private static final String FAILED_RED = ANSI_ESCAPE + "31m";
-		private static final String ABORTED_YELLOW = ANSI_ESCAPE + "33m";
-		private static final String SKIPPED_MAGENTA = ANSI_ESCAPE + "35m";
-		private static final String RESET = ANSI_ESCAPE + "0m";
-		// Single color palette ANSI codes
-		private static final String BOLD = ANSI_ESCAPE + "1m";
-		private static final String REVERSE = ANSI_ESCAPE + "7m";
-		private static final String UNDERLINE = ANSI_ESCAPE + "4m";
-		private static final String STRIKETHROUGH = ANSI_ESCAPE + "9m";
+		private static final String ESC = "\u001B[";
+		private static final String RESET = ESC + "0m";
 
 		@Test
 		void verbose_default() {
@@ -143,13 +133,14 @@ class ColorPaletteTests {
 			demoTestRun(listener);
 
 			String output = stringWriter.toString();
+			// @formatter:off
 			assertThat(output).contains(
-					"My Test",
-					SUCCESSFUL_GREEN,
-					FAILED_RED,
-					ABORTED_YELLOW,
-					SKIPPED_MAGENTA,
-					RESET);
+				withAnsi("{green}[OK] SUCCESSFUL"),
+				withAnsi("{red}[X] FAILED"),
+				withAnsi("{yellow}[A] ABORTED"),
+				withAnsi("{magenta}[S] SKIPPED")
+			);
+			// @formatter:on
 		}
 
 		@Test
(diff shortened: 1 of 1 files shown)


platform-tests/src/test/java/org/junit/platform/console/output/ColorPaletteTests.java
```diff
@@ -116,72 +117,130 @@ void cannotOverrideNone() {
...
+		private static final String BOLD = ANSI_ESCAPE + "1m";
+		private static final String REVERSE = ANSI_ESCAPE + "7m";
+		private static final String UNDERLINE = ANSI_ESCAPE + "4m";
+		private static final String STRIKETHROUGH = ANSI_ESCAPE + "9m";
+
 		@Test
 		void verbose_default() {
-			PrintWriter out = new PrintWriter(System.out);
+			StringWriter stringWriter = new StringWriter();
+			PrintWriter out = new PrintWriter(stringWriter);
```
sormuras:
```suggestion
			PrintWriter out = new PrintWriter(stringWriter, true);
```

Using the `PrintWriter(Writer out, boolean autoFlush)` constructor let's you remove the `out.flush();` line below.

Same goes for all other modified methods as well.

platform-tests/src/test/java/org/junit/platform/console/output/ColorPaletteTests.java
```diff
@@ -116,72 +117,130 @@ void cannotOverrideNone() {
...
 			TestExecutionListener listener = new VerboseTreePrintingListener(out, ColorPalette.DEFAULT, 16,
 				Theme.ASCII);
 
 			demoTestRun(listener);
-
-			assertDoesNotThrow(out::flush);
+			out.flush();
```
sormuras:
```suggestion
```

Remove after applying above's suggestion to use `PrintWriter(OutputStream out, boolean autoFlush)`.

rogozhinp: thank you! much nicer)

platform-tests/src/test/java/org/junit/platform/console/output/ColorPaletteTests.java
```diff
@@ -116,72 +117,124 @@ void cannotOverrideNone() {
...
+			StringWriter stringWriter = new StringWriter();
+			PrintWriter out = new PrintWriter(stringWriter, true);
 			TestExecutionListener listener = new VerboseTreePrintingListener(out, ColorPalette.DEFAULT, 16,
 				Theme.ASCII);
 
 			demoTestRun(listener);
 
-			assertDoesNotThrow(out::flush);
+			String output = stringWriter.toString();
+			assertTrue(output.contains("My Test"), "Output should contain test name");
```
sbrannen: I'd prefer that we use AssertJ here -- for example, `assertThat(output).contains(xxx, yyy, zzz);` -- in which case, I don't think we really need the "Output should contain..." error messages.

platform-tests/src/test/java/org/junit/platform/console/output/ColorPaletteTests.java
```diff
@@ -143,12 +143,12 @@ void verbose_default() {
...
-			assertTrue(output.contains(FAILED_RED), "Output should contain red for failed");
-			assertTrue(output.contains(ABORTED_YELLOW), "Output should contain yellow for aborted");
-			assertTrue(output.contains(SKIPPED_MAGENTA), "Output should contain magenta for skipped");
-			assertTrue(output.contains(RESET), "Output should contain ANSI reset codes");
+			assertThat(output).contains("My Test");
+			assertThat(output).contains(SUCCESSFUL_GREEN);
+			assertThat(output).contains(FAILED_RED);
+			assertThat(output).contains(ABORTED_YELLOW);
+			assertThat(output).contains(SKIPPED_MAGENTA);
+			assertThat(output).contains(RESET);
```
sbrannen: Thanks for switching to AssertJ!

Though, I think it might be clearer to collapse all of those into a single assertion -- something like the following.

```java
assertThat(output).contains(
		"My Test",
		SUCCESSFUL_GREEN,
		...
	);
```

rogozhinp: Thank you very much for your advise!

sbrannen: You're welcome, and thanks for making the change.

Though, I suppose the build is now failing for you due to formatting violations.

To have that nice one-value-per-line formatting, I guess you'll either have to add ` //` to the end of the each line or wrap those blocks of code in the following.

```java
// @formatter:off
<block of code>
// @formatter:on
```

sbrannen:
Please make sure you include the [DCO sign-off](https://github.com/junit-team/junit-framework/pull/5337/checks?check_run_id=62007679750) in every commit, or alternatively squash all commits into a single commit (with the DCO sign-off) and force-push.

mpkorstanje (changes requested):
Currently the tests don't assert an outcome that is meaningful. The output should not just contain some color, the context in which that color is used also matters.

So I think it would make sense if the output was asserted against a text block. And because ANSI escapes aren't quite readable, some template formatting might also be needed. For example:

```java
assertThat(output).contains(withAnsi(
    """
    {green}A passing test{/green}
    {yellow}A aborted test{/yellow}
    ect,
    ect
    """
));
```

Where `withAnsi` would replace the place holders with their ansi escapes.

sbrannen:
This has been merged into `main` in 3940bf3037ce4d0435003a4b937a7390bad63c5d and revised in 5befa089942698e062c46740b43b2a29b4cc5fb8.

Thanks 👍

rogozhinp:
Thank you!

Issue #5346, opened 2026-02-02
Deadlock in `NamespacedHierarchicalStore.computeIfAbsent()`


muellerj2:
#5231 introduced a deadlock in `NamespacedHierarchicalStory.computeIfAbsent()`. I observed the following thread states when running tests concurrently on JUnit 6.0.2:

Thread A:
```
   java.lang.Thread.State: BLOCKED (on object monitor)
        at java.util.concurrent.ConcurrentHashMap.transfer(java.base@21.0.8/ConcurrentHashMap.java:2483)
        - waiting to lock <0x00000000820f3f10> (a java.util.concurrent.ConcurrentHashMap$Node)
        at java.util.concurrent.ConcurrentHashMap.addCount(java.base@21.0.8/ConcurrentHashMap.java:2354)
        at java.util.concurrent.ConcurrentHashMap.compute(java.base@21.0.8/ConcurrentHashMap.java:2002)
        at org.junit.platform.engine.support.store.NamespacedHierarchicalStore.computeIfAbsent(NamespacedHierarchicalStore.java:272)
        at org.junit.platform.engine.support.store.NamespacedHierarchicalStore.computeIfAbsent(NamespacedHierarchicalStore.java:353)
        at org.junit.jupiter.engine.execution.NamespaceAwareStore.lambda$computeIfAbsent$1(NamespaceAwareStore.java:92)
        at org.junit.jupiter.engine.execution.NamespaceAwareStore$$Lambda/0x000001e4d52e50c0.get(Unknown Source)
        at org.junit.jupiter.engine.execution.NamespaceAwareStore.accessStore(NamespaceAwareStore.java:120)
        at org.junit.jupiter.engine.execution.NamespaceAwareStore.computeIfAbsent(NamespaceAwareStore.java:93)
        at org.junit.jupiter.engine.extension.TimeoutExtension.getGlobalTimeoutConfiguration(TimeoutExtension.java:174)
```

Thread B:
```
   java.lang.Thread.State: WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@21.0.8/Native Method)
        - parking to wait for  <0x000000008cb5ee90> (a java.util.concurrent.FutureTask)
        at java.util.concurrent.locks.LockSupport.park(java.base@21.0.8/LockSupport.java:221)
        at java.util.concurrent.FutureTask.awaitDone(java.base@21.0.8/FutureTask.java:500)
[...]

PR #5348, opened 2026-02-03
Fix deadlock in NamespacedHierarchicalStore.computeIfAbsent()

mpkorstanje:
In #5231 evaluation of the default value creator was deferred until after an entry was created in the concurrent hash map backing the store. This enables recursive updates.  However, by calling `evaluateIfNotNull` inside `compute` threads that lost the race would wait inside `Map::compute` for the value to be evaluated. 

https://github.com/junit-team/junit-framework/blob/bfd70fdbd6d2303019120bc997f6bb77847a50b6/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java#L272-L280

Waiting here is a problem when the backing concurrent hashmap is reaching capacity. The thread waiting inside compute holds a lock while at the same time the thread that won the race will try to acquire a lock to resize the map.

By making the threads that lost the race wait outside of `Map::compute` we can prevent this dead lock. This does require that each thread retries when another thread failed to insert a value. But eventually either one other thread either successfully inserts a value or the thread uses its own default value creator.

Fixes: #5346

---

---

### Definition of Done

### DIFF
--- a/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
+++ b/junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
@@ -237,7 +237,7 @@ public final class NamespacedHierarchicalStore<N> implements AutoCloseable {
 					// Condition c guards against race conditions (repeated from
 					// getStoredValue) this filters out failures inserted by
 					// computeIfAbsent.
-					if (oldStoredValue != null && (oldStoredValue.isDone() || oldStoredValue.isPresent())) {
+					if (oldStoredValue != null && (!oldStoredValue.isDone() || oldStoredValue.isPresent())) {
 						return oldStoredValue;
 					}
 					rejectIfClosed();


junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -223,22 +223,37 @@ public void close() {
...
-				return candidateStoredValue;
-			});
 
-		// Only the caller that created the candidateStoredValue may run it
-		if (candidateStoredValue.equals(storedValue)) {
-			return candidateStoredValue.execute();
+		for (;;) {
+			var storedValue = storedValues.compute(compositeKey, //
+				(__, oldStoredValue) -> {
+					// The old store value is remains if a) there is an old stored
```
marcphilipp:
```suggestion
					// The old stored value remains if a) there is an old stored
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -223,22 +223,37 @@ public void close() {
...
+					// value and b) the old stored value is not yet evaluated or c) the old
+					// stored value evaluated to a present value.
```
marcphilipp:
```suggestion
					// stored value was evaluated to a present value.
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -269,29 +284,43 @@ public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? e
...
 		});
-		var storedValue = storedValues.compute(compositeKey, (__, oldStoredValue) -> {
-			// guard against race conditions
-			// computeIfAbsent replaces both null and absent values
-			if (StoredValue.evaluateIfNotNull(oldStoredValue) != null) {
-				return oldStoredValue;
+
+		for (;;) {
+			var storedValue = storedValues.compute(compositeKey, (__, oldStoredValue) -> {
+				// The old store value is remains if a) there is an old stored
```
marcphilipp:
```suggestion
				// The old stored value remains if a) there is an old stored
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/store/NamespacedHierarchicalStore.java
```diff
@@ -269,29 +284,43 @@ public <K, V> Object computeIfAbsent(N namespace, K key, Function<? super K, ? e
...
+				// value and b) the old stored value is not yet evaluated or c) the old
+				// stored value evaluated to null.
```
marcphilipp, line 294:
```suggestion
				// stored value was evaluated to null.
```

marcphilipp (approved):
Thanks for taking care of this! The reasoning makes sense to me. I only added a few nitpicks regarding explanatory comment.

Issue #3930, opened 2024-08-20
Support arbitrary Java versions with `JRE` conditions


sbrannen:
## Overview

Inspired by #3918 and related discussions, we should consider introducing a mechanism for users to supply arbitrary Java feature versions that are not yet covered by the `JRE` enum constants.

Note that a "feature" version refers to the leading number in a Java version -- for example, the `21` in `21.0.4`. In Java 9, this was referred to as the _major_ number ([JEP 223](https://openjdk.org/jeps/223)), but the terminology changed to _feature_ in Java 10 ([JEP 322](https://openjdk.org/jeps/322)).

## Deliverables

- [x] Investigate options.
- [x] Revise `JRE` (via `JRE.java.jte`) to support feature versions.
- [x] Introduce `versions` attribute in `@EnabledOnJre` and revise `EnabledOnJreCondition` accordingly.
- [x] Introduce `versions` attribute in `@DisabledOnJre` and revise `DisabledOnJreCondition` accordingly.
- [x] Introduce `minVersion` and `maxVersion` attributes in `@EnabledForJreRange` and revise `EnabledForJreRangeCondition` accordingly.
- [x] Introduce `minVersion` and `maxVersion` attributes in `@DisabledForJreRange` and revise `DisbledForJreRangeCondition` accordingly.
- [x] Revise `EnabledOnJreIntegrationTests`, `DisabledOnJreIntegrationTests`, `EnabledOnJreConditionTests`, `DisabledOnJreConditionTests`, etc.
- [x] There are no TODOs left in the code.
- [x] Method [preconditions](https://junit.org/junit5/docs/snapshot/api/org.junit.platform.commons/org/junit/platform/commons/util/Preconditions.html) are checked and documented in the method's Javadoc.
- [x] [Coding conventions](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#coding-conventions) (e.g. for logging) have been followed.
- [x] Change is covered by [automated tests](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling.
- [x] Public API has [Javadoc](https://github.com/junit-team/junit5/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html).
[...]

PR #5352, opened 2026-02-04
Reliably support `JRE.OTHER` with `@EnabledOnJre` and `@DisabledOnJre`

### LINKED ISSUE #5341: `@EnabledOnJre` and `@DisabledOnJre` no longer work reliably with `JRE.OTHER`
## Overview

We migrated from JUnit 5.10 to 5.14, and `@DisabledOnJre(JRE.OTHER)` no longer works as it used to before.

For example, on 5.14 the highest supported JRE is 26, which means if I run with Java 27-ea then I would expect the test not to run. However, the test with `@DisabledOnJre(JRE.OTHER)` does run with Java 27-ea.

## Steps to reproduce

```java
public class ExampleTest {

    @Test
    @DisabledOnJre(JRE.OTHER)
    void shouldNotRunOnJava27Ea() {
        fail("Should not run on Java 27-ea");
    }
}
```

## Context

 - Used versions (Jupiter/Vintage/Platform): 5.14.2, 6.0.2
 - Build Tool/IDE:

## Related Issues

- #3930

sbrannen:
## Overview

In JUnit Jupiter 5.12, I added support for arbitrary Java versions with JRE conditions; however, I accidentally introduced a regression regarding `JRE.OTHER`. Specifically, prior to this commit, `JRE.OTHER` no longer worked reliably when used with `@EnabledOnJre` and `@DisabledOnJre`.

To address that, this commit revises the logic in `JRE.isCurrentVersion(int)` to account for situations where the supplied version value is `Integer.MAX_VALUE` (representing `@EnabledOnJre(OTHER)` or `@DisabledOnJre(OTHER)`), and the current version of the JVM is greater than the version of the last `JRE.JAVA_*` enum constant — for example, when running on Java 27 and the last JRE enum constant is `JAVA_26`.

## Related Issues

- #3930
- #5341
- #5364

---

---

### Definition of Done

- [x] Change is documented in the [Release Notes](https://docs.junit.org/snapshot/release-notes.html)

### DIFF
--- a/jupiter-tests/src/test/java/org/junit/jupiter/api/condition/EnabledForJreRangeIntegrationTests.java
+++ b/jupiter-tests/src/test/java/org/junit/jupiter/api/condition/EnabledForJreRangeIntegrationTests.java
@@ -26,6 +26,7 @@ import static org.junit.jupiter.api.condition.JavaVersionPredicates.onJava21;
 import static org.junit.jupiter.api.condition.JavaVersionPredicates.onJava22;
 import static org.junit.jupiter.api.condition.JavaVersionPredicates.onJava23;
 import static org.junit.jupiter.api.condition.JavaVersionPredicates.onKnownVersion;
+import static org.junit.jupiter.api.condition.JavaVersionPredicates.onOtherVersion;
 
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
@@ -152,7 +153,7 @@ class EnabledForJreRangeIntegrationTests {
 	@Test
 	@EnabledForJreRange(min = JAVA_20)
 	void min20() {
-		assertTrue(onKnownVersion());
+		assertTrue(onKnownVersion() || onOtherVersion());
 		assertTrue(JRE.currentVersionNumber() >= 20);
 		assertTrue(CURRENT_JRE.compareTo(JAVA_20) >= 0);
 		assertTrue(CURRENT_JRE.version() >= 20);
@@ -241,7 +242,7 @@ class EnabledForJreRangeIntegrationTests {
 	@Test
 	@EnabledForJreRange(minVersion = 21, maxVersion = Integer.MAX_VALUE)
 	void minVersion21MaxVersionMaxInteger() {
-		assertTrue(onKnownVersion());
+		assertTrue(onKnownVersion() || onOtherVersion());
 		assertTrue(JRE.currentVersionNumber() >= 21);
 	}
 
--- a/junit-jupiter-api/src/templates/resources/main/org/junit/jupiter/api/condition/JRE.java.jte
+++ b/junit-jupiter-api/src/templates/resources/main/org/junit/jupiter/api/condition/JRE.java.jte
@@ -169,17 +169,28 @@ public enum JRE {
 	}
 
 	/**
-	 * {@return {@code true} if the supplied version number is known to be the
-	 * Java Runtime Environment version for the currently executing JVM or if
-	 * the supplied version number is {@code -1} and the current JVM version
-	 * could not be determined}
+	 * Determine if the supplied version number is considered to be the current
+	 * JRE version.
 	 *
+	 * <p>Returns {@code true} if either of the following is {@code true}.
+	 *
+	 * <ul>
+	 * <li>The supplied version number is known to be the Java Runtime Environment
+	 * version for the currently executing JVM.</li>
+	 * <li>The supplied version number is {@link Integer#MAX_VALUE} and the current
+	 * {@code JRE} is {@link JRE#OTHER OTHER}.</li>
+	 * </ul>
+	 *
+	 * @return {@code true} if the supplied version number is considered to be
+	 * the current JRE version
 	 * @since 5.12
 	 * @see Runtime.Version#feature()
+	 * @see #isCurrentVersion()
+	 * @see #currentJre()
 	 */
 	@API(status = MAINTAINED, since = "5.13.3")
 	public static boolean isCurrentVersion(int version) {
-		return version == CURRENT_VERSION;
+		return (version == CURRENT_VERSION) || (version == JRE.currentJre().version);
 	}
 
 	static boolean isCurrentVersionWithinRange(int min, int max) {
--- a/junit-jupiter-api/src/templates/resources/testFixtures/org/junit/jupiter/api/condition/JavaVersionPredicates.java.jte
+++ b/junit-jupiter-api/src/templates/resources/testFixtures/org/junit/jupiter/api/condition/JavaVersionPredicates.java.jte
@@ -19,4 +19,9 @@ public class JavaVersionPredicates {
 		return @for(var jre : ForSupport.of(supportedJres))onJava${jre.get().getVersion()}()@if(!jre.isLast()) //
 				|| @endif@endfor;
 	}
+
+	static boolean onOtherVersion() {
+		return JRE.OTHER.isCurrentVersion();
+	}
+
 }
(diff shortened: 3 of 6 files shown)


mpkorstanje (commented):
I suppose this could be tested by generating a `TestDoubleJRE` from the `JRE.java.jte` template with a smaller set of configured JREs.

sbrannen:
> I suppose this could be tested by generating a `TestDoubleJRE` from the `JRE.java.jte` template with a smaller set of configured JREs.

I suppose that would be one way to make it "testable".

Though, we'd likely have to duplicate the [jre.yaml](https://github.com/junit-team/junit-framework/blob/main/gradle/base/code-generator-model/src/main/resources/jre.yaml) file, and we'd have to ensure that we have a CI build running a Java version greater than the last `JAVA_*` version in our `TestDoubleJRE`. Of course, I suppose the latter would always be the case these days.

Perhaps @marcphilipp can comment on the feasibility of such an approach, since he introduced the JTE (code generation) support in the Gradle build.

sbrannen:
> Though, we'd likely have to duplicate the [jre.yaml](https://github.com/junit-team/junit-framework/blob/main/gradle/base/code-generator-model/src/main/resources/jre.yaml) file,

Or perhaps we could just keep that file as-is and then _filter_ the JREs we process in our `TestDoubleJRE.java.jte` template file?

https://github.com/junit-team/junit-framework/blob/bfd70fdbd6d2303019120bc997f6bb77847a50b6/junit-jupiter-api/src/templates/resources/main/org/junit/jupiter/api/condition/JRE.java.jte#L25

sbrannen:
The thing I don't like about having something like `TestDoubleJRE.java.jte` is that we would have to sync any changes we make to the real `JRE.java.jte` to the dummy `TestDoubleJRE.java.jte`, and we (or future maintainers of JUnit) might neglect to do that.

And I doubt we could have any kind of inheritance between those [JTE](https://jte.gg/) files, but I also have to confess my ignorance: I'm a JTE novice.

mpkorstanje:
The code generation looks all quite convention based. So my idea would be to a) Parameterize the class name in `X.java.jte`. Then. generate a `X` and `TestDoubleX` for each template and then b) use `jre.yaml` to generate `X` and then filter `jre.yaml` to generate `TestDoubleX`. 

Because everything comes from the same source files, there would be no manual work needed aside from updating tests when new versions are added.

sbrannen:
> The code generation looks all quite convention based. So my idea would be to a) Parameterize the class name in `X.java.jte`. Then. generate a `X` and `TestDoubleX` for each template and then b) use `jre.yaml` to generate `X` and then filter `jre.yaml` to generate `TestDoubleX`.

So, basically, we'd then want to implement most of the "logic" to support this here:

https://github.com/junit-team/junit-framework/blob/bfd70fdbd6d2303019120bc997f6bb77847a50b6/gradle/plugins/code-generator/src/main/kotlin/junitbuild/generator/GenerateJreRelatedSourceCode.kt#L60-L77

Plus one or more additional JTE parameters that can be used within the `JRE.java.jte` to control how many `JAVA_*` enum constants are generated.

Is that what you have in mind?

> Because everything comes from the same source files, there would be no manual work needed aside from updating tests when new versions are added.

Indeed, if we have a single "source" that's parameterized in that limited fashion, that would avoid duplication and/or missing out on changes to the "real" template.

mpkorstanje:
> Plus one or more additional JTE parameters that can be used within the JRE.java.jte to control how many JAVA_* enum constants are generated.
> Is that what you have in mind?

Something like this, but cleaner.

```diff
diff --git a/gradle/plugins/code-generator/src/main/kotlin/junitbuild/generator/GenerateJreRelatedSourceCode.kt b/gradle/plugins/code-generator/src/main/kotlin/junitbuild/generator/GenerateJreRelatedSourceCode.kt
index d85b9dd79a..27222906fd 100644
--- a/gradle/plugins/code-generator/src/main/kotlin/junitbuild/generator/GenerateJreRelatedSourceCode.kt
+++ b/gradle/plugins/code-generator/src/main/kotlin/junitbuild/generator/GenerateJreRelatedSourceCode.kt
@@ -60,6 +60,7 @@ abstract class GenerateJreRelatedSourceCode : DefaultTask() {
             val minRuntimeVersion = 17
             val supportedJres = jres.filter { it.version >= minRuntimeVersion }
             val params = mapOf(
+                "classNamePrefix" to "",
                 "minRuntimeVersion" to minRuntimeVersion,
                 "allJres" to jres,
                 "supportedJres" to supportedJres,
@@ -75,6 +76,24 @@ abstract class GenerateJreRelatedSourceCode : DefaultTask() {
                     templateEngine.render(safePath, params, output)
                 }
             }
+
+            val testDoubleParams = mapOf(
+                "classNamePrefix" to "TestDouble",
+                "minRuntimeVersion" to minRuntimeVersion,
+                "allJres" to jres.dropLast(1),
+                "supportedJres" to supportedJres.dropLast(1),
+                "supportedJresSortedByStringValue" to supportedJres.dropLast(1).sortedBy { it.version.toString() },
+                "licenseHeader" to licenseHeaderFile.asFile.get().readText().trimEnd() + "\n",
+            )
+            templates.forEach {
+                val testTargetFile = testTargetDir.toPath().resolve(it.resolveSibling("TestDouble" + it.nameWithoutExtension).path)
+
+                FileOutput(testTargetFile).use { output ->
+                    // JTE does not support Windows paths, so we need to replace them
+                    val safePath = it.path.replace('\\', '/')
+                    templateEngine.render(safePath, testDoubleParams, output)
+                }
+            }
         }
     }
```

sbrannen:
> Something like this, but cleaner.

- See #5364

sbrannen:
> [ ] Change is covered by [automated tests](https://github.com/junit-team/junit-framework/blob/HEAD/CONTRIBUTING.md#tests) including corner cases, errors, and exception handling

Automated testing for corner cases will be implemented once #5364 has been completed and merged.

jupiter-tests/src/test/java/org/junit/jupiter/api/condition/EnabledForJreRangeIntegrationTests.java
```diff
@@ -241,7 +242,7 @@ void minVersion20MaxVersion21() {
 	@Test
 	@EnabledForJreRange(minVersion = 21, maxVersion = Integer.MAX_VALUE)
 	void minVersion21MaxVersionMaxInteger() {
-		assertTrue(onKnownVersion());
+		assertTrue(onKnownVersion() || onOtherVersion());
```
mpkorstanje, line 245: Should this change, and other uses of  `onOtherVersion` be moved to https://github.com/junit-team/junit-framework/pull/5364?

mpkorstanje (approved):
Changes to the template look good. But the I'd nitpick that we can pull out the test code to a different PR.

marcphilipp (approved):

jupiter-tests/src/test/java/org/junit/jupiter/api/condition/EnabledForJreRangeIntegrationTests.java
sbrannen: No, those tests fail if you run on a higher Java version -- for example if you run the tests with Java 27 EA.

On `main` this doesn't _currently_ matter, since we already have a `JAVA_27` constant in `JRE`.

However, on `6.0.x` and `5.14.x`, we only have a `JAVA_26` constant in `JRE`.

In other words, those tests are logically broken. It's just that we never ran them on a higher Java version than the last defined `JAVA_*` constant.

PR #5373, opened 2026-02-13
Make Kotlin suspend function checks `ClassLoader`-agnostic

geoand:
## Overview

This is done because with the current version of the code, in order for a method to be considered a Kotlin suspending method, Kotlin and the test class need to be loaded from the same `ClassLoader`, but in Quarkus that is not the case.

## Related Issues

- https://github.com/quarkusio/quarkus/issues/52480

----


marcphilipp:
@geoand Have you verified that this change solves the issue on the Quarkus side?

geoand:
I've verified that it gets me past the JUnit part and on to the part we have to fix eventually in Quarkus

marcphilipp:
Let's take a step back and start with an issue and a (minimal) reproducer so we can ensure we fully understand the problem before discussing a potential solution.

@geoand Could you please create an issue and supply a reproducer?

geoand:
Of course!

I opened https://github.com/junit-team/junit-framework/issues/5375

PR #5379, opened 2026-02-14
Move Jupiter's `Constants` to API package

marcphilipp:
Prior to this commit, the class was in the unexported
`org.junit.jupiter.engine` package so it could not be referenced from
the module path.

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
@@ -100,6 +100,22 @@ public final class Constants {
 	public static final String EXTENSIONS_AUTODETECTION_EXCLUDE_PROPERTY_NAME = "junit.jupiter.extensions.autodetection.exclude";
 
 	/**
+	 * Property name used to enable auto-detection and registration of extensions via
+	 * Java's {@link java.util.ServiceLoader} mechanism: {@value}
+	 *
+	 * <p>The default behavior is not to perform auto-detection.
+	 */
+	public static final String EXTENSIONS_AUTODETECTION_ENABLED_PROPERTY_NAME = "junit.jupiter.extensions.autodetection.enabled";
+
+	/**
+	 * Property name used to enable auto-closing of {@link AutoCloseable} instances: {@value}
+	 *
+	 * <p>By default, auto-closing is enabled.
+	 *
+	 */
+	public static final String CLOSING_STORED_AUTO_CLOSEABLE_ENABLED_PROPERTY_NAME = "junit.jupiter.extensions.store.close.autocloseable.enabled";
+
+	/**
 	 * Property name used to provide patterns for deactivating conditions: {@value}
 	 *
 	 * <h4>Pattern Matching Syntax</h4>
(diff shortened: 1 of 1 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
```diff
@@ -0,0 +1,415 @@
...
+import org.junit.jupiter.api.extension.PreInterruptCallback;
+import org.junit.jupiter.api.extension.TestInstantiationAwareExtension.ExtensionContextScope;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.api.parallel.Execution;
+
+/**
+ * Collection of configuration constants for the Jupiter test engine.
+ *
+ * @since 6.1
+ * @see <a href="https://docs.junit.org/6.0.2/running-tests/configuration-parameters.html">User Guide</a>
```
sbrannen:
```suggestion
 * @see <a href="https://docs.junit.org/current/running-tests/configuration-parameters.html">User Guide</a>
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
```diff
@@ -0,0 +1,415 @@
...
+	 * <li>{@code *System*, *Dev*}: excludes every extension whose FQCN contains
+	 * {@code System} or {@code Dev}.
+	 * <li>{@code org.example.MyExtension, org.example.TheirExtension}: excludes
+	 * extensions whose FQCN is exactly {@code org.example.MyExtension} or
+	 * {@code org.example.TheirExtension}.
+	 * </ul>
+	 *
+	 * <p>Note: A class that matches both an inclusion and exclusion pattern will be excluded.
+	 */
+	public static final String EXTENSIONS_AUTODETECTION_EXCLUDE_PROPERTY_NAME = "junit.jupiter.extensions.autodetection.exclude";
```
sbrannen, line 100: While we're at it, I think I'd go ahead and _group_ together all constants related to `junit.jupiter.extensions.*` so that they are adjacent within the source code.

Obviously not a big deal, but it seems more logical. 😉

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
```diff
@@ -0,0 +1,415 @@
...
+	 */
+	public static final String DEFAULT_TEMP_DIR_FACTORY_PROPERTY_NAME = TempDir.DEFAULT_FACTORY_PROPERTY_NAME;
+
+	/**
+	 * Property name used to set the default extension context scope for
+	 * extensions that participate in test instantiation: {@value}
+	 *
+	 * @see org.junit.jupiter.api.extension.TestInstantiationAwareExtension
+	 */
+	public static final String DEFAULT_TEST_CLASS_INSTANCE_CONSTRUCTION_EXTENSION_CONTEXT_SCOPE_PROPERTY_NAME = ExtensionContextScope.DEFAULT_SCOPE_PROPERTY_NAME;
```
sbrannen, line 409: Wow... that's a mouthful! 😮

But I suppose it's challenging to make it much shorter.

junit-jupiter-engine/src/main/java/module-info.java
```diff
@@ -27,8 +27,6 @@
 	requires org.junit.platform.engine;
 	requires org.opentest4j;
 
-	// exports org.junit.jupiter.engine; // Constants...
-
```
sormuras, line 31: Gone for good!

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/Constants.java
sormuras, line 1: In an ideal modular world, this class could have been deleted rigth away. Deprecating it for removal is the correct classpath-y way, though.

I assume, it's no longer referenced by any Jupiter Engine code from now on.

marcphilipp: Correct, it isn't referenced anymore.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
marcphilipp: Done in 3b4279611f89007cbd6bb9c98ca7404e5d2e9ce5

PR #5424, opened 2026-03-02
Introduce `TempDirDeletionStrategy`

### LINKED ISSUE #4567: Support ignoring errors during `@TempDir` cleanup
## Overview

There are various circumstances that can cause the cleanup of a temporary directory to fail. Some of these errors may indicate open or leaking file handles, issues in the implementation being tested, or problems in the test itself. Others may occur only occasionally on slow filesystems, race-conditions in asynchronous process and so on.

Currently, a test encountering such a problem will fail with something like:

<details>
<summary>org.junit.platform.commons.JUnitException: Failed to close extension context</summary>

```
org.junit.platform.commons.JUnitException: Failed to close extension context
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
Caused by: java.io.IOException: Failed to delete temp directory C:\Windows\TEMP\junit-6301891027881835454. The following paths could not be deleted (see suppressed exceptions for details): <root>, ws
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.stream.SortedOps$RefSortingSink.end(SortedOps.java:395)
	at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:258)
	at java.base/java.util.stream.Sink$ChainedReference.end(Sink.java:258)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:510)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596)
	... 2 more
	Suppressed: java.nio.file.DirectoryNotEmptyException: C:\Windows\TEMP\junit-6301891027881835454
		at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:272)
		at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
		at java.base/java.nio.file.Files.delete(Files.java:1152)
		at java.base/java.nio.file.Files.walkFileTree(Files.java:2828)
		at java.base/java.nio.file.Files.walkFileTree(Files.java:2882)
		... 13 more
	Suppressed: java.nio.file.DirectoryNotEmptyException: C:\Windows\TEMP\junit-6301891027881835454\ws
		at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:272)
		at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
		at java.base/java.nio.file.Files.delete(Files.java:1152)
		at java.base/java.nio.file.Files.walkFileTree(Files.java:2828)
		at java.base/java.nio.file.Files.walkFileTree(Files.java:2882)
		... 13 more
```
</details>

In my opinion, this is unfortunate and could also be considered a regression compared to JUnit 4. In JUnit 4, it was possible to work around these kinds of issues using `TemporaryFolder#assureDeletion`, which allowed ignoring failed deletion of the temporary directory.

As of now, there is no way to bypass these issues in JUnit 5 unless you define `@TempDir(cleanup = CleanupMode.NEVER)` or avoid using `@TempDir` altogether - both of which seem undesirable.

See #4549 for a working proposal.

marcphilipp:
Resolves #4567.

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirDeletionStrategy.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirDeletionStrategy.java
@@ -152,6 +152,10 @@ public interface TempDirDeletionStrategy {
 	 * Standard {@link TempDirDeletionStrategy} implementation that recursively
 	 * deletes all files and directories within the temporary directory.
 	 *
+	 * <p>Symbolic and other types of links, such as junctions on Windows, are
+	 * not followed. A warning is logged when deleting a link that targets a
+	 * location outside the temporary directory.
+	 *
 	 * <p>If a file or directory cannot be deleted, its permissions are reset
 	 * and deletion is attempted again. If deletion still fails, the path is
 	 * scheduled for deletion on JVM exit via
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDir.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDir.java
@@ -15,7 +15,6 @@ import static org.apiguardian.api.API.Status.MAINTAINED;
 import static org.apiguardian.api.API.Status.STABLE;
 
 import java.io.File;
-import java.io.IOException;
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
@@ -63,17 +62,16 @@ import org.junit.jupiter.api.extension.ParameterResolutionException;
  * {@code static} field or on a parameter of a
  * {@link org.junit.jupiter.api.BeforeAll @BeforeAll} method.
  *
- * <h2>Clean Up</h2>
+ * <h2>Cleanup/Deletion</h2>
  *
  * <p>By default, when the end of the scope of a temporary directory is reached,
  * &mdash; when the test method or class has finished execution &mdash; JUnit will
  * attempt to clean up the temporary directory by recursively deleting all files
  * and directories in the temporary directory and, finally, the temporary directory
- * itself. Symbolic and other types of links, such as junctions on Windows, are
- * not followed. A warning is logged when deleting a link that targets a
- * location outside the temporary directory. In case deletion of a file or
- * directory fails, an {@link IOException} will be thrown that will cause the
- * test or test class to fail.
+ * itself.
+ *
+ * <p>Two attributes allow customizing <em>when</em> (see {@link #cleanup()})
+ * and <em>how</em> (see {@link #deletionStrategy()}) to clean up.
  *
  * <p>The {@link #cleanup} attribute allows you to configure the {@link CleanupMode}.
  * If the cleanup mode is set to {@link CleanupMode#NEVER NEVER}, the temporary
@@ -83,6 +81,13 @@ import org.junit.jupiter.api.extension.ParameterResolutionException;
  * {@link CleanupMode#ALWAYS ALWAYS} clean up mode will be used, but this can be
  * configured globally by setting the {@value #DEFAULT_CLEANUP_MODE_PROPERTY_NAME}
  * configuration parameter.
+ *
+ * <p>The {@link #deletionStrategy()} attribute defines the strategy for
+ * performing the cleanup and dealing with errors such as undeletable files.
+ * By default, the {@link TempDirDeletionStrategy.Standard Standard} strategy is
+ * used which will cause a test or test class to fail in case deletion of a file
+ * or directory fails. This can be configured globally by setting the
+ * {@value #DEFAULT_DELETION_STRATEGY_PROPERTY_NAME} configuration parameter.
  *
  * @since 5.4
  */


strangelookingnerd:
Looks great so far! If I understand correctly `cleanup` is the _when_ and `deletionStrategy`is the _how_ temporary files are deleted. Using the synergy between these options should allow to handle most if not all cases we discussed in #4567

Overall I'd hope that there would be more implementations of `TempDirDeletionStrategy` besides `Standard` shipped with JUnit. 
I don't think it is feasible for each consumer to implement their own strategy. Further one could have already do so via `TempDir#factory` by implementing a `TempDirFactory`.

Having the option to _ignore_ errors is still relevant for me and other users seem to agree. 
Also the option to delete on VM exit - even though JUnit has no control over that - looks interesting enough for a best-effort option. Maybe it makes more sense to have this as a `TempDir#cleanup` implementation?

I fully understand that JUnit does not want to provide numerous options to not have the burden of maintaining them, still at least a curated selection should be available.

marcphilipp:
> If I understand correctly `cleanup` is the _when_ and `deletionStrategy`is the _how_ temporary files are deleted.

I'd say `cleanup` is _if_ not _when_ (in a temporal sense).

> Overall I'd hope that there would be more implementations of `TempDirDeletionStrategy` besides `Standard` shipped with JUnit.
> I don't think it is feasible for each consumer to implement their own strategy.

I agree but that can be a second step, maybe even in a later release after we've seen a few implementations.

> Further one could have already do so via `TempDir#factory` by implementing a `TempDirFactory`.

How do you mean that?

> Having the option to _ignore_ errors is still relevant for me and other users seem to agree.

The implementation could be as simple as that:

```java
public class ErrorIgnoringTempDirDeletionStrategy implements TempDirDeletionStrategy {

	@Override
	public Map<Path, Exception> delete(Path tempDir, AnnotatedElementContext elementContext,
			ExtensionContext extensionContext) throws IOException {
		Standard.INSTANCE.delete(tempDir,  elementContext, extensionContext);
		return emptyMap();
	}
}
```

Whether we want to provide this as a built-in implementation (and thus encouraging users to ignore errors), remains to be discussed.

strangelookingnerd:
> > Further one could have already do so via `TempDir#factory` by implementing a `TempDirFactory`.
> 
> How do you mean that?
> 

I was under the impression that one could already implement custom deletion behavior via `TempDirFactory#close` - on second look I see that this is only possible in a very limited way as `#close` is called just after the deletion.

> > Having the option to _ignore_ errors is still relevant for me and other users seem to agree.
> 
> The implementation could be as simple as that:
> 
> ```java
> public class ErrorIgnoringTempDirDeletionStrategy implements TempDirDeletionStrategy {
> 
> 	@Override
> 	public Map<Path, Exception> delete(Path tempDir, AnnotatedElementContext elementContext,
> 			ExtensionContext extensionContext) throws IOException {
> 		Standard.INSTANCE.delete(tempDir,  elementContext, extensionContext);
> 		return emptyMap();
> 	}
> }
> ```
> 
> Whether we want to provide this as a built-in implementation (and thus encouraging users to ignore errors), remains to be discussed.

Just want to emphasize that with the default being `Standard` (throwing exceptions on errors) you would not really _encourage_ users to ignore errors, but rather providing an option to do so by having _sensible_ alternatives. At least from an argumentative standpoint that's a different thing.

marcphilipp:
**Team decision:**

- Create proper return type rather than `Map<Path, Exception>`
- Add `TempDirDeletionStrategy.IgnoreFailures` that logs all failures but does not throw an exception

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirDeletionStrategy.java
```diff
@@ -0,0 +1,328 @@
...
+
+		@Override
+		public DeletionResult delete(Path tempDir, AnnotatedElementContext elementContext,
+				ExtensionContext extensionContext) throws IOException {
+
+			var result = delegate.delete(tempDir, elementContext, extensionContext);
+
+			if (!result.isSuccessful()) {
+				var exception = result.toException();
+				LOGGER.warn(exception, exception::getMessage);
```
mpkorstanje: Because this is tied to `isSuccessful` I would expect optional-like semantics.

Something like 

```
result.toException().ifPresent(exception -> {
    LOGGER.warn(exception, exception::getMessage);
});
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/DefaultDeletionResult.java
```diff
@@ -0,0 +1,86 @@
...
+				.map(DeletionFailure::path) //
+				.sorted() //
+				.distinct() //
+				.map(path -> relativizeSafely(rootDir(), path)) //
+				.map(path -> emptyPath.equals(path) ? "<root>" : path.toString()) //
+				.collect(joining(", "));
+		var exception = new DeletionException("Failed to delete temp directory " + rootDir().toAbsolutePath()
+				+ ". The following paths could not be deleted (see suppressed exceptions for details): " + joinedPaths);
+		failures().stream() //
+				.sorted(comparing(DeletionFailure::path)) //
```
mpkorstanje, line 48: Why sorted here? Shouldn't the suppressed exceptions be presented in the order they occurred?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirDeletionStrategy.java
```diff
@@ -0,0 +1,328 @@
...
+			this(Standard.INSTANCE);
+		}
+
+		IgnoreFailures(TempDirDeletionStrategy delegate) {
+			this.delegate = delegate;
+		}
+
+		@Override
+		public DeletionResult delete(Path tempDir, AnnotatedElementContext elementContext,
+				ExtensionContext extensionContext) throws IOException {
```
mpkorstanje, line 116: Would it be useful to include the `elementContext` in the log message so users can quickly find the offending test?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/DefaultDeletionResult.java
```diff
@@ -0,0 +1,86 @@
...
+		if (isSuccessful()) {
+			throw new IllegalStateException("Cannot create an exception for a successful deletion result");
+		}
+		var emptyPath = Path.of("");
+		var joinedPaths = failures().stream() //
+				.map(DeletionFailure::path) //
+				.sorted() //
+				.distinct() //
+				.map(path -> relativizeSafely(rootDir(), path)) //
+				.map(path -> emptyPath.equals(path) ? "<root>" : path.toString()) //
```
mpkorstanje: The `emptyPath` uses the default file system, while `path` may use any file system. So `path.toString()` may still yield an empty string.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirDeletionStrategy.java
marcphilipp: Done in b6d4430cc6094ac1d5f429865bcaf907a6fa5175

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/DefaultDeletionResult.java
marcphilipp: They are sorted the same way the undeletable paths are sorted in the message for consistency.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirDeletionStrategy.java
marcphilipp: Done in 8c55968928b5711cbdb4f710e8c2fa7abd23064a

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/DefaultDeletionResult.java
marcphilipp: Good catch! Fixed in ad967e102594247e2cdfe8f4ef1044e7ea702aee

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDirDeletionStrategy.java
```diff
@@ -0,0 +1,470 @@
...
+			String type = isConstructor ? "constructor" : "method";
+			String name = isConstructor ? executable.getDeclaringClass().getSimpleName() : executable.getName();
+			return "%s %s(%s)".formatted(type, name,
+				ClassUtils.nullSafeToString(Class::getSimpleName, executable.getParameterTypes()));
+		}
+	}
+
+	/**
+	 * Standard {@link TempDirDeletionStrategy} implementation that recursively
+	 * deletes all files and directories within the temporary directory.
```
mpkorstanje, line 153: "within the temporary directory" is doing some heavy lifting here. Would it be appropriate to mention symlinks and junctions here?

marcphilipp: Done in edbdfde79e629f94919d1346787336d9a002d8a6

PR #5432, opened 2026-03-04
Add Kotlin snippets for six User Guide sections

yalishevant:
Adds Kotlin tabbed code snippets to six more User Guide sections,
continuing the effort started in #5316.

**Pages updated:**
- Nested Tests
- Dynamic Tests
- Assumptions
- Exception Handling
- Repeated Tests
- Test Classes and Methods

Each Java code block is wrapped in `[tabs]` with a corresponding
Kotlin example that mirrors the same test logic using idiomatic
Kotlin: property access, string templates, reified generics,
trailing lambdas, `companion object` + `@JvmStatic` for lifecycle
methods, etc.

---

---

### Definition of Done

- ~Method preconditions are checked~ _(N/A — documentation only)_
- [x] Change is covered by automated tests
- ~Public API has Javadoc and `@API` annotations~ _(N/A — documentation only)_
- [x] Change is documented in the User Guide

### DIFF
--- a/documentation/modules/ROOT/pages/writing-tests/repeated-tests.adoc
+++ b/documentation/modules/ROOT/pages/writing-tests/repeated-tests.adoc
@@ -8,6 +8,11 @@ desired. Each invocation of a repeated test behaves like the execution of a regu
 The following example demonstrates how to declare a test named `repeatedTest()` that
 will be automatically repeated 10 times.
 
+[tabs]
+====
+Java::
++
+--
 [source,java]
 ----
 @RepeatedTest(10)
@@ -15,6 +20,20 @@ void repeatedTest() {
 	// ...
 }
 ----
+--
+
+Kotlin::
++
+--
+[source,kotlin]
+----
+@RepeatedTest(10)
+fun repeatedTest() {
+    // ...
+}
+----
+--
+====
 
 `@RepeatedTest` can be configured with a failure threshold which signifies the number of
 failures after which remaining repetitions will be automatically skipped. Set the
@@ -125,10 +144,26 @@ INFO: About to execute repetition 4 of 5 for repeatedTestInGerman
 INFO: About to execute repetition 5 of 5 for repeatedTestInGerman
 ....
 
-[source,java]
+[tabs]
+====
+Java::
++
+--
+[source,java,indent=0]
 ----
 include::example$java/example/RepeatedTestsDemo.java[tags=user_guide]
 ----
+--
+
+Kotlin::
++
+--
+[source,kotlin,indent=0]
+----
+include::example$kotlin/example/kotlin/RepeatedTestsDemo.kt[tags=user_guide]
+----
+--
+====
 
 When using the `ConsoleLauncher` with the unicode theme enabled, execution of
 `RepeatedTestsDemo` results in the following output to the console.
--- a/documentation/src/test/kotlin/example/kotlin/DynamicTestsDemo.kt
+++ b/documentation/src/test/kotlin/example/kotlin/DynamicTestsDemo.kt
@@ -1,0 +1,220 @@
+/*
+ * Copyright 2015-2026 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package example.kotlin
+
+// tag::user_guide[]
+import example.util.Calculator
+import example.util.StringUtils.isPalindrome
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertFalse
+import org.junit.jupiter.api.Assertions.assertNotNull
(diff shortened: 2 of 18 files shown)


documentation/src/test/kotlin/example/kotlin/DynamicTestsDemo.kt
```diff
@@ -0,0 +1,215 @@
...
+            dynamicTest("6th dynamic test") { assertEquals(4, calculator.multiply(2, 2)) }
+        ).iterator()
+
+    @TestFactory
+    fun dynamicTestsFromArray(): Array<DynamicTest> =
+        arrayOf(
+            dynamicTest("7th dynamic test") { assertTrue(isPalindrome("madam")) },
+            dynamicTest("8th dynamic test") { assertEquals(4, calculator.multiply(2, 2)) }
+        )
+
```
marcphilipp, line 74: I think we should add an example using a Kotlin `Sequence` here

documentation/src/test/kotlin/example/kotlin/exception/ExceptionAssertionExactDemo.kt
```diff
@@ -0,0 +1,40 @@
...
+    @ExpectToFail
+    // tag::user_guide[]
+    @Test
+    fun testExpectedExceptionIsThrown() {
+        // The following assertion succeeds because the code under assertion throws
+        // IllegalArgumentException which is exactly equal to the expected type.
+        // The assertion also returns the thrown exception which can be used for
+        // further assertions like asserting the exception message.
+        val exception =
+            assertThrowsExactly(IllegalArgumentException::class.java) {
```
marcphilipp: Note to self: we should add a Kotlin-specific variant with reified generics

documentation/src/test/kotlin/example/kotlin/ParameterizedClassDemo.kt
```diff
@@ -0,0 +1,61 @@
...
+    inner class FruitTests {
+        @Parameter
+        lateinit var fruit: String
+
+        @Nested
+        @ParameterizedClass
+        @ValueSource(ints = [23, 42])
+        inner class QuantityTests {
+            @Parameter
+            var quantity: Int = 0
```
marcphilipp, line 39: Doesn't `lateinit` work here?

```suggestion
            lateinit var quantity: Int
```

documentation/modules/ROOT/pages/writing-tests/repeated-tests.adoc
marcphilipp, line 22: What about this one?

documentation/src/test/kotlin/example/kotlin/DynamicTestsDemo.kt
marcphilipp: Done in

- https://github.com/junit-team/junit-framework/pull/5436

documentation/src/test/kotlin/example/kotlin/ParameterizedClassDemo.kt
yalishevant: `lateinit` cannot be used with primitive types (Int, Long, Boolean, etc.) in Kotlin. This is a language-level restriction because lateinit uses null as an internal sentinel to track initialization, and JVM primitives cannot be null.

documentation/src/test/kotlin/example/kotlin/DynamicTestsDemo.kt
yalishevant: Good suggestion! Added dynamicTestsFromSequence() which demonstrates Sequence<DynamicTest> as a return type - the idiomatic Kotlin equivalent of Stream.

documentation/src/test/kotlin/example/kotlin/exception/ExceptionAssertionExactDemo.kt
yalishevant: In the thread above you wrote that it's done, so I rebased onto latest main and updated ExceptionAssertionExactDemo.kt to use the new reified assertThrowsExactly<T> from #5436.

marcphilipp: Thanks! I meant to post my comment in this thread. 😅

documentation/modules/ROOT/pages/writing-tests/repeated-tests.adoc
yalishevant: Done.

yalishevant:
Thanks for the review! I've addressed all the comments:
- Added a Sequence<DynamicTest> example in DynamicTestsDemo.kt
- Wrapped the inline snippet in repeated-tests.adoc in tabs
- Rebased onto latest main and updated ExceptionAssertionExactDemo.kt to use the new reified assertThrowsExactly<T> from #5436
- Replied to the lateinit question in the thread

By the way - was six pages a reasonable scope for one PR, or would you prefer smaller/larger batches going forward?

marcphilipp:
> By the way - was six pages a reasonable scope for one PR, or would you prefer smaller/larger batches going forward?

I think that's a good size. If you'd prefer slightly larger to reduce the time you have to wait for my reviews, I'd be fine with that, though.

PR #5441, opened 2026-03-06
Include class template invocation index in legacy reporting names

### LINKED ISSUE #5409: Legacy XML reports generated for `@ParameterizedClass` tests do not contain parameters
This issue stems from #5296 which was closed because the report in question was being generated by Gradle, but the same issue exists in JUnit 6.0.3.

If you have a test using `@ParameterizedClass` and get it to generate legacy reports, the parameters do not appear in the report and you just get lots of duplicate entries for each method.  This makes it hard to tell which parameters failed and hard for other tools to process the report.

## Steps to reproduce

I used the `junit-platform-console-standalone-6.0.3.jar` to run the test from the command-line

The test was:

```java
import org.junit.jupiter.params.ParameterizedClass;

import java.util.Locale;

import static org.junit.jupiter.api.Assertions.*;

@ParameterizedClass(name="candidate={0}")
@ValueSource(strings = { "first", "second", "third" })
class ClassParamTest {

  private String candidate;

  public ClassParamTest(String candidate) {
    this.candidate = candidate;
  }

  @Test
  void notEmpty() {
    assertFalse(candidate.isEmpty());
  }

  @Test
  void isLowercase() {
    assertEquals(candidate.toLowerCase(Locale.ROOT), candidate);
  }
}
```

The build script was:

```bash
#!/bin/bash
cd ~/code/junit-test
JAR=junit-platform-console-standalone-6.0.3.jar
if [ ! -f "$JAR" ]; then
  curl -fL -o "$JAR" "https://repo1.maven.org/maven2/org/junit/platform/junit-platform-console-standalone/6.0.3/$JAR"
fi
mkdir -p out/test-classes reports
javac -cp "$JAR" -d out/test-classes test/ClassParamTest.java
java -cp "$JAR:out/test-classes" -jar "$JAR" execute -cp out/test-classes --reports-dir=reports --config=junit.platform.reporting.output.dir=reports --config=junit.platform.reporting.legacy.xml.enabled=true --scan-class-path
```

and the report generated was:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="JUnit Jupiter" tests="6" skipped="0" failures="0" errors="0" time="0.061" hostname="co1devps22.yb.lmax" timestamp="2026-02-26T17:39:08">
<properties>
<property name="file.encoding" value="UTF-8"/>
</properties>
<testcase name="isLowercase()" classname="ClassParamTest" time="0">
<system-out><![CDATA[
unique-id: [engine:junit-jupiter]/[class-template:ClassParamTest]/[class-template-invocation:#2]/[method:isLowercase()]
display-name: isLowercase()
]]></system-out>
</testcase>
<testcase name="isLowercase()" classname="ClassParamTest" time="0">
<system-out><![CDATA[
unique-id: [engine:junit-jupiter]/[class-template:ClassParamTest]/[class-template-invocation:#3]/[method:isLowercase()]
display-name: isLowercase()
]]></system-out>
</testcase>
<testcase name="notEmpty()" classname="ClassParamTest" time="0">
<system-out><![CDATA[
unique-id: [engine:junit-jupiter]/[class-template:ClassParamTest]/[class-template-invocation:#3]/[method:notEmpty()]
display-name: notEmpty()
]]></system-out>
</testcase>
<testcase name="notEmpty()" classname="ClassParamTest" time="0">
<system-out><![CDATA[
unique-id: [engine:junit-jupiter]/[class-template:ClassParamTest]/[class-template-invocation:#2]/[method:notEmpty()]
display-name: notEmpty()
]]></system-out>
</testcase>
<testcase name="notEmpty()" classname="ClassParamTest" time="0">
<system-out><![CDATA[
unique-id: [engine:junit-jupiter]/[class-template:ClassParamTest]/[class-template-invocation:#1]/[method:notEmpty()]
display-name: notEmpty()
]]></system-out>
</testcase>
<testcase name="isLowercase()" classname="ClassParamTest" time="0.013">
<system-out><![CDATA[
unique-id: [engine:junit-jupiter]/[class-template:ClassParamTest]/[class-template-invocation:#1]/[method:isLowercase()]
display-name: isLowercase()
]]></system-out>
</testcase>
<system-out><![CDATA[
unique-id: [engine:junit-jupiter]
display-name: JUnit Jupiter
]]></system-out>
</testsuite>
```

Note that I removed most of the properties as they are not really relevant

I would expect the test names to be something like:

```xml
<testcase name="isLowerCase[first]" classname="ClassParamTest" time="0.019"/>
```

## Context

 - Uses Jupiter 6.0.3

marcphilipp:
Resolves #5409.

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DynamicNodeTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/DynamicNodeTestDescriptor.java
@@ -45,7 +45,7 @@ abstract class DynamicNodeTestDescriptor extends JupiterTestDescriptor {
 	}
 
 	@Override
-	protected String getLegacyReportingBaseName() {
+	protected final String getLegacyReportingBaseName() {
 		// @formatter:off
 		return getParent()
 				.map(it -> {
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/MethodBasedTestDescriptor.java
@@ -87,7 +87,7 @@ public abstract class MethodBasedTestDescriptor extends JupiterTestDescriptor
 	}
 
 	@Override
-	protected String getLegacyReportingBaseName() {
+	protected final String getLegacyReportingBaseName() {
 		return "%s(%s)".formatted(getTestMethod().getName(),
 			ClassUtils.nullSafeToString(Class::getSimpleName, getTestMethod().getParameterTypes()));
 	}


jupiter-tests/src/test/java/org/junit/jupiter/params/ParameterizedClassIntegrationTests.java
```diff
@@ -464,6 +466,18 @@ void supportsNestedParameterizedClass(Class<?> classTemplateClass) {
...
+					.map(e -> e.getTestDescriptor().getLegacyReportingName());
+			assertThat(legacyReportingNames).containsExactly( //
+				"test(boolean, TestReporter)[1][1][1]", //
+				"test(boolean, TestReporter)[1][1][2]", //
+				"test(boolean, TestReporter)[1][2][1]", //
+				"test(boolean, TestReporter)[1][2][2]", //
+				"test(boolean, TestReporter)[2][1][1]", //
+				"test(boolean, TestReporter)[2][1][2]", //
+				"test(boolean, TestReporter)[2][2][1]", //
+				"test(boolean, TestReporter)[2][2][2]");
```
marcphilipp, line 480: These are the new names for a parameterized test within two parameterized classes.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
```diff
@@ -78,6 +81,33 @@ public abstract class JupiterTestDescriptor extends AbstractTestDescriptor
...
+
+	private IntStream getLegacyReportingIndexes() {
+		var ownIndex = getLegacyReportingIndex();
+		return getParent() //
+				.flatMap(it -> {
+					if (it instanceof JupiterTestDescriptor parent) {
+						return Optional.of(parent.getLegacyReportingIndexes());
+					}
+					return Optional.empty();
+				}) //
```
mpkorstanje: Wrapping with optional isn't needed because we're mapping an optional.

```suggestion
				.map(it -> it instanceof JupiterTestDescriptor parent ? parent.getLegacyReportingIndexes() : null) //
```

platform-tooling-support-tests/src/test/resources/platform/tooling/support/tests/AntStarterTests_snapshots/open-test-report.xml.snapshot
```diff
@@ -16,20 +16,20 @@ test-method: ant_starter
...
     <operatingSystem>Linux</operatingSystem>
     <cpuCores>16</cpuCores>
-    <java:javaVersion>24.0.1</java:javaVersion>
+    <java:javaVersion>25.0.2</java:javaVersion>
     <java:fileEncoding>UTF-8</java:fileEncoding>
     <java:heapSize max="16282288128"/>
   </infrastructure>
   
-  <e:started id="1" name="JUnit Jupiter" time="2025-07-22T05:46:39.391132136Z">
+  <e:started id="1" name="JUnit Jupiter" time="2026-03-06T12:56:11.394645169Z">
```
mpkorstanje, line 24: Not important for this PR, but it would be nice if the time could be stubbed. Then I think we might have noticed earlier that the naming for templates e.o wasn't great.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
```diff
@@ -128,7 +128,7 @@ public final Type getType() {
 	}
 
 	@Override
-	public final String getLegacyReportingName() {
+	protected final String getLegacyReportingBaseName() {
```
mpkorstanje, line 131: Why is this final, but other `getLegacyReportingBaseName` overrides are not?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/JupiterTestDescriptor.java
marcphilipp: Done in 07ee87c2e59ca25b60e5686f2f57dc0b55ea4093

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/descriptor/ClassBasedTestDescriptor.java
marcphilipp: I've now made the others `final` as well, where it made sense.

platform-tooling-support-tests/src/test/resources/platform/tooling/support/tests/AntStarterTests_snapshots/open-test-report.xml.snapshot
marcphilipp: I agree that change sets that update these snapshots are quite noisy and that it's hard to spot the relevant bits.

I think stubbing time would only work if we made it constant. That might then hide timestamp-related bugs, though. Moreover, this file is written from a JVM that's started by a JVM that runs Gradle/Maven/Ant so nothing we control directly.

We could maybe normalize all timestamps before the assertion but, again, it doesn't feel like a good solution to me, either. I'll ponder it for a bit.

PR #5524, opened 2026-03-17
Include full display name in legacy XML reports

### LINKED ISSUE #5409: Legacy XML reports generated for `@ParameterizedClass` tests do not contain parameters

marcphilipp:
Rather than just including the display name of the test, the display
names of all ancestors are now included as well, separated by the `>`
character. This includes display names of parents otherwise not
represented in the legacy XML format (for example, the invocation of a
`@ParameterizedClass`).

Resolves #5409.

### DIFF
--- a/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java
+++ b/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java
@@ -18,6 +18,7 @@ import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.counting;
 import static java.util.stream.Collectors.groupingBy;
 import static java.util.stream.Collectors.mapping;
+import static java.util.stream.Collectors.toCollection;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 import static org.junit.platform.commons.util.ExceptionUtils.readStackTrace;
@@ -36,6 +37,7 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.text.NumberFormat;
 import java.time.LocalDateTime;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.EnumSet;
@@ -44,10 +46,12 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.TreeSet;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import javax.xml.stream.XMLOutputFactory;
 import javax.xml.stream.XMLStreamException;
@@ -111,8 +115,8 @@ class XmlReportWriter {
 	private void writeXmlReport(TestIdentifier testIdentifier, Map<TestIdentifier, AggregatedTestResult> tests,
 			Writer out) throws XMLStreamException {
 
-		try (XmlReport report = new XmlReport(out)) {
-			report.write(testIdentifier, tests);
+		try (var report = new XmlReport(out)) {
+			report.write(testIdentifier, tests, this.reportData.getTestPlan());
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java
```diff
@@ -111,8 +115,8 @@ private boolean shouldInclude(TestPlan testPlan, TestIdentifier testIdentifier)
 	private void writeXmlReport(TestIdentifier testIdentifier, Map<TestIdentifier, AggregatedTestResult> tests,
 			Writer out) throws XMLStreamException {
 
-		try (XmlReport report = new XmlReport(out)) {
-			report.write(testIdentifier, tests);
+		try (var report = new XmlReport(out)) {
+			report.write(testIdentifier, tests, this.reportData.getTestPlan());
```
jbduncan, line 119: I've had a look at `XmlReport` and it looks like it uses `reportData.getTestPlan()` directly in a few places.

Have you considered either:
1. Inlining this parameter, or
2. Replacing all uses of `reportData.getTestPlan()` with this parameter?

jbduncan (commented):
Hi @marcphilipp - just a super minor nit, but otherwise this PR looks fantastic to me. Good work! 👍

junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java
marcphilipp: Thanks for the review! I went with option 1. Done in f71017219088244e623f00c6c24a1d0c6af45032

jbduncan: Brilliant, I'm happy with that. Thanks for making the fix so quickly! 👍

PR #5554, opened 2026-03-25
Adopt naming conventions for constant fields

mpkorstanje:
A constant field is a `static final` field whose value is immutable. If a static final field has a primitive type or an immutable reference type it is a constant field. Constants should be named with uppercase words separated by underscores. Non-constant `static final` fields should be named with camel case.

For example a `public static final String KEY` is immutable and a constant, while `private static final Logger logger` holds a reference to the logging system, is mutable and not a constant.  

While I've not checked all occurrences, we generally follow this convention in production code.  However when discussing [#5424](https://github.com/junit-team/junit-framework/pull/5424/changes#diff-d07eefdcb5de4095f89b0a210342e61731f794e581f6ea41910e416ae52c82e2R97) we noticed a drift towards `LOGGER`. Specifically:
  
 * all public constants are upper case.
 * we have 67 lowercase private constants, of which 34 are named `logger`.
 * we have 150 uppercase private constants, of which 5 are named `LOGGER`.

To ensure this drift doesn't persist:
* we should explicitly adopt a naming convention for constants.
* enforce that all non-private static final fields are named using the constant naming conventions. 
* enforce that loggers are never named following the constant naming conventions.

Out of scope: enforce that non-private static final fields are actually constants.

---

---

### Definition of Done


jbduncan:
Making loggers lowercase sounds reasonable to me. 

The rationale I've seen elsewhere for making loggers lowercase is that they're not really constants so much as side-effecting "do-ers" that just so happen to be `private static final` because it stops extra logger instances from being created. 

On the other hand, uppercase naming for `private static final` fields is near-universal across the ecosystem, regardless if they're constants or not, so I'd be equally happy with making everything uppercase.

I haven't searched the code for any of the other `private static final` fields, so I'm afraid I can't comment on them.

mpkorstanje:
From Joshua Blogs, Effective Java v2, p238

> The sole exception to the previous rule concerns “constant fields,” whose names should consist of one or more uppercase words separated by the underscore character, for example, VALUES or NEGATIVE_INFINITY. A constant field is a static final field whose value is immutable. If a static final field has a primitive type or an immutable reference type (Item 15), then it is a constant field. For example, enum constants are constant fields. If a static final field has a mutable reference type, it can still be a constant field if the referenced object is immutable.

The mutability is important here and I think we can consider loggers to be mutable. They're holding a reference to the logging system which holds state somewhere.

And combined with the advice to minimize accessibility and mutability for public parts of the API, I think we could adopt the convention that `public/protected/package-private static final` fields must be immutable and follow upcase naming convention. With private static final fields allowed to be mutable and named with either upper or lowercase. Except for logger, which is mutable.

CONTRIBUTING.md
```diff
@@ -54,6 +54,8 @@ I hereby agree to the terms of the JUnit Contributor License Agreement.
 
 ### Naming Conventions
 
+Acronym are words.
```
mpkorstanje:
```suggestion
Acronyms are words.
```

marcphilipp (approved):

PR #5559, opened 2026-03-26
Support multiple junit-platform.properties

### LINKED ISSUE #2794: Support multiple `junit-platform.properties` on the classpath
I implemented a [`ClassOrderer` for Quarkus](https://github.com/quarkusio/quarkus/blob/2.6.0.CR1/test-framework/junit5/src/main/java/io/quarkus/test/junit/util/QuarkusTestProfileAwareClassOrderer.java) and the idea was to enable that orderer by default for all Quarkus users (and also for the Quarkus CI).
But I overlooked the fact that `junit-platform.properties` is only expected once on the classpath and if there already is a project-specific properties file (e.g. with `junit.jupiter.extensions.autodetection.enabled`) there are many warnings issued by this code: https://github.com/junit-team/junit5/blob/r5.8.2/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherConfigurationParameters.java#L232-L236

Btw, that orderer is configurable, so the original idea also was to just add (as a Quarkus user) a custom `junit-platform.properties` to your project _with only the respective property_ (_without_ repeating `junit.jupiter.testclass.order.default`).

It would come in very handy if _all_ `junit-platform.properties` on the classpath would be merged, in the order as they appear on the cp.
I'm not sure whether a flag to go back to the previous behavior would be required, which is a bit of a chicken and egg situation I guess (where to put it?).

holly-cummins:
Ingest properties from all junit-platform.properties files on the classpath, rather than just the first. Where the same property exists in multiple files, maintain the current behaviour.

Resolves [#2794](https://github.com/junit-team/junit-framework/issues/2794)

This ended up being a fairly straightforward change, because all the testing was already in place, and just needed a slight extension for the new behaviour. What I've done is loaded *all* properties files with the expected name. I reverse the order so that if the same property is present in multiple files, the first occurrence 'wins'. 

I also kept the warning (but with amended wording, of course). If we wanted to be really fancy we could only warn if there actually is a conflict, but I wasn't sure if the extra effort was warranted.

I strengthened the assertions in the test to equals rather than contains, since it's easy to accidentally use similar values in test properties files.

I looked through the docs but couldn't find anywhere where it says you can only have one properties file, so I don't think there's anything to change there.

---

---

### Definition of Done

- [X] NOT APPLICABLE Public API has [Javadoc](https://github.com/junit-team/junit-framework/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [X]  Change is documented in the [User Guide](https://docs.junit.org/snapshot/) and [Release Notes](https://docs.junit.org/snapshot/release-notes.html)

### DIFF
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.0-M2.adoc
@@ -38,6 +38,8 @@ repository on GitHub.
   objects.
 * New `selectClasspathResources(String...)` and `selectClasspathResources(List<String)`
   methods in `DiscoverySelectors`.
+* Multiple `junit-platform.properties` files are now supported. Where the same property is set
+  in more than one file, the first value will be used.
 
 
 [[v6.1.0-M2-junit-jupiter]]
--- a/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherConfigurationParameters.java
+++ b/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/LauncherConfigurationParameters.java
@@ -255,9 +255,12 @@ class LauncherConfigurationParameters implements ConfigurationParameters {
 		Properties props = new Properties();
 
 		try {
-			URL configFileUrl = findConfigFile(configFileName);
-			if (configFileUrl != null) {
-				loadClasspathResource(configFileUrl, props);
+			List<URL> configFileUrls = findConfigFiles(configFileName);
+			// Reverse the list, so that files early on the classpath take priority where there's a conflict
+			Collections.reverse(configFileUrls);
+
+			for (URL url : configFileUrls) {
+				loadClasspathResource(url, props);
 			}
 		}
 		catch (Exception ex) {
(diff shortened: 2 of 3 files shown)


sormuras:
Hi Holly, I started the CI checks - and expect only the "Cross-Version / OpenJDK 27 (ea)" one to fail (due to #5605).

I note that the underlying issue

- #2794 

is still under discussion.

holly-cummins:
Oh dear, it looks like the Leyden job is failing as well (https://github.com/junit-team/junit-framework/actions/runs/23615551087/job/71287894863?pr=5559). It looks like either a hang or excessive slowness. I can't see anything obviously related to my changes in the thread dumps.

(Thanks for kicking the jobs off, @sormuras!)

marcphilipp:
Thank you for the PR, @holly-cummins! We're putting it "on hold" for the time being, though. Please see https://github.com/junit-team/junit-framework/issues/2794#issuecomment-4278245940 for details.

PR #5561, opened 2026-03-27
Replace `Arrays.asList()` and `addAll()` with `Collections.addAll()`

kawshikbuet17:
This PR simplifies a few internal collection update sites by replacing
Arrays.asList(...) used only as an intermediate step before addAll(...)
with Collections.addAll(...).

That makes the affected code paths more direct and easier to follow,
since the intent is simply to add array or varargs elements into an
existing collection. It also avoids creating intermediate list views for
those updates.

In ListenerRegistry, null-element validation is kept explicit before
adding the listeners, so the refactoring preserves the expected checks
while removing the extra conversion step.

No functional change intended.


sbrannen:

Thanks

PR #5564, opened 2026-03-27
Console selectors: accept Windows paths and spaces

kawshikbuet17:
## Summary
- Allow --select-file and --select-resource to accept Windows drive paths and paths with spaces.
- Preserve URI handling and line/column query parsing.
- Add unit tests for Windows paths with spaces and classpath resources with spaces.

## Motivation
Console selectors currently treat inputs as URIs, which breaks common Windows paths (e.g., C:\...) and quoted paths with spaces. This change makes the CLI more robust while keeping existing URI support.

## Testing
- ./gradlew :junit-platform-console:test --tests org.junit.platform.console.options.SelectorConverterTests

### DIFF
--- a/junit-platform-console/junit-platform-console.gradle.kts
+++ b/junit-platform-console/junit-platform-console.gradle.kts
@@ -22,6 +22,8 @@ dependencies {
 	osgiVerification(projects.junitJupiterEngine)
 	osgiVerification(projects.junitPlatformLauncher)
 	osgiVerification(libs.openTestReporting.tooling.spi)
+
+	testImplementation(projects.junitJupiter)
 }
 
 tasks {
@@ -73,5 +75,8 @@ tasks {
 		manifest {
 			attributes("Main-Class" to "org.junit.platform.console.ConsoleLauncher")
 		}
+	}
+	test {
+		useJUnitPlatform()
 	}
 }
--- a/junit-platform-console/src/main/java/org/junit/platform/console/options/SelectorConverter.java
+++ b/junit-platform-console/src/main/java/org/junit/platform/console/options/SelectorConverter.java
@@ -21,7 +21,9 @@ import static org.junit.platform.engine.discovery.DiscoverySelectors.selectUniqu
 import static org.junit.platform.engine.discovery.DiscoverySelectors.selectUri;
 
 import java.net.URI;
+import java.util.regex.Pattern;
 
+import org.jspecify.annotations.Nullable;
 import org.junit.platform.commons.PreconditionViolationException;
 import org.junit.platform.commons.util.ResourceUtils;
 import org.junit.platform.engine.DiscoverySelectorIdentifier;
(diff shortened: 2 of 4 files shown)


mpkorstanje:
I've had a quick look at your PR and  I get the impression this is still a work in progress so I've marked it as a draft.

Please keep the following in mind:

*  Make sure your pull request description includes the [pull request template](https://github.com/junit-team/junit-framework/blob/main/.github/PULL_REQUEST_TEMPLATE.md).
* Follow the Definition of Done from the pull request template.
* Give the [Contributing Guide](https://github.com/junit-team/junit-framework/blob/main/CONTRIBUTING.md) a read from top to bottom.

And as a general idea towards a solution if the string is not URI like, you don't have to fix the whitepspace. Eventually the string is passed to `new File()` which should handle whitespace just fine so an intermediate conversion to `URI` isn't nessesary or wanted.

~edit: On second thought, it would be good to be more specific when to apply the alternative behaviour and limit the potential blast radius. So it would make sense to check if the operating system is windows and only if the path contains a windows drive letter use the windows specific parsing.~

edit: No. It's more complicated. I've had to solve a similar problem in a different project in the past. Have a look at [FeaturePath.parse](https://github.com/cucumber/cucumber-jvm/blob/c24528d2acd44840a08001f0ae29b11cef99f3db/cucumber-core/src/main/java/io/cucumber/core/feature/FeaturePath.java#L35) for inspiration.

junit-platform-console/src/main/java/org/junit/platform/console/options/SelectorConverter.java
```diff
@@ -130,4 +134,63 @@ public DiscoverySelectorIdentifier convert(String value) {
...
+			if (parsed != null) {
+				return parsed;
+			}
+			if (value.indexOf(' ') >= 0) {
+				parsed = parseUri(value.replace(" ", "%20"));
+				if (parsed != null) {
+					return parsed;
+				}
+			}
+			URI.create(value);
```
mpkorstanje, line 149: ??

junit-platform-console/src/test/java/org/junit/platform/console/options/SelectorConverterTests.java
```diff
@@ -0,0 +1,44 @@
...
+
+	@Test
+	void fileSelectorAcceptsWindowsPathWithSpacesAndQuery() {
+		SelectorConverter.File converter = new SelectorConverter.File();
+		FileSelector selector = converter.convert("C:\\work\\My Tests\\FooTest.java?line=12&column=34");
+
+		assertEquals("C:\\work\\My Tests\\FooTest.java", selector.getRawPath());
+		FilePosition position = selector.getPosition().orElseThrow();
+		assertEquals(12, position.getLine());
+		assertEquals(34, position.getColumn().orElseThrow());
```
mpkorstanje, line 31: `FileSelector` implements equals so instead of testing each individual property, you can create the expected file selector and test it with a single `assertEquals` against the actual selector.

junit-platform-console/junit-platform-console.gradle.kts
```diff
@@ -22,6 +22,8 @@ dependencies {
 	osgiVerification(projects.junitJupiterEngine)
 	osgiVerification(projects.junitPlatformLauncher)
 	osgiVerification(libs.openTestReporting.tooling.spi)
+
+	testImplementation(projects.junitJupiter)
```
mpkorstanje, line 26: This shouldn't be needed. Please read the `README.md` in the test folder.

PR #5590, opened 2026-04-09
Add Kotlin snippets for ten User Guide sections

yalishevant:
Adds Kotlin tabbed code snippets to ten more User Guide sections,
continuing the effort started in #5316 and #5432.

**Pages updated (Writing Tests):**
- Built-in Extensions (@TempDir, @AutoClose, Locale/TimeZone, SystemProperty)
- Conditional Test Execution
- Display Names
- Annotations
- Timeouts
- Test Interfaces and Default Methods

**Pages updated (Extensions):**
- Parameter Resolution
- Keeping State in Extensions
- Test Lifecycle Callbacks
- Registering Extensions

---

---

### Definition of Done

- ~Method preconditions are checked~ _(N/A — documentation only)_
- [x] Change is covered by automated tests
- ~Public API has Javadoc and `@API` annotations~ _(N/A — documentation only)_
- [x] Change is documented in the User Guide


documentation/src/test/kotlin/example/kotlin/DefaultLocaleTimezoneExtensionDemo.kt
```diff
@@ -0,0 +1,137 @@
...
+    @DefaultTimeZone("Africa/Juba")
+    fun test_with_long_zone_id() {
+        assertThat(TimeZone.getDefault()).isEqualTo(TimeZone.getTimeZone("Africa/Juba"))
+    }
+    // end::default_timezone_zone[]
+
+    @Nested
+    // tag::default_timezone_class_level[]
+    @DefaultTimeZone("CET")
+    inner class MyTimeZoneTests {
```
marcphilipp: I guess there's no way to hide the `inner` keyword because of the formatter? I think we have two options: include `@Nested` as well or move it to a separate file. Thoughts?

documentation/src/test/kotlin/example/kotlin/SystemPropertyExtensionDemo.kt
```diff
@@ -0,0 +1,143 @@
...
+        assertThat(System.getProperty("1st property")).isNull()
+        assertThat(System.getProperty("2nd property")).isNull()
+        assertThat(System.getProperty("3rd property")).isEqualTo("new value")
+    }
+    // end::systemproperty_using_set_and_clear[]
+
+    @Nested
+    // tag::systemproperty_using_at_class_level[]
+    @ClearSystemProperty(key = "some property")
+    inner class MySystemPropertyTest {
```
marcphilipp: Ditto

documentation/src/test/kotlin/example/kotlin/SystemPropertyExtensionDemo.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    @Nested
+    @TestClassOrder(ClassOrderer.OrderAnnotation::class)
+    inner class SystemPropertyRestoreExample {
+        @Nested
+        @Order(1)
+        @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+        // tag::systemproperty_class_restore_setup[]
+        @RestoreSystemProperties
+        inner class MySystemPropertyRestoreTest {
```
marcphilipp, line 88: We should probably include the `@TestInstance` annotation here so one doesn't wonder why `beforeAll()` works. Same for the Java variant.

documentation/src/test/kotlin/example/kotlin/SystemPropertyExtensionDemo.kt
```diff
@@ -0,0 +1,143 @@
...
+                assertThat(System.getProperty("C")).isNull()
+            }
+        }
+        // end::systemproperty_class_restore_setup[]
+
+        @Nested
+        @Order(2)
+        // tag::systemproperty_class_restore_isolated_class[]
+        @ReadsSystemProperty
+        inner class SomeOtherTestClass {
```
marcphilipp: Ditto

documentation/modules/ROOT/pages/writing-tests/display-names.adoc
```diff
@@ -4,10 +4,26 @@ Test classes and test methods can declare custom display names via `@DisplayName
...
 include::example$java/example/DisplayNameDemo.java[tags=user_guide]
 ----
+--
+
+Kotlin::
++
+--
+[source,kotlin,indent=0]
+----
+include::example$kotlin/example/kotlin/DisplayNameDemo.kt[tags=user_guide]
```
marcphilipp, line 23: I wonder if we should a Kotlin-only example using backticks so spaces can be used in the function/test name directly. WDYT? 🤔

documentation/src/test/kotlin/example/kotlin/TimeoutDemo.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Tag
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.Timeout
+import org.junit.jupiter.api.Timeout.ThreadMode
+import java.util.concurrent.TimeUnit
+
+// tag::user_guide[]
+@Tag("timeout")
```
marcphilipp, line 21: We're not showing the tag in the Java snippet. The formatter enforces this ordering so I guess the only thing we can do to make it consistent is to include it in the Java version as well.

documentation/src/test/kotlin/example/kotlin/HttpServerDemo.kt
```diff
@@ -0,0 +1,51 @@
...
+        val address = server.address
+        val requestUrl = URI("http://${address.hostName}:${address.port}/example").toURL()
+
+        val responseBody = sendRequest(requestUrl)
+
+        assertEquals("This is a test", responseBody)
+    }
+
+    companion object {
+        @JvmStatic
```
marcphilipp: There's not a strong reason for this method/function to be static. Let's make it non-static to simplify the Kotlin version.

documentation/modules/ROOT/pages/extensions/keeping-state-in-extensions.adoc
marcphilipp, line 135: What about this one?

documentation/src/test/kotlin/example/kotlin/registration/DocumentationDemo.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.RegisterExtension
+import java.nio.file.Path
+
+// tag::user_guide[]
+class DocumentationDemo {
+    @RegisterExtension
+    val docs: DocumentationExtension =
+        DocumentationExtension.forPath(lookUpDocsDir())
```
marcphilipp, line 22: This needs `@JvmField`, doesn't it?

```suggestion
    @JvmField
    @RegisterExtension
    val docs =
        DocumentationExtension.forPath(lookUpDocsDir())
```

documentation/src/test/kotlin/example/kotlin/timing/TimingExtension.kt
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext.Store
+import java.util.logging.Logger
+
+// end::user_guide[]
+
+// tag::user_guide[]
+class TimingExtension :
+    BeforeTestExecutionCallback,
+    AfterTestExecutionCallback {
+    private val logger: Logger = Logger.getLogger(TimingExtension::class.java.name)
```
marcphilipp: Move into companion object?

marcphilipp (commented):
Thank you for the PR! 👍

documentation/src/test/kotlin/example/kotlin/DefaultLocaleTimezoneExtensionDemo.kt
yalishevant: Went with including `@Nested` - moved the `// tag::` markers before `@Nested` in both Java and Kotlin files so the annotation is now visible in the rendered snippets. This way the reader sees the full context without needing a separate file. IMO, some readers may find it useful. Let me know if you want to remove it from Java and keep only for Kotin.

documentation/src/test/kotlin/example/kotlin/SystemPropertyExtensionDemo.kt
yalishevant: Done - same fix applied here

documentation/src/test/kotlin/example/kotlin/SystemPropertyExtensionDemo.kt
yalishevant: Good point - moved the `// tag::` markers before `@TestInstance(PER_CLASS)` in both Java and Kotlin so readers can see why `beforeAll()` works without a companion object.

documentation/src/test/kotlin/example/kotlin/SystemPropertyExtensionDemo.kt
yalishevant: ✅

documentation/modules/ROOT/pages/writing-tests/display-names.adoc
yalishevant: Sure, great idea - added a TIP block after the `@DisplayName` example mentioning Kotlin's backtick-enclosed function names as an alternative for simple display names, with a short inline code sample.

Or you would rather wanted to see it as one more function inside `DisplayNameDemo` class? It will break symmetry but save space 🙂

documentation/src/test/kotlin/example/kotlin/TimeoutDemo.kt
yalishevant: Agreed - moved the `// tag::user_guide[]` marker before `@Tag("timeout")` in the Java file

documentation/src/test/kotlin/example/kotlin/HttpServerDemo.kt
yalishevant: Right - removed the companion object wrapper and @JvmStatic.

documentation/modules/ROOT/pages/extensions/keeping-state-in-extensions.adoc
yalishevant: Thanks! Added a Kotlin tab for the inline `MyResource` example in the Migration Note. Used a longer delimiter `======` for the outer TIP block to allow nesting the [tabs] block inside it.

documentation/src/test/kotlin/example/kotlin/registration/DocumentationDemo.kt
yalishevant: You're right! Added `@JvmField` before `@RegisterExtension`. Thank you

documentation/src/test/kotlin/example/kotlin/timing/TimingExtension.kt
yalishevant: Right, it mirrors `private static final` semantics

documentation/src/test/kotlin/example/kotlin/DefaultLocaleTimezoneExtensionDemo.kt
marcphilipp: Thanks! I found a way now using `@formatter:off`/`on` tags and have adjusted the samples accordingly.

marcphilipp (approved):

PR #5630, opened 2026-04-23
Introduce `type` annotation attribute for `EmptySource`

### LINKED ISSUE #4801: `@EmptySource` not working with `ArgumentConverter` accepting `@EmptySource` supported types
Currently, `EmptyArgumentsProvider` directly checks [the type of the first declared parameter](https://github.com/junit-team/junit-framework/blob/ebb561acdc79db2c26210c3bd059f349a59d1750/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptyArgumentsProvider.java#L51-L93) and does not know of potential converters declared for that parameter, so it will throw a `PreconditionViolationException` if the declared type doesn't match the expectations.

Looking ahead to when #853 is integrated, a globally-configured converter might even make the situation more complicated.

## Steps to reproduce

Given an `ArgumentConverter` accepting strings, like:

```java
class StringToBytesConverter extends TypedArgumentConverter<String, byte[]> {

	protected StringToBytesConverter() {
		super(String.class, byte[].class);
	}

	@Override
	protected byte[] convert(String source) {
		return source.getBytes();
	}

}
```

the following test passes:

```java
@ParameterizedTest
@FieldSource
void fieldSourceTest(@ConvertWith(StringToBytesConverter.class) byte[] bytes) {
    assertEquals(0, bytes.length);
}

static List<String> fieldSourceTest = List.of("");
```

but the following test does not:

```java
@ParameterizedTest
@EmptySource
void emptySourceTest(@ConvertWith(StringToBytesConverter.class) byte[] bytes) {
    assertEquals(0, bytes.length);
}
```

failing with:

```
org.junit.jupiter.api.extension.ParameterResolutionException: Error converting parameter at index 0: MyConverter cannot convert objects of type [[B]. Only source objects of type [java.lang.String] are supported.
```

Similar reproducers can be composed for all other `@EmptySource` supported types, like `Collection`, `Map`, etc.

## Context

 - Used versions (Jupiter/Vintage/Platform): 5.13.4
 - Build Tool/IDE: Maven/IntelliJ IDEA

## Deliverables

- [ ] `@EmptySource` provides an empty argument if there is a converter that can accept one of its supported types

marcphilipp:
Resolves #4801.

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptySource.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptySource.java
@@ -10,6 +10,7 @@
 
 package org.junit.jupiter.params.provider;
 
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.STABLE;
 
 import java.lang.annotation.Documented;
@@ -20,13 +21,14 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 import org.apiguardian.api.API;
+import org.junit.jupiter.params.converter.ArgumentConverter;
 
 /**
  * {@code @EmptySource} is an {@link ArgumentsSource} which provides a single
  * <em>empty</em> argument to the annotated {@code @ParameterizedClass}
  * or {@code @ParameterizedTest}.
  *
- * <h2>Supported Parameter Types</h2>
+ * <h2 id="supported-parameter-types">Supported Parameter Types</h2>
  *
  * <p>This argument source will only provide an empty argument for the following
  * parameter types.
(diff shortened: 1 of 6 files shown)


junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptySource.java
```diff
@@ -67,4 +73,18 @@
...
+	 * <p>Must be one of the
+	 * <a href="#supported-parameter-types">supported parameter types</a>.
+	 *
+	 * <p>Setting this attribute is usually not necessary because the type will
+	 * be derived from the first (and only) parameter. Setting it explicitly
+	 * allows using an {@link ArgumentConverter} to convert from the specified
+	 * type to the actual parameter type.
+	 */
+	@API(status = EXPERIMENTAL, since = "6.1")
+	Class<?> type() default EmptyArgumentsProvider.Unset.class;
```
mpkorstanje: Does this improve things?

```suggestion
	Class<?> type() default EmptyArgumentsProvider.Derived.class;
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/EmptySource.java
```diff
@@ -48,6 +50,10 @@
  * <li>object arrays &mdash; for example {@code String[]}, {@code Integer[][]}, etc.</li>
  * </ul>
  *
+ * <p>Unless the {@link #type()} is set, the parameter type is determined from
+ * the first (and only) parameter of the annotated {@code @ParameterizedClass}
```
mpkorstanje:
```suggestion
 * <p>Unless the {@link #type()} is set, the parameter type is derived from
 * the first (and only) parameter of the annotated {@code @ParameterizedClass}
```

Or `determined by the`.

mpkorstanje (commented):
LGTM but for some nitpicks.

PR #5633, opened 2026-04-24
 Reduce NullAway suppressions in throwing code

JarvisCraft:
Instead of using `throwFoo(); return null;` with suppressed `NullAway` lint just do `throw newFoo()`.

This removes the need for `NullAway` suppressions since now Java type system is happy deducing the return type of the methods to the bottom type which coerces to anything.

---

---

### Definition of Done

- [x] Change is documented in the [User Guide](https://docs.junit.org/snapshot/) and [Release Notes](https://docs.junit.org/snapshot/release-notes.html) (the change is effectively internal)

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
@@ -49,14 +49,14 @@ class AssertInstanceOf {
 	private static <T> T assertInstanceOf(Class<T> expectedType, @Nullable Object actualValue,
 			@Nullable Object messageOrSupplier) {
 		if (!expectedType.isInstance(actualValue)) {
-			assertionFailure() //
+			throw assertionFailure() //
 					.message(messageOrSupplier) //
 					.reason(actualValue == null ? "Unexpected null value" : "Unexpected type") //
 					.expected(expectedType) //
 					.actual(actualValue == null ? null : actualValue.getClass()) //
 					.cause(actualValue instanceof Throwable t ? t : null) //
 					.trimStacktrace(Assertions.class) //
-					.buildAndThrow();
+					.build();
 		}
 		return expectedType.cast(actualValue);
 	}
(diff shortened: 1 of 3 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
```diff
@@ -46,18 +46,17 @@ static <T> T assertInstanceOf(Class<T> expectedType, @Nullable Object actualValu
...
-			assertionFailure() //
+			throw assertionFailure() //
 					.message(messageOrSupplier) //
 					.reason(actualValue == null ? "Unexpected null value" : "Unexpected type") //
 					.expected(expectedType) //
 					.actual(actualValue == null ? null : actualValue.getClass()) //
 					.cause(actualValue instanceof Throwable t ? t : null) //
 					.trimStacktrace(Assertions.class) //
-					.buildAndThrow();
+					.build();
```
marcphilipp: I've just pushed 85d360fc87cbf82f4515ad03579c1765bcbfc463 which makes the change to this file unnecessary.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
```diff
@@ -117,10 +117,9 @@ protected Assertions() {
 	 * generic return type {@code V}.
 	 */
 	@Contract(" -> fail")
-	@SuppressWarnings({ "NullAway", "TypeParameterUnusedInFormals" })
+	@SuppressWarnings({ "TypeParameterUnusedInFormals" })
```
marcphilipp:
```suggestion
	@SuppressWarnings("TypeParameterUnusedInFormals")
```

and so on below

marcphilipp (changes requested):
Thanks for the PR! 👍

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Assertions.java
JarvisCraft: Good catch; fixed it

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertInstanceOf.java
JarvisCraft: I've rebased on master but I would still suggest keeping the explicit `throw` here since it operates on Java's type system level (e.g. it participates in reachability checks of javac, a la guaranteed local var initialization) which is less prone to breakages and more obvious when reviewing the code.

marcphilipp (approved):

PR #5654, opened 2026-05-04
Lock discussion on resolved issues

mpkorstanje:
There is a small trickle of people who comment on an old already closed issue with a question or feature request. Unfortunately, by doing this their comment falls outside of workflow. We generally only look at and discuss new issues.

Presumably people who do this have put in some effort to research our backlog for related items. By locking the thread with an instructional explanation we can hopefully nudge them into the right direction.

By default, [`dessant/lock-threads`](https://github.com/dessant/lock-threads) will lock the conversation on resolved issues, pull requests and discussions after 365 days. This has been limited to just issues.

---

---

### Definition of Done

### DIFF
--- a/.github/workflows/lock-resolved-issues.yml
+++ b/.github/workflows/lock-resolved-issues.yml
@@ -1,0 +1,26 @@
+name: Lock resolved issues
+
+on:
+  schedule:
+    - cron: '13 7 * * 6'
+  workflow_dispatch:
+
+concurrency:
+  group: lock-resolved-issues
+
+jobs:
+  action:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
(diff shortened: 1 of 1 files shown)


scordio:
@mpkorstanje what an email burst we're getting now 😅 

Maybe it's too late and a one-off, but what about just locking them without a comment? People would anyway raise a new issue when commenting is no longer allowed.

mpkorstanje:
Done 006583dfbb4846b30946a243eba33adb28241a46.

mpkorstanje:
> By locking the thread with an instructional explanation we can hopefully nudge them into the right direction.

We close about 15-20 issues every month. So this will result in a trickle of notifications bigger than the original problem. So for now we'll have to do without the friendly nudge.

PR #5701, opened 2026-05-26
Prefer @Nullable to Optional for InterceptingExecutableInvoker

mpkorstanje:
Same as https://github.com/junit-team/junit-framework/pull/5700, but for the `InterceptingExecutableInvoker` specifically. 

An interesting aspect of this API was that we open taken in nullable values, wrap them in an `Optional`, pass them down the call chain and then unwrap them later on. With JSpecify available this isn't necessary at all and we can also remove this behaviour.

---

---

### Definition of Done

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ReflectiveInvocationContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ReflectiveInvocationContext.java
@@ -17,6 +17,7 @@ import java.util.List;
 import java.util.Optional;
 
 import org.apiguardian.api.API;
+import org.jspecify.annotations.Nullable;
 
 /**
  * {@code ReflectiveInvocationContext} encapsulates the <em>context</em> of
@@ -56,7 +57,7 @@ public interface ReflectiveInvocationContext<T extends Executable> {
 	 * @return the arguments of the executable in this invocation context;
 	 * immutable and never {@code null}
 	 */
-	List<Object> getArguments();
+	List<@Nullable Object> getArguments();
 
 	/**
 	 * Get the target object of this invocation context, if available.
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ReflectiveInvocationContext.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ReflectiveInvocationContext.java
@@ -55,7 +55,7 @@ public interface ReflectiveInvocationContext<T extends Executable> {
 	 * Get the arguments of the executable in this invocation context.
 	 *
 	 * @return the arguments of the executable in this invocation context;
-	 * immutable and never {@code null}
+	 * immutable and never {@code null} but may contain {@code null}.
 	 */
 	List<@Nullable Object> getArguments();
 
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/InterceptingExecutableInvoker.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/InterceptingExecutableInvoker.java
@@ -16,7 +16,6 @@ import static org.junit.jupiter.engine.execution.ParameterResolutionUtils.resolv
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Executable;
 import java.lang.reflect.Method;
-import java.util.Optional;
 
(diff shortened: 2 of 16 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ReflectiveInvocationContext.java
```diff
@@ -56,7 +57,7 @@ public interface ReflectiveInvocationContext<T extends Executable> {
 	 * @return the arguments of the executable in this invocation context;
 	 * immutable and never {@code null}
 	 */
-	List<Object> getArguments();
+	List<@Nullable Object> getArguments();
```
mpkorstanje, line 60: This nullable annotation was missing. A method may be invoked with `null` as on of it's arguments.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/InterceptingExecutableInvoker.java
```diff
@@ -89,12 +88,9 @@ public void invokeVoid(Method method, @Nullable Object target, ExtensionContext
 			ExtensionContext extensionContext, ExtensionRegistry extensionRegistry,
 			ReflectiveInterceptorCall<Method, T> interceptorCall) {
 
-		@SuppressWarnings({ "unchecked", "rawtypes" })
-		Optional<Object> optionalTarget = (target instanceof Optional optional ? optional
-				: Optional.ofNullable(target));
```
mpkorstanje, line 94: I couldn't figure out what this is for. It looks like at some point `invoke` could be called with an optional target. I suppose to make it easier to use in combination with `ExtensionContext::getTestInstance`. But that idea never was completed. So I've just removed it.

sbrannen: Well, if it's no longer used, removing that sounds good.

Please update the `@param target` Javadoc as well.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/MethodInvocation.java
```diff
@@ -25,23 +25,26 @@
...
 		this.arguments = arguments;
 	}
 
 	@Override
 	public Class<?> getTargetClass() {
-		return this.target.<Class<?>> map(Object::getClass).orElseGet(this.method::getDeclaringClass);
+		if (target != null) {
+			return this.target.getClass();
+		}
+		return this.method.getDeclaringClass();
```
sbrannen: We often use a ternary operator for situations like that.

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/MethodInvocation.java
```diff
@@ -50,15 +53,15 @@ public Method getExecutable() {
 	}
 
 	@Override
-	public List<Object> getArguments() {
-		return unmodifiableList(Arrays.asList(this.arguments));
+	public List<@Nullable Object> getArguments() {
+		List<@Nullable Object> list = Arrays.asList(this.arguments);
+		return unmodifiableList(list);
```
sbrannen, line 55: Did you have to extract the `list` in order to appease NullAway?

sbrannen (approved):

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/MethodInvocation.java
mpkorstanje: Yes.

```
> Task :junit-jupiter-engine:compileJava
/home/mpkorstanje/Projects/github/junit-team/junit5/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/execution/MethodInvocation.java:54: warning: [NullAway] incompatible types: List<Object> cannot be converted to List<@Nullable Object>
                return unmodifiableList(Arrays.asList(this.arguments));
```

sormuras:
> An interesting aspect of this API is that we open taken in nullable values, wrap them in an `Optional`, pass them down the call chain and then unwrap them later on.

So, better store `null`s directly (in fields)?

mpkorstanje:
> So, better store nulls directly (in fields)?

Should have used past tense.: 😛

PR #5708, opened 2026-05-28
Document inheritance support

swatichauhan814:
Issue: https://github.com/junit-team/junit-framework/issues/938

### DIFF
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
@@ -30,6 +30,8 @@ import org.opentest4j.AssertionFailedError;
  * <p>Using this builder ensures consistency in how failure message are formatted
  * within JUnit Jupiter and for custom user-defined assertions.
  *
+ * <p>This class is not intended to be extended by clients.
+ *
  * @since 5.9
  * @see AssertionFailedError
  */
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
@@ -30,7 +30,10 @@ import org.opentest4j.AssertionFailedError;
  * <p>Using this builder ensures consistency in how failure message are formatted
  * within JUnit Jupiter and for custom user-defined assertions.
  *
- * <p>This class is not intended to be extended by clients.
+ * <h2>Extensibility</h2>
+ *
+ * <p>Although it is technically possible to extend this class, extension is
+ * strongly discouraged.
  *
  * @since 5.9
  * @see AssertionFailedError
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
@@ -22,6 +22,8 @@ import org.junit.platform.commons.JUnitException;
  * {@link ArgumentsAccessor} if an error occurs while accessing
  * or converting an argument.
  *
+ * <p>This class is not intended to be extended by clients.
+ *
  * @since 5.2
  * @see ArgumentsAccessor
  */
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
@@ -22,7 +22,9 @@ import org.junit.platform.commons.JUnitException;
  * {@link ArgumentsAccessor} if an error occurs while accessing
  * or converting an argument.
  *
- * <p>This class is not intended to be extended by clients.
+ * <h2>Extensibility</h2>
+ *
+ * <p>This class is designed for extension.
  *
  * @since 5.2
  * @see ArgumentsAccessor
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
@@ -24,7 +24,8 @@ import org.junit.platform.commons.JUnitException;
  *
  * <h2>Extensibility</h2>
  *
- * <p>This class is designed for extension.
+ * <p>Although it is technically possible to extend this class, extension is
+ * strongly discouraged.
  *
  * @since 5.2
  * @see ArgumentsAccessor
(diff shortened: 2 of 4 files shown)


mpkorstanje:
> I'd be very happy to review your work!

@jbduncan all yours.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertionFailureBuilder.java
```diff
@@ -30,6 +30,8 @@
  * <p>Using this builder ensures consistency in how failure message are formatted
  * within JUnit Jupiter and for custom user-defined assertions.
  *
+ * <p>This class is not intended to be extended by clients.
```
jbduncan: Looks good. I think it would be even better with wording and a header like this (inspired by `Assertions.java`) to make things more prominent:

```
 * <h2>Extensibility</h2>
 *
 * <p>Although it is technically possible to extend this class, extension is
 * strongly discouraged.
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
```diff
@@ -22,6 +22,8 @@
  * {@link ArgumentsAccessor} if an error occurs while accessing
  * or converting an argument.
  *
+ * <p>This class is not intended to be extended by clients.
```
jbduncan: Exceptions are a bit unusual. The base `JUnitException` is definitely designed for extension (as is `RuntimeException` for that matter), and I think it's fair to say that the JUnit maintainers may want to extend `ArgumentAccessException` in the future too, so we should give users the same right.

How about a header like this (inspired by `Assertions.java`) to make the wording on extension more prominent?

```
 * <h2>Extensibility</h2>
 *
 * <p>This class is designed for extension.
```

@mpkorstanje, do you have any thoughts on all this?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentsAggregationException.java
```diff
@@ -22,6 +22,8 @@
  * {@link ArgumentsAggregator} when an error occurs while aggregating
  * arguments.
  *
+ * <p>This class is not intended to be extended by clients.
```
jbduncan: Like my other comment on `ArgumentAccessException`, I think we should allow users to extend this class and we should use wording like this: 

```
 * <h2>Extensibility</h2>
 *
 * <p>This class is designed for extension.
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/AnnotationBasedArgumentsProvider.java
```diff
@@ -30,6 +30,8 @@
  * {@link ArgumentsProvider} implementations that also need to consume an
  * annotation in order to provide the arguments.
  *
+ * <p>This class is designed for extension.
```
jbduncan, line 35: Looks good. Like my other comments, I think it would be even better with a header like this: 

```
 * <h2>Extensibility</h2>
 *
 * <p>This class is designed for extension.
```

jbduncan (commented):
This is a great start! I have some thoughts on how it could be improved further; see below.

@mpkorstanje, I was wondering if I could have your input on these too as a JUnit maintainer?

jbduncan (commented):
@swatichauhan814 LGTM! Consider it approved on my end. Great work. 👍

I believe we need a review from a JUnit maintainer to get this merged in, so now we wait.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
mpkorstanje: > I think it's fair to say that the JUnit maintainers may want to extend ArgumentAccessException in the future too, so we should give users the same right.

As far as I can tell, the `ArgumentAccessException` shouldn't ever be created from user code. So this would have to be the "This class is not intended to be extended by clients." warning. But I'm not 100% sure about that.

In the user guide I see that users might implement the `ArgumentsAccessor` interface which would require them to be able to throw `ArgumentAccessException`. 

https://docs.junit.org/6.1.0/writing-tests/parameterized-classes-and-tests.html#argument-aggregation

> In this context, an indexed parameter is an argument for a given index in the [Arguments](https://docs.junit.org/6.1.0/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/Arguments.html) provided by an [ArgumentsProvider](https://docs.junit.org/6.1.0/api/org.junit.jupiter.params/org/junit/jupiter/params/provider/ArgumentsProvider.html) that is injected into a field annotated with `@Parameter(index)`. An aggregator is any `@Parameter`-annotated field of type [ArgumentsAccessor](https://docs.junit.org/6.1.0/api/org.junit.jupiter.params/org/junit/jupiter/params/aggregator/ArgumentsAccessor.html) or any field annotated with [@AggregateWith](https://docs.junit.org/6.1.0/api/org.junit.jupiter.params/org/junit/jupiter/params/aggregator/AggregateWith.html).

But I'm not sure if the user guide is correct here, I couldn't find a code path that would make that possible and I think that instead of `ArgumentsAccessor` the guide should omit the type and say:

> An aggregator is any `@Parameter`-annotated field or any field annotated with @AggregateWith.

 @jbduncan you could help me out here by double checking if this indeed the case.

mpkorstanje (changes requested):
Looks good! 

@jbduncan one remark for you to look into.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/ArgumentAccessException.java
jbduncan: @mpkorstanje I think you're right. From my testing with JUnit 6.0.3 on a personal project, it looks like `ArgumentsAccessor`-typed fields annotated with `@Parameter` aren't allowed. So the wording in the user guide can be fixed separately from this PR.

Furthermore, I can't see why users would want to implement `ArgumentsAccessor` if its Javadoc already says that it "is not intended to be implemented by clients" anyway.

Therefore, you're right, users should have no reason to subclass `ArgumentAccessException`, never mind instantiate it.

jbduncan: @swatichauhan814 So, from my discussion with @mpkorstanje just now, it turns out I was wrong; `ArgumentAccessException` is _not_ intended for extension after all. So could you change the wording to this, please? Sorry for misleading you.

```
 * <h2>Extensibility</h2>
 *
 * <p>Although it is technically possible to extend this class, extension is
 * strongly discouraged.
```

jbduncan:
> Looks good!
> 
> @jbduncan one remark for you to look into.

Thanks for the ping @mpkorstanje! I've just replied to your remark.

@swatichauhan814 As soon as [my comment](https://github.com/junit-team/junit-framework/pull/5708#discussion_r3368240145) is addressed, this PR looks good to go from my end!

jbduncan (commented):
LGTM! :+1:

mpkorstanje (approved):

mpkorstanje (changes requested):
@swatichauhan814 quick check, it looks like you're using multiple emails in b9cd9a1754e7761dcb5c931f837c355623d273ae. They'll all end up in the co-author tags. Is that intentional?

You can run `git log` on the command line on your branch to see exactly what is being committed.

mpkorstanje (approved):

PR #5719, opened 2026-06-01
Allow system properties abstraction

TomaszGaweda:
This PR allows building abstractions very similar to SystemPropertiesExtension.

The use case in my case is related to Hazelcast (#hazelcast/hazelcast), where we have our class wrapper for ClusterProperties - `ClusterProperty`. Cluster properties can be set via system properties.

It's common to see then patterns like `System.setProperty(SOME_PROPERTY.name(), "value")` in the code. We are now slowly migrating to JUnit 5 and it would be awesome if we could use `@SetSystemProperty` - however, Java does not allow us to call `.name()` method in annotation.

With this PR merged we will be able to make similar annotations like:
```java
@SetClusterProperty(LOGGING_TYPE, "log4j2")
```

Obviously it is possible to make such extensions right now, however it will duplicate existing extension - I'd even say, only changes will be other annotation names. I think there will be more places where libraries would like to have derivatives of the SetSystemProperty & Co. annotations.

Notes to reviewers:
- `AbstractSystemPropertiesExtension` is mostly a copy of old `SystemPropertiesExtension`
- new `SystemPropertiesExtension` implements `AbstractSystemPropertiesExtension` with current and standard set of annotations
- There is a test added `SystemPropertiesExtensionViaCustomAnnotationTests`, which is also mostly copied from `SystemPropertiesExtensionTests`, testing new abstraction and showing the usage.
- Please let me know if documentation added is enough. I didn't write too much as it's only a technical abstraction, that I think everybody will see quickly how it works, but I am not sure.

---

---

### Definition of Done

### DIFF
--- a/documentation/antora.yml
+++ b/documentation/antora.yml
@@ -175,6 +175,7 @@ asciidoc:
     RestoreSystemProperties: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/util/RestoreSystemProperties.html[@RestoreSystemProperties]'
     ReadsSystemProperty: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/util/ReadsSystemProperty.html[@ReadsSystemProperty]'
     WritesSystemProperty: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/util/WritesSystemProperty.html[@WritesSystemProperty]'
+    AbstractSystemPropertiesExtension: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/util/AbstractSystemPropertiesExtension.html[@AbstractSystemPropertiesExtension]'
     # Jupiter Parallel API
     Execution: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/parallel/Execution.html[@Execution]'
     Isolated: '{javadoc-root}/org.junit.jupiter.api/org/junit/jupiter/api/parallel/Isolated.html[@Isolated]'
--- a/documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
+++ b/documentation/modules/ROOT/pages/writing-tests/built-in-extensions.adoc
@@ -905,3 +905,16 @@ need to be annotated with the respective annotation:
 
 Tests annotated in this way will never execute in parallel with tests annotated with
 `@ClearSystemProperty`, `@SetSystemProperty`, or `@RestoreSystemProperties`.
+
+
+[[system-properties-extensibility]]
+=== System Property extensibility
+
+If you want to have customized annotations, that will for example limit keys to some enum values, you can use
+{AbstractSystemPropertiesExtension} as a base of your extension. With this abstraction you need to provide 3 class types:
+
+ - annotation used to set system property
+ - annotation used to clear system property
+ - annotation used to restore system properties.
+
+Annotations provided by those functions should be annotated with {WritesSystemProperty}.
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/AbstractSystemPropertiesExtension.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/util/AbstractSystemPropertiesExtension.java
@@ -1,0 +1,344 @@
+/*
+ * Copyright 2015-2026 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.api.util;
+
+import static java.util.stream.Collectors.toSet;
+import static org.junit.platform.commons.support.AnnotationSupport.findRepeatableAnnotations;
+import static org.junit.platform.commons.support.AnnotationSupport.isAnnotated;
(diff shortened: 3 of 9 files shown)


TomaszGaweda:
Thanks @mpkorstanje for adding the labels and forwarding to the Team. Please let me know when you will need changes to the PR, questions answered or anything else. Have a great day!

mpkorstanje:
To clarify, because I only got this after looking at your test code.

> It's common to see then patterns like System.setProperty(SOME_PROPERTY.name(), "value") in the code. We are now slowly migrating to JUnit 5 and it would be awesome if we could use @SetSystemProperty - however, Java does not allow us to call .name() method in annotation.

So you've defined [`ClusterProperty`](https://github.com/hazelcast/hazelcast/blob/d4dbe90ea646f3bfca462d8d7128a55ef082594d/hazelcast/src/main/java/com/hazelcast/spi/properties/ClusterProperty.java#L613-L618) like this:

```java
    /**
     * The interval at which member heartbeat messages are sent
     */
    public static final HazelcastProperty HEARTBEAT_INTERVAL_SECONDS
            = new HazelcastProperty("hazelcast.heartbeat.interval.seconds", 5, SECONDS);
```

And currently you'd write a [SystemClockChangeTest](https://github.com/hazelcast/hazelcast/blob/d4dbe90ea646f3bfca462d8d7128a55ef082594d/hazelcast/src/test/java/com/hazelcast/cluster/SystemClockChangeTest.java#L38-L42) like this:


```java
    @BeforeClass
    public static void setSystemProps() {
        System.setProperty(ClusterProperty.HEARTBEAT_INTERVAL_SECONDS.getName(), "1");
        System.setProperty(ClusterProperty.MAX_NO_HEARTBEAT_SECONDS.getName(), "10");
    }
```

And you can't use the system properties extension because `@SetSystemProperty(key = ClusterProperty.HEARTBEAT_INTERVAL_SECONDS.getName(), "1")` isn't valid java.

mpkorstanje:
> Obviously it is possible to make such extensions right now, however it will duplicate existing extension - I'd even say, only changes will be other annotation names. I think there will be more places where libraries would like to have derivatives of the SetSystemProperty & Co. annotations.

I don't quite see what your extension would look like. You wouldn't be able to use write 
`@SetClusterProperty(key = ClusterProperty.HEARTBEAT_INTERVAL_SECONDS, "1")` because `ClusterProperty` isn't an enum.

If you are going to refactor that, have you instead considered extracting the property names as constants? It is a fairly common pattern.

```java
    public static final String NETWORK_STATS_REFRESH_INTERVAL_SECONDS_PROPERTY_NAME = "hazelcast.network.stats.refresh.interval.seconds";
    public static final HazelcastProperty NETWORK_STATS_REFRESH_INTERVAL_SECONDS
            = new HazelcastProperty(NETWORK_STATS_REFRESH_INTERVAL_SECONDS_PROPERTY_NAME, 3, SECONDS);
```

TomaszGaweda:
Hi @mpkorstanje!

The key can be any object - you can even use int, that will be then translated to property with switch case (although I don't think anybody would use it this way, just an example :) ).

> And you can't use the system properties extension because @SetSystemProperty(key = ClusterProperty.HEARTBEAT_INTERVAL_SECONDS.getName(), "1") isn't valid java.

Exactly

> I don't quite see what your extension would look like. You wouldn't be able to use write
@SetClusterProperty(key = ClusterProperty.HEARTBEAT_INTERVAL_SECONDS, "1")

The key could be any value, that would be translated into string by `setAnnotationKey` function. Since `ClusterProperty.HEARTBEAT_INTERVAL_SECONDS` is a constant, it can be used, `setAnnotationKey` will do `property.getName()`.

> If you are going to refactor that, have you instead considered extracting the property names as constants? It is a fairly common pattern.

That's some solution, but also lacks some of protections:
- key is still any String. That's ok in general, but if you want to limit what can be set, it's not possible now.
- this won't handle e.g. default values (custom annotation can have `value` as optional parameter)

mpkorstanje:
> The key could be any value, that would be translated into string by `setAnnotationKey` function. Since `ClusterProperty.HEARTBEAT_INTERVAL_SECONDS` is a constant, it can be used, `setAnnotationKey` will do `property.getName()`.

As per https://docs.oracle.com/javase/specs/jls/se17/html/jls-9.html#jls-9.6.1 annotation members can only be primitives, strings, ennums, annotation types or an array of any of these. You would not be able to write `@SetClusterProperty(key = ClusterProperty.HEARTBEAT_INTERVAL_SECONDS, "1")` because `ClusterProperty.HEARTBEAT_INTERVAL_SECONDS` is an object and not one of the allowed member types. So as I understand it right now, your proposed use case is not possible.

Furthermore, I think your proposal would benefit from a more concrete example. I would suggest that you implement your idea in Hazelcast first. It will remove any ambiguity as to what you are proposing and provide us with a concrete and working example to look at. At the same time you won't have to wait for us to get through our backlog.

After creating a concrete example, you can re-propose this by creating an issue. Unlike a PR it will lets us discuss different solutions to a usecase without also having to discuss a specific implementation.

TomaszGaweda:
> As per [docs.oracle.com/javase/specs/jls/se17/html/jls-9.html#jls-9.6.1](https://docs.oracle.com/javase/specs/jls/se17/html/jls-9.html#jls-9.6.1) annotation members can only be primitives, strings, ennums, annotation types or an array of any of these.

That's indeed my mistake, sorry! 

> Furthermore, I think your proposal would benefit from a more concrete example. I would suggest that you implement your idea in Hazelcast first

Will do. I made a small PoC, but didn't notice it's enum and won't work for target object. My bad. Thanks for correction, closing PR for now.

PR #5720, opened 2026-06-01
Restore original SetSystemProperty values in a ParameterizedTest

### LINKED ISSUE #5717: Using `@SetSystemProperty` in a `@ParameterizedTest` fails to restore original values
## Overview

Important: I see this as the EXACT same problem as https://github.com/junit-pioneer/junit-pioneer/issues/875

The problem is that when using `@SetSystemProperty` on a `@ParameterizedTest` then after that test has completed the properties are not reset to their original state. The effect is that these changes pollute the environment any later test runs in.

The mentioned https://github.com/junit-pioneer/junit-pioneer/issues/875 is the exact same problem, but there it is the `@SetEnvironmentVariable` (which uses the exact same code to handle these annotations).

## Steps to reproduce

After having run a test like this; the system property set via `@SetSystemProperty` will NOT have its original value in the tests that are run after this test. It will still have the `new value` from this test.

```java
@ParameterizedTest
@ValueSource(ints = { 1 })
@SetSystemProperty(key = "foo", value = "new value")
void testParamWithoutRestore(int dummy) {
   ...
}
```

I wrote some tests to reproduce this problem: https://github.com/junit-team/junit-framework/pull/5718

This is essentially the same set of tests I provided for the JUnit Pioneer project: https://github.com/junit-pioneer/junit-pioneer/pull/876

## Context

- Used versions (Jupiter/Vintage/Platform): 6.1.0
- Build Tool/IDE: IntelliJ

mpkorstanje:
Restore original SetSystemProperty values in a ParameterizedTest

The system properties extension would for each context and all its
ancestors process the system property annotations. After applying each
context it would also create an incremental backup. But it would only
store the latest backup.

This is a problem when parametrized tests are used. The ancestral
context includes the test method element twice:

```
0 = {Optional@5653} "Optional[void RestoreAfterParameterizedTest.testParamWithoutRestore(int)]"
1 = {Optional@5654} "Optional[void RestoreAfterParameterizedTest.testParamWithoutRestore(int)]"
2 = {Optional@5655} "Optional[class SystemPropertiesExtensionTests$RestoreAfterParameterizedTest]"
3 = {Optional@5656} "Optional[class SystemPropertiesExtensionTests]"
4 = {Optional@4689} "Optional.empty"
 ``` 

As such the system  property will be backed up, modified, the modified
value will be backed up again and then modified (idempotent). Because
only the last incremental backup is stored, the original value is
never restored.
 
This could be solved by either:
 
 a) Only processing the original context and none of its ancestors.
 b) Using a compound key of the original context and its current ancestor.
 
Option a mostly works, but will not properly restore programmatically
modified keys when nested tests are sparsely annotated. So it must be
option b.

Fixes: #5717
Supersedes: #5718

---

---

### Definition of Done

### DIFF
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.1.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.1.1.adoc
@@ -35,7 +35,7 @@ repository on GitHub.
 [[v6.1.1-junit-jupiter-bug-fixes]]
 ==== Bug Fixes
 
-* ❓
+* Restore original values set by `@SetSystemProperty` when used with a `@ParameterizedTest`.
 
 [[v6.1.1-junit-jupiter-deprecations-and-breaking-changes]]
 ==== Deprecations and Breaking Changes
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
@@ -41,7 +41,7 @@ repository on GitHub.
 [[v6.2.0-M1-junit-jupiter-bug-fixes]]
 ==== Bug Fixes
 
-* Restore original values set by `@SetSystemProperty` when used with a `@ParameterizedTest`.
+* ❓
 
 [[v6.2.0-M1-junit-jupiter-deprecations-and-breaking-changes]]
 ==== Deprecations and Breaking Changes
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
@@ -38,7 +38,7 @@ repository on GitHub.
 [[v6.2.0-M1-junit-jupiter-bug-fixes]]
 ==== Bug Fixes
 
-* ❓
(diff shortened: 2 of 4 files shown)


marcphilipp:
> Using a compound key of the original context and its current ancestor

I'm proposing a slightly different solution that avoids the compound key by storing the incremental backups in their corresponding context:

- #5744 

WDYT?

mpkorstanje:
> I'm proposing a slightly different solution that avoids the compound key by storing the incremental backups in their corresponding context:

This does look more elegant. I kinda like it.

* It does leave garbage behind though. When the current context goes out of scope, the ancestral contexts retain information about it.
* I imagine it will be harder to debug because you can't just inspect the current context for information.

I think the first item we can come up with some solutions, but then we're trading complexity for complexity. So probably not worth it.

More fundamentally though, the need for composite keys or storing in different context comes from using `findAllExtensionContexts`. And  I don't quite understand why we are using `findAllExtensionContexts` when then extension also uses before and after all callbacks. Looking at the history in JUnit Pioneer, it seems they started with just before and after each call backs and then later the all variants.

marcphilipp:
> It does leave garbage behind though

That's a good point! Instead of creating _n_ incremental backups, we could create a single merged one, though, couldn't we?

mpkorstanje:
We could create single operation. Instead of applying the operations the system properties, they could be applied to map. And then only at the end apply the final operation and create the backup.

marcphilipp (approved):

PR #5741, opened 2026-06-07
Close Socket in testPlanExecutionStarted

tejas-ae:
While tracing testPlanExecutionStarted in junit-platform-reporting/src/main/java/org/junit/platform/reporting/open/xml/OpenTestReportGeneratingListener.java I noticed what might be an issue: The resource opened there can leak if testPlanExecutionStarted exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path. Let me know if the approach doesn’t fit.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


marcphilipp:
This change would close the socket too early. It needs to stay open while the test plan is being executed. It's already closed when its `OutputStream` is closed in `testPlanExecutionFinished` which is guaranteed to be called by the JUnit Platform except in extreme cases.

PR #5765, opened 2026-06-13
Document inheritance support - part 2

swatichauhan814:
Issue: https://github.com/junit-team/junit-framework/issues/938


junit-jupiter-api/src/main/java/org/junit/jupiter/api/DisplayNameGenerator.java
```diff
@@ -44,6 +44,10 @@
  * <p>Display names are typically used for test reporting in IDEs and build
  * tools and may contain spaces, special characters, and even emoji.
  *
+ * <h2>Extensibility</h2>
+ *
+ * <p>This interface is designed to be implemented by clients.
+ *
  * <p>Concrete implementations must have a <em>default constructor</em>.
  *
  * <p>A {@link DisplayNameGenerator} can be configured <em>globally</em> for the
```
jbduncan, line 55: It's worth giving this paragraph its own header so that users don't confuse configuration for extension. Something like this: 

```
 * <h2>Configuration</h2>
 * 
 * <p>A {@link DisplayNameGenerator} can be configured <em>globally</em> for the
 ...
```

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/SimpleArgumentsAggregator.java
```diff
@@ -23,6 +23,10 @@
  * {@link ArgumentsAggregator} implementations that do not need to distinguish
  * between fields and method/constructor parameters.
  *
+ * <h2>Extensibility</h2>
+ *
+ * <p>This class is designed for extension.
+ *
```
jbduncan, line 29: The public `aggregateArguments` methods in this class have to [document](https://github.com/junit-team/junit-framework/issues/938#issue-242019121) that they call the protected abstract `aggregateArguments` method, as that method is technically overridable.

jbduncan, line 29: Also, let's add this extension wording to the class Javadoc of ArgumentsAggregator too, just so that the whole type hierarchy is consistent.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/HierarchicalTestEngine.java
jbduncan, line 60: This method should document that it uses the overridable method `createThrowableCollectorFactory`. Something like this: 

```
	 * Create an {@linkplain #createExecutorService(ExecutionRequest) executor
	 * service}; create an initial {@linkplain #createExecutionContext execution
	 * context}; execute the behavior of all {@linkplain Node nodes} in the
	 * hierarchy starting with the supplied {@code request}'s
	 * {@linkplain ExecutionRequest#getRootTestDescriptor() root}, collecting every thrown throwable using the {@linkplain org.junit.platform.engine.support.hierarchical.ThrowableCollector throwable collector} returned by {@linkplain #createThrowableCollectorFactory}; and notify
	 * its {@linkplain ExecutionRequest#getEngineExecutionListener() execution
	 * listener} of test execution events.
```

...which will need some reformatting with `./gradlew spotlessApply`.

jbduncan: But I'd like to run this past @mpkorstanje first, as I can't tell on a code search on my phone how `ThrowableCollector` is being used exactly and how the collected throwables are being reported, so it may influence the wording used in the Javadoc here.

@mpkorstanje, what do you think?

jbduncan (commented):
@swatichauhan814 Fantastic start! 👍 

I've added some suggestions, including one which will need a JUnit maintainer to check, but once everything is addressed then it should all be good.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/HierarchicalTestEngine.java
mpkorstanje: The purpose of the `ThrowableCollector` is to collect exceptions that happen in a hierarchical context and convert those, if any, into a `TestExecutionResult`. The use-case for extension is to make it possible to customize what exceptions contitute an aborted execution (e.g. from `Assumptions.fail()`.

jbduncan: Okay, so it sounds like my proposed wording is probably fine. The user can search up exactly what `createThrowableCollectorFactory` does by default by reading its Javadoc and looking up `OpenTest4JAwareThrowableCollector`. Thoughts?

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/SimpleArgumentsAggregator.java
```diff
@@ -36,12 +36,18 @@ public abstract class SimpleArgumentsAggregator implements ArgumentsAggregator {
 	public SimpleArgumentsAggregator() {
 	}
 
+	/**
+	 * Delegates to {@link #aggregateArguments(ArgumentsAccessor, Class, AnnotatedElementContext, int)}.
+	 */
```
jbduncan, line 41: Minor nitpick: Let's use the wording "By default, this method delegates to...".

But I won't stop this PR from being merged in as-is, this wording is pretty clear already! Thank you for doing this. 👍

junit-jupiter-params/src/main/java/org/junit/jupiter/params/aggregator/SimpleArgumentsAggregator.java
```diff
@@ -36,12 +36,18 @@ public abstract class SimpleArgumentsAggregator implements ArgumentsAggregator {
...
 	@Override
 	public @Nullable Object aggregateArguments(ArgumentsAccessor accessor, ParameterContext context)
 			throws ArgumentsAggregationException {
 		return aggregateArguments(accessor, context.getParameter().getType(), context, context.getIndex());
 	}
 
+	/**
+	 * Delegates to {@link #aggregateArguments(ArgumentsAccessor, Class, AnnotatedElementContext, int)}.
+	 */
```
jbduncan, line 50: Ditto.

jbduncan (commented):
@swatichauhan814 LGTM! If you have the time to address a minor nitpick (see below), then great, but if not then no worries, this is looking really good. Well done! 👍

I'll leave it to @mpkorstanje to give the final approval.

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/HierarchicalTestEngine.java
mpkorstanje: Yeah. Sounds about right. The `ThrowableCollector` could be better to but that is out scope for this PR.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ConditionEvaluationResult.java
```diff
@@ -22,6 +22,11 @@
 /**
  * The result of evaluating an {@link ExecutionCondition}.
  *
+ * <h2>Extensibility</h2>
+ *
+ * <p>Although it is technically possible to extend this class, extension is
+ * strongly discouraged.
```
mpkorstanje: @jbduncan in what category does classes fall that are public, but have a private constructor?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
```diff
@@ -26,6 +26,11 @@
  * {@code DynamicNode} serves as the abstract base class for a container or a
  * test case generated at runtime.
  *
+ * <h2>Extensibility</h2>
+ *
+ * <p>Although it is technically possible to extend this class, extension is
+ * strongly discouraged.
```
mpkorstanje: @jbduncan in what category does classes fall that are public, but have a package private constructor?

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/LoggingListener.java
```diff
@@ -30,6 +30,11 @@
  * for all events via a {@link BiConsumer} that consumes {@code Throwable}
  * and {@code Supplier<String>}.
  *
+ * <h2>Extensibility</h2>
+ *
+ * <p>Although it is technically possible to extend this class, extension is
+ * strongly discouraged.
```
mpkorstanje: @jbduncan private constructor?

mpkorstanje (changes requested):
Looks good, but looks like we caught some classes with private and package private constructors. I don't think those are extensible in the first place.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ConditionEvaluationResult.java
jbduncan: @mpkorstanje Good catch! In this case, it can't be subclassed precisely because the constructor is private. So it might as well be made `final` instead of having this wording added to the Javadoc, as implied by https://github.com/junit-team/junit-framework/issues/938#issue-242019121:

> **Prohibit subclassing** by making the class final or replacing all public constructors with public static factory methods.

@swatichauhan814 Would you mind removing this Javadoc wording and making the class `final`, please?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/DynamicNode.java
jbduncan: Similarly to my other comment, it can't be subclassed because the constructor isn't public. So we can make the class `final` instead as implied by https://github.com/junit-team/junit-framework/issues/938#issue-242019121:

> **Prohibit subclassing** by making the class final or replacing all public constructors with public static factory methods.

@swatichauhan814 Would you mind removing this Javadoc wording, please? We would make it `final`, but it's an abstract class which means it's impossible to do so. However, because the constructor is package-private, it means the class can't be extended by JUnit users anyway, so it's fine that we can't make it `final`. :)

junit-platform-launcher/src/main/java/org/junit/platform/launcher/listeners/LoggingListener.java
jbduncan: Good catch again!

@swatichauhan814 Let's remove this Javadoc wording and make the class `final`, please.

jbduncan (commented):
@swatichauhan814 Sorry for the delay, I've submitted my latest review in the comments with @mpkorstanje above, let me know what you think. :)

jbduncan:
@swatichauhan814 LGTM! I think this is ready to be merged in. 👍

mpkorstanje (approved):
LGTM.

I've reverted a56956a0a0a95515062627a28fdb5627e125f8b6. While technically not a breaking change, people can't extend these classes in the first place, I also don't want to surprise anyone who's been mocking them.

It also keeps the scope of this PR limited to documentation which simplifies things and making these classes final can de deffered until v7 (https://github.com/junit-team/junit-framework/issues/5827).

PR #5776, opened 2026-06-17
Add action to mirror repo to Codeberg

marcphilipp:


.github/workflows/codeberg.yml
```diff
@@ -0,0 +1,22 @@
+name: Codeberg
+
+on: [ push ]
```
mpkorstanje, line 3: I think we should take care not to abuse codeberg. I don't see a good reason to push anything but main, release and website branches.

marcphilipp: I think having a full mirror in case GitHub could be useful in case GitHub goes down. I don't think this will cause a lot of traffic for Codeberg that would be qualified as abuse. Moreover, we've applied to become a member of Codeberg e.V. and are supporting them that way.

PR #5812, opened 2026-06-29
Validate argument max length configuration parameter

### LINKED ISSUE #5810: Non-positive `junit.jupiter.params.displayname.argument.maxlength` crashes all parameterized tests
When the documented configuration parameter `junit.jupiter.params.displayname.argument.maxlength` is set to a non-positive value (e.g. `0` or a negative number), every parameterized test in the run fails with an internal `StringIndexOutOfBoundsException` instead of a clear configuration error.

## Steps to reproduce

`src/test/resources/junit-platform.properties`:

```properties
junit.jupiter.params.displayname.argument.maxlength=0
```

```java
import org.junit.jupiter.params.ParameterizedTest;

import static org.junit.jupiter.api.Assertions.assertNotNull;

class MaxLengthDemoTests {

    @ParameterizedTest
    @ValueSource(strings = "x")
    void test(String s) {
        assertNotNull(s);
    }
}
```

**Expected:** the configuration parameter is rejected with a clear configuration error (or a non-positive value is handled gracefully). The user should never see an internal `StringIndexOutOfBoundsException`.

**Actual:** every invocation fails. `ParameterizedInvocationNameFormatter` throws, wrapped as:

```
org.junit.platform.commons.JUnitException: Failed to format display name for parameterized test
  Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 1
```

With `maxlength=-1` it is even worse: `argument.length() > -1` is `true` even for an empty string, so every argument (including `""`) triggers the failure.

### Root cause

`ParameterizedInvocationNameFormatter` reads the configuration parameter without any lower-bound validation, then truncates with `argument.substring(0, argumentMaxLength - 1)`. For a non-positive `argumentMaxLength`, `substring` is called with a negative end index and throws. (The sibling CSV attribute `maxCharsPerColumn` is validated, so this is also an inconsistency in input validation.)

## Context

 - Used versions (Jupiter/Vintage/Platform): 6.1.1 (and earlier); reproduced on `main`
 - Build Tool/IDE: any (Gradle/Maven/IDE)

## Deliverables

- [x] Validate `junit.jupiter.params.displayname.argument.maxlength` and fail fast with a clear configuration error when the value is less than `1`
- [x] Add a regression test covering a non-positive configured value

big-cir:
This validates the `junit.jupiter.params.displayname.argument.maxlength` configuration parameter and fails fast with a clear configuration error when the value is less than `1`.

Previously the value was read without any lower-bound validation. A non-positive value caused the display-name formatter to call `substring(0, argumentMaxLength - 1)` with a negative end index, throwing a `StringIndexOutOfBoundsException` that was wrapped as a `JUnitException` and failed every parameterized test in the run:

```
org.junit.platform.commons.JUnitException: Failed to format display name for parameterized test
  Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 1
```

The sibling CSV attribute `maxCharsPerColumn` is already validated, so this also aligns the input validation across the module.

Closes: #5810

---

---

### Definition of Done

- [x] Change is covered by automated tests including corner cases, errors, and exception handling
- [x] Method preconditions are checked and documented in the method's Javadoc — n/a; this validates a configuration parameter rather than adding a new public method
- [x] Public API has Javadoc and `@API` annotations — n/a; no public API was added or changed
- [x] Change is documented in the User Guide and Release Notes — the User Guide already documents the parameter; a Release Notes entry can be added if preferred

### DIFF
--- a/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedInvocationNameFormatter.java
+++ b/junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedInvocationNameFormatter.java
@@ -74,6 +74,8 @@ class ParameterizedInvocationNameFormatter {
 
 		int argumentMaxLength = extensionContext.getConfigurationParameter(ARGUMENT_MAX_LENGTH_KEY, Integer::parseInt) //
 				.orElse(512);
+		Preconditions.condition(argumentMaxLength > 0,
+			() -> ARGUMENT_MAX_LENGTH_KEY + " must be a positive number: " + argumentMaxLength);
 
 		return new ParameterizedInvocationNameFormatter(pattern, extensionContext.getDisplayName(), declarationContext,
 			argumentMaxLength);
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
@@ -39,7 +39,9 @@ repository on GitHub.
 [[v6.2.0-M1-junit-jupiter-bug-fixes]]
 ==== Bug Fixes
 
-* ❓
+* The `junit.jupiter.params.displayname.argument.maxlength` configuration parameter is
+  now validated and rejected with a clear error when set to a non-positive value instead
+  of causing parameterized tests to fail with an internal exception.
 
 [[v6.2.0-M1-junit-jupiter-deprecations-and-breaking-changes]]
 ==== Deprecations and Breaking Changes
--- a/jupiter-tests/src/test/java/org/junit/jupiter/params/ParameterizedTestIntegrationTests.java
+++ b/jupiter-tests/src/test/java/org/junit/jupiter/params/ParameterizedTestIntegrationTests.java
@@ -492,8 +492,8 @@ class ParameterizedTestIntegrationTests extends AbstractJupiterTestEngineTests {
 				.selectors(selectMethod(TestCase.class, "testWithCsvSource", String.class.getName())) //
 				.execute();
 		results.allEvents().assertThatEvents() //
-				.haveExactly(1, event(container(), finishedWithFailure(message(msg -> msg.contains(
-					ParameterizedInvocationNameFormatter.ARGUMENT_MAX_LENGTH_KEY + " must be a positive number")))));
+				.haveExactly(1, event(container(), finishedWithFailure(message(
+					ParameterizedInvocationNameFormatter.ARGUMENT_MAX_LENGTH_KEY + " must be a positive number: 0"))));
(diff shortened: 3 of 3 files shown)


junit-jupiter-params/src/main/java/org/junit/jupiter/params/ParameterizedInvocationNameFormatter.java
```diff
@@ -74,6 +74,9 @@ static ParameterizedInvocationNameFormatter create(ExtensionContext extensionCon
 
 		int argumentMaxLength = extensionContext.getConfigurationParameter(ARGUMENT_MAX_LENGTH_KEY, Integer::parseInt) //
 				.orElse(512);
+		Preconditions.condition(argumentMaxLength >= 1,
+			() -> "Configuration parameter '%s' must be greater than or equal to 1 but was %d.".formatted(
```
mpkorstanje: Similar checks elsewhere in the code base use `arg > 0` and say `arg must be a positive number`. It would be good to keep this language consistent.

jupiter-tests/src/test/java/org/junit/jupiter/params/ParameterizedInvocationNameFormatterTests.java
```diff
@@ -275,6 +281,20 @@ void truncatesArgumentsThatExceedMaxLength() {
 		assertThat(formattedName).isEqualTo("\"fo\", \"foo\", \"fo…\"");
 	}
 
+	@Test
+	void throwsExceptionWhenArgumentMaxLengthIsNotPositive() {
+		ExtensionContext extensionContext = mock();
```
mpkorstanje: Instead of mocking, we usually test through an integration test. The test kit makes this relatively easy.

For example the other precondition from the `ParameterizedInvocationNameFormatter.create` is tested here.
https://github.com/junit-team/junit-framework/blob/e49e70cff390b08920bd15b9a942688601200fc3/jupiter-tests/src/test/java/org/junit/jupiter/params/ParameterizedTestIntegrationTests.java#L422 you can follow that pattern (and move the test)

mpkorstanje (changes requested):
Cheers! The current PR does things a bit different from now they're commonly done in the code base. It would be good to keep that consists.

mpkorstanje:
> Change is documented in the User Guide and Release Notes — the User Guide already documents the parameter; a Release Notes entry can be added if preferred

Please do add a release note.

big-cir:
Hi, mpkorstanje!

Thank you for the review! I have addressed all of your feedback and updated the PR accordingly. 
Could you please take another look when you have a chance?

Thanks!

mpkorstanje (approved):
LGTM.

PR #5816, opened 2026-06-30
Address EditorConfig violations

BaumiCoder:
This Pull Request is part of an academic research project. More preciously it is for my master's thesis, more background information [here](https://codeberg.org/BaumiCoder/Masterthesis).

## Definition: EditorConfig violations

The goal of this Pull Request is to address all [EditorConfig](https://editorconfig.org) violations in the project. With EditorConfig you have set for example `trim_trailing_whitespace = true` for most files, but there were files with trailing spaces. I call this an EditorConfig violation as a file content violates the respective EditorConfig property. They can have different kind of impacts. For example trailing spaces, which get remove with the next save, may make the next diff / Pull Request to this file less readable.

## How to address them?

To check for such violations and fix them I developed the CLI tool [ecformat](https://codeberg.org/BaumiCoder/ecformat). However, there are different ways to address the violations:

1. **Adjust EditorConfig:** The EditorConfig properties for a file can be wrong. For example many of your `.xml.snapshot` files have trailing spaces, but these have to be exactly the expected output. Therefore, it is wrong to remove them. The correct way to address that is to set `trim_trailing_whitespace = false` for them, which avoids automatic removal of them on save.
2. **Fix file content:** Another way is to fix the content of the file. Which means for trailing spaces to remove them.
   To do so `ecformat fix` can be helpful.
3. **Ignore files:** For some files it is not meaningful to consider the EditorConfig violations, which `ecformat check` finds. I added a `.ecformat_ignore` file for this. It has the common ignore syntax such as a `.gitignore`. To use such ignore files, provide the name of them with an CLI switch, `ecformat check --ignore-file .ecformat_ignore`. (The `.gitignore` files are automatically considered inside a git repository.) There are different reasons to ignore files:
   1. As EditorConfig is only for text files, considering binary files does not make sense.
   2. EditorConfig supports to set the property on the file level. However, there can be files with intentional differences inside. Such files need to be ignored fully by `ecformat` as it currently (version 0.2.0) has no support to ignore only a single property for some files. (I found no such cases in your project.)

## Changes in this Pull Request

For this Pull Request, I addressed all violations in the project in the way, which looks the most correct one to me. (I focused on the Properties of the current [EditorConfig specification](https://spec.editorconfig.org/#supported-pairs) 0.17.2.) If I fixed file contents, I manually reviewed all of them, making adjustments in the `ecformat fix` changes were necessary. Furthermore, `ecformat` detects the charset wrong for some files. Therefore, I would suggest disabling this property for whole project, with `--disable-charset` or `-c`. (I plan to improve the charset detection in a later version.)

If I picked the wrong way to address the violations for some files, 
please let me know with a respective Review comment.
Then I will address it in the correct way.
To support the Review of this Pull Request, I tried to group the changes in meaningful commits
with respective commit messages. Therefore, may take a look at the commit history.

## Avoid EditorConfig violations in the future

Such EditorConfig violations can creep in again. Possible reasons can be that the editor of some contributors has no EditorConfig support, or it would be required to install a respective plugin for the support.

If you want to avoid them in the future, I would suggest integrating a respective utility. For example as part of the CI. One possibility is to use `ecformat` with `ecformat check --disable-charset --ignore-file .ecformat_ignore`.
Another one is [editorconfig-checker](https://editorconfig-checker.github.io), which is a more sophisticated tool without the possibility to fix violations. It provides more configuration options, such as ignoring violations in single lines with marker comments.

Let me know if you are interested using one of them. I could add one in this Pull Request or a follow-up Pull Request.
If you not consider to use `ecformat` in your project, the `.ecformat_ignore` file is maybe not worth to merge into main. I can remove it if desired.

---

---

### Definition of Done

(Most of this bullet points may not be really applicable for this format changes.)


platform-tests/src/test/resources/do_not_delete_me.txt
```diff
@@ -1,3 +1,3 @@
 I am used by tests, so...
 
-Do NOT delete me!
```
BaumiCoder, line 3: Here I also added only a final newline. I don't know why GitHub does not show the "⛔" symbol as in the other files, that here was no final newline before.

.ecformat_ignore
mpkorstanje, line 1: This file isn't used by anything in the JUnit project, correct?

BaumiCoder: Yes, that is correct. It is only for using it with ecformat like this: `ecformat check --disable-charset --ignore-file .ecformat_ignore` as mentioned in the PR description.

.editorconfig
```diff
@@ -18,3 +18,15 @@ ij_kotlin_allow_trailing_comma_on_call_site = false
 
 [*.kts]
 indent_style = tab
+
+[*.xml.snapshot]
+# The expected XML files contain trailing spaces. Removing them let the tests fail.
+trim_trailing_whitespace = false
```
mpkorstanje, line 24: > The EditorConfig properties for a file can be wrong. For example many of your .xml.snapshot files have trailing spaces, but these have to be exactly the expected output. Therefore, it is wrong to remove them.

@marcphilipp why do we produce xml reports with trailing whitespace anyway?

.editorconfig
```diff
@@ -18,3 +18,15 @@ ij_kotlin_allow_trailing_comma_on_call_site = false
...
+
+[.idea/*.xml]
+# Config files from IntelliJ are without a final newline.
+insert_final_newline = false
+
+[documentation/modules/ROOT/images/extensions_StoreHierarchy.svg]
+# SVG files from draw.io are without a final newline
+insert_final_newline = false
```
mpkorstanje, line 32: @BaumiCoder, for these last two does anything bad happen if a new line is added?

.ecformat_ignore
mpkorstanje: Then please omit it from the PR.

mpkorstanje (commented):
Cheers! Looks good overall. I reckon we can merge this without `.ecformat_ignore`. 

> Let me know if you are interested using one of them. I could add one in this Pull Request or a follow-up Pull Request.
If you not consider to use ecformat in your project, the .ecformat_ignore file is maybe not worth to merge into main. I can remove it if desired.

We'll have to discuss that internally and I've labeled the PR as such.

.editorconfig
BaumiCoder: - The `.idea/*.xml` file is normally modified via saving the respective IntelliJ settings. This saves the file always without a final newline, tested with IntelliJ IDEA 2026.1.3. 
- This one SVG is from draw.io. (It contains a comment that it should only be modified with it.) I tested with the current online version of draw.io, and it saves SVG files always without a final newline.

Therefore, adding a final newline now or automatically on save in your editor would work against these behaviors. Next time, you are saving it with the respective software the new line would get lost again.

.ecformat_ignore
BaumiCoder: Done with c34f2aa8f6604fc4b7aa1f9ae77a74743c26dc95

.editorconfig
marcphilipp: We don't. These files are produced by the `snapshot-tests` library.

marcphilipp (approved):

mpkorstanje:
> Let me know if you are interested using one of them. I could add one in this Pull Request or a follow-up Pull Request.

@BaumiCoder thanks for offering.

We currently use Spotless for all our formatting needs. It integrates with Gradle so we can manage our build through a single tool. Spotless doesn't support all file formats so it is not the perfect solution to enforce editor config rules, but given that the bulk of the drift is in the yaml files we would consider formatting only those to be sufficient.

From the wording of your offer I get the impression that you are only interested in setting up either `ecformat` or `editorconfig-checker`. But would you perhaps be interested in configuring spotless to format yaml files instead (https://github.com/junit-team/junit-framework/issues/5822)?

BaumiCoder:
I understand the need to keep the set of tools small. I am not very familiar with Gradle, but I want to try it. You can assign me the issue #5822, and I will work on it in 1–2 weeks. There would be also a Gradle Plugin for EditorConfig. However, I never tried it and it is maybe broken see https://github.com/ec4j/editorconfig-gradle-plugin/issues/19.

mpkorstanje:
Cheers!

PR #5819, opened 2026-06-30
Add assertThrows variants which accept exceptions as values

mdedetrich:
Currently there exists `org.junit.jupiter.api.assertThrows` and its various variants to be used with JUnit and Kotlin that are also compatible with coroutines (i.e. `suspend func`). The current design uses reified inline functions, which generally work aside from one situation, specifically

1. You are using `@ParameterizedTest` that produces exceptions classes as values and you intend to use `assertThrows` on those exception values.
2. The function/method you are testing is a kotlin coroutine (suspend func)

In this case, the current `org.junit.jupiter.api.assertThrows` doesn't work because in order to use it you would have to make your test function (the one annotated with `@ParameterizedTest`) an `inline func` with a reified `T: Throwable` however this won't be picked up by JUnit as `@ParameterizedTest` uses reflection.

This PR adds variants of the currently existing `org.junit.jupiter.api.assertThrows` that have an `expectedType: Class<T>` as the first parameter (which is the same convention as `org.junit.jupiter.api.Assertions.assertThrows`) but since these added variants are `inline fun`, functions/methods used in the body can be `suspend func` (this is verified with the added tests).

The current implementation is a bit of boilerplate however this is inevitable to a degree since we cannot use the currently existing `org.junit.jupiter.api.assertThrows` and its variants implementations as they only accept reified `T: Throwable` which then defeats the purpose of what we are trying to achieve.

---

---

### Definition of Done

- [X] Public API has [Javadoc](https://github.com/junit-team/junit-framework/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [X] Change is documented in the [User Guide](https://docs.junit.org/snapshot/) and [Release Notes](https://docs.junit.org/snapshot/release-notes.html)


marcphilipp:
@mdedetrich Can't you use `Assertions.assertThrows` directly?

mdedetrich:
> @mdedetrich Can't you use `Assertions.assertThrows` directly?

Not with a kotlin suspend function that you are asserting, thats the precise issue. It will fail to compile for the reasons I stated in the PR. Kotlin suspend only works if the `assertThrow` definitions are `inline func` (which `Assertions.assertThrows` is not since its written in Java) or if the block parameter argument has `suspend func () ->` definition.

Of course you could just do

```kotlin
Assertions.assertThrows(IllegalArgumentException::class.java) {
  runBlocking {
    someKotlinSuspendFunc()
  }
}
```

But one of the whole point of `org.junit.jupiter.api.*` assert functions existing is so that they can work with kotlin coroutines directly without the using of `runBlocking`/`runTest`.

I have just rebased the PR against latest main to resolve conflicts.

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -300,6 +333,25 @@ inline fun <reified T : Throwable> assertThrows(
...
+ * }
+ * assertEquals("Talk to a duck", exception.message)
+ * ```
+ *
+ * This version of assertThrows is intended to be used in situations where reified Throwable's cannot be used,
+ * i.e. when using a @ParameterizedTest that provides exception classes as values.
+ * @see Assertions.assertThrows
+ */
+inline fun <T : Throwable> assertThrows(
+    expectedType: Class<T>,
```
marcphilipp: Should we use `KClass` here to get rid of the `.java`?

marcphilipp:
> Kotlin suspend only works if the `assertThrow` definitions are `inline func`

Sorry for making you re-explain it. I didn't read carefully enough. :grimacing:

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
mdedetrich: Yes fair point, as only Kotlin people will be using this API. I will update it and let you know when its done.

mdedetrich: I have just rebased and pushed the PR to use `KClass` instead of `Class`

marcphilipp:
I wonder whether having the reified and the new variants is confusing and we should deprecate the old ones. WDYT? 🤔

mdedetrich:
> I wonder whether having the reified and the new variants is confusing and we should deprecate the old ones. WDYT? 🤔

I personally see value in both and actually prefer the original variants. Its just that there are some cases where you are using exceptions as values (i.e. `@ParameterizedTest`) and there isn't anything you can do about it.

While its true that there are a lot of overloads, the design is intelligent in that the ones that use exception values always have exception as first arg so the quality of life of users with IDE's should be a non issue (talking about ergonomics of finding the right overload).

tld;r I think its a non issue and there is value in both cases, like generally with JVM languages in some cases you treat types as values and in other cases as type parameters.

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
```diff
@@ -285,6 +286,39 @@ inline fun <reified T : Throwable> assertThrows(executable: () -> Unit): T {
...
+ *     throw IllegalArgumentException("Talk to a duck")
+ * }
+ * assertEquals("Talk to a duck", exception.message)
+ * ```
+ *
+ * This version of assertThrows is intended to be used in situations where reified Throwable's cannot be used,
+ * i.e. when using a @ParameterizedTest that provides exception classes as values.
+ * @see Assertions.assertThrows
+ */
+inline fun <T : Throwable> assertThrows(
```
marcphilipp, line 304: Please add `@API(status = EXPERIMENTAL, since = "6.2.0")` to all new functions.

marcphilipp (changes requested):

junit-jupiter-api/src/main/kotlin/org/junit/jupiter/api/Assertions.kt
mdedetrich: Done, I have added it to all functions, rebased and pushed.

marcphilipp (approved):
Thanks! 👍

mdedetrich:
> Thanks! 👍

Thanks for spending the time on this and fixing up pr 😅

Issue #5017, opened 2025-10-02
Support text blocks in `@ValueSource` and deprecate `textBlock` attribute in `@CsvSource`


manoelcampos:
## Introduction

JUnit 6 requires JDK 17 and text blocks are already supported in `@CsvSource(textBlock = ...)`. However, we need to explicitly choose between using a text block or a regular string.

[JEP 378](https://openjdk.org/jeps/378) aimed for the interchangeability of regular Strings and Text Blocks, but that `textBlock` attribute in `@CsvSource()` forces users to choose between them.

## `@CsvSource`

The code snippet below shows that we need to use `textBlock` property to pass a multiline String and the test is run 3 times:

```java
@ParameterizedTest
@CsvSource(textBlock =
        """
        2,1
        6,3
        10,5
        """)
void csvSourceTextBlock(int a, int b) {
    assertEquals(a, b*2);
}
```

But using the `value` default property, only the first line is parsed:

```java
@ParameterizedTest
@CsvSource(
        """
        100,50
        200,100
        300,150
        """)
void csvSourceTextBlockUsingValueProp(int a, int b) {
    assertEquals(a, b*2);
}
```

## `@ValueSource`

This other annotation doesn’t even allow using text blocks as one would expect. Despite using a text block the code compiles, it doesn't give the expected result:

```java
@ParameterizedTest
@ValueSource(strings =
        """
        L1
        L2
        L3
        """)
void valueSourceTextBlock(final String value) {
    assertFalse(value.contains("\n"), "Each value must represent a single line only (without line break)");
}
```

Instead of running the test 3 times (one line at a time), it runs a single time with value "L1 L2 L3".

## Motivation

Leveraging the utilization of text blocks and providing and more natural way to use them interchangeably with regular Strings will make tests easier and cleaner. 

This code below:

```java
@ValueSource(
        """
        One
        Two
        Three
        """
)
```

is much cleaner than:

```java
@ValueSource(strings = {
        "One",
        "Two",
        "Three"
})
```

### Proposal for `@ValueSource`

Add a `String value` default property and leave the `strings` for backward compatibility and for users that prefer passing values as an array. Internally, if the `value` has a line break, split it and run as if a String array was given to the `strings` property.

### Proposal for `@CsvSource`

- Make the annotation split the `value` property at line breaks and run the test once for each resulting line.
- Deprecate the `textBlock` property

# Final Thoughts

[...]

PR #5852, opened 2026-07-09
Parse CsvSource.value[] elements  separately

mpkorstanje:
In JUnit v5, when using `CsvSource.value[]` the expectation is that user declares individual CSV records. When declaring multiple records either through a text block, or explicit new-line, any records after the first were ignored.

However in v6 we switched parsers to FastCSV in https://github.com/junit-team/junit-framework/pull/4606. This changed the implementation. Instead of treating each `value` as a single record, all the `value[]` elements are joined with `\n` and then treated as if they were a single CSV document.

This has the benefit that multiple records can now be declared in a single string, but it is now also possible that unclosed qoutes from one record can bleed into the next record. We don't consider the latter to be desirable.

For example:

```
@CsvSource(
  "a, 'b", 
  "c', d"
)
```

Should provide two records, containing two fields each. Instead we find fields `a`, `b\nc` and `d`. This should be fields `a`, `b`, `c'` and `d`. Note: The parser leniently closes the unclosed quote for b.

By treating each `value[]` element as if it was an dependent document we prevent quotes from bleeding over while still allowing each `value[]` element to contain multiple records

Contributes to: #5017

---

---

### Definition of Done


junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderFactory.java
```diff
@@ -88,7 +89,8 @@ static CsvReader<? extends CsvRecord> createReaderFor(CsvSource csvSource, Strin
 				Set.of(csvSource.nullValues()),
 				csvSource.ignoreLeadingAndTrailingWhitespace(),
 				csvSource.maxCharsPerColumn(),
-				csvSource.useHeadersInDisplayName()
+				// For CsvSource.value we manually process the header.
+				hasTextBlock && csvSource.useHeadersInDisplayName()
```
mpkorstanje: Implicit repeated condition in provideArgumentsFromValueArray. De-duplicate and pass as parameter.

mpkorstanje: Maybe wrap `CsvSource` and `CsvFileSource` so we have all this logic in one place.

marcphilipp (commented):
Looks very reasonable! Is there anything you'd like feedback on?

mpkorstanje:
> Looks very reasonable! Is there anything you'd like feedback on?

No. Biggest point of discussion would be what to decide what, if anything, to do after this.

This PR is far from done though. At the moment there are two effective configurations spread out over three different code paths. I'd like to reduce the rube-goldberg-esq nature somewhat.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvSource.java
```diff
@@ -122,20 +118,54 @@
 	 *     // ...
 	 * }</pre>
 	 *
+	 * <h4>Text Blocks</h4>
```
mpkorstanje, line 121: The implementation for `CsvSource.value` supports whole CSV documents while the documentation pretends it only supports records. Trying to fix that pretense makes the explanation of the basic use case unreadable, so I've duplicated most of the the `textBlock` documentation here as pretend "different" usecase.

I'm open to better ideas.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderConfiguration.java
```diff
@@ -0,0 +1,323 @@
...
+
+import de.siegmar.fastcsv.reader.CommentStrategy;
+
+import org.junit.platform.commons.util.Preconditions;
+
+/**
+ * Wrapper for {@link CsvSource} with {@code value}, {@link CsvSource} with a
+ * {@code textBlock} and {@link CsvFileSource}.
+ */
+abstract class CsvReaderConfiguration {
```
mpkorstanje: This class is rather verbose, but it also makes it explicit that we have 3 similar but different use-cases (`CsvSource.value`, `CsvSource.textBlock`, `CsvFileSource`) and that `CsvSource.textBlock` are `CsvFileSource` identical.

I prefer it over having two different but similar methods to create `CsvParsers`, but that is just a preference.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvSource.java
marcphilipp: I think that's fine and I don't have a better idea.

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderConfiguration.java
marcphilipp: Probably a nit-pick, but I wonder if it would be clearer/shorter if this class was a `record` which would avoid the delegation pattern. I'll let you decide, though.

marcphilipp (approved):

junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/CsvReaderConfiguration.java
mpkorstanje: It does look much better.

PR #5881, opened 2026-07-17
Clean up Spotless conventions to avoid scanning non-project files

marcphilipp:


sormuras (approved):
Ran flawless on my machine: https://ge.junit.org/s/3vezuw7e73mmc

PR #5930, opened 2026-07-30
Support verbose console output with parallel execution

### LINKED ISSUE #5922: Support verbose console launcher output in combination with parallel execution
When the ConsoleLauncher is run with `--details=verbose` and parallel execution is enabled,
the reported per-test durations do not correspond to the tests they are printed for, and
individual output lines are interleaved and thus garbled. `--details=tree` is not affected.

## Steps to reproduce

Two test classes with clearly separated durations:

```java

public class LongTest {
    @Test void t01() throws Exception { Thread.sleep(400); }
    // ... t02 .. t10, all sleeping 400 ms
}
```

```java

public class ShortTest {
    @Test void t01() throws Exception { Thread.sleep(10); }
    // ... t02 .. t10, all sleeping 10 ms
}
```

Run with the released standalone JAR:

```
java -jar junit-platform-console-standalone-6.1.1.jar execute \
  --class-path classes --select-class LongTest --select-class ShortTest \
  --details=verbose --disable-ansi-colors --disable-banner \
  --config junit.jupiter.execution.parallel.enabled=true \
  --config junit.jupiter.execution.parallel.mode.default=concurrent \
  --config junit.jupiter.execution.parallel.mode.classes.default=concurrent \
  --config junit.jupiter.execution.parallel.config.strategy=fixed \
  --config junit.jupiter.execution.parallel.config.fixed.parallelism=4
```

**Expected:** 20 reported durations — ten close to 400 ms and ten close to 10 ms.

**Actual:** the reported durations do not match the tests that were executed. Ten tests slept
400 ms each, but only four of the twenty reported durations were above 300 ms:

```
reported durations, parallel execution
   1x     0 ms     3x    28 ms
   1x     1 ms     1x    41 ms
   1x     2 ms     1x    42 ms
   1x     9 ms     1x    44 ms
   2x    11 ms     1x   348 ms
   3x    12 ms     1x   351 ms
   1x    13 ms     1x   363 ms
                   1x   406 ms
```

With parallel execution disabled, the same classes report correctly — ten durations between
13 ms and 15 ms and ten between 405 ms and 427 ms.

A single-test class shows the inconsistency within one run, since the container duration and
its only test's duration have to agree:

```
│  │  │   duration: 1954 ms
│  └─ SlowTest finished after 2027 ms.
```

### Interleaved output

In the same runs, individual lines were garbled because several threads write to the same
`PrintWriter` while printing one event:

```
│  │  │  │  │  │  │  │  │  │  │  │     status:    status: ✔ SUCCESSFUL
│  │  │  │   duration:  duration: 2 ms
│  │  │  │  │  │  │  │   uniqueId:    parent:  uniqueId: [engine:junit-jupiter]/[class:SlowTest]/[method:slow()]
```

In one run of 151 tests, only 140 `duration:` values remained parseable for this reason. No
garbled lines occur with parallel execution disabled.

### Root cause

`VerboseTreePrintingListener` stores its timing state in a shared stack and a single field:

```java
private final Deque<Long> frames;        // an ArrayDeque; :37, :48
private long executionStartedMillis;     // :39
```

`executionStartedMillis` is written in `executionStarted` (:86) and read in
`executionFinished` (:111), and `frames` is pushed in `executionStarted` (:91) and popped in
`executionFinished` (:105). `TestExecutionListener` documents that these methods "could be
called from different threads concurrently in case one or multiple test engines execute tests
in parallel", so neither a single field nor a stack can attribute a start time to the
execution that is finishing. `ArrayDeque` is also not thread-safe.

`TreePrintingListener` solves the same problem by keeping the state per test:
`Map<UniqueId, TreeNode>` in a `ConcurrentHashMap` (:36), with each `TreeNode` holding its own
`creation` timestamp and computing `duration` when its result is set (`TreeNode.java:31, :46,
:78`).

## Context

 - Used versions (Jupiter/Vintage/Platform): reproduced with the 6.1.1 standalone JAR; the same logic is present on `main` (6.2.0-SNAPSHOT)
 - Build Tool/IDE: ConsoleLauncher, JDK 21

## Deliverables

- [ ] Track the start time per execution, keyed by `UniqueId`, instead of a shared stack and field
- [ ] Print each event as a single write so concurrent events cannot interleave within a line
- [ ] Add a regression test covering concurrent executions with distinguishable durations

big-cir:
When tests ran in parallel, `--details=verbose` reported the duration of other executions, indented the tree by the wrong amount, and merged output lines.

The listener kept the start time in a single field, derived the depth from a shared stack, and wrote each line to the `PrintWriter` in several steps.

### Changes

* Start times are now kept in a map keyed by unique ID and removed when an execution finishes.
* The nesting level is derived from the number of ancestors in the test plan.
* The output of an event is collected and written at once.

### Verification

Ran two classes: `SlowTest` with one test sleeping 3000 ms and `TickTest` with ten tests sleeping 300 ms each, using `--details=verbose` and `parallelism=4`.

**Before:**

```
│  │  └─ TickTest finished after 1271 ms.
│  │  │   duration: 2068 ms
│  │  │     status: ✔ SUCCESSFUL
│  └─ SlowTest finished after 3025 ms.
```

**After:**

```
│  └─ TickTest finished after 1254 ms.
│  │  │   duration: 3020 ms
│  │  │     status: ✔ SUCCESSFUL
│  └─ SlowTest finished after 3026 ms.
```

`SlowTest` has a single test that slept 3000 ms; it now reports its own duration instead of the one of a `TickTest` test. `TickTest` is a child of the engine and is no longer indented one level too deep.

Across the same run, the number of lines carrying two labels went from 3 to 0. Sequential output is unchanged.

Closes: #5922

---

---

### Definition of Done

- [x] Change is covered by automated tests including corner cases, errors, and exception handling — three tests cover overlapping executions: per-execution durations, indentation, and non-interleaved lines
- [x] Method preconditions are checked and documented in the method's Javadoc — n/a; no new public method
- [x] Public API has Javadoc and `@API` annotations — n/a; no public API was added or changed
- [x] Change is documented in the User Guide and Release Notes — Release Notes entry added; the User Guide does not describe the output format


junit-platform-console/src/main/java/org/junit/platform/console/output/VerboseTreePrintingListener.java
```diff
@@ -34,201 +38,242 @@ public class VerboseTreePrintingListener implements DetailsPrintingListener {
...
+	 * Determine the nesting level of the supplied test identifier, i.e. the number
+	 * of its ancestors in the test plan, with test engines being at level 0.
+	 */
+	private int nestingLevel(TestIdentifier testIdentifier) {
+		TestPlan testPlan = requireNonNull(this.testPlan);
+		int nestingLevel = 0;
+		TestIdentifier current = testIdentifier;
+		// roots are the only identifiers without a parent in the test plan, so
+		// getParent(...) does not fail for anything else
+		while (!testPlan.getRoots().contains(current)) {
```
mpkorstanje: While it is likely that the root of a test plan is also a test identifier without a parent, it would be more accurate for the loop condition to halt when current node has no parent.

Typically this would be written like this:

```
while ((parent = testPlan.getParent().orElse(null)) != null) {
```

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
```diff
@@ -64,46 +83,208 @@ void fileEntryPublished() {
...
+				"test" + i + "()");
+			engine.addChild(test);
+			identifiers.add(TestIdentifier.from(test));
+		}
+
+		var stringWriter = new StringWriter();
+		var listener = listener(stringWriter, engine);
+		var barrier = new CyclicBarrier(NUM_THREADS);
+		var executor = Executors.newFixedThreadPool(NUM_THREADS);
+		try {
```
mpkorstanje:
```suggestion
		try (var executor = Executors.newFixedThreadPool(NUM_THREADS)) {
```

mpkorstanje: And await the completion of all submitted tasks instead of waiting for the executor in the `finally` block. You can use the `@Timeout` annotation on the test method to make sure the whole test completes in a timely manner.

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
```diff
@@ -64,46 +83,208 @@ void fileEntryPublished() {
...
+
+	@Test
+	void linesAreNotInterleavedWhenExecutionsArePrintedConcurrently() throws Exception {
+		var engine = new TestDescriptorStub(UniqueId.forEngine("demo-engine"), "demo-engine");
+		var identifiers = new ArrayList<TestIdentifier>();
+		for (int i = 0; i < TESTS_PER_THREAD * NUM_THREADS; i++) {
+			var test = new TestDescriptorStub(engine.getUniqueId().append("method", "test" + i + "()"),
+				"test" + i + "()");
+			engine.addChild(test);
+			identifiers.add(TestIdentifier.from(test));
```
mpkorstanje: This would look nicer if you used `IntStream.range(0, TESTS_PER_THREAD * NUM_THREADS)` and then mapped the ints into test descriptor stubs which are then added to the engine with `testDescriptors.forEach(engine::addChild)` and then separately mapped into test identifiers.

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
```diff
@@ -64,46 +83,208 @@ void fileEntryPublished() {
...
+		engine.addChild(slow);
+		engine.addChild(fast);
+
+		var stringWriter = new StringWriter();
+		var listener = listener(stringWriter, engine);
+		var slowIdentifier = TestIdentifier.from(slow);
+		var fastIdentifier = TestIdentifier.from(fast);
+
+		listener.executionStarted(slowIdentifier);
+		Thread.sleep(100);
```
mpkorstanje: Using `Thread.sleep` in a test is a signal that the code under test should use a `Clock`. You can then mock the clock to simulate the progress of time in a controlled manner. Then instead of storing the started milis in `startedMillisByUniqueId` you could store instances of `Instant` instead.

This also means that in other tests where you've used a wildcard to match the duration, you can now match an exact number. And you can probably compare two multi-line strings instead of matching individual lines because the output should be completely deterministic.

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
```diff
@@ -64,46 +83,208 @@ void fileEntryPublished() {
...
+	private VerboseTreePrintingListener listener(StringWriter stringWriter, TestDescriptor engineDescriptor) {
+		var listener = new VerboseTreePrintingListener(new PrintWriter(stringWriter), ColorPalette.NONE, 16,
+			Theme.ASCII);
+		listener.testPlanExecutionStarted(
+			TestPlan.from(true, Set.of(engineDescriptor), mock(), dummyOutputDirectoryCreator()));
+		discardOutput(stringWriter);
+		return listener;
+	}
+
+	private static final TestDescriptor TEST_DESCRIPTOR = new TestDescriptorStub(UniqueId.forEngine("demo-engine"),
```
mpkorstanje: This doesn't have to be a constant.

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
```diff
@@ -64,46 +83,208 @@ void fileEntryPublished() {
...
-		return new VerboseTreePrintingListener(new PrintWriter(stringWriter), ColorPalette.NONE, 16, Theme.ASCII);
+		return listener(stringWriter, TEST_DESCRIPTOR);
 	}
 
+	private VerboseTreePrintingListener listener(StringWriter stringWriter, TestDescriptor engineDescriptor) {
+		var listener = new VerboseTreePrintingListener(new PrintWriter(stringWriter), ColorPalette.NONE, 16,
+			Theme.ASCII);
+		listener.testPlanExecutionStarted(
+			TestPlan.from(true, Set.of(engineDescriptor), mock(), dummyOutputDirectoryCreator()));
+		discardOutput(stringWriter);
```
mpkorstanje: This doesn't look right. There should be no reason to discard output from the tests. Rather you can update the existing tests to include the new output.

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
```diff
@@ -64,46 +83,208 @@ void fileEntryPublished() {
...
 
 	private VerboseTreePrintingListener listener(StringWriter stringWriter) {
-		return new VerboseTreePrintingListener(new PrintWriter(stringWriter), ColorPalette.NONE, 16, Theme.ASCII);
+		return listener(stringWriter, TEST_DESCRIPTOR);
 	}
 
+	private VerboseTreePrintingListener listener(StringWriter stringWriter, TestDescriptor engineDescriptor) {
+		var listener = new VerboseTreePrintingListener(new PrintWriter(stringWriter), ColorPalette.NONE, 16,
+			Theme.ASCII);
+		listener.testPlanExecutionStarted(
```
mpkorstanje: Changing the state of the object being tested in a helper method is unexpected. I would expect individual test methods to call this. It does create some duplication, but that also makes the tests more clear.

mpkorstanje (changes requested):
Looks good! 

Aside from the mostly stylistic comments below, obtaining the time from a `Clock` in the `VerboseTreePrintingListener` would simplify the assertions in the tests quite a bit.

Because this makes the printer deterministic you can probably also compare the actual output against the expected output written as a text block. This would give a better impression of what the actual output should look like.

junit-platform-console/src/main/java/org/junit/platform/console/output/VerboseTreePrintingListener.java
big-cir: Applied, thanks!

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
big-cir: Applied, thanks!

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
big-cir: Applied, thanks!

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
big-cir: Applied, thanks! 

The listener now takes an injected `Clock` and stores `Instant` values instead of millis.

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
big-cir: Applied, thanks!

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
big-cir: Applied, thanks!

platform-tests/src/test/java/org/junit/platform/console/output/VerboseTreePrintingListenerTests.java
big-cir: Applied, thanks!

big-cir:
> Looks good!
> 
> Aside from the mostly stylistic comments below, obtaining the time from a `Clock` in the `VerboseTreePrintingListener` would simplify the assertions in the tests quite a bit.
> 
> Because this makes the printer deterministic you can probably also compare the actual output against the expected output written as a text block. This would give a better impression of what the actual output should look like.

Thanks for the review!

I've updated the listener to use an injected `Clock`, removed `Thread.sleep` from the tests, assert the verbose output as text blocks where it's deterministic, and refactored the test setup as suggested.

mpkorstanje (approved):
Cheers! I think we're done.

The original `VerboseTreeListenerTests` wasn't great. So I've pushed a few additional changes to address nitpicks, lack of coverage and other legacy issues. I get the impression that you've been intentionally quite conservative here, so that is okay. Please do have a look at test to get an idea of the style we're going for. It will help speed up future contributions.

The most important things to note are:

1. The Arrange-Act-Assert pattern in each test
2. Removal of incidental details (where reasonable, the concurrency test still has some)
3. Assert against the whole output as a textblock using `assertLinesMatch` under the hood for pattern matching.

big-cir:
> Cheers! I think we're done.

Thanks for the thorough review and all the improvements to the tests.

I learned a lot from this change. I’ll keep the Arrange-Act-Assert style and text block assertions in mind for future contributions!

PR #5932, opened 2026-07-30
Generate configuration parameters overview 

### LINKED ISSUE #4330: Configuration Parameters Overview
The user guide mentions several [configuration parameters](https://junit.org/junit5/docs/current/user-guide/#running-tests-config-params) for the Jupiter Test Engine. However, they are a bit scattered throughout the chapters.

What do you think about adding an overview to the [appendix](https://junit.org/junit5/docs/current/user-guide/#appendix), in the style of the Spring Boot [Common Application Properties](https://docs.spring.io/spring-boot/appendix/application-properties/index.html)?

In addition to the user guide benefit, I would like to ask JetBrains if they are open to support code completion in `junit-platform.properties` files. Having such an overview in the docs might help with that request.

## Related Issues

- #2516

## Deliverables

- [ ] Add a configuration parameters overview to the user guide

mpkorstanje:
Generate `META-INF/junit-platform-configuration-metadata.json` in the format of [Spring Boots Configuration Metadata](https://docs.spring.io/spring-boot/specification/configuration-metadata/format.html).

### Minimal example

The minimal usage:

```java
/**
 * A brief multi-line description of
 * this property: {@value}.
 *
 * <p>Followed by an additional paragraph.
 */
@ConfigurationParameter
public static final String EXAMPLE_PROPERTY_NAME = "org.example.property";
```

Produces:

```json
{
  "properties": [
	{
	  "name": "org.example.property",
      "description": "A brief multi-line description of this property.",
	  "sourceType": "com.example.app.Constants"
	}
  ]
}
```					

This takes all the information from the obvious places. Of note is that the first paragraph from the doc string is used and the `: {@value}.` that we conventionally use is replaced with a `.`.

### Deprecation

```java
@Deprecated
@ConfigurationParameter
public static final String EXAMPLE_PROPERTY_NAME = "org.example.property";
```

Produces:

```json
{
  "properties": [
	{
	  "name": "org.example.property",
      "description": "A brief multi-line description of this property.",
	  "sourceType": "com.example.app.Constants",
      "deprecated": { }
	}
  ]
}
```

The empty `deprecated` object is sufficient to mark a field as deprecated. To fill out the fields in the `deprecated` object the `@Deprecation` annotation can be used:

```java
@Deprecated
@ConfigurationParameter(deprecation = @Deprecation(reason = "A an example reason", replacement = "com.example.app.Constants#EXAMPLE_PROPERTY_NAME_V2", since = "2.0.0"))
public static final String EXAMPLE_PROPERTY_NAME_V1 = "org.example.property";
```

## Type and default

The type of the parameter can be provided using the `type` property of the annotation.

```java
@ConfigurationParameter(type = String.class)
public static final String EXAMPLE_PROPERTY_NAME = "org.example.property";
```

Though in practice this will be rarely needed. If the type is not provided, we can take the type of the default value.

```java
@ConfigurationParameter(defaultValue = @Value(stringValue = "default"))
public static final String EXAMPLE_PROPERTY_NAME = "org.example.property";
```

When using Enums it is useful to use both type and default value, because arbitrary enums can't be used in an annotation.

```java
@ConfigurationParameter(defaultValue = @Value(intValue = 42))
public static final String INTS_PROPERTY_NAME = "org.example.ints";
```

## Limitations

- `Metadata.groups`, `Metadata.hints`, `Metadata.ignored` are not yet implemented. See: https://github.com/junit-team/junit-framework/pull/6059
- While `Property.default` supports lists of values, the implementation is intentionally limited to a single value.

Closes: #4330

---

---

### Definition of Done


gradle/plugins/backward-compatibility/src/main/kotlin/junitbuild.backward-compatibility.gradle.kts
```diff
@@ -89,6 +89,7 @@ val checkBackwardCompatibility = tasks.register("checkBackwardCompatibility") {
 	dependsOn(roseau)
 }
 
-tasks.check {
-	dependsOn(checkBackwardCompatibility)
-}
+// TODO: Breaks because the configuration-api hasn't been released yet.
+//tasks.check {
+//	dependsOn(checkBackwardCompatibility)
```
mpkorstanje: @marcphilipp could you look at this?

junit-platform-configuration-processor/junit-platform-configuration-processor.gradle.kts
marcphilipp, line 1: I think we should build this as a fat/shadowed jar to make it easy to put it on the annotation processor classpath. No need to do that now, but let's discuss it.

junit-platform-configuration-api/junit-platform-configuration-api.gradle.kts
```diff
@@ -0,0 +1,17 @@
+plugins {
+	id("junitbuild.java-library-conventions")
+}
+
+description = "JUnit Platform Configuration API"
+
+dependencies {
+	api(platform(projects.junitBom))
+	api(projects.junitPlatformCommons)
```
marcphilipp: Does it need that?

junit-jupiter-api/junit-jupiter-api.gradle.kts
```diff
@@ -35,7 +38,7 @@ eclipseConventions {
 
 tasks {
 	compileJava {
-		options.compilerArgs.add("-Xlint:-module") // due to qualified exports
+		options.compilerArgs.add("-Xlint:-module,-processing") // -module: due to qualified exports, -processing: not all annotations need to be processed
```
marcphilipp, line 42: What warnings does this emit without `-processing`?

junit-jupiter-migrationsupport/junit-jupiter-migrationsupport.gradle.kts
```diff
@@ -12,6 +12,7 @@ dependencies {
 
 	compileOnlyApi(libs.apiguardian)
 	compileOnlyApi(libs.jspecify)
+	compileOnlyApi(projects.junitPlatformConfigurationApi)
```
marcphilipp: I think `compileOnly` should suffice since retention is `SOURCE`.

junit-platform-configuration-api/junit-platform-configuration-api.gradle.kts
mpkorstanje: No, but the way OSGI is setup makes it a requirement for every module.

junit-jupiter-api/junit-jupiter-api.gradle.kts
mpkorstanje: Complain about all present annotations that weren't claimed by a processor.

Similar too:

https://stackoverflow.com/questions/69439506/no-processor-claimed-any-of-these-annotations

junit-platform-configuration-processor/junit-platform-configuration-processor.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+dependencies {
+	api(platform(projects.junitBom))
+	api(projects.junitPlatformConfigurationApi)
+	api(projects.junitPlatformCommons)
+
+	compileOnlyApi(libs.apiguardian)
+	compileOnlyApi(libs.jspecify)
+
+	// TODO: Shade, but it's non trivial.
+	implementation(libs.jackson.databind)
```
mpkorstanje: @marcphilipp I'm having my doubts about picking Jackson. 

* It is a CVE magnet, so we'd have to backport dependency updates quite frequently. Though not many people would actually use the processor.
* Is overkill for what we need.
* Shading Jackson is quite complicated because of various service descriptors and miscellaneous junk in the jar files.

The spring-boot-configuration-processor` resolves this by [copying a minimal json implementation](https://github.com/spring-projects/spring-boot/tree/main/configuration-metadata/spring-boot-configuration-metadata/src/json-shade). But that implementation is Apache licensed.

There is https://github.com/stleary/JSON-java but the "vibes" are off. No clear license, calendar based versioning, ect.

marcphilipp: > It is a CVE magnet, so we'd have to backport dependency updates quite frequently. Though not many people would actually use the processor.

That's a good point! Can we get by with just jackson-core which has fewer CVEs?

Some potential alternatives:
- https://github.com/google/gson
- https://github.com/eclipse-ee4j/parsson

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationMetadataAnnotationProcessor.java
```diff
@@ -0,0 +1,155 @@
...
+		return SourceVersion.latestSupported();
+	}
+
+	@Override
+	public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
+		process(roundEnv);
+		if (roundEnv.processingOver()) {
+			writeMetaData();
+		}
+		return false;
```
marcphilipp: Shouldn't this return `true` if `annotations` contains only `@ConfigurationParameter`?

junit-platform-configuration-api/junit-platform-configuration-api.gradle.kts
marcphilipp: It does actually use it now (for `Preconditions`)

junit-platform-configuration-processor/junit-platform-configuration-processor.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+dependencies {
+	api(platform(projects.junitBom))
+	api(projects.junitPlatformConfigurationApi)
+	api(projects.junitPlatformCommons)
+
+	compileOnlyApi(libs.apiguardian)
+	compileOnlyApi(libs.jspecify)
+
+	shadowed(libs.jakarta.json.api)
+	// TODO: Check if shadowed artifact is standalone
```
marcphilipp: I wasn't sure what you meant here so I left the todo in place.

junit-platform-configuration-api/src/main/java/org/junit/platform/configuration/api/ConfigurationProperty.java
```diff
@@ -0,0 +1,24 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.apiguardian.api.API;
+
+@API(status = API.Status.EXPERIMENTAL)
+@Retention(RetentionPolicy.SOURCE)
+@Target(ElementType.FIELD)
+public @interface ConfigurationProperty {
```
marcphilipp: Shouldn't this be called `ConfigurationParameter`?

junit-platform-configuration-api/junit-platform-configuration-api.gradle.kts
marcphilipp: Actually, I mixed things up with the processor. I've now removed the dependency in 433b3b94852a49096d74e985e68ff4a85df42f91.

junit-platform-configuration-processor/junit-platform-configuration-processor.gradle.kts
mpkorstanje: Effectively:

* Inspect the contents of the artifact. 
* Inspect the contents of `pom.xml`
* Test the processor with a Maven project.

junit-platform-configuration-api/src/main/java/org/junit/platform/configuration/api/ConfigurationProperty.java
mpkorstanje: We already have a `@ConfigurationParameter` annotation that is specified as a key-value pair and that seems to be common language. What do you think about `ConfigurationParameterKey`?

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationMetadataAnnotationProcessor.java
mpkorstanje: If want to claim the annotation, yes. But I don't see a reason to claim it and deny the annotation to other processors.

platform-tooling-support-tests/projects/jar-describe-module/junit-platform-commons.expected.txt
```diff
@@ -16,4 +16,4 @@ requires org.apiguardian.api static transitive
...
 qualified exports org.junit.platform.commons.logging to org.junit.jupiter.api org.junit.jupiter.engine org.junit.jupiter.migrationsupport org.junit.jupiter.params org.junit.platform.console org.junit.platform.engine org.junit.platform.launcher org.junit.platform.reporting org.junit.platform.suite.api org.junit.platform.suite.engine org.junit.platform.testkit org.junit.vintage.engine
-qualified exports org.junit.platform.commons.util to org.junit.jupiter.api org.junit.jupiter.engine org.junit.jupiter.migrationsupport org.junit.jupiter.params org.junit.platform.console org.junit.platform.engine org.junit.platform.launcher org.junit.platform.reporting org.junit.platform.suite.api org.junit.platform.suite.engine org.junit.platform.testkit org.junit.vintage.engine
+qualified exports org.junit.platform.commons.util to org.junit.jupiter.api org.junit.jupiter.engine org.junit.jupiter.migrationsupport org.junit.jupiter.params org.junit.platform.configuration.processor org.junit.platform.console org.junit.platform.engine org.junit.platform.launcher org.junit.platform.reporting org.junit.platform.suite.api org.junit.platform.suite.engine org.junit.platform.testkit org.junit.vintage.engine
```
mpkorstanje: The `org.junit.platform.configuration.processor` can be removed now that the dependency on commons has been dropped.

mpkorstanje: Not yet. But can be once we report preconditions problems with `processingEnvironment().getMessager().printMessage()` instead.

gradle/plugins/common/src/main/kotlin/junitbuild/shadow/ConfigurationMetadataMergingTransformer.kt
```diff
@@ -0,0 +1,64 @@
...
+        val merged = mergedTree
+        if (merged == null) {
+            mergedTree = tree
+        } else {
+            tree.properties().forEach { (name, value) ->
+                val existing = merged.get(name)
+                if (existing is ArrayNode && value is ArrayNode) {
+                    existing.addAll(value)
+                } else {
+                    merged.set(name, value)
```
mpkorstanje, line 46: Shouldn't this be an exception since we don't expect only array nodes here?

platform-tooling-support-tests/projects/jar-describe-module/junit-platform-commons.expected.txt
marcphilipp: The processor still uses it, just the api doesn't.

gradle/plugins/common/src/main/kotlin/junitbuild/shadow/ConfigurationMetadataMergingTransformer.kt
marcphilipp: Done in fa5ab928a74230f0944dc741610b8bfbc6f96cf5

junit-platform-configuration-processor/junit-platform-configuration-processor.gradle.kts
marcphilipp: I checked the dependency graph already when suggesting Parsson and it looks self-contained apart from the api dependency.

> Test the processor with a Maven project

We can do that in `platform-tooling-support-test` as an integration test

junit-platform-configuration-api/src/main/java/org/junit/platform/configuration/api/ConfigurationProperty.java
marcphilipp: > We already have a `@ConfigurationParameter` annotation that is specified as a key-value pair and that seems to be common language

True, but I think we could reuse that name here. The annotation from suite-api isn't visible in any of the projects we're using this new annotation.

> What do you think about `ConfigurationParameterKey`?

It annotates a key but documents its values as well so I'm not convinced that's better.

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationMetadataAnnotationProcessor.java
marcphilipp: I think this is mostly done for performance reasons. We only use this annotation processor but other projects might have additional ones set up.

mpkorstanje: The intake is covered by `@SupportedAnnotationTypes` / `AbstractProcessor::getSupportedAnnotationTypes`.

But I don't see a harm in enabling it either.

junit-platform-configuration-api/src/main/java/org/junit/platform/configuration/api/ConfigurationProperty.java
mpkorstanje: Reasonable enough. Done in 7cf5a6d43757ba12c2c81d4fb6b82fa936f440ee.

We'll have to document that Maven users should declare the dependency as optional and provided (I think).

platform-tests/src/test/java/org/junit/platform/configuration/processor/ConfigurationMetadataAnnotationProcessorTests.java
```diff
@@ -0,0 +1,346 @@
...
+		void defaults() {
+			compiler.compile(Defaults.class);
+			assertMetaDataIsEqualTo("""
+										{
+					  "properties": [
+					    {
+					      "name": "org.example.shorts",
+					      "type": "java.lang.Short",
+					      "sourceType": "org.junit.platform.configuration.testcases.Defaults",
+					      "defaultValue": "1"
```
mpkorstanje: Todo: Check if these should be strings or numbers.

junit-platform-configuration-api/src/main/java/org/junit/platform/configuration/api/ConfigurationParameter.java
```diff
@@ -0,0 +1,133 @@
...
+		 */
+		String since() default "";
+	}
+
+	@interface Value {
+
+		/**
+		 * The {@code short} values to use as the defaults; must not be empty.
+		 */
+		short[] shorts() default {};
```
mpkorstanje: Todo: Can these be made singular?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
```diff
@@ -107,6 +119,8 @@ public final class Constants {
 	 *
 	 * <p>The default behavior is not to perform auto-detection.
 	 */
+	// TODO: Handle @link references.
+	@ConfigurationParameter(defaultValue = @Value(booleanValue = false))
```
mpkorstanje: @marcphilipp to keep the documentation and implementation consistent I think it would help to extract all the default values to constants and use them when parsing configuration parameters. Thoughts?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
```diff
@@ -277,6 +310,7 @@ public final class Constants {
 	 * <p>Value must be a positive decimal number; defaults to {@code 1}.
 	 *
 	 */
+	@ConfigurationParameter(type = BigDecimal.class, defaultValue = @Value(doubleValue = 1.0))
```
mpkorstanje: Do we really need infinite precision here?

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
```diff
@@ -293,13 +329,17 @@ public final class Constants {
 	 *
 	 * @see Timeout#DEFAULT_TIMEOUT_PROPERTY_NAME
 	 */
+	// TODO: TimeoutDuration is not part of the public API
+	@ConfigurationParameter /*(type = TimeoutDuration.class)*/
```
mpkorstanje: Not sure about this yet. We could use a string here, but `TimeoutDuration` has a specific format. 

TODO: Investigate how this works with `Duration` in Spring.

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/AnnotationMirrorUtil.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.AnnotationValue;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+
+import org.jspecify.annotations.Nullable;
+
+class AnnotationMirrorUtil {
+
+	private AnnotationMirrorUtil() {
```
mpkorstanje, line 30: TODO: Make these methods less ad-hoc.

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationMetadataAnnotationProcessor.java
mpkorstanje: Done a235f3e80ad7103354f7773aee98f145e49f69f1.

platform-tooling-support-tests/projects/jar-describe-module/junit-platform-commons.expected.txt
mpkorstanje: Done in 3a6ce7f3d8cb0090609eb3152de155be29c3822a.

junit-platform-configuration-processor/junit-platform-configuration-processor.gradle.kts
mpkorstanje: Done 024414719e99281018c6f5a5aea177195f449a71.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Constants.java
marcphilipp: String plus a regex?

junit-platform-configuration-api/src/test/README.md
mpkorstanje, line 1: TODO: Document circular dependency between tests and annotation processor.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/io/TempDir.java
```diff
@@ -132,13 +132,19 @@
 	@API(status = MAINTAINED, since = "5.13.3")
 	Class<? extends TempDirFactory> factory() default TempDirFactory.class;
 
+	/**
+	 * Default value for {@value #DEFAULT_CLEANUP_MODE_PROPERTY_NAME} is {@value}.
+	 */
+	@API(status = MAINTAINED, since = "6.2.0")
+	String DEFAULT_TEMP_DIR_CLEANUP_MODE_DEFAULT = "always";
```
mpkorstanje, line 139: @marcphilipp I can't infer what the intended location for constants is.

I would expect them to all sit in Constants, but we have `DEFAULT_CLEANUP_MODE_PROPERTY_NAME` duplicated here. This pattern is also present in Execution, TestInstance and Timeout.

Should new constants go here, or into Constants?

mpkorstanje:
TODO:

- [ ] Document the annotation processor in the user guide
- [ ] Support hints (new PR)
- [ ] Support groups (new PR)

documentation/modules/ROOT/pages/advanced-topics/engines.adoc
```diff
@@ -147,3 +147,27 @@ be used as a `Predicate` or `Consumer`. Please refer to the implementations of t
 
 Moreover, xref:advanced-topics/testkit.adoc#engine-discovery[Engine Test Kit] provides a way to write tests for
 reported discovery issues.
+
+[[configuration-parameters]]
+== Configuration Parameters overview
+
+To assist IDEs with code completion test engines may generate machine-readable overview
```
mpkorstanje, line 154: TODO: Also works for extensions.

documentation/modules/ROOT/pages/advanced-topics/junit-platform-configuration.adoc
```diff
@@ -0,0 +1,36 @@
...
+Processing the annotations with `junit-platform-configuration-processor` will produce a
+`META-INF/junit-platform-configuration-metadata.json` file in the format of
+{SpringBootConfigurationMetadata}. This file can then be used to generate documentation
+or provide code-completion by IDEs. An example is shown below.
+
+[source,json,indent=0]
+[subs=attributes+]
+.A minimal `META-INF/junit-platform-configuration-metadata.json`
+----
+{
```
mpkorstanje: @marcphilipp is there a way to use the generated `documentation/build/classes/java/test/META-INF/junit-platform-configuration-metadata.json` instead?

mpkorstanje: I tried to add a `collector` in `antora.yml`, but that seems to do nothing and I can't figure out where it creates the collection.

documentation/modules/ROOT/nav.adoc
```diff
@@ -60,6 +60,7 @@
 ** xref:advanced-topics/testkit.adoc[]
 ** xref:advanced-topics/launcher-api.adoc[]
 ** xref:advanced-topics/engines.adoc[]
+** xref:advanced-topics/junit-platform-configuration.adoc[]
```
marcphilipp: I think the file name (which is also used as the published path) could be more specific. Maybe something like this?

```suggestion
** xref:advanced-topics/configuration-parameter-documentation.adoc[]
```

documentation/modules/ROOT/pages/advanced-topics/configuration-parameter-documentation.adoc
```diff
@@ -0,0 +1,36 @@
+= Documenting Configuration Parameters
+
+Authors of test engines and JUnit Jupiter extensions may generate a machine-readable
+overview of configuration parameters by annotating their configuration parameters
+with `{PlatformConfigurationParameter}` from the `junit-platform-configuration-api`
```
marcphilipp, line 5: The new modules also need to be included in the "Dependency Metadata" section of the appendix and the component diagram.

documentation/modules/ROOT/pages/advanced-topics/junit-platform-configuration.adoc
```diff
@@ -0,0 +1,36 @@
...
+[source,java,indent=0]
+.A small example
+----
+include::example$java/example/ConfigurationParametersDemo.java[tags=user_guide]
+----
+
+Processing the annotations with `junit-platform-configuration-processor` will produce a
+`META-INF/junit-platform-configuration-metadata.json` file in the format of
+{SpringBootConfigurationMetadata}. This file can then be used to generate documentation
+or provide code-completion by IDEs. An example is shown below.
```
marcphilipp:
```suggestion
or provide code completion by IDEs. An example is shown below.
```

documentation/modules/ROOT/pages/advanced-topics/junit-platform-configuration.adoc
marcphilipp: Done in 4550c4992d5a33214eec1adc73adbb8e3bf1cd6f

documentation/src/test/java/example/ConfigurationParametersDemo.java
```diff
@@ -0,0 +1,29 @@
...
+// tag::user_guide[]
+public class ConfigurationParametersDemo {
+
+	/**
+	 * Property name used to set the default test execution mode: {@value}.
+	 * <p>
+	 * Test can be executed with either a fixed number of threads, a number based on
+	 * the available cores or a custom strategy.
+	 */
+	@ConfigurationParameter(type = ExecutionMode.class, defaultValue = @Value(stringValue = "fixed"))
```
marcphilipp, line 25: Since this is referencing `org.junit.jupiter.api.parallel.ExecutionMode`, it should have one of its enum constants as default value. Alternatively, we could create an enum just for the docs.

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationParameterHandler.java
```diff
@@ -0,0 +1,157 @@
...
+		}
+		if (defaultValues.size() != 1) {
+			messager.printMessage(ERROR, "@ConfigurationParameter must have exactly one default value", field.element(),
+				field.annotationMirror());
+		}
+		var entry = defaultValues.entrySet().iterator().next();
+		var values = entry.getValue();
+		if (values.size() != 1) {
+			messager.printMessage(ERROR, "@ConfigurationParameter must have exactly one default value", field.element(),
+				field.annotationMirror());
```
marcphilipp, line 123: Shouldn't these checks return `null` rather than causing an exception below when `values.get(0)` is called?

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/JsonConverter.java
```diff
@@ -0,0 +1,144 @@
...
+
+		return builder.build();
+	}
+
+	private void addObjectValue(JsonObjectBuilder builder, String name, Object defaultValue) {
+		if (defaultValue instanceof Short v) {
+			builder.add(name, v);
+		}
+		else if (defaultValue instanceof Byte v) {
+			builder.add(name, "%02X".formatted(v));
```
marcphilipp: Why are we using the hex representation here?

junit-vintage-engine/src/main/java/org/junit/vintage/engine/Constants.java
```diff
@@ -58,8 +74,15 @@ public final class Constants {
 	 * @since 5.12
 	 */
 	@API(status = MAINTAINED, since = "5.13.3")
+	@ConfigurationParameter(defaultValue = @Value(booleanValue = PARALLEL_EXECUTION_ENABLED_DEFAULT))
```
marcphilipp:
```suggestion
	@ConfigurationParameter(defaultValue = @Value(booleanValue = PARALLEL_CLASS_EXECUTION_DEFAULT))
```

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
```diff
@@ -221,8 +239,20 @@ static ParallelExecutionConfiguration toConfiguration(ConfigurationParameters co
...
 
 	static ParallelExecutionConfigurationStrategy getStrategy(ConfigurationParameters configurationParameters) {
-		return valueOf(
-			configurationParameters.get(CONFIG_STRATEGY_PROPERTY_NAME).orElse("dynamic").toUpperCase(Locale.ROOT));
+		return configurationParameters.get(CONFIG_STRATEGY_PROPERTY_NAME, value -> {
+			try {
+				return valueOf(value.toUpperCase(Locale.ROOT));
+			}
+			catch (Exception e) {
+				throw new JUnitException(
```
marcphilipp: `ConfigurationParameters.get` wraps this again and includes both key and value. Therefore, I think wrapping it here is unnecessary.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Execution.java
```diff
@@ -71,7 +77,7 @@
 	 * <p>Supported values include names of enum constants defined in
 	 * {@link ExecutionMode}, ignoring case.
 	 *
-	 * <p>If not specified, the default is "same_thread" which corresponds to
+	 * <p>If not specified, the default is {@link #DEFAULT_EXECUTION_MODE_DEFAULT} which corresponds to
```
marcphilipp:
```suggestion
	 * <p>If not specified, the default is {@value #DEFAULT_EXECUTION_MODE_DEFAULT} which corresponds to
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Execution.java
```diff
@@ -61,6 +61,12 @@
 @Inherited
 public @interface Execution {
 
+	/**
+	 * Default value for {@value #DEFAULT_EXECUTION_MODE_PROPERTY_NAME} is {@value}.
+	 */
+	@API(status = MAINTAINED, since = "6.2.0")
```
marcphilipp: We usually only include major and minor in case the patch part is 0.

```suggestion
	@API(status = MAINTAINED, since = "6.2")
```

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationMetaData.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * See <a href="https://docs.spring.io/spring-boot/specification/configuration-metadata/format.html">Spring Boot - Specifications - Configuration Metadata - Metadata Format</a>.
+ */
+final class ConfigurationMetaData {
```
marcphilipp: For consistency?

```suggestion
final class ConfigurationMetadata {
```

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationMetaData.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	record Deprecation( //
+			@Nullable Level level, //
+			@Nullable String reason, //
+			@Nullable String replacement, //
+			@Nullable String since //
+	) {
+
+		enum Level {
```
marcphilipp: Is this used anywhere? Deprecation is always created with `null`, isn't it?

junit-platform-configuration-processor/src/main/resources/META-INF/gradle/incremental.annotation.processors
marcphilipp, line 1: According to https://docs.gradle.org/current/userguide/java_plugin.html#sec:example_registering_incremental_annotation_processors_dynamically, incremental annotation processors can only read `CLASS` or `RUNTIME` retention annotations. Therefore, I think we should delete this file.

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationParameterHandler.java
mpkorstanje: No. Because then the earlier `defaultValues` would be empty. But I suppose it doesn't hurt for consistency.

documentation/modules/ROOT/pages/advanced-topics/configuration-parameter-documentation.adoc
mpkorstanje: Done. 990b117d0aebe6a1345f7cb844d00aa45c914bf7

documentation/src/test/java/example/ConfigurationParametersDemo.java
mpkorstanje: Done. 747779b8e1ce72af21db8979fce9705360f70824

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationParameterHandler.java
mpkorstanje: Done. d110ac3ec9895ba0e5c36bac752ec9dd568f08f1

junit-platform-engine/src/main/java/org/junit/platform/engine/support/hierarchical/DefaultParallelExecutionConfigurationStrategy.java
mpkorstanje: I copied this from the `EnumConfigurationParameterConverter`. It might not need it either then. Or at least not in every context.

mpkorstanje: Done. e8ba0dfd86a5cd7f7dcd4c5113e4fdbf7940ebe9

junit-platform-configuration-processor/src/main/resources/META-INF/gradle/incremental.annotation.processors
mpkorstanje: Done. 422988821ce6e775542a9042cc0500deb9054f0b

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationMetaData.java
mpkorstanje: Done. 572e59541f5b913531a6674dbb841fd132f1c5bb

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/ConfigurationMetaData.java
mpkorstanje: Done. a80336052c63f82b1ae2935e3c2d1c5a487caaa9

junit-jupiter-api/src/main/java/org/junit/jupiter/api/parallel/Execution.java
mpkorstanje: Done. 186c38e4fb24d5998756077d55fa6ea388ca4b5d

junit-platform-configuration-processor/src/main/java/org/junit/platform/configuration/processor/JsonConverter.java
mpkorstanje: Mistaken assumption. Done. 74fb180e15dfc8253e564ae3433784b0127817de

junit-jupiter-api/src/main/java/org/junit/jupiter/api/Timeout.java
```diff
@@ -297,27 +297,39 @@
...
 	 *
-	 * <h4>Supported timeout mode values (case insensitive):</h4>
+	 * <h4>Supported timeout mode values (case-insensitive):</h4>
 	 * <ul>
-	 * <li>{@code ENABLED}: enables timeouts
-	 * <li>{@code DISABLED}: disables timeouts
-	 * <li>{@code DISABLED_ON_DEBUG}: disables timeouts while debugging
+	 * <li>{@code enabled}: enables timeouts
+	 * <li>{@code disabled}: disables timeouts
+	 * <li>{@code disabled_on_debug}: disables timeouts while debugging
```
mpkorstanje: While working on https://github.com/junit-team/junit-framework/pull/6059 I realized that we should use upper case by by default while still supporting lowercase (instead of the other way around).

Otherwise we will render the hints in lower case and thus force every test engine author to support lowercase.

mpkorstanje: Done 3a8c32266a42c36c8b1ab90e32f2ff7754a4c437.

marcphilipp (approved):

marcphilipp:
Excellent work! :+1:

PR #6001, opened 2026-08-21
Update to NullAway 0.14.1 and fix new warnings

msridhar:
NullAway 0.14.0 release notes: https://github.com/uber/NullAway/releases/tag/v0.14.0
NullAway 0.14.1 release notes: https://github.com/uber/NullAway/releases/tag/v0.14.1

Most of the new warnings are due to inconsistent type variable upper bounds on method overrides, which are easily fixed.

In `Arguments.java` there is a new warning due to an inference limitation in NullAway related to wildcards.  This limitation is actually fixed when the `JSpecifyExperimental` flag is enabled (see [release notes](https://github.com/uber/NullAway/releases/tag/v0.14.0)), but that flag causes many other new warnings.  So for now, we add an explicit type argument and a comment.

---

### DIFF
--- a/documentation/src/test/java/example/CustomLauncherInterceptor.java
+++ b/documentation/src/test/java/example/CustomLauncherInterceptor.java
@@ -18,6 +18,7 @@ import java.net.URI;
 import java.net.URL;
 import java.net.URLClassLoader;
 
+import org.jspecify.annotations.Nullable;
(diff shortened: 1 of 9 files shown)


mpkorstanje (approved):
Cheers! Much appreciated.

PR #6010, opened 2026-08-26
Document that extensions should be stateless

### LINKED ISSUE #6004: Document how many instances of an extension class are created
Affects 6.1.3, probably all older versions as well.

https://docs.junit.org/6.1.3/extensions/registering-extensions.html documents how to create an extension instance, but it does not document instance reuse.
E.g. if you create an extension for parameters, and want to keep around some data for possible later error messages, you need to know whether you record the associated class, method, or the parameter index as well as the value itself.

I believe (but have not fully validated) that it depends on where the extension is registered:
- For class extensions, there is on instance per test class.
- For method extensions, one instance per test method.
- For parameter extensions, one instance per (surprisingly but usefully) test method.
- I don't know if it is possible to have an extension that exists just once per test run. It could sure be useful to have that, e.g. to output some background information once after it has been determined if a specific situation happened at least once.

I think this is a bit of a complication for extensions that wish to have slightly different semantics depending on whether they are registered globally, on a class, on a method, or on a parameter.
I may be barking up the wrong tree here, as it is entirely possible I'm fundamentally misunderstanding the extension instances' lifecycles.

JonnasFigueiredo:
Closes #6004.

The "Registering Extensions" page explains how to register extensions but does not mention how often an extension is instantiated or that a single instance is typically reused across tests, which led to the confusion in #6004.

As @mpkorstanje noted there, it shouldn't matter how often an extension is created — state should be kept in the `ExtensionContext.Store`, not in the extension instance. This adds a short "Extension Instances and State" section to `registering-extensions.adoc` making that explicit and linking to "Keeping State in Extensions".

### DIFF
--- a/documentation/modules/ROOT/pages/extensions/registering-extensions.adoc
+++ b/documentation/modules/ROOT/pages/extensions/registering-extensions.adoc
@@ -492,3 +492,13 @@ programmatically via `@RegisterExtension` or `@ExtendWith` on fields in a subcla
 NOTE: A specific extension implementation can only be registered once for a given
 extension context and its parent contexts. Consequently, any attempt to register a
 duplicate extension implementation will be ignored.
+
+[[registration-instances]]
+== Extension Instances and State
+
+You should not make any assumptions about when or how often the JUnit Platform instantiates
+a registered extension. In particular, a single extension instance is typically reused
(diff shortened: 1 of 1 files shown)


marcphilipp (approved):

marcphilipp:
@JonnasFigueiredo Thank you for your contribution! :+1:

Issue #5292, opened 2026-01-18
Support CompletionStage friendly method signature


rmannibucau:
It is common to have promise friendly code - and virtual threads will not change that so much, at least for libs.

And all the time in code we do `promise.get()` to be able to comply to junit signature support.

Request is to enable to return a promise (`@Test CompletionStage<?> myTest() { ... }`).

This will enable to define the test code as most of the code base with  promise related API and not have to block randomly sometimes leveraging better the CPU in concurrent mode.

Then junit runtome will try/catch around the test method as today but if result is a promise it will await it using `whenComplete` it to check if there is a failure or not.

The framework will guarantee all promises are done before exiting - as today implicitly.

mpkorstanje:
> It is common to have promise friendly code - and virtual threads will not change that so much, at least for libs.

From your description I get the impression that the test would be executed inside the promise. To avoid any ambiguity, could you provide a simple but non-trivial example of what this would look like?

rmannibucau:
```java
@Test
CompletionStage<?> mytest(MyTestHelper helper, MyInjectedService service) {
    return helper
       .init()
        .thenCompose(data -> service.compute(new Request(data.getXxx()))
        .thenApply(result -> assertAll( .... ));
}
```

so yes the "given, when, then" pattern is actually a chain of promises but as soon as you have > 1 promise it is worse it (previous snippet has only 2 levels but in practise projects embrassing completionstage pattern often make it even more composed).

Indeed having an `assertAsync` or alike can work but will have the drawback to block so still not leverage the machine CPU accurately whereas giving it back to the caller (framework/runtime) it enables to make it run faster.

marcphilipp:
> It is common to have promise friendly code - and virtual threads will not change that so much, at least for libs.
> 
> And all the time in code we do `promise.get()` to be able to comply to junit signature support.

I haven't come across this testing style in the wild. Is there an open source project that does this?

> Then junit runtome will try/catch around the test method as today but if result is a promise it will await it using `whenComplete` it to check if there is a failure or not.

This is similar to the recent addition of coroutine support for Kotlin. There we wrap the `suspend` function in `runBlocking` to wait for it to complete before proceeding with the enclosing container (usually the test class and its `@AfterAll` etc. behavior).

We could do something like this if we decide to support `CompletionStage`-returning methods:

```java
CompletionStage<?> returnValue = // call to test method via reflection
try {
	returnValue.toCompletableFuture().join()
[...]

PR #6012, opened 2026-08-26
Enable JDK promise support and start to support reactive pipeline all the way

rmannibucau:
proposed solution to https://github.com/junit-team/junit-framework/issues/5292

---

---

### Definition of Done

- [X] Public API has [Javadoc](https://github.com/junit-team/junit-framework/blob/HEAD/CONTRIBUTING.md#javadoc) and [`@API` annotations](https://apiguardian-team.github.io/apiguardian/docs/current/api/org/apiguardian/api/API.html)
- [X] Change is documented in the [User Guide](https://docs.junit.org/snapshot/) and [Release Notes](https://docs.junit.org/snapshot/release-notes.html)

### DIFF
--- a/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/support/AsyncReturnTypeSupport.java
+++ b/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/support/AsyncReturnTypeSupport.java
@@ -1,0 +1,258 @@
+/*
+ * Copyright 2015-2026 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+
+package org.junit.jupiter.engine.support;
+
+import static java.util.Collections.synchronizedMap;
+import static org.apiguardian.api.API.Status.INTERNAL;
+
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Future;
+
+import org.apiguardian.api.API;
+import org.jspecify.annotations.Nullable;
+import org.junit.jupiter.api.extension.AsyncReturnValueHandler;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.Extension;
+import org.junit.jupiter.engine.extension.ExtensionRegistry;
+import org.junit.platform.commons.util.AnnotationUtils;
+import org.junit.platform.commons.util.ExceptionUtils;
+import org.junit.platform.commons.util.LruCache;
+import org.junit.platform.commons.util.Preconditions;
+
+/**
+ * {@code AsyncReturnTypeSupport} provides support for test methods that return
+ * an asynchronous completion signal.
+ *
+ * <p>Returning a {@link CompletionStage}, {@link CompletableFuture}, or
+ * {@link Future} from a {@code @Test} method is interpreted as a promise that
+ * the test's asynchronous work has terminated; its payload (if any) is ignored.
+ * Only the termination itself is awaited, so the return type is treated as an
+ * opaque signal.
+ *
+ * @since 6.2
+ */
+@API(status = INTERNAL, since = "6.2", consumers = "org.junit.jupiter.engine")
+public final class AsyncReturnTypeSupport {
+
+	private AsyncReturnTypeSupport() {
+		/* no-op */
+	}
+
+	/**
+	 * Caches, per test {@link Method}, whether it has an asynchronous return
+	 * type. This avoids invoking user-provided
+	 * {@link AsyncReturnValueHandler#supports(Type, AnnotatedElement)
+	 * handlers} and re-walking {@link ExtendWith @ExtendWith} annotations over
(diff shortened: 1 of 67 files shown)


jupiter-tests/src/test/java/org/junit/jupiter/engine/descriptor/CompletionStageTestMethodsTests.java
```diff
@@ -0,0 +1,153 @@
...
+			return CompletableFuture.failedFuture(new IllegalStateException("async boom"));
+		}
+	}
+
+	static class LifecycleCallbackTestCase {
+
+		static final AtomicInteger beforeEachCount = new AtomicInteger();
+		static final AtomicInteger afterEachCount = new AtomicInteger();
+
+		@BeforeEach
```
FroMage, line 116: Are lifecycle methods that return an async type also supported?

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/support/AsyncReturnTypeSupport.java
```diff
@@ -0,0 +1,87 @@
...
+	 * asynchronous completion signal.
+	 *
+	 * @param type the (raw) return type; never {@code null}
+	 * @return {@code true} if the type is a supported asynchronous signal
+	 * @throws org.junit.platform.commons.PreconditionViolationException if
+	 * {@code type} is {@code null}
+	 */
+	public static boolean isFullySupported(Class<?> type) {
+		Preconditions.notNull(type, "type must not be null");
+		return CompletionStage.class.isAssignableFrom(type) || Future.class.isAssignableFrom(type);
```
FroMage, line 80: The problem with this is that it artificially limits the usable async types. We don't use those types in Quarkus, we use `Uni`, in Vert.x they use a different `Future`, in Spring they use `Mono`…

All of which can be converter to a `CompletionStage` easily, but this PR is missing the SPI to achieve it.

Otherwise every test will have to end with `.convertToCompletionStage()` which is just unnecessary boilerplate for users.

FroMage (commented):
This is much larger than my proposal, and more complete in terms of supporting async types for parallel execution.

Unlike my proposal though, it does not support pluggable async types, so as such it is not very reactive-framework-friendly.

It also does not support a pluggable way to offload the original test method onto the proper reactive context (the event loop), but I have to investigate if this could be achieved by a JUnit interceptor (depends if they have access to the method return value, I guess).

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/support/AsyncReturnTypeSupport.java
rmannibucau: yes, my approach is that it is less critical to have to add a call which doesn't break the execution model than to not be able to use a programming model at all but agree we'll need to enable what you PR goal is. I just didn't want to do it before the API/SPI is fully reactive friendly cause it is likely it misses the line otherwise (it is already a lot of rework in this PR to say it differently).

does it sound ok to finish this one then add "your" feature on top - and I totally agree it will be better but what I'm not sure today is that a new API will be needed, I hope we can make it functional with interceptors or something existing even if totally unsure yet?

jupiter-tests/src/test/java/org/junit/jupiter/engine/descriptor/CompletionStageTestMethodsTests.java
rmannibucau: Very good point and no I missed it, will work on that

junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/support/AsyncReturnTypeSupport.java
FroMage: Well, if you don't introduce a new API/SPI, you'll have to use converting libraries such as https://github.com/smallrye/smallrye-reactive-converters/tree/main/reactive-converters which would be fine for us.

rmannibucau:
up?

marcphilipp:
@rmannibucau Thanks for the PR! Since #5292 is waiting for additional interest from the community, reviewing this PR is not a priority for us at the moment.

FroMage:
@rmannibucau I just tested it and it does work for us. It's a little less than what I provided in my PR in the sense that we can't control the execution of the test method to offload it to another thread (we can only convert its return value), but we can achieve that offloading via another sort of interceptor, so we can work around it.

rmannibucau:
@FroMage 

> we can't control the execution of the test method to offload it to another thread

my idea was that the promise controls the thread so it was built-in by design, did I miss sthg?

FroMage:
As far as I can see, I can only control what happens in the `CompletionStage`, which means the test method is still executed outside of my control. There's two phases happening in the test method: the test method itself, which creates the `CompletionStage`, and the body of the `CompletionStage` which is executed later. I can only control the second half, not the first.

But again, I can work around that.

The other thing I can think of is that `CompletionStage` is eager, so converting from our promise type to it will trigger subscription and execution, but again, in this case I don't think it's an actual issue.

rmannibucau:
@FroMage yep but the first one is the trigger which must not block (else it is pointless to use this programming model) so second is sufficient from my view and is always workaround-able using supplyAsync(,pool)

FroMage:
It's not just about blocking: in our case the creation of these async tasks must happen on an event loop, so we need to redirect the execution of the test method on the event loop, because we can start making those promises. Anyway, as I said, we can work around it using other kinds of interceptors. The biggest hurdle was the method return type being disallowed.

rmannibucau:
@FroMage the point is more that you can always do it in a promise or using an injected helper which does facade a queue push/pop using completionstages. That said I get it but also brings the context in the loop issue generally speaking so maybe your interceptor (or a parameter/field injection) is saner :thinking: . Anyway if this PR already enables your case I see it as  a baby step forward ;)

FroMage:
Yeah, as I said, at this point, both my PR (less complete, but smaller) or yours achieve our goals, so I'm good with either one, I just hope we can get any merged soon so I don't have to commit atrocious classloader mocks to workaround this 😬

PR #6045, opened 2026-09-09
Use iterators to compare equally sized lines

JunggiKim:
For equally sized lists, `AssertLinesMatch` compared every line through indexed
access. This repeatedly walks a `LinkedList` to reach each index.

Use each list's iterator instead. The equal-size check is a precondition of that
comparison, so it now lives inside `matchesLineByLine()` rather than at the call
site. The comparison order, first-mismatch behavior, and existing fast-forward
fallback remain unchanged.

---

---

### Definition of Done

- [x] Existing preconditions and Javadoc remain unchanged
- [x] Existing automated tests cover the affected line-matching behavior
- [x] No public API is affected
- [x] No User Guide or Release Notes change is needed

### Verification

```text
./gradlew :jupiter-tests:test --tests org.junit.jupiter.api.AssertLinesMatchAssertionsTests \
  -Pjunit.develocity.predictiveTestSelection.enabled=false --no-scan --console=plain

./gradlew :junit-jupiter-api:checkstyleMain \
  -Pjunit.develocity.predictiveTestSelection.enabled=false --no-scan --console=plain

./gradlew spotlessCheck -Pjunit.develocity.predictiveTestSelection.enabled=false \
  --no-scan --console=plain
```

All commands completed successfully (28 tests executed, 0 failures).


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
```diff
@@ -98,7 +97,7 @@ void assertLinesMatch() {
 
 			// simple case: both list are equally sized, compare them line-by-line
 			if (expectedSize == actualSize) {
-				if (IntStream.range(0, expectedSize).allMatch(i -> matches(expectedLines.get(i), actualLines.get(i)))) {
+				if (matchesLineByLine()) {
```
mpkorstanje: The `expectedSize == actualSize` is a precondition for the correct functioning of `matchesLineByLine` and shouldn't be separated from it.

The `expectedSize == actualSize` can be moved into `matchesLineByLine`.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
```diff
@@ -107,6 +106,17 @@ void assertLinesMatch() {
 			assertLinesMatchWithFastForward();
 		}
 
+		private boolean matchesLineByLine() {
+			var expectedIterator = expectedLines.iterator();
+			var actualIterator = actualLines.iterator();
+			while (expectedIterator.hasNext()) {
+				if (!matches(expectedIterator.next(), actualIterator.next())) {
```
sormuras, line 115:
```suggestion
        assert actualIterator.hasNext(); // ensure same-sized lists
				if (!matches(expectedIterator.next(), actualIterator.next())) {
```

sormuras: Either add this assertion or change the name of the method to reflect the "same-sizeness" of both lists, please.

sormuras (changes requested):
Looks good and more efficient in general; minor nit about the method name.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
sormuras: Or like this, which is even better.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
marcphilipp: @sormuras We don't use `assert` anywhere, do we?

JunggiKim: @sormuras @marcphilipp Thanks for the pointers!

I checked the codebase as suggested and confirmed that the `assert` keyword is not used anywhere under `src/main`, so I dropped that option.

That left the rename, but @mpkorstanje's suggestion of moving the equal-size check into the method — which you agreed was even better — removes the need for it: `matchesLineByLine()` now verifies the sizes itself and returns `false` when they differ, which falls through to the fast-forward matching exactly as before. Since the "same-sizeness" is now part of the method's own contract rather than an unstated precondition, I kept the name as is. Happy to rename it if you would still prefer something more explicit.

Applied in 71f7a09.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
JunggiKim: @mpkorstanje Good point — the caller could have broken the method without any signal. Moved the equal-size check into `matchesLineByLine()` in 71f7a09, so the method no longer depends on its call site. Thanks!

sormuras (approved):

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertLinesMatch.java
mpkorstanje: Please, take it easy on the AI generated writing.

marcphilipp (approved):

marcphilipp:
@JunggiKim Thank you for your contribution! :+1:

PR #6057, opened 2026-09-09
Assert timeout with sub millisecond precision

### LINKED ISSUE #6053: Timeout assertions lose sub-millisecond precision
## Steps to reproduce

The relevant implementation is unchanged between JUnit Jupiter 6.1.3 and
current `main` (`17ce9f1d4fd65fb01f5f23c87c3daecf7a185c94`).

Run the following test with JUnit Jupiter:

```java
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertTimeout;

import java.time.Duration;

class AssertTimeoutSubMillisecondTests {

	@Test
	void rejectsEveryOverrun() {
		int falsePasses = 0;
		for (int i = 0; i < 100; i++) {
			try {
				assertTimeout(Duration.ofNanos(500_000), () -> spinForNanos(700_000));
				falsePasses++;
			}
			catch (AssertionError expected) {
			}
		}

		assertEquals(0, falsePasses,
				"each 700 us execution exceeds the 500 us timeout");
	}

	private static void spinForNanos(long nanos) {
		long end = System.nanoTime() + nanos;
		while (System.nanoTime() < end) {
			Thread.onSpinWait();
		}
	}
}
```

Expected: every invocation of `assertTimeout()` rejects the overrun, so
`falsePasses` is zero.

Actual: invocations whose two wall-clock reads fall in the same millisecond
return normally. The exact count is timing-dependent. On revision
`f10989d3073dc0c6e3a1af4426fbd68ec4c119e4`, the test above observed 27 false
passes out of 100; a separate 50 us / 1 ns probe observed 94 out of 100.

[`AssertTimeout`](https://github.com/junit-team/junit-framework/blob/17ce9f1d4fd65fb01f5f23c87c3daecf7a185c94/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java#L67-L87)
converts the requested `Duration` with `toMillis()` and measures elapsed time
using `System.currentTimeMillis()`. Thus every positive timeout below one
millisecond becomes zero, elapsed time is quantized to milliseconds, and
wall-clock adjustments can affect the result.

The preemptive path has the inverse symptom.
[`PreemptiveTimeoutUtils`](https://github.com/junit-team/junit-framework/blob/17ce9f1d4fd65fb01f5f23c87c3daecf7a185c94/junit-jupiter-api/src/main/java/org/junit/jupiter/api/timeout/PreemptiveTimeoutUtils.java#L90-L105)
calls `Future.get(timeout.toMillis(), MILLISECONDS)`, so all positive values
from 1 to 999,999 ns become the same zero-length poll. Whether a particular
no-op invocation succeeds is scheduler-dependent; the deterministic defect is
the conversion to `Future.get(0, MILLISECONDS)`, not a particular failure rate.

This utility is used by both `assertTimeoutPreemptively()` and
[`SeparateThreadTimeoutInvocation`](https://github.com/junit-team/junit-framework/blob/17ce9f1d4fd65fb01f5f23c87c3daecf7a185c94/junit-jupiter-engine/src/main/java/org/junit/jupiter/engine/extension/SeparateThreadTimeoutInvocation.java#L33-L38),
so `@Timeout(..., threadMode = SEPARATE_THREAD)` has the same wait-precision
issue. Same-thread `@Timeout` uses a different path.

Assertion failures also render a positive sub-millisecond timeout as `0 ms`.

Expected behavior is to honor the supplied `Duration` precision and to use a
monotonic elapsed-time clock for non-preemptive assertions.

Compatibility points that should be decided explicitly:

- nanosecond elapsed-time measurement will newly detect overruns of an
  integer-millisecond timeout that are smaller than one millisecond;
- `Duration.toNanos()` overflows much earlier than the current `toMillis()`
  conversion, so a direct replacement would regress very large durations; and
- zero and negative `Duration` values are currently not rejected or covered by
  tests, so their behavior should be preserved deliberately or specified.

I searched open and closed issues and pull requests for combinations of
`assertTimeout`, `toMillis`, `currentTimeMillis`, `nanoTime`, `nanoseconds`,
`sub-millisecond`, `precision`, `truncation`, and `Future.get`. I did not find
a directly matching public report. The closest result, #2239, concerns access
to the worker `Future`/executor for diagnostics rather than timeout precision.

## Context

 - Used versions (Jupiter/Vintage/Platform): JUnit Jupiter 6.1.3; also current 6.2.0-SNAPSHOT `main` at `17ce9f1d4fd65fb01f5f23c87c3daecf7a185c94`
 - Build Tool/IDE: JUnit repository Gradle build; JDK 25; Windows 11

This is not known to be a regression; the same implementation is present in
the inspected history.

## Deliverables

- [ ] Preserve sub-millisecond precision in `assertTimeout()` and the shared preemptive timeout path.
- [ ] Use a monotonic clock to measure elapsed time in `assertTimeout()`.
- [ ] Report positive sub-millisecond timeout values without displaying `0 ms`.
- [ ] Add deterministic regression tests for elapsed-duration comparison and the `(value, TimeUnit)` passed to `Future.get()`.
- [ ] Cover whole-millisecond boundaries, very large durations, and the chosen zero/negative-duration behavior.
- [ ] Document the compatibility change if accepted.

mpkorstanje:
Measure in nanoseconds the execution time of the `supplier` supplied to `assertTimeout`, `assertTimeoutPreemptively` and methods annotated with `@Timeout`.

Additionally, limit the timeout provided to both methods and annotation to a positive total duration less than 2<sup>63</sup> nanoseconds (approximately 292 years).

While this limitation may cause a test suite to fail, non-positive timeout values always result in an assertion failure so we wouldn't expect to see this used in reasonable test scenarios. Likewise, asserting that a method completes with in more than 292 years seems equally unreasonable.

Note: The `@Timeout` annotation already restricted the duration to positive values. 

Closes: #6053

---

---

### Definition of Done

### DIFF
--- a/documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
+++ b/documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
@@ -62,7 +62,9 @@ repository on GitHub.
   quotes.
 * When using `@CsvSource.value[]` the `commentCharacter` must be distinct from both
   `delimiter` and `quoteCharacter`.
-
+* When the timeout provided to `assertTimeout`, `assertTimeoutPreemptively` and test
+  methods annotated with `@Timeout` is not a positive value less than 2^63^ nanoseconds
+  (approximately 292 years) a `PreconditionViolationException` is thrown.
 
 [[v6.2.0-M1-junit-jupiter-new-features-and-improvements]]
 ==== New Features and Improvements
@@ -71,7 +73,8 @@ repository on GitHub.
   and `@DisabledOnOs`.
 * Failures caused by `@Timeout` expirations now include a hint about enabling
   xref:writing-tests/timeouts.adoc#debugging-thread-dump[thread dumps].
-
+* The `assertTimeout`, `assertTimeoutPreemptively` and methods annotated with `@Timeout`
+  now measure the execution time with sub-millisecond precision.
 
 [[v6.2.0-M1-junit-vintage]]
 === JUnit Vintage
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
@@ -10,7 +10,9 @@
 
 package org.junit.jupiter.api;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.junit.jupiter.api.AssertionFailureBuilder.assertionFailure;
+import static org.junit.jupiter.api.timeout.TimeoutUtils.isRepresentableInNanos;
 import static org.junit.platform.commons.util.ExceptionUtils.throwAsUncheckedException;
 
 import java.time.Duration;
@@ -19,6 +21,7 @@ import java.util.function.Supplier;
 import org.jspecify.annotations.Nullable;
 import org.junit.jupiter.api.function.Executable;
 import org.junit.jupiter.api.function.ThrowingSupplier;
+import org.junit.platform.commons.util.Preconditions;
 
 /**
  * {@code AssertTimeout} is a collection of utility methods that support asserting
@@ -66,8 +69,12 @@ class AssertTimeout {
 
 	private static <T extends @Nullable Object> T assertTimeout(Duration timeout, ThrowingSupplier<T> supplier,
 			@Nullable Object messageOrSupplier) {
-		long timeoutInMillis = timeout.toMillis();
-		long start = System.currentTimeMillis();
+		Preconditions.notNull(timeout, () -> "timeout must not be null");
+		Preconditions.condition(!timeout.isNegative() && !timeout.isZero(), () -> "timeout must be positive");
+		Preconditions.condition(isRepresentableInNanos(timeout),
+			() -> "timeout must be less than approximately 292 years (2^63 nanoseconds)");
+
+		long start = System.nanoTime();
 		T result;
 		try {
 			result = supplier.get();
@@ -75,17 +82,34 @@ class AssertTimeout {
 		catch (Throwable ex) {
 			throw throwAsUncheckedException(ex);
 		}
-
-		long timeElapsed = System.currentTimeMillis() - start;
-		if (timeElapsed > timeoutInMillis) {
+		// Creates a false positive if the tests runs for more than 292 years.
+		var timeElapsed = Duration.ofNanos(System.nanoTime() - start);
+		if (timeElapsed.compareTo(timeout) > 0) {
 			assertionFailure() //
 					.message(messageOrSupplier) //
-					.reason("execution exceeded timeout of " + timeoutInMillis + " ms by "
-							+ (timeElapsed - timeoutInMillis) + " ms") //
+					.reason(createExecutionExceededTimeoutMessage(timeout, timeElapsed)) //
 					.trimStacktrace(Assertions.class) //
 					.buildAndThrow();
 		}
 		return result;
+	}
+
+	private static String createExecutionExceededTimeoutMessage(Duration timeout, Duration timeElapsed) {
+		var significantNanoFraction = timeout.toNanos() - MILLISECONDS.toNanos(timeout.toMillis()) != 0;
+		var timeoutExceeded = timeElapsed.minus(timeout);
+		boolean timeoutExceededOnlyByNanoSeconds = timeoutExceeded.toMillis() == 0;
+		return "execution exceeded timeout of %s by %s" //
+				.formatted(formatDuration(timeout, significantNanoFraction), //
+					formatDuration(timeoutExceeded, significantNanoFraction || timeoutExceededOnlyByNanoSeconds));
+	}
+
+	private static String formatDuration(Duration duration, boolean includeNanoSeconds) {
+		long milliseconds = duration.toMillis();
+		if (!includeNanoSeconds) {
+			return "%d ms".formatted(milliseconds);
+		}
+		long nanoFraction = duration.toNanosPart() - MILLISECONDS.toNanos(milliseconds);
+		return "%d.%06d ms".formatted(milliseconds, nanoFraction);
 	}
 
 }
--- a/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeoutPreemptively.java
+++ b/junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeoutPreemptively.java
@@ -10,6 +10,7 @@
 
 package org.junit.jupiter.api;
 
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.junit.jupiter.api.AssertionFailureBuilder.assertionFailure;
 import static org.junit.jupiter.api.timeout.PreemptiveTimeoutUtils.executeWithPreemptiveTimeout;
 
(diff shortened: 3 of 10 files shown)


junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
```diff
@@ -66,26 +69,45 @@ static void assertTimeout(Duration timeout, Executable executable, Supplier<@Nul
...
 		}
 		catch (Throwable ex) {
 			throw throwAsUncheckedException(ex);
 		}
-
-		long timeElapsed = System.currentTimeMillis() - start;
-		if (timeElapsed > timeoutInMillis) {
+		// Creates a false positive if the tests runs for more than 292 years.
+		var timeElapsed = Duration.ofNanos(System.nanoTime() - start);
+		if (timeElapsed.compareTo(timeout) > 0) {
```
mpkorstanje, line 87: In theory, because we're using `Duration::compareTo`, the timeout does not have to be represntable in nanos.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/timeout/PreemptiveTimeoutUtils.java
```diff
@@ -92,7 +97,7 @@ public class PreemptiveTimeoutUtils {
 			Supplier<@Nullable Thread> threadSupplier, TimeoutFailureFactory<E> failureFactory)
 			throws E, RuntimeException {
 		try {
-			return future.get(timeout.toMillis(), TimeUnit.MILLISECONDS);
+			return future.get(timeout.toNanos(), TimeUnit.NANOSECONDS);
```
mpkorstanje, line 102: We could check the significant digits and use a different timeout. But with 272 years available, I don't think that is worth the effort.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/timeout/PreemptiveTimeoutUtils.java
```diff
@@ -61,6 +62,10 @@ public class PreemptiveTimeoutUtils {
 			ThrowingSupplier<T> supplier, @Nullable Supplier<@Nullable String> messageSupplier,
 			TimeoutFailureFactory<E> failureFactory) throws E {
 
+		Preconditions.condition(!timeout.isNegative(), () -> "timeout must be non-negative");
+		DurationPreconditions.representableInNanos(timeout,
+			() -> "timeout must be less than approximately 292 years (2^63 nanoseconds)");
```
mpkorstanje: This check could be pushed up, but the fan-out is huge.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
marcphilipp: If we check it here, we should also check it in `TimeoutDuration` for consistency.

documentation/modules/ROOT/partials/release-notes/release-notes-6.2.0-M1.adoc
```diff
@@ -62,7 +62,9 @@ repository on GitHub.
   quotes.
 * When using `@CsvSource.value[]` the `commentCharacter` must be distinct from both
   `delimiter` and `quoteCharacter`.
-
+* When the timeout provided to `assertTimeout`, `assertTimeoutPreemptively` and test
+  methods annotated with `@Timeout` is not a positive value less than 2^63^ nanoseconds
```
marcphilipp, line 66: TIL that `2^63^` is Asciidoc syntax.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeout.java
```diff
@@ -66,26 +69,47 @@ static void assertTimeout(Duration timeout, Executable executable, Supplier<@Nul
...
+				.formatted(formatDuration(timeout, significantNanoFraction), //
+					formatDuration(timeoutExceeded, significantNanoFraction || timeoutExceededOnlyByNanoSeconds));
+	}
+
+	private static String formatDuration(Duration duration, boolean includeNanoSeconds) {
+		long milliseconds = duration.toMillis();
+		if (!includeNanoSeconds) {
+			return "%d ms".formatted(milliseconds);
+		}
+		long nanoFraction = duration.toNanosPart() - MILLISECONDS.toNanos(milliseconds);
```
marcphilipp:
```suggestion
		long nanoFraction = duration.toNanos() - MILLISECONDS.toNanos(milliseconds);
```

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeoutPreemptively.java
```diff
@@ -71,12 +72,21 @@ private static AssertionFailedError createAssertionFailure(Duration timeout,
...
 	}
 
+	private static String formatDuration(Duration duration) {
+		long milliseconds = duration.toMillis();
+		long nanoFraction = duration.toNanos() - MILLISECONDS.toNanos(milliseconds);
+		if (nanoFraction == 0) {
+			return "%d ms".formatted(milliseconds);
+		}
+		return "%d.%06d ms".formatted(milliseconds, nanoFraction);
+	}
```
marcphilipp: Can we extract a package-private helper function to remove the duplication with `AssertTimeout`?

jupiter-tests/src/test/java/org/junit/jupiter/engine/extension/TimeoutDurationTests.java
```diff
@@ -40,4 +46,29 @@ void fulfillsEqualsAndHashCodeContract() {
...
+
+		@Test
+		@SuppressWarnings("DataFlowIssue")
+		void nonNullUnit() {
+			assertPreconditionViolationNotNullFor("timeout unit", () -> new TimeoutDuration(1, null));
+		}
+
+		@Test
+		void representableInNanos() {
+			var maxNanoRepresentableDays = Duration.ofNanos(Long.MAX_VALUE).toDays();
```
marcphilipp: What about `new TimeoutDuration(Long.MAX_VALUE, DAYS)`. If I'm reading the code correctly, that would overflow in `toDuration()` before `isRepresentableInNanos` is even called.

marcphilipp (commented):
I think we should also update Javadoc for `assertTimeout{Preemptively}` and `@Timeout` to document the valid range.

jupiter-tests/src/test/java/org/junit/jupiter/engine/extension/TimeoutDurationTests.java
mpkorstanje: Should work fine. `Duration` tracks time internally using seconds and nano seconds. So we got `Duration.ofSeconds(Long.MAX_VALUE)` -1 nano second before anything overflows. Adding a test doesn't hurt though.

mpkorstanje: Okay. I see the problem.

mpkorstanje: Done ec2f234244f9c1001807a3e7fc12e361e889f2a3.

junit-jupiter-api/src/main/java/org/junit/jupiter/api/AssertTimeoutPreemptively.java
mpkorstanje: Mmh. It's not pretty but 2d2e2717983ed10f23aa5f6c3fdb8c1b7380af38.

mpkorstanje:
> I think we should also update Javadoc for assertTimeout{Preemptively} and @Timeout to document the valid range.

Done. I've asserted all the preconditions for both assert timeout methods for consistency. The documentation is a bit awkward because we don't use the `@param` documentation on assertions for some reason.

9e3d7b28d82e8792ea69045c63a06209224160b and 0f627af0db35dd5cc24781b37978620f297f8687

marcphilipp (approved):

PR #6077, opened 2026-09-18
Change default executor service implementation to `WORKER_THREAD_POOL`

### LINKED ISSUE #5291: Change default executor service implementation to `worker_thread_pool`
In 6.1.0 (see #5060) the `worker_thread_pool` executor service for parallel test execution was introduced as an opt-in feature. We should swap the default in 6.2.0 unless there is negative feedback on the new implementation.

## Deliverables

- [ ] Change default from `fork_join_pool` to `worker_thread_pool`

marcphilipp:
Resolves #5291.


documentation/modules/ROOT/pages/writing-tests/parallel-execution.adoc
```diff
@@ -164,22 +164,18 @@ concurrently. You can configure which implementation of `HierarchicalTestExecuto
 is used be setting the `junit.jupiter.execution.parallel.config.executor-service`
 configuration parameter to one of the following options:
 
-`fork_join_pool` (default)::
+`worker_thread_pool` (default)::
+Use an executor service that is backed by a regular thread pool and does not create
+additional threads if test or production code uses `ForkJoinPool` or calls a blocking
+API in the JDK.
```
mpkorstanje, line 170: It might not be apparent to all readers that a `ThreadPoolExecutor` is the regular thread pool because this option now precedes the FJP. Additionally, I don't think we have to mention that this doesn't interact with the FJP. That would be the default assumption.

So something like:

```suggestion
`worker_thread_pool` (default)::
Use an executor service that is backed by a `ThreadPoolExecutor`. The
executor service tries to maintain the desired parallelism but does
not compensate for test or production code uses blocking API in the JDK.
```

mpkorstanje (changes requested):

